PR #401, opened 2024-10-11
Str 523 sync event rpcs

delbonis:
The same as the PR I just merged into main.


delbonis:
I think github is confused about what I actually intended to merge into this branch because it's including a bunch of unrelated commits.  I don't know how to resolve this with the UI and the branch protections prevent me from pushing commits directly.

PR #406, opened 2024-10-13
Fixes for STR-522

delbonis:
## Description

Something unclear is broken with the CSM stuff that's preventing us from correctly advancing the chain.  This PR is incomplete and is trying to investigate and fix those issues.

### Type of Change

-   [ ] Documentation update
-   [x] Refactor

## Checklist


delbonis:
Here's what I've done so far:

* reorganized the CSM modules a bit, split out some exec behavior so we might be able to refactor it some more in the future
* improved logging with span so that we can see from the logs more easily what the CSM is trying to do
* added sync event retry and exponential backoff so if it really is a transient issue we'll just retry it in a second or a few
* if there's ever an error processing a CSM message we quit the CSM worker, which crashes the client that we can identify more clearly instead of leaving it in some weird broken state

delbonis:
The actual issue was resolved, but these are still useful changes that are in service of longer term goals, so we should still wrap this up.

bin/strata-client/src/helpers.rs
```diff
@@ -1,3 +1,5 @@
+// TODO clean up this module
+
```
storopoli, line 2: This should be moved to an issue. I don't like TODO comments in the codebase. They are easily forgotten

bin/strata-client/src/l1_reader.rs
```diff
@@ -1,7 +1,9 @@
+// TODO clean up this module
+
```
storopoli, line 2: same

crates/consensus-logic/src/csm/config.rs
```diff
@@ -0,0 +1,27 @@
+use std::time;
+
+const RETRY_BACKOFF_BASE: u32 = 1024;
+
+/// Run-time config for CSM executor.  This is *not* like system params.
```
storopoli, line 5:
```suggestion
/// Run-time config for CSM executor. This is *not* like system params.
```

crates/consensus-logic/src/csm/ctl.rs
```diff
@@ -5,7 +5,7 @@ use strata_state::sync_event::SyncEvent;
 use tokio::sync::{mpsc, oneshot};
 use tracing::*;
 
-use crate::message::CsmMessage;
+use super::message::CsmMessage;
```
storopoli, line 8: This change is not needed

crates/consensus-logic/src/csm/mod.rs
```diff
@@ -0,0 +1,14 @@
+//! Consensus state machine
+//!
+//! This is responsible for managing the final view of the checkpointing state,
+//! tracking unrecognized state from L1, and determining the basis for which
+//! unfinalized blocks are committed.
+// TODO clean up this module so that specific items are directly exported and
+// modules don't have to be
```
storopoli, line 7: same as above

crates/consensus-logic/src/csm/state_tracker.rs
```diff
@@ -11,7 +11,8 @@ use strata_state::{
 };
 use tracing::*;
 
-use crate::{client_transition, errors::Error};
+use super::client_transition;
+use crate::errors::Error;
```
storopoli, line 15: I don't think this change is needed. Also it is not consistent...

crates/consensus-logic/src/csm/worker.rs
```diff
@@ -0,0 +1,372 @@
+//! Consensus logic worker task.
+
+// TODO massively refactor this module
```
storopoli, line 3: same as above

crates/consensus-logic/src/csm/worker.rs
```diff
@@ -0,0 +1,372 @@
...
+    time,
+};
+use tracing::*;
+
+use super::{
+    config::CsmExecConfig,
+    message::{ClientUpdateNotif, CsmMessage, ForkChoiceMessage},
+    state_tracker,
+};
+use crate::{errors::Error, genesis};
```
storopoli, line 28: We should be consistent with these imports. Either `super` or `crate`. I prefer `crate`.

crates/consensus-logic/src/csm/worker.rs
```diff
@@ -0,0 +1,372 @@
...
+
+/// Mutable worker state that we modify in the consensus worker task.
+///
+/// Unable to be shared across threads.  Any data we want to export we'll do
+/// through another handle.
+#[allow(unused)]
+pub struct WorkerState<D: Database> {
+    /// Consensus parameters.
+    params: Arc<Params>,
```
storopoli, line 37: This is a lib I don't think we should `Arc` stuff here. Maybe fine in binaries.
What if we
```suggestion
    params: &Params,
```
and let the user define if it needs `Arc` or no in the actual impl?

crates/consensus-logic/src/csm/worker.rs
```diff
@@ -0,0 +1,372 @@
...
+
+    /// Underlying database hierarchy that writes ultimately end up on.
+    // TODO should we move this out?
+    database: Arc<D>,
+
+    /// L2 block manager.
+    l2_block_manager: Arc<L2BlockManager>,
+
+    /// Checkpoint manager.
+    checkpoint_manager: Arc<CheckpointDbManager>,
```
storopoli, line 50: same as above

crates/consensus-logic/src/csm/worker.rs
```diff
@@ -0,0 +1,372 @@
...
+
+impl<D: Database> WorkerState<D> {
+    /// Constructs a new instance by reconstructing the current consensus state
+    /// from the provided database layer.
+    pub fn open(
+        params: Arc<Params>,
+        database: Arc<D>,
+        l2_block_manager: Arc<L2BlockManager>,
+        cupdate_tx: broadcast::Sender<Arc<ClientUpdateNotif>>,
+        checkpoint_manager: Arc<CheckpointDbManager>,
```
storopoli, line 67: same as above

crates/consensus-logic/src/csm/worker.rs
```diff
@@ -0,0 +1,372 @@
...
+        })
+    }
+
+    /// Gets the index of the current state.
+    pub fn cur_event_idx(&self) -> u64 {
+        self.state_tracker.cur_state_idx()
+    }
+
+    /// Gets a ref to the consensus state from the inner state tracker.
+    pub fn cur_state(&self) -> &Arc<ClientState> {
```
storopoli, line 103: same as above

crates/consensus-logic/src/csm/worker.rs
```diff
@@ -0,0 +1,372 @@
...
+}
+
+/// Receives messages from channel to update consensus state with.
+// TODO consolidate all these channels into container/"io" types
+pub fn client_worker_task<D: Database, E: ExecEngineCtl>(
+    shutdown: ShutdownGuard,
+    mut state: WorkerState<D>,
+    engine: Arc<E>,
+    mut msg_rx: mpsc::Receiver<CsmMessage>,
+    status_tx: Arc<StatusTx>,
```
storopoli, line 120: same as above

crates/consensus-logic/src/csm/worker.rs
```diff
@@ -0,0 +1,372 @@
...
+    info!("consensus task exiting");
+
+    Ok(())
+}
+
+fn process_msg<D: Database>(
+    state: &mut WorkerState<D>,
+    engine: &impl ExecEngineCtl,
+    msg: &CsmMessage,
+    status_rx: &Arc<StatusTx>,
```
storopoli, line 157: same as above

storopoli (changes requested):
Some nits.
Correct if I am wrong, but I don't think we need to shove down the throat `Arc`s here? They should be (mutable) references and then the user decides to use `Arc` or not in the implementation, i.e. the binaries.

crates/consensus-logic/src/csm/ctl.rs
delbonis: It should be, I moved the message module too, right?

crates/consensus-logic/src/csm/state_tracker.rs
delbonis: Again this was from modules that got moved. :P

crates/consensus-logic/src/csm/worker.rs
delbonis: I prefer `super` for neighboring modules deeper in a hierarchy since ideally we'd slim down the high-level exports and those might not be accessible through `crate::`.

crates/consensus-logic/src/csm/worker.rs
delbonis: We construct it as an Arc and pass it around everywhere as one.  Plus then we'd have to expose a lifetime param.

crates/consensus-logic/src/csm/worker.rs
delbonis: What if we want to clone the state?  We never mutate it in place outside of the internal logic, so it makes sense to expose the arc.

delbonis:
Will address the other comments.

crates/consensus-logic/src/csm/mod.rs
delbonis: Created as [STR-550](https://alpenlabs.atlassian.net/browse/STR-550).

crates/consensus-logic/src/csm/worker.rs
delbonis: There's tickets for this one.

crates/consensus-logic/src/csm/worker.rs
delbonis: This was just moving code around that already exists.

PR #408, opened 2024-10-15
STR-506: test seed encryption

AaronFeickert:
## Description

This PR tests successful seed encryption and decryption, as well as failed decryption caused by  a bad passphrase or malleation.

To do this, it adds the ability to create a `Password` from a string; this is necessary since current functionality requires interactive user input, which is not suitable for automated testing.

### Type of Change

-   [ ] Documentation update
-   [ ] Refactor

## Checklist

## Related Issues

None.

### DIFF
--- a/bin/strata-cli/src/seed/password.rs
+++ b/bin/strata-cli/src/seed/password.rs
@@ -29,6 +29,13 @@ impl HashVersion {
 }
 
 impl Password {
+    /// Constructs a password from a string. (The complexity of the password is not checked.)
+    pub fn new(password: String) -> Self {
+        Self { inner: password }
+    }
+
+    /// Constructs a password from user interaction. (The complexity of the password is not
+    /// checked.)
     pub fn read(new: bool) -> Result<Self, dialoguer::Error> {
         let mut input = InputPassword::new().allow_empty_password(true);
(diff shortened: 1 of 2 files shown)


bin/strata-cli/src/seed/password.rs
```diff
@@ -29,6 +29,14 @@ impl HashVersion {
 }
 
 impl Password {
+    /// Sets a password from a string. (The complexity of the password is not checked.)
+    pub fn new(password: &str) -> Self {
+        Self {
+            inner: password.to_string(),
```
delbonis, line 35: This could take `String` directly right?  Also this would be "constructs" rather than "sets" couldn't it?

delbonis (changes requested):

bin/strata-cli/src/seed/password.rs
AaronFeickert: It certainly could, but I wanted to make it more general and not require heap allocation by the caller. Thoughts on this?

Good call on the docstrings. Reading it again now, the current wording feels awkward.

delbonis (approved):

delbonis:
Going to wait for tests to pass.

PR #410, opened 2024-10-15
STR-504: add sanity test for key generation

AaronFeickert:
## Description

It wasn't clear how the library handles generation of ECDSA signing keys when provided with bytes representing a value that overflows the curve order, since simply performing modular reduction induces bias. It turns out that this case is properly rejected, even though the documentation wasn't particularly clear that this would occur.

This PR adds a sanity test demonstrating rejection cases (and a success case, just for the heck of it).

### Type of Change

-   [ ] Documentation update
-   [ ] Refactor
-   [x] New or updated tests

## Checklist

## Related Issues

None.

### DIFF
--- a/bin/strata-cli/src/seed.rs
+++ b/bin/strata-cli/src/seed.rs
@@ -261,8 +261,24 @@ pub mod password;
 #[cfg(test)]
 mod test {
     use rand::rngs::OsRng;
+    use sha2::digest::generic_array::GenericArray;
 
     use super::*;
+
+    #[test]
+    // Sanity check on private key construction
+    fn invalid_keys() {
+        // The key can't be zero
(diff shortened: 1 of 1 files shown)


storopoli (approved):
ACK 456675d4fce78a8497ee61f534a504467f794534

Maybe I might be wrong but this is not ECDSA, but Shnorr BIP-340 signatures, since we are using Taproot in all wallet/user-facing CLI stuff, right?

If I am not mistaken, then we can change the PR description to mention that instead of ECDSA.

AaronFeickert:
Yeah, I noted ECDSA keys only to match the underlying library structure that generates them. I can change to make it more clear to the reader.

PR #419, opened 2024-10-17
Checkpoint proof verification

prajwolrg:
## Description

* Correctly update `HeaderVerificationState` in the client state
* Updates risc0 version to 1.1.2
* Introduced a new `ProofWithMetadata` that includes unique Id and remote Id
* Added the capability to run simulations within the ZKVM environment, allowing for better testing and validation of proof generation.
* Updated SP1-related tests to support the newly added simulation feature, enhancing the test coverage and reliability.
* Introduced logging functionality in SP1 ZKVM tests to provide better visibility into the test execution and debugging process.
* Moved the calculation of transaction filter rules from BTC blockspace logic to the L1 batch to improve proving cycles.

### Type of Change

-   [ ] Documentation update
-   [x] Refactor

## Checklist


crates/zkvm/adapters/sp1/src/prover.rs
```diff
@@ -28,44 +36,116 @@ impl ZKVMHost for SP1Host {
...
+                block_on(network_prover.request_proof(&self.elf, prover_input.clone(), mode))?;
+
+            let proof_data: SP1ProofWithPublicValues =
+                block_on(network_prover.wait_proof(&remote_id, None))?;
+
+            let filename: String = format!("{}.{}proof", remote_id, self.prover_options);
+            let mut file = File::create(filename).unwrap();
+            file.write_all(&bincode::serialize(&proof_data).unwrap())
+                .unwrap();
+
```
MdTeach, line 98: I think this is done for the debugging purpose only right ?

crates/state/src/l1/id.rs
```diff
@@ -54,3 +62,20 @@ impl fmt::Display for L1BlockId {
...
+    #[test]
+    fn test_l1_blkid() {
+        let block = get_btc_mainnet_block();
+        let l1_blkid: L1BlockId = block.block_hash().into();
+        // This makes sense to be equal?
+        let _str1 = format!("{}", block.block_hash());
+        let _str2 = format!("{:?}", l1_blkid);
+        // assert_eq!(str1, str2);
+    }
+}
```
MdTeach, line 81: This test looks incomplete

crates/zkvm/adapters/sp1/src/prover.rs
```diff
@@ -28,44 +36,116 @@ impl ZKVMHost for SP1Host {
...
         }
         if self.prover_options.stark_to_snark_conversion {
             prover = prover.groth16();
         }
 
-        let proof = prover.run()?;
+        let (remote_id, proof_data) = if client.prover.id() == ProverType::Network {
+            println!("Remote prover");
+            let network_prover =
+                unsafe { &*(client.prover.as_ref() as *const _ as *const NetworkProver) };
```
MdTeach, line 76: * Refactor `println!` to log
* Also adding some context for the need of `unsafe` block would be helpful.

crates/zkvm/adapters/sp1/src/prover.rs
```diff
@@ -28,44 +36,116 @@ impl ZKVMHost for SP1Host {
...
+
+            let mode = match (
+                self.prover_options.enable_compression,
+                self.prover_options.stark_to_snark_conversion,
+            ) {
+                (_, true) => ProofMode::Groth16,
+                (true, _) => ProofMode::Compressed,
+                (_, _) => ProofMode::default(),
+            };
+            println!("Prover Mode: {:?}", mode);
```
MdTeach, line 86: Refactoring `println!` to log

provers/sp1/guest-cl-agg/src/main.rs
```diff
@@ -10,7 +10,10 @@ fn main() {
         "At least one CL proof is required for aggregation"
     );
 
-    let cl_proof_pp_start = read_and_validate_next_proof();
+    // TODO: Do this in better
+    let mut cl_proof_pp_start = read_and_validate_next_proof();
+    cl_proof_pp_start.initial_snapshot.slot += 1;
+
```
MdTeach, line 16: Need to add context why `cl_proof_pp_start.initial_snapshot.slot += 1` was done.

MdTeach (changes requested):

bin/prover-client/src/proving_ops/btc_ops.rs
```diff
@@ -25,19 +18,23 @@ use crate::{
...
+            btc_client,
+            rollup_params,
+        }
     }
 }
 
 #[async_trait]
 impl ProvingOperations for BtcOperations {
-    type Input = (Block, RollupParams);
+    type Input = (Block, Vec<TxFilterRule>, CredRule);
```
delbonis, line 37: Should probably wrap this in a struct at this point.

bin/prover-client/src/proving_ops/cl_ops.rs
```diff
@@ -19,14 +20,20 @@ use crate::{
...
+    rollup_params: Arc<RollupParams>,
 }
 
 impl ClOperations {
     /// Creates a new CL operations instance.
-    pub fn new(cl_client: HttpClient, el_dispatcher: Arc<TaskDispatcher<ElOperations>>) -> Self {
+    pub fn new(
+        cl_client: HttpClient,
+        el_dispatcher: Arc<TaskDispatcher<ElOperations>>,
+        rollup_params: Arc<RollupParams>,
```
delbonis, line 31: Should be `Arc<Params>`, we don't pass around the inner `RollupParams` in an arg.  Although this might all get refactored soon anyways.

crates/btcio/src/reader/query.rs
```diff
@@ -299,28 +304,129 @@ pub async fn get_verification_state(
...
+    // This fetches timestamps of `vh-10`,`vh-9`, ... `vh-1`, `vh`
+    for i in 0..N {
+        if vh >= i as u64 {
+            let height_to_fetch = vh - i as u64;
+            let h = client.get_block_at(height_to_fetch).await?;
+            timestamps[N - 1 - i] = h.header.time;
         } else {
-            timestamps[i] = 0;
+            // No more blocks to fetch; the rest remain zero
+            timestamps[N - 1 - i] = 0;
```
delbonis, line 319: Curious about what's changing here.  And couldn't we load this out of the db instead of fetching it again?

crates/consensus-logic/src/l1_handler.rs
```diff
@@ -169,6 +169,8 @@ pub fn verify_proof(
     checkpoint: &BatchCheckpoint,
     rollup_params: &RollupParams,
 ) -> anyhow::Result<()> {
+    debug!(?checkpoint, ?rollup_params, "Verifying checkpoint proof");
```
delbonis, line 172: Pls don't log the entire rollup params, that makes the logs very verbose.

crates/proof-impl/btc-blockspace/src/logic.rs
```diff
@@ -2,59 +2,68 @@
...
+};
 
 #[derive(Debug, BorshSerialize, BorshDeserialize)]
 pub struct BlockspaceProofOutput {
     pub header_raw: Vec<u8>,
     pub deposits: Vec<DepositInfo>,
     pub prev_checkpoint: Option<BatchCheckpoint>,
-    pub rollup_params_commitment: Buf32,
+    pub tx_filters_commitment: Buf32,
+    pub cred_rule: CredRule,
```
delbonis, line 20: Is this only used for checking checkpoint signatures?  This will have to be refactored when we have multiple sequencers eventually.

delbonis (changes requested):
Minor comments, some concerns about which types being used.

crates/zkvm/adapters/sp1/src/prover.rs
prajwolrg: Yup.

bin/prover-client/src/proving_ops/cl_ops.rs
prajwolrg: Why shouldn't we pass `RollupParams`? In the `prover-client`, non-rollup params are not required.

crates/btcio/src/reader/query.rs
prajwolrg: It turns out there was error on how the timestamps were ordered previously. This fixes that. Now, timestamps[0] will be the timestamp of the oldest block. 
We can load this from the db. Right now this is called only once by the sequencer for the genesis block. The HeaderVerificationState is then in the client state and is updated there. This is also called from the prover-client independently right now. We want to fetch the HeaderVerificationState directly from the sequencer later on.

crates/proof-impl/btc-blockspace/src/logic.rs
prajwolrg: Yes the `cred_rule` is only used for validating signatures. `tx_filters_commitment` is used so that we don't have to compute it on every blockspace proof. We later verify it is from rollup params in the batch proof.

bin/prover-client/src/proving_ops/btc_ops.rs
prajwolrg: Will work on a better interface for handling this. Planning to move this into `proof-impl` crate for better consistency.

crates/consensus-logic/src/client_transition.rs
```diff
@@ -58,16 +58,15 @@ pub fn process_event<D: Database>(
...
-                        updated_l1vs = updated_l1vs
-                            .check_and_update_continuity_new(&header, &get_btc_params());
-                    }
+
+                for block_height in (l1_vs_height + 1..=*height) {
+                    let block_mf = l1prov
+                        .get_block_manifest(block_height)?
+                        .ok_or(Error::MissingL1BlockHeight(block_height))?;
+                    let header: Header =
+                        bitcoin::consensus::deserialize(block_mf.header()).unwrap();
```
storopoli, line 67: This needs to be imported in the top as

```rs
using bitcoin::consensus
```
then
```suggestion
                        consensus::deserialize(block_mf.header()).unwrap();
```

crates/consensus-logic/src/l1_handler.rs
```diff
@@ -189,11 +191,14 @@ pub fn verify_proof(
             Risc0Verifier::verify_groth16(proof, vk.as_ref(), &public_params_raw)
         }
         RollupVerifyingKey::SP1VerifyingKey(vk) => {
-            SP1Verifier::verify_groth16(proof, vk.as_ref(), &public_params_raw)
+            SP1Verifier::verify_groth16_raw(proof, vk.as_ref(), &public_params_raw)
         }
     });
     match res {
-        Ok(Ok(())) => Ok(()),
+        Ok(Ok(())) => {
```
storopoli, line 198:
```suggestion
        Ok(Ok(_)) => {
```
We don't care about the type of this nested OK right?
This might also catch bugs in the future.

crates/proof-impl/btc-blockspace/src/logic.rs
```diff
@@ -2,59 +2,68 @@
...
-        rollup_params,
-    } = input;
-    assert!(check_merkle_root(block));
-    // assert!(check_witness_commitment(block));
+pub fn process_blockspace_proof(
+    serialized_block: &[u8],
+    cred_rule: &CredRule,
+    serialized_tx_filters: &[u8],
+) -> BlockspaceProofOutput {
+    let block: Block = bitcoin::consensus::deserialize(serialized_block).unwrap();
```
storopoli, line 28: Same as above

bin/prover-client/src/prover.rs
```diff
@@ -110,13 +109,15 @@ where
             Vm::Input::new().write(&el_input)?.build()?
         }
 
-        ZKVMInput::BtcBlock(block, rollup_params) => Vm::Input::new()
-            .write(&rollup_params)?
+        ZKVMInput::BtcBlock(block, cred_rule, tx_filters) => Vm::Input::new()
+            .write(&cred_rule)?
             .write_serialized(&bitcoin::consensus::serialize(&block))?
```
storopoli, line 114: Same as above

crates/proof-impl/btc-blockspace/src/logic.rs
```diff
@@ -2,59 +2,68 @@
...
 
-    let (deposits, prev_checkpoint) = extract_relevant_info(block, rollup_params);
-    let rollup_params_commitment = rollup_params.compute_hash();
+    let tx_filters: Vec<TxFilterRule> = borsh::from_slice(serialized_tx_filters).unwrap();
+
+    let (deposits, prev_checkpoint) = extract_relevant_info(&block, &tx_filters, cred_rule);
+    let tx_filters_commitment = strata_primitives::hash::raw(serialized_tx_filters);
 
     BlockspaceProofOutput {
         header_raw: bitcoin::consensus::serialize(&block.header),
```
storopoli, line 39: same as above

crates/proof-impl/btc-blockspace/src/logic.rs
```diff
@@ -2,59 +2,68 @@
...
+        let tx_filters = derive_tx_filter_rules(rollup_params).unwrap();
+        let serialized_tx_filters = borsh::to_vec(&tx_filters).unwrap();
+
         let btc_block = get_btc_chain().get_block(40321).clone();
-        let input = BlockspaceProofInput {
-            block: btc_block,
-            rollup_params: rollup_params.clone(),
-        };
-        let _ = process_blockspace_proof(&input);
+        let serialized_btc_block = bitcoin::consensus::serialize(&btc_block);
```
storopoli, line 61: same as above

bin/prover-client/Cargo.toml
```diff
@@ -41,6 +41,7 @@ rayon = "1.8.0"
 reth-rpc-types = { workspace = true }
 rockbound = { workspace = true }
 serde = { workspace = true }
+serde_json = { workspace = true }
```
storopoli, line 44: Can we change all this `foo = { workspace = true }` to `foo.workspace = true` in this file to match our root and other `Cargo.toml`s in the repo?

bin/prover-client/src/proving_ops/cl_ops.rs
storopoli: Why not just a `&RollupParams` and let the user `Arc`it if necessary?

storopoli: Or `&Params` just don't shove an `Arc` down the throat.

crates/state/src/l1/id.rs
```diff
@@ -37,12 +38,30 @@ impl From<Buf32> for L1BlockId {
...
+// /// Ignored for now because this is a breaking change!
+// impl fmt::Debug for L1BlockId {
+//     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+//         let reversed_bytes: Vec<u8> = self.0 .0.iter().rev().cloned().collect();
+//         let mut buf = [0; 64];
+//         hex::encode_to_slice(&reversed_bytes, &mut buf).expect("buf: enc hex");
+//         f.write_str(unsafe { str::from_utf8_unchecked(&buf) })
+//     }
+// }
+
```
storopoli, line 64: Don't need an empty line here
```suggestion
```

crates/state/src/l1/timestamp_store.rs
```diff
@@ -3,6 +3,7 @@ use borsh::{BorshDeserialize, BorshSerialize};
 use serde::{Deserialize, Serialize};
 
 const N: usize = 11;
+const MID: usize = N / 2;
```
storopoli, line 6: What?!
usize 5.5?
This needs a docstring...
Does not make sense

crates/state/src/l1/timestamp_store.rs
```diff
@@ -16,28 +17,66 @@ const N: usize = 11;
     BorshDeserialize,
     Arbitrary,
 )]
+/// A fixed-size ring buffer that stores timestamps.
+/// It overwrites the oldest timestamps when new ones are inserted.
+
```
storopoli, line 22:
```suggestion
```

crates/zkvm/adapters/sp1/src/utils.rs
```diff
@@ -72,3 +74,20 @@ pub fn get_proving_keys(
         client.setup(guest_code)
     }
 }
+
+/// Extracts the raw Groth16 proof bytes from a given `Proof`.
```
storopoli, line 78:
```suggestion
/// Extracts the raw Groth16 proof bytes from a given [`Proof`].
```

crates/zkvm/adapters/sp1/src/verifier.rs
```diff
@@ -46,9 +48,30 @@ impl ZKVMVerifier for SP1Verifier {
...
     }
 
-    fn verify_groth16(proof: &Proof, vkey_hash: &[u8], committed_values_raw: &[u8]) -> Result<()> {
-        let vk = GROTH16_VK_BYTES;
+    fn verify_groth16(
+        raw_sp1_proof: &Proof,
+        vkey_hash: &[u8],
+        committed_values_raw: &[u8],
+    ) -> Result<()> {
+        let sp1_proof: SP1ProofWithPublicValues = bincode::deserialize(raw_sp1_proof.as_bytes())
```
storopoli, line 56: I don't like hidden imports.
Please add `use bincode::deserialize` on the top?

crates/zkvm/zkvm/src/lib.rs
```diff
@@ -36,6 +48,68 @@ impl From<&Proof> for Vec<u8> {
...
+    /// A unique identifier for the proof. This identifier is consistent for a given fixed input
+    /// and fixed program, meaning that for the same inputs and program, the `id` will always
+    /// remain the same.
+    id: String,
+    proof: Proof,
+    /// An optional identifier that uniquely identifies the proof on a remote server. Unlike `id`,
+    /// this field might vary for the same input and program depending on the context, such as
+    /// server-side processes or external systems. If the proof is generated locally and not
+    /// associated with a remote server, this field will be `None`.
+    remote_id: Option<String>,
```
storopoli, line 80: Amazing docstrings, but we need some padding here:
```suggestion
    /// A unique identifier for the proof. This identifier is consistent for a given fixed input
    /// and fixed program, meaning that for the same inputs and program, the `id` will always
    /// remain the same.
    id: String,
    
    proof: Proof,
    
    /// An optional identifier that uniquely identifies the proof on a remote server. Unlike `id`,
    /// this field might vary for the same input and program depending on the context, such as
    /// server-side processes or external systems. If the proof is generated locally and not
    /// associated with a remote server, this field will be `None`.
    remote_id: Option<String>,
```

storopoli (changes requested):
minor nits

bin/prover-client/Cargo.toml
prajwolrg: What's the preferred convention for enabling workspace dependencies: using the form `foo = { workspace = true }` versus `foo.workspace = true`? Are there any specific guidelines or best practices to follow?

I remember @delbonis asking to use `foo = { workspace = true }`.

prajwolrg:
> Why not just a &RollupParams and let the user Arcit if necessary? https://github.com/alpenlabs/strata/pull/419#discussion_r1808704471

This was done to avoid cloning. I'm not sure if this is a good way. `RollupParams` is needed in multiple structs. 
I tried using reference to `RollupParams` as a parameter in a struct using lifelines but couldn't properly figure out to use that properly within asynchronous contexts.

What do you recommend here?

bin/prover-client/src/proving_ops/cl_ops.rs
delbonis: The goal was to only have one instance of the `Params` type and its constitutents, so that's why we'd be passing it around as an `Arc`.

delbonis:
@prajwolrg The issue is we're being sloppy with how we init the client.  The `resolve_and_validate_rollup_params` function should return `Arc<Params>` (and really it maybe should be fetching from RPC but we don't have that :thinking:), which is the toplevel type that we decode from the json that we'll distribute for the network parameters.  For the `ClOperations` type it *feels* like it shouldn't be in the arc.  I wouldn't worry too much about making it perfect since we'll be refactoring the params structure to better reflect the proof structure soon.

bin/prover-client/Cargo.toml
storopoli: We changed this when we renamed to `strata`. See the root-level `Cargo.toml` and a bunch of others `Cargo.toml`s on `main`.
Not putting words into Trey's mouth but I think he said that we should do whatever we want but be consistent.

Since the `foo.workspace = true` is more succinct, we are using that one. Now the only `Cargo.toml` that I could find that is inconsistent is this one.

delbonis (approved):

MdTeach (approved):

storopoli (approved):
ACK 82da6bc26f3973b9f8d3bf8ba0c99e4abc0902dd

Thanks for cleaning up my suggestions

PR #421, opened 2024-10-17
Functional tests: Bridge Deposits

storopoli:
## Description

Adds the bridge deposit functional test for the happy path.

Still draft, because this should work and I don't know why it is not working...

### Type of Change

-   [ ] Documentation update
-   [ ] Refactor
-   [x] New or updated tests

## Checklist

## Related Issues

STR-483

### DIFF
--- a/functional-tests/fn_bridge_deposit_happy.py
+++ b/functional-tests/fn_bridge_deposit_happy.py
@@ -2,10 +2,10 @@ import time
 
 import flexitest
 from bitcoinlib.services.bitcoind import BitcoindClient
-from strata_utils import deposit_request_transaction, get_address
+from strata_utils import deposit_request_transaction, get_address, drain_wallet
 
 from entry import BasicEnvConfig
-from utils import get_logger
+from utils import get_logger, wait_until, get_bridge_pubkey
 
 
 @flexitest.register
@@ -25,12 +25,14 @@ class BridgeDepositHappyTest(flexitest.Test):
         self.logger = get_logger("BridgeDepositHappyTest")
 
     def main(self, ctx: flexitest.RunContext):
-        return True  # this will be handled in a future PR
         btc = ctx.get_service("bitcoin")
         seq = ctx.get_service("sequencer")
         reth = ctx.get_service("reth")
+
         el_address_1 = "deadf001900dca3ebeefdeadf001900dca3ebeef"
         el_address_2 = "deedf001900dca3ebeefdeadf001900dca3ebeef"
+        self.logger.debug(f"EL address 1: {el_address_1}")
+        self.logger.debug(f"EL address 2: {el_address_2}")
 
         seqrpc = seq.create_rpc()
         btcrpc: BitcoindClient = btc.create_rpc()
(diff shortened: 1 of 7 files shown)


functional-tests/fn_bridge_deposit_happy.py
```diff
@@ -44,114 +46,97 @@ def main(self, ctx: flexitest.RunContext):
...
-                print("Found expected tx in mempool")
-                tx_published = True
-                break
-            except Exception as e:
-                print(e)
-        assert tx_published, "Tx was not published"
-        # time to mature FIXME: is this 6 blocks?
+        # time to mature DT
         btcrpc.proxy.generatetoaddress(6, seq_addr)
+        time.sleep(10)
```
storopoli, line 121: This is not working and I don't know why. The sequencer does not pick up the transaction, despite being successfully broadcasted.
Any ideas @Rajil1213, @delbonis?

storopoli: Then it explodes here in L68
https://github.com/alpenlabs/strata/blob/d81b827fb59ad7a97d447dca208c0a11671fe3dd/functional-tests/fn_bridge_deposit_happy.py#L66-L68

storopoli: @Rajil1213 figured it out. The magic bytes change and I hard-coded the damn thing to be  `b"alpenstrata"` 🤦🏻‍♂️

delbonis (changes requested):
Minor comments.

functional-tests/fn_bridge_deposit_happy.py
```diff
@@ -44,114 +46,93 @@ def main(self, ctx: flexitest.RunContext):
...
+        original_balance = int(rethrpc.eth_getBalance(f"0x{el_address_2}"), 16)
+        self.logger.debug(f"Balance before deposit (EL address 2): {original_balance}")
+        btcrpc.proxy.generatetoaddress(102, addr_2)[0]
+        # Send DRT from Address 2 to EL Address 2
+        self.make_drt(ctx, el_address_2, bridge_pk)
+        new_balance = int(rethrpc.eth_getBalance(f"0x{el_address_2}"), 16)
+        self.logger.debug(f"Balance after deposit (EL address 2): {new_balance}")
+        assert new_balance > original_balance, "balance did not increase"
+        # Drain wallet back to sequencer so that we cannot use address 1 or change anymore
+        self.drain_wallet(ctx)
```
Rajil1213, line 82: Should refactor this to a separate function that does the checks. The following additional checks must be added:

1) Check the initial balance in the EL address
2) Assert that the difference in the balance after deposit is exactly equal to the `deposit_amount` (set in the rollup params).
3) Use the `getCurrentDeposits` RPC on the sequencer to check if the number of deposits increases as deposits are made.
4) Also check that multiple deposits to the same address works.

It's fine to split these tests out but it's preferable to not do that since setting up and tearing down takes a LOT of time and all these cases are part of the same "happy-case" environment.

Rajil1213 (changes requested):
Looks good but requires some additional checks for the sake of completeness/rigor.

Rajil1213 (changes requested):
Requires some minor refactoring.

functional-tests/fn_bridge_deposit_happy.py
```diff
@@ -25,133 +26,163 @@ def __init__(self, ctx: flexitest.InitContext):
...
+        self.logger.debug(f"Sequencer Address: {seq_addr}")
+        self.logger.debug(f"Address: {address}")
+
+        # Make sure that the el_address has zero balance
+        original_balance = int(rethrpc.eth_getBalance(f"0x{el_address}"), 16)
+        self.logger.debug(f"Balance before deposit (EL address): {original_balance}")
+        if new_address:
+            assert original_balance == 0, "balance is not zero"
+        else:
+            assert original_balance > 0, "balance is zero"
```
Rajil1213, line 169: This is not strictly necessary. For the purposes of this test, we only care whether the deposit works i.e, whether the balance for a particular account increases by a certain amount.

functional-tests/fn_bridge_deposit_happy.py
```diff
@@ -25,133 +26,163 @@ def __init__(self, ctx: flexitest.InitContext):
...
+        self.make_drt(ctx, el_address, bridge_pk)
+
+        # Make sure that the balance has increased
+        new_balance = int(rethrpc.eth_getBalance(f"0x{el_address}"), 16)
+        self.logger.debug(f"Balance after deposit (EL address): {new_balance}")
+        assert new_balance > original_balance, "balance did not increase"
+
+        # Make sure that the balance is 10 BTC in Strata "wei"
+        expected_balance = rollup_deposit_amount * (10**10) * (2 if not new_address else 1)
+        assert new_balance == expected_balance, "balance is not the default rollup_deposit_amount"
```
Rajil1213, line 184: And this assertion should only check the difference i.e., `new_balance - expected_balance == rollup_deposit_amount`.

functional-tests/fn_bridge_deposit_happy.py
delbonis: It would say it's fine an thorough to check that other accounts that we know we didn't deposit to are *not* credited.

Rajil1213: > I would say it's fine an thorough to check that other accounts that we know we didn't deposit to are not credited.

Well, if we _were_ making that assertion, sure. All this asserts right now is that some random address was not credited, and not that no other account was erroneously credited.

functional-tests/fn_bridge_deposit_happy.py
```diff
@@ -25,133 +26,163 @@ def __init__(self, ctx: flexitest.InitContext):
...
+
+        # 3rd deposit, now to a previously used address
+        n_deposits_pre_3 = len(seqrpc.strata_getCurrentDeposits())
+        self.logger.debug(f"Current deposits: {n_deposits_pre_3}")
+        assert self.test_deposit(ctx, addr_3, el_address_1, new_address=False)
+        n_deposits_post_3 = len(seqrpc.strata_getCurrentDeposits())
+        self.logger.debug(f"Current deposits: {n_deposits_post_3}")
+        assert n_deposits_post_3 == n_deposits_pre_3 + 1
+
+        # 4th deposit, now to a previously used address
```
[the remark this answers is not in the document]
storopoli: Removed in 0e5bbb07597354c168807d85323b0957a07d36cd

functional-tests/fn_bridge_deposit_happy.py
```diff
@@ -25,133 +26,163 @@ def __init__(self, ctx: flexitest.InitContext):
...
+        self.logger.debug(f"Balance after deposit (EL address): {new_balance}")
+        assert new_balance > original_balance, "balance did not increase"
+
+        # Make sure that the balance is 10 BTC in Strata "wei"
+        expected_balance = rollup_deposit_amount * (10**10) * (2 if not new_address else 1)
+        assert new_balance == expected_balance, "balance is not the default rollup_deposit_amount"
+
+        # Drain wallet back to sequencer so that we cannot use address 1 or change anymore
+        self.drain_wallet(ctx)
+        return True
```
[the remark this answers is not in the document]
storopoli: Done in 0e5bbb07597354c168807d85323b0957a07d36cd
thanks!

functional-tests/fn_bridge_deposit_happy.py
```diff
@@ -25,133 +26,163 @@ def __init__(self, ctx: flexitest.InitContext):
...
-        seqrpc = seq.create_rpc()
+        addr_1 = get_address(0)
+        addr_2 = get_address(1)
+        addr_3 = get_address(2)
+        addr_4 = get_address(3)
+
+        # 1st deposit
+        n_deposits_pre_1 = len(seqrpc.strata_getCurrentDeposits())
+        self.logger.debug(f"Current deposits: {n_deposits_pre_1}")
+        assert self.test_deposit(ctx, addr_1, el_address_1)
```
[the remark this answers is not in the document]
storopoli: Refactored in 0e5bbb07597354c168807d85323b0957a07d36cd

functional-tests/fn_bridge_deposit_happy.py
storopoli: good suggestion, done in 0e5bbb07597354c168807d85323b0957a07d36cd

functional-tests/fn_bridge_deposit_happy.py
```diff
@@ -44,114 +127,46 @@ def main(self, ctx: flexitest.RunContext):
...
-                break
-            except Exception as e:
-                print(e)
-        assert tx_published, "Tx was not published"
-        # time to mature FIXME: is this 6 blocks?
-        btcrpc.proxy.generatetoaddress(6, seq_addr)
-        time.sleep(1)
+        n_deposits_post = len(seqrpc.strata_getCurrentDeposits())
+        self.logger.debug(f"Current deposits: {n_deposits_post}")
+        assert n_deposits_post == n_deposits_pre + 1
```
Rajil1213, line 159: This assertion needs a failure message.

functional-tests/fn_bridge_deposit_happy.py
```diff
@@ -25,12 +26,94 @@ def __init__(self, ctx: flexitest.InitContext):
...
+        # 1st deposit
+        self.test_deposit(ctx, addr_1, el_address_1)
+
+        # 2nd deposit
+        self.test_deposit(ctx, addr_2, el_address_2)
+
+        # 3rd deposit, now to a previously used address
+        self.test_deposit(ctx, addr_3, el_address_1, new_address=False)
+
+        return True
```
Rajil1213, line 45: This return isn't necessary either. AFAIK, flexitest only requires that the functions don't throw any exceptions.

Rajil1213 (changes requested):
Some minor nits but otherwise looks good. Nice work!

functional-tests/fn_bridge_deposit_happy.py
storopoli: `ruff` complains :/

storopoli:
Cache is rugpulling me in the CI failure but this runs OK locally and used to run fine here also (don't know what's happening)

functional-tests/fn_bridge_deposit_happy.py
storopoli: Done in 1df4fb61dd69c50e6fdcfc65a281d1092000d54e

Rajil1213:
> Cache is rugpulling me in the CI failure but this runs OK locally and used to run fine here also (don't know what's happening)

Does removing the `if` check in the `Install dependencies` step help? I think python is smart enough to not reinstall deps if the version has not changed.

On an unrelated point, a bigger issue that I just now discovered is that `cargo b --workspace` fails on mac with the arm chips even after setting `PYTHON_CONFIGURE_OPTS="--enable-shared"` as suggested in the [pyo3 user guide](https://pyo3.rs/v0.18.1/getting_started#virtualenvs). Will post something on Slack to see if anybody else is facing the same issue. On the other hand, `poetry install --no-root` works just fine 🤷 .

functional-tests/fn_bridge_deposit_happy.py
Rajil1213: That's strange. Didn't complain when I ran it locally.

storopoli:
> > Cache is rugpulling me in the CI failure but this runs OK locally and used to run fine here also (don't know what's happening)
> 
> Does removing the `if` check in the `Install dependencies` step help? I think python is smart enough to not reinstall deps if the version has not changed.
> 
> On an unrelated point, a bigger issue that I just now discovered is that `cargo b --workspace` fails on mac with the arm chips even after setting `PYTHON_CONFIGURE_OPTS="--enable-shared"` as suggested in the [pyo3 user guide](https://pyo3.rs/v0.18.1/getting_started#virtualenvs). Will post something on Slack to see if anybody else is facing the same issue. On the other hand, `poetry install --no-root` works just fine 🤷 .

I have a Mac now and I don't have that ENV var and `poetry install --no-root` works.

Rajil1213:
> I have a Mac now and I don't have that ENV var and poetry install --no-root works.

Yeah that works. Does `cargo b --workspace` work too?

storopoli:
> > I have a Mac now and I don't have that ENV var and poetry install --no-root works.
> 
> Yeah that works. Does `cargo b --workspace` work too?
No, mine is broken as well :(

functional-tests/fn_bridge_deposit_happy.py
storopoli: I get this:

```console
     ├╴  Method "main" overrides class "Test" in an incompatible manner
     │      Return type mismatch: base method returns type "Literal[True]", override returns type "None"
     │        "None" is not assignable to "Literal[True]" Pyright (reportIncompatibleMethodOverride) [28, 9]
```

storopoli: Oh, this is pyright.

Rajil1213:
> No, mine is broken as well :(

Could you create a ticket for this? It's not a blocker for this PR specifically. But it is definitely a blocker for almost any development on `main` (at least for the majority of people that are now using a mac with arm chips).

functional-tests/fn_bridge_deposit_happy.py
Rajil1213: You could add an explicit `-> None` annotation for the return type of `main` and that should fix it, I think. No idea why pyright is assuming that no return type annotation means `Literal[True]`. 🤷

functional-tests/fn_bridge_deposit_happy.py
storopoli: https://alpenlabs.atlassian.net/browse/STR-541

storopoli:
> > No, mine is broken as well :(
> 
> Could you create a ticket for this? It's not a blocker for this PR specifically. But it is definitely a blocker for almost any development on `main` (at least for the majority of people that are now using a mac with arm chips).

https://alpenlabs.atlassian.net/browse/STR-541

functional-tests/fn_bridge_deposit_happy.py
storopoli: This needs a change in https://codeberg.org/treyd/flexitest/src/commit/7a1483f7b9c14ee8e23003632a5b138367f19be1/flexitest/__init__.py#L35-L36

Rajil1213: The explicit annotation did not work either? 🫨 
If that's the case, then this can be left as is.

In `flexitest`, the function should probably just be `pass`?

crates/util/python-utils/build.rs
```diff
@@ -0,0 +1,9 @@
+//! Needed for MacOS to link the Python library.
+//! See: https://pyo3.rs/v0.14.2/building_and_distribution.html#macos
+
+fn main() {
+    #[cfg(target_os = "macos")]
+    pyo3_build_config::add_extension_module_link_args();
+    #[cfg(not(target_os = "macos"))]
+    ()
```
delbonis, line 8: Redundant?  Maybe print a warning?

delbonis (approved):
This looks good broadly aside from the, like, major issue with it.

Rajil1213 (approved):

storopoli:
> This looks good broadly aside from the, like, major issue with it.

What is the major issue?

PR #424, opened 2024-10-22
refactor(btcio): remove bitcoind_json_rpc_types

storopoli:
It does not make sense to have this dep since we're already half-baking
our own custom types on top of it and using only a tiny minority of the
types as being the ones provided by bitcoind_json_rpc_types.


storopoli:
One less dependency @delbonis.

PR #426, opened 2024-10-22
Prover Manager Functional Test for Deposit & Withdrwal

MdTeach:
## Description
* Requires https://github.com/alpenlabs/strata/pull/419 to be merged
* Functional test to ensure proof generation works for the block containing Deposits & Withdrawal txns
* Minor updates on the EL, CL and Checkpoint guest code public params to assert deposit info matches from L1 and L2

### Type of Change

-   [ ] Documentation update
-   [ ] Refactor
-   [ ] New or updated tests

## Checklist


bin/prover-client/Cargo.toml
```diff
@@ -41,6 +41,7 @@ rayon = "1.8.0"
 reth-rpc-types = { workspace = true }
 rockbound = { workspace = true }
 serde = { workspace = true }
+serde_json = { workspace = true }
 sp1-sdk = "2.0.0"
 strata-tx-parser = { workspace = true }
 thiserror = { workspace = true }
```
storopoli, line 47: Can we change all the `foo = { workspace = true }` in this file for `foo = workspace.true`?

storopoli (changes requested):
This is a backport no?

bin/prover-client/src/ckp_runner.rs
delbonis, line 1: Consistency: rename to `ckpt_runner`

crates/proof-impl/checkpoint/src/lib.rs
```diff
@@ -21,7 +21,7 @@ pub struct ChainStateSnapshot {
 
 #[derive(Debug, Clone, BorshSerialize, BorshDeserialize)]
 pub struct L2BatchProofOutput {
-    pub deposits: Vec<DepositInfo>,
+    pub deposits: Vec<ELDepositData>,
```
delbonis, line 24: Can you elaborate on what this field represents now?  Having the indexes in here is redundant since they'll always be ordered, etc..  This type wasn't meant to exist as a long term solution, it was part of the `Op` stopgap.  What goal are we trying to achieve here?

Since we can figure out the indexes from other externally observed data, we can construct everything from purely the index that was processed by looking at the deposit intents queue on the exec env.

crates/proof-impl/cl-stf/src/lib.rs
```diff
@@ -40,25 +48,57 @@ fn verify_l2_block(
...
+                gwei_to_sats(deposit_request.amount),
+                deposit_request.address.as_slice().to_vec(),
+            )
+        })
+        .collect();
+
+    let applied_ops = el_deposit_data
+        .iter()
+        .map(|deposit_data| Op::Deposit(deposit_data.clone()))
+        .collect();
```
delbonis, line 88: Try to avoid using anything with this `Op` type.  The deposit intents queue is the source of truth.  See other comment.

crates/proof-impl/evm-ee-stf/src/processor.rs
```diff
@@ -33,7 +33,7 @@ use revm::{
     primitives::{SpecId, TransactTo, TxEnv},
     Database, DatabaseCommit, Evm,
 };
-use strata_reth_evm::set_evm_handles;
+use strata_reth_evm::{constants::BRIDGEOUT_ADDRESS, set_evm_handles};
```
delbonis, line 36: Is this really the normal way to do this as opposed to pulling it from the network params file?

crates/proof-impl/evm-ee-stf/src/processor.rs
```diff
@@ -315,9 +315,13 @@ impl EvmProcessor<InMemoryDB> {
                 storage_root,
                 code_hash: account.info.code_hash,
             };
-            state_trie
-                .insert_rlp(&state_trie_index, state_account)
-                .unwrap();
+
+            // Skip adding BridgeOut precompile (stateless contract) to the state tree.
```
delbonis, line 319: Why are we skipping this exactly?

crates/proof-impl/btc-blockspace/src/block.rs
```diff
@@ -76,8 +91,63 @@ pub fn compute_witness_root(l1_tx: &L1Tx) -> Buf32 {
...
+    let coinbase = &block.txdata[0];
+    if !coinbase.is_coinbase() {
+        return false;
+    }
+
+    // The commitment is recorded in a scriptPubKey of the coinbase transaction. It must be at least
+    // 38 bytes, with the first 6-byte of 0x6a24aa21a9ed, that is:
+    //
+    // 1-byte - OP_RETURN (0x6a)
+    // 1-byte - Push the following 36 bytes (0x24)
```
delbonis, line 120: "at least" or "exactly"?

delbonis (changes requested):
Some comments on the approach with the use of the `ELDepositData` type, should be focusing on the intents queue instead.  Overall seems okay for now.

crates/proof-impl/checkpoint/src/lib.rs
MdTeach: The idea is to verify that the deposits identified from the L1 scan match those made on the EL layer. This ensures that the sequencer did not overlook any deposit transactions settled on L1.

On the current CL-EL implementation, we are passing the deposits as Eth-POS `Withdrawals` while building the block. So, from the EL-CL proof we can recover `ELDepositData` not `DepositInfo` so following changes were done.

delbonis: Ah we gotta rework this post devnet.

PR #429, opened 2024-10-25
Issue str 100/macro buf impl

sudikshyapant:
## Description

This PR refactors the Buf20, Buf32, and Buf64 types to use a macro for generating their common implementations.

### Type of Change

-   [ ] Documentation update
-   [ ] Refactor
-   [ ] New or updated tests

## Checklist

### DIFF
--- a/crates/primitives/src/macros.rs
+++ b/crates/primitives/src/macros.rs
@@ -34,12 +34,31 @@ macro_rules! impl_buf {
             }
         }
 
-        impl Into<[u8; $len]> for $name {
-            fn into(self) -> [u8; $len] {
-                self.0
+        impl<'a> From<&'a [u8; $len]> for $name {
+            fn from(data: &'a [u8; $len]) -> Self {
+                Self(*data)
+            }
+        }
+
+        // Conditionally implement Default
+        impl Default for $name {
+            fn default() -> Self {
+                Self([0; $len])
             }
         }
 
         // Add any other common impls here if needed.
     };
 }
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Buf20([u8; 20]);
+impl_buf!(Buf20, 20);
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Buf32([u8; 32]);
+impl_buf!(Buf32, 32);
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Buf64([u8; 64]);
+impl_buf!(Buf64, 64);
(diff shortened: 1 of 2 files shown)


crates/primitives/src/macros.rs
```diff
@@ -0,0 +1,108 @@
...
+    };
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Buf20([u8; 20]);
+impl_buf!(Buf20, 20);
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Buf32([u8; 32]);
+impl_buf!(Buf32, 32);
```
storopoli, line 65: Shouldn't this also remove the `Buf32` definition somewhere?

crates/primitives/src/macros.rs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        impl Default for $name {
+            fn default() -> Self {
+                Self([0; $len])
+            }
+        }
+
+        // Add any other common impls here if needed.
+    };
```
delbonis, line 56: This is a good start, but we should add `Debug` and `Display` impls for these too.

Note this pattern to avoid allocs:

```rs

impl fmt::Debug for Buf20 {
        let mut buf = [0; 40];
        hex::encode_to_slice(self.0, &mut buf).expect("buf: enc hex");
        f.write_str(unsafe { str::from_utf8_unchecked(&buf) })
    }
}

impl fmt::Display for Buf20 {
        let mut buf = [0; 6];
        hex::encode_to_slice(&self.0[..3], &mut buf).expect("buf: enc hex");
        f.write_str(unsafe { str::from_utf8_unchecked(&buf) })?;
        f.write_str("..")?;
        hex::encode_to_slice(&self.0[17..], &mut buf).expect("buf: enc hex");
        f.write_str(unsafe { str::from_utf8_unchecked(&buf) })?;
        Ok(())
    }
}
```

crates/primitives/src/macros.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Buf20([u8; 20]);
```
delbonis, line 60: In the `buf` module we were using `pub FixedBytes<20>`.  I don't remember specifically why, although it would be smart to keep this up.

delbonis (changes requested):
Thank you for getting to this.  Also maybe include `.into_inner()` too.

crates/primitives/src/macros.rs
storopoli: Yes, something needs to be replaced somewhere.

sudikshyapant:
will address all the comments in a new PR

PR #455, opened 2024-11-07
Define ZkVm Trait and Implement for Risc0 and Sp1

prajwolrg:
## Description

This PR introduces a new ZkVm trait, implemented by both the Risc0 and SP1 adapters, consolidating guest code to avoid duplication. This allows us to move more logic into the proof-impl, ensuring a consistent structure across implementations in the provers/ directory.

### Type of Change

-   [X] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
-   [ ] Documentation update
-   [X] Refactor
-   [ ] New or updated tests

## Checklist


crates/zkvm/zkvm/src/lib.rs
```diff
@@ -204,3 +204,102 @@ impl AggregationInput {
         &self.vk
     }
 }
+
+/// A trait representing a Zero-Knowledge Virtual Machine (ZkVM) interface.
+/// Provides methods for reading inputs, committing outputs, and verifying proofs
+/// within the ZkVM environment.
+pub trait ZkVm {
```
delbonis, line 211: Shouldn't this be named something like `ZkVmHost`, `ZkVmInterface`, or `ZkVmIo`?  A name like this makes me think that it's to generically interface with the zkvm runtime from the *outside*.

crates/zkvm/zkvm/src/lib.rs
```diff
@@ -204,3 +204,102 @@ impl AggregationInput {
...
+    /// Reads a serialized object from the guest code, deserializing it using Serde.
+    ///
+    /// The input is expected to be written with [`ZKVMInputBuilder::write`].
+    fn read<T: DeserializeOwned>(&self) -> T;
```
delbonis, line 215: Should be `read_serde` or `read_bincode` to be more explicit.

crates/zkvm/zkvm/src/lib.rs
```diff
@@ -204,3 +204,102 @@ impl AggregationInput {
...
+
+    /// Reads a serialized byte slice from the guest code.
+    ///
+    /// The input is expected to be written with [`ZKVMInputBuilder::write_serialized`].
+    fn read_slice(&self) -> Vec<u8>;
```
delbonis, line 220: Should be `read_buf`.

crates/zkvm/zkvm/src/lib.rs
```diff
@@ -204,3 +204,102 @@ impl AggregationInput {
...
+
+    /// Commits a Serde-serializable object to the public values stream.
+    ///
+    /// Values that are committed can be proven as public parameters.
+    fn commit<T: Serialize>(&self, output: &T);
```
delbonis, line 225: Should be `commit_serde` or `commit_bincode` to be more explicit.

crates/zkvm/zkvm/src/lib.rs
```diff
@@ -204,3 +204,102 @@ impl AggregationInput {
...
+
+    /// Commits a pre-serialized byte slice to the public values stream.
+    ///
+    /// This method is intended for cases where the data has already been serialized
+    /// outside of the ZkVM's standard serialization methods. It allows you to provide
+    /// serialized outputs directly, bypassing any further serialization.
+    fn commit_slice(&self, raw_output: &[u8]);
```
delbonis, line 232: This is fine as-is, although maybe if you change the other one it also makes sense to rename to `commit_buf`.

crates/zkvm/zkvm/src/lib.rs
```diff
@@ -204,3 +204,102 @@ impl AggregationInput {
...
+    /// The input is expected to be written with [`ZKVMInputBuilder::write_borsh`].
+    fn read_borsh<T: BorshSerialize + BorshDeserialize>(&self) -> T {
+        let slice = self.read_slice();
+        borsh::from_slice(&slice).expect("borsh serialization failed")
+    }
+
+    /// Commits a Borsh-serializable object to the public values stream.
+    ///
+    /// Values that are committed can be proven as public parameters.
+    fn commit_borsh<T: BorshSerialize + BorshDeserialize>(&self, output: &T) {
```
delbonis, line 271: Do these both need to be both `BorshSerialize` and `BorshDeserialize`?

crates/zkvm/zkvm/src/lib.rs
```diff
@@ -204,3 +204,102 @@ impl AggregationInput {
...
+    /// # Returns
+    ///
+    /// Returns `Ok(())` if the proof verifies successfully, or an `anyhow::Error` if verification
+    /// fails.
+    fn verify_groth16(
+        &self,
+        proof: &[u8],
+        verification_key: &[u8],
+        public_params_raw: &[u8],
+    ) -> anyhow::Result<()>;
```
delbonis, line 258: First maybe should be `verify_native_proof`?  And then these I would put after the io fns to keep them together.

crates/zkvm/zkvm/src/lib.rs
```diff
@@ -204,3 +204,102 @@ impl AggregationInput {
...
+    /// Borsh.
+    ///
+    /// This function is similar to [`ZkVm::read_verified`], but is intended for guest commitments
+    /// committed via [`ZkVm::commit_borsh`]. The output is expected to be Borsh-serializable.
+    /// It then verifies the proof using the internal verification key context.
+    ///
+    /// This is equivalent to calling [`ZkVm::read_borsh`] and [`ZkVm::verify_proof`], but avoids
+    /// double serialization and deserialization. The function will panic if the proof fails to
+    /// verify.
+    fn read_verified_borsh<T: BorshSerialize + BorshDeserialize>(&self, vk_digest: &[u32; 8]) -> T {
```
delbonis, line 300: These two methods feel kinda off to me.  What would be a likely scenario to override these by downstream zkvm impls?  Could they be broken out to an impl fn?

Also see above comment about `_serde`/`_bincode` names.

delbonis (changes requested):
Broad strokes looks good, main comments are on ensuring clarity with this new abstraction.

crates/zkvm/zkvm/src/lib.rs
prajwolrg: `ZkVmHost` interface is already defined and is meant for outside of zkvm runtime.
Does renaming to `ZkVmEnv` makes sense?

crates/zkvm/zkvm/src/lib.rs
prajwolrg: Have updated this to be more consistent.

crates/proof-impl/cl-agg/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+        rollup_params_commitment,
+    };
+
+    zkvm.commit_borsh(&public_params);
+}
+
+fn validate_proof_consistency(
+    current_proof_cs_snap: &L2BatchProofOutput,
+    next_proof_cs_snap: &L2BatchProofOutput,
+) {
```
MdTeach, line 43: Since, the function `validate_proof_consistency` is quite small and been called every time during the loop so here I think wrapping the function `validate_proof_consistency` with `#[inline]` would be appropriate to avoid function call overhead.

MdTeach (dismissed):

crates/zkvm/zkvm/src/lib.rs
delbonis: That makes sense yeah.

bin/prover-client/src/prover.rs
```diff
@@ -107,18 +107,18 @@ where
     let zkvm_input = match zkvm_input {
         ZKVMInput::ElBlock(el_input) => {
             let el_input: ELProofInput = bincode::deserialize(&el_input.data)?;
-            Vm::Input::new().write(&el_input)?.build()?
+            Vm::Input::new().write_serde(&el_input)?.build()?
```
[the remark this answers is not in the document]
prajwolrg: No we can't do that. SP1 uses bincode natively, wheras Risc0 has it's own serde-compatible de/serializer.

MdTeach (dismissed):

crates/zkvm/zkvm/src/verifier.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    fn verify_groth16(
+        proof: &Proof,
+        verification_key: &[u8],
+        public_params_raw: &[u8],
+    ) -> anyhow::Result<()>;
+
+    /// Extracts the public output from the given proof using standard `serde`
+    /// serialization/deserialization.
+    fn extract_public_output<T: Serialize + DeserializeOwned>(proof: &Proof) -> anyhow::Result<T>;
```
sapinb, line 29: `extract_serde_public_output` ?

Rajil1213 (dismissed):

crates/zkvm/zkvm/src/verifier.rs
```diff
@@ -0,0 +1,54 @@
...
+    fn extract_borsh_public_output<T: BorshSerialize + BorshDeserialize>(
+        proof: &Proof,
+    ) -> anyhow::Result<T>;
+}
+
+/// Verification Key required to verify proof generated from `ZkVmHost`
+#[derive(
+    Debug, Clone, Serialize, Deserialize, BorshSerialize, BorshDeserialize, PartialEq, Eq, Arbitrary,
+)]
+pub struct VerificationKey(pub Vec<u8>);
```
delbonis, line 44: Let's make this an opaque type, remove the `pub`?

crates/zkvm/zkvm/src/verifier.rs
```diff
@@ -0,0 +1,54 @@
...
+        proof: &Proof,
+        verification_key: &[u8],
+        public_params_raw: &[u8],
+    ) -> anyhow::Result<()>;
+
+    /// Extracts the public output from the given proof using standard `serde`
+    /// serialization/deserialization.
+    fn extract_serde_public_output<T: Serialize + DeserializeOwned>(
+        proof: &Proof,
+    ) -> anyhow::Result<T>;
```
delbonis, line 31: Same thing here as with the last round, I didn't pick this up.  `bincode` is the actual format, serde is just the framework.

prajwolrg: Yup. SP1 is using `bincode` while Risc0 has it's own serde compatible de/serializer.

delbonis (approved):

MdTeach (approved):

sapinb (approved):

PR #458, opened 2024-11-07
l1-filter: Add TxFilterConfig

bewakes:
### Type of Change

-   [ ] Documentation update
-   [ ] Refactor
-   [ ] New or updated tests

## Checklist


crates/tx-parser/src/filter.rs
```diff
@@ -24,6 +25,77 @@ pub enum TxFilterRule {
     DepositRequest(DepositTxParams),
     /// Deposit transaction with deposit config and address
     Deposit(DepositTxParams),
+    /// Addresses that are spent to
+    SpentToAddrs(Vec<BitcoinAddress>),
+    /// Blob ids that are expected
+    BlobIds(Vec<Buf32>),
```
bewakes, line 31: These vecs can be sorted and then we can do binary search.
Having `Vec` instead of Individual items lets us group the items, sort and do binary search as we want.

crates/tx-parser/src/filter.rs
```diff
@@ -47,46 +119,45 @@ pub fn derive_tx_filter_rules(params: &RollupParams) -> anyhow::Result<Vec<TxFil
...
-/// Filter all the relevant [`Transaction`]s in a block based on given [`TxFilterRule`]s
-pub fn filter_relevant_txs(block: &Block, filters: &[TxFilterRule]) -> Vec<ProtocolOpTxRef> {
+
+/// Filter protocol operatios as refs from relevant [`Transaction`]s in a block based on given
+/// [`TxFilterRule`]s
+pub fn filter_protocol_op_tx_refs(
+    block: &Block,
+    filter_config: TxFilterConfig,
+) -> Vec<ProtocolOpTxRef> {
+    let filter_rules = filter_config.into_rules();
```
bewakes, line 129: The config is converted to existing rules with added variants. We can add complex logic here if needed. But this should probably suffice.

crates/tx-parser/src/filter.rs
```diff
@@ -24,6 +25,77 @@ pub enum TxFilterRule {
...
+    rollup_name: RollupName,
+
+    /// For addresses that we expect spends to.
+    expected_addrs: Vec<BitcoinAddress>,
+
+    /// For blobs we expect to be written.
+    expected_blobs: Vec<Buf32>,
+
+    /// For deposits that might be spent from.
+    expected_outpoints: Vec<(Buf32, u32)>,
```
delbonis, line 48: Should use a dedicated type for this.

crates/tx-parser/src/filter.rs
```diff
@@ -24,6 +25,77 @@ pub enum TxFilterRule {
...
+    // eventually, in future version
+    // For bridge V1 deposits that do bitmap flagging for the multisig addr.
+    // operator_key_tbl: PublickeyTable,
+    /// EE addr length
+    ee_addr_len: u8,
```
delbonis, line 53: Should be max length.

crates/tx-parser/src/filter.rs
```diff
@@ -24,6 +25,77 @@ pub enum TxFilterRule {
...
+
+    /// Deposit denomination
+    deposit_denomination: u64, // sats
```
delbonis, line 56: Don't bake this in as a single value, it should be capable of being extended to support multiple denominations easily even if we don't support this today.

delbonis (changes requested):

crates/tx-parser/src/filter.rs
```diff
@@ -47,46 +119,45 @@ pub fn derive_tx_filter_rules(params: &RollupParams) -> anyhow::Result<Vec<TxFil
...
         .collect()
 }
 
 ///  if a [`Transaction`] is relevant based on given [`RelevantTxType`]s then we extract relevant
 ///  info
-fn check_and_extract_relevant_info(
-    tx: &Transaction,
-    filters: &[TxFilterRule],
-) -> Option<ProtocolOperation> {
+fn extract_protocol_op(tx: &Transaction, filters: &[TxFilterRule]) -> Option<ProtocolOperation> {
```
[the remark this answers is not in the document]
bewakes: This is used internally and not exposed to the reader module. The reader module calls the extraction function by passing the config. The config is converted to the TxFilterRule that exists already.

crates/tx-parser/src/filter.rs
```diff
@@ -24,6 +25,77 @@ pub enum TxFilterRule {
...
+
+    /// Operators addr
+    operator_addr: BitcoinAddress,
+}
+
+impl TxFilterConfig {
+    pub fn from_rollup_params(rollup_params: &RollupParams) -> anyhow::Result<Self> {
+        let operator_wallet_pks = get_operator_wallet_pks(rollup_params);
+        let address = generate_taproot_address(&operator_wallet_pks, rollup_params.network)?;
+
```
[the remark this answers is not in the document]
bewakes: Yes, when we do that it won't make sense here.

crates/tx-parser/src/filter.rs
bewakes: Yes, as I think about it more now, having a separate rule with enum variant looks kind of redundant. Basically I want a particular type/variant to produce a specific ProtocolOp,k but yes, that can be done with structs that wrap some value without the need of these enums.

crates/tx-parser/src/filter.rs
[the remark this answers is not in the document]
delbonis: Yeah that's redundant, see [this comment](https://github.com/alpenlabs/strata/pull/458#discussion_r1832977528).

crates/tx-parser/src/filter.rs
bewakes: Let's do that when we have concrete decisions/specs on what kind of denominations we will have.

crates/tx-parser/src/filter.rs
bewakes: Why max len? Is it because we'll have multiple EEs? I believe length is constant for the single ee we have now.

crates/tx-parser/src/filter.rs
```diff
@@ -24,6 +25,77 @@ pub enum TxFilterRule {
...
+    // For bridge V1 deposits that do bitmap flagging for the multisig addr.
+    // operator_key_tbl: PublickeyTable,
+    /// EE addr length
+    ee_addr_len: u8,
+
+    /// Deposit denomination
+    deposit_denomination: u64, // sats
+
+    /// Operators addr
+    operator_addr: BitcoinAddress,
```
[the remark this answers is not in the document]
bewakes: Possibly. What addresses do you suppose client will be watching for besides the operators multisig address? Parsing txs that spend to different address might result in different ProtocolOps. So, I am thinking of doing something like:
```rust
pub struct ExpectedAddr {
   address: Address,
   // type will indicate what exactly we need to parse from the transaction so that we can have all addrs sorted in a vec
   type: Operator | ...
}

// This one for sorting the addrs
impl PartialOrd for ExpectedAddr {
   fn partial_cmp() { // compare address }
}
```

crates/tx-parser/src/filter.rs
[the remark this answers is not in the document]
bewakes: Yes, the enum variants will have a single presence anyways, so makes sense to use a struct.

delbonis (changes requested):
Was thinking about this some more, in relation to how we'd support multiple denominations.  We should just bite the bullet here and directly have the list of deposit configs, we'd just only specify one at a time *right now*, which would gives us a better ability to extend it in the future.

bewakes:
> We should just bite the bullet here and directly have the list of deposit configs.

I don't see much significance of having multiple deposit configs. In that case won't there be redundant information across the configs like max len, magic bytes, address, etc? I think we should just have a sorted list of denominations. Also, there might be a case when we support any denominations greater than some value. In that case a list for denominations wouldn't work.

bewakes:
Removed the intermediate `TxFilterRule` enum. Still some changes, cleanups and testings to do.

delbonis (changes requested):

crates/primitives/src/sorted_vec.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// Creates a new, empty `SortedVec`.
+    pub fn new() -> Self {
+        Self { inner: Vec::new() }
+    }
+
+    /// Inserts an element into the `SortedVec`, maintaining sorted order.
+    pub fn insert(&mut self, value: T) {
+        let pos = self.inner.binary_search(&value).unwrap_or_else(|e| e);
+        self.inner.insert(pos, value);
+    }
```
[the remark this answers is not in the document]
bewakes: This was named `insert` to make the interface as similar to that of `Vec`. Also, inserting in a sorted vec implies the result should be sorted.  We can keep it as is after having a comment/docstring that says the operation is linear time.

Yeah, a merge operation would be a good addition.

crates/primitives/src/l1.rs
```diff
@@ -352,6 +352,28 @@ impl BorshDeserialize for BitcoinAddress {
...
+    }
+}
+
+impl Ord for BitcoinAddress {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.address
+            .script_pubkey()
+            .cmp(&other.address.script_pubkey())
+    }
+}
```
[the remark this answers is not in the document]
bewakes: The goal here was to compare the script pubkey, and that we won't be using addresses with different networks. Ideally, BitcoinAddress shouldn't even have a network field and should wrap a network-checked address, but that's something for later.

There is also no natural ordering of Network. So this should be fine imo.

crates/primitives/src/sorted_vec.rs
```diff
@@ -0,0 +1,140 @@
...
+            true
+        } else {
+            false
+        }
+    }
+
+    /// Checks if the `SortedVec` contains the given value.
+    pub fn contains(&self, value: &T) -> bool {
+        self.inner.binary_search(value).is_ok()
+    }
```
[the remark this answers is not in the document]
bewakes: Added a `binary_search` method to be consistent with Vec interface.

crates/btcio/src/reader/query.rs
```diff
@@ -26,42 +26,42 @@ use crate::{
...
-pub async fn bitcoin_data_reader_task<D: Database + 'static>(
+struct ReaderContext<R: Reader> {
+    /// Bitcoin reader client
+    client: Arc<R>,
+    /// L1Event sender
+    event_tx: mpsc::Sender<L1Event>,
+    /// Config
+    config: Arc<ReaderConfig>,
+    /// Status transmitter
+    status_tx: Arc<StatusTx>,
```
storopoli, line 37: I honestly think that we should not `Arc` everything in `crates/*` since they are libs.
`Arc`ing should be done by the implementor.

Hence:
```suggestion
    client: &R,
    /// L1Event sender
    event_tx: mpsc::Sender<L1Event>,
    /// Config
    config: &ReaderConfig,
    /// Status transmitter
    status_tx: &StatusTx,
```
Then `Arc`ing it in the `bin/*` stuff.

This comment applies to all `Arc`ing in the `crates/*` changes in this PR.

crates/btcio/src/reader/query.rs
```diff
@@ -26,42 +26,42 @@ use crate::{
     status::{apply_status_updates, L1StatusUpdate},
 };
 
-// TODO: remove this
-pub async fn bitcoin_data_reader_task<D: Database + 'static>(
+struct ReaderContext<R: Reader> {
```
storopoli, line 29: Missing docstring.

crates/btcio/src/reader/query.rs
```diff
@@ -26,42 +26,42 @@ use crate::{
...
+    client: Arc<R>,
+    /// L1Event sender
+    event_tx: mpsc::Sender<L1Event>,
+    /// Config
+    config: Arc<ReaderConfig>,
+    /// Status transmitter
+    status_tx: Arc<StatusTx>,
+}
+
+pub async fn bitcoin_data_reader_task(
```
storopoli, line 40: missing docstring

crates/btcio/src/reader/query.rs
```diff
@@ -26,42 +26,42 @@ use crate::{
...
-    .await
+        status_tx,
+    };
+    do_reader_task(ctx, target_next_block).await
 }
 
-async fn do_reader_task(
-    client: &impl Reader,
-    event_tx: &mpsc::Sender<L1Event>,
+async fn do_reader_task<R: Reader>(
```
storopoli, line 56: missing docstring.

crates/btcio/src/reader/query.rs
```diff
@@ -232,37 +216,37 @@ async fn find_pivot_block(
     Ok(None)
 }
 
-async fn fetch_and_process_block(
+async fn fetch_and_process_block<R: Reader>(
```
storopoli, line 219: missing docstrings

crates/primitives/src/l1.rs
```diff
@@ -237,7 +237,7 @@ pub struct L1Status {
 /// some useful traits on it such as [`serde::Deserialize`], [`borsh::BorshSerialize`] and
 /// [`borsh::BorshDeserialize`].
 // TODO: implement [`arbitrary::Arbitrary`]?
-#[derive(Debug, Clone, Serialize, PartialEq, Eq)]
+#[derive(Debug, Clone, Serialize, Eq)]
```
storopoli, line 240: Why you implemented `PartialEq` manually and still left `Eq` as derived?
Is it the case that the derived `Eq` will use the `PartialEq` manual impl?

crates/primitives/src/sorted_vec.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    /// Inserts an element into the `SortedVec`, maintaining sorted order. This runs in O(n) because
+    /// of shifting of elements.
+    pub fn insert(&mut self, value: T) {
+        let pos = self.inner.binary_search(&value).unwrap_or_else(|e| e);
+        self.inner.insert(pos, value);
+    }
+
+    /// Removes an element from the `SortedVec`. Returns `true` if the element was found and
```
storopoli, line 24:
```suggestion
    /// Removes an element from the [`SortedVec`]. Returns `true` if the element was found and
```

crates/primitives/src/sorted_vec.rs
```diff
@@ -0,0 +1,194 @@
...
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize)]
+pub struct SortedVec<T> {
+    inner: Vec<T>,
+}
+
+impl<T: Ord + Clone> SortedVec<T> {
+    /// Creates a new, empty `SortedVec`.
+    pub fn new() -> Self {
+        Self { inner: Vec::new() }
+    }
```
storopoli, line 15: We should also do an [`pub fn with_capacity`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity).

crates/primitives/src/sorted_vec.rs
```diff
@@ -0,0 +1,194 @@
+use std::cmp::Ordering;
```
storopoli, line 1: Have you checked?
https://crates.io/crates/sorted-vec or https://crates.io/crates/sortedvec ?

crates/tx-parser/src/utils.rs
```diff
@@ -80,3 +84,12 @@ pub fn generate_taproot_address(
 
     Ok(addr)
 }
+
+/// Reads the operator wallet public keys from Rollup params. Returns None if
+/// not yet bootstrapped
+/// FIXME: This is only for devnet as these pks have to be read from the chain state
```
storopoli, line 90: This should not be a docstring:
```suggestion
// FIXME: This is only for devnet as these PKs have to be read from the chain state
```

crates/tx-parser/src/filter_types.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// For addresses that we expect spends to.
+    pub expected_addrs: SortedVec<BitcoinAddress>,
+
+    /// For blobs we expect to be written.
+    pub expected_blobs: SortedVec<Buf32>,
+
+    /// For deposits that might be spent from.
+    pub expected_outpoints: SortedVec<Outpoint>,
+
+    /// Deposit config that defines the structure we expect in the utxo
```
storopoli, line 25:
```suggestion
    /// Deposit config that defines the structure we expect in the utxo.
```
The only docstring in this struct that does not have a `.` in the end. 😓
Also I don't like "we" in docstrings?
Who are "we"? Alpen Labs, Strata, Users, Developers, Sequencer, Bridge Operator, Humans?

crates/tx-parser/src/filter_types.rs
```diff
@@ -0,0 +1,61 @@
...
+};
+
+use crate::utils::{generate_taproot_address, get_operator_wallet_pks};
+
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize)]
+pub struct TxFilterConfig {
+    /// For checkpoint update inscriptions.
+    pub rollup_name: String,
+
+    /// For addresses that we expect spends to.
```
storopoli, line 16: Who are we?

crates/tx-parser/src/filter_types.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub expected_addrs: SortedVec<BitcoinAddress>,
+
+    /// For blobs we expect to be written.
```
storopoli, line 19: Who are we?

crates/tx-parser/src/filter_types.rs
```diff
@@ -0,0 +1,61 @@
...
+    buf::Buf32,
+    l1::BitcoinAddress,
+    params::{DepositTxParams, RollupParams},
+    sorted_vec::SortedVec,
+};
+
+use crate::utils::{generate_taproot_address, get_operator_wallet_pks};
+
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize)]
+pub struct TxFilterConfig {
```
storopoli, line 12: docstring missing in the struct.

crates/tx-parser/src/filter_types.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// For deposits that might be spent from.
+    pub expected_outpoints: SortedVec<Outpoint>,
+
+    /// Deposit config that defines the structure we expect in the utxo
+    pub deposit_config: DepositTxParams,
+}
+
+impl TxFilterConfig {
+    // TODO: this will need chainstate too in the future
+    pub fn derive_from(rollup_params: &RollupParams) -> anyhow::Result<Self> {
```
storopoli, line 31: docstring

crates/tx-parser/src/filter_types.rs
```diff
@@ -0,0 +1,61 @@
...
+        })
+    }
+}
+
+/// Outpoint of a bitcoin tx
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, BorshSerialize, BorshDeserialize)]
+pub struct Outpoint {
+    pub txid: Buf32,
+    pub vout: u32,
+}
```
storopoli, line 61: This should be moved to `crates/primitives/src/l1.rs`.
See example of `BitcoinAddress`: https://github.com/alpenlabs/strata/blob/77cae83ebf5319b17ae10fa8ce5df5876f2dbec0/crates/primitives/src/l1.rs#L236-L247

crates/tx-parser/src/filter.rs
```diff
@@ -1,92 +1,94 @@
...
                 .map(|relevant_tx| ProtocolOpTxRef::new(i as u32, relevant_tx))
         })
         .collect()
 }
 
 ///  if a [`Transaction`] is relevant based on given [`RelevantTxType`]s then we extract relevant
-///  info
-fn check_and_extract_relevant_info(
+///  info.
+///  TODO: make this function return multiple ops as a single tx can have multiple outpoints that's
```
storopoli, line 33: Not part of docstring:
```suggestion
// TODO: make this function return multiple ops as a single tx can have multiple outpoints that's
```

crates/tx-parser/src/filter.rs
```diff
@@ -1,92 +1,94 @@
...
+            extract_protocol_op(tx, &filter_config)
                 .map(|relevant_tx| ProtocolOpTxRef::new(i as u32, relevant_tx))
         })
         .collect()
 }
 
 ///  if a [`Transaction`] is relevant based on given [`RelevantTxType`]s then we extract relevant
-///  info
-fn check_and_extract_relevant_info(
+///  info.
```
storopoli, line 32:
```suggestion
/// info.
```
Strange padding

crates/tx-parser/src/filter.rs
```diff
@@ -1,92 +1,94 @@
...
+///  TODO: make this function return multiple ops as a single tx can have multiple outpoints that's
+///  relevant
```
storopoli, line 34:
```suggestion
// relevant
```

crates/tx-parser/src/filter.rs
```diff
@@ -1,92 +1,94 @@
...
-                    }
-                }
-            }
-            None
-        }
+    // Currently all we have are inscription txs, deposits and deposit requests
+    parse_inscription_checkpoint(tx, filter_conf)
+        .map(ProtocolOperation::Checkpoint)
+        .or_else(|| parse_deposit(tx, filter_conf).map(ProtocolOperation::Deposit))
+        .or_else(|| parse_deposit_request(tx, filter_conf).map(ProtocolOperation::DepositRequest))
```
storopoli, line 43: I really would like to encapsulate the fact that we are parsing first DTs then DRTs in another place rather than the ordering of `or_else` in a `extract_foo` function.
Can't we move this logic elsewhere?

storopoli: Maybe below in a `parse` function that uses these sequentially in a proper manner/ordering?

crates/tx-parser/src/filter.rs
```diff
@@ -1,92 +1,94 @@
...
+) -> Option<DepositRequestInfo> {
+    extract_deposit_request_info(tx, &filter_conf.deposit_config)
+}
+
+fn parse_deposit(tx: &Transaction, filter_conf: &TxFilterConfig) -> Option<DepositInfo> {
+    extract_deposit_info(tx, &filter_conf.deposit_config)
+}
+
+/// Parses inscription from the given transaction. The inscription currently recognized is the
+/// checkpoint inscription.
```
storopoli, line 80:
```suggestion
/// Parses inscription from the given transaction. Currently, the only inscription recognizable is the
/// checkpoint inscription.
```
I think this is less ambiguous and captures your intentions more clearly, no?

storopoli (changes requested):
Some points of discussions and minor nits.
I think we need to reconsider our `Arc`ing in `lib` crates and move these to `bin` crates.

We are currently shoving `Arc` down the throat of anyone who wants to use the libraries:

image

We should think this through and document in the Rust guidelines.

storopoli:
@bewakes I'm fine with keeping the `Arc`s the way they are if that's too cumbersome.
My point of view is that the `lib` crates should not enforce impl details like `Arc`/`Mutex` etc into the functions.
That is for the implementer in `bin` crates or other implementations.

For example suppose we want to create an FFI WASM module for a browser extension.
But now, we have to refactor since the function we wanted to use in a certain `lib` crate hardcodes `Arc`.

crates/primitives/src/l1.rs
bewakes: Yes, Eq is there to imply equivalence relation and implies PartialEq.

crates/primitives/src/sorted_vec.rs
bewakes: Yes, I had looked into the first one. The second one seems to use macro magic. I just wanted to keep the dependencies minimal for small added feature.

crates/tx-parser/src/filter_types.rs
bewakes: [We are the champions](https://www.youtube.com/watch?v=04854XqcfCY)!! 😄  Removed the "we" references.

crates/tx-parser/src/filter_types.rs
bewakes: I am not sure about this. The Outpoint is supposed to be more "local" than the BitcoinAddress. That's why it is here.

crates/tx-parser/src/filter.rs
bewakes: Do you mean, a single transaction can be both DT and DRT? If so then we have a [ticket](https://alpenlabs.atlassian.net/browse/STR-402) that says we should be able to parse multiple items from a transaction. So, the changes you suggested can possible done while working on that ticket.

bewakes:
Fixed the nits and other comments

crates/primitives/src/l1.rs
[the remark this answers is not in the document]
delbonis: Then it shouldn't implement `Ord`.  The `Eq` impl is also wrong since it doesn't check all fields.  Either remove the ord/eq impls or make them also check the network field.

crates/tx-parser/src/filter_types.rs
delbonis: I always use it as the instructive we that an instructor would use when working through a problem on a whiteboard.  Related to but different from the royal we.

crates/tx-parser/src/filter.rs
delbonis: We should be counting *all* of the ops that match, not just the first.

crates/primitives/src/l1.rs
```diff
@@ -352,6 +352,28 @@ impl BorshDeserialize for BitcoinAddress {
     }
 }
 
+impl PartialEq for BitcoinAddress {
+    fn eq(&self, other: &Self) -> bool {
+        self.address
+            .script_pubkey()
+            .eq(&other.address.script_pubkey())
+    }
+}
```
delbonis, line 361: This also needs to include the network field, missed it the previous time around.

delbonis (changes requested):
Also see other comments.

crates/tx-parser/src/filter_types.rs
storopoli: I think that this invites code duplication.
It is way easier to have it in `strata/crates/primitives/src/l1.rs` so that we can use it in the future if necessary.
Especially if we want to `Borsh` stuff and with the `Buf32`.

bewakes:
@delbonis , I've updated it to parse multiple inscriptions. Parsing all possible items is slightly more involved and will be done for the ticket mentioned in the comment. 
Also, I've resorted to the default implementations of `Ord`, `Eq`, for `BitcoinAddress`. How it needs to be sorted will be handled when we actually work on handling multiple spent addresses.

> I think that this invites code duplication.
It is way easier to have it in strata/crates/primitives/src/l1.rs so that we can use it in the future if necessary.

@storopoli, that's true. Thanks for pointing out. Updated accordingly.

crates/primitives/src/sorted_vec.rs
storopoli: Ok makes sense. Thanks!

storopoli (approved):
ACK ede3ac764448800ba59bb09e15c82406cadb36ff
func-tests might be fixed by a rebase.

prajwolrg (approved):
ACK https://github.com/alpenlabs/strata/commit/ede3ac764448800ba59bb09e15c82406cadb36ff

delbonis (approved):

manishbista28 (approved):

prajwolrg (dismissed):

storopoli (approved):
ACK 2b115cfa22a2885da168ab67004341a1f0162023

manishbista28 (approved):

delbonis (approved):

prajwolrg (approved):

PR #467, opened 2024-11-13
Str-613: prefund addresses in the fn_tests

evgenyzdanovich:
## Description

Enhance functional tests by exposing address utils through the common LiveEnv: get unique EL addresses and get funded btc addresses. 

### Type of Change

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Documentation update
-   [ ] Refactor
-   [x] New or updated tests

## Checklist

-   [x] I have performed a self-review of my code.
-   [x] I have commented my code where necessary.
-   [ ] I have updated the documentation if needed.
-   [x] My changes do not introduce new warnings.
-   [ ] I have added tests that prove my changes are effective or that my feature works.
-   [x] New and existing tests pass with my changes.

PR #468, opened 2024-11-13
func-tests: bridge withdraw happy

storopoli:
## Description

Adds the bridge fn-test withdraw happy path. 

### Type of Change

-   [ ] Documentation update
-   [ ] Refactor
-   [x] New or updated tests

## Checklist

## Related Issues

STR-447


functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -0,0 +1,184 @@
...
+
+        # FIXME: I don't know why the hell the math is wrong
+        #        and I need to manually add 5 zeros with 100_000
+        #
+        # lower bound is 10 BTC - 0.01 BTC (fees) - gas with a leeway
+        lower_bound = ((rollup_deposit_amount - 1_000_000) * SATS_TO_WEI * 100_000) - gas * 2
+        # upper bound is 10 BTC + 0.01 BTC (fees)
+        upper_bound = (rollup_deposit_amount + 1_000_000) * SATS_TO_WEI * 100_000
+        self.logger.debug(f"Balance lower bound: {lower_bound}")
+        self.logger.debug(f"Balance upper bound: {upper_bound}")
```
storopoli, line 110: I still need help debugging this math.
I don't understand why I had to add 5 zeros despite the fact that I was already doing the conversion with `SATS_TO_WEI`

Cc @delbonis

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -0,0 +1,184 @@
...
+
+        # Mine blocks
+        btcrpc.proxy.generatetoaddress(12, UNSPENDABLE_ADDRESS)
+        time.sleep(0.5)
+
+        # Make sure that the balance in the BTC wallet is 10 BTC-ish
+        btc_balance = get_balance(change_address, btc_url, btc_user, btc_password)
+        self.logger.debug(f"BTC balance: {btc_balance}")
+        # FIXME: This only gives me 4.97 BTC WTF?
+        assert btc_balance >= 9_900_000_000, "BTC balance is not 10 BTC"
```
storopoli, line 124: We should get `10 BTC`-ish back.
But how the hell I got less than half?
Cc @delbonis

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -0,0 +1,179 @@
...
+        self.logger.debug(f"Bridge address: {PRECOMPILE_BRIDGEOUT_ADDRESS}")
+
+        data_bytes = bytes.fromhex(change_address_pk)
+
+        transaction = {
+            "from": el_address,
+            "to": PRECOMPILE_BRIDGEOUT_ADDRESS,
+            # FIXME: I don't know why the hell the math is wrong
+            #        and I need to manually add 5 zeros with 100_000
+            "value": DEFAULT_ROLLUP_PARAMS["deposit_amount"] * SATS_TO_WEI,
```
delbonis, line 174: Is this still applicable?

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -0,0 +1,179 @@
...
+        assert balance >= lower_bound, "balance is not in the expected range"
+
+        # Mine blocks
+        btcrpc.proxy.generatetoaddress(12, UNSPENDABLE_ADDRESS)
+        time.sleep(1)
+
+        # Make sure that the balance in the BTC wallet is 10 BTC-ish
+        btc_balance = get_balance(withdraw_address, btc_url, btc_user, btc_password)
+        self.logger.debug(f"BTC balance: {btc_balance}")
+        btc_expected_balance = rollup_deposit_amount - 100_000_000
```
delbonis, line 115: Should we shift out these constants to somewhere else?

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -0,0 +1,179 @@
...
+        btc_user = btc.props["rpc_user"]
+        btc_password = btc.props["rpc_password"]
+        seq_addr = seq.get_prop("address")
+
+        # Create the deposit request transaction
+        tx = bytes(
+            deposit_request_transaction(
+                el_address, musig_bridge_pk, btc_url, btc_user, btc_password
+            )
+        ).hex()
```
delbonis, line 139: These `bytes(_).hex()` conversions seem odd, what's happening here?  Does this return some `Transaction` type natively that we then have to convert later?

delbonis (changes requested):
Some comments about comprehension.  Broadly seems fine though.

functional-tests/fn_bridge_withdraw_happy.py
storopoli: This gives raw bytes from the rust side of things. Then I convert to hex.

functional-tests/fn_bridge_withdraw_happy.py
storopoli: No, will fix

functional-tests/fn_bridge_withdraw_happy.py
storopoli: https://github.com/alpenlabs/strata/blob/ac8e8f2917441d6d68c44caa59bb28b7a049ecf5/crates/util/python-utils/src/drt.rs#L22-L52

functional-tests/fn_bridge_withdraw_happy.py
storopoli: I don't know this is just used in this withdraw case. I can make it better by using relative amounts like

```py
0.9 * rollup_deposit_amount
# or
rollup_deposit_amount // 1.1
```

storopoli: I've cleaned them up in 7a414eca5abf79b72ed921d945c06301005bca77
and also added tighter margins for lower bounds

storopoli: I've cleaned them up in 7a414eca5abf79b72ed921d945c06301005bca77
and also added tighter margins for lower bounds

functional-tests/fn_bridge_withdraw_happy.py
storopoli: Done in 7a414eca5abf79b72ed921d945c06301005bca77

delbonis (dismissed):

storopoli:
We need an approval from @alpenlabs/ci

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -0,0 +1,182 @@
...
+        self.logger.debug(f"Sent withdrawal transaction with hash: {l2_tx_hash}")
+
+        # Wait for the withdrawal to be processed
+        time.sleep(DEFAULT_BLOCK_TIME_SEC * 2)
+
+        l2_tx_state = web3.eth.get_transaction_receipt(l2_tx_hash)
+        self.logger.debug(f"Transaction state (web3): {l2_tx_state}")
+
+        # Make sure that the balance is zero
+        balance = int(rethrpc.eth_getBalance(el_address), 16)
```
Rajil1213, line 106: This should just check if the difference in balance before and after the withdrawal equals the `DEPOSIT_DENOMINATION` - `gas`. Doing it this way makes the tests a bit flaky (for example, if we want to test multiple withdrawals).

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -0,0 +1,182 @@
...
+        self.logger.debug(f"Strata Balance after withdrawal: {balance}")
+
+        lower_bound = STRATA_DEPOSIT_LOWER_BOUND
+        self.logger.debug(f"Strata Balance lower bound: {lower_bound}")
+
+        assert balance >= lower_bound, "balance is not in the expected range"
```
Rajil1213, line 112: This lower bound check shouldn't be necessary. At this point in the test, all we need to assert is that the difference in balance is equal to the `DEPOSIT_DENOMINATION` - `gas_fees`. In the current scenario, the fees are guaranteed to be constant (`22_000`).

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -0,0 +1,182 @@
...
+
+# Local constants
+# D BTC
+ROLLUP_DEPOSIT_AMOUNT = DEFAULT_ROLLUP_PARAMS["deposit_amount"]
+# Gas for the withdrawal transaction
+GAS = 22_000
+# lower bound is D BTC - fees - gas with a leeway
+STRATA_DEPOSIT_LOWER_BOUND = ((ROLLUP_DEPOSIT_AMOUNT // 1.01) * SATS_TO_WEI) - (
+    2 * GAS * GWEI_TO_WEI
+)
```
Rajil1213, line 32: I don't think is entirely accurate. Where does the `1.01` come from (this magic number should either be a descriptive constant or explained in the comment)?

The only place I've seen this number is in the faucet but that should not matter for the withdrawal because the deposit amount is always exactly 10 BTC.

I expect the calculation to simply be the following (but do correct me if I'm missing something):

```suggestion
# lower bound is D BTC - fees
STRATA_DEPOSIT_LOWER_BOUND = ROLLUP_DEPOSIT_AMOUNT * SATS_TO_WEI - GAS * GWEI_TO_WEI
```

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -0,0 +1,182 @@
...
+        assert balance >= lower_bound, "balance is not in the expected range"
+
+        # Mine blocks
+        btcrpc.proxy.generatetoaddress(12, UNSPENDABLE_ADDRESS)
+        time.sleep(1)
+
+        # Make sure that the balance in the BTC wallet is D BTC-ish
+        btc_balance = get_balance(withdraw_address, btc_url, btc_user, btc_password)
+        self.logger.debug(f"BTC balance: {btc_balance}")
+        btc_balance_lower_bound = ROLLUP_DEPOSIT_AMOUNT // 1.075
```
Rajil1213, line 121: Where does the `1.075` value come from? This is a case where a lower bound makes sense. But even here the fees are almost guaranteed to be constant. There is an `OPERATOR_FEE` constant somewhere in the code that should be used here and the transaction fees are also going to be more or less the same in the test environment. The assertion should again be on the difference in balance before and after the withdrawal confirmation.

Rajil1213 (changes requested):
Apologies for the late review.

Requesting some changes and clarification for some of the assertions here.

functional-tests/fn_bridge_withdraw_happy.py
delbonis: That magic 1.01 number was the old value for the deposit denomination in the test env for some reason, we changed that to just use 10 BTC.

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -0,0 +1,192 @@
...
+        l2_tx_state = web3.eth.get_transaction_receipt(l2_tx_hash)
+        self.logger.debug(f"Transaction state (web3): {l2_tx_state}")
+
+        # Make sure that the balance is less than the initial balance
+        balance_post_withdraw = int(rethrpc.eth_getBalance(el_address), 16)
+        self.logger.debug(f"Strata Balance after withdrawal: {balance_post_withdraw}")
+        difference = balance_post - balance_post_withdraw
+        self.logger.debug(f"Strata Balance difference: {difference}")
+        # NOTE: Difference is D BTC - gas_fee
+        assert difference == 10_000_023_000_818_888_144, "balance difference is not expected"
```
Rajil1213, line 122: It's more ergonomic to have descriptive constants here rather than comments. Right now, the tests are failing at this point and the error message just prints this line out (without the comment). We'd know exactly what failed if this line said: `assert difference == DEPOSIT_AMOUNT - WITHDRAWAL_GAS_FEE` instead.

You might find this interesting: https://youtu.be/Bf7vDBBOBUA?si=kQQbb18nJ69MCplM

functional-tests/fn_bridge_withdraw_happy.py
Rajil1213: Might also make sense here to assert on the upper bound. The balance should be in the range: $(0, D - (fee_{operator} + fee_{miner})]$.

Rajil1213 (changes requested):
Left some comment on the failing test and a review from the earlier round that hasn't been addressed yet.

functional-tests/fn_bridge_withdraw_happy.py
AaronFeickert: Agreed. Much cleaner and future proof to define constants and use arithmetic expressions where needed (unless the expression evaluation itself were to be prohibitively expensive, which it isn't here).

storopoli: Rajil thanks for the review, and yes you're totally correct.
I will mark this as a draft and work on it this week.
Will mark as ready once I think is in a bulletproof state to not waste review time.
Thanks!

functional-tests/fn_bridge_withdraw_happy.py
storopoli: Because it returns a list of bytes... Since it is a Vector underneath.

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -0,0 +1,239 @@
...
+        bridge_pk = get_bridge_pubkey(seqrpc)
+
+        # Generate plenty of BTC to address
+        btcrpc.proxy.generatetoaddress(102, address)
+
+        # Deposit to the EL address
+        # NOTE: we need 2 deposits to make sure we have funds for gas
+        self.make_drt(ctx, el_address, bridge_pk)
+        self.make_drt(ctx, el_address, bridge_pk)
+        wait_until(lambda: int(rethrpc.eth_getBalance(el_address), 16) > 0)
```
bewakes, line 121: I think just having a `> 0` might not account for both DRT to be processed. Maybe we can be certain with two wait_until checks. For practical purpose, this should probably be fine though.

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -0,0 +1,239 @@
...
+        # Make sure that the balance is expected
+        balance_post_withdraw = int(rethrpc.eth_getBalance(el_address), 16)
+        self.logger.debug(f"Strata Balance after withdrawal: {balance_post_withdraw}")
+        difference = DEPOSIT_AMOUNT * SATS_TO_WEI - total_gas_used
+        self.logger.debug(f"Strata Balance difference: {difference}")
+        assert difference == balance_post_withdraw, "balance difference is not expected"
+
+        # Mine blocks
+        btcrpc.proxy.generatetoaddress(12, UNSPENDABLE_ADDRESS)
+        time.sleep(1)
```
bewakes, line 152: Ideally this would be replaced by wait_until to indicate what exactly we are sleeping/waiting here for, and also to prevent longer waits if the condition meets early

bewakes (commented):
Ideally we would avoid `time.sleep()` calls with `wait_until` to pass early and also to indicate what/why we are waiting. The rest looks good.

functional-tests/fn_bridge_withdraw_happy.py
storopoli: Done in a663e7424a18213078973394beca7685ab05d45f.

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -0,0 +1,239 @@
...
+        # Send funds to the bridge precompile address for a withdrawal
+        change_address_pk = extract_p2tr_pubkey(withdraw_address)
+        self.logger.debug(f"Change Address PK: {change_address_pk}")
+        estimated_withdraw_gas = self.estimate_withdraw_gas(ctx, change_address_pk)
+        self.logger.debug(f"Estimated withdraw gas: {estimated_withdraw_gas}")
+        l2_tx_hash = self.make_withdraw(ctx, change_address_pk, estimated_withdraw_gas).hex()
+        self.logger.debug(f"Sent withdrawal transaction with hash: {l2_tx_hash}")
+
+        # Wait for the withdrawal to be processed
+        time.sleep(DEFAULT_BLOCK_TIME_SEC * 2)
```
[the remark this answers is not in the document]
storopoli: Done in a663e7424a18213078973394beca7685ab05d45f.

functional-tests/fn_bridge_withdraw_happy.py
storopoli: Done in a663e7424a18213078973394beca7685ab05d45f.

delbonis (approved):

Rajil1213 (approved):
Looks good. Ideally, we don't _have_ to care about the initial balance in the EL address  because we only care about the difference but since this simplifies the test WLOG, this is fine too.

storopoli:
#467 with this PR on `main` broke CI. Working on a fix.

PR #469, opened 2024-11-13
STR-929 Core STF refactors

delbonis:
## Description

The goal of this PR is to change the OL bookkeeping so we're only actually checking in with the L1 at the end of the epoch when we're preparing the epoch checkpoint.

This is incomplete, and may require some storage refactors if we go deep enough.  Not sure there yet.

### Type of Change

-   [ ] Documentation update
-   [x] Refactor
-   [ ] New or updated tests

## Checklist


bewakes (changes requested):

crates/chaintsn/src/epoch.rs
```diff
@@ -0,0 +1,246 @@
...
+    let num_deposit_ents = state.state().deposits_table().len();
+
+    // This determines how long we'll keep trying to service a withdrawal before
+    // updating it or doing something else with it.  This is also what we use
+    // when we decide to reset an assignment.
+    let cur_block_height = state.l1_safe_height();
+    let new_exec_height = cur_block_height as u32 + params.dispatch_assignment_dur;
+
+    // Sequence in which we assign the operators to the deposits.  This is kinda
+    // shitty because it might not account for available funds but it works for
```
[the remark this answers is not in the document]
storopoli:
```suggestion
    // suboptimal because it might not account for available funds but it works for
```

functional-tests/utils.py
```diff
@@ -102,6 +102,19 @@ def wait_until_with_value(
...
+    """Waits until the current checkpoint index increases."""
+    first_ckpt_idx = seqrpc.strata_getLatestCheckpointIndex()
+
+    def _f():
+        status = seqrpc.strata_syncStatus()
+        print("waiting for epoch, at chain tip slot", status)
+        cur_ckpt_index = seqrpc.strata_getLatestCheckpointIndex()
+        return cur_ckpt_idx > first_ckpt_idx
+
+    wait_until_with_value(_f, lambda v: v, "Epoch never finalized", timeout, step)
```
bewakes, line 115: The caller doesn't seem to use the returned value, so could just be simplified with `wait_until`.
```python
wait_until(_f, "Error never vinalizesd", timeout, step)
```

crates/chaintsn/src/clock.rs
```diff
@@ -0,0 +1,26 @@
...
+    slot / params.target_l2_batch_size as u64
+}
+
+pub fn get_epoch_initial_slot(epoch: u64, params: &RollupParams) -> u64 {
+    epoch * params.target_l2_batch_size
+}
+
+pub fn get_epoch_final_slot(epoch: u64, params: &RollupParams) -> u64 {
+    let epoch_init_slot = get_epoch_initial_slot(epoch, params);
+    epoch_init_slot + (params.target_l2_batch_size - 1)
```
bewakes, line 15: This is simpler?
```suggestion
    get_epoch_initial_slot(epoch + 1, params) - 1
```

crates/chaintsn/src/epoch.rs
```diff
@@ -0,0 +1,381 @@
...
+        // TODO make sure that the block hashes all connect up sensibly.
+
+        // Collect the indexes of current operators.
+        let op_idxs: Vec<_> = state.epoch_state().operator_table().indices().collect();
+
+        for e in l1seg.new_payloads() {
+            // TODO maybe consolidate these to have ops for each tx
+
+            for tx in e.deposit_update_txs() {
+                if let ProtocolOperation::Deposit(deposit_info) = tx.tx().protocol_operation() {
```
bewakes, line 76: Asking for deposit_update_txs and then checking if the pattern match and `.tx().protocol_operation()` look a bit weird, the interface could be simpler I guess, just get the deposit infos from payload.

crates/chaintsn/src/epoch.rs
```diff
@@ -0,0 +1,381 @@
...
+        let have_ready_intent = next_intent_to_assign < ready_withdrawals_cnt;
+
+        match ent.deposit_state() {
+            DepositState::Created(_) => {
+                // TODO I think we can remove this state
+            }
+
+            DepositState::Accepted => {
+                // If we have an intent to assign, we can dispatch it to this deposit.
+                if have_ready_intent {
```
bewakes, line 172: Can move this to match line to avoid one nesting

crates/chaintsn/src/epoch.rs
```diff
@@ -0,0 +1,381 @@
...
+                        new_exec_height as u64,
+                    );
+
+                    next_intent_to_assign += 1;
+                }
+            }
+
+            DepositState::Dispatched(dstate) => {
+                // Check if the deposit is past the threshold.
+                if cur_block_height >= dstate.exec_deadline() {
```
bewakes, line 195: Same for this maybe?

bewakes (commented):
Some nit comments. I think the `process_deposit_updates` function can be made better with helper functions.

functional-tests/utils.py
delbonis: Oh good point, I didn't realize what I was doing.

crates/state/src/chain_state.rs
```diff
@@ -30,37 +30,16 @@ pub struct Chainstate {
...
+    /// Epoch-level state that we only change as of the last block of an epoch.
+    // TODO this might be reworked to be managed separately
+    pub(crate) epoch_state: EpochState,
 
     /// Pending withdrawals that have been initiated but haven't been sent out.
     pub(crate) pending_withdraws: StateQueue<bridge_ops::WithdrawalIntent>,
 
     /// Execution environment state.  This is just for the single EE we support
     /// right now.
     pub(crate) exec_env_state: exec_env::ExecEnvState,
```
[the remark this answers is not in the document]
delbonis: Perhaps, could you expand on what the value of this would be?  I'm trying to be as data-oriented as possible here since I anticipate some refactors in the future here when we support async EE consensus.

delbonis:
@bewakes Appreciate the comments.  Going to take a while to circle back to most of these since I'm working on the more tricky areas right now for stuff that breaks due to the consensus/bridging refactors.

delbonis:
Monster rebase.

delbonis:
Closing since we merged the other one.

PR #477, opened 2024-11-15
func-test: bridge deposit reclaim operator seen DRT

storopoli:
## Description

Test case same as bridge deposit reclaim but now the operators have a high `message_interval` and cannot coordinate to spend the DRT into a DT.

### Type of Change

-   [ ] Documentation update
-   [ ] Refactor
-   [x] New or updated tests

## Checklist

## Related Issues

STR-624

### DIFF
--- a/functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
+++ b/functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
@@ -1,5 +1,3 @@
-import time
-
 import flexitest
 from bitcoinlib.services.bitcoind import BitcoindClient
 from strata_utils import (
@@ -14,7 +12,7 @@ from constants import (
     DEFAULT_TAKEBACK_TIMEOUT,
     UNSPENDABLE_ADDRESS,
 )
-from utils import get_bridge_pubkey, get_logger
+from utils import get_bridge_pubkey, get_logger, wait_until
 
 # Local constants
 # D BTC
@@ -56,8 +54,13 @@ class BridgeDepositReclaimDrtSeenTest(flexitest.Test):
 
         bridge_pk = get_bridge_pubkey(seqrpc)
         el_address = ctx.env.gen_el_address()
+        user_addr = ctx.env.gen_ext_btc_address()
         refund_addr = ctx.env.gen_ext_btc_address()
         recovery_addr = ctx.env.gen_rec_btc_address()
+        self.logger.debug(f"El address: {el_address}")
+        self.logger.debug(f"User address: {user_addr}")
+        self.logger.debug(f"Refund address: {refund_addr}")
+        self.logger.debug(f"Recovery address: {recovery_addr}")
 
         # First let's see if the EVM side has no funds
         # Make sure that the el_address has zero balance
@@ -76,7 +79,6 @@ class BridgeDepositReclaimDrtSeenTest(flexitest.Test):
         # DRT same block
         txid_drt = self.make_drt(ctx, el_address, bridge_pk)
         self.logger.debug(f"Deposit Request Transaction: {txid_drt}")
-        time.sleep(1)
 
         # Now we need to generate a bunch of blocks
         # since they will be able to spend the DRT output.
@@ -90,10 +92,11 @@ class BridgeDepositReclaimDrtSeenTest(flexitest.Test):
             self.logger.debug(f"Generating {blocks_to_generate} blocks in chunk {i+1}/{chunks}")
             btcrpc.proxy.generatetoaddress(blocks_to_generate, UNSPENDABLE_ADDRESS)
 
-        # wait up a little bit
-        time.sleep(0.25)
-
         # Make sure that the BTC refund address has the expected balance
+        wait_until(
+            lambda: get_balance(refund_addr, btc_url, btc_user, btc_password)
+            == initial_refund_btc_balance
+        )
         refund_btc_balance = get_balance(
             refund_addr,
             btc_url,
@@ -110,10 +113,20 @@ class BridgeDepositReclaimDrtSeenTest(flexitest.Test):
         self.logger.debug("Take back tx generated")
 
         # Send the transaction to the Bitcoin network
+        original_recovery_balance = get_balance_recovery(
+            recovery_addr,
+            bridge_pk,
+            btc_url,
+            btc_user,
+            btc_password,
+        )
         txid = btcrpc.proxy.sendrawtransaction(take_back_tx)
         self.logger.debug(f"sent take back tx with txid = {txid} for address {el_address}")
         btcrpc.proxy.generatetoaddress(2, UNSPENDABLE_ADDRESS)
-        time.sleep(1)
+        wait_until(
+            lambda: get_balance_recovery(recovery_addr, bridge_pk, btc_url, btc_user, btc_password)
+            < original_recovery_balance
+        )
 
         # Make sure that the BTC recovery address has 0 BTC
         btc_balance = get_balance_recovery(
@@ -166,6 +179,5 @@ class BridgeDepositReclaimDrtSeenTest(flexitest.Test):
         # Send the transaction to the Bitcoin network
         txid = btcrpc.proxy.sendrawtransaction(tx)
         self.logger.debug(f"sent deposit request with txid = {txid} for address {el_address}")
-        time.sleep(1)  # I don't know how to get rid of this sleep
 
         return txid
--- a/functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
+++ b/functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
@@ -14,57 +14,72 @@ from strata_utils import (
 from constants import (
     DEFAULT_ROLLUP_PARAMS,
     DEFAULT_TAKEBACK_TIMEOUT,
-    SATS_TO_WEI,
     UNSPENDABLE_ADDRESS,
 )
 from entry import BasicEnvConfig
-from utils import get_bridge_pubkey, get_logger
+from utils import get_bridge_pubkey, get_logger, wait_until
 
 # Local constants
-ROLLUP_DEPOSIT_AMOUNT = DEFAULT_ROLLUP_PARAMS["deposit_amount"]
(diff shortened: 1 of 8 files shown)


functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
```diff
@@ -0,0 +1,292 @@
...
+        self.logger.debug(f"DRT BTC balance (after generating blocks): {btc_recovery_balance_post}")
+
+        # DRT same block
+        txid_drt = self.make_drt(ctx, el_address, bridge_pk, maturity=0)
+        self.logger.debug(f"Deposit Request Transaction: {txid_drt}")
+        time.sleep(1)
+
+        # Kill the operator so the bridge does not process the DRT transaction
+        # NOTE: The operator is aware of the DRT transaction before stopping
+        bridge_operator.stop()
```
Rajil1213, line 137: Instead of stopping the bridge which may or may not cause the deposit transaction to be published, we can set the `--message-interval` flag when instantiating the bridge client. This will require changing the bridge factory to allow that flag to be passed. Setting the value of this flag to a high-enough value means that the nonces and signatures will be polled very slowly which would delay the signature aggregation and the Deposit Transaction would not be published. You can take the additional measure to stop the bridge operators too.

functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
```diff
@@ -0,0 +1,292 @@
...
+
+    It tests the functionality of broadcasting a deposit request transaction (DRT) for the bridge
+    and verifying that the reclaim path works as expected.
+
+    In the test we stop one of the bridge operators to prevent the DRT from being processed.
+    However, we let the operator be aware of the DRT before stopping it.
+    """
+
+    def __init__(self, ctx: flexitest.InitContext):
+        ctx.set_env(BasicEnvConfig(pre_generate_blocks=101))
```
Rajil1213, line 41: For the purposes of this test, you might want to set `auto_generate_blocks` to `False` so that there is more granular control over how many blocks are generated and when.

functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
```diff
@@ -0,0 +1,292 @@
...
+            btc_user,
+            btc_password,
+        )
+        self.logger.debug(f"Refund BTC balance (after takeback): {refund_btc_balance_post}")
+        expected_balance_lower_bound = btc_recovery_balance_pre - 100_000_000
+        self.logger.debug(f"Expected balance lower bound: {expected_balance_lower_bound}")
+        assert (
+            refund_btc_balance_post >= expected_balance_lower_bound
+        ), "BTC balance is not as expected"
+
```
Rajil1213, line 241: Shouldn't the expected balance be `DRT amount (1.01 BTC) - transaction fees`? Specifically, what makes the refund address to have a balance greater than this amount?

functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
```diff
@@ -0,0 +1,292 @@
...
+        # Now let's see if the EVM side has no funds
+        # Make sure that the el_address has zero balance
+        balance = int(rethrpc.eth_getBalance(f"0x{el_address}"), 16)
+        self.logger.debug(f"EVM balance (after takeback): {balance}")
+        self.logger.debug(f"Expected balance lower bound: {STRATA_DEPOSIT_LOWER_BOUND}")
+        assert balance >= STRATA_DEPOSIT_LOWER_BOUND, "EVM balance is not as expected"
```
Rajil1213, line 247: Shouldn't this assert that the balance in the EVM address is zero? A better approach is to check the difference in the EVM balance before and after the DRT is published (that difference must be zero).

Rajil1213 (changes requested):
Left some questions and requesting a few changes.

functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
storopoli: Done in 923d4ffea9601a37f8e044ba953340e80176104a

functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
storopoli: Done in 2b2cc8e6db5166e7342d03ae12c41dfa8917c1ac

functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
storopoli: Done in 2b2cc8e6db5166e7342d03ae12c41dfa8917c1ac

delbonis (approved):

functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
```diff
@@ -0,0 +1,230 @@
...
+            bridge_pk,
+            btc_url,
+            btc_user,
+            btc_password,
+        )
+        self.logger.debug(f"DRT BTC balance (before DRT): {btc_recovery_balance}")
+        assert btc_recovery_balance == 0, "Recovery BTC balance is not zero"
+
+        # Generate Plenty of BTC to address for the DRT
+        self.logger.debug(f"Generating 102 blocks to address: {addr}")
```
evgenyzdanovich, line 117: feel free to use the "funded addresses" functionality, if it can help here. #467

evgenyzdanovich (approved):
Look good to me, one minor suggestion

Rajil1213 (changes requested):
Looks good overall.

The only change that I'd like to see is replacing the `time.sleep()` calls with `wait_until` as suggested by @bewakes (in this PR or somewhere else). I suspect that we're doing some unnecessary `sleep` calls in some place (for example, waiting inside the `make_drt` and then waiting again after the call). Having `wait_until` helps us reason about what we're waiting for and might help speed up the test in CI by surfacing redundant calls.

functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
storopoli: Done in 421a23dca071bcb1e198b18fb3ec6adbab752a7f

functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
```diff
@@ -0,0 +1,171 @@
...
+            deposit_request_transaction(
+                el_address, musig_bridge_pk, btc_url, btc_user, btc_password
+            )
+        ).hex()
+        self.logger.debug(f"Deposit request tx: {tx}")
+
+        # Send the transaction to the Bitcoin network
+        txid = btcrpc.proxy.sendrawtransaction(tx)
+        self.logger.debug(f"sent deposit request with txid = {txid} for address {el_address}")
+        time.sleep(1)  # I don't know how to get rid of this sleep
```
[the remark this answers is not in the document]
storopoli: That would be ideal.
But this transaction is not made "in-wallet" with respect to `bitcoind`.
It is made in the `python-strata-utils`.
Hence I cannot use [`gettransaction`](https://developer.bitcoin.org/reference/rpc/gettransaction.html) and [`getrawtransaction`](https://developer.bitcoin.org/reference/rpc/getrawtransaction.html) only works with mempool transactions, this is already mined.

That's why I don't know how to get rid of the `time.sleep`.

Do you have a good suggestion?

functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
```diff
@@ -0,0 +1,171 @@
...
+            btc_url,
+            btc_user,
+            btc_password,
+        )
+        self.logger.debug(f"Initial refund BTC balance: {initial_refund_btc_balance}")
+
+        # DRT same block
+        txid_drt = self.make_drt(ctx, el_address, bridge_pk)
+        self.logger.debug(f"Deposit Request Transaction: {txid_drt}")
+        time.sleep(1)
```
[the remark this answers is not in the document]
storopoli: No, we can remove it. We are duplicating it anyways.

functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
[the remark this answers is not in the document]
Rajil1213: Does it not work when you set `-txindex=1` when you instantiate the test bitcoind instance? Also, isn't DRT signed with a wallet available on the bitcoind instance? Easiest would be to use the same bitcoind instance but separate wallets so that we can do this. Another option would be to query the current best block height after publishing DRT  and then wait for some number of blocks to be published. 

But if none of these are feasible, leaving a comment regarding what we're waiting for here would be okay, I guess.

functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
```diff
@@ -0,0 +1,171 @@
...
+
+        # Make sure that the BTC refund address has the expected balance
+        refund_btc_balance = get_balance(
+            refund_addr,
+            btc_url,
+            btc_user,
+            btc_password,
+        )
+        self.logger.debug(f"Refund BTC balance (before takeback): {refund_btc_balance}")
+        assert refund_btc_balance == initial_refund_btc_balance, "BTC balance has changed"
```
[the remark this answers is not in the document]
storopoli: I got rid of all `time.sleep` in f84f072913c902683ae2caf644eb3b37f010b014
🎉

functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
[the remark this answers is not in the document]
storopoli: I got rid of all `time.sleep` in f84f072913c902683ae2caf644eb3b37f010b014
🎉

functional-tests/fn_bridge_deposit_reclaim_drt_seen.py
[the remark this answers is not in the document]
storopoli: I got rid of all `time.sleep` in f84f072913c902683ae2caf644eb3b37f010b014
🎉

Rajil1213 (approved):
Looks great! Thanks for all the effort.

PR #481, opened 2024-11-20
Str 542 zkvm native mode

prajwolrg:
### Type of Change

-   [ ] Documentation update
-   [ ] Refactor
-   [ ] New or updated tests

## Checklist


provers/tests/src/hosts/native.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+pub fn checkpoint() -> NativeHost {
+    NativeHost {
+        process_proof: Arc::new(Box::new(move |zkvm: &NativeMachine| {
+            process_checkpoint_proof_outer(zkvm, &[0u32; 8], &[0u32; 8]);
+            Ok(())
+        })),
+    }
+}
```
MdTeach, line 62: In native mode, a mock vk using &[0u32; 8] is passed. It is somewhat hard to read initially; perhaps it could be replaced with a constant, such as MOCK_VK, to enhance readability.

crates/zkvm/adapters/native/src/host.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    fn extract_borsh_public_output<T: borsh::BorshDeserialize>(proof: &Proof) -> ZkVmResult<T> {
+        Ok(borsh::from_slice(proof.as_bytes()).expect("ser"))
+    }
+
+    fn extract_serde_public_output<T: serde::Serialize + serde::de::DeserializeOwned>(
+        proof: &Proof,
+    ) -> ZkVmResult<T> {
+        Ok(bincode::deserialize(proof.as_bytes()).expect("ser"))
+    }
```
MdTeach, line 45: Instead of expect can we use `ZkVmError` ?

crates/zkvm/adapters/native/src/host.rs
```diff
@@ -0,0 +1,60 @@
...
+            vec![]
+        } else {
+            output_ref[0].clone()
+        };
+        Ok((Proof::new(output), VerificationKey::new(vec![])))
+    }
+
+    fn get_verification_key(&self) -> VerificationKey {
+        VerificationKey::new(vec![])
+    }
```
MdTeach, line 35: Since we are using empty/mock verification key it would it be simpler if we could have `VerificationKey::default()`

crates/zkvm/adapters/native/src/host.rs
```diff
@@ -0,0 +1,60 @@
...
+        prover_input: <Self::Input<'a> as ZkVmInputBuilder<'a>>::Input,
+        _proof_type: ProofType,
+    ) -> ZkVmResult<(Proof, VerificationKey)> {
+        (self.process_proof)(&prover_input)?;
+        let output_ref = prover_input.output.borrow();
+
+        let output = if output_ref.is_empty() {
+            vec![]
+        } else {
+            output_ref[0].clone()
```
MdTeach, line 28: Why are we indexing the first element only `output_ref[0]` ?
Will it fail if there were multiple outputs ?

crates/zkvm/adapters/native/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "strata-native-zkvm-adapter"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+bincode.workspace = true
+borsh.workspace = true
+serde.workspace = true
+strata-zkvm.workspace = true
+tracing.workspace = true
```
storopoli, line 11: I prefer to order them as internal vs external:
```suggestion
strata-zkvm.workspace = true

bincode.workspace = true
borsh.workspace = true
serde.workspace = true
tracing.workspace = true
```

crates/zkvm/adapters/risc0/Cargo.toml
```diff
@@ -6,6 +6,7 @@ version = "0.1.0"
 [dependencies]
 bincode.workspace = true
 borsh.workspace = true
+hex.workspace = true
```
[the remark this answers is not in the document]
prajwolrg: Sure.

provers/tests/Cargo.toml
```diff
@@ -13,6 +13,7 @@ num-bigint = "0.4.6"
 num-traits = "0.2.19"
 serde_json.workspace = true
 sp1-sdk = "3.0.0"
+strata-native-zkvm-adapter.workspace = true
 strata-primitives.workspace = true
 strata-proofimpl-btc-blockspace.workspace = true
 strata-proofimpl-checkpoint.workspace = true
```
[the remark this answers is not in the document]
prajwolrg: Done. Also, right now I've separated optional features below. Is there a better way to organize this?

crates/zkvm/zkvm/src/host.rs
```diff
@@ -1,18 +1,15 @@
+use std::fmt::Display;
+
 use borsh::BorshDeserialize;
 use serde::{de::DeserializeOwned, Serialize};
 
 use crate::{input::ZkVmInputBuilder, proof::Proof, ProofType, VerificationKey, ZkVmResult};
 
 /// A trait implemented by the prover ("host") of a zkVM program.
-pub trait ZkVmHost: Send + Sync + Clone + 'static {
+pub trait ZkVmHost: Send + Sync + Clone + Display + 'static {
```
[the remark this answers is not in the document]
prajwolrg: The host contains the ELF file which would make `#[derive(Debug)]` a bit messy.
I guess we can implement the `Debug` trait manually as done for the `Display`.

crates/zkvm/adapters/native/src/host.rs
prajwolrg: Updated this so that it only has a single output. This is similar to [Risc0 journal](https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/receipt.rs#L274) and [SP1 public values buffer](https://github.com/succinctlabs/sp1/blob/dev/crates/primitives/src/io.rs#L9). 

Since all our guest programs uses a single output, this should be fine for now. In case we need multiple outputs, we can revisit this.

crates/zkvm/adapters/native/src/input.rs
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    fn write_proof(&mut self, item: strata_zkvm::AggregationInput) -> ZkVmResult<&mut Self> {
+        // TODO: figure this out
+        self.write_buf(item.proof().as_bytes())
+    }
+
+    fn build(&mut self) -> ZkVmResult<Self::Input> {
+        Ok(self.0.clone())
+    }
```
delbonis, line 36: This is relatively minor but since we clone in `.build` and *also* make a copy in `.write_buf` then we have to make *two* copies of any buffer we pass to `.write_buf`.  Ideally in the builder we could just store a ref to the buf passed, but I can see that this interferes with how the other write fns here work because they only own the copy in the scope of the function.  Not sure the best way to proceed with that.  At the very least you could require passing a `Vec<u8>` to try to avoid the extra clone.  Maybe similar with adding a `.into_bytes()` on `AggregationInput`?

This probably doesn't make any measurable performance difference since the only place it even *might* matter is the L1 proofs, which we only do once every 10 mins.

crates/zkvm/adapters/native/src/zkvm.rs
```diff
@@ -0,0 +1,90 @@
+use std::cell::{Cell, RefCell};
+
+use strata_zkvm::{Proof, ZkVmEnv};
+
+#[derive(Debug, Clone)]
+/// A native implementation of the [`ZkVmEnv`]
```
delbonis, line 6: Comments before attributes. :)

crates/zkvm/adapters/native/src/zkvm.rs
```diff
@@ -0,0 +1,90 @@
...
+    ///
+    /// Uses `Cell` for interior mutability, allowing `input_ptr` to be incremented within methods
+    /// that have an immutable reference to `self`. This keeps track of the next input to read.
+    pub input_ptr: Cell<usize>,
+
+    /// A vector for collecting serialized output data chunks.
+    ///
+    /// Wrapped in a `RefCell` to allow mutable access even when `self` is immutable. This stores
+    /// the outputs produced.
+    pub output: RefCell<Vec<u8>>,
```
delbonis, line 27: Why is it important that this uses interior mutability instead of taking `&mut self` on methods?  Would it be cleaner to move the mutable parts of the impl into an interior container to encapsulate it better?

delbonis (changes requested):
A few questions about design choices but broadly looks good.  Maybe most of those aren't actually actionable.

crates/zkvm/adapters/native/src/zkvm.rs
prajwolrg: Oops, my bad.

crates/zkvm/adapters/native/src/zkvm.rs
prajwolrg: `&mut self` is not required by both Risc0, SP1. They internally use sys-calls to write and read between host and the ZkVm Env. 

Are you suggesting an interior state as below?
```rust
/// Encapsulates the mutable state of the NativeMachine.
pub struct NativeMachineState {
    pub input_ptr: Cell<usize>,
    pub output: RefCell<Vec<u8>>,
}

pub struct NativeMachine {
    pub input: Vec<Vec<u8>>,
    pub state: NativeMachineState,
}
```

crates/zkvm/adapters/native/src/input.rs
prajwolrg: Not sure on how to proceed. Leaving it as is for now.

crates/zkvm/adapters/native/src/zkvm.rs
delbonis: More like

```rs

prajwolrg: This looks cleaner. Thanks.

delbonis (approved):

MdTeach (approved):

storopoli (approved):
ACK 20f52dbbb86f4b3e119ce159711664448945123e

PR #484, opened 2024-11-24
ProverManager: Avoid Cloning VM for Each Proof Request

MdTeach:
## Description
In the current implementation, the ZkVm host is cloned for each proof request. This results in significant  memory overhead, especially in the SP1 host, where the proving key is stored. 

### Type of Change

-   [ ] Documentation update
-   [x] Refactor
-   [ ] New or updated tests

## Checklist

### DIFF
--- a/bin/prover-client/src/primitives/vms.rs
+++ b/bin/prover-client/src/primitives/vms.rs
@@ -13,7 +13,7 @@ pub enum ProofVm {
 }
 
 pub struct ZkVMManager<Vm: ZkVmHost> {
-    vms: HashMap<ProofVm, Vm>,
+    vms: HashMap<ProofVm, &'static Vm>,
     prover_config: ProverOptions,
 }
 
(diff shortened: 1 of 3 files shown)


bin/prover-client/src/primitives/vms.rs
```diff
@@ -13,7 +13,7 @@ pub enum ProofVm {
 }
 
 pub struct ZkVMManager<Vm: ZkVmHost> {
-    vms: HashMap<ProofVm, Vm>,
+    vms: HashMap<ProofVm, &'static Vm>,
```
AaronFeickert, line 16: Are there any expected cases where this lifetime won't be valid?

MdTeach: I don't think it will be invalid.

ZkVmHost is loaded once as:
https://github.com/alpenlabs/strata/blob/84752e3ba0c367876dacc03f2448532df9ffa50d/bin/prover-client/src/prover.rs#L187-L193

The instances in `vms: HashMap<ProofVm, &'static Vm>` are expected to persist for the entire program’s lifetime. 
Based on the proof request i.e. `ProofVm` the prover worker can be provided with a reference to `Vm: ZkVmHost`

prajwolrg (approved):

bin/prover-client/src/primitives/vms.rs
storopoli: Even if it won't be invalid. The compiler will enforce it, hence it will be a compile error rather than a runtime error.

storopoli: But I guess this PR is one of the many that we need to do to make the prover faster.
Correct me if I am wrong.

PR #485, opened 2024-11-26
Refactor status types

bewakes:
## Description
This PR updates the interfaces of status types that were leaking implementation details. It also makes FCM update the Chainstate status every time a new block is accepted.

### Type of Change

-   [ ] Documentation update
-   [x] Refactor
-   [ ] New or updated tests

## Checklist

### DIFF
--- a/crates/btcio/src/status.rs
+++ b/crates/btcio/src/status.rs
@@ -1,8 +1,5 @@
-use std::sync::Arc;
-
 use bitcoin::Txid;
-use strata_status::StatusTx;
-use tracing::error;
+use strata_status::StatusChannel;
 
 #[derive(Debug, Clone)]
 pub enum L1StatusUpdate {
@@ -15,9 +12,9 @@ pub enum L1StatusUpdate {
     IncrementInscriptionCount,
 }
 
-pub async fn apply_status_updates(status_updates: &[L1StatusUpdate], status_tx: Arc<StatusTx>) {
-    let mut l1_status = status_tx.l1.borrow().clone();
-    for event in status_updates {
+pub async fn apply_status_updates(st_updates: &[L1StatusUpdate], st_chan: &StatusChannel) {
+    let mut l1_status = st_chan.l1_status();
+    for event in st_updates {
         match event {
             L1StatusUpdate::CurHeight(height) => l1_status.cur_height = *height,
             L1StatusUpdate::LastUpdate(epoch_time) => l1_status.last_update = *epoch_time,
@@ -31,7 +28,5 @@ pub async fn apply_status_updates(status_updates: &[L1StatusUpdate], status_tx: 
         }
     }
 
-    if let Err(err) = status_tx.l1.send(l1_status.clone()) {
-        error!(%err, "error updating l1status");
-    }
+    st_chan.update_l1_status(l1_status);
 }
--- a/crates/consensus-logic/src/csm/worker.rs
+++ b/crates/consensus-logic/src/csm/worker.rs
@@ -11,7 +11,7 @@ use strata_db::{
 use strata_eectl::engine::ExecEngineCtl;
 use strata_primitives::prelude::*;
 use strata_state::{client_state::ClientState, csm_status::CsmStatus, operation::SyncAction};
-use strata_status::StatusTx;
+use strata_status::StatusChannel;
 use strata_storage::{managers::checkpoint::CheckpointDbManager, L2BlockManager};
 use strata_tasks::ShutdownGuard;
 use tokio::{
@@ -117,10 +117,16 @@ pub fn client_worker_task<D: Database, E: ExecEngineCtl>(
     mut state: WorkerState<D>,
     engine: Arc<E>,
     mut msg_rx: mpsc::Receiver<CsmMessage>,
-    status_tx: Arc<StatusTx>,
+    status_channel: StatusChannel,
 ) -> Result<(), Error> {
     while let Some(msg) = msg_rx.blocking_recv() {
-        if let Err(e) = process_msg(&mut state, engine.as_ref(), &msg, &status_tx, &shutdown) {
+        if let Err(e) = process_msg(
+            &mut state,
+            engine.as_ref(),
+            &msg,
+            &status_channel,
+            &shutdown,
+        ) {
             error!(err = %e, ?msg, "failed to process sync message, aborting!");
             break;
         }
@@ -140,7 +146,7 @@ fn process_msg<D: Database>(
     state: &mut WorkerState<D>,
     engine: &impl ExecEngineCtl,
     msg: &CsmMessage,
-    status_rx: &Arc<StatusTx>,
+    status_channel: &StatusChannel,
     shutdown: &ShutdownGuard,
 ) -> anyhow::Result<()> {
     match msg {
@@ -153,7 +159,7 @@ fn process_msg<D: Database>(
                 if ev_idx < *idx {
                     warn!(%ev_idx, "Applying missed sync event.");
                 }
-                handle_sync_event_with_retry(state, engine, ev_idx, status_rx, shutdown)?;
+                handle_sync_event_with_retry(state, engine, ev_idx, status_channel, shutdown)?;
             }
 
             Ok(())
@@ -167,7 +173,7 @@ fn handle_sync_event_with_retry<D: Database>(
     state: &mut WorkerState<D>,
     engine: &impl ExecEngineCtl,
     ev_idx: u64,
-    status_tx: &Arc<StatusTx>,
+    status_channel: &StatusChannel,
     shutdown: &ShutdownGuard,
 ) -> anyhow::Result<()> {
     // Fetch the sync event so that we can debug print it.
@@ -190,7 +196,7 @@ fn handle_sync_event_with_retry<D: Database>(
         // TODO demote to trace after we figure out the current issues
         debug!("trying sync event");
 
-        let Err(e) = handle_sync_event(state, engine, ev_idx, status_tx) else {
+        let Err(e) = handle_sync_event(state, engine, ev_idx, status_channel) else {
             // Happy case, we want this to happen.
             trace!("completed sync event");
             break;
@@ -220,7 +226,7 @@ fn handle_sync_event<D: Database>(
     state: &mut WorkerState<D>,
     engine: &impl ExecEngineCtl,
     ev_idx: u64,
-    status_tx: &Arc<StatusTx>,
+    status_channel: &StatusChannel,
 ) -> anyhow::Result<()> {
     // Perform the main step of deciding what the output we're operating on.
     let (outp, new_state) = state.state_tracker.advance_consensus_state(ev_idx)?;
@@ -228,7 +234,7 @@ fn handle_sync_event<D: Database>(
 
     // Apply the actions produced from the state transition.
     for action in outp.actions() {
-        apply_action(action.clone(), state, engine)?;
+        apply_action(action.clone(), state, engine, status_channel)?;
     }
 
     // Make sure that the new state index is set as expected.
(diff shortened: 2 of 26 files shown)


crates/consensus-logic/src/csm/worker.rs
```diff
@@ -153,7 +153,7 @@ fn process_msg<D: Database>(
                 if ev_idx < *idx {
                     warn!(%ev_idx, "Applying missed sync event.");
                 }
-                handle_sync_event_with_retry(state, engine, ev_idx, status_rx, shutdown)?;
+                handle_sync_event_with_retry(state, engine, ev_idx, status_tx, shutdown)?;
```
storopoli, line 156: This diff seems wrong no?
I expect that you would never replace a `foo_rx` for a `foo_tx` no?

crates/primitives/src/l1.rs
```diff
@@ -227,10 +227,13 @@ pub struct L1Status {
     pub cur_tip_blkid: String,
 
     /// Last published txid where L2 blob was present
-    pub last_published_txid: Option<String>,
+    pub last_published_txid: Option<Txid>,
```
storopoli, line 230: Yes! Every time that we use `String` or `Vec<u8>` instead of proper types a puppy dies.

crates/status/src/status_manager.rs
```diff
@@ -16,93 +16,121 @@ pub enum StatusError {
...
+}
+
+impl StatusReceiver {
+    /// Borrow [`ClientState`] from channel. Be careful while using this to avoid deadlock which can
+    /// occur if the borrow lives too long during which sender tries to send through this.
+    pub fn borrow_client_state(&self) -> Ref<ClientState> {
+        self.cl.borrow()
+    }
+
+    /// Gets the lastest [`ClientState`].
```
storopoli, line 73:
```suggestion
    /// Gets the latest [`ClientState`].
```

crates/status/src/status_manager.rs
```diff
@@ -16,93 +16,121 @@ pub enum StatusError {
...
+    pub fn client_state(&self) -> ClientState {
+        self.cl.borrow().clone()
+    }
+
+    /// Borrow [`Chainstate`] from channel. Be careful while using this to avoid deadlock which can
```
storopoli, line 78:
```suggestion
    /// Borrows [`Chainstate`] from channel. Be careful while using this to avoid deadlock which can
```

You are using "Gets" above, so be consistent.

crates/status/src/status_manager.rs
```diff
@@ -16,93 +16,121 @@ pub enum StatusError {
...
+    /// occur if the borrow lives too long during which sender tries to send through this.
+    pub fn borrow_chainstate(&self) -> Ref<Option<Chainstate>> {
+        self.chs.borrow()
+    }
+
+    /// Gets the lastest [`ChainState`].
```
storopoli, line 84:
```suggestion
    /// Gets the latest [`ChainState`].
```

crates/status/src/status_manager.rs
```diff
@@ -16,93 +16,121 @@ pub enum StatusError {
...
+    pub fn chainstate(&self) -> Option<Chainstate> {
+        self.chs.borrow().clone()
+    }
+
+    /// Borrow [`L1Status`] from channel. Be careful while using this to avoid deadlock which can
```
storopoli, line 89:
```suggestion
    /// Borrows [`L1Status`] from channel. Be careful while using this to avoid deadlock which can
```

crates/status/src/status_manager.rs
```diff
@@ -16,93 +16,121 @@ pub enum StatusError {
...
+    /// occur if the borrow lives too long during which sender tries to send through this.
+    pub fn borrow_l1_status(&self) -> Ref<L1Status> {
+        self.l1.borrow()
     }
+
+    /// Gets the lastest [`L1Status`].
```
storopoli, line 95:
```suggestion
    /// Gets the latest [`L1Status`].
```

crates/status/src/status_manager.rs
```diff
@@ -16,93 +16,121 @@ pub enum StatusError {
...
-    pub cl: watch::Sender<ClientState>,
-    pub l1: watch::Sender<L1Status>,
-    pub chs: watch::Sender<Option<Chainstate>>,
+pub struct StatusSender {
+    cl: watch::Sender<ClientState>,
+    l1: watch::Sender<L1Status>,
+    chs: watch::Sender<Option<Chainstate>>,
 }
 
 /// initializes the StatusRx and StatusTx watch channel wrapper
```
storopoli, line 115: I think you need to update this docstring.

storopoli (changes requested):
Minor nits and maybe a logic error.

delbonis (changes requested):
The description of this ticket said to improve the interface and hide the implementation details by providing context-aware accessors for the common generally-useful fields (see ticket) that make sense for how we want to wait on them instead of just returning the full states and still exposing implementation details (`Ref`).

Also combining the send side and the receive side feels sloppy as it could obscure the real dataflows between different systems.  What was the rationale for consolidating it?

crates/consensus-logic/src/csm/worker.rs
bewakes: That's a valid concern. Actually, the function argument was wrongly named, this change uses the correct name. :D

bewakes:
> Also combining the send side and the receive side feels sloppy as it could obscure the real dataflows between different systems. What was the rationale for consolidating it?

The idea was to expose interfaces via a "Hub" that could be passed around. But then, later it felt like providing sending capabilities to components that only require receiving and vice versa is not ideal, and hence exposure to `receiver()` and `sender()`. But I guess we are better off with how it was earlier.

This wasn't actually meant to be marked as a ready PR though.

bewakes:
Fixed the docstrings and did some more cleanups and interface update.

delbonis:
I'm not against consolidating it, I just want to make sure it's well-reasoned.  A consolidated approach should hide the internals in any case, removing those two fns and directly exposing the actual operation fns on the "hub" type.

crates/status/src/status_manager.rs
```diff
@@ -16,93 +20,131 @@ pub enum StatusError {
     Other(String),
 }
 
-/// Bundle wrapper for Status receiver
-pub struct StatusRx {
-    csm: watch::Receiver<CsmStatus>,
+pub type StatusChannel = (Arc<StatusSender>, Arc<StatusReceiver>);
```
delbonis, line 23: This is awkward because it makes you do the `.0`/`.1` syntax and doesn't really abstract over anything.  It's structurally equivalent to the hub concept but less ergonomic and with less ability to rework the internals.

crates/status/src/status_manager.rs
```diff
@@ -16,93 +20,131 @@ pub enum StatusError {
...
+            return Ok(self.cl.borrow().clone());
+        }
+
+        // If not wait for it
+        loop {
+            rx.changed().await?;
+            let state = rx.borrow();
+            if state.has_genesis_occurred() {
+                return Ok(state.clone());
+            }
```
delbonis, line 124: See if `.wait_for` does what we need here.

delbonis (changes requested):

crates/status/src/status_manager.rs
bewakes: Actually it doesn't because there's one more check besides changed() we need to do which is check if genesis happened. But this method can be simplified like this:
```rust
pub async fn wait_until_genesis(&self) -> Result<ClientState, RecvError> {
        let mut rx = self.cl.clone();
        loop {
            if rx.borrow().has_genesis_occurred() {
                return Ok(rx.borrow().clone());
            }
            rx.changed().await?;
        }
    }
```

crates/status/src/status_manager.rs
bewakes: Actually I don't want to pass whole of the channel/hub thing everywhere, especially to components which needs only tx or rx parts. And yes, `.0/.1` is not good. I should convert to `.receiver()` and `sender()`. I don't quite feel like the channel should hide both receiver and sender and expose what those pieces do. This is very similar to our database traits that had provider and store. It makes sense to merge them for db, but in my opinion, that should not be done here. Because, we will always have receiving and sending ends in a channel and all the abstractions can happen within those pieces.

bewakes: Well, I kept receiver() and sender(). But didn't look good. So, removed them and directly exposed methods via `StatusChannel`. Thanks for the suggestion.

crates/status/src/status_manager.rs
```diff
@@ -16,93 +20,147 @@ pub enum StatusError {
...
+            rx.changed().await?;
+        }
+    }
+
+    // Sender methods
+
+    /// Sends the updated `Chainstate` to the chain state receiver. Logs a warning if the receiver
+    /// is dropped.
+    pub fn update_chainstate(&self, post_state: &Chainstate) {
```
delbonis, line 124: This take ownership of the chainstate so it should take it by value instead of by ref.

crates/status/src/status_manager.rs
```diff
@@ -16,93 +20,147 @@ pub enum StatusError {
...
     }
 
+    /// Sends the updated `L1Status` to the L1 status receiver. Logs a warning if the receiver is
+    /// dropped.
     pub fn update_l1_status(&self, post_state: &L1Status) {
-        if self.l1.send(post_state.clone()).is_err() {
+        if self.sender.l1.send(post_state.clone()).is_err() {
             warn!("l1 status receiver dropped");
         }
     }
```
delbonis, line 144: Same for these as well.  If you always unconditionally clone it then there's no point in taking it by ref.

crates/status/src/status_manager.rs
```diff
@@ -16,93 +20,147 @@ pub enum StatusError {
...
 
-    pub fn update_csm_status(&self, post_state: &CsmStatus) {
-        if self.csm.send(post_state.clone()).is_err() {
-            warn!("csm status receiver dropped");
-        }
+    /// Subscribes to updates for `ClientState` and returns a receiver.
+    pub fn subscribe_cl(&self) -> watch::Receiver<ClientState> {
+        self.sender.cl.subscribe()
```
delbonis, line 148: This doesn't seem to be used anywhere, ideally we shouldn't leak it.

delbonis (changes requested):
Generally good, some more minor cleanliness comments.

crates/status/src/status_manager.rs
bewakes: Oh yes, I forgot it. Removed.

crates/status/src/status_manager.rs
bewakes: True. Changed.

crates/status/src/status_manager.rs
bewakes: Done.

delbonis (approved):

bewakes:
@alpenlabs/crypto @delbonis @storopoli requesting for re-review and approval again.

storopoli (approved):
ACK 3c18b9557a282312438f0a7519c62b0f870cf269

crates/btcio/src/status.rs
```diff
@@ -15,9 +12,9 @@ pub enum L1StatusUpdate {
     IncrementInscriptionCount,
 }
 
-pub async fn apply_status_updates(status_updates: &[L1StatusUpdate], status_tx: Arc<StatusTx>) {
-    let mut l1_status = status_tx.l1.borrow().clone();
-    for event in status_updates {
+pub async fn apply_status_updates(st_updates: &[L1StatusUpdate], st_chan: &StatusChannel) {
```
delbonis, line 15: We can simplify this more.  The way this `L1StatusUpdate` concept works is just adding more unnecessary steps.

delbonis (approved):
Change isn't really required, but would be nice.  Approving anyways.

PR #486, opened 2024-11-26
Factor out common Buf* implementations into the macro rule.

evgenyzdanovich:
## Description

Factor out common Buf* implementations into the macro rule.

### Type of Change

-   [ ] Documentation update
-   [x] Refactor
-   [ ] New or updated tests

## Checklist

### DIFF
--- a/crates/evmexec/src/engine.rs
+++ b/crates/evmexec/src/engine.rs
@@ -398,7 +398,7 @@ fn to_bridge_withdrawal_intent(
     rpc_withdrawal_intent: strata_reth_node::WithdrawalIntent,
 ) -> bridge_ops::WithdrawalIntent {
     let strata_reth_node::WithdrawalIntent { amt, dest_pk } = rpc_withdrawal_intent;
-    bridge_ops::WithdrawalIntent::new(BitcoinAmount::from_sat(amt), XOnlyPk::new(Buf32(dest_pk)))
+    bridge_ops::WithdrawalIntent::new(BitcoinAmount::from_sat(amt), XOnlyPk::new(Buf32(*dest_pk)))
 }
 
 #[cfg(test)]
@@ -529,8 +529,8 @@ mod tests {
 
         let timestamp = 0;
         let el_ops = vec![];
-        let safe_l1_block = Buf32(FixedBytes::<32>::random());
-        let prev_l2_block = Buf32(FixedBytes::<32>::random()).into();
+        let safe_l1_block = FixedBytes::<32>::random().into();
+        let prev_l2_block = Buf32(FixedBytes::<32>::random().into()).into();
 
         let payload_env = PayloadEnv::new(timestamp, prev_l2_block, safe_l1_block, el_ops);
 
@@ -573,7 +573,7 @@ mod tests {
         let fcs = ForkchoiceState::default();
 
         let el_payload = ElPayload {
-            base_fee_per_gas: Buf32(U256::from(10).into()),
+            base_fee_per_gas: FixedBytes::<32>::from(U256::from(10)).into(),
             parent_hash: Default::default(),
             fee_recipient: Default::default(),
             state_root: Default::default(),
--- a/crates/primitives/src/buf.rs
+++ b/crates/primitives/src/buf.rs
@@ -1,84 +1,32 @@
-use std::{
-    fmt,
-    io::{self, Read, Write},
-    str,
-};
+use std::str::FromStr;
 
-use arbitrary::Arbitrary;
 use bitcoin::{
     hashes::Hash,
     secp256k1::{schnorr, SecretKey, XOnlyPublicKey},
     BlockHash, Txid,
 };
-use borsh::{BorshDeserialize, BorshSerialize};
-use reth_primitives::revm_primitives::alloy_primitives::FixedBytes;
-use serde::{Deserialize, Serialize};
+use reth_primitives::revm_primitives::alloy_primitives::hex;
 
-use crate::errors::ParseError;
+use crate::{errors::ParseError, macros::internal};
 
 // 20-byte buf
-#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Default)]
-pub struct Buf20(pub FixedBytes<20>);
-
-impl Buf20 {
-    pub fn zero() -> Self {
-        Self([0; 20].into())
-    }
-}
-
-impl From<[u8; 20]> for Buf20 {
-    fn from(value: [u8; 20]) -> Self {
-        Self(FixedBytes::from(value))
-    }
-}
-
-impl fmt::Debug for Buf20 {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let mut buf = [0; 40];
-        hex::encode_to_slice(self.0, &mut buf).expect("buf: enc hex");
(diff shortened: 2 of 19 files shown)


bewakes:
Just a thought, could using const generics be possible here?

evgenyzdanovich:
> Just a thought, could using const generics be possible here?

@bewakes 

I guess yes, but I can't think of any significant advantages here. Do you have something in mind?

bewakes:
Maybe as an alternative to the macro usage.

evgenyzdanovich:
> Maybe as an alternative to the macro usage.

Ah, I see. But I don't think we'll get rid of macro completely, right? 

Anyways, no strong opinion on this point. @delbonis @storopoli any thoughts?

crates/evmexec/src/engine.rs
```diff
@@ -529,8 +529,8 @@ mod tests {
 
         let timestamp = 0;
         let el_ops = vec![];
-        let safe_l1_block = Buf32(FixedBytes::<32>::random());
-        let prev_l2_block = Buf32(FixedBytes::<32>::random()).into();
+        let safe_l1_block = FixedBytes::<32>::random().into();
+        let prev_l2_block = Buf32(FixedBytes::<32>::random().into()).into();
```
delbonis, line 533: Could we get rid of this extra `Buf32` here too?

crates/primitives/src/buf.rs
```diff
@@ -1,84 +1,32 @@
...
-    fn from(value: [u8; 32]) -> Self {
-        Self(FixedBytes::from(value))
-    }
-}
-
-impl From<Buf32> for FixedBytes<32> {
-    fn from(value: Buf32) -> Self {
-        value.0
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        hex::decode_to_array(s).map(Self::new)
```
delbonis, line 29: This is great :+1:.

crates/primitives/src/macros.rs
```diff
@@ -0,0 +1,203 @@
...
+                pub fn as_bytes(&self) -> &[u8] {
+                    self.0.as_slice()
+                }
+
+                pub fn zero() -> Self {
+                    Self([0; $len].into())
+                }
+
+                pub fn is_zero(&self) -> bool {
+                    *self.as_ref() == [0; $len]
```
delbonis, line 28: Is this better than `self.iter().all(|v| == 0)`?

delbonis (approved):
This is great, appreciate the cleanups downstream.  Just a couple questions though.

delbonis:
@bewakes @evgenyzdanovich It doesn't make a huge improvement and the generics can confuse rust-analyzer.  This is a clean 1:1 generalization, I can't see anywhere we could take advantage of it being generic at currently.

storopoli:
> > Maybe as an alternative to the macro usage.
> 
> Ah, I see. But I don't think we'll get rid of macro completely, right?
> 
> Anyways, no strong opinion on this point. @delbonis @storopoli any thoughts?

I like the idea. Thanks!
Agree with all @delbonis's comments.

crates/evmexec/src/engine.rs
evgenyzdanovich: we can, but then `L2BlockId` (in the crate `state`) would need an additional `From` impl, I guess.

Given it's in the tests only (I didn't see this pattern in the code, but maybe I overlooked somewhere), I'd leave it as is, if you don't mind.

crates/primitives/src/buf.rs
evgenyzdanovich: thanks :)

Rajil1213 (approved):
Looks good. No issues with the parts of this PR that touch the bridge crates.

storopoli (dismissed):
ACK 1f0f0280a19f826bb0c2d0cbcd6181318e3a7c5c

evgenyzdanovich:
> Looks good but didn't address all imports. I see `std`, `serde`, `arbitrary` etc not being imported via absolute crate path. `Default` too

Right, done (hopefully).

crates/evmexec/src/engine.rs
delbonis: I think adding a `From<Buf32>` makes sense there.

Would it make sense add another macro that can instantiate the impls for types that are just wrappers around hashes like `L2BlockId` and `L1BlockId`?

evgenyzdanovich: Good idea one another macro, done.

> I think adding a `From<Buf32>` makes sense there.

@delbonis Sorry, did you mean `From<FixedBytes>`? `From<Buf32>` is already there. 
I now think adding `From<FixedBytes>` actually makes little sense. One obvious drawback - it adds a dependency on `reth_pritimitives` for the `strata-state` crate. Right now, `strata-state` depends only on `strata-primitives` which I find ergonomic.

Nonetheless, not a strong opinion, let me know what do you think, so I can wrap this piece of work soon :)

crates/primitives/src/macros.rs
```diff
@@ -0,0 +1,244 @@
...
+                fn default() -> Self {
+                    Self([0; $len])
+                }
+            }
+
+            impl ::std::fmt::Debug for $name {
+                fn fmt(&self, f: &mut ::std::fmt::Formatter<'_>) -> ::std::fmt::Result {
+                    // twice as large, required by the hex::encode_to_slice.
+                    let mut buf = [0; $len * 2];
+                    hex::encode_to_slice(self.0, &mut buf).expect("buf: enc hex");
```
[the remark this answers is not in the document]
evgenyzdanovich: thanks, done

crates/primitives/src/macros.rs
```diff
@@ -0,0 +1,244 @@
...
+                    f.write_str(unsafe { ::core::str::from_utf8_unchecked(&buf) })
+                }
+            }
+
+            impl ::std::fmt::Display for $name {
+                fn fmt(&self, f: &mut ::std::fmt::Formatter<'_>) -> ::std::fmt::Result {
+                    // fmt only first and last bits of data.
+                    let mut buf = [0; 6];
+                    hex::encode_to_slice(&self.0[..3], &mut buf).expect("buf: enc hex");
```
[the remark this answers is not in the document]
evgenyzdanovich: thanks, done

crates/primitives/src/macros.rs
```diff
@@ -0,0 +1,244 @@
...
+                    f.write_str(unsafe { ::core::str::from_utf8_unchecked(&buf) })?;
+                    f.write_str("..")?;
+                    hex::encode_to_slice(&self.0[$len - 3..], &mut buf).expect("buf: enc hex");
```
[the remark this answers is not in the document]
evgenyzdanovich: thanks, done

evgenyzdanovich (commented):
Thanks for your thorough review, @Zk2u ! PTAL (hopefully final one:)

prajwolrg (approved):
LGTM!

crates/evmexec/src/engine.rs
delbonis: Oh then I guess we're all set.  Yeah we want to avoid extra deps on `reth_primitives`.

delbonis (approved):

storopoli (approved):
ACK df5234738a2cbbf586fc823925b8bcd3ec51749d

PR #497, opened 2024-11-29
feat(storage) `CacheTable` rewrite

Zk2u:
## Description

This PR rewrites the `CacheTable` used for the L2 block manager (and soon the L1BM) to increase performance.

It adds the following dependencies:

| Dependency | About                       | Reason for addition |
|--------------|--------------------|--------------------|
| concurrent-map | A lock-free B+ tree based on [sled](https://github.com/spacejam/sled)’s internal index structure, but supporting richer Rust types as keys and values than raw bytes. This structure supports atomic compare and swap operations with the [ConcurrentMap::cas](https://docs.rs/concurrent-map/latest/concurrent_map/struct.ConcurrentMap.html#method.cas) method. | Powers the new in-memory cache, replaces the effective Mutex around a hashmap + arc<rwlock> rubbish we had going on before. |
| cache-advisor | A simple, high performance scan-resistant eviction manager. Tells you when to evict items from a cache. Features: two-segment LRU, protects against cache pollution from single-hit items, 256 shards accessed via non-blocking flatcombining, local access buffer that must fill up before accessing shared state, compresses the costs associated with each item to a u8 using a compression technique that will converge to the overall true sum of costs over time, but allows for much less memory to be used for accounting. | Concurrent map isn't an LRU by itself, so this gives us LRU functionality |
| kanal| Provides multi-producer and multi-consumer channels with advanced features and lock-free one-shot channels that only allocate the size of a pointer in the heap, allowing for fast communication. The library has a focus on unifying message passing between synchronous and asynchronous portions of Rust code through a combination of synchronous and asynchronous APIs while maintaining high performance. | Fastest channel design in Rust, so I want to move our internal channels to use this eventually. However, it was needed here for another reason. Tokio's `broadcast::Receiver`cannot be cloned - you need the `Sender`to create new `Receiver`s. Kanal's channels are cooler and let you do this, which gets rid of the Arcing we were previously doing, allowing concurrent map to clone `Slot`s when necessary |
| ahash | Rust's fastest `Hasher`implementation. Uses hardware AES when available. Non-cryptographic. | `CacheAdvisor`identifies objects via `u64` keys. To keep the cache generic, we require `V: Hash`and then internally use `u64`s to identify things in the map. So a bit of a hybrid between a B+ tree and a hash map. |

Note that the new `CacheTable`is `Send`but `!Sync`. You cannot share a single instantiation between threads. This is the biggest API-level change. The way to share it across threads is by cloning, and giving a new clone of the `CacheTable` whenever it might switch threads, like for a tokio task.

I haven't yet tested the performance of this implementation, but I've built something using the same tooling and similar algorithms before. These things are usually blisteringly quick.

### Type of Change

-   [ ] Documentation update
-   [x] Refactor
-   [ ] New or updated tests

## Checklist

## Related Issues

STR-627

### DIFF
--- a/crates/storage/src/cache.rs
+++ b/crates/storage/src/cache.rs
@@ -1,32 +1,50 @@
 //! Generic cache utility for what we're inserting into the database.
 
-use std::{hash::Hash, num::NonZeroUsize, sync::Arc};
+use std::{cell::RefCell, cmp::PartialEq, hash::Hash, marker::PhantomData, num::NonZeroUsize};
 
+use ahash::RandomState;
+use cache_advisor::CacheAdvisor;
+use concurrent_map::{CasFailure, ConcurrentMap};
+use kanal::{bounded_async, AsyncReceiver};
 use strata_db::{DbError, DbResult};
-use tokio::sync::{broadcast, Mutex, RwLock};
 use tracing::*;
 
 use crate::exec::DbRecv;
 
-/// Entry for something we can put into the cache without actually knowing what it is, and so we can
-/// keep the reservation to it.
-type CacheSlot<T> = Arc<RwLock<SlotState<T>>>;
-
 /// Describes a cache entry that may be occupied, reserved for pending database read, or returned an
 /// error from a database read.
-#[derive(Debug)]
-pub enum SlotState<T> {
+#[derive(Debug, Clone, PartialEq)]
+pub enum Entry<T: Clone> {
     /// Authentic database entry.
     Ready(T),
 
     /// A database fetch is happening in the background and it will be updated.
-    Pending(broadcast::Receiver<T>),
-
-    /// An unspecified error happened fetching from the database.
-    Error,
+    Pending(Receiver<DbResult<T>>),
 }
 
-impl<T: Clone> SlotState<T> {
+// Do not use outside this module.
+//
+// A wrapper around a async kanal receiver that implements PartialEq so we can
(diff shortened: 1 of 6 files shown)


delbonis:
> Note that the new CacheTableis Sendbut !Sync. You cannot share a single instantiation between threads. This is the biggest API-level change. The way to share it across threads is by cloning, and giving a new clone of the CacheTable whenever it might switch threads, like for a tokio task.

Yeah this is a big problem.  We're likely to be fetching stuff from DB on different tasks/threads, so this probably weakens our use of caching here.

crates/storage/src/cache.rs
```diff
@@ -1,32 +1,50 @@
...
+impl<T> AsRef<AsyncReceiver<T>> for Receiver<T> {
+    fn as_ref(&self) -> &AsyncReceiver<T> {
+        &self.0
+    }
 }
 
-impl<T: Clone> SlotState<T> {
+impl<T> PartialEq for Receiver<T> {
+    fn eq(&self, _other: &Self) -> bool {
+        // this is only so we can
```
bewakes, line 41: Is this comment incomplete?

storopoli:
> Anyway, not gonna do this anymore. I spent way too much time working on something like this last week and it was basically just premature optimisations and the replacement data structure that was built is very buggy atm.

But I liked the idea of `parking_lot` + `hashbrown`. That was an easy refactor and it was faster than what we have now no?

PR #502, opened 2024-12-01
Changes in ArbitraryGenerator

purusang:
## Description
- Used 128 bytes of default buffer size vs 2^24 bytes buffer size earlier.
- Used thread_rng to generate random number in each call as compared to using slices from pre-initialized huge buffer
- Updated unit tests accordingly to work with new changes.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues
STR-551

### DIFF
--- a/crates/test-utils/src/lib.rs
+++ b/crates/test-utils/src/lib.rs
@@ -18,16 +18,39 @@ impl Default for ArbitraryGenerator {
     }
 }
 impl ArbitraryGenerator {
+    /// Creates a new `ArbitraryGenerator` with a default buffer size.
+    ///
+    /// # Returns
+    ///
+    /// A new instance of `ArbitraryGenerator`.
     pub fn new() -> Self {
         ArbitraryGenerator {
             buf: vec![0u8; ARB_GEN_LEN],
         }
     }
+
+    /// Creates a new `ArbitraryGenerator` with a specified buffer size.
+    ///
+    /// # Arguments
+    ///
+    /// * `s` - The size of the buffer to be used.
+    ///
+    /// # Returns
+    ///
+    /// A new instance of `ArbitraryGenerator` with the specified buffer size.
     pub fn new_with_size(s: usize) -> Self {
         ArbitraryGenerator { buf: vec![0u8; s] }
     }
 
-    /// Legacy interface: no arguments
+    /// Generates an arbitrary instance of type `T` using the default RNG, [`OsRng`].
+    ///
+    /// # Returns
+    ///
+    /// An arbitrary instance of type `T`.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if it fails to generate an arbitrary instance.
     pub fn generate<'a, T>(&'a mut self) -> T
     where
         T: Arbitrary<'a> + Clone,
@@ -35,7 +58,21 @@ impl ArbitraryGenerator {
         self.generate_with_rng::<T, OsRng>(None)
     }
 
-    /// Core function: accepts an optional RNG
+    /// Generates an arbitrary instance of type `T` using an optional RNG.
+    ///
+    /// # Arguments
+    ///
+    /// * `rng` - An optional RNG to be used for generating the arbitrary instance. If `None`, the
+    ///   default RNG, [`OsRng`], will be used. The provided RNG must implement the [`RngCore`] and
+    ///   [`CryptoRng`] traits.
+    ///
+    /// # Returns
+    ///
+    /// An arbitrary instance of type `T`.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if it fails to generate an arbitrary instance.
     pub fn generate_with_rng<'a, T, R>(&'a mut self, rng: Option<&mut R>) -> T
     where
         T: Arbitrary<'a> + Clone,
--- a/crates/test-utils/src/lib.rs
+++ b/crates/test-utils/src/lib.rs
@@ -1,5 +1,5 @@
 use arbitrary::{Arbitrary, Unstructured};
-use rand::{thread_rng, RngCore};
+use rand::{rngs::OsRng, CryptoRng, RngCore};
 
 pub mod bitcoin;
 pub mod bridge;
@@ -10,8 +10,7 @@ pub mod l2;
 const ARB_GEN_LEN: usize = 128;
 
 pub struct ArbitraryGenerator {
-    rng: rand::rngs::ThreadRng, // Thread-local RNG
-    buf: Vec<u8>,               // Persistent buffer
+    buf: Vec<u8>, // Persistent buffer
 }
 impl Default for ArbitraryGenerator {
     fn default() -> Self {
@@ -21,19 +20,34 @@ impl Default for ArbitraryGenerator {
 impl ArbitraryGenerator {
     pub fn new() -> Self {
         ArbitraryGenerator {
-            rng: thread_rng(),
             buf: vec![0u8; ARB_GEN_LEN],
         }
     }
     pub fn new_with_size(s: usize) -> Self {
-        ArbitraryGenerator {
-            rng: thread_rng(),
-            buf: vec![0u8; s],
-        }
+        ArbitraryGenerator { buf: vec![0u8; s] }
     }
 
-    pub fn generate<'a, T: Arbitrary<'a> + Clone>(&'a mut self) -> T {
-        self.rng.fill_bytes(&mut self.buf);
+    /// Legacy interface: no arguments
+    pub fn generate<'a, T>(&'a mut self) -> T
+    where
+        T: Arbitrary<'a> + Clone,
+    {
+        self.generate_with_rng::<T, OsRng>(None)
+    }
+
+    /// Core function: accepts an optional RNG
+    pub fn generate_with_rng<'a, T, R>(&'a mut self, rng: Option<&mut R>) -> T
+    where
+        T: Arbitrary<'a> + Clone,
+        R: RngCore + CryptoRng,
+    {
+        let mut thread_rng = OsRng; // Default secure RNG
+        let rng: &mut dyn RngCore = match rng {
+            Some(r) => r,
+            None => &mut thread_rng as &mut dyn RngCore,
+        };
+
+        rng.fill_bytes(&mut self.buf);
         let mut u = Unstructured::new(&self.buf);
         T::arbitrary(&mut u).expect("Failed to generate arbitrary instance")
     }
--- a/crates/test-utils/src/lib.rs
+++ b/crates/test-utils/src/lib.rs
@@ -6,8 +6,8 @@ pub mod bridge;
 pub mod evm_ee;
 pub mod l2;
 
-// Smaller buffer size as compared to 2^24
-const ARB_GEN_LEN: usize = 128;
+/// The default buffer size for the `ArbitraryGenerator`.
+const ARB_GEN_LEN: usize = 1024;
 
 pub struct ArbitraryGenerator {
     buf: Vec<u8>, // Persistent buffer
@@ -24,9 +24,7 @@ impl ArbitraryGenerator {
     ///
     /// A new instance of `ArbitraryGenerator`.
     pub fn new() -> Self {
-        ArbitraryGenerator {
-            buf: vec![0u8; ARB_GEN_LEN],
-        }
+        Self::new_with_size(ARB_GEN_LEN)
     }
 
     /// Creates a new `ArbitraryGenerator` with a specified buffer size.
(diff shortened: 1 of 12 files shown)


crates/test-utils/src/lib.rs
```diff
@@ -1,46 +1,40 @@
...
-        let mut u = Unstructured::new(&self.buf[off..]);
-        let prev_off = u.len();
-        let inst = T::arbitrary(&mut u).expect("failed to generate arbitrary instance");
-        let additional_off = prev_off - u.len();
-        self.off.store(off + additional_off, Ordering::Relaxed);
-        inst
+    pub fn generate<'a, T: Arbitrary<'a> + Clone>(&'a mut self) -> T {
+        self.rng.fill_bytes(&mut self.buf);
+        let mut u = Unstructured::new(&self.buf);
+        T::arbitrary(&mut u).expect("Failed to generate arbitrary instance")
```
delbonis, line 38: The way this is written kinda makes having the random buf be a field pointless.  Since we sample new randomness on each call it might as well just be a local and then we can just have this take `&self`.  Also what happens if we consume more than the length of randomness in a call to `generate`?  I'm not sure the best design approach here since it'd be nice to be able to make it deterministic if desired.

Also should verify if it makes sense to include the rng as a field as opposed to just calling `thread_rng()` on every invocation.  I'm not sure if there's a practical difference in terms of speed/determinism there.

delbonis (changes requested):

crates/test-utils/src/lib.rs
bewakes: I think the buf field is required because of lifetime issues. The random buf is generated inside the method which can't leave the method scope.

I don't really like having `mut` everywhere though. Maybe a `RefCell` can be used for buffer, which should be fine since we probably won't be using a generator instance across multiple threads. What are your thoughts on this?

purusang: > The way this is written kinda makes having the random buf be a field pointless. 
The reason for having buf field is due to lifetime requirement, like @bewakes said.
> Since we sample new randomness on each call it might as well just be a local and then we can just have this take `&self`. 
We needed `&mut` due to having to update randomness on every call. 
> Also what happens if we consume more than the length of randomness in a call to `generate`? I'm not sure the best design approach here since it'd be nice to be able to make it deterministic if desired.
It's a good concern, but I think having new_with_size() serves that purpose.
> Also should verify if it makes sense to include the rng as a field as opposed to just calling `thread_rng()` on every invocation. I'm not sure if there's a practical difference in terms of speed/determinism there.
Both approaches definitely work, and don't see either way to be an issue.

purusang: > I think the buf field is required because of lifetime issues. The random buf is generated inside the method which can't leave the method scope.

That's what I observed too.
> I don't really like having `mut` everywhere though. Maybe a `RefCell` can be used for buffer(and for rng?), which should be fine since we probably won't be using a generator instance across multiple threads. What are your thoughts on this?

Yeah, initially I did not like the idea of having to change multiple files to pass mutable arbitrary generator but due to lifetime requirement using RefCell was not helpful there.

delbonis: Oh christ I'm realizing how this api kinda sucks.  I didn't see that `Arbitrary` has a lifetime param.

Taking `&mut self` means I think you can have multiple invocations now can you?  Since you'd overwrite the buffer that we have a hypothetical ref to in the value returned in the previous invocation.

Is this an issue that we can deal with by generating the randomness upfront?

bewakes: > I think you can have multiple invocations now can you?

Multiple `arb.generate()`? Yes.

delbonis: Sorry, I meant *can't*.  Since you'd need to take an exclusive borrow while there was a shared borrow active.

crates/test-utils/src/lib.rs
```diff
@@ -1,46 +1,40 @@
...
 pub mod l2;
 
-const ARB_GEN_LEN: usize = 1 << 24; // 16 MiB
+// Smaller buffer size as compared to 2^24
+const ARB_GEN_LEN: usize = 128;
 
 pub struct ArbitraryGenerator {
-    buf: Vec<u8>,
-    off: AtomicUsize,
+    rng: rand::rngs::ThreadRng, // Thread-local RNG
```
AaronFeickert, line 13: What's the rationale for the use of `ThreadRng` versus `OsRng`? We've been standardizing on the use of `OsRng` elsewhere.

delbonis (approved):

crates/test-utils/src/lib.rs
AaronFeickert: ^ pinging @delbonis as well, given recent approval

delbonis: It's faster, right?  I don't think it really matters since it's only being used in tests.

AaronFeickert: In theory it's faster, but I think it's worth being consistent to avoid it being copy-pasted outside of this context.

AaronFeickert: Something something premature optimization...

bewakes (dismissed):

crates/test-utils/src/lib.rs
purusang: > being consistent to avoid it being copy-pasted outside of this context.

Sorry, I could not get the context of it. @AaronFeickert

storopoli: Yes we can move it outside of the `struct`. Let it take a (mutable) reference to an PRNG, and the user bring either a fast one (in the case of testing and fuzzing) or a slow and safe (CSPRNG) one.

storopoli: Just put a trait bound on https://docs.rs/rand/latest/rand/trait.RngCore.html and let the user bring either a https://docs.rs/rand/latest/rand/rngs/struct.OsRng.html which impls `RngCore` or something faster like https://docs.rs/rand/latest/rand/trait.RngCore.html#impl-RngCore-for-ChaCha12Rng that also impls `RngCore`.

storopoli, line 13: Also please import `ThreadRng` and not use absolute hard-coded imports.

AaronFeickert: I recommend requiring `CryptoRngCore` (which is just `RngCore + CryptoRng`) to mitigate against the unintentional misuse of obviously bad generators. Even for fast testing operations, surely something like `ChaCha8Rng` would be performant enough.

Also note that `CryptoRngCore` does _not_ guarantee the generator's output is of high entropy!

delbonis: I mean, it's only used for tests, there's no such thing as a bad generator.

AaronFeickert: [orly](https://xkcd.com/221/)

AaronFeickert: I kid, but I think this highlights a suboptimal design for which better design patterns exist.

AaronFeickert: Plus I am generally a fan of having tests reflect optimal behavior as much as possible. You never know who will take a design pattern from a test and use it where they shouldn't.

crates/test-utils/src/lib.rs
```diff
@@ -1,46 +1,54 @@
...
+use rand::{rngs::OsRng, CryptoRng, RngCore};
 
 pub mod bitcoin;
 pub mod bridge;
 pub mod evm_ee;
 pub mod l2;
 
-const ARB_GEN_LEN: usize = 1 << 24; // 16 MiB
+// Smaller buffer size as compared to 2^24
+const ARB_GEN_LEN: usize = 128;
```
storopoli, line 10: Upgrade this comment to a full blown docstring.

crates/test-utils/src/lib.rs
```diff
@@ -1,46 +1,54 @@
...
+    pub fn new_with_size(s: usize) -> Self {
+        ArbitraryGenerator { buf: vec![0u8; s] }
     }
 
-    pub fn new_with_size(n: usize) -> Self {
-        let mut buf = vec![0; n];
-        OsRng.fill_bytes(&mut buf); // 128 wasn't enough
-        let off = AtomicUsize::new(0);
-        ArbitraryGenerator { buf, off }
+    /// Legacy interface: no arguments
```
storopoli, line 30: Is this deprecated?
Maybe add a https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute ?

Or maybe add a note that this uses `OsRng` in the docstrings.

crates/test-utils/src/lib.rs
```diff
@@ -1,46 +1,54 @@
...
-    pub fn generate<'a, T: Arbitrary<'a> + Clone>(&'a self) -> T {
-        // Doing hacky atomics to make this actually be reusable, this is pretty bad.
-        let off = self.off.load(Ordering::Relaxed);
-        let mut u = Unstructured::new(&self.buf[off..]);
-        let prev_off = u.len();
-        let inst = T::arbitrary(&mut u).expect("failed to generate arbitrary instance");
-        let additional_off = prev_off - u.len();
-        self.off.store(off + additional_off, Ordering::Relaxed);
-        inst
+    /// Core function: accepts an optional RNG
```
storopoli, line 38: What is the difference between a `Legacy` and a `Core`?
Please reword. Imagine someone in a code editor trying to call these functions and seeing something like "Ok do I want to call the legacy interface or the core function?".

storopoli (changes requested):
I still think that the interface how it is presented is confusing.
I would need to inspect the code to use these functions in a code editor when seeing the docstrings of the functions.

Which is really bad experience, right?

crates/test-utils/src/lib.rs
purusang: Sure, will make it clear.

crates/test-utils/src/lib.rs
```diff
@@ -1,46 +1,54 @@
...
+    pub fn generate_with_rng<'a, T, R>(&'a mut self, rng: Option<&mut R>) -> T
+    where
+        T: Arbitrary<'a> + Clone,
+        R: RngCore + CryptoRng,
+    {
```
delbonis, line 43: This interface is clunky.  Have a fn that take an rng unconditionally and the other fn will just call that with a default ThreadRng/OsRng/whatever.

crates/test-utils/src/lib.rs
delbonis: I wouldn't call it "legacy", it's a convenience function.

crates/test-utils/src/lib.rs
delbonis: Maybe bump it up to 1024 at least?

crates/test-utils/src/lib.rs
```diff
@@ -1,46 +1,54 @@
...
         Self::new()
     }
 }
-
 impl ArbitraryGenerator {
     pub fn new() -> Self {
-        Self::new_with_size(ARB_GEN_LEN)
+        ArbitraryGenerator {
+            buf: vec![0u8; ARB_GEN_LEN],
+        }
```
delbonis, line 24: `Self::new_with_sice(ARB_GEN_LEN)`

crates/test-utils/src/lib.rs
```diff
@@ -1,46 +1,54 @@
...
+    }
+    pub fn new_with_size(s: usize) -> Self {
+        ArbitraryGenerator { buf: vec![0u8; s] }
```
delbonis, line 27: Formatting and should be `Self { ... }`.

delbonis (changes requested):
Some stylistic things.

AaronFeickert (changes requested):
Agree with the other review comments.

crates/test-utils/src/lib.rs
purusang: It's not deprecated actually, I updated the docs which provides context.

bewakes (dismissed):

crates/test-utils/src/lib.rs
purusang: Agree

delbonis (approved):

delbonis:
Another thing I just realized we *could* do is to *allocate* a large buffer for entropy (like 16M) upfront, but only write to it 1K at a time as requested by the type we're generating.  We'd have to screw around a little bit to make the lifetimes work but it'd be fine and safe.

AaronFeickert (approved):
LGTM

storopoli (approved):
ACK 1deafdf338f8347e5d1a51b4e94dfd84b8260020

Nice job!

PR #508, opened 2024-12-01
Build bitcoind image from official source

purusang:
## Description
Use of bitcoin source to build image
- uses v28.0 tag from bitcoin repo
- applies a specific commits from a pr that provisions custom blocktime for signet
- updates the tag of bitcoin client in docker compose to use/build new image

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues
STR-533

### DIFF
--- a/docker/bitcoind/Dockerfile
+++ b/docker/bitcoind/Dockerfile
@@ -1,47 +1,58 @@
-# Use the official Ubuntu 24.04 base image
-FROM ubuntu:24.04
+# Stage 1: Build Bitcoin from source
+FROM ubuntu:24.04 AS build
 
-WORKDIR /app
+WORKDIR /build
 
-# Install necessary dependencies
+# Install build dependencies
 RUN apt-get update && \
-    apt-get install -y software-properties-common build-essential libtool autotools-dev automake pkg-config bsdmainutils python3 && \
-    apt-get install -y libevent-dev libboost-dev libsqlite3-dev && \
-    apt-get install -y libssl-dev libboost-all-dev git
-
-# Set Git with dummy user configuration
-# This configuration is needed for cherry-picking the specific commit
-RUN git config --global user.email "john@example.com" && \
-    git config --global user.name "John Doe"
+    apt-get install -y software-properties-common build-essential libtool autotools-dev automake pkg-config bsdmainutils python3 \
+    libevent-dev libboost-dev libsqlite3-dev libssl-dev libboost-all-dev git
 
 # Clone Bitcoin repository and checkout the commit for v28.0 tag
-RUN git clone https://github.com/bitcoin/bitcoin.git && \
-    cd bitcoin && \
-    git checkout 5de225f5c145368f70cb5f870933bcf9df6b92c8
+RUN git clone -b 'v28.0' --single-branch --depth 1 https://github.com/bitcoin/bitcoin.git
 
-# Apply the specific pr https://github.com/bitcoin/bitcoin/pull/27446/commits
-# which includes the changes needed for configurable block time for signet
-RUN cd bitcoin && \
-    git fetch origin pull/27446/head && \
-    git cherry-pick d8434da3c14ed6723d86ef2cd266008d366e1413
+# Apply a diff file to the Bitcoin source code to enable configurable block
+# time for signet. The diff file is generated from the following PR:
+# https://github.com/bitcoin/bitcoin/pull/27446/commits
+COPY configurable_blocktime_patch.diff configurable_blocktime_patch.diff
+
+# Apply the patch to Bitcoin source code
+RUN cd bitcoin && git apply /build/configurable_blocktime_patch.diff
 
 # Build Bitcoin from source
+# Run the build in parallel using one less than the number of available CPUs
 RUN cd bitcoin && \
     ./autogen.sh && \
     ./configure && \
-    make && \
+    make -j$(($(nproc) - 1)) && \
     make install
+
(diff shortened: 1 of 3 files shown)


delbonis (approved):

docker/bitcoind/Dockerfile
```diff
@@ -1,19 +1,36 @@
...
+
+# Set Git with dummy user configuration
+# This configuration is needed for cherry-picking the specific commit
+RUN git config --global user.email "john@example.com" && \
+    git config --global user.name "John Doe"
+
+# Clone Bitcoin repository and checkout the commit for v28.0 tag
+RUN git clone https://github.com/bitcoin/bitcoin.git && \
+    cd bitcoin && \
+    git checkout 5de225f5c145368f70cb5f870933bcf9df6b92c8
```
bewakes, line 20: Would it make sense to use a specific tag here? I can see they have a tag `v28.0`

purusang: Yeah, that would be more readable but hash is kinda source of truth here so leaving as is.

docker/bitcoind/Dockerfile
```diff
@@ -1,19 +1,36 @@
...
+
```
storopoli, line 21: Yes @bewakes is right, use https://github.com/bitcoin/bitcoin/releases/tag/v28.0
```suggestion
RUN git clone -b 'v28.0' --single-branch --depth 1 https://github.com/bitcoin/bitcoin.git && 
    cd bitcoin
```
Also you don't need to pull all the ~crap~ things from the repo, you can do `--single-branch` and `--depth 1`

> [!TIP]
> You might want to make `v28.0` configurable.

docker/bitcoind/Dockerfile
```diff
@@ -1,19 +1,36 @@
...
+# Apply the specific pr https://github.com/bitcoin/bitcoin/pull/27446/commits
+# which includes the changes needed for configurable block time for signet
+RUN cd bitcoin && \
+    git fetch origin pull/27446/head && \
+    git cherry-pick d8434da3c14ed6723d86ef2cd266008d366e1413
```
storopoli, line 26: I don't like this. The PR can be changed and we can get rug-pulled (EDIT: not exactly since we are `cherry-pick`ing by the long SHA, but still my tinfoil-hat spider-senses are going crazy)
Can you grab that diff and put in a `30_sec_block.diff` file and then apply that diff?
This way we are totally sure that the diff will not change under us.

storopoli (changes requested):

docker/bitcoind/Dockerfile
```diff
@@ -1,19 +1,36 @@
...
-    apt-get update && \
-    apt-get install -y \
-    bitcoind \
-    curl && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
-
+    apt-get install -y software-properties-common build-essential libtool autotools-dev automake pkg-config bsdmainutils python3 && \
+    apt-get install -y libevent-dev libboost-dev libsqlite3-dev && \
+    apt-get install -y libssl-dev libboost-all-dev git
```
voidash, line 10: I think it is better to replace three `apt-get` command with single one if we are concerned about number of layers in docker image.

docker/bitcoind/Dockerfile
storopoli: Source: https://stackoverflow.com/a/31666461

docker/bitcoind/Dockerfile
voidash: I think tagging would be better. 
 `git clone --branch v28.0 --depth 1 https://github.com/bitcoin/bitcoin.git && cd bitcoin` would simplify things

docker/bitcoind/Dockerfile
purusang: I think this will create only 1 layer.

docker/bitcoind/Dockerfile
purusang: Gonna update accordingly, agree on cloning only the specific tag.

docker/bitcoind/Dockerfile
purusang: Yeah, having hash assures nothing else is applied. But I like the idea of having a patch file which show what difference is being on top of the tag. Testing in local.

docker/bitcoind/Dockerfile
```diff
@@ -1,30 +1,58 @@
...
+
+# Apply the patch to Bitcoin source code
+RUN cd bitcoin && git apply /build/configurable_blocktime_patch.diff
+
+# Build Bitcoin from source
+# Run the build in parallel using one less than the number of available CPUs
+RUN cd bitcoin && \
+    ./autogen.sh && \
+    ./configure && \
+    make -j$(($(nproc) - 1)) && \
```
storopoli, line 27: This looks like a crazy child of FORTRAN and Lisp.

storopoli (approved):
ACK c1f1ea72fe747210aacf3995cbae9d8c001edb92

docker/bitcoind/Dockerfile
purusang: Anyway addressed this.

docker/bitcoind/Dockerfile
delbonis: This makes it less reproducible fwiw.

delbonis (approved):

voidash (approved):

PR #510, opened 2024-12-02
Introduce `ProofReceipt` struct

prajwolrg:
## Description

The current `Proof` type in `zkvm/src/proof` is overloaded, sometimes including public values and sometimes not. This PR tries to address this ambiguity needs by creating `ProofReceipt` struct .

```rust
pub struct ProofReceipt {
  proof: Proof,
  public_values: PublicValues
}
```

### Type of Change

- [ ] Documentation update
- [X] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


crates/zkvm/zkvm/src/proof.rs
```diff
@@ -2,80 +2,98 @@ use arbitrary::Arbitrary;
...
-    }
-
-    pub fn as_bytes(&self) -> &[u8] {
-        &self.0
-    }
+pub struct ProofReceipt {
+    /// The validity proof.
+    pub proof: Proof,
+    /// The public values associated with the proof.
+    pub public_values: PublicValues,
```
storopoli, line 59: I like the idea bu to be honest I think we need to make these fields private and implement a pub getter. What do you think?
I am not that happy with something important as proofs receipt having public fields. I think that it kinda future-proofs it if someone wants to add something else to the `ProofReceipt` since everything is private with public getters.

storopoli (changes requested):
Nice work with the docstrings. I think this is an overall good idea. Just have 1 thing to discuss.

crates/zkvm/adapters/risc0/src/host.rs
```diff
@@ -197,8 +185,9 @@ outputs",
 
         let filename = "proof-groth16.bin";
         let mut file = File::create(filename).unwrap();
-        file.write_all(proof.as_bytes()).unwrap();
+        file.write_all(&bincode::serialize(&proof).expect("bincode serialization failed"))
+            .unwrap();
```
delbonis, line 189: Should probably be a little better about this.  I didn't catch the `.unwrap()` before.  Also *must* this be bincode here or can it be something more generic/stable?

crates/zkvm/adapters/risc0/src/input.rs
```diff
@@ -36,9 +38,10 @@ impl<'a> ZkVmInputBuilder<'a> for Risc0ProofInputBuilder<'a> {
         Ok(self)
     }
 
-    fn write_proof(&mut self, item: strata_zkvm::AggregationInput) -> ZkVmResult<&mut Self> {
+    fn write_proof(&mut self, item: &AggregationInput) -> ZkVmResult<&mut Self> {
         // Learn more about assumption and proof compositions at https://dev.risczero.com/api/zkvm/composition
-        let receipt: Receipt = bincode::deserialize(item.proof().as_bytes())?;
+        let journal = item.receipt().public_values.as_bytes().to_vec();
+        let inner: InnerReceipt = bincode::deserialize(item.receipt().public_values.as_bytes())?;
```
delbonis, line 44: Same here about bincode.  It might be fine, not sure what is expected or what needs to be stable.

crates/zkvm/adapters/sp1/src/host.rs
```diff
@@ -63,23 +67,22 @@ impl ZkVmHost for SP1Host {
...
+        let proof_info = prover
             .run()
             .map_err(|e| ZkVmError::ProofGenerationError(e.to_string()))?;
 
         // Proof serialization
-        let serialized_proof = bincode::serialize(&proof)?;
-
-        Ok((Proof::new(serialized_proof), self.get_verification_key()))
-    }
+        let proof = Proof::new(bincode::serialize(&proof_info.proof)?);
```
delbonis, line 75: Same here.  Isn't the proof already just a `[u8; 256]` or similar?

crates/zkvm/zkvm/src/proof.rs
```diff
@@ -2,80 +2,98 @@ use arbitrary::Arbitrary;
...
 
-/// Validity proof generated by the `ZkVmHost`
+/// Represents a wrapper around a byte vector for various proofs and keys.
+///
+/// Provides common utilities such as byte access and emptiness checks.
 #[derive(
     Debug, Clone, Serialize, Deserialize, BorshSerialize, BorshDeserialize, PartialEq, Eq, Arbitrary,
 )]
-pub struct Proof(Vec<u8>);
+pub struct ByteWrapper(Vec<u8>);
```
delbonis, line 11: Maybe a better name for this is like `ProverBuf` or something?  Why can't we just be using `Vec<u8>` directly in all these places?

delbonis (changes requested):

crates/zkvm/zkvm/src/proof.rs
prajwolrg: Yup makes sense! Thanks!

crates/zkvm/adapters/risc0/src/host.rs
prajwolrg: It need not be `bincode`. 
The `.unwrap()` is here used in the test to save the proof locally since generating it takes some time.

prajwolrg: But since the proof is already serialized, we can skip this serialization. Thanks!

crates/zkvm/adapters/risc0/src/input.rs
prajwolrg: Right now, `Proof` that we have defined actually is the serialized version of:
- `InnerReceipt` in case of Risc0
- `SP1Proof` in case of SP1

To serialize/deserialize we can use any supported serializer/deserializer. It should be consistent in the 
- `ZkVmHost::prove` - where the serialization happens
- `ZkVmHost::verify` - where they are converted back to their expected type
- `ZkVmInputBuilder::write_proof` - where the proof is written as input for the next program.

crates/zkvm/zkvm/src/proof.rs
prajwolrg: Yes, `ProofBuf` makes sense. `ByteWrapper` has been introduced, because we were exposing same functions for the `Proof`, `PublicValues` and the `VerificationKey` all of which have a different type for different ZkVm.

crates/zkvm/adapters/risc0/src/host.rs
delbonis: Oh my bad.

delbonis (approved):

MdTeach (approved):

PR #512, opened 2024-12-02
STR-211: CI build checks for the prover guest code

evgenyzdanovich:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/.github/workflows/prover.yml
+++ b/.github/workflows/prover.yml
@@ -15,21 +15,25 @@ jobs:
         uses: dtolnay/rust-toolchain@nightly
         with:
           toolchain: nightly-2024-07-27
-      - uses: Swatinem/rust-cache@v2
+      
+      # Installs mold (modern ld), a drop-in replacement for lld.
+      # Under the hood, the following actions symlinks mold binary onto lld,
+      # so everything is linked faster (hopefully).
+      - name: Install mold linker
+        uses: rui314/setup-mold@v1
+
+      - name: Set up Clang
+        uses: egor-tensin/setup-clang@v1
         with:
-          cache-on-failure: true
+          version: latest
+          platform: x64
 
       - name: Install SP1 Toolchain
         run: |
           curl -L https://sp1.succinct.xyz | bash && ~/.sp1/bin/sp1up
           export PATH=~/.sp1/bin:$PATH
           cargo prove --version
-      - name: Set up Clang
-        uses: egor-tensin/setup-clang@v1
-        with:
-          version: latest
-          platform: x64
-  
+
       - name: Checkout code
         uses: actions/checkout@v4
 
--- a/.github/workflows/prover.yml
+++ b/.github/workflows/prover.yml
@@ -34,4 +34,4 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Build guest code
-        run: cd provers/sp1/ && cargo build --release -F "prover"
+        run: cd provers/sp1/ && RUSTFLAGS="-C link-arg=-fuse-ld=lld" cargo build --profile prover-ci -F "prover"
--- a/.github/workflows/prover.yml
+++ b/.github/workflows/prover.yml
@@ -48,8 +48,8 @@ jobs:
         env:
           RUSTFLAGS: "-C link-arg=-fuse-ld=lld"
   
-      - name: Build prover client
+      - name: Build prover client binary
         run: cargo build --profile prover-ci -F "prover"
-        working-directory: provers/sp1
+        working-directory: bin/prover-client
         env:
           RUSTFLAGS: "-C link-arg=-fuse-ld=lld"
--- a/.github/workflows/prover.yml
+++ b/.github/workflows/prover.yml
@@ -16,8 +16,13 @@ jobs:
         with:
           toolchain: nightly-2024-07-27
       
+      - name: Use Cargo cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: true
+      
       # Installs mold (modern ld), a drop-in replacement for lld.
-      # Under the hood, the following actions symlinks mold binary onto lld,
+      # Under the hood, the following action symlinks mold binary onto lld,
       # so everything is linked faster (hopefully).
       - name: Install mold linker
         uses: rui314/setup-mold@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/prover.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: egor-tensin/setup-clang@v1
+        with:
+          version: latest
+          platform: x64
+  
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Build guest code
+        run: cd provers && cargo build --release -F "prover"
```
MdTeach, line 37: Can we save some build time if we build only these instead of entire project:
* `bin/prover-client`
*  `provers/sp1`

MdTeach (commented):
Should we also add build check for the prover-client binary here ?
I think in normal CI build it will be also skipped

https://github.com/alpenlabs/strata/blob/08ea6465983d85e761e0d33702ca791d1bf15848/bin/prover-client/Cargo.toml#L6-L9

evgenyzdanovich:
> Should we also add build check for the prover-client binary here ? I think in normal CI build it will be also skipped
> 
> https://github.com/alpenlabs/strata/blob/08ea6465983d85e761e0d33702ca791d1bf15848/bin/prover-client/Cargo.toml#L6-L9

@MdTeach Thanks a lot!!! You are completely right. I missed one directory, didn't realize that there's no `cargo.toml` under `provers/` and it was building the whole thing (including reth and all other binaries). 

Trying to restrict it.

delbonis:
Why were all those files deleted?

evgenyzdanovich:
@delbonis to not bloat the actions tab on the repository, because I experiment a lot and run CI on this branch frequently.
I'll add it back definitely once the PR is ready for the review.

.github/workflows/prover.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Set up Rust
+        uses: dtolnay/rust-toolchain@nightly
+        with:
+          toolchain: nightly-2024-07-27
+      
+      # Installs mold (modern ld), a drop-in replacement for lld.
+      # Under the hood, the following actions symlinks mold binary onto lld,
+      # so everything is linked faster (hopefully).
+      - name: Install mold linker
+        uses: rui314/setup-mold@v1
```
storopoli, line 23: `mold` is awesome. Great addition.

.github/workflows/prover.yml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Set up Clang
+        uses: egor-tensin/setup-clang@v1
+        with:
+          version: latest
+          platform: x64
+
+      - name: Install SP1 Toolchain
+        run: |
+          curl -L https://sp1.succinct.xyz | bash && ~/.sp1/bin/sp1up
```
storopoli, line 33:
```suggestion
          curl -fsSL --proto '=https' --tlsv1.2 https://sp1.succinct.xyz | bash && ~/.sp1/bin/sp1up
```
til-foil hat mode

.github/workflows/prover.yml
```diff
@@ -0,0 +1,41 @@
...
+          export PATH=~/.sp1/bin:$PATH
+          cargo prove --version
+
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Build guest code
+        run: cd provers/sp1/ && RUSTFLAGS="-C link-arg=-fuse-ld=lld" cargo build --profile prover-ci -F "prover"
```
storopoli, line 41:
```suggestion
        run: cargo build --profile prover-ci -F "prover"
        working-directory: provers/sp1
        env:
          RUSTFLAGS: "-C link-arg=-fuse-ld=lld"
```

see https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsworking-directory

and https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsenv

.github/workflows/prover.yml
evgenyzdanovich: Agree. 

Honestly, I didn't know much about other linkers, but just recently discovered it as part of this task (while digging various articles about how people **actually** **optimize compile time**). It shows super promising results!

.github/workflows/prover.yml
evgenyzdanovich: done :)

.github/workflows/prover.yml
evgenyzdanovich: Hm, nice, thanks!

storopoli:
You should also probably cache this thing. Use the setup that we're doing in other GH actions: https://github.com/alpenlabs/strata/blob/74250151c2561bbbf6625595312cf487d6341dbe/.github/workflows/docs.yml#L22-L24

.github/workflows/prover.yml
storopoli: Too bad that it's abandone for MacOS

.github/workflows/prover.yml
```diff
@@ -0,0 +1,47 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Set up Rust
+        uses: dtolnay/rust-toolchain@nightly
+        with:
+          toolchain: nightly-2024-07-27
+      - uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: true
```
storopoli, line 20: Final nit
In every step in this file you're doing with `name`.
Why in this it is straight into `uses`?
And why this is the only step that is not surrounded by empty lines?

evgenyzdanovich: >In every step in this file you're doing with name.
I've seen the same pattern in `functional.yml`

>And why this is the only step that is not surrounded by empty lines?
I treat it like a code. Usually, block of the code that delivers some logic goes line-by-line and blocks of logic are separated by an empty line. With that said, I think setting up rust and configuring its cache is the same "logical block", while, installing `clang` / `mold` / `sp1` are separate "logical steps/blocks"

Also I've seen the same in `functional.yml`.

Feel free to counter it and tell me if it's ugly, I'd be happy to fix it.

storopoli:
```suggestion

      - name: Rust cache
        uses: Swatinem/rust-cache@v2
        with:
          cache-on-failure: true
```
Let's just be consistent in this file. We can standardize later if necessary across the YAML CI files.

storopoli (approved):
ACK 8d38d21778bb6350a19ebf8f64221a8ef5a31bec

MdTeach (dismissed):

delbonis (dismissed):

.github/workflows/prover.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: cargo build --profile prover-ci -F "prover"
+        working-directory: provers/sp1
+        env:
+          RUSTFLAGS: "-C link-arg=-fuse-ld=lld"
+  
+      - name: Build prover client
+        run: cargo build --profile prover-ci -F "prover"
+        working-directory: provers/sp1
+        env:
+          RUSTFLAGS: "-C link-arg=-fuse-ld=lld"
```
storopoli, line 55: What is the difference between `Build prover guest code` and `Build prover client`?
image

storopoli (changes requested):

.github/workflows/prover.yml
evgenyzdanovich: renamed.

storopoli: No, that's not it.
It is the SAME command with the SAME env vars and the SAME working dir.

evgenyzdanovich: ah, right. Thanks for noticing! Fixed

storopoli (approved):
ACK 1a8b67797d63eb8297e1743e87bf4aa87cd86893

storopoli (approved):
ACK 0ea3f662c9172a1c893d4d9e0ad755ae859e2e37

PR #514, opened 2024-12-02
Str 106 csm test util

purusang:
## Description
Consolidates repeated testing logic to a single function.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
Please suggest on handling test_l1_reorg(). Do we not apply writes for this particular case?

## Checklist

## Related Issues
STR-106

### DIFF
--- a/crates/consensus-logic/src/csm/client_transition.rs
+++ b/crates/consensus-logic/src/csm/client_transition.rs
@@ -353,238 +353,6 @@ mod tests {
     use super::*;
     use crate::genesis;
 
-    #[test]
-    fn test_genesis() {
-        // TODO there's a ton of duplication in this test, we could wrap it up so we just run
-        // through a table and call a function with it
-
-        let database = get_common_db();
-        let params = gen_params();
-        let mut state = gen_client_state(Some(&params));
-
-        assert!(!state.is_chain_active());
-
-        let horizon = params.rollup().horizon_l1_height;
-        let genesis = params.rollup().genesis_l1_height;
-
-        let chain = get_btc_chain();
-        let l1_chain = chain.get_block_manifests(horizon as u32, 10);
-        let l1_verification_state =
-            chain.get_verification_state(genesis as u32 + 1, &MAINNET.clone().into());
-
-        let l1_db = database.l1_db();
-        for (i, b) in l1_chain.iter().enumerate() {
-            l1_db
-                .put_block_data(i as u64 + horizon, b.clone(), Vec::new())
-                .expect("test: insert blocks");
-        }
-
-        let blkids: Vec<L1BlockId> = l1_chain.iter().map(|b| b.block_hash().into()).collect();
-
-        // at horizon block
-        {
-            let height = horizon;
-            let idx = (height - horizon) as usize;
-            let l1_block_id = l1_chain[idx].block_hash().into();
-            let event = SyncEvent::L1Block(height, l1_block_id);
-
-            let output = process_event(&state, &event, database.as_ref(), &params).unwrap();
-
-            let writes = output.writes();
-            let actions = output.actions();
-
-            let expected_writes = [ClientStateWrite::AcceptL1Block(l1_block_id)];
-            let expected_actions = [];
-
-            assert_eq!(writes, expected_writes);
-            assert_eq!(actions, expected_actions);
-
-            operation::apply_writes_to_state(&mut state, writes.iter().cloned());
-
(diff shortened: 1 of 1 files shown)


crates/consensus-logic/src/csm/client_transition.rs
```diff
@@ -353,17 +353,61 @@ mod tests {
...
+        event: SyncEvent,
+        expected_writes: Vec<ClientStateWrite>,
+        expected_actions: Vec<SyncAction>,
+    }
+
+    struct TestCase {
+        description: &'static str,
+        events: Vec<TestEvent>, // List of events to process
+        state_assertions: Box<dyn Fn(&ClientState)>, // Closure to verify state after all events
+    }
```
bewakes, line 366: Since we are having state_assertions as a function, can we make this generic? something like:

```rust
struct TestRunner<Ctx, Inp, Out, State> {
    description: &'static str,
    inputs: Vec<TestCase<Inp, Out>>,
    state_assertions: Box<dyn Fn(&Ctx, State)>  
}

struct TestCase<I, O> {
    input: I,
    expected_output: O
}
```
And then we can have it in `test-utils`.

crates/consensus-logic/src/csm/client_transition.rs
```diff
@@ -372,196 +416,203 @@ mod tests {
...
-
-            assert_eq!(output1.writes(), expected_writes1);
-            assert_eq!(output1.actions(), expected_actions1);
-
-            assert_eq!(output2.writes(), expected_writes2);
-            assert_eq!(output2.actions(), expected_actions2);
-
-            operation::apply_writes_to_state(&mut state, output1.writes().iter().cloned());
-            operation::apply_writes_to_state(&mut state, output2.writes().iter().cloned());
+        let test_cases = vec![
```
bewakes, line 430: It seems to me like the test cases are more structured now, but are not very readable. Maybe some parameterized helper functions could make these more readable.

crates/consensus-logic/src/csm/client_transition.rs
delbonis: That would be such a pain in the ass and very much confuse the autocompletion.  It's meant to be streamlined and easy to reason about.

bewakes: Really? I thought that could be a nice abstraction to common patterns that occur on tests.

crates/consensus-logic/src/csm/client_transition.rs
```diff
@@ -353,22 +353,22 @@ mod tests {
     use super::*;
     use crate::genesis;
 
-    struct TestEvent {
+    struct TestEvent<'a> {
         event: SyncEvent,
-        expected_writes: Vec<ClientStateWrite>,
-        expected_actions: Vec<SyncAction>,
+        expected_writes: &'a [ClientStateWrite],
+        expected_actions: &'a [SyncAction],
```
bewakes, line 359: I think that these make more sense to be inside `TestCase` struct.
Since these expected values will eventually be used to assert things, I think they be together with `state_assertions`.

What do you think?

bewakes: Ah, I see it now after a discussion, these expected values are meant to run for each event and state assertions are meant to run after processing all of the events.

purusang: Yeah, that's right.

crates/consensus-logic/src/csm/client_transition.rs
delbonis: We really just care about the csm logic here.  Sounds like premature abstraction imo.

delbonis (approved):

crates/consensus-logic/src/csm/client_transition.rs
purusang: I think this could be a good addition if we could have test cases generalized for all cases. For this particular case, different type of assertions are before and after application of writes. There could be such special assertion scenarios in other test cases too. That means different impl. May be we need to take some more time to generalize the struct for all the test cases if needed. So for the scope of this work I'm going to defer generalization for now and may be we can have it in another pr.

bewakes: Agree with both of you. Since this is just a single place which we are doing, it makes sense to defer it until the pattern naturally emerges in other testcases.

bewakes (approved):

PR #515, opened 2024-12-03
Update ProverDB

prajwolrg:
## Description

This PR updates the `ProverDB` implementation to replace UUID-based indexing with a more intuitive and readable `ProofId`. 

### Type of Change

- [ ] Documentation update
- [X] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

 - `crates/primitives/src/vk.rs` has been renamed to `proof.rs` since it now also includes `ProofId`.
 - In the `prover-client`, saving and getting proof is broken. This is alright for now, because the proof was never read and is always stored in memory.

## Checklist

## Related Issues

STR-735


bin/prover-client/src/prover.rs
```diff
@@ -338,22 +339,23 @@ impl Prover {
...
 
-    fn save_proof_to_db(&self, task_id: Uuid, proof: &Proof) -> Result<(), anyhow::Error> {
-        self.db
-            .prover_task_db()
-            .insert_new_task_entry(*task_id.as_bytes(), proof.into())?;
+    // TODO: replace `task_id` with ProofId
+    fn save_proof_to_db(&self, _task_id: Uuid, _proof: &Proof) -> Result<(), anyhow::Error> {
+        // Note: This works fine for now because proof is never read from DB right now
+        // let proof_id = ProofId::BtcBlockspace(1);
+        // self.db.proof_db().insert_proof(proof_id, proof.clone())?;
```
bewakes, line 346: Is this supposed to be commented out?

crates/rocksdb-store/src/lib.rs
```diff
@@ -60,8 +60,7 @@ use bridge::schemas::{
 };
 pub const PROVER_COLUMN_FAMILIES: &[ColumnFamilyName] = &[
     SequenceSchema::COLUMN_FAMILY_NAME,
-    prover::schemas::ProverTaskIdSchema::COLUMN_FAMILY_NAME,
-    prover::schemas::ProverTaskSchema::COLUMN_FAMILY_NAME,
```
bewakes, line 64: Why do we not need the prover tasks though?

prajwolrg: Prover Tasks will be implied by the use of `ProofId`.
For example `ProofId::L1Batch(u64, u64)` will automatically imply that it should get proofs from`ProofId::BtcBlockspace(u64)` till `ProofId::BtcBlockspace(u64)`.

bin/prover-client/src/prover.rs
prajwolrg: Yes, for now! Since we don't use DB is the prover-client, this is fine for now.
This will be used on the upcoming PR that'll actually use this.
This is done so to reduce the scope of this PR.

crates/primitives/src/proof.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use crate::buf::Buf32;
+
+#[derive(Clone, Debug, Copy, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum RollupVerifyingKey {
+    // Verifying Key used to verify proof created using SP1
+    SP1VerifyingKey(Buf32),
+    // Verifying Key used to verify proof created using Risc0
+    Risc0VerifyingKey(Buf32),
```
delbonis, line 12: While we're here can we just do `#[serde(rename = "sp1")]` and `#[serde(rename = "r0")]` so we don't have the annoying `s_p1_verifiying_key` thing.

crates/primitives/src/proof.rs
```diff
@@ -0,0 +1,38 @@
...
+}
+
+/// `ProofId` is an enumeration representing various identifiers for proofs.
+/// It is used to uniquely identify the proofing tasks for different types of proofs.
+#[derive(
+    Debug, Clone, Copy, PartialEq, BorshSerialize, BorshDeserialize, Serialize, Deserialize,
+)]
+pub enum ProofId {
```
delbonis, line 20: Maybe a better name would be `proof_key` since there's more semantic information associated with it now?

crates/primitives/src/proof.rs
```diff
@@ -0,0 +1,38 @@
...
+    /// Represents the height of the EVM Execution Environment (EE) block for which
+    /// the State Transition Function (STF) proof is being generated.
+    EvmEeStf(u64),
+
+    /// Represents the height of the Consensus Layer (CL) block for which
+    /// the State Transition Function (STF) proof is being generated.
+    ClStf(u64),
+
+    /// Represents the index of the checkpoint that is being proven.
+    Checkpoint(u64),
```
delbonis, line 37: All of these should *probably* be based on blkids since we want to be able to handle reorgs somehow.

crates/db/src/traits.rs
```diff
@@ -265,32 +266,21 @@ pub trait BlobDatabase {
...
-    /// Get next index to be inserted to
-    fn get_next_task_idx(&self) -> DbResult<u64>;
-
-    /// Get transaction id for index
-    fn get_taskid(&self, idx: u64) -> DbResult<Option<[u8; 16]>>;
+pub trait ProofDatabase {
+    /// Inserts a proof into the database.
+    ///
+    /// Returns `Ok(())` on success, or an error on failure.
+    fn insert_proof(&self, proof_id: ProofId, proof: Proof) -> DbResult<()>;
```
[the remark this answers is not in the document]
prajwolrg: Sure. I have also added `del_proof`.
Not sure what can be other way to search the proof. I think we should be able to derive the proof ID easily given access to CL / EL / BTC client.

functional-tests/rollup_params_cfg.py
```diff
@@ -18,7 +18,7 @@ def get_operators_pubkeys(self) -> list[str]:
 
 
 class RollupVk(BaseModel):
-    s_p1_verifying_key: str
+    sp1: str
```
MdTeach, line 21: Good to see `s_p1_verifying_key` go away 🎉

crates/primitives/src/proof.rs
```diff
@@ -0,0 +1,42 @@
...
+    #[serde(rename = "risc0")]
+    Risc0VerifyingKey(Buf32),
+}
+
+/// `ProofKey` is an enumeration representing various identifiers for proofs.
+/// It is used to uniquely identify the proving tasks for different types of proofs.
+#[derive(
+    Debug, Clone, Copy, PartialEq, BorshSerialize, BorshDeserialize, Serialize, Deserialize,
+)]
+pub enum ProofKey {
```
MdTeach, line 22: Do you think the term `ProofKey` will create ambiguity with the term ProvingKey ?

MdTeach (approved):

crates/primitives/src/proof.rs
prajwolrg: Not so sure. 
But since we will be interacting with host (which internally has the ProvingKey), this should not create much ambiguity. AFAIK we haven't declared ProvingKey anywhere in our codebase.

crates/db/src/traits.rs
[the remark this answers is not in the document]
delbonis: Yeah it's probably fine to leave it out actually.

crates/rocksdb-store/src/lib.rs
bewakes: That's cool

crates/primitives/src/proof.rs
prajwolrg: I used blkids for a bit. But realized that the Prover works based on the `RpcCheckpointInfo` which is defined mostly in terms of heights.
```rust 
pub struct RpcCheckpointInfo {
    /// The index of the checkpoint
    pub idx: u64,
    /// L1 height  the checkpoint covers
    pub l1_range: (u64, u64),
    /// L2 height the checkpoint covers
    pub l2_range: (u64, u64),
    /// L2 block that this checkpoint covers
    pub l2_blockid: L2BlockId,
}
```
I have created a separate ticket to track this.

storopoli (approved):
ACK 9a5fae2da376e794dd1b7c65abb6209691ef7f21

CODEOWNERS:

- `@alpenlabs/bin`: renaming and minor workaround the `save_proof_to_db`
- `@alpenlabs/ci`: renaming constants in `constants.py`

crates/primitives/src/proof.rs
delbonis: That's rather annoying.  Internally it *definitely* makes sense to do it based on blkids because those are objective.  Especially for the L1 proofs since it's view-dependent.  The RPCs we can rework.

bewakes (dismissed):

delbonis (dismissed):

delbonis:
Doesn't this also close [STR-555](https://alpenlabs.atlassian.net/browse/STR-555)?

prajwolrg:
> Doesn't this also close [STR-555](https://alpenlabs.atlassian.net/browse/STR-555)?

Yes!

MdTeach (dismissed):

mukeshdroid (dismissed):

MdTeach (approved):

bin/prover-client/src/prover.rs
```diff
@@ -3,13 +3,15 @@ use std::{
     sync::{Arc, RwLock},
 };
 
-use strata_db::traits::{ProverDatabase, ProverTaskDatabase};
+use strata_db::traits::{ProofDatabase, ProverDatabase};
+use strata_primitives::proof::{ProofContext, ProofKey, ProofZkVm};
 use strata_proofimpl_evm_ee_stf::ELProofInput;
 use strata_rocksdb::{
     prover::db::{ProofDb, ProverDB},
```
bewakes, line 10: These two structs are likely redundant, could you please verify and merge them into one?

crates/db/src/traits.rs
```diff
@@ -265,32 +270,43 @@ pub trait BlobDatabase {
...
+    /// Returns `Some(proof)` if found, or `None` if not.
+    fn get_proof(&self, proof_id: ProofKey) -> DbResult<Option<ProofReceipt>>;
 
-    /// Fetch [`Vec<u8>`] from db
-    fn get_task_entry_by_id(&self, txid: [u8; 16]) -> DbResult<Option<Vec<u8>>>;
+    /// Deletes a proof by its key.
+    ///
+    /// Tries to delete a proof by its key, returning if it really
+    /// existed or not.  
+    fn del_proof(&self, proof_id: ProofKey) -> DbResult<bool>;
```
bewakes, line 293: the argument name and the type do not quite match each other

delbonis (approved):

bewakes (approved):

MdTeach (approved):

mukeshdroid (approved):

storopoli (approved):
ACK d29181f817d129c569350ce8d930b01b0b6229d2

PR #518, opened 2024-12-04
STR-807: Rework on  Inscription Structure

voidash:
## Description
Following are the main changes made for Inscription Transaction
- Multiple Envelopes in a single script is supported
- Envelope format 
```regex
OP_FALSE
OP_IF
<Tag> (Either Checkpoint or DA)
Data [0..520]
....
OP_ENDIF 
(OP_FALSE
OP_IF // No rollup name required for further envelopes
<Tag> (Either Checkpoint or DA)
Data [0..520]
....
OP_ENDIF
)*
```
- Pivot from inscription type naming system 
### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

STR-807


crates/state/src/tx.rs
```diff
@@ -41,18 +43,46 @@ pub struct DepositRequestInfo {
...
+impl BlobType {
+    /// Used to convert Integer to Type mainly for parsing purpose
+    pub fn from_u32(value: u32) -> Option<Self> {
+        match value {
+            0 => Some(Self::Checkpoint),
+            1 => Some(Self::DA),
+            _ => None,
+        }
+    }
+}
```
delbonis, line 72: Needs a method like:

```rs
pub fn get_tag_bytes(&self, params: &RollupParams) -> &[u8] {
    match self {
        Self::Checkpoint => params.envelope_ckpt_tag.as_bytes()
        Self::DA => params.envelope_da_tag.as_bytes()
    }
}
```

Of course, those fields might be encapsulated in a different structure, but that's the idea.

crates/tx-parser/src/filter.rs
```diff
@@ -162,13 +170,20 @@ mod test {
...
         let inp_tx = create_test_tx(vec![create_test_txout(100000000, &address)]);
         let signed_checkpoint: SignedBatchCheckpoint = ArbitraryGenerator::new().generate();
-        let inscription_data = InscriptionData::new(borsh::to_vec(&signed_checkpoint).unwrap());
+        let inscription_data = (0..num_envelopes)
+            .map(|_| {
+                InscriptionBlob::new(
+                    BlobType::Checkpoint,
+                    borsh::to_vec(&signed_checkpoint).unwrap(),
+                )
+            })
```
delbonis, line 183: Should probably do most of these as DA and only one of them as a checkpoint.

crates/tx-parser/src/filter_types.rs
```diff
@@ -54,3 +55,19 @@ impl TxFilterConfig {
         })
     }
 }
+
+/// Wrapper to hold Inscription Data after parsing and filtering.
+#[allow(clippy::large_enum_variant)]
+pub(crate) enum FilteredInscriptionType {
+    Checkpoint(SignedBatchCheckpoint),
+    DA(Vec<u8>),
+}
```
delbonis, line 64: What does this do?

crates/tx-parser/src/filter_types.rs
delbonis, line 1: Need to extend the filter config to include the `envelope_ckpt_tag` and `envelope_da_tag` fields to properly do the filtering, copied from the rollup params.

crates/tx-parser/src/inscription.rs
```diff
@@ -19,8 +21,8 @@ pub enum InscriptionParseError {
     #[error("Invalid/Missing envelope(NO OP_IF..OP_ENDIF): ")]
     InvalidEnvelope,
     /// Does not have a valid name tag
-    #[error("Invalid/Missing name tag")]
-    InvalidNameTag,
+    #[error("Invalid/Missing Batch tag")]
+    InvalidBatchTag,
```
delbonis, line 25: This and `InvalidBlobTag` don't make sense anymore.  We can treat them both as `UnknownTag` since we have multiple tags we're paying attention to now in the simplified format.

crates/tx-parser/src/inscription.rs
delbonis, line 1: These `_TAG` consts can be removed now.

crates/btcio/src/writer/builder.rs
```diff
@@ -494,26 +489,28 @@ fn assert_correct_address(
...
+            .push_int(inscription.data_type() as i64) // Batch Tag
+            .push_int(inscription.data().len() as i64);
+
+        trace!(batchdata_size = %inscription.data().len(), "Inserting batch data");
+        for chunk in inscription.data().chunks(520) {
+            trace!(size=%chunk.len(), "inserting chunk");
+            builder = builder.push_slice(PushBytesBuf::try_from(chunk.to_vec())?);
+        }
+        builder = builder.push_opcode(OP_ENDIF);
     }
```
delbonis, line 513: This needs to be reworked some more to reflect the format as written in the ticket description.

```
OP_FALSE
OP_IF
  PUSH(tag)
  PUSH(payload[..512])
  PUSH(payload[512..1024])
  ...
OP_ENDIF
```

We're using the type tags also as the magic numbers, so it doesn't need to take `rollup_name` anymore, but instead it should be using the tag strings as resolved from the `InscriptionBlob` type.  This pushes an int based on the variant, which is also implicitly making it dependent on the order that they're listed in the enum.

Perhaps we should lower it down again, resolve the tag bytes from the params based on the blob types ahead of time, and have a function that takes blobs with explicit tag bytes as slices instead?

I'm fairly sure we don't need to also push the length of the payload there do we?  It's implicit based on the number of pushes after that until the ENDIF.

crates/btcio/src/writer/builder.rs
```diff
@@ -494,26 +489,28 @@ fn assert_correct_address(
...
+            builder = builder.push_slice(PushBytesBuf::try_from(rollup_name.as_bytes().to_vec())?);
+        }
+
+        builder = builder
+            .push_int(inscription.data_type() as i64) // Batch Tag
+            .push_int(inscription.data().len() as i64);
+
+        trace!(batchdata_size = %inscription.data().len(), "Inserting batch data");
+        for chunk in inscription.data().chunks(520) {
+            trace!(size=%chunk.len(), "inserting chunk");
```
delbonis, line 509: This log is a big excessive.  Just write unit tests for it all.

crates/state/src/tx.rs
```diff
@@ -41,18 +43,46 @@ pub struct DepositRequestInfo {
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize, Arbitrary)]
+#[borsh(use_discriminant = true)]
+pub enum BlobType {
+    Checkpoint = 0,
+    DA = 1,
+}
+
+impl BlobType {
+    /// Used to convert Integer to Type mainly for parsing purpose
+    pub fn from_u32(value: u32) -> Option<Self> {
```
delbonis, line 65: Shouldn't be necessary here.

crates/tx-parser/src/filter.rs
```diff
@@ -65,15 +66,21 @@ fn parse_deposits(
 fn parse_inscription_checkpoints<'a>(
```
delbonis, line 66: This does more than just checkpoints now.

crates/state/src/da_blob.rs
```diff
@@ -115,3 +118,30 @@ impl BlobIntent {
...
+        });
+        Self(commitment)
+    }
+
+    pub fn verify_against_payload(&self, payload: &[InscriptionBlob]) -> bool {
+        let commitment = payload.iter().fold(Buf32::zero(), |acc, blob| {
+            hash::raw(&[acc.0, hash::raw(blob.data()).0].concat())
+        });
+        commitment == self.0
+    }
```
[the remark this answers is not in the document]
bewakes: We had a short discussion on whether to use merkle trees or not. Do we actually need to prove existence any entry? I couldn't think of cases when we would need to prove the entries.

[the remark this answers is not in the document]
delbonis: We don't.  They get tracked in the chain with the granularity of blobs so we should recognize them in terms of blobs.  This chained commitment scheme would break since we wouldn't know what we should be looking for in the L1 proofs.

crates/state/src/da_blob.rs
```diff
@@ -115,3 +118,30 @@ impl BlobIntent {
         }
     }
 }
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, Arbitrary, BorshDeserialize, BorshSerialize)]
+pub struct BlobCommitment(pub Buf32);
```
delbonis, line 123: Avoid pub here.

crates/btcio/src/writer/builder.rs
voidash: > instead it should be using the tag strings as resolved from the InscriptionBlob type.
I still think, `rollup_name` makes more sense than the different tag strings. We just have to include  the rollup name for the first envelope then it can be used for filtering. Later included envelopes can just include a integer based tag which is byte efficient than posting the tag string every time.

crates/tx-parser/src/filter_types.rs
voidash: This was a shim enum, which facilitated conversion to ProtocolOp. I have removed it

crates/btcio/src/writer/test_utils.rs
```diff
@@ -4,41 +4,41 @@ use bitcoin::{Address, Network};
...
-/// Returns `Arc` of `BroadcastDatabase` for testing
+/// Returns `Arc` of [`BroadcastDatabase`] for testing
 pub fn get_broadcast_db() -> Arc<impl BroadcastDatabase> {
     let (db, dbops) = get_rocksdb_tmp_instance().unwrap();
     let bcastdb = Arc::new(L1BroadcastDb::new(db, dbops));
     Arc::new(BroadcastDb::new(bcastdb))
 }
 
-/// Returns `Arc` of `L1BroadcastHandle` for testing
+/// Returns `Arc` of [`L1BroadcastHandle`] for testing
```
storopoli, line 41: Probably doclink all `Arc`s here as well.

storopoli (approved):
ACK 59c60e60fe5c6b032baa8e1e9a5777cc4cd02934

Just saw the parts that I am a CODEOWNER: CI and BTC.

crates/btcio/src/writer/builder.rs
delbonis: (discussed in DMs)

Ideally we should want to avoid special casing the parsing and having to treat all of the envelopes together.  Because then they're not logically distinct envelopes, they're a single data structure.  Also raises the question of there being envelopes in different inputs, do we key those on the field from a different envelope on a different input?

bin/strata-client/src/rpc_server.rs
```diff
@@ -636,44 +636,61 @@ impl StrataAdminApiServer for AdminServerImpl {
...
         Ok(())
     }
+}
+
+#[async_trait]
+impl StrataSequencerApiServer for SequencerServerImpl {
+    async fn submit_da_blobs(&self, blobs: Vec<HexBytes>) -> RpcResult<()> {
+        let blob_vec: Vec<EnvelopePayload> = blobs
+            .into_iter()
+            .map(|blob| EnvelopePayload::new(PayloadTypeTag::DA, blob.into_inner()))
```
sapinb, line 683: As we are adding a new RPC endpoint, could we generalize this to allow for other types of da tag items too ?

crates/tx-parser/src/envelope.rs
```diff
@@ -0,0 +1,153 @@
...
+/// # Errors
+///
+/// This function errors if it cannot extract the [`EnvelopePayload`] from the bitcoin instructions.
+pub fn parse_envelope(
+    index: u32,
+    instructions: &mut Peekable<Instructions<'_>>,
+    rollup_name: &str,
+) -> Result<EnvelopePayload, EnvelopeParseError> {
+    // Parse name
+    if index == 0 {
```
bewakes, line 61: Is this actually needed if we are not having nested envelopes?

crates/state/src/tx.rs
```diff
@@ -14,7 +14,9 @@ pub enum ProtocolOperation {
     DepositRequest(DepositRequestInfo),
     /// Checkpoint data
     Checkpoint(SignedBatchCheckpoint),
-    // TODO: add other kinds like Proofs and statediffs
+    /// DA data. can be made through `submit_da_blob` RPC.
+    DA(Vec<u8>),
```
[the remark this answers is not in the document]
voidash: Discussed with @prajwolrg about just adding in the commitment for proofs. He said to leave as is.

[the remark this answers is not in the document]
delbonis: There must have been a miscommunication somewhere.  We have the full data in the L1 proofs since it's part of the block, the chain just needs to observe *that* the DA has been published (based on the commitment as a reference), it doesn't need to inspect it.

The actual DA payloads would be stored separately so we can fetch them directly more efficiently when it comes time to do state reconstruction.  It doesn't make sense to store them like this.

delbonis (changes requested):

crates/db/src/traits.rs
```diff
@@ -245,27 +245,28 @@ pub trait CheckpointDatabase {
...
+    fn get_entry_by_id(&self, id: Buf32) -> DbResult<Option<CommitRevealEntry>>;
 
     /// Get the blob ID corresponding to the index
-    fn get_blob_id(&self, blobidx: u64) -> DbResult<Option<Buf32>>;
+    fn get_id(&self, entryidx: u64) -> DbResult<Option<Buf32>>;
 
     /// Get the last blob index
-    fn get_last_blob_idx(&self) -> DbResult<Option<u64>>;
+    fn get_last_idx(&self) -> DbResult<Option<u64>>;
 }
```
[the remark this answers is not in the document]
voidash: Renamed it to data bundle to match with other names

storopoli (approved):
ACK 12d1ad72ec39516f371037dbedcbd8513f22525a
The `btcio/traits` -> `btcio-types` refactor is nice.

This is just from the CI, bin, and btc CODEOWNERS

bin/datatool/src/args.rs
```diff
@@ -172,6 +185,15 @@ pub(crate) struct SubcParams {
         description = "permit blank proofs after timeout in millis (default strict)"
     )]
     pub(crate) proof_timeout: Option<u32>,
+
+    #[argh(option, description = "fixed fee for writer transaction")]
+    pub(crate) fixed_fee: Option<u64>,
+
```
bewakes, line 191: I guess this is more descriptive.
```suggestion
    #[argh(option, description = "Fee in sats for reveal transactions. Will be determined dynamically if None")]
    pub(crate) reveal_fee: Option<u64>,
```

bewakes (changes requested):
Looks good overall, the use of visitor pattern seems a bit complicated though.

Need to add unit tests for parsing multiple protocol ops from a single transaction.

bin/datatool/src/args.rs
```diff
@@ -172,6 +185,18 @@ pub(crate) struct SubcParams {
         description = "permit blank proofs after timeout in millis (default strict)"
     )]
     pub(crate) proof_timeout: Option<u32>,
+
+    #[argh(
+        option,
+        description = "fee in sats for reveal transactions. Will be determined dynamically if None"
+    )]
+    pub(crate) fixed_fee: Option<u64>,
```
bewakes, line 193: Also, could you also rename this attribute to something like `reveal_fee`? Right now it's not clear what fee it is referring to.

crates/transaction/reveal/src/visitor.rs
```diff
@@ -0,0 +1,66 @@
...
+
+pub trait EnvelopeVisitor<'t> {
+    type Type;
+    fn visit_envelope_payload_chunks(chunks: impl Iterator<Item = &'t [u8]>) -> Self::Type;
+}
+// CheckpointVisitor
+pub trait CheckpointVisitor<'t>: EnvelopeVisitor<'t> {
+    fn visit_checkpoint(&self) -> Option<SignedBatchCheckpoint>;
+}
+
```
delbonis: This isn't how visitors work.  They update their internal state bookkeeping (meaning the `visit_` fns have to take `&mut self`).  This `visit_checkpoint` fn doesn't make sense since it isn't taking an argument of what's being visited.

crates/transaction/reveal/src/visitor.rs
```diff
@@ -0,0 +1,66 @@
...
+#[derive(Debug, Clone)]
+pub struct CommitmentVisitor {
+    commitment: PayloadCommitment,
```
delbonis: Would be a list of payload commitments.

crates/transaction/reveal/src/visitor.rs
```diff
@@ -0,0 +1,66 @@
...
+
+impl CommitmentVisitor {
+    pub fn commitment(&self) -> PayloadCommitment {
+        self.commitment
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct ExtractorVisitor<'a> {
+    chunks: Vec<&'a [u8]>,
```
delbonis: Would be a `Vec<Vec<u8>>` or similar, as we discussed in DMs.  We coalese the chunks into a single entry since they're semantically contiguous.

delbonis (changes requested):
What we discussed in DMs was a sensible approach.  This is very different.  What was the reasoning for doing this?  There's a few issues with it.  It duplicates the high level envelope parsing logic and making 2 passes for checkpoints and DA.  If we were to add another type of payload (like a forced inclusion) this would mean another instance of it.

We *always* want to take and parse the full contents of checkpoint payloads, so using the visitor-based approach doesn't help us much, it just complicates the code.  The goal of using the visitors for DA was to ideally avoid copies depending on the context.

delbonis:
Closing in favor of plan for @bewakes to take over and consolidate into another streamlined PR.

PR #519, opened 2024-12-04
STR-723: Structured Logging

voidash:
## Description

- sets a logger for each test run with stream and file handles 
- replaces print statements on tests with logger debug statements (level can be set through `LOG_LEVEL` env
- logs are stored on `_dd/test_dir/logs/{test_file_name}.log`

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

STR-723


Rajil1213:
Might be a good idea to set a default log level instead of `NOTSET` when the env variable is not set. The logs in the CI are _very_ noisy. If there's a way to set a different log level for dependencies (the way we do in Rust), then that works too.

If we're not setting a default log level, then we should set a proper log level in the CI workflow.

functional-tests/utils.py
```diff
@@ -432,3 +415,55 @@ def get_bridge_pubkey_from_cfg(cfg_params) -> str:
...
+    log_dir = os.path.join(datadir_root, "logs")
+    os.makedirs(log_dir, exist_ok=True)
+
+    # Common formatter
+    formatter = logging.Formatter(
+        "%(asctime)s - %(levelname)s - %(filename)s:%(lineno)d - %(message)s"
+    )
+    # Set up individual loggers for each test
+    for test_name in tests:
+        logger = logging.getLogger(test_name)
```
delbonis, line 453: This is essentially doing mutable global state, which you usually want to avoid.

More of this logic should be integrated into the runtime, which we can do now, see toplevel review comment.

functional-tests/utils.py
```diff
@@ -49,9 +49,9 @@ def generate_n_blocks(bitcoin_rpc: BitcoindClient, n: int):
     print(f"generating {n} blocks to address", addr)
     try:
         blk = bitcoin_rpc.proxy.generatetoaddress(n, addr)
-        print("made blocks", blk)
+        print(f"made blocks {blk}")
     except Exception as ex:
-        log.warning(f"{ex} while generating address")
+        logging.warning(f"{ex} while generating address")
```
delbonis, line 54: Does this direct call via the module end up getting captured by the test-level logger?

delbonis (changes requested):
I made [a few changes](https://codeberg.org/treyd/flexitest/commit/80655dad026723394705e3d1cfb28bb5090ff408) to flexitest that would help with this.

Since this 

```py
        self.logger = logging.getLogger(Path(__file__).stem)
```

pattern is duplicated across all instances of the tests, what we should do instead is to

* create a parent class `StrataTest` that all the tests inherit from, which takes the logger from the context in the new `premain` function
* create a child of `TestRuntime` that constructs the logger for the test itself, setting it on the context, instead of relying on the test to do it and letting us configure the logger more nicely

We could also add convenience fns on the new test type like `.info`, `.debug`, etc as wrappers instead of doing `.logger.info`.

This wasn't in scope of the original ticket, but if it's feasible it'd also be nice to have another set of loggers for each env in its `EnvContext` too, so that we can collect that as well.

functional-tests/fn_btcio_broadcast.py
```diff
@@ -28,27 +32,23 @@ def main(self, ctx: flexitest.RunContext):
...
+
+        wait_until_with_value(
+            lambda: btcrpc.gettransaction(txid),
+            predicate=lambda v: v is not None,
+            error_with="Tx was not published",
+            timeout=10,
+        )
+
+        # Also check from strata rpc
+        wait_until_with_value(
```
bewakes, line 48: can use `wait_until` if the return value is not used.
```python
wait_until(
    lambda: seqrpc.strata_getTxStatus(txid)["status"] in ("Confirmed", "Finalized"),
    error_with ...
    ...
)
```

functional-tests/entry.py
```diff
@@ -21,22 +21,23 @@ class BasicLiveEnv(flexitest.LiveEnv):
...
-    def __init__(self, srvs, bridge_pk):
+    def __init__(self, srvs, path, bridge_pk):
         super().__init__(srvs)
         self._el_address_gen = (
             f"deada00{x:04X}dca3ebeefdeadf001900dca3ebeef" for x in range(16**4)
         )
         self._ext_btc_addr_idx = 0
         self._rec_btc_addr_idx = 0
         self._bridge_pk = bridge_pk
+        self.path = path
```
bewakes, line 32: This does not seem to be used.

functional-tests/entry.py
```diff
@@ -45,7 +46,7 @@ def gen_ext_btc_address(self) -> str | List[str]:
         self._ext_btc_addr_idx += 1
         return tr_addr
 
-    def gen_rec_btc_address(self) -> str | List[str]:
+    def gen_rec_btc_address(self) -> str | list[str]:
```
storopoli, line 49: If I am not mistaken these `ruff` fixes were already done in #511. You should rebase.

functional-tests/fn_el_bridge_precompile.py
```diff
@@ -23,9 +25,10 @@
 class ElBridgePrecompileTest(flexitest.Test):
     def __init__(self, ctx: flexitest.InitContext):
         ctx.set_env("basic")
+        self.logger = logging.getLogger(Path(__file__).stem)
 
     def main(self, ctx: flexitest.RunContext):
-        print("SKIPPING TEST fn_el_bridge_precompile")
+        self.logger.debug("SKIPPING TEST fn_el_bridge_precompile")
```
storopoli, line 31: This should be an `info` no?
```suggestion
        self.logger.info("SKIPPING TEST fn_el_bridge_precompile")
```

storopoli (changes requested):
Thanks for going over the func-tests and improving them! ❤️ 
Some minor NITs.

functional-tests/entry.py
delbonis: It's possible some were missed.

functional-tests/fn_el_bridge_precompile.py
delbonis: Actually it should be a warning, just so it jumps out at us.  Although this whole test can probably be removed now since we have a end-to-end withdrawal test, right?

storopoli: Yes. Happy path only since that's what currently supported in devnet.

functional-tests/utils.py
voidash: Since this is root level logger, i don't think it's captured by test-level logger.

voidash:
Pulled instead of rebasing. Will revert the changes again

functional-tests/setup.py
```diff
@@ -0,0 +1,37 @@
+import flexitest
```
storopoli, line 1: Oh god another file?
We already have `entry.py`, `factory.py`, `constants.py`, `utils.py`, `net_settings.py`, etc.

Can't we merge this into `entry.py`?

Also, `setup.py` is very ambiguous. Maybe a better name if we really need a new file.

storopoli (changes requested):

functional-tests/setup.py
voidash: Merged into `entry.py`. These new classes will act as parent class for each test so i thought this is a setup for the tests. But then again alot of setup stuff is in entry.py

functional-tests/entry.py
```diff
@@ -16,12 +16,47 @@
...
+class RunContext(flexitest.RunContext):
+    """
+    Custom run context which provides access to services and some test specific variables.
+    To be used by ExtendedTestRuntime
+    """
+
+    def __init__(self, datadir_root: str, name: str, env: flexitest.LiveEnv):
+        self.name = name
+        self.datadir_root = datadir_root
+        super().__init__(env)
```
delbonis, line 51: Better names to avoid conflicts and be more consistent would be:

* `StrataTest`
* `StrataTestRuntime`
* `StrataRunContext`

Also maybe should put this into a new `testenv` module to keep "libraries" separate from the setup code?

Generally doesn't `super().__init__(env)` go on the first line of the constructor?

functional-tests/entry.py
```diff
@@ -16,12 +16,47 @@
...
+
+
 class BasicLiveEnv(flexitest.LiveEnv):
```
delbonis, line 54: While we're making a new module for this more general code it'd be nice to move these env classes over there too to clean things up.

functional-tests/fn_bridge_deposit_happy.py
```diff
@@ -5,11 +5,12 @@
...
 
 from constants import DEFAULT_ROLLUP_PARAMS
-from utils import get_bridge_pubkey, get_logger
+from entry import TestStrata
+from utils import get_bridge_pubkey
 
 
 @flexitest.register
-class BridgeDepositHappyTest(flexitest.Test):
+class BridgeDepositHappyTest(TestStrata):
```
delbonis, line 13: It'd be fine to just directly say `testenv.StrataTest` here (and in all the other places) to avoid cluttering the imports.

functional-tests/utils.py
```diff
@@ -432,3 +415,55 @@ def get_bridge_pubkey_from_cfg(cfg_params) -> str:
...
+
+    # Common formatter
+    formatter = logging.Formatter(
+        "%(asctime)s - %(levelname)s - %(filename)s:%(lineno)d - %(message)s"
+    )
+    # Set up individual loggers for each test
+    logger = logging.getLogger(test_name)
+    # Avoid duplicate handlers
+    if logger.hasHandlers():
+        return logger
```
delbonis, line 456: Is this still needed?

delbonis (changes requested):
I wonder if there's a *good* solution to make the logs in the utils crate end up in the per-test log file (ie. not just setting some global logger to the test logger when we enter and not passing it down in args everywhere).

functional-tests/fn_bridge_deposit_happy.py
voidash: StrataTest gets the file get picked up as test file, so i had to resort to `StrataTester`

delbonis: That's annoying, I'll make an adjustment to the autoconfiguration later.

delbonis (approved):

functional-tests/fn_bridge_deposit_happy.py
delbonis: `StrataTestBase` maybe would have been more appropriate but that's fine I'll just fix it later.

storopoli (approved):
ACK d5bdafee19b81b06398070709569eaeef747480a

PR #524, opened 2024-12-06
STR-687: improve NUMS unspendable public key

AaronFeickert:
## Description

The codebase currently uses a nonstandard instance of a [BIP-341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki) method for producing unspendable public keys. It does so by applying an arbitrary offset to a known unspendable key. This offset is not verifiably random, and provides no security benefit. It is therefore unnecessary.

This PR replaces all uses of this method. Instead, we hash a fixed string and use it to produce a curve group generator. Note that this is _not_ guaranteed to be a uniformly-sampled curve point! However, this is not relevant here, as the only security requirement is that the result not have an efficiently-computable discrete logarithm relation against the standard generator.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

There are multiple uses of this construction. All should be carefully reviewed to ensure that corresponding changes are done consistently.

## Checklist

## Related Issues

None.


delbonis (approved):

delbonis:
```

error: unresolved link to `UNSPENDABLE_INTERNAL_KEY`
  --> crates/bridge-tx-builder/src/operations.rs:70:34
   |
70 |     /// internal key being the [`static@UNSPENDABLE_INTERNAL_KEY`].
   |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no item named `UNSPENDABLE_INTERNAL_KEY` in scope
   |
   = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
   = note: `-D rustdoc::broken-intra-doc-links` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(rustdoc::broken_intra_doc_links)]`
```

That's annoying.

AaronFeickert:
@delbonis: Fixed.

crates/primitives/src/constants.rs
```diff
@@ -14,3 +19,26 @@ pub const SEC_NONCE_SIZE: usize = 64;
...
 /// The size (in bytes) of a Hash (such as [`Txid`](bitcoin::Txid)).
 pub const HASH_SIZE: usize = 32;
+
+/// A verifiably unspendable public key, produced by hashing a fixed string to a curve group
+/// generator.
+///
+/// Note that this is _not_ necessarily a uniformly-sampled curve point!
+///
+/// But this is fine; we only need a generator with no efficiently-computable discrete logarithm
+/// relation against the standard generator.
```
storopoli, line 29: We should mention BIP 341 and link to it here.

crates/primitives/src/constants.rs
```diff
@@ -14,3 +19,26 @@ pub const SEC_NONCE_SIZE: usize = 64;
...
+pub static UNSPENDABLE_PUBLIC_KEY: LazyLock<XOnlyPublicKey> = LazyLock::new(|| {
+    XOnlyPublicKey::from_slice(sha256::Hash::hash(b"Strata unspendable").as_byte_array())
```
storopoli, line 31: Maybe make the `"Strata unspendable"` also a constant?

What if we want to use it somewhere in the future?

storopoli (changes requested):
Very minor nits. Please feel free to push back if you feel so.

crates/primitives/src/constants.rs
AaronFeickert: Fair enough. I removed the link since we aren't strictly using the offset technique, and the idea of hashing to a generator isn't unique to that BIP. But it could still be useful for a reader or reviewer, so I'll add it back.

crates/primitives/src/constants.rs
AaronFeickert: I agree that this would be better, to make the use of magic values as obvious and clear as possible. Of course, in this case we probably want to _avoid_ reuse elsewhere, but same idea.

storopoli (approved):
ACK 9a9d52610290d5743e03271f9566579eef75168b

storopoli (approved):
ACK 811fbdf67ebcbb0c945dc3162655f378d8b697f2

delbonis (approved):

PR #525, opened 2024-12-09
fix: fullnode sync restart

purusang:
- l2 block finalization is done only if the l2 block actually exists.
- check point finalization is triggered only once in events L1Block or NewTipBlock.

### Reason of failing to restart
When fullnode is in initial syncing state, l1 reader/processor reads/processes blocks way faster than l2 block reader/processor. As a result l2blocks that did not exist in l2 db were marked as finalized. 

After restart, fullnode tries to perform state transitions from the point where l2 block was last finalized. If l2 block does not exist it crashes.

### Solution
In this pr I have changed the behavior of how we actually mark the finalize block. Previously, we were updating finalized block based on checkpoints processed by reading l1 blocks only. 
After this pr, we do as previously only if the l2 block actually exist in the l2 db. And for each new l2 block tip we look for any corresponding checkpoint and finalize it if the l2blockid in checkpoint matches.

### No effect on Sequencer
Sequencer does not get in this situation so it's work should not be affected.

### No effect on fully synced nodes
If fullnode is in sync with realtime chain state then no way it receives any l1 block which will have future l2 checkpoint.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

I want to add it's test case in a separate pr as creating this rare condition is quite tricky in test cases. Any suggestions on test cases for this condition will be highly appreciated.
I have tested these changes against our devnet releases/0.1.0 and the fix must be applied asap to releases.
## Checklist

## Related Issues
STR-647

### DIFF
--- a/crates/consensus-logic/src/csm/client_transition.rs
+++ b/crates/consensus-logic/src/csm/client_transition.rs
@@ -234,6 +234,27 @@ pub fn process_event<D: Database>(
     Ok(ClientUpdateOutput::new(writes, actions))
 }
 
+/// Handles the maturation of L1 height by finalizing checkpoints and emitting
+/// sync actions.
+///
+/// This function checks if there are any verified checkpoints at or before the
+/// given `maturable_height`. If such checkpoints exist, it attempts to
+/// finalize them by checking if the corresponding L2 block is available in the
+/// L2 database. If the L2 block is found, it marks the checkpoint as finalized
+/// and emits a sync action to finalize the L2 block. If the L2 block is not
+/// found, it logs a warning and skips the finalization.
+///
+/// # Arguments
+///
+/// * `maturable_height` - The height at which L1 blocks are considered mature.
+/// * `state` - A reference to the current client state.
+/// * `database` - A reference to the database interface.
+///
+/// # Returns
+///
+/// A tuple containing:
+/// * A vector of [`ClientStateWrite`] representing the state changes to be written.
+/// * A vector of [`SyncAction`] representing the actions to be synchronized.
 fn handle_mature_l1_height(
     maturable_height: u64,
     state: &ClientState,
@@ -285,6 +306,27 @@ fn handle_mature_l1_height(
     (writes, actions)
 }
 
+/// Handles the finalization of a checkpoint by processing the corresponding L2
+/// block ID.
+///
+/// This function checks if the given L2 block ID corresponds to a verified
+/// checkpoint. If it does, it calculates the maturable height based on the
+/// rollup parameters and the current state. If the L1 height associated with
+/// the L2 block ID is less than the maturable height, it calls
+/// [`handle_mature_l1_height`] and returns writes and sync actions.
+///
+/// # Arguments
+///
+/// * `state` - A reference to the current client state.
+/// * `blkid` - A reference to the L2 block ID to be finalized.
+/// * `params` - A reference to the rollup parameters.
+/// * `database` - A reference to the database interface.
+///
+/// # Returns
+///
+/// A tuple containing:
+/// * A vector of [`ClientStateWrite`] representing the state changes to be written.
+/// * A vector of [`SyncAction`] representing the actions to be synchronized.
 fn handle_checkpoint_finalization(
     state: &ClientState,
     blkid: &L2BlockId,
@@ -319,7 +361,7 @@ fn handle_checkpoint_finalization(
     (writes, actions)
 }
 
-/// Searches for a given `L2BlockId` within a slice of `L1Checkpoint` structs
+/// Searches for a given [`L2BlockId`] within a slice of [`L1Checkpoint`] structs
 /// and returns the height of the corresponding L1 block if found.
 fn find_l1_height_for_l2_blockid(
     checkpoints: &[L1Checkpoint],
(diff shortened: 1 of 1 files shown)


crates/consensus-logic/src/csm/client_transition.rs
```diff
@@ -261,6 +285,52 @@ fn handle_maturable_height(
...
+        }
+        None => {
+            debug!(%blkid, "L2 block not found in verified checkpoints, possibly not a last block in the checkpoint.");
+        }
+    }
+    (writes, actions)
+}
+
+/// Searches for a given `L2BlockId` within a slice of `L1Checkpoint` structs
+/// and returns the height of the corresponding L1 block if found.
```
storopoli, line 323:
```suggestion
/// Searches for a given [`L2BlockId`] within a slice of [`L1Checkpoint`] structs
/// and returns the height of the corresponding L1 block if found.
```

crates/consensus-logic/src/csm/client_transition.rs
```diff
@@ -261,6 +285,52 @@ fn handle_maturable_height(
...
+    checkpoints: &[L1Checkpoint],
+    target_l2_blockid: &L2BlockId,
+) -> Option<u64> {
+    checkpoints
+        .binary_search_by(|checkpoint| checkpoint.batch_info.l2_blockid.cmp(target_l2_blockid))
+        .ok()
+        .map(|index| checkpoints[index].height)
+}
+
 /// Filters a list of `BatchCheckpoint`s, returning only those that form a valid sequence
```
storopoli, line 334:
```suggestion
/// Filters a list of [`BatchCheckpoint`]s, returning only those that form a valid sequence
```

crates/consensus-logic/src/csm/client_transition.rs
```diff
@@ -261,6 +285,52 @@ fn handle_maturable_height(
     (writes, actions)
 }
 
+fn handle_checkpoint_finalization(
```
storopoli, line 288: Can you add docstrings here?

crates/consensus-logic/src/csm/client_transition.rs
```diff
@@ -224,15 +224,20 @@ pub fn process_event<D: Database>(
...
+            writes.extend(wrs);
+            actions.extend(acts);
         }
     }
 
     Ok(ClientUpdateOutput::new(writes, actions))
 }
 
-fn handle_maturable_height(
+fn handle_mature_l1_height(
```
storopoli, line 237: Add docstrings here?

storopoli (changes requested):
Some nits on docstrings.
The logic looks fine, but I am not a codeowner in this part of the codebase.

delbonis (dismissed):
We'll be coming back around to working on some of this later.

crates/consensus-logic/src/csm/client_transition.rs
```diff
@@ -224,15 +224,41 @@ pub fn process_event<D: Database>(
...
+///
+/// # Arguments
+///
+/// * `maturable_height` - The height at which L1 blocks are considered mature.
+/// * `state` - A reference to the current client state.
+/// * `database` - A reference to the database interface.
+///
+/// # Returns
+///
+/// A tuple containing:
```
storopoli, line 255:
```suggestion
/// A tuple containing:
///
```

crates/consensus-logic/src/csm/client_transition.rs
```diff
@@ -261,7 +306,74 @@ fn handle_maturable_height(
...
+/// # Arguments
+///
+/// * `state` - A reference to the current client state.
+/// * `blkid` - A reference to the L2 block ID to be finalized.
+/// * `params` - A reference to the rollup parameters.
+/// * `database` - A reference to the database interface.
+///
+/// # Returns
+///
+/// A tuple containing:
```
storopoli, line 327:
```suggestion
/// A tuple containing:
///
```

crates/consensus-logic/src/csm/client_transition.rs
```diff
@@ -261,7 +306,74 @@ fn handle_maturable_height(
...
+            }
+        }
+        None => {
+            debug!(%blkid, "L2 block not found in verified checkpoints, possibly not a last block in the checkpoint.");
+        }
+    }
+    (writes, actions)
+}
+
+/// Searches for a given [`L2BlockId`] within a slice of [`L1Checkpoint`] structs
```
storopoli, line 364:
```suggestion
/// Searches for a given [`L2BlockId`] within an array of [`L1Checkpoint`]
```

crates/consensus-logic/src/csm/client_transition.rs
```diff
@@ -270,12 +382,13 @@ fn handle_maturable_height(
 /// # Arguments
 ///
 /// * `state` - The client's current state, which provides the L1 view and pending checkpoints.
-/// * `checkpoints` - A slice of `BatchCheckpoint`s to be filtered.
+/// * `checkpoints` - A slice of [`BatchCheckpoint`]s to be filtered.
```
storopoli, line 385:
```suggestion
/// * `checkpoints` - An array of [`BatchCheckpoint`]s to be filtered.
```

storopoli (changes requested):
Amazing docstrings.
Just some suggestions.
I still think "slice" is less intuitive than "array". But feel free to disagree and go with "slice".

crates/consensus-logic/src/csm/client_transition.rs
purusang: Yeah for reading arrays are more intuitive but these are slices so keeping as it it.

storopoli (approved):
ACK c58fac0218e7536834be6235e57fac5e620d12e8

delbonis (approved):

PR #527, opened 2024-12-09
Str 294 update l1 manifest

bewakes:
## Description

This PR adds the epoch information to `L1BlockManifest` and rescans recent blocks if the scan rules have changed in the new epoch.
This also does some minor refactoring in `btcio::query.rs` to become more "testing" friendly.

### Idea
If there is a new epoch and filtering rules change on that epoch, that means recent blocks needs to be reparsed with the new rule.

But what defines recent blocks?

An epoch change is only identified when the L1 block containing the epoch is finalized. So, between the time the epoch block appears on  L1(Say at L1 height `h`) and the time it is buried and identified in chain, we would have scanned L1 blocks using the previous epoch's rules. This means, we would have scanned blocks from `h` to current block(at the time of updating new epoch), using the old rules. Those are the "recent blocks".

So, once the rule changes, a `L1Event` is emitted to revert to height `h` and reader's state is also updated accordingly.
### Assumption
The `recent_blocks` field in `ReaderState` holds enough blocks to be able to revert to the point where checkpoint appeared.

## Type of change
- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

We still have to derive the filter rules from chainstate which will be the part of another Ticket.

## Checklist

## Related Issues
STR-716

### DIFF
--- a/crates/btcio/src/reader/query.rs
+++ b/crates/btcio/src/reader/query.rs
@@ -5,8 +5,8 @@ use std::{
 };
 
 use anyhow::bail;
-use bitcoin::{hashes::Hash, BlockHash};
-use strata_primitives::buf::Buf32;
+use bitcoin::{hashes::Hash, Block, BlockHash};
+use strata_primitives::{buf::Buf32, l1::Epoch};
 use strata_state::l1::{
     get_btc_params, get_difficulty_adjustment_height, BtcParams, HeaderVerificationState,
     L1BlockId, TimestampStore,
@@ -21,7 +21,10 @@ use tokio::sync::mpsc;
 use tracing::*;
 
 use crate::{
-    reader::{config::ReaderConfig, state::ReaderState},
+    reader::{
+        config::ReaderConfig,
+        state::{EpochFilterConfig, ReaderState},
+    },
     rpc::traits::Reader,
     status::{apply_status_updates, L1StatusUpdate},
 };
@@ -73,6 +76,8 @@ async fn do_reader_task<R: Reader>(
         let cur_best_height = state.best_block_idx();
         let poll_span = debug_span!("l1poll", %cur_best_height);
 
+        check_and_update_epoch_scan_rules_change(&ctx, &mut state).await?;
+
         if let Err(err) = poll_for_new_blocks(&ctx, &mut state, &mut status_updates)
             .instrument(poll_span)
             .await
@@ -103,6 +108,68 @@ async fn do_reader_task<R: Reader>(
 
         apply_status_updates(&status_updates, &ctx.status_channel).await;
     }
+}
+
+/// Checks for epoch and scan rule update. If so, rescan recent blocks to account for possibly
+/// missed relevant txs.
+async fn check_and_update_epoch_scan_rules_change<R: Reader>(
+    ctx: &ReaderContext<R>,
+    state: &mut ReaderState,
+) -> anyhow::Result<()> {
+    let latest_epoch = ctx.status_channel.epoch().unwrap_or(0);
+    // TODO: check if latest_epoch < current epoch. should panic if so?
+    let curr_epoch = match state.epoch() {
+        Epoch::Exact(e) => *e,
+        Epoch::AtTransition(_from, to) => *to,
+    };
+    let is_new_epoch = curr_epoch == latest_epoch;
+    let curr_filter_config = match state.filter_config() {
+        EpochFilterConfig::AtEpoch(c) => c.clone(),
+        EpochFilterConfig::AtTransition(_, c) => c.clone(),
+    };
+    if is_new_epoch {
+        // TODO: pass in chainstate to `derive_from`
+        let new_config = TxFilterConfig::derive_from(ctx.config.params.rollup())?;
+
+        // If filter rule has changed, first revert recent scanned blocks and rescan them with new
+        // rule
+        if new_config != curr_filter_config {
+            // Send L1 revert so that the recent txs can be appropriately re-filtered
+            let revert_ev =
+                L1Event::RevertTo(state.best_block_idx() - state.recent_blocks().len() as u64);
+            if ctx.event_tx.send(revert_ev).await.is_err() {
+                warn!("unable to submit L1 reorg event, did persistence task exit?");
+            }
+
+            state.set_filter_config(EpochFilterConfig::AtTransition(
+                curr_filter_config,
+                new_config,
+            ));
+            state.set_epoch(Epoch::AtTransition(curr_epoch, latest_epoch));
+
+            rescan_recent_blocks(ctx, state).await?;
+        } else {
+            state.set_filter_config(EpochFilterConfig::AtEpoch(curr_filter_config));
+            state.set_epoch(Epoch::Exact(curr_epoch));
+        }
+    }
+    // TODO: anything else?
+    Ok(())
+}
+
+async fn rescan_recent_blocks<R: Reader>(
+    ctx: &ReaderContext<R>,
+    state: &mut ReaderState,
+) -> anyhow::Result<()> {
+    let recent_len = state.recent_blocks().len() as u64;
+    let mut status_updates: Vec<L1StatusUpdate> = Vec::new(); // TODO: is this necessary
+    let recent = state.recent_blocks().to_vec();
+    for (i, blk_hash) in recent.iter().enumerate() {
+        let height = state.best_block_idx() - recent_len + i as u64;
+        let block = ctx.client.get_block(blk_hash).await?;
+        process_block(ctx, state, &mut status_updates, height, block).await?;
+    }
+    Ok(())
 }
 
 /// Inits the reader state by trying to backfill blocks up to a target height.
@@ -140,7 +207,16 @@ async fn init_reader_state<R: Reader>(
         real_cur_height = height;
     }
 
-    let state = ReaderState::new(real_cur_height + 1, lookback, init_queue);
+    let params = ctx.config.params.clone();
+    let filter_config = TxFilterConfig::derive_from(params.rollup())?;
+    let epoch = ctx.status_channel.epoch().unwrap_or(0);
+    let state = ReaderState::new(
+        real_cur_height + 1,
+        lookback,
+        init_queue,
+        EpochFilterConfig::AtEpoch(filter_config), // TODO: decide if it is transitional or exact
+        Epoch::Exact(epoch),                       // TODO: decide if it is transitional or exact
+    );
     Ok(state)
 }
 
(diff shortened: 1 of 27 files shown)


crates/btcio/src/reader/query.rs
```diff
@@ -105,6 +110,68 @@ async fn do_reader_task<R: Reader>(
...
+        EpochFilterConfig::AtTransition(_, c) => c.clone(),
+    };
+    if is_new_epoch {
+        // TODO: pass in chainstate to `derive_from`
+        let new_config = TxFilterConfig::derive_from(ctx.config.params.rollup())?;
+
+        // If filter rule has changed, first revert recent scanned blocks and rescan them with new
+        // rule
+        if new_config != curr_filter_config {
+            // Send L1 revert so that the recent txs can be appropriately re-filtered
```
bewakes, line 137: I couldn't think of better way to re-scan recent blocks without sending the revert event.

crates/primitives/src/l1.rs
```diff
@@ -116,6 +116,46 @@ impl L1TxProof {
...
+    manifest: L1BlockManifest,
+    /// Epoch, whose scan rules are applied to this manifest
+    epoch: Epoch,
+}
+
+/// Epoch that is either exact or transitioning from one to another.
+#[derive(Clone, Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize, Arbitrary)]
+pub enum Epoch {
+    /// Exact epoch
+    Exact(u64),
```
delbonis, line 132: Probably makes more sense to say "within" instead of "exact"?

Should also have some accessor that takes away the nuance this is trying to add for when we just need a specific epoch number.  The name `Epoch` here seems too broad.  Really the epoch index should always be represented as a single number.

crates/primitives/src/l1.rs
```diff
@@ -116,6 +116,46 @@ impl L1TxProof {
     }
 }
 
+/// Includes `L1BlockManifest` along with scan rules that it is applied to
+#[derive(Clone, Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize, Arbitrary)]
+pub struct EpochedL1BlockManifest {
+    /// The actual l1 manifest
+    manifest: L1BlockManifest,
+    /// Epoch, whose scan rules are applied to this manifest
+    epoch: Epoch,
```
delbonis, line 125: Doesn't make sense to use the below type for this.  It's always referring to the rules that were derived as of the particular checkpoint, and the checkpoints correspond 1:1 with the epochs they are checkpointing.

crates/primitives/src/l1.rs
```diff
@@ -116,6 +116,46 @@ impl L1TxProof {
     }
 }
 
+/// Includes `L1BlockManifest` along with scan rules that it is applied to
+#[derive(Clone, Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize, Arbitrary)]
+pub struct EpochedL1BlockManifest {
```
delbonis, line 121: This is somewhat of a clumsy name.  Maybe `L1ManifestContainer`?  Or change it so *this* is just `L1BlockManifest` and the other is `L1BlockManifestData` or something?  Or take "block" out of those types since it's pretty implied?

crates/btcio/src/reader/query.rs
```diff
@@ -105,6 +110,68 @@ async fn do_reader_task<R: Reader>(
     }
 }
 
+/// Checks for epoch and scan rule update. If so, rescan recent blocks to account for possibly
+/// missed relevant txs.
+async fn check_and_update_epoch_scan_rules_change<R: Reader>(
+    ctx: &ReaderContext<R>,
+    state: &mut ReaderState,
+) -> anyhow::Result<()> {
+    let latest_epoch = ctx.status_channel.epoch().unwrap_or(0);
```
delbonis, line 119: `last_epoch` is more precise as "latest" implies it's what `curr_epoch` would be.

crates/btcio/src/reader/query.rs
```diff
@@ -229,11 +305,28 @@ async fn fetch_and_process_block<R: Reader>(
...
-    let filter_config = TxFilterConfig::derive_from(params.rollup())?;
-    let filtered_txs = filter_protocol_op_tx_refs(&block, filter_config);
+    let filtered_txs = match state.filter_config() {
+        EpochFilterConfig::AtEpoch(config) => filter_protocol_op_tx_refs(&block, config),
+        EpochFilterConfig::AtTransition(from_config, to_config) => {
+            let mut txs = filter_protocol_op_tx_refs(&block, from_config);
+            let txs1 = filter_protocol_op_tx_refs(&block, to_config);
+            txs.extend_from_slice(&txs1);
+            txs
+        }
```
delbonis, line 328: Why are we looking at both epochs here?  We only care about a single epoch at a time.

If the same tx matched under both sets of rules, then the entry would be present in the list twice, which doesn't make sense.

delbonis (changes requested):
Need more rationale for why this new design that tracks epoch transitions was needed.  It's rather complicated and the core scanning logic shouldn't need it, since we're only ever using one epoch's rules at a time.

bewakes:
@delbonis, scanning using only one epoch rule is definitely the ideal scenario. But as discussed somewhere in slack, it seems like having to scan blocks(at least the deposit transactions within) requires two(or more probably?) rules to adhere to. That's why I've used the transitions for epochs. I have a feeling that this might still not be sufficient, but I can't see how transitions can be avoided or simplified to use a single rule(at the very least for deposits).

crates/btcio/src/reader/query.rs
bewakes: Oh yes, this is kind of doesn't make sense right now. We can uniquely identify protocol ops, how I haven't thought of yet. This would be unnecessary if there were no transitions, but transitions seem inevitable don't they?

crates/primitives/src/l1.rs
bewakes: Yeah the name `Epoch` is too broad in this context. It's true that epoch should mean a single number, but in order to handle valid transactions that can fall under both old and new epoch, such a construct was used.

crates/primitives/src/l1.rs
bewakes: All boils down to the same point mentioned in other comments and the slack thread.

crates/primitives/src/l1.rs
bewakes: Agree. Thanks.

bewakes: Renamed: `L1BlockManifest -> L1BlockRecord` and `EpochedL1BlockManifest -> L1BlockManifest`, which I think is sensible enough, isn't it?

delbonis:
@bewakes 

> But as discussed somewhere in slack, it seems like having to scan blocks(at least the deposit transactions within) requires two(or more probably?) rules to adhere to. That's why I've used the transitions for epochs.

Can you DM me a link to this conversation?  This doesn't seem right.

crates/state/src/client_state.rs
```diff
@@ -115,6 +115,13 @@ impl ClientState {
     }
 }
 
+#[cfg(feature = "test_utils")]
+impl ClientState {
+    pub fn set_last_finalized_checkpoint(&mut self, chp: L1Checkpoint) {
+        self.local_l1_view.last_finalized_checkpoint = Some(chp);
+    }
```
delbonis, line 122: Given how this is used, shouldn't this be `#[cfg(test)]`?

delbonis (changes requested):

crates/state/src/client_state.rs
bewakes: The thing is, this cannot be done because this method is used in unit test of another crate, and using the feature flag is the only way. I had the same assumption and spent a whole lot of time trying to do it with `#[cfg(test)]`

delbonis: Oh that's unfortunate.

crates/btcio/src/reader/state.rs
```diff
@@ -13,24 +14,52 @@ pub struct ReaderState {
...
+
+    /// Current epoch
+    epoch: u64,
 }
 
 impl ReaderState {
     /// Constructs a new reader state instance using some context about how we
     /// want to manage it.
-    pub fn new(next_height: u64, max_depth: usize, recent_blocks: VecDeque<BlockHash>) -> Self {
+    // FIXME: the args list is getting long
```
storopoli, line 28: Can you remove this `FIXME`?
This will be handled by chippy's `too_many_arguments`: https://rust-lang.github.io/rust-clippy/master/index.html#too_many_arguments
The default is 7 so we are good so far.

crates/primitives/src/l1.rs
```diff
@@ -128,10 +128,41 @@ impl L1TxProof {
     }
 }
 
+/// Includes `L1BlockManifest` along with scan rules that it is applied to
```
storopoli, line 131:
```suggestion
/// Includes [`L1BlockManifest`] along with scan rules that it is applied to
```

crates/rpc/types/src/types.rs
```diff
@@ -91,7 +91,7 @@ impl RpcL1Status {
             last_rpc_error: l1s.last_rpc_error,
             cur_height: l1s.cur_height,
             cur_tip_blkid: l1s.cur_tip_blkid,
-            last_published_txid: l1s.last_published_txid,
+            last_published_txid: l1s.last_published_txid.map(Into::into),
```
storopoli, line 94: Does this work?

```suggestion
            last_published_txid: l1s.last_published_txid.into(),
```

storopoli (changes requested):
Cosmetic nits

crates/rpc/types/src/types.rs
bewakes: Oh yes, it should have worked, but interestingly it did not work.

storopoli (approved):
ACK f33a5ec99594df53cc75b632d377ba5782b75378

delbonis (changes requested):
Have some more thorough comments now that this is in a more current state.

crates/primitives/src/l1.rs
```diff
@@ -128,10 +128,41 @@ impl L1TxProof {
...
+        Self { record, epoch }
+    }
+
+    pub fn header(&self) -> &[u8] {
+        self.record.header()
+    }
+
+    pub fn block_hash(&self) -> Buf32 {
+        self.record.block_hash()
+    }
```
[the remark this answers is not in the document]
bewakes: Maybe, but I think its fine.

crates/btcio/src/reader/state.rs
```diff
@@ -13,24 +14,51 @@ pub struct ReaderState {
...
         }
     }
 
     pub fn next_height(&self) -> u64 {
         self.next_height
     }
 
+    pub fn recent_blocks(&self) -> &[BlockHash] {
+        // TODO: is this correct?
+        self.recent_blocks.as_slices().0
```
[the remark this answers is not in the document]
bewakes: I see, thanks. I've updated the interface to return an iterator instead.

bewakes:
@delbonis, those are great suggestions! I've updated them accordingly.

storopoli (approved):
ACK 4148f40105a35c4460e4a43cfd4ca09358d79188

delbonis (approved):
I think this is fine now.

MdTeach (approved):

manishbista28 (approved):

PR #531, opened 2024-12-10
fix(storage): weird caching behaviour with `Option<T>` & replace Mutex with `parking_lot`

Zk2u:
## Description

This PR addresses some weird caching behaviour I found while attempting a rewrite of the CacheTable's inner structure. TLDR is that instead of caching V, we cache Option<V>. That means if it returns None, it will keep returning None until that key is either purged or expired. I'm not sure if this could lead to other issues, but this seemed wrong to me. So this fixes that. 

Also, `parking_lot`'s `Mutex` is so much faster than tokio's or std's, especially considering microcontention cases like this. It has other advantages too, and is just a better Mutex all around. So I'm switching the CacheTable to use that too.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This is quite a small PR.

## Checklist

## Related Issues

- STR-704
- STR-705


crates/storage/src/cache.rs
```diff
@@ -20,17 +20,17 @@ pub enum SlotState<T> {
     Ready(T),
 
     /// A database fetch is happening in the background and it will be updated.
-    Pending(broadcast::Receiver<T>),
+    Pending(broadcast::Receiver<Option<T>>),
```
storopoli, line 23: I don't get it. Aren't we replacing the `Option<T>` in this PR?
Or are we caching a `T` and maybe we don't have it cached and we return `None`?

storopoli: Oh, makes sense.

delbonis:
> That means if it returns None, it will keep returning None until that key is either purged or expired. 

This is the correct behavior.

delbonis:
Right?  Is there something wrong with my reasoning?

delbonis:
There aren't stale `None`s though, since we (are supposed to) purge the existing cache entries when we write new values.  Adjusting the managers to make it writethrough in the meantime isn't a bad idea.

What I would like in some future version is a new type of cache that's context-aware and meant for storing things that act like `Option<V>` that separately tracks `None`s and doesn't count them against the LRU limit (maybe using a separate LRU limit that's higher instead?).

delbonis:
Also are you *sure* it makes sense to use parking_lot here?  I know we don't hold the lock for lock but this *could* result in a tokio runtime thread being blocked which would be worse.

PR #535, opened 2024-12-12
Additional rpcs

purusang:
## Description
These additional rpcs are aimed to be helpful for debugging purpose.
  - strata_getBlockById
  - strata_getBlocksAtIdx
  - strate_getChainstateAtIdx
  - strata_getClientStateAtIdx
  - strata_getLastBroadcastEntry
  - strata_getBroadcastEntryByIdx

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
Looking forward to get suggestions on fields to be exposed from L2Block::body and ChainState.

## Checklist

- [x] I have tested them locally. 

## Related Issues
STR-801

### DIFF
--- a/bin/strata-client/src/rpc_server.rs
+++ b/bin/strata-client/src/rpc_server.rs
@@ -143,10 +143,18 @@ impl<D: Database + Send + Sync + 'static> StrataApiServer for StrataRpcImpl<D> {
 
     async fn get_clientstate_at_idx(&self, idx: u64) -> RpcResult<Option<ClientState>> {
         let database = self.database.clone();
-        let client_state_db = database.client_state_db().as_ref();
-        let cs = reconstruct_state(client_state_db, idx)
-            .map_err(|e| Error::Db(strata_db::DbError::Other(e.to_string())))?;
-        Ok(Some(cs))
+        let cs = wait_blocking("clientstate_at_idx", move || {
+            let client_state_db = database.client_state_db();
+            match reconstruct_state(client_state_db.as_ref(), idx) {
+                Ok(client_state) => Ok(Some(client_state)),
+                Err(e) => {
+                    error!(%idx, %e, "failed to reconstruct client state");
+                    Err(Error::Other(e.to_string()))
+                }
+            }
+        })
+        .await?;
+        Ok(cs)
     }
 
     async fn protocol_version(&self) -> RpcResult<u64> {
--- a/bin/strata-client/src/rpc_server.rs
+++ b/bin/strata-client/src/rpc_server.rs
@@ -106,6 +106,75 @@ fn conv_blk_header_to_rpc(blk_header: &impl L2Header) -> RpcBlockHeader {
 
 #[async_trait]
 impl<D: Database + Send + Sync + 'static> StrataApiServer for StrataRpcImpl<D> {
+    async fn get_blocks_at_idx(&self, idx: u64) -> RpcResult<Vec<HexBytes32>> {
+        let db = self.database.clone();
+        let blk_ids = wait_blocking("l2_blockid", move || {
+            db.l2_db()
+                .get_blocks_at_height(idx)
+                .map_err(Error::Db)
+                .map(|ids| {
+                    ids.into_iter()
+                        .map(|id| HexBytes32(*id.as_ref()))
+                        .collect::<Vec<HexBytes32>>()
+                })
+        })
+        .await?;
+        Ok(blk_ids)
+    }
+
+    async fn get_block_by_id(&self, block_id: L2BlockId) -> RpcResult<Option<L2Block>> {
+        let db = self.database.clone();
+        let l2blk = fetch_l2blk::<D>(db.l2_db(), block_id)?;
+        Ok(Some(L2Block {
+            header: conv_blk_header_to_rpc(l2blk.header()),
+            body: BlockBody { todo: 0 },
+        }))
+    }
+    async fn block_number(&self) -> RpcResult<u64> {
+        let sync = {
+            let cl = self.status_rx.cl().borrow();
+            cl.sync().ok_or(Error::ClientNotStarted)?.clone()
+        };
+        Ok(sync.chain_tip_height())
+    }
+
+    async fn get_last_chainstate_idx(&self) -> RpcResult<u64> {
+        let db = self.database.clone();
+        let last_idx = wait_blocking("last_chain_state_idx", move || {
+            db.chain_state_db().get_last_state_idx().map_err(Error::Db)
+        })
+        .await?;
+        Ok(last_idx)
+    }
+
+    // async fn get_chainstate_at_idx(&self, idx: u64) -> RpcResult<Option<ChainState>> {
+    //     let db = self.database.clone();
+    //     let chain_state = wait_blocking("chain_state_at_idx", move || {
+    //         db.chain_state_db().get_state_at(idx).map_err(Error::Db)
+    //     })
+    //     .await?;
+    //     Ok(chain_state)
+    // }
+
+    async fn get_last_clientstate_idx(&self) -> RpcResult<u64> {
+        let db = self.database.clone();
+        let last_idx = wait_blocking("last_client_state_idx", move || {
+            db.client_state_db()
+                .get_last_client_state_idx()
+                .map_err(Error::Db)
+        })
+        .await?;
+        Ok(last_idx)
+    }
+
+    async fn get_clientstate_at_idx(&self, idx: u64) -> RpcResult<Option<ClientState>> {
+        let db = self.database.clone();
+        let client_state = wait_blocking("client_state_at_idx", move || {
+            db.client_state_db().get_state_at(idx).map_err(Error::Db)
+        })
+        .await?;
+        Ok(client_state)
+    }
     async fn protocol_version(&self) -> RpcResult<u64> {
         Ok(1)
     }
--- a/bin/strata-client/src/rpc_server.rs
+++ b/bin/strata-client/src/rpc_server.rs
@@ -25,7 +25,7 @@ use strata_primitives::{
     hash,
     params::Params,
 };
-use strata_rpc_api::{StrataAdminApiServer, StrataApiServer, StrataSequencerApiServer};
+use strata_rpc_api::{StrataAdminApiServer, StrataApiServer, StrataSequencerApiServer, StrataDebugApiServer};
 use strata_rpc_types::{
     errors::RpcServerError as Error, DaBlob, HexBytes, HexBytes32, L2BlockStatus, RpcBlockBody,
     RpcBlockHeader, RpcBridgeDuties, RpcChainState, RpcCheckpointInfo, RpcClientStatus,
@@ -34,7 +34,7 @@ use strata_rpc_types::{
 use strata_rpc_utils::to_jsonrpsee_error;
 use strata_state::{
     batch::BatchCheckpoint,
-    block::L2BlockBundle,
+    block::{L2BlockBundle, L2Block},
     bridge_duties::BridgeDuty,
     bridge_ops::WithdrawalIntent,
     client_state::ClientState,
@@ -120,40 +120,6 @@ impl<D: Database + Send + Sync + 'static> StrataApiServer for StrataRpcImpl<D> {
         Ok(block_ids)
     }
 
-    async fn get_block_by_id(&self, block_id: L2BlockId) -> RpcResult<Option<RpcL2Block>> {
-        let l2_block_manager = self.l2_block_manager.clone();
-        let l2_block = l2_block_manager
-            .get_block_async(&block_id)
-            .await
-            .map_err(Error::Db)?;
-
-        let header = l2_block
-            .as_ref()
-            .ok_or(Error::MissingL2Block(block_id))?
-            .header();
-        Ok(Some(RpcL2Block {
-            header: conv_blk_header_to_rpc(header),
-            body: RpcBlockBody { todo: 0 },
-        }))
-    }
-    async fn block_number(&self) -> RpcResult<u64> {
-        let sync = self
-            .status_channel
-            .sync_state()
-            .ok_or(Error::ClientNotStarted)?
-            .clone();
-        Ok(sync.chain_tip_height())
-    }
-
-    async fn get_last_chainstate_idx(&self) -> RpcResult<u64> {
-        let db = self.database.clone();
-        let last_idx = wait_blocking("last_chain_state_idx", move || {
-            db.chain_state_db().get_last_state_idx().map_err(Error::Db)
-        })
-        .await?;
-        Ok(last_idx)
-    }
-
     async fn get_chainstate_at_idx(&self, idx: u64) -> RpcResult<Option<RpcChainState>> {
         let db = self.database.clone();
         let chain_state = wait_blocking("chain_state_at_idx", move || {
@@ -162,23 +128,12 @@ impl<D: Database + Send + Sync + 'static> StrataApiServer for StrataRpcImpl<D> {
         .await?;
         match chain_state {
             Some(cs) => Ok(Some(RpcChainState {
-                last_block: cs.chain_tip_blockid(),
-                slot: cs.chain_tip_slot(),
-                epoch: cs.epoch(),
+                tip_blkid: cs.chain_tip_blockid(),
+                tip_slot: cs.chain_tip_slot(),
+                cur_epoch: cs.epoch(),
             })),
             None => Ok(None),
         }
-    }
-
-    async fn get_last_clientstate_idx(&self) -> RpcResult<u64> {
-        let db = self.database.clone();
-        let last_idx = wait_blocking("last_client_state_idx", move || {
-            db.client_state_db()
-                .get_last_client_state_idx()
-                .map_err(Error::Db)
-        })
-        .await?;
-        Ok(last_idx)
     }
 
     async fn get_clientstate_at_idx(&self, idx: u64) -> RpcResult<Option<ClientState>> {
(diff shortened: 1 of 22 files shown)


crates/rpc/types/src/types.rs
```diff
@@ -164,6 +163,17 @@ pub struct RpcBlockHeader {
     pub state_root: [u8; 32],
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct RpcBlockBody {
+    /// The fields of block body is to be decided.
+    pub todo: u64,
+}
```
delbonis, line 170: We don't want to directly expose the block bodies as json since that makes consumers dependent on the json representation which we don't want to directly expose.  If we do want to expose it for testing purposes, we should make it under an explicitly-unstable `debug_` namespace.

crates/rpc/types/src/types.rs
```diff
@@ -314,3 +324,14 @@ pub enum L2BlockStatus {
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RpcChainState {
+    /// Most recent seen block.
+    pub last_block: L2BlockId,
+
+    /// The slot of the last produced block.
+    pub slot: u64,
+
+    pub epoch: u64,
```
delbonis, line 336: `tip_blkid`, `tip_slot`, `cur_epoch`

crates/rpc/api/src/lib.rs
```diff
@@ -133,6 +162,14 @@ pub trait StrataAdminApi {
 #[cfg_attr(not(feature = "client"), rpc(server))]
 #[cfg_attr(feature = "client", rpc(server, client))]
 pub trait StrataSequencerApi {
+    /// Get the last broadcast entry
+    #[method(name = "strata_getLastBroadcastEntry")]
+    async fn get_last_broadcast_entry(&self) -> RpcResult<Option<L1TxEntry>>;
+
+    /// Get the broadcast entry by its id
+    #[method(name = "strata_getBroadcastEntryByIdx")]
+    async fn get_broadcast_entry_by_idx(&self, idx: HexBytes32) -> RpcResult<Option<L1TxEntry>>;
```
delbonis, line 171: Maybe should be on the rollup client namespace?  Since the sequencer won't be directly connected to the bitcoin client in the future.

crates/rpc/api/src/lib.rs
```diff
@@ -1,18 +1,47 @@
...
+    #[method(name = "blockNumber")]
+    async fn block_number(&self) -> RpcResult<u64>;
+
+    /// Get the last ChainState index
+    #[method(name = "getLastChainstateIdx")]
+    async fn get_last_chainstate_idx(&self) -> RpcResult<u64>;
+
+    /// Get the ChainState at a certain index
+    #[method(name = "getChainstateAtIdx")]
+    async fn get_chainstate_at_idx(&self, idx: u64) -> RpcResult<Option<RpcChainState>>;
```
delbonis, line 35: These should be queried by the slot index.

crates/rpc/api/src/lib.rs
```diff
@@ -1,18 +1,47 @@
...
+
+    /// Get the last ClientState index
+    #[method(name = "getLastClientStateIdx")]
+    async fn get_last_clientstate_idx(&self) -> RpcResult<u64>;
+
+    /// Get the ClientState at a certain index
+    #[method(name = "getClientStateAtIdx")]
+    async fn get_clientstate_at_idx(&self, idx: u64) -> RpcResult<Option<ClientState>>;
```
delbonis, line 43: The last client state index is *by definition* the index of the last sync event.  Should be queried based on that.

crates/rpc/api/src/lib.rs
```diff
@@ -1,18 +1,47 @@
...
+    #[method(name = "getBlockById")]
+    async fn get_block_by_id(&self, block_id: L2BlockId) -> RpcResult<Option<RpcL2Block>>;
+
+    /// Get blocks at a certain height
+    #[method(name = "getBlocksAtIdx")]
+    async fn get_blocks_at_idx(&self, idx: u64) -> RpcResult<Vec<HexBytes32>>;
+
+    /// Get chain tip block id
+    #[method(name = "blockNumber")]
+    async fn block_number(&self) -> RpcResult<u64>;
```
delbonis, line 27: Duplicates `tip_slot` from the current chainstate, don't need a RPC just for this.

delbonis (changes requested):

delbonis:
Also should take a look at STR-697, as this overlaps.

crates/rpc/api/src/lib.rs
purusang: Removed get_last_chainstate_idx.

crates/rpc/api/src/lib.rs
purusang: Removed `get_last_clientstate_idx`.

crates/rpc/api/src/lib.rs
purusang: We can move this later when sequencer don't have direct connection.

bin/strata-client/src/rpc_server.rs
```diff
@@ -660,6 +701,21 @@ impl SequencerServerImpl {
...
+    async fn get_last_broadcast_entry(&self) -> RpcResult<Option<L1TxEntry>> {
+        let broadcast_handle: Arc<L1BroadcastHandle> = self.broadcast_handle.clone();
+        let txentry = broadcast_handle.get_last_broadcast_entry().await;
+        Ok(txentry.map_err(|e| Error::Other(e.to_string()))?)
+    }
+
+    async fn get_broadcast_entry_by_idx(&self, txid: HexBytes32) -> RpcResult<Option<L1TxEntry>> {
+        let mut txid = txid.0;
+        txid.reverse();
+        let id = Buf32::from(txid);
```
delbonis, line 713: We shouldn't be doing this parsing all on the fly, should have some better way to deal with it.

Also this says "idx" in the fn name but it's taking a txid, what's the deal with that?

bin/strata-client/src/rpc_server.rs
```diff
@@ -105,6 +108,44 @@ fn conv_blk_header_to_rpc(blk_header: &impl L2Header) -> RpcBlockHeader {
...
 impl<D: Database + Send + Sync + 'static> StrataApiServer for StrataRpcImpl<D> {
+    async fn get_blocks_at_idx(&self, idx: u64) -> RpcResult<Vec<HexBytes32>> {
+        let l2_block_manager = self.l2_block_manager.clone();
+        let l2_blocks = l2_block_manager
+            .get_blocks_at_height_async(idx)
+            .await
+            .map_err(Error::Db)?;
+        let block_ids = l2_blocks
+            .iter()
+            .map(|id| HexBytes32(*id.as_ref()))
```
delbonis, line 119: We should add a `From` impl or even just `new` that makes this a little cleaner.

crates/db/src/traits.rs
```diff
@@ -150,6 +150,11 @@ pub trait ClientStateDatabase {
 
     /// Gets a state checkpoint at a previously written index, if it exists.
     fn get_state_checkpoint(&self, idx: u64) -> DbResult<Option<ClientState>>;
+
+    fn get_last_client_state_idx(&self) -> DbResult<u64>;
+
+    /// Get client state at idx
+    fn get_state_at(&self, idx: u64) -> DbResult<Option<ClientState>>;
```
delbonis, line 157: No!  You don't want to make the database do this, it creates circular dependencies.  The consumer should be doing the fetches itself and reconstruct the client state itself.

crates/db/src/stubs/chain_state.rs
```diff
@@ -160,4 +160,19 @@ impl ChainstateDatabase for StubChainstateDb {
...
+        }
+
+        let mut state = st.toplevels.get(&0).cloned().unwrap();
+        for i in 1..=idx {
+            let batch = st.write_batches.get(&i).cloned().unwrap();
+            state = state_op::apply_write_batch_to_chainstate(state, &batch);
+        }
+
+        Ok(Some(state))
+    }
```
delbonis, line 177: See other comment, the database shouldn't be doing this.

crates/rocksdb-store/src/chain_state/db.rs
```diff
@@ -49,13 +49,18 @@ impl ChainstateDatabase for ChainstateDb {
     }
 
     // TODO: define what toplevel means more clearly
+    // TODO: Why is it called toplevel when it is returning any record based on index?
```
delbonis, line 52: Consider in Ethereum where there's a whole state trie of accounts.  We don't load all of those into memory at once.  We can think of that as being the "deep" state.  We don't have deep state like that on the Strata chain right now, but we *might*.  So "toplevel" in this context refers to all of the state that we *can* load into memory unconditionally.

crates/db/src/traits.rs
```diff
@@ -226,6 +231,8 @@ pub trait ChainstateDatabase {
 
     /// Gets the toplevel chain state at a particular block index (height).
     fn get_toplevel_state(&self, idx: u64) -> DbResult<Option<Chainstate>>;
+
+    fn get_state_at(&self, idx: u64) -> DbResult<Option<Chainstate>>;
```
delbonis, line 235: Same thing as before, don't make the database reconstruct this.  We can reconstruct it ourselves.  And right now we store all states anyways, so you can just call `get_toplevel_state`.

crates/rocksdb-store/src/chain_state/db.rs
```diff
@@ -49,13 +49,18 @@ impl ChainstateDatabase for ChainstateDb {
...
     fn get_toplevel_state(
         &self,
         idx: u64,
     ) -> DbResult<Option<strata_state::chain_state::Chainstate>> {
         Ok(self.db.get::<ChainstateSchema>(&idx)?)
     }
 
+    fn get_state_at(&self, idx: u64) -> DbResult<Option<strata_state::chain_state::Chainstate>> {
+        Ok(self.db.get::<ChainstateSchema>(&idx)?)
+    }
```
delbonis, line 62: Remove this.

delbonis (changes requested):
As a rule of thumb, if implementing some feature involves reaching down multiple layers and adding new functionality deep down, then either some more discussion is needed to properly scope the ticket and design the new functionality better *or* there's a misunderstanding somewhere.

bin/strata-client/src/rpc_server.rs
purusang: Thanks! It should be by idx.

crates/rocksdb-store/src/chain_state/db.rs
purusang: Got it, thanks!

crates/db/src/traits.rs
purusang: Doing it in `rpc_server.rs` is fine? Please comment on the next commit that I'm gonna make.

crates/db/src/traits.rs
purusang: Reconstructing it in `rpc_server.rs` is fine? Please comment on the next commit that I'm gonna make.

bin/strata-client/src/rpc_server.rs
```diff
@@ -138,12 +141,11 @@ impl<D: Database + Send + Sync + 'static> StrataApiServer for StrataRpcImpl<D> {
...
     async fn get_clientstate_at_idx(&self, idx: u64) -> RpcResult<Option<ClientState>> {
-        let db = self.database.clone();
-        let client_state = wait_blocking("client_state_at_idx", move || {
-            db.client_state_db().get_state_at(idx).map_err(Error::Db)
-        })
-        .await?;
-        Ok(client_state)
+        let database = self.database.clone();
+        let client_state_db = database.client_state_db().as_ref();
+        let cs = reconstruct_state(client_state_db, idx)
```
delbonis, line 146: This is blocking, still needs to wrap it in `wait_blocking`.  Eventually this would be replaced with the client state manager which does the reconstruction internally.

delbonis (changes requested):

crates/rpc/api/src/lib.rs
```diff
@@ -133,6 +147,14 @@ pub trait StrataAdminApi {
 #[cfg_attr(not(feature = "client"), rpc(server))]
 #[cfg_attr(feature = "client", rpc(server, client))]
 pub trait StrataSequencerApi {
+    /// Get the last broadcast entry
+    #[method(name = "strata_getLastBroadcastEntry")]
+    async fn get_last_broadcast_entry(&self) -> RpcResult<Option<L1TxEntry>>;
+
+    /// Get the broadcast entry by its id
```
bewakes, line 154:
```suggestion
    /// Get the broadcast entry by its idx
```

crates/rpc/api/src/lib.rs
```diff
@@ -1,18 +1,32 @@
...
+    #[method(name = "getBlocksAtIdx")]
+    async fn get_blocks_at_idx(&self, idx: u64) -> RpcResult<Vec<HexBytes32>>;
+
+    /// Get the ChainState at a certain index
+    #[method(name = "getChainstateAtIdx")]
+    async fn get_chainstate_at_idx(&self, idx: u64) -> RpcResult<Option<RpcChainState>>;
+
+    /// Get the ClientState at a certain index
+    #[method(name = "getClientStateAtIdx")]
+    async fn get_clientstate_at_idx(&self, idx: u64) -> RpcResult<Option<ClientState>>;
```
bewakes, line 28: Should these also be a part of debug RPC?

Also, would it make sense to have a new type `RpcClientState` instead of exposing the `ClientState` if this is not a part of debug RPC?

crates/rocksdb-store/src/broadcaster/db.rs
```diff
@@ -82,6 +82,14 @@ impl L1BroadcastDatabase for L1BroadcastDb {
             )))
         }
     }
+
+    fn get_last_broadcast_entry(&self) -> DbResult<Option<L1TxEntry>> {
```
bewakes, line 86: We should also add a corresponding unit test for this.

bin/strata-client/src/rpc_server.rs
```diff
@@ -660,6 +711,18 @@ impl SequencerServerImpl {
...
 #[async_trait]
 impl StrataSequencerApiServer for SequencerServerImpl {
+    async fn get_last_broadcast_entry(&self) -> RpcResult<Option<L1TxEntry>> {
+        let broadcast_handle: Arc<L1BroadcastHandle> = self.broadcast_handle.clone();
+        let txentry = broadcast_handle.get_last_broadcast_entry().await;
+        Ok(txentry.map_err(|e| Error::Other(e.to_string()))?)
+    }
+
+    async fn get_broadcast_entry_by_idx(&self, idx: u64) -> RpcResult<Option<L1TxEntry>> {
+        let broadcast_handle = self.broadcast_handle.clone();
```
bewakes, line 721: Probably don't need to clone here and above?

bin/strata-client/src/rpc_server.rs
```diff
@@ -734,3 +797,26 @@ impl StrataSequencerApiServer for SequencerServerImpl {
...
+
+#[async_trait]
+impl StrataDebugApiServer for StrataDebugRpcImpl {
+    async fn get_block_by_id(&self, block_id: L2BlockId) -> RpcResult<Option<L2Block>> {
+        let l2_block_manager = self.l2_block_manager.clone();
+        let l2_block = l2_block_manager
+            .get_block_async(&block_id)
+            .await
+            .map_err(Error::Db)?
+            .map(|b| b.block().clone());
```
bewakes, line 819: Not sure but maybe we don't need a clone here?

bin/strata-client/src/rpc_server.rs
```diff
@@ -734,3 +797,26 @@ impl StrataSequencerApiServer for SequencerServerImpl {
...
+impl StrataDebugRpcImpl {
+    pub fn new(l2_block_manager: Arc<L2BlockManager>) -> Self {
+        Self { l2_block_manager }
+    }
+}
+
+#[async_trait]
+impl StrataDebugApiServer for StrataDebugRpcImpl {
+    async fn get_block_by_id(&self, block_id: L2BlockId) -> RpcResult<Option<L2Block>> {
+        let l2_block_manager = self.l2_block_manager.clone();
```
bewakes, line 814: And here?

bewakes (commented):
Looks good, minor nits.

crates/rpc/api/src/lib.rs
purusang: I think it does make sense to have them in debug_. The notion behind types Rpc* is due to Serialization trait not being derived for those structs or to have a stripped down version of the struct. If we move these to debug then this should not be an issue.
cc: @delbonis

bin/strata-client/src/rpc_server.rs
purusang: .block() returns reference so it's needed in this case.

crates/rocksdb-store/src/broadcaster/db.rs
purusang: also, feels like the name should be `get_last_tx_entry` ?

purusang:
In a discussion with @delbonis, regarding debug apis, he suggested a nice idea of returning native byte representation as hex so that we do not need additional `Serialize`. What are your thoughts on it? @bewakes @sapinb . I'm positive about it.

crates/rocksdb-store/src/broadcaster/db.rs
bewakes: sounds good

bewakes:
> returning native byte representation as hex so that we do not need additional Serialize

As long as we can inspect the data with ease, it should be fine.

sapinb:
> In a discussion with @delbonis, regarding debug apis, he suggested a nice idea of returning native byte representation as hex so that we do not need additional `Serialize`. What are your thoughts on it @bewakes @sapinb?. I'm positive about it.

My opinion is to keep things human readable for the time being. these are meant to help us in debugging. An opaque data that needs another tool to parse first will just make it harder to use. Hex representation would be good for communication between 2 nodes.

delbonis (approved):
Alright then I guess this is in sufficient working order for now.

storopoli (approved):
ACK f1cf675b204cbfbb96df56d64e3bc9cbb7bd091d

We need to manually update the private docs for the new/updated RPCs.

PR #539, opened 2024-12-13
STR-722: refactor random generation to `CryptoRngCore`

AaronFeickert:
## Description

This PR refactors the trait bound `CryptoRng + RngCore` to `CryptoRngCore` from `rand_core` and performs associated import cleanup.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [x] Dependency Update

## Notes to Reviewers

None.

## Checklist

## Related Issues

None.


storopoli:
The idea is to break up STR-722 into small easy-to-review PRs?
Then in the final one remove `rand` altogether for `rand-core`?

AaronFeickert:
> The idea is to break up STR-722 into small easy-to-review PRs? Then in the final one remove `rand` altogether for `rand_core`?

The primary goal is to use `CryptoRngCore` instead of `CryptoRng + RngCore` to reduce the likelihood of missing `CryptoRng`. However, `CryptoRngCore` is only provided by `rand_core`, whereas `CryptoRng` and `RngCore` are _also_ provided by `rand`. With a few small extra import updates, it happened to be possible to swap out `rand` for `rand_core` as a dependency in a couple of crates, so I went ahead and did that to clean things up a bit.

There are a few places where we really do need `rand` to get access to sampling and shuffling, so it will stick around. Initially, I went a bit wild with swapping out to `rand_core`, but realized that's out of scope here and not the intent of the issue.

bin/strata-cli/src/cmd/faucet.rs
```diff
@@ -71,15 +70,17 @@ pub async fn faucet(args: FaucetArgs, seed: Seed, settings: Settings) {
...
         prehash.clone(),
         challenge.difficulty,
         solution.to_le_bytes(),
     ) {
         solution += 1;
-        if (0..100).sample_single(&mut OsRng) == 0 {
+        if counter % 100 == 0 {
             pb.set_message(format!("Trying {solution}"));
         }
+        counter += 1;
```
storopoli, line 83: This is not calling `sample_single` anymore.
Wasn't this needed for the PoW?

storopoli (changes requested):
Just one question to clarify otherwise good to go.

storopoli (approved):
ACK 4605cf31980e1dd6934daa3a371cbaa1da226d9f

delbonis (approved):

PR #544, opened 2024-12-16
chore(storage) use `parking_lot` mutex in `CacheTable`

Zk2u:
## Description

Switches the `tokio::sync::Mutex` in the storage crate's `CacheTable` to a `parking_lot::Mutex` for improved performance. Meant to submit last friday but didn't open PR.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

New mutex is sync not async, so it's imperative to keep the critical path as short as possible and not hold across awaits. I believe I've done this but this is worth double checking.

## Checklist

## Related Issues

STR-705
STR-237


delbonis (dismissed):
This should also close [this task](https://alpenlabs.atlassian.net/browse/STR-237) since we're renaming the fns as described.

I wonder if there's a good way here to print a warning if the executor does yield in the async calls?

storopoli (approved):
ACK 0900918ea0350e71407c98603f9fb85213506c24
Nice work.
Can you override the `clippy` warning so that CI passes?

storopoli (approved):
ACK 8dec44ebabbfcd5813af35859b16fd7898c9008a

delbonis:
@Zk2u Does parking_lot have a way to specifically wake a thread after unlocking a mutex?  I don't mean across an async yield point, I mean if two different executors try to acquire the lock at the same time.

I guess this would be exceptionally unlikely to actually happen right now.  The only way this would happen is on a RPC full node and while being hammered by requests.

delbonis:
I *know* this isn't a significant issue but we can't say it's the same behavior.

PR #537, opened 2024-12-13
STR-676: updates rust nightly and sets a CI autoupdate cron job

storopoli:
## Description

Updates rust nightly version and adds a CI cron job to do a PR monthly with my GH token (with very limited scope and permissions) saved in repository secrets.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

None so far.

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have commented my code where necessary.
- [ ] I have updated the documentation if needed.
- [ ] My changes do not introduce new warnings.
- [ ] I have added tests that prove my changes are effective or that my feature works.
- [ ] New and existing tests pass with my changes.

## Related Issues

STR-676

PR #545, opened 2024-12-16
ci: fix automated `Cargo.lock` dependency update

storopoli:
## Description

Fixes the automated `Cargo.lock` dependency update.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update
- [x] CI

## Notes to Reviewers

I think this will work with the default `GITHUB_TOKEN`.

Since:

> Note that due to [token restrictions on public repository forks](https://docs.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#permissions-for-the-github_token), workflows for this use case do not work for pull requests raised from forks. Private repositories can be configured to [enable workflows](https://docs.github.com/en/github/administering-a-repository/disabling-or-limiting-github-actions-for-a-repository#enabling-workflows-for-private-repository-forks) from forks to run without restriction.

Source: https://github.com/peter-evans/create-pull-request/blob/main/docs/examples.md#use-case-create-a-pull-request-to-modifyfix-pull-requests

## Checklist

## Related Issues

https://github.com/alpenlabs/strata/pull/536#issuecomment-2546199070

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -6,12 +6,11 @@ on:
   schedule:
     # Run at the first of every month
     - cron: "0 0 1 * *"
-  workflow_dispatch:
-    # Needed so we can run it manually
+  workflow_dispatch: # Needed so we can run it manually
 
 env:
-  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  BRANCH: cargo-update
+  AUTHOR: Update Cargo.lock Bot <no-reply@alpenlabs.io>
+  BRANCH: create-pull-request/automated-cargo-update
   TITLE: "chore(deps): monthly `cargo update`"
   BODY: |
     Automation to keep dependencies in `Cargo.lock` current.
@@ -30,6 +29,9 @@ jobs:
   update:
     name: Update
     runs-on: ubuntu-latest
+    permissions:
+      contents: write # Needed to create commits
+      pull-requests: write # Needed to create a PR
     steps:
       - uses: actions/checkout@v4
         with:
@@ -68,6 +70,7 @@ jobs:
         with:
           add-paths: ./Cargo.lock
           commit-message: ${{ steps.msg.outputs.commit_message }}
+          author: ${{ env.AUTHOR }}
           title: ${{ env.TITLE }}
           body: ${{ steps.msg.outputs.body }}
           branch: ${{ env.BRANCH }}


Rajil1213:
The issue with the `Dependency Update` workflow was that it requires write permissions to the repo as per https://github.com/peter-evans/create-pull-request/issues/968. Has that been fixed? If the required permissions have been provided, I don't think this PR needs any changes to the actual workflow file.

.github/workflows/dependencies.yml
```diff
@@ -66,8 +68,10 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v7
         with:
+          token: ${{ secrets.STOROPOLI_CREATE_PR_TOKEN }}
```
Rajil1213, line 71: This looks like a bad idea. Probably best to have a separate token for github actions (and named as such).

storopoli:
> The issue with the `Dependency Update` workflow was that it requires write permissions to the repo as per [peter-evans/create-pull-request#968](https://github.com/peter-evans/create-pull-request/issues/968). Has that been fixed? If the required permissions have been provided, I don't think this PR needs any changes to the actual workflow file.

I prefer to keep the repo-wide permissions off and enable on a per-workflow basis.
That's what this does: https://github.com/alpenlabs/strata/blob/5bb20c42b7ba6fc395af675d61720aa2ec218276/.github/workflows/dependencies.yml#L32-L34

.github/workflows/dependencies.yml
storopoli: We had a brief discussion about the token here: https://alpenlabs.atlassian.net/browse/STR-676

Rajil1213: Replied to the ticket. But the idea is to basically use a separate account, or a token managed by the repo/org owners. Using a personal token should really be our last resort if all else fails.

Rajil1213:
> I prefer to keep the repo-wide permissions off and enable on a per-workflow basis.
> That's what this does:

I see. Looks fine then. Could you check if there is a way to not use personal tokens though?

.github/workflows/dependencies.yml
Rajil1213: So the restriction on PRs created from within this repo by github actions has been lifted?

storopoli: AFAIK, you need to add the permissions as I did in https://github.com/alpenlabs/strata/blob/5bb20c42b7ba6fc395af675d61720aa2ec218276/.github/workflows/dependencies.yml#L32-L34

.github/workflows/dependencies.yml
```diff
@@ -6,12 +6,11 @@ on:
...
     # Run at the first of every month
     - cron: "0 0 1 * *"
-  workflow_dispatch:
-    # Needed so we can run it manually
+  workflow_dispatch: # Needed so we can run it manually
 
 env:
-  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  BRANCH: cargo-update
+  AUTHOR: Update Cargo.lock Bot <bot@example.com>
```
AaronFeickert, line 12: Why are we using a generic domain?

storopoli: No reason. I was just copying `rust-bitcoin` template since I'm familiar with that.
Any suggestions?

storopoli: NOTE: All suggestion discussed (and adopted) here I am also applying to #537.

AaronFeickert: Is it required to include an email? If so, maybe just use a `no-reply@alpenlabs.io`.

storopoli: Done.

storopoli:
We are in a stalemate here.
Why don't we merge this and I'll trigger the manual CI and report back to you?

Rajil1213 (approved):

PR #547, opened 2024-12-18
Fix resubmit checkpoint bug

voidash:
## Description

Since anyone can post inscriptions, a malicious user can resend checkpoint inscription that has already been published and sequencer stops working. This PR fixes that particular bug

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist

This 

## Related Issues

STR-795


functional-tests/fn_btcio_resubmit_checkpoint.py
```diff
@@ -0,0 +1,85 @@
...
+    op_endif = "68"
+    op_pushbytes_33 = "21"
+    op_false = "00"
+    start_position = inp.index(f"{op_false}{op_if}")
+    end_position = inp.index(f"{op_endif}{op_pushbytes_33}", start_position)
+    op_if_block = inp[start_position + 3 : end_position]
+    op_pushdata = "4d"
+    pushdata_position = op_if_block.index(f"{op_pushdata}")
+    # we don't want PUSHDATA + num bytes b401
+    return op_if_block[pushdata_position + 2 + 4 :]
```
storopoli, line 85: Are we gonna use this in other tests?
If yes we should move this to `utils.py`.

delbonis:
How does this change fix the bug?  I would expect that the change would be to check if the epoch is less than or equal to the current finalized epoch then we just ignore it the checkpoint.

voidash:
> How does this change fix the bug? I would expect that the change would be to check if the epoch is less than or equal to the current finalized epoch then we just ignore it the checkpoint. I don't follow what was changed.

There was already a check present in the code. Instead of passing those filtered checkpoints, we were passing the initial checkpoint vector. Changing it to filtered checkpoint fixes the problem

functional-tests/fn_btcio_resubmit_checkpoint.py
voidash: This might be helpful in other tests but we can move to utils when such scenario arrives.

storopoli: Alright then do it please?

functional-tests/utils.py
```diff
@@ -464,3 +465,30 @@ def setup_test_logger(datadir_root: str, test_name: str) -> logging.Logger:
...
+    pushdata_position = op_if_block.index(f"{op_pushdata}")
+    # we don't want PUSHDATA + num bytes b401
+    return op_if_block[pushdata_position + 2 + 4 :]
+
+
+def submit_da_blob(btcrpc: BitcoindClient, seqrpc: JsonrpcClient, blobdata: str):
+    _ = seqrpc.strataadmin_submitDABlob(blobdata)
+
+    # Allow some time for sequencer to publish blob
+    time.sleep(SEQ_PUBLISH_BATCH_INTERVAL_SECS)
```
storopoli, line 488: Can we replace this for `wait_until`?
Only if we cannot or does not make sense to use `wait_until` we should use `time.sleep`.

`wait_until` is more efficient. Especially if you customize the kwargs.

voidash: submitDABlob doesn't provide txid so there is no direct way other than comparing previous last_l1_txid status. I guess i should just do that. Also i think there is this edge case where our new checkpoint batch might be posted after posting submitted DA blob, which is also be handled.

storopoli (approved):
ACK 055793df07b959b153c32541cd101b9d35d42cd9

delbonis (approved):

PR #552, opened 2024-12-19
feat(storage) `L1BlockManager`

Zk2u:
## Description

This PR adds a `L1BlockManager` struct that handles caching of L1 block data.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

I'm not using `inst_ops!` because it's kinda manky 🦨

- [ ] open an additional ticket to clean the whole storage crate up lol

## Checklist

## Related Issues

STR-797


delbonis:
> I'm not using inst_ops! because it's kinda manky 🦨

But now you're introducing new paradigms and consumers have to import additional traits to use it in addition to using the type itself.  Why?

Can we discuss things in #eng or in meetings before we start changing strategy here so that we consider all the needs across the team?

delbonis:
This also still retains the `L1Database` type parameter.  Hiding that baggage through the `inst_ops!` macro and then building shared logic around it is the whole goal of having the ops primitive in the first place.  Yes there's a tiny perf hit because of the indirect call, but it's calling IO and it's only a single indirect call per database call so it's fine (and it's not in a BoxFuture that would need an indirect call each time it's polled).

crates/storage/src/managers/l1.rs
```diff
@@ -0,0 +1,124 @@
...
+};
+use strata_state::{block::L2BlockBundle, header::L2Header, id::L2BlockId, l1::L1Tx};
+use threadpool::ThreadPool;
+use tokio::sync::oneshot::{self, error::RecvError};
+use tracing::warn;
+
+use crate::cache::{self, CacheTable};
+
+/// Caching manager of L1 blocks in the block database.
+pub struct L1BlockManager<DB>
```
bewakes, line 20: One of the major reasons to have managers was to be free from these Db generics. The `Ops` construct was to facilitate this.

delbonis: You have to pass it down through everything that wants to interact with the database / manager, which becomes cumbersome.

delbonis:
The need is that it cleanly wraps up all the machinery to implement async and blocking calls over a simpler blocking database interface and give some extra space for future indirection in the `Context` type.

delbonis:
The main issue right now is that this doesn't remove the need to pass around the generic type parameter as in the current strategy.  We *don't* want the manager to just transparently implement the database traits but with additional caching, we want it to be the *high level* interface that can hide the implementation details.  This is more apparent in the [intention for the chainstate manager](https://alpenlabs.atlassian.net/browse/STR-182), which we want to be the main thing handling reconstructing arbitrary states from the checkpointed states and writes.  The architecture with the other managers/ops also gives a lot of flexibility for transparently adding things like metrics collection to improve visibility.

The generics are great for performance sensitive code (because of monomorphization helping with automatic inlining) or logic that's highly abstracted (with associated types).  But this isn't that, we're just using them for indirection between different underlying database impls.

bewakes:
Since Azz is out, I'll be taking this over.

bewakes:
Closing this in favor of another PR: https://github.com/alpenlabs/strata/pull/568

PR #546, opened 2024-12-17
Str-700: Preparation work for the prover CI performance metrics

evgenyzdanovich:
## Description

Preparation work for the prover CI performance metrics. Includes:
- define `ProofReport` simple structure
- factor out `bin/prover-client/zkvm` into a separate crate. Reuse this crate in provers/tests. Later use it for in prover CI performance metrics.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have commented my code where necessary.
- [ ] I have updated the documentation if needed.
- [ ] My changes do not introduce new warnings.
- [ ] I have added tests that prove my changes are effective or that my feature works.
- [ ] New and existing tests pass with my changes.

PR #555, opened 2024-12-20
Str-796: ZkVMs prover performance metrics

evgenyzdanovich:
## Description

Introduce performance metrics for provers. Currently, only displays `cycle_count` for each ZkVm program.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

STR-796


provers/perf/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+mod reports;
+/// A proof report containing a performance stats about proof generation.
+#[derive(Debug, Clone)]
+pub struct ProofReport {
+    pub cycles: u64,
+    pub report_name: String,
+}
+
+/// An extension trait that supports performance report for [`ZkVmHost`].
+pub trait ZkVmHostPerf: ZkVmHost {
```
evgenyzdanovich, line 19: Previously, in the preparation work (#546), the `ZkVmHost` has been extended with this method `perf_report`, see `zkvm/src/host.rs`. 

However, after thinking for a while, I decided to make it as a separate subtrait of `ZkVmHost` that's defined and implemented in the place where it "semantically" belongs - `provers/perf` folder.

Prajwol and I have different opinions on this approach: I'm for the split, Prajwol is in favor of `perf_report` staying a part of `ZkVmHost`

**My rationales, quoting:**

my concern during thinking about that and why I decided to split it:
- perf-report can potentially come in different shapes. It can be like now where it only contains 
cycles , but it also can be a full execution report (with execution/proof/verify timings)
 and having it defined by the caller makes more sense to me.
  - Having it defined by the zkvm, we are "polluting" core host with different
 (quite unrelated to the core logic) methods.
- if we define it on the host level, this means we are required to bring all the dependencies 
that require is to measure performance to the adapters crait
- Also, if we provide zkvm as a library (separate repo) where it can be useful for other devs -
they may not care that much about perf and implementing unnecessary rigid methods 
can be a bit annoying (instead of scaffolding-like approach, where you implement only what's necessary)


**Prajwol's rationales, quoting:**

1. totally agree on perf-report coming in different shapes and having it be defined by the caller makes more sense.
my rationale is the end-user should be able to request for the perf-metrics and be available to them regardless of the zkvm being used. they should have a coherent experience across different zkvms
2. i think the largest dependency will be zkvm specific sdk (might be totally wrong). And we already have that there in the zkvm-specific-host
3. i think perf metrics is something that's really needed and most zkvms will support it in some way.


For now, we agreed to gather the feedback from other team members and make a decision as a team.

**Kindly asking to spend some time and express your opinion here**: @delbonis @storopoli @MdTeach @prajwolrg @Rajil1213 (please feel free to list a person who may have strong grounded opinion here).

prajwolrg: > However, after thinking for a while, I decided to make it as a separate subtrait of ZkVmHost that's defined and implemented in the place where it "semantically" belongs - provers/perf folder.

> Prajwol and I have different opinions on this approach: I'm for the split, Prajwol is in favor of perf_report staying a part of ZkVmHost

To clarify, I am not opposed to splitting perf_report from `ZkVmHost`, but I believe it should remain part of the crate specific to a ZkVm. For example, importing `sp1-adapter` should include perf_report (it can be behind a feature flag but I should not need to import other crate specific to sp1). Everything related to a particular ZkVm should be accessible through its adapter, and the behavior across different ZkVm should be consistent to ensure a coherent user experience.

evgenyzdanovich: @prajwolrg thanks for your clarification!

To others, just to be clear what's the topic here:
> For example, importing `sp1-adapter` should include `perf_report`

this is exactly where me and Prajwol diverge. I believe it shall be a part of `perf` crait, not adapter.

> behavior across different ZkVm should be consistent to ensure a coherent user experience.
To ensure consistent behavior, we can expose/extend `perf` crait where all the logic related to performance reporting belongs

storopoli: I prefer modularity and "feature-gate" for stuff that is non-essential. I see valid arguments in both sides. But for me this principle is paramount.

Hence I think that it should be in a different crate, feature-gated, `provers/perf`.

This will also help with compile times which feels like a monotonically increasing function over time, and we should push back on this.

delbonis: Yeah I agree with making it feature-gated.  It's super useful and important, but it's also valuable to be able to make prod mainnet builds with it disabled.  I'm not sure the best strategy for achieving that.

delbonis (approved):
Approved but still maybe pending some action about what's discussed in that thread?

provers/perf/src/lib.rs
evgenyzdanovich: @delbonis I guess the best way to achieve prod builds without perf is to simply exclude it from the build with `--exclude` option: `cargo build --release --workspace --exclude strata-provers-perf`.

Feature-gating would not really help, as having `perf` in the workspace members would still make `provers/perf` (at least minimal `provers/perf/src/main.rs` no matter how deep we feature-gate it) into the build.

Alternatively, we can exclude `provers/perf` from the root workspace members  (the same way prover guest zkvms are not a part of the workspace members, for example) and thus making it completely separate thingy.

evgenyzdanovich:
@delbonis thanks! I'll discuss with Prajwol to make sure we are on the same page and resolve the conversation/make necessary adjustments.

I'll ping you to reapprove in case there were some changes required

provers/perf/src/lib.rs
evgenyzdanovich: Discussed with Prajwol, decided to leave as is:

- option to have a prod build without `perf` can be achieved as described above: `cargo build --release --workspace --exclude strata-provers-perf`
- for now, we will stay with `perf_report` in a separate trait, we will revisit this approach with the usage and duing porting zkvm into a separate repo.

Resolving for now, thanks everyone.

PR #557, opened 2024-12-23
STR-734: OPERATOR REASSIGNMENT 

voidash:
## Description

Functional test where we make deposits, then trigger the withdrawal. client associated with assigned operator id is stopped. We produce l1 blocks so that the dispatch assignment duration is over, and later check if new operator is being assigned or not

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

STR-734

### DIFF
--- a/bin/bridge-client/src/args.rs
+++ b/bin/bridge-client/src/args.rs
@@ -1,4 +1,5 @@
 //! Parses command-line arguments for the bridge-client CLI.
+
 use std::fmt::Display;
 
 use argh::FromArgs;
@@ -9,73 +10,104 @@ use crate::errors::InitError;
 #[argh(name = "strata-bridge-client")]
 #[argh(description = "The bridge client for Strata")]
 pub(crate) struct Cli {
+    /// Specifies the mode to run the client in: either Operator (alias: op) or Challenger (alias:
+    /// ch).
     #[argh(
         positional,
         description = "what mode to run the client in, either Operator (alias: op) or Challenger (alias: ch)"
     )]
     pub mode: String,
 
+    /// Path to the directory where RocksDB databases are stored.
+    /// Defaults to `$HOME/.local/share/strata/` if not specified.
     #[argh(
         option,
         description = "path to the directory where to store the rocksdb databases (default: $HOME/.local/share/strata/)"
     )]
     pub datadir: Option<String>,
 
+    /// Master operator key in xpriv format. Defaults to the environment variable
+    /// `STRATA_OP_MASTER_XPRIV` if not provided.
     #[argh(
         option,
         description = "xpriv to be used as the master operator's key (default: envvar STRATA_OP_MASTER_XPRIV)"
     )]
     pub master_xpriv: Option<String>,
 
+    /// Path to the file containing the master operator's xpriv.
+    /// Should not be used with the `--master-xpriv` option or `STRATA_OP_MASTER_XPRIV` environment
+    /// variable.
     #[argh(
         option,
         description = "path to the file containing the master operator's xpriv (don't use with --master-xpriv or the envvar STRATA_OP_MASTER_XPRIV)"
     )]
     pub master_xpriv_path: Option<String>,
 
+    /// Host address for the RPC server. Defaults to `127.0.0.1` if not specified.
     #[argh(
         option,
         description = "host to run the rpc server on (default: 127.0.0.1)"
     )]
     pub rpc_host: Option<String>,
 
+    /// Port number for the RPC server. Defaults to `4781` if not specified.
     #[argh(option, description = "port to run the rpc server on (default: 4781)")]
     pub rpc_port: Option<u32>,
 
+    /// URL for the Bitcoin RPC endpoint.
     #[argh(option, description = "url for the bitcoin RPC")]
     pub btc_url: String,
 
+    /// Username for accessing the Bitcoin RPC.
     #[argh(option, description = "username for bitcoin RPC")]
     pub btc_user: String,
 
+    /// Password for accessing the Bitcoin RPC.
     #[argh(option, description = "password for bitcoin RPC")]
     pub btc_pass: String,
 
+    /// URL for the Rollup RPC server.
     #[argh(option, description = "url for the rollup RPC server")]
     pub rollup_url: String,
 
+    /// Interval for polling bridge duties in milliseconds.
+    /// Defaults to the block time according to the Rollup RPC.
     #[argh(
         option,
         description = "bridge duty polling interval in milliseconds (default: block time according to rollup RPC)"
     )]
-    pub duty_interval: Option<u64>, // default: same as rollup block time
+    pub duty_interval: Option<u64>,
 
+    /// Interval for polling bridge messages in milliseconds.
+    /// Defaults to half the block time according to the Rollup RPC.
     #[argh(
         option,
         description = "bridge message polling interval in milliseconds (default: half of the block time according to the rollup RPC client)"
     )]
-    pub message_interval: Option<u64>, // default: half of the rollup block time
+    pub message_interval: Option<u64>,
 
+    /// Number of retries for RocksDB database operations. Defaults to `3`.
     #[argh(
         option,
         description = "retry count for the rocksdb database (default: 3)"
     )]
     pub retry_count: Option<u16>,
+
+    /// Timeout duration for duties in seconds. Defaults to `600`.
+    #[argh(
+        option,
+        description = "duty timeout duration in seconds (default: 600)"
+    )]
+    pub duty_timeout_duration: Option<u64>,
 }
 
+/// Represents the operational mode of the client.
+///
+/// - `Operator`: Handles deposits, withdrawals, and challenges.
+/// - `Challenger`: Verifies and challenges Operator claims.
 #[derive(Debug, Clone)]
 pub(super) enum OperationMode {
-    /// Run client in Operator mode to handle deposits, withdrawals and challenging.
+    /// Run client in Operator mode to handle deposits, withdrawals, and challenging.
     Operator,
 
     /// Run client in Challenger mode to verify/challenge Operator claims.
--- a/bin/bridge-client/src/constants.rs
+++ b/bin/bridge-client/src/constants.rs
@@ -4,5 +4,6 @@ pub(super) const DEFAULT_RPC_PORT: u32 = 4781; // first 4 digits in the sha256 o
 
 pub(super) const DEFAULT_RPC_HOST: &str = "127.0.0.1";
 
+pub(super) const DEFAULT_DUTY_TIMEOUT_DURATION: u64 = 600;
 /// The default bridge rocksdb database retry count, if not overridden by the user.
 pub(super) const ROCKSDB_RETRY_COUNT: u16 = 3;
--- a/bin/bridge-client/src/constants.rs
+++ b/bin/bridge-client/src/constants.rs
@@ -4,6 +4,6 @@ pub(super) const DEFAULT_RPC_PORT: u32 = 4781; // first 4 digits in the sha256 o
 
 pub(super) const DEFAULT_RPC_HOST: &str = "127.0.0.1";
 
-pub(super) const DEFAULT_DUTY_TIMEOUT_DURATION: u64 = 600;
+pub(super) const DEFAULT_DUTY_TIMEOUT_SEC: u64 = 600;
 /// The default bridge rocksdb database retry count, if not overridden by the user.
 pub(super) const ROCKSDB_RETRY_COUNT: u16 = 3;
--- a/bin/bridge-client/src/modes/operator/bootstrap.rs
+++ b/bin/bridge-client/src/modes/operator/bootstrap.rs
@@ -28,7 +28,7 @@ use super::{constants::DB_THREAD_COUNT, task_manager::TaskManager};
 use crate::{
     args::Cli,
     constants::{
-        DEFAULT_DUTY_TIMEOUT_DURATION, DEFAULT_RPC_HOST, DEFAULT_RPC_PORT, ROCKSDB_RETRY_COUNT,
+        DEFAULT_DUTY_TIMEOUT_SEC, DEFAULT_RPC_HOST, DEFAULT_RPC_PORT, ROCKSDB_RETRY_COUNT,
     },
     db::open_rocksdb_database,
     rpc_server::{self, BridgeRpc},
@@ -150,9 +150,9 @@ pub(crate) async fn bootstrap(args: Cli) -> anyhow::Result<()> {
         Duration::from_millis,
     );
 
-    let duty_timeout_duration = args.duty_timeout_duration.map_or(
-        Duration::from_secs(DEFAULT_DUTY_TIMEOUT_DURATION),
-        Duration::from_secs,
+    let duty_timeout_duration = Duration::from_secs(
+        args.duty_timeout_duration
+            .unwrap_or(DEFAULT_DUTY_TIMEOUT_SEC),
     );
 
     // TODO: wrap these in `strata-tasks`
--- a/bin/bridge-client/src/modes/operator/bootstrap.rs
+++ b/bin/bridge-client/src/modes/operator/bootstrap.rs
@@ -27,7 +27,9 @@ use tracing::{error, info};
 use super::{constants::DB_THREAD_COUNT, task_manager::TaskManager};
 use crate::{
     args::Cli,
-    constants::{DEFAULT_RPC_HOST, DEFAULT_RPC_PORT, ROCKSDB_RETRY_COUNT},
+    constants::{
+        DEFAULT_DUTY_TIMEOUT_DURATION, DEFAULT_RPC_HOST, DEFAULT_RPC_PORT, ROCKSDB_RETRY_COUNT,
+    },
     db::open_rocksdb_database,
     rpc_server::{self, BridgeRpc},
     xpriv::resolve_xpriv,
@@ -148,9 +150,17 @@ pub(crate) async fn bootstrap(args: Cli) -> anyhow::Result<()> {
         Duration::from_millis,
     );
 
+    let duty_timeout_duration = args.duty_timeout_duration.map_or(
+        Duration::from_secs(DEFAULT_DUTY_TIMEOUT_DURATION),
+        Duration::from_secs,
+    );
+
     // TODO: wrap these in `strata-tasks`
     let duty_task = tokio::spawn(async move {
-        if let Err(e) = task_manager.start(duty_polling_interval).await {
+        if let Err(e) = task_manager
+            .start(duty_polling_interval, duty_timeout_duration)
+            .await
+        {
             error!(error = %e, "could not start task manager");
         };
     });
--- a/bin/bridge-client/src/modes/operator/task_manager.rs
+++ b/bin/bridge-client/src/modes/operator/task_manager.rs
@@ -15,7 +15,10 @@ use strata_rpc_api::StrataApiClient;
 use strata_rpc_types::RpcBridgeDuties;
 use strata_state::bridge_duties::{BridgeDuty, BridgeDutyStatus};
 use strata_storage::ops::{bridge_duty::BridgeDutyOps, bridge_duty_index::BridgeDutyIndexOps};
-use tokio::{task::JoinSet, time::sleep};
+use tokio::{
+    task::JoinSet,
+    time::{sleep, timeout},
+};
 use tracing::{error, info, trace, warn};
 
 pub(super) struct TaskManager<L2Client, TxBuildContext, Bcast>
@@ -36,13 +39,20 @@ where
     TxBuildContext: BuildContext + Sync + Send + 'static,
     Bcast: Broadcaster + Sync + Send + 'static,
 {
-    pub(super) async fn start(&self, duty_polling_interval: Duration) -> anyhow::Result<()> {
+    pub(super) async fn start(
+        &self,
+        duty_polling_interval: Duration,
+        duty_timeout_duration: Duration,
+    ) -> anyhow::Result<()> {
+        info!(?duty_polling_interval, "Starting to poll for duties");
         loop {
             let RpcBridgeDuties {
                 duties,
                 start_index,
                 stop_index,
             } = self.poll_duties().await?;
+
+            info!(num_duties = duties.len(), "got duties");
 
             let mut handles = JoinSet::new();
             for duty in duties {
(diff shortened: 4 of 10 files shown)


functional-tests/fn_bridge_reassignment.py
```diff
@@ -0,0 +1,73 @@
...
+    Makes two DRT deposits, then triggers the withdrawal.
+    The bridge client associated with assigned operator id is stopped.
+    After the dispatch assignment duration is over,
+    Check if new operator is being assigned or not
+    """
+
+    def __init__(self, ctx: flexitest.InitContext):
+        # Example: we spin up 5 operators
+        ctx.set_env(testenv.BasicEnvConfig(101, n_operators=5))
+
```
storopoli, line 28: Do we need 5? What if we do it with two and take one out? Or maybe even 3?
The tests should be as minimal as possible with respect to resources to replicate the context/environment that we need to test.

functional-tests/fn_bridge_reassignment.py
```diff
@@ -0,0 +1,73 @@
...
+import testenv
+from constants import UNSPENDABLE_ADDRESS
+from fn_bridge_withdraw_happy import (
+    DEPOSIT_AMOUNT,
+    check_initial_eth_balance,
+    deposit_twice,
+    do_withdrawal,
+    setup_services,
+)
```
storopoli, line 13: Maybe we should move these to utils?
What you think?
They are being used elsewhere than `fn_bridge_withdraw_happy`.

functional-tests/fn_bridge_reassignment.py
```diff
@@ -0,0 +1,73 @@
...
+        self.debug(f"Address: {address}")
+        self.debug(f"Change Address: {withdraw_address}")
+
+        btc, seq, reth, seqrpc, btcrpc, rethrpc, web3, el_address = setup_services(ctx, self.debug)
+
+        # Check initial balance is 0
+        check_initial_eth_balance(rethrpc, el_address, self.debug)
+
+        # Perform two deposits
+        deposit_twice(ctx, web3, seqrpc, rethrpc, el_address, self.debug)
```
storopoli, line 41: I think that there's a way for you to setup an environment with already balances in BTC and Strata, AFAIK @evgenyzdanovich implemented that feature.

storopoli: This aligns with the "minimal resources" principle above.

functional-tests/fn_bridge_reassignment.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    Makes two DRT deposits, then triggers the withdrawal.
+    The bridge client associated with assigned operator id is stopped.
+    After the dispatch assignment duration is over,
+    Check if new operator is being assigned or not
+    """
+
+    def __init__(self, ctx: flexitest.InitContext):
+        # Example: we spin up 5 operators
+        ctx.set_env(testenv.BasicEnvConfig(101, n_operators=5))
```
storopoli, line 27: I think that you can also disable auto-generate BTC blocks. This is important to avoid flaky tests when you need total control of the block production (i.e. waiting for 64 blocks below).

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -46,210 +42,292 @@ def __init__(self, ctx: flexitest.InitContext):
...
-        - 6 blocks to mature the DRT
-        - 6 blocks to mature the DT
-        """
-        btcrpc.proxy.generatetoaddress(number_of_blocks, UNSPENDABLE_ADDRESS)
-        wait_until(
-            lambda: get_balance(withdraw_address, btc_url, btc_user, btc_password)
-            > original_balance
-        )
+
+def make_drt(ctx: flexitest.RunContext, el_address, musig_bridge_pk):
```
storopoli, line 89: Maybe we should move this utils?

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -46,210 +42,292 @@ def __init__(self, ctx: flexitest.InitContext):
...
+    # time to mature DRT
+    btcrpc.proxy.generatetoaddress(6, seq_addr)
+    time.sleep(3)
+
+    # time to mature DT
+    btcrpc.proxy.generatetoaddress(6, seq_addr)
+    time.sleep(3)
+
+
+def make_withdraw(
```
storopoli, line 120: This as well?

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -46,210 +42,292 @@ def __init__(self, ctx: flexitest.InitContext):
...
+        "to": PRECOMPILE_BRIDGEOUT_ADDRESS,
+        "value": deposit_amount * SATS_TO_WEI,
+        "gas": gas,
+        "data": data_bytes,
+    }
+    l2_tx_hash = web3.eth.send_transaction(transaction)
+    return l2_tx_hash
+
+
+def estimate_withdraw_gas(web3: Web3, el_address, change_address_pk, deposit_amount):
```
storopoli, line 143: This as well?

functional-tests/fn_bridge_reassignment.py
```diff
@@ -0,0 +1,73 @@
...
+        # Re-check duties
+        duties = seqrpc.strata_getBridgeDuties(0, 0)["duties"]
+        withdraw_duty = [d for d in duties if d["type"] == "FulfillWithdrawal"][0]
+        new_assigned_op_idx = withdraw_duty["payload"]["assigned_operator_idx"]
+        new_assigned_operator = ctx.get_service(f"bridge.{new_assigned_op_idx}")
+
+        # Ensure a new operator is assigned
+        assert new_assigned_operator != assigned_operator, "No new operator was assigned"
+
+        return True
```
storopoli, line 73: We are testing only 2 out of 3 things that are highlighted in the original ticket (STR-734):

1. ~Is the user’s withdrawal request confirmed in a Strata batch?~
2. ~After the first operator fails to initiate the withdrawal, is the withdrawal request properly assigned to another operator?~
3. After the withdrawal request is assigned to another operator, is it signed by all other operators, broadcast, and confirmed on signet?

We need to find a way to test number 3 above.

storopoli (changes requested):
Maybe we should create a proper parent class for tests that needs withdrawal/deposits that can have all the yet-to-move `utils.py` functions that I commented below in the NITs.

Then all the DRT, get balances, withdraw etc functionality is encapsulated in it and we can create proper concrete classes for the other tests that inherit from this parent class.

Sorry to push all this refactoring into your plate.
Since you're already editing these files it is convenient.
However, feel free to reject and open a ticket for this so that we can do it later.

delbonis:
I think it's reasonable to include the fix for [this ticket](https://alpenlabs.atlassian.net/browse/STR-781) in this PR too, since it's all related.

functional-tests/fn_bridge_reassignment.py
delbonis: Yeah I guess once we observe that it's been reassigned we can restart the operator we shut down and then get it signed.

functional-tests/fn_bridge_reassignment.py
delbonis: We discovered that doing it with just 2 breaks some stuff.  Created a ticket (see comment) for that.

functional-tests/fn_bridge_reassignment.py
```diff
@@ -0,0 +1,73 @@
...
+)
+
+
+@flexitest.register
+class BridgeWithdrawReassignmentTest(testenv.StrataTester):
+    """
+    Makes two DRT deposits, then triggers the withdrawal.
+    The bridge client associated with assigned operator id is stopped.
+    After the dispatch assignment duration is over,
+    Check if new operator is being assigned or not
```
bewakes, line 22: This is a great comment! Concise and clear.

functional-tests/fn_bridge_withdraw_happy.py
```diff
@@ -46,210 +42,292 @@ def __init__(self, ctx: flexitest.InitContext):
...
+
+
+def deposit_twice(ctx, web3, seqrpc, rethrpc, el_address, debug_fn=print):
+    """
+    Make two DRT deposits to ensure the EL address has enough funds for gas
+    and for subsequent withdrawals. Wait until the deposit is reflected on L2.
+    """
+    # Get bridge pubkey
+    bridge_pk = utils.get_bridge_pubkey(seqrpc)
+    if debug_fn:
```
bewakes, line 232: Can we not just omit this if clause? Looks annoying. The default value in the `debug_fn` argument indicates that this clause will always return true, unless `None` is passed(but this can be avoided by enforcing argument type).

Rajil1213: In that case, I second the approach by @bewakes i.e., annotating with `Callable[str, None]` perhaps and removing the `if` clause.

bin/bridge-client/src/modes/operator/task_manager.rs
```diff
@@ -54,23 +64,28 @@ where
...
+                }
+            } else {
+                error!(?duty_timeout_duration, "some duties timed out");
+            }
 
             // if none of the duties failed, update the duty index so that the
             // next batch is fetched in the next poll.
             //
             // otherwise, don't update the index so that the current batch is refetched and
             // ones that were not executed successfully are executed again.
```
Rajil1213, line 88: Need to move these comments on top of the `if !any_failed` check as well.

crates/bridge-exec/src/handler.rs
```diff
@@ -297,6 +297,9 @@ where
                 if let Ok(msg) = msg {
                     let payload = msg.payload();
                     let parsed_payload = borsh::from_slice::<Payload>(payload);
+                    let scope = msg.scope();
+                    let parsed_scope = borsh::from_slice::<Scope>(scope);
+                    debug!(msg=?msg, payload=?parsed_payload, scope=?parsed_scope, "ABC: got message from the L2 Client");
```
Rajil1213, line 302:
```suggestion
                    debug!(msg=?msg, payload=?parsed_payload, scope=?parsed_scope, "got message from the L2 Client");
```

functional-tests/fn_bridge_reassignment.py
```diff
@@ -0,0 +1,107 @@
...
+            new_assigned_op_idx = withdraw_duty["payload"]["assigned_operator_idx"]
+            new_assigned_operator = ctx.get_service(f"bridge.{new_assigned_op_idx}")
+            self.debug(f"new assigned operator is {new_assigned_op_idx}")
+            return new_assigned_operator
+
+        new_assigned_operator = wait_until_with_value(
+            my_val,
+            predicate=lambda v: v != assigned_operator,
+            timeout=10,
+        )
```
Rajil1213, line 89: I think this check can be removed now since the reassignment logic is supposed to always compute a different assignee than the current one (by computing offset).

functional-tests/fn_bridge_reassignment.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # Ensure a new operator is assigned
+        assert new_assigned_operator != assigned_operator, "No new operator was assigned"
+        assigned_operator.start()
+        abc = assigned_operator.create_rpc()
```
Rajil1213, line 94:
```suggestion
        bridge_rpc = assigned_operator.create_rpc()
```

functional-tests/fn_bridge_reassignment.py
```diff
@@ -0,0 +1,107 @@
...
+class BridgeWithdrawReassignmentTest(testenv.StrataTester):
+    """
+    Makes two DRT deposits, then triggers the withdrawal.
+    The bridge client associated with assigned operator id is stopped.
+    After the dispatch assignment duration is over,
+    Check if new operator is being assigned or not
+    """
+
+    def __init__(self, ctx: flexitest.InitContext):
+        # Example: we spin up 5 operators
```
Rajil1213, line 30:
```suggestion
        # Example: we spin up 3 operators
```

Actually, you can just remove this comment as it is very self-evident.

bin/bridge-client/src/constants.rs
```diff
@@ -4,5 +4,6 @@ pub(super) const DEFAULT_RPC_PORT: u32 = 4781; // first 4 digits in the sha256 o
 
 pub(super) const DEFAULT_RPC_HOST: &str = "127.0.0.1";
 
+pub(super) const DEFAULT_DUTY_TIMEOUT_DURATION: u64 = 65;
```
Rajil1213, line 7: Should increase this default value to at least 600. I know this is a temporary measure but this is to avoid breaking stuff if we ever deploy this change our signet.

crates/chaintsn/src/transition.rs
```diff
@@ -297,10 +297,13 @@ fn process_deposit_updates(
...
-                    let new_op_pos = if num_operators > 0 {
-                        let op_off = rng.next_u32() % (num_operators - 1);
-                        (dstate.assignee() + op_off) % num_operators
+                    let new_op_pos = if num_operators > 1 {
+                        // Compute a random offset from 1 to (num_operators - 1),
+                        // ensuring we pick a different operator than the current one.
+                        let offset = 1 + (rng.next_u32() % (num_operators - 1));
+                        (dstate.assignee() + offset) % num_operators
                     } else {
+                        // If there is 0 or 1 operator, we remain with the current assignee.
```
Rajil1213, line 306:
```suggestion
                        // If there is only a single operator, we remain with the current assignee.
```
There can't technically be 0 operators or `dstate.assignee()` wouldn't work.

functional-tests/fn_bridge_reassignment.py
```diff
@@ -0,0 +1,107 @@
...
+from utils import wait_until, wait_until_with_value
+
+
+@flexitest.register
+class BridgeWithdrawReassignmentTest(testenv.StrataTester):
+    """
+    Makes two DRT deposits, then triggers the withdrawal.
+    The bridge client associated with assigned operator id is stopped.
+    After the dispatch assignment duration is over,
+    Check if new operator is being assigned or not
```
Rajil1213, line 26:
```suggestion
    Check if new operator is being assigned or not
    Ensure that the withdrawal resumes and completes successfully.
```

Rajil1213 (changes requested):

crates/bridge-exec/src/handler.rs
delbonis: Also fix the formatting here.  If the name of the log field is the same as the variable being used for it, you can just omit the field name and make it `?msg` (or `%msg` for `Display` print).  In any case, there should be spaces around the `=`.  Rustfmt can't understand the macro conventions so it can't automatically fix this.

crates/chaintsn/src/transition.rs
delbonis: Yeah I think we refuse to start the client with 0 operators.

bin/bridge-client/src/args.rs
```diff
@@ -71,6 +71,9 @@ pub(crate) struct Cli {
         description = "retry count for the rocksdb database (default: 3)"
     )]
     pub retry_count: Option<u16>,
+
+    #[argh(option, description = "duty timeout duration in seconds (default: 65)")]
+    pub duty_timeout_duration: Option<u64>,
```
storopoli, line 76: Can I annoy you and ask you, since you're editing the `args.rs` and this might be fresh on your mind, to add docstring to all of these fields? The `argh` macros help but they don't provide LSP introspection in the editor/IDE.

functional-tests/fn_bridge_reassignment.py
voidash: Total block production control is not required. The event is triggered after 64 blocks and not at exact 64 blocks so i think leaving it to auto generate will work

functional-tests/fn_bridge_reassignment.py
voidash: The bug has been fixed and now this is also being checked

functional-tests/fn_bridge_withdraw_happy.py
voidash: These functions with `debug_fn` are no longer valid and has been covered by `BridgeTestBase` which has access to self.debug itself

bin/bridge-client/src/args.rs
voidash: Done

functional-tests/fn_bridge_reassignment.py
voidash: I think we don't have environment with already balances in strata.

bin/bridge-client/src/args.rs
```diff
@@ -9,73 +10,104 @@ use crate::errors::InitError;
...
 
+    /// Path to the directory where RocksDB databases are stored.
+    /// Defaults to `$HOME/.local/share/strata/` if not specified.
     #[argh(
         option,
         description = "path to the directory where to store the rocksdb databases (default: $HOME/.local/share/strata/)"
     )]
     pub datadir: Option<String>,
 
+    /// Master operator's key in xpriv format. Defaults to the environment variable
```
Rajil1213, line 29:
```suggestion
    /// Master operator key in xpriv format. Defaults to the environment variable
```

functional-tests/fn_bridge_reassignment.py
```diff
@@ -0,0 +1,106 @@
...
+        assigned_op_idx = withdraw_duty["payload"]["assigned_operator_idx"]
+        assigned_operator = ctx.get_service(f"bridge.{assigned_op_idx}")
+        self.debug(f"Assigned operator index: {assigned_op_idx}")
+
+        # Stop assigned operator
+        self.debug("Stopping assigned operator ...")
+        assigned_operator.stop()
+
+        # Let enough blocks pass so the assignment times out
+        # (64 is the dispatch_assignment_duration)
```
Rajil1213, line 77: This comment is no longer necessary, I think because `64` has been moved to a constant.

functional-tests/testenv.py
```diff
@@ -45,6 +58,161 @@ def __init__(self, datadir_root: str, name: str, env: flexitest.LiveEnv):
         super().__init__(env)
 
 
+class BridgeTestBase(StrataTester):
+    """
+    Testbase for bridge specific test.
+    Provides methods for setting up service, making DRT, withdraw transaction
+    """
+
+    def premain(self, ctx: flexitest.RunContext):
```
Rajil1213, line 67: I can't find any usages for this method. Is there some place where this is being called implicitly?

Rajil1213 (changes requested):
Some final nits and a question.

functional-tests/testenv.py
voidash: Yes, `premain(ctx)` is part of the flexitest `TestRuntime`.
https://codeberg.org/treyd/flexitest/src/commit/216e87270994ebb023ba9b0c1a38134cc59da1f7/flexitest/runtime.py#L164

storopoli: TIL

storopoli (dismissed):
ACK 5eeff365c68b526d6c176cac1e9ad07e5994bd42

bin/bridge-client/src/modes/operator/bootstrap.rs
```diff
@@ -148,9 +150,17 @@ pub(crate) async fn bootstrap(args: Cli) -> anyhow::Result<()> {
         Duration::from_millis,
     );
 
+    let duty_timeout_duration = args.duty_timeout_duration.map_or(
+        Duration::from_secs(DEFAULT_DUTY_TIMEOUT_DURATION),
+        Duration::from_secs,
+    );
```
delbonis, line 156: `Duration::from_secs(stuff.unwrap_or(DEFAULT_DUTY_TIMEOUT_DURATION))`

bin/bridge-client/src/constants.rs
```diff
@@ -4,5 +4,6 @@ pub(super) const DEFAULT_RPC_PORT: u32 = 4781; // first 4 digits in the sha256 o
 
 pub(super) const DEFAULT_RPC_HOST: &str = "127.0.0.1";
 
+pub(super) const DEFAULT_DUTY_TIMEOUT_DURATION: u64 = 600;
```
delbonis, line 7: Units?

I can infer that this is seconds, so it should be named `DEFAULT_DUTY_TIMEOUT_SEC`.  We can leave out the `DURATION` here since it's a timeout and it should be obvious from context.

crates/bridge-exec/src/handler.rs
```diff
@@ -297,6 +297,9 @@ where
                 if let Ok(msg) = msg {
                     let payload = msg.payload();
                     let parsed_payload = borsh::from_slice::<Payload>(payload);
+                    let scope = msg.scope();
+                    let parsed_scope = borsh::from_slice::<Scope>(scope);
+                    debug!(msg = ?msg, payload = ?parsed_payload, scope = ?parsed_scope, "got message from the L2 Client");
```
delbonis, line 302: `?msg`

Really I would do

* payload -> raw_payload
* parsed_payload -> payload

And the same for the scope so that you can also avoid the extra `=` and make it simpler in the consumer.

delbonis (changes requested):

functional-tests/logs
delbonis, line 1: Don't forget to delete.

bin/bridge-client/src/modes/operator/task_manager.rs
```diff
@@ -54,22 +64,26 @@ where
...
+                    if let Err(e) = self
+                        .bridge_duty_idx_db_ops
+                        .set_index_async(stop_index)
+                        .await
+                    {
+                        error!(error = %e, %start_index, %stop_index, "could not update duty index");
+                    }
                 }
+            } else {
+                error!(?duty_timeout_duration, "some duties timed out");
```
delbonis, line 86: Can use % here, although I'm not sure how useful it is to even have this here.

crates/bridge-exec/src/handler.rs
```diff
@@ -295,13 +295,16 @@ where
...
+                    let raw_scope = msg.scope();
+                    let scope = borsh::from_slice::<Scope>(raw_scope);
+                    debug!(?msg, ?payload, ?scope , "got message from the L2 Client");
 
-                    match parsed_payload {
+                    match payload {
                         Ok(payload) => Some((msg.source_id(), payload)),
                         Err(err) => {
-                            warn!(?scope, ?payload, error=?err, "skipping faulty message payload");
+                            warn!(?scope, ?raw_payload, error=?err, "skipping faulty message payload");
```
delbonis, line 307: If you're able to parse the payload here then you'd want to print the parse dpayload instead of the raw bytes.

Also fix `=?` formatting.

functional-tests/testenv.py
delbonis, line 1: Everywhere here that this is doing `svc.props["key"]` should be `svc.get_prop("key")`.  I should really update that field to have a `_` prefix.

delbonis (changes requested):

functional-tests/testenv.py
delbonis: Yeah this is somewhat new.

bin/bridge-client/src/modes/operator/task_manager.rs
voidash: Duration doesn't have default formatter. But i think removing this makes sense as we don't exactly track which duty got timed out

Rajil1213 (approved):

bewakes (approved):

storopoli (approved):
ACK 27bd0294b7bd001e84a4ae335ecdebdf52f73ac4

PR #558, opened 2024-12-23
STR-782: add a warning about partial MuSig2 verification

AaronFeickert:
## Description

This PR updates the documentation for partial MuSig2 signature verification to caution against assuming unforgeability of partial signatures.

### Type of Change

- [x] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/bridge-sig-manager/src/operations.rs
+++ b/crates/bridge-sig-manager/src/operations.rs
@@ -113,6 +113,21 @@ pub fn sign_state_partial(
 }
 
 /// Verify that a partial MuSig2 signature is correct.
+///
+/// # Warning
+///
+/// It is possible for an adversary to forge a partial signature under certain conditions.
+/// You _must not_ rely on this verification to imply that a partial signature is unforgeable.
+///
+/// See the footnote in section 4.2 of <https://eprint.iacr.org/archive/2020/1261/20231020:081535> or
+/// <https://gist.github.com/AdamISZ/ca974ed67889cedc738c4a1f65ff620b> for details.
+///
+/// It is _only_ safe to assume that:
+///
+/// - failed verification of _any_ partial signature represents a fault, and the signing session
+///   should be aborted
+/// - successful verification of _all_ partial signatures for a signing session will result in a
+///   successful signing session
 pub fn verify_partial_sig(
     tx_state: &BridgeTxState,
     signature_info: &OperatorPartialSig,


crates/bridge-sig-manager/src/operations.rs
```diff
@@ -113,6 +113,14 @@ pub fn sign_state_partial(
 }
 
 /// Verify that a partial MuSig2 signature is correct.
+///
+/// WARNING: It is possible for an adverary to forge a partial signature under certain conditions.
```
Rajil1213, line 117: Could you link to any relevant resources for what these conditions are? This is both for this documentation and for myself. 😅

Rajil1213 (changes requested):

crates/bridge-sig-manager/src/operations.rs
AaronFeickert: Sure, I'll add that. It's in a footnote of section 4.2 in [the preprint](https://eprint.iacr.org/2020/1261), which links out [here](https://gist.github.com/AdamISZ/ca974ed67889cedc738c4a1f65ff620b).

AaronFeickert: I also need to change the statement about failed partial verification guaranteeing a failed signing session. It's possible for an adversary to manipulate multiple (otherwise valid) partial signatures such that each will fail, but the signing session will not. It's something of a weird edge case.

storopoli: Maybe we should move these to H1? The same way we suggest here: https://www.notion.so/Engineering-Rust-Best-Practices-14c901ba000f8025a974ec8e2a70b4af?pvs=4#14d901ba000f8020aebbc58bee7c0a4c

as in:

```rust
/// Foo bla bla bla
///
/// # Warning
///
/// It is possible for an adversary to forge
```

And how did `codespell` not got this `adverary`?

crates/bridge-sig-manager/src/operations.rs
```diff
@@ -113,6 +113,20 @@ pub fn sign_state_partial(
...
+/// You _must not_ rely on this verification to imply that a partial signature is unforgeable.
+///
+/// See the footnote in section 4.2 of <https://eprint.iacr.org/archive/2020/1261/20231020:081535> or
+/// <https://gist.github.com/AdamISZ/ca974ed67889cedc738c4a1f65ff620b> for details.
+///
+/// It is _only_ safe to assume that:
+/// - failed verification of _any_ partial signature represents a fault, and the signing session
+///   should be aborted
+/// - successful verification of _all_ partial signatures for a signing session will result in a
+///   successful signing session
```
storopoli, line 129: Just a final nit. Empty line around lists
```suggestion
///
/// - failed verification of _any_ partial signature represents a fault, and the signing session
///   should be aborted
/// - successful verification of _all_ partial signatures for a signing session will result in a
///   successful signing session
```

storopoli (changes requested):

storopoli (approved):
ACK afd48f89482dca390ba709dcc6fcfae90aa88c3b

Rajil1213 (approved):

PR #560, opened 2024-12-24
block template manager and rpc

sapinb:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

STR-945
STR-946
STR-948


bin/strata-client/src/main.rs
```diff
@@ -408,7 +409,7 @@ fn start_sequencer_tasks(
         engine,
         bitcoin_client,
         ..
```
delbonis, line 411: Don't really like how this is being cloned but I do like the idea of passing it to the thing that uses it on 449.

Could this be reworked to repeat the pattern with the other launcher fns?

delbonis (changes requested):
Broad strokes looks good and cleanly written.  There's a lot of stuff in here that ideally should have some unit tests, although I realize that's hard with the stuff that's being ran here.

Also, does it really make sense to use L2BlockId as the template ID?  I know it works out but it feels like it's not quite right and there should be a distinct thing that refers to templates as they're being constructed.  Not sure.

bin/strata-client/src/rpc_server.rs
```diff
@@ -791,6 +814,94 @@ impl StrataSequencerApiServer for SequencerServerImpl {
...
+        if !verify_checkpoint_sig(&signed_checkpoint, &self.params) {
+            Err(Error::InvalidCheckpointSignature(idx))?;
+        }
+
+        let payload = borsh::to_vec(&signed_checkpoint).map_err(|e| Error::Other(e.to_string()))?;
+        let sighash = signed_checkpoint.checkpoint().get_sighash();
+
+        let blob_intent = BlobIntent::new(BlobDest::L1, sighash, payload);
+        self.inscription_handle
+            .submit_intent_async(blob_intent)
```
delbonis, line 899: Submitting the checkpoint to L1 should be done somewhere else after we've saved it durably.  There should be something that tracks the status of these outstanding payloads.  If we've saved the signed checkpoint but we don't see a writer entry for it then we should create it.

This will have to be updated sooner or later to use the `Payload`-oriented terminology since "blob" refers more to DA.

bin/strata-client/src/rpc_server.rs
```diff
@@ -791,6 +814,94 @@ impl StrataSequencerApiServer for SequencerServerImpl {
...
+        {
+            Err(Error::CheckpointAlreadyPosted(idx))?;
+        }
+
+        let checkpoint = BatchCheckpoint::from(entry);
+        let signed_checkpoint = SignedBatchCheckpoint::new(checkpoint, sig.0.into());
+
+        if !verify_checkpoint_sig(&signed_checkpoint, &self.params) {
+            Err(Error::InvalidCheckpointSignature(idx))?;
+        }
```
delbonis, line 892: Missing `return`s?

crates/sequencer/src/block_template/error.rs
```diff
@@ -0,0 +1,17 @@
...
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("unknown blockId: {0}")]
+    UnknownBlockId(L2BlockId),
+    #[error("invalid signature: {0}, {1}")]
+    InvalidSignature(L2BlockId, Buf64),
+    #[error("{0}")]
+    DbError(#[from] DbError),
+    #[error("{0}")]
+    ConsensusError(#[from] strata_consensus_logic::errors::Error),
```
delbonis, line 14: Should have `db:` and `consensus:` prefixes.

crates/sequencer/src/block_template/error.rs
```diff
@@ -0,0 +1,17 @@
+use strata_db::DbError;
+use strata_primitives::{buf::Buf64, l2::L2BlockId};
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("unknown blockId: {0}")]
+    UnknownBlockId(L2BlockId),
+    #[error("invalid signature: {0}, {1}")]
+    InvalidSignature(L2BlockId, Buf64),
```
delbonis, line 10: Would make these

* `unknown block id {0}`
* `invalid signature for block {0}`

and leave it at that without actually printing the sig.  It's not really useful as something printed and makes the logs more verbose.  It's still in the type if we want to work with it internally.

crates/sequencer/src/block_template/handle.rs
delbonis, line 1: See earlier comment about having some helper fn that handles submitting the message on the channel and waiting for a response so the send/receive logic doesn't have to be duplicated in each of the high level functions.

crates/sequencer/src/block_template/manager.rs
```diff
@@ -0,0 +1,119 @@
...
+            status_channel,
+            pending_templates: HashMap::new(),
+            pending_by_parent: HashMap::new(),
+        }
+    }
+
+    pub(crate) fn insert_template(&mut self, block_id: L2BlockId, template: BlockTemplateFull) {
+        let parent = *template.header().parent();
+        self.pending_templates.insert(block_id, template);
+        self.pending_by_parent.insert(parent, block_id);
```
delbonis, line 60: Maybe this should print a warning when evicting a child?  Or do some other handling if we're doing something weird with conflicting templates so we at least can know if it's happening.

crates/sequencer/src/block_template/manager.rs
```diff
@@ -0,0 +1,119 @@
...
+    completion: &BlockCompletionData,
+) -> bool {
+    let sighash = header.get_sighash();
+    match &rollup_params.cred_rule {
+        CredRule::Unchecked => true,
+        CredRule::SchnorrKey(pubkey) => {
+            verify_schnorr_sig(completion.signature(), &sighash, pubkey)
+        }
+    }
+}
```
delbonis, line 119: This should probably call out to the actual validation logic instead of reimplementing it here.

crates/sequencer/src/errors.rs
```diff
@@ -0,0 +1,105 @@
+//! common error types for sequencer duty
+
+use strata_chaintsn::errors::TsnError;
+use strata_eectl::errors::EngineError;
+use strata_state::{id::L2BlockId, l1::L1BlockId};
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum Error {
```
delbonis, line 9: I order these as with:

* firstparty error variants we expect to generate
* wrapper variants for errors from dependencies
* `Unimplemented`
* `Other(String)` and similar

This way we don't mix in interesting errors that we're looking for with stuff that's just wrappers.

functional-tests/tests/cl_checkpoint.py
```diff
@@ -24,10 +24,13 @@ def main(self, ctx: flexitest.RunContext):
...
-        ckp_idx = seqrpc.strata_getLatestCheckpointIndex()
-        assert ckp_idx is not None
+        ckp_idx = wait_until_with_value(
+            lambda: seqrpc.strata_getLatestCheckpointIndex(),
+            predicate=lambda idx: idx is not None,
+            error_with="Checkpoint was not generated in time",
+        )
 
-        self.debug(f"checkpoint: {ckp_idx} finalized")
+        self.debug(f"checkpoint: {ckp_idx} found")
```
delbonis, line 33: Just make this `info`.

functional-tests/factory/factory.py
```diff
@@ -127,6 +124,39 @@ def create_sequencer(
...
+            "strata-sequencer-client",
+            "--sequencer-key", seqkey_path,
+            "--rpc-host", sequencer_rpc_host,
+            "--rpc-port", str(sequencer_rpc_port),
+        ]
+        # fmt: on
+
+        props = {
+            "seqkey": seqkey_path,
+        }
```
delbonis, line 153: Expose RPCs data here so we can call it?

crates/sequencer/src/worker.rs
```diff
@@ -0,0 +1,177 @@
+//! Executes duties.
+
+use std::{
+    sync::Arc,
+    time::{self},
```
delbonis, line 5: Weird import.

delbonis (changes requested):
I think this got rebased so the github comments got confused.  Redid review on new content, maybe duplicated some comments.

crates/block-assembly/src/handle.rs
```diff
@@ -0,0 +1,79 @@
...
+
+#[derive(Debug, Clone)]
+pub struct TemplateManagerHandle {
+    tx: mpsc::Sender<TemplateManagerRequest>,
+}
+
+impl TemplateManagerHandle {
+    pub fn new(tx: mpsc::Sender<TemplateManagerRequest>) -> Self {
+        Self { tx }
+    }
```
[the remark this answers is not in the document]
sapinb: handle has to be pub as its created in the strata-client bin.

functional-tests/factory/factory.py
sapinb: signer does not serve any rpcs. Its purely does signing and is stateless right now.

crates/sequencer/src/worker.rs
sapinb: this should have been autocorrected 😄

bin/strata-client/src/rpc_server.rs
sapinb: `?` used on the Err should be equivalent to returning it. Should be equivalent to 

`return Err(Error::MissingCheckpointProof(idx).into());`

crates/sequencer/src/block_template/error.rs
sapinb: updated

crates/sequencer/src/errors.rs
sapinb: There were a lot of unused errors here. This has been cleaned up.

crates/sequencer/src/block_template/manager.rs
sapinb: updated.

bin/strata-client/src/rpc_server.rs
sapinb: Currently, we are depending on the L1 writer to ensure durability, else the sequencer will receive signing duty for same checkpoint again and provide another signature. 
The durably saving part should likely come on sequencer side to prevent signing duplicate/conflicting checkpoints ?

crates/block-assembly/src/error.rs
```diff
@@ -0,0 +1,17 @@
...
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("unknown blockId: {0}")]
+    UnknownBlockId(L2BlockId),
+    #[error("invalid signature: {0}, {1}")]
+    InvalidSignature(L2BlockId, Buf64),
+    #[error("{0}")]
+    DbError(#[from] DbError),
+    #[error("{0}")]
+    ConsensusError(#[from] strata_consensus_logic::errors::Error),
```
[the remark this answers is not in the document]
sapinb: updated

crates/block-assembly/src/types.rs
```diff
@@ -0,0 +1,98 @@
...
+    pub fn new(header: L2BlockHeader, body: L2BlockBody, accessory: L2BlockAccessory) -> Self {
+        Self {
+            header,
+            body,
+            accessory,
+        }
+    }
+
+    pub fn block_id(&self) -> L2BlockId {
+        self.header.get_blockid()
```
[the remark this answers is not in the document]
sapinb: updated

crates/block-assembly/src/worker.rs
```diff
@@ -0,0 +1,105 @@
...
+    while let Some(request) = rx.recv().await {
+        match request {
+            TemplateManagerRequest::GenerateBlockTemplate(config, sender) => {
+                let _ = sender.send(generate_block_template(&mut manager, config));
+            }
+            TemplateManagerRequest::CompleteBlockTemplate(template_id, completion, sender) => {
+                let _ = sender.send(manager.complete_block_template(template_id, completion));
+            }
+            TemplateManagerRequest::GetBlockTemplate(block_id, sender) => {
+                let _ = sender.send(manager.get_block_template(block_id));
```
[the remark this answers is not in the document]
sapinb: added warning logs if this errors. 
Didnt get the `completion` instead of `sender` part though.

crates/block-assembly/src/types.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+impl BlockGenerationConfig {
+    pub fn parent_block_id(&self) -> L2BlockId {
+        self.parent_block_id
+    }
+
+    pub fn ts(&self, default: u64) -> u64 {
+        self.ts.unwrap_or(default)
+    }
```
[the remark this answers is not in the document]
sapinb: updated

crates/block-assembly/src/types.rs
```diff
@@ -0,0 +1,98 @@
...
+    // ..
+}
+
+#[derive(Debug, Clone)]
+pub struct BlockTemplateFull {
+    header: L2BlockHeader,
+    body: L2BlockBody,
+    accessory: L2BlockAccessory,
+}
+
```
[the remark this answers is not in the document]
sapinb: updated

crates/block-assembly/src/handle.rs
```diff
@@ -0,0 +1,79 @@
...
+    ) -> Result<BlockTemplate, Error> {
+        let (tx, rx) = oneshot::channel();
+        self.tx
+            .send(TemplateManagerRequest::GenerateBlockTemplate(config, tx))
+            .await
+            .map_err(|_| Error::ChannelError("send"))?;
+
+        match rx.await {
+            Err(_) => Err(Error::ChannelError("oneshot tx dropped")),
+            Ok(res) => res,
```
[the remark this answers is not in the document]
sapinb: added more descriptive message

crates/sequencer/src/block_template/error.rs
sapinb: updated.

crates/sequencer/src/block_template/handle.rs
sapinb: updated

crates/block-assembly/src/handle.rs
```diff
@@ -0,0 +1,79 @@
...
+        }
+    }
+
+    pub async fn get_block_template(&self, template_id: L2BlockId) -> Result<BlockTemplate, Error> {
```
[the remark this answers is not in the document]
delbonis: This isn't what I had in mind, the main thing we want to avoid is the worker thread having to wake up unless it has to, which it has to do for a lot with this call/response style pattern.  The goal would involve removing this `GetBlockTemplate` variant.

Very loose skeleton of the structure:

```rs
struct Shared {
    templates: HashMap<SomeKey, Template>
}

struct Handle {
    shared: Arc<Mutex<Shared>>
    tx: ...
}

fn worker(rx: ..., shared: Arc<Shared>) {
    // when finish new template, put into shared.templates
}
```

Then we can always check the templates table first for the ready templates before sending of a new request.  Maybe I shouldn't have used the term "cache".

crates/sequencer/src/block_template/handle.rs
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+
+    async fn request<R>(
+        &self,
+        build_request: impl FnOnce(oneshot::Sender<Result<R, Error>>) -> TemplateManagerRequest,
+    ) -> Result<R, Error> {
+        let (tx, rx) = oneshot::channel();
+        self.tx.send(build_request(tx)).await.map_err(|_| {
+            Error::ChannelError("failed to send request to template manager, worker likely exited")
```
delbonis, line 44: This always only happens if the channel is closed.  Just make this its own error variant.

crates/sequencer/src/block_template/handle.rs
```diff
@@ -0,0 +1,101 @@
...
+        })?;
+
+        match rx.await {
+            Ok(res) => res,
+            // oneshot tx is dropped
+            Err(_) => Err(Error::ChannelError("tx dropped, worker likely exited")),
```
delbonis, line 50: Same here.

delbonis (changes requested):

crates/block-assembly/src/worker.rs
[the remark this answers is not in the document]
delbonis: The term "completion" is the more generally used term for what channel represents in this kind of pattern, using it instead of just "sender" makes it more clear what's happening.

I realize that conflicts with the var named "completion" here already, might have to rework that.

bin/strata-client/src/rpc_server.rs
delbonis: Oh I didn't see that.  Yeah that's a bit unclear.  It's fine.

delbonis:
The main issue is just fastpathing requests for already-built templates to avoid waking up the template manager thread.

crates/block-assembly/src/handle.rs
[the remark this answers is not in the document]
sapinb: updated

crates/block-assembly/src/worker.rs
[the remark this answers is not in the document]
sapinb: opting for response here

crates/sequencer/src/block_template/handle.rs
sapinb: updated

crates/sequencer/src/block_template/handle.rs
sapinb: updated

bewakes (commented):
Looks good overall. Some nits, mostly around documents and commented out codes.

Do you think the types in the `sequencer` crate could possibly be moved to primitives? Because the `strata-sequencer-client` depends on the `sequencer` crate which also has `strata-client` specific worker/handle modules, which the `sequencer-client` has nothing to do with.

bin/strata-sequencer-client/src/main.rs
```diff
@@ -0,0 +1,135 @@
...
+        duty_executor_worker(rpc, duty_rx, handle.clone(), idata),
+    );
+
+    task_manager.start_signal_listeners();
+    task_manager.monitor(Some(Duration::from_millis(SHUTDOWN_TIMEOUT_MS)))?;
+
+    Ok(())
+}
+
+async fn duty_fetcher_worker<R>(
```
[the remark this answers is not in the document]
sapinb: moved this out to separate file

crates/sequencer/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+parking_lot.workspace = true
+serde.workspace = true
+thiserror.workspace = true
+tokio.workspace = true
+tracing.workspace = true
+
+[dev-dependencies]
+# rand = { version = "0.8.5", features = ["getrandom", "std_rng"] }
+# strata-db = { workspace = true, features = ["stubs"] }
+# strata-rocksdb = { workspace = true, features = ["test_utils"] }
```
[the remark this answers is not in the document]
sapinb: removed

crates/sequencer/src/block_template/handle.rs
```diff
@@ -0,0 +1,83 @@
+use strata_primitives::l2::L2BlockId;
+use strata_state::block::L2BlockBundle;
+use tokio::sync::{mpsc, oneshot};
+
+use super::{BlockCompletionData, BlockGenerationConfig, BlockTemplate, Error, SharedState};
+
+#[derive(Debug)]
+pub enum TemplateManagerRequest {
```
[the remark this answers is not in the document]
sapinb: updated.

bin/strata-sequencer-client/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+jsonrpsee = { workspace = true, features = ["server", "macros"] }
+serde.workspace = true
+thiserror.workspace = true
+tokio.workspace = true
+tracing.workspace = true
+zeroize.workspace = true
+
+[dev-dependencies]
+# arbitrary.workspace = true
+# strata-test-utils.workspace = true
```
storopoli, line 33: Why these are commented? Do we plan to use them in this PR? Or is this a leftover?

crates/sequencer/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+strata-test-utils.workspace = true
+
+[lints]
+rust.missing_debug_implementations = "warn"
+# rust.missing_docs = "warn"
+rust.rust_2018_idioms = { level = "deny", priority = -1 }
+rust.unreachable_pub = "warn"
+rust.unused_crate_dependencies = "deny"
+rust.unused_must_use = "deny"
+# rustdoc.all = "warn"
```
storopoli, line 35: I dare you uncommenting these docs lints 😂.

storopoli (changes requested):
Minor nits for now.

bin/strata-sequencer-client/Cargo.toml
sapinb: Forgot to remove them. Thanks for the catch.

crates/sequencer/Cargo.toml
sapinb: Updated the docs where possible.

bin/strata-sequencer-client/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+edition = "2021"
+name = "strata-sequencer-client"
```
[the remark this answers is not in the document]
sapinb: added new crates to CODEOWNERS

crates/sequencer/Cargo.toml
```diff
@@ -24,12 +24,3 @@ tracing.workspace = true
...
 strata-test-utils.workspace = true
-
-[lints]
-rust.missing_debug_implementations = "warn"
-# rust.missing_docs = "warn"
-rust.rust_2018_idioms = { level = "deny", priority = -1 }
-rust.unreachable_pub = "warn"
-rust.unused_crate_dependencies = "deny"
-rust.unused_must_use = "deny"
-# rustdoc.all = "warn"
```
storopoli: I mean these lints are good.
You can leave the missing_docs stuff commented out.
Can you undo this change?
Or are they already enforced at workspace level?

Cc @Rajil1213

Rajil1213: Yes, definitely in favor of these rustc lints (they cannot be done at the workspace level, I think).

That said, we should put it in our engineering guidelines to enable these for all new crates. For existing ones, we already have a ticket: https://alpenlabs.atlassian.net/browse/STR-198 since that would require adding documentation to all the code which is a massive undertaking.

In this case, is there a reason we are explicitly opting out of these lints if these were already enabled?

Rajil1213: If these were missed in some other PR and re-enabling these would require adding docs all over, I think it's fine to just remove the docs lints for now and add it to the pile in the other ticket as Jose is suggesting.

storopoli: > I think it's fine to just remove the docs lints for now and add it to the pile in the other ticket as Jose is suggesting.

Yeah just undo the diff I've highlighted

.github/CODEOWNERS
```diff
@@ -43,12 +44,13 @@ tests/                 @alpenlabs/rust @alpenlabs/ci
...
-bin/prover*/        @alpenlabs/bin @alpenlabs/prover
-bin/strata-client/  @alpenlabs/bin @alpenlabs/bin-strata-client
-bin/strata-reth/    @alpenlabs/bin @alpenlabs/evm
+bin/                          @alpenlabs/bin                   # catch all
+bin/bridge-client/            @alpenlabs/bin @alpenlabs/bridge
+bin/datatool/                 @alpenlabs/bin @alpenlabs/bin-datatool
+bin/prover*/                  @alpenlabs/bin @alpenlabs/prover
+bin/strata-client/            @alpenlabs/bin @alpenlabs/bin-strata-client
+bin/strata-reth/              @alpenlabs/bin @alpenlabs/evm
+bin/strata-sequencer-client/  @alpenlabs/bin @alpenlabs/bin-strata-client
```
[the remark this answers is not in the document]
sapinb: sequencer-client is split off from strata-client and keeping it under same owners should be good for now

crates/sequencer/Cargo.toml
sapinb: updated.

Rajil1213 (changes requested):
Do these changes also affect the files in `docker/`? If so, could we update those as well.  
I personally use it for running the bridge node locally but there may be other use cases that I'm not aware of.

delbonis (dismissed):

.github/CODEOWNERS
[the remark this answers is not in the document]
storopoli: Ok!

storopoli (dismissed):
ACK cc943eb011df24b012e0a54a2c3aae51bd6f62d2

bewakes:
Needs a rebase.

storopoli (dismissed):
ACK 33622c221d5f1ee60014bfd55a671a2a03c8eefe

delbonis:
What's still broken here?

storopoli:
> What's still broken here?

Might be a flaky one. Re-triggered the functional tests.

storopoli:
func tests are still failing and also the `cargo-audit`.

delbonis (dismissed):

voidash (dismissed):

delbonis (approved):

sapinb:
functional tests fixed. 
cargo audit issue will be resolved with reth update which should be done separately.

delbonis:
Looking at the flaky tests is looks like some of them have just *always* been flaking.

delbonis:
Merging now to keep the ball rolling.

PR #561, opened 2024-12-25
Str 786 main refactor part1

bewakes:
## Description
This is the part 1 of main refactor which moves around a couple of things:
1. Move db init to rocksdb-store crate
2. Move `Config` to `crates/config` as it is where it should be. Corresponding minor changes to `ReaderConfig` and `Args` methods.
3. Adds `NodeStorage`, which at the moment consolidates two managers. It will get extended and included in the context in future.
4. Expose `Handle` from `TaskManager` and use it instead of passing around `runtime` everywhere.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
This is a part of bigger refactor that lies ahead and I kindly request to review this as soon as possible.

## Checklist


bin/strata-client/src/args.rs
```diff
@@ -57,3 +62,114 @@ pub struct Args {
...
 }
+
+impl Args {
+    pub fn derive_config(&self) -> Result<Config, String> {
+        let args = self.clone();
+        Ok(Config {
+            bitcoind_rpc: BitcoindConfig {
+                rpc_url: args
+                    .bitcoind_host
+                    .ok_or_else(|| "args: no bitcoin --rpc-url provided".to_string())?,
```
delbonis, line 73: I would make this a toplevel function or even like `from_args` on `Config`, it doesn't really make sense to be a member function on `Args`.  This also makes some sense to use `anyhow` instead of manipulating string error messages.

bin/strata-client/src/args.rs
```diff
@@ -57,3 +62,114 @@ pub struct Args {
...
+            relayer: RelayerConfig {
+                refresh_interval: 10,
+                stale_duration: 120,
+                relay_misc: true,
+            },
+        })
+    }
+
+    pub fn update_config(&self, config: &mut Config) {
+        let args = self.clone();
```
delbonis, line 135: Same here, doesn't make sense to be a member function on `Args`.

bin/strata-client/src/args.rs
```diff
@@ -57,3 +62,114 @@ pub struct Args {
...
+                max_reorg_depth: 4,
+                client_poll_dur_ms: 200,
+                client_checkpoint_interval: 10,
+            },
+            exec: ExecConfig {
+                reth: RethELConfig {
+                    rpc_url: args.reth_authrpc.unwrap_or("".to_string()), // TODO: sensible default
+                    secret: args.reth_jwtsecret.unwrap_or_default(),      /* TODO: probably
+                                                                           * secret should be
+                                                                           * Option */
```
delbonis, line 123: There is no sensible default here and we *always* need a secret.  We're not at the point where we can run without reth, so make these required.

crates/storage/src/managers/mod.rs
```diff
@@ -1,2 +1,37 @@
...
 pub mod checkpoint;
 pub mod l2;
+
+/// A consolidation of database managers and ops. Ideally ops should not be exposed.
+#[derive(Clone)]
+pub struct DbManager {
+    l2_block_manager: Arc<l2::L2BlockManager>,
+    checkpoint_manager: Arc<checkpoint::CheckpointDbManager>,
+}
+impl DbManager {
```
delbonis, line 14: Not really a manager since it's just a dumb container for other data.  It makes sense to put this all under an umbrella type something like "`NodeStorageInterface`" and not in the `managers` module.

Should also have a separate one of these for the sequencer databases.

crates/storage/src/managers/mod.rs
```diff
@@ -1,2 +1,37 @@
...
+    pub fn l2(&self) -> Arc<l2::L2BlockManager> {
+        self.l2_block_manager.clone()
+    }
+
+    pub fn checkpoint(&self) -> Arc<checkpoint::CheckpointDbManager> {
+        self.checkpoint_manager.clone()
+    }
```
delbonis, line 21: Don't force these clones.

crates/rocksdb-store/src/lib.rs
```diff
@@ -99,3 +113,72 @@ impl DbOpsConfig {
...
+        STORE_COLUMN_FAMILIES.iter().map(|s| s.to_string()),
+        &opts,
+    )
+    .context("opening database")?;
+
+    Ok(Arc::new(rbdb))
+}
+
+pub type CommonDb =
+    CommonDatabase<L1Db, L2Db, SyncEventDb, ClientStateDb, ChainstateDb, RBCheckpointDB>;
```
delbonis, line 145: I would like to get rid of this `CommonDatabase` abstraction since it's not useful anymore.  But since there's a lot of stuff still using the `Database` trait I'm not sure how feasible that is yet.

crates/common/src/config.rs
delbonis, line 1: This crate is for things that are common to every possible service, like logging.  This configuration is specific to the strata client.  Do you think it makes sense move it to a new `node` crate like Reth has and shift more init logic there?

delbonis (changes requested):

crates/storage/src/managers/mod.rs
bewakes: What do you suggest? Returning refs?

bin/strata-client/src/args.rs
bewakes: The idea behind this was that Args seems to be more inside the bin whereas Config seems somewhere common which other crates can/do use. To me, it doesn't make sense to pull in Arg dependency where config is defined.

I think config should not be in bin. It is just being used everywhere so it makes sense to keep it somewhere common and it should not have dependency on Args.

bin/strata-client/src/args.rs
bewakes: Same reasoning as above.

bin/strata-client/src/args.rs
bewakes: Oh these, I haven't had a change to go through in detail. These were just being moved around. Will make it required.

crates/storage/src/managers/mod.rs
bewakes: We need a good name for the container of similar items. :)

crates/rocksdb-store/src/lib.rs
bewakes: The feasibility is yet to be assessed. Once things are in place where they should be,  we can then get rid of the stuffs.

crates/common/src/config.rs
bewakes: Yes, I think that would be better. I've mentioned the rationale behind this in one of the comments above.

I am also working on a new crate which I've named as "client" so far. But I do like the name "node" more.

crates/storage/src/managers/mod.rs
bewakes: I could just return `&Arc<...>`

crates/storage/src/managers/mod.rs
bewakes: But as I think more about it, I think it is fine. It is something similar to the `Database` trait isn't it?

crates/common/src/config.rs
bewakes: So yeah, let's keep `Config` under `common` as it is fairly serves its purpose(being "common" to almost all other crates in some ways). Can move to node crate in the subsequent PRs.

crates/storage/src/managers/mod.rs
delbonis: Yeah that's the most straightforward way to do it.

crates/common/src/config.rs
delbonis: No it doesn't serve its purpose.  That's just *not* what the crate is intended for.

bin/strata-client/src/args.rs
delbonis: If we're separating the actual *service* config from the inner *client*/*node* config internally then yeah moving it apart makes sense.

But how these fns are organized here does feel weird to me.

crates/storage/src/managers/mod.rs
delbonis: Structurally yes, but it's not serving a comparable role to make sense to be called that the `Database`-related traits aren't managers either, they're more like "providers".  The term "manager" implies there's some nontrivial amount of underlying complexity being *managed*, but it's not, it's *just* a container for the things that actually do that.

crates/common/src/config.rs
delbonis: The term "common" refers to what's shared underlying infra between all of the *binaries*.

crates/storage/src/managers/mod.rs
bewakes: I see. Probably `ManagersContainer` then?

> and not in the managers module.

Why not? I see the container to be just a container that will be passed around other services and tasks. It won't have different and extra functionalities of its own compared to what the current managers have.

crates/common/src/config.rs
bewakes: I think then it makes sense to keep it in primitives? Or perhaps in a new crate(`node` that you mentioned somewhere)? It should definitely be somewhere outside bin because reader, writer, bridge, etc have their configs derived from this `Config`.

crates/storage/src/managers/mod.rs
delbonis: How about `NodeStorage`?

crates/common/src/config.rs
delbonis: It's not a consensus-related type like what the primitives crate is for.

Yes as I said before it would make sense to have a new node crate like Reth has.

>because reader, writer, bridge, etc have their configs derived from this Config.

But we don't want to make btcio dependent on this new node crate, the node crate makes sense to be dependent on btcio instead.  The btcio services have their own configuration that we build from our config.

crates/storage/src/managers/mod.rs
bewakes: That's better 👍

crates/common/src/config.rs
bewakes: Should we have a new crate `config`? Well, do you agree that `Config` should be somewhere outside of bin?

delbonis: It just makes sense in that node crate.

bewakes: Thought some more and decided to have a new `config` crate. This way, crates depending on `config` don't have to pull in the `node` crate.

delbonis: What would be depending on this config crate but not the node crate?

delbonis (approved):
Ok I think I'm fine with this as-is.  It's still a work in progress but this is an improvement.

sapinb (approved):

PR #563, opened 2024-12-26
STR-799: Upgrade CL guest code to prove range of N blocks

MdTeach:
## Description
* Upgrades the current CL guest code and prover manager to support proving the range of CL blocks.
* Refactor proof job in terms of the block IDs for El, CL and CL_Agg

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


bin/prover-client/src/operators/checkpoint.rs
```diff
@@ -55,6 +59,24 @@ impl CheckpointOperator {
...
+            .cl_client
+            .get_headers_at_idx(block_num)
+            .await
+            .inspect_err(|_| error!(%block_num, "Failed to fetch l2_headers"))
+            .map_err(|e| ProvingTaskError::RpcError(e.to_string()))?;
+
+        let cl_stf_id_buf: Buf32 = l2_headers
+            .expect("invalid height")
+            .first()
+            .expect("at least one l2 blockid")
```
delbonis, line 74: Should handle this by printing an error and failing out in case the database is corrupted instead of just panicking.

bin/prover-client/src/operators/cl_stf.rs
```diff
@@ -42,56 +42,67 @@ impl ClStfOperator {
...
-            .get_header_by_id(id)
+            .get_header_by_id(block_id)
             .await
-            .map_err(|e| ProvingTaskError::RpcError(e.to_string()))?
-            .expect("invalid blkid");
-        Ok(header.block_idx)
+            .inspect_err(|_| error!(%block_id, "Failed to fetch l2_header"))
+            .map_err(|e| ProvingTaskError::RpcError(e.to_string()))?;
+
+        let prev_block: Buf32 = l2_block.expect("invalid height").prev_block.into();
```
delbonis, line 70: Similar here, should handle it.

bin/prover-client/src/operators/checkpoint.rs
```diff
@@ -86,10 +108,18 @@ impl ProvingOp for CheckpointOperator {
...
+        let end_l2_idx = self.get_l2id(checkpoint_info.l2_range.1).await?;
+        let l2_range = vec![(start_l2_idx, end_l2_idx)];
+
         let l2_batch_keys = self
             .l2_batch_operator
-            .create_task(checkpoint_info.l2_range, task_tracker.clone(), db)
+            .create_task(l2_range, task_tracker.clone(), db)
             .await?;
+
         let l2_batch_id = l2_batch_keys.first().expect("at least one").context();
```
delbonis, line 123: Should be checked ahead of time before we do anything so that we are less likely to have dangling jobs.

bin/prover-client/src/operators/cl_agg.rs
```diff
@@ -29,28 +30,25 @@ impl ClAggOperator {
...
         task_tracker: Arc<Mutex<TaskTracker>>,
         db: &ProofDb,
     ) -> Result<Vec<ProofKey>, ProvingTaskError> {
-        let (start_height, end_height) = params;
+        let mut cl_stf_deps = Vec::with_capacity(batches.len());
 
-        let len = (end_height - start_height) as usize + 1;
-        let mut cl_stf_deps = Vec::with_capacity(len);
+        let start_blkid = batches.first().expect("Proof request with empty batch").0;
+        let end_blkid = batches.last().expect("Proof request with empty batch").1;
```
delbonis, line 44: These too.

bin/prover-client/src/operators/cl_stf.rs
```diff
@@ -42,56 +42,67 @@ impl ClStfOperator {
...
+        let header = self
             .cl_client
-            .get_headers_at_idx(block_num)
+            .get_header_by_id(cl_block_id)
             .await
-            .inspect_err(|_| error!(%block_num, "Failed to fetch l2_headers"))
-            .map_err(|e| ProvingTaskError::RpcError(e.to_string()))?;
+            .inspect_err(|_| error!(%cl_block_id, "Failed to fetch corresponding ee data"))
+            .map_err(|e| ProvingTaskError::RpcError(e.to_string()))?
+            .expect("invalid height");
```
delbonis, line 52: Here.

bin/prover-client/src/operators/cl_stf.rs
```diff
@@ -42,56 +42,67 @@ impl ClStfOperator {
...
+                (el_start_block_id, el_end_block_id),
+                task_tracker.clone(),
+                db,
+            )
             .await?;
+
         let evm_ee_id = evm_ee_tasks
             .first()
             .expect("creation of task should result on at least one key")
             .context();
```
delbonis, line 103: But what if it doesn't?

bin/prover-client/src/operators/checkpoint.rs
```diff
@@ -86,10 +108,18 @@ impl ProvingOp for CheckpointOperator {
             .await?;
         let l1_batch_id = l1_batch_keys.first().expect("at least one").context();
```
delbonis, line 109: Here.

functional-tests/fn_prover_cl_dispatch.py
```diff
@@ -13,17 +19,27 @@ def __init__(self, ctx: flexitest.InitContext):
...
+
         self.debug(f"using task id: {task_id}")
         assert task_id is not None
 
         time_out = 10 * 60
         wait_for_proof_with_time_out(prover_client_rpc, task_id, time_out=time_out)
+
+    def blockidx_2_blockid(self, seqrpc, blockidx):
+        l2_blks = seqrpc.strata_getHeadersAtIdx(blockidx)
+        return l2_blks[0]["block_id"]
```
delbonis, line 45: Have been trying to use the term "slot" more often since it's more clear than blockidx and less prone to accidental misreading/mistypes.

Since this doesn't reference `self`, should we move it out of the class to a toplevel fn or even into `utils`?

functional-tests/fn_prover_el_dispatch.py
```diff
@@ -20,16 +20,24 @@ def __init__(self, ctx: flexitest.InitContext):
...
         self.debug(f"got task ids: {task_ids}")
         task_id = task_ids[0]
         self.debug(f"using task id: {task_id}")
         assert task_id is not None
 
         time_out = 10 * 60
         wait_for_proof_with_time_out(prover_client_rpc, task_id, time_out=time_out)
+
+    def blockidx_2_blockid(self, rethrpc, blockid):
+        return rethrpc.eth_getBlockByNumber(hex(blockid), False)["hash"]
```
delbonis, line 43: Not sure of a better name for this.

delbonis (changes requested):
Looks good, main issue is that error handling should be a little more deliberate instead of panicking if things seem wrong.  I know a lot of this was written with `.expect` in places but since we're in here making changes to related code we have good reason to clean it up a bit more.

bin/prover-client/src/operators/checkpoint.rs
```diff
@@ -59,6 +61,32 @@ impl CheckpointOperator {
...
+        let l2_headers = self
+            .cl_client
+            .get_headers_at_idx(block_num)
+            .await
+            .inspect_err(|_| error!(%block_num, "Failed to fetch l2_headers"))
+            .map_err(|e| ProvingTaskError::RpcError(e.to_string()))?;
+
+        let headers = l2_headers.ok_or_else(|| {
+            error!(%block_num, "No L2 headers found at block height");
+            ProvingTaskError::WitnessNotFound
```
delbonis, line 75: I don't think this is the most appropriate error here.

bin/prover-client/src/operators/cl_agg.rs
```diff
@@ -29,28 +31,33 @@ impl ClAggOperator {
...
         db: &ProofDb,
     ) -> Result<Vec<ProofKey>, ProvingTaskError> {
-        let (start_height, end_height) = params;
+        let mut cl_stf_deps = Vec::with_capacity(batches.len());
 
-        let len = (end_height - start_height) as usize + 1;
-        let mut cl_stf_deps = Vec::with_capacity(len);
+        // Extract first and last block IDs from batches, error if empty
+        let (start_blkid, end_blkid) = match (batches.first(), batches.last()) {
+            (Some(first), Some(last)) => (first.0, last.1),
```
delbonis, line 46: I'm unsure of if this is the best way to structure the proof params here.  Isn't this aggregation always "one epoch at a time", extending the proof we have for the previous epoch by the next epoch we just made?

bin/prover-client/src/operators/cl_stf.rs
```diff
@@ -49,56 +49,76 @@ impl ClStfOperator {
...
             .await
-            .map_err(|e| ProvingTaskError::RpcError(e.to_string()))?
-            .expect("invalid blkid");
-        Ok(header.block_idx)
+            .inspect_err(|_| error!(%block_id, "Failed to fetch l2_header"))
+            .map_err(|e| ProvingTaskError::RpcError(e.to_string()))?;
+
+        let prev_block: Buf32 = l2_block
+            .ok_or_else(|| {
+                error!(%block_id, "Block not found");
```
delbonis, line 82: Couple of things:

* it's probably be better to have a method like `get_block_ancestors(id: L2BlockId, n_ancestors: u32) -> Result<Vec<L2BlockId>, _>`, which means we don't have to repeatedly call a "get parent" method to walk back over the chain, and it aligns slightly better with the use case since we know how long we *expect* the batches we're proving to be and can validate the block we expect it to be later
* The logging strings here are rather vague and imprecise.  Seeing "block not found" in the log message without context isn't helpful.  Also should be consistent about `block_id`, `blockid`, `blkid`, `l2blkid`, etc.
* Also not really comfortable with the RPC error variants being sorta handwaved like this.  Not sure what an ideal solution here is though.

bin/prover-client/src/operators/cl_stf.rs
```diff
@@ -112,26 +132,37 @@ impl ProvingOp for ClStfOperator {
...
+            let witness = raw_witness.ok_or(ProvingTaskError::WitnessNotFound)?;
+            stf_witness_payloads.push(witness);
+
+            if blkid == start_block_hash {
+                break;
+            } else {
+                blkid = self.get_prev_block_id(blkid).await?;
+            }
+        }
+        stf_witness_payloads.reverse();
```
delbonis, line 157: We need some safety hatch here so we bail out if we're trying to prove more than a couple thousand blocks instead of trying to load all the way back to genesis.  Just in the event that there's a bug and we have an invalid start ID we can identify it early.  See other comment about loading ancestors.

provers/tests/src/provers/el.rs
```diff
@@ -16,16 +16,16 @@ impl<H: ZkVmHost> ElProofGenerator<H> {
...
+        let (start_block, end_block) = block_range;
+        let evm_segment = EvmSegment::initialize_from_saved_ee_data(*start_block, *end_block);
+
+        Ok(evm_segment.get_inputs().clone())
     }
 
-    fn get_proof_id(&self, block_num: &u64) -> String {
-        format!("el_{}", block_num)
+    fn get_proof_id(&self, block_range: &(u64, u64)) -> String {
+        format!("el_{}_{}", block_range.0, block_range.1)
```
delbonis, line 28: Should be `evmee_`.

Also can we not apply the same rework to use EVM blkids / payload IDs instead of heights here?

delbonis (changes requested):
Still not really happy with the name "operators" it kinda doesn't make sense and it's sorta an overloaded term.  A term like "controller" or "dispatcher" or even maybe just "adaptor" makes a *lot* more sense to me.

bin/prover-client/src/operators/cl_agg.rs
MdTeach: Yes, `cl_agg` is one epoch at a time. 
This `create_task` allows the creator to specify blocks in mini-batches that can be proven independently and subsequently aggregated.

bin/prover-client/src/operators/checkpoint.rs
MdTeach: Did minor refactoring here.

Ideally we can get rid of entire `fn get_l2id(&self, block_num: u64) -> Result<L2BlockId, ProvingTaskError>` once we have `RpcCheckpointInfo` to have block_ids instead of block_idx on [STR-755](https://alpenlabs.atlassian.net/browse/STR-755)

provers/tests/src/provers/el.rs
MdTeach: Since, these test rely on [static witness data](https://github.com/alpenlabs/strata/tree/main/crates/test-utils/data/evm_ee). So, we are planning to get [rid of these](https://alpenlabs.atlassian.net/browse/STR-760) and have migrated to [functional test](https://github.com/alpenlabs/strata/blob/main/functional-tests/fn_prover_el_dispatch.py).

bin/prover-client/src/operators/checkpoint.rs
```diff
@@ -59,6 +61,31 @@ impl CheckpointOperator {
...
+            .first()
+            .ok_or_else(|| {
+                ProvingTaskError::InvalidWitness(format!("Invalid L2 block height {}", block_num))
+            })?
+            .block_id
+            .into();
+
+        Ok(first_header.into())
+    }
+
```
prajwolrg, line 88: This should be part of CL Operator.

bin/prover-client/src/operators/checkpoint.rs
```diff
@@ -88,13 +115,27 @@ impl ProvingOp for CheckpointOperator {
...
             .create_task(checkpoint_info.l1_range, task_tracker.clone(), db)
             .await?;
-        let l1_batch_id = l1_batch_keys.first().expect("at least one").context();
+        let l1_batch_id = l1_batch_keys
+            .first()
+            .ok_or_else(|| ProvingTaskError::NoTasksFound)?
+            .context();
+
+        // Doing the manual block idx to id transformation. Will be removed once checkpoint_info
+        // include the range interms of block_id.
```
prajwolrg, line 124: typo: ~~interms~~  in terms

bin/prover-client/src/operators/constants.rs
```diff
@@ -0,0 +1,6 @@
+/// Maximum number of blocks allowed in a proving range.
+///
+/// This constant serves as a safety limit to prevent proving tasks from processing
+/// an excessively large number of blocks. If the number of blocks to prove exceeds
+/// this limit, the task will be aborted early.
+pub const MAX_PROVING_BLOCK_RANGE: u64 = 1024;
```
prajwolrg, line 6: Does it make sense it to be passed as an argument? I don't have a strong opinion on this. Also let's add a rationale for using `1024` instead of other number.

bin/strata-client/src/rpc_server.rs
```diff
@@ -298,38 +298,24 @@ impl<D: Database + Send + Sync + 'static> StrataApiServer for StrataRpcImpl<D> {
...
-        })
-        .await?;
-
-        // Check if blk_ids is empty
-        let blkid = match blk_ids.first() {
-            Some(id) => id.to_owned(),
-            None => return Ok(None),
-        };
-
+    async fn get_cl_block_witness_raw(&self, blkid: L2BlockId) -> RpcResult<Option<Vec<u8>>> {
```
prajwolrg, line 301: I think this should not return an option. It might make sense to return Error when the information is not found.

crates/rpc/api/src/lib.rs
```diff
@@ -44,7 +44,7 @@ pub trait StrataApi {
     async fn get_exec_update_by_id(&self, block_id: L2BlockId) -> RpcResult<Option<RpcExecUpdate>>;
 
     #[method(name = "getCLBlockWitness")]
-    async fn get_cl_block_witness_raw(&self, index: u64) -> RpcResult<Option<Vec<u8>>>;
+    async fn get_cl_block_witness_raw(&self, block_id: L2BlockId) -> RpcResult<Option<Vec<u8>>>;
```
prajwolrg, line 47: Same as above. I think this should return error if not found instead of None.

crates/rpc/prover-client-api/Cargo.toml
```diff
@@ -13,6 +13,7 @@ rust.unused_must_use = "deny"
 rustdoc.all = "warn"
 
 [dependencies]
+strata-primitives = { workspace = true }
 strata-rpc-types = { workspace = true }
```
prajwolrg, line 17: Use
`.workspace = true`

provers/tests/src/checkpoint.rs
```diff
@@ -52,7 +52,7 @@ impl<H: ZkVmHost> ProofGenerator for CheckpointProofGenerator<H> {
 
         let l2_batch_proof = self
             .l2_batch_prover
-            .get_proof(&(l2_start_height, l2_end_height))
+            .get_proof(&vec![(l2_start_height, l2_end_height)])
```
prajwolrg, line 55: If we are passing reference, we might not need vec.

prajwolrg (changes requested):

bin/prover-client/src/operators/checkpoint.rs
MdTeach: Yes, but since get_l2id is only temporarily present and will be removed soon, transitioning to the CL Operator and invoking it from the Checkpoint Operator would necessitate introducing additional getter methods.
https://github.com/alpenlabs/strata/blob/ad3a19b344154a13cf3f8ed2dc5ccefe5d9461ec/bin/prover-client/src/operators/checkpoint.rs#L123-L128

bin/prover-client/src/operators/constants.rs
MdTeach: I was thinking on this. I think better would be defering to [STR-828](https://alpenlabs.atlassian.net/browse/STR-828 ) where we define the sub-batches of block to be proven in a single proof request.

MdTeach: Reverting `MAX_PROVING_BLOCK_RANGE` check for now

delbonis (approved):
Ok I feel comfortable with this now.

bewakes (approved):

storopoli (approved):
Nice work!
ACK 7c26c7b519517e0f4df8519c999fd3e7ba71a58e

mukeshdroid (approved):
Only change in crypto owned file is adding two CL block ID instead of one in the identifier for STF proof.

prajwolrg (approved):

PR #564, opened 2024-12-26
STR-885 Fix Bridge Client Bug and add functional Test 

voidash:
## Description
The nature of websocket client is that it requires restart once the connection drops. Due to this `bridge-client` stops  if the connection to `strata-client` drops which is not the desired and must be handled. This PR fixes the bug by using managed connection pool called`deadpool`. 

> [!NOTE]
> The functional test deliberately contains duplicated code which can be fixed after STR-734 is merged. STR-734 contains `BridgeTestBase` class that will be utilized by this functional test after rebase. I opened this PR to discuss the bug fix part

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

STR-885


crates/bridge-exec/src/handler.rs
```diff
@@ -19,22 +20,22 @@ use strata_rpc_api::StrataApiClient;
...
+pub struct ExecHandler<TxBuildContext: BuildContext + Sync + Send> {
     /// The build context required to create transaction data needed for signing.
     pub tx_build_ctx: TxBuildContext,
 
     /// The signature manager that handles nonce and signature aggregation.
     pub sig_manager: SignatureManager,
 
     /// The RPC client to connect to the RPC full node.
-    pub l2_rpc_client: L2Client,
+    pub l2_rpc_client_pool: Pool<WsClientManager>,
```
delbonis, line 38: Perhaps make a type alias like `type WsClientPool = Pool<WsClientManager>;` that can be re-exported from where `WsClientManager` is defined instead of having to make downstream consumers import the `Pool` type.

crates/bridge-exec/src/handler.rs
```diff
@@ -286,9 +290,13 @@ where
         Payload: BorshDeserialize + Debug,
     {
         let raw_scope: HexBytes = scope.into();
-        info!(scope=?scope, "getting messages from the L2 Client");
-        let received_payloads = self
-            .l2_rpc_client
+        info!(scope = ?scope, "getting messages from the L2 Client");
```
delbonis, line 293: Can just be `?scope`.

crates/bridge-exec/src/ws_client.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// Returns a `WsClient` wrapped in a `Working` or `NotWorking` state depending on
+    /// the success of the operation.
+    async fn create(&self) -> Result<Self::Type, Self::Error> {
+        let client = WsClientBuilder::default()
+            .build(self.config.url.clone())
+            .await;
+        let bl = match client {
+            Ok(cl) => WsClientState::Working(cl),
+            Err(_) => WsClientState::NotWorking,
+        };
```
delbonis, line 78: If we try to create a new instance and it starts off as being `NotWorking` that should be an error right?  When do we update the state from `Working` to `NotWorking`?

crates/bridge-exec/src/ws_client.rs
```diff
@@ -0,0 +1,159 @@
...
+        batch: BatchRequestBuilder<'a>,
+    ) -> Result<BatchResponse<'a, R>, ClientError>
+    where
+        R: DeserializeOwned + fmt::Debug + 'a,
+    {
+        match &self.0 {
+            WsClientState::Working(inner) => inner.batch_request(batch).await,
+            WsClientState::NotWorking => Err(ClientError::Transport(BoxError::from(
+                "Client is Not Working".to_string(),
+            ))),
```
delbonis, line 156: As a cleanliness thing you could have a function like `make_notworking_err` to dedup constructing this complex error instance.

functional-tests/fn_bridge_deposit_seq_unreliable.py
```diff
@@ -0,0 +1,149 @@
...
+# Local constants
+# D BTC
+DEPOSIT_AMOUNT = DEFAULT_ROLLUP_PARAMS["deposit_amount"]
+# Gas for the withdrawal transaction
+WITHDRAWAL_GAS_FEE = 22_000  # technically is 21_000
+# Ethereum Private Key
+# NOTE: don't use this private key in production
+ETH_PRIVATE_KEY = "0x0000000000000000000000000000000000000000000000000000000000000001"
+# BTC Operator's fee for withdrawal
+OPERATOR_FEE = DEFAULT_ROLLUP_PARAMS["operator_fee"]
```
delbonis, line 28: Spaces between each of these after each assignment before the next comment to make them easier to read.

crates/bridge-exec/src/handler.rs
```diff
@@ -122,9 +122,13 @@ where
...
             .expect("should be able to borsh serialize raw message");
 
-        self.l2_rpc_client
-            .submit_bridge_msg(raw_message.into())
-            .await?;
+        let l2_rpc_client = self
+            .l2_rpc_client_pool
+            .get()
+            .await
+            .map_err(|err| ExecError::WsPool(err))?;
```
delbonis, line 129: Maybe make an accessor on the type that wraps this repeated pattern so you can just do something like `self.get_ready_rpc_client().await?`.

bin/bridge-client/src/modes/operator/task_manager.rs
```diff
@@ -18,57 +18,56 @@ use strata_storage::ops::{bridge_duty::BridgeDutyOps, bridge_duty_index::BridgeD
...
-            for duty in duties {
-                let exec_handler = self.exec_handler.clone();
-                let bridge_duty_ops = self.bridge_duty_db_ops.clone();
-                let broadcaster = self.broadcaster.clone();
-                handles.spawn(async move {
-                    process_duty(exec_handler, bridge_duty_ops, broadcaster, &duty).await
-                });
-            }
+            }) = self.poll_duties().await
+            {
```
delbonis, line 44: Changing this to an `if let` like this is a minor change but it makes the structure much different.  This should reason about the error conditions more explicitly.

I don't even see the error case being handled here, we need to do that.  From what I can tell, if this errors immediately (like because the connection is broken) then we just spin in a hot loop eating a CPU core until we don't encounter the error anymore.

delbonis (changes requested):
The way errors are reasoned about here needs to be improved, it's written assuming they can't really happen.

crates/bridge-exec/src/ws_client.rs
voidash: This is handled by `deadpool` itself. after the client we receive from the pool is dropped, it goes to recycle phase and if there is any problem then due to recycle error we create new client and at that point it can be set as NotWorking or working. This also means that for every NotWorking client, we regularly need to create and check if it is working or not

bin/bridge-client/src/modes/operator/task_manager.rs
voidash: I have implemented exponential backoff with max retry count which allows Bridge to handle the ws related issues more gracefully

functional-tests/fn_bridge_deposit_seq_unreliable.py
voidash: This will be changed completely after Operator assignment PR is merged

crates/bridge-exec/src/ws_client.rs
delbonis: Right but what's the point of returning a client that we know as soon as it's constructed that it's `NotWorking`?  Is this a normal approach?

I would think that in the `create` impl we'd keep retrying creating the client (with some backoff) until we actually get a working connection that can be used immediately.

bin/bridge-client/src/modes/operator/task_manager.rs
delbonis: Are you *completely sure* that this will not hot spin under any circumstances?  If so then please add a comment explaining why.

voidash: Well it's in new commit. there is `max retry count` added with backoff, which doesn't let it looping forever

bin/bridge-client/src/modes/operator/task_manager.rs
```diff
@@ -34,39 +37,65 @@ where
...
-                        .set_index_async(stop_index)
-                        .await
-                    {
-                        error!(error = %e, %start_index, %stop_index, "could not update duty index");
+                            // Exponential backoff
+                            let delay = Duration::from_secs(1) * 2u32.pow(retries as u32 - 1);
+                            sleep(delay).await;
+                        }
+                        _ => {
+                            bail!(err.to_string());
```
delbonis, line 97: Use structured errors with `thiserror`.

bin/bridge-client/src/modes/operator/task_manager.rs
```diff
@@ -89,14 +118,16 @@ where
             .l2_rpc_client_pool
             .get()
             .await
-            .expect("cannot get rpc client");
+            .map_err(|_| PollDutyError::WsPool)?;
```
delbonis, line 121: Ideally this should propagate up the error that we returned while trying to get a RPC client.

Also like I said elsewhere, wrapping acquiring an RPC client in another fn so we can just do `self.get_ready_rpc_client().await?` would make this cleaner.

delbonis (changes requested):

bin/bridge-client/src/constants.rs
```diff
@@ -6,3 +6,6 @@ pub(super) const DEFAULT_RPC_HOST: &str = "127.0.0.1";
 
 /// The default bridge rocksdb database retry count, if not overridden by the user.
 pub(super) const ROCKSDB_RETRY_COUNT: u16 = 3;
+
+/// The default Rpc retry count, if not overridden by the user.
```
storopoli, line 10:
```suggestion
/// The default RPC retry count, if not overridden by the user.
```

bin/bridge-client/src/modes/operator/task_manager.rs
voidash: It is behing exec_handler so changed it to `self.exec_handler.get_ready_rpc_client()`

bin/bridge-client/src/modes/operator/bootstrap.rs
```diff
@@ -62,10 +65,20 @@ pub(crate) async fn bootstrap(args: Cli) -> anyhow::Result<()> {
...
+    let l2_rpc_client_pool = managed::Pool::<WsClientManager>::builder(manager)
+        .max_size(5)
+        .build()
+        .unwrap();
+
+    let l2_rpc_client = l2_rpc_client_pool
+        .get()
         .await
-        .expect("failed to connect to the rollup RPC server");
+        .expect("cannot get rpc client from pool");
```
storopoli, line 81:
```suggestion
        .expect("cannot get RPC client from pool");
```

bin/bridge-client/src/modes/operator/task_manager.rs
```diff
@@ -18,57 +19,84 @@ use strata_storage::ops::{bridge_duty::BridgeDutyOps, bridge_duty_index::BridgeD
...
+                            .await
+                        {
+                            error!(error = %e, %start_index, %stop_index, "could not update duty index");
+                        }
+                    }
+                }
+                Err(err) => {
+                    match err {
+                        PollDutyError::RpcError(err) => {
+                            error!(%err, "could not get rpc response");
```
storopoli, line 84:
```suggestion
                            error!(%err, "could not get RPC response");
```

bin/bridge-client/src/modes/operator/task_manager.rs
```diff
@@ -18,57 +19,84 @@ use strata_storage::ops::{bridge_duty::BridgeDutyOps, bridge_duty_index::BridgeD
...
+                            retries += 1;
+                            if retries >= max_retries {
+                                error!(%err, "Exceeded maximum retries to acquire client. Failing gracefully");
+
+                                bail!("Exceeded maximum retries to acquire client")
```
storopoli, line 89: `bail!` is not "Failing gracefully" in my opinion.
Either we fail really gracefully or we change the error message.

crates/bridge-exec/src/errors.rs
```diff
@@ -47,6 +47,10 @@ pub enum ExecError {
     /// Signer does not have access to the [`Xpriv`](bitcoin::bip32::Xpriv)
     #[error("bitcoin signer do not have access to the private keys, i.e. xpriv")]
     Xpriv,
+
+    /// Error getting the web socket client from pool
+    #[error("fetching websocket client from pool failed")]
```
storopoli, line 52: `web socket` ≠ `websocket`.

Decide which one to use.

storopoli: I prefer "WebSocket" as is officially documented: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket

crates/bridge-exec/src/handler.rs
```diff
@@ -19,22 +20,25 @@ use strata_rpc_api::StrataApiClient;
 use strata_rpc_types::HexBytes;
 use tracing::{debug, info, warn};
 
-use crate::errors::{ExecError, ExecResult};
+use crate::{
+    errors::{ExecError, ExecResult},
+    ws_client::{WsClient, WsClientManager},
+};
+
+/// Websocket client pool
```
storopoli, line 28:
```suggestion
/// WebSocket client pool
```

crates/bridge-exec/src/ws_client.rs
```diff
@@ -0,0 +1,158 @@
+//! Wrapper for managing a WebSocket client that allows recycling or restarting
+//! the client without restarting the entire program.
```
storopoli, line 2:
```suggestion
//! Wrapper for managing a WebSocket client that supports connection recycling and client restarting.
```

crates/bridge-exec/src/ws_client.rs
```diff
@@ -0,0 +1,158 @@
...
+        params::BatchRequestBuilder,
+        traits::ToRpcParams,
+        BoxError, ClientError, DeserializeOwned,
+    },
+    ws_client::{WsClient as WebsocketClient, WsClientBuilder},
+};
+
+/// Configuration for the WebSocket client.
+///
+/// settings that are necessary to initialize and configure
```
storopoli, line 20:
```suggestion
/// Settings that are necessary to initialize and configure
```

crates/bridge-exec/src/ws_client.rs
```diff
@@ -0,0 +1,158 @@
...
+/// the WebSocket client, ie. URL of the WebSocket server.
+#[derive(Clone, Debug)]
+pub struct WsClientConfig {
+    /// The URL of the WebSocket server.
+    pub url: String,
+}
+
+/// Manager for creating and recycling WebSocket clients.
+///
+/// This struct manages the lifecycle of WebSocket clients, including creating
```
storopoli, line 30:
```suggestion
/// Manages the lifecycle of WebSocket clients, including creating
```

"this struct" is ambiguous. It is already a `pub struct Foo`.

crates/bridge-exec/src/ws_client.rs
```diff
@@ -0,0 +1,158 @@
...
+///
+/// - `Working`: The client is connected and operational.
+/// - `NotWorking`: The client is not connected or has encountered an error.
+#[derive(Debug)]
+pub enum WsClientState {
+    /// The WebSocket client is connected and operational.
+    Working(WebsocketClient),
+
+    /// The WebSocket client is not connected or is in a failed state.
+    NotWorking,
```
storopoli, line 48: We can doclink:

```suggestion
    /// The [`WebsocketClient`] is connected and operational.
    Working(WebsocketClient),

    /// The [`WebsocketClient`] is not connected or is in a failed state.
    NotWorking,
```

crates/bridge-exec/src/ws_client.rs
```diff
@@ -0,0 +1,158 @@
...
+///
+/// This struct manages the lifecycle of WebSocket clients, including creating
+/// new clients and determining whether existing clients are still valid.
+#[derive(Clone, Debug)]
+pub struct WsClientManager {
+    /// The configuration used to initialize WebSocket clients.
+    pub config: WsClientConfig,
+}
+
+/// Represents the state of a WebSocket client.
```
storopoli, line 38:
```suggestion
/// Represents the state of a WebSocket client:
```

crates/bridge-exec/src/ws_client.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// The WebSocket client is connected and operational.
+    Working(WebsocketClient),
+
+    /// The WebSocket client is not connected or is in a failed state.
+    NotWorking,
+}
+
+/// Wrapper for the WebSocket client state.
+///
+/// This struct encapsulates the `WsClientState`, enabling unified management of
```
storopoli, line 53:
```suggestion
/// Encapsulates the [`WsClientState`], enabling unified management of
```

crates/bridge-exec/src/ws_client.rs
```diff
@@ -0,0 +1,158 @@
...
+/// both connected and failed client states.
+#[derive(Debug)]
+pub struct WsClient(WsClientState);
+
+/// Implements the `Manager` trait for managing WebSocket clients.
```
storopoli, line 58:
```suggestion
/// Implements the [`Manager`] trait for managing WebSocket clients.
```

crates/bridge-exec/src/ws_client.rs
```diff
@@ -0,0 +1,158 @@
...
+///
+/// The `WsClientManager` provides methods to create new clients and recycle
```
storopoli, line 60:
```suggestion
/// The [`WsClientManager`] provides methods to create new clients and recycle
```

crates/bridge-exec/src/ws_client.rs
```diff
@@ -0,0 +1,158 @@
...
+/// existing ones, ensuring that clients remain in a valid state.
+impl Manager for WsClientManager {
+    type Type = WsClient;
+    type Error = jsonrpsee::core::StringError;
+
+    /// Creates a new WebSocket client.
+    ///
+    /// Attempts to initialize a new WebSocket client using the provided configuration.
+    /// Returns a `WsClient` wrapped in a `Working` or `NotWorking` state depending on
```
storopoli, line 69:
```suggestion
    /// Returns a [`WsClient`] wrapped in a [`WsClientState::Working`] or [`WsClientState::NotWorking`] state depending on
```

crates/bridge-exec/src/ws_client.rs
storopoli: Let's make unwanted or unreachable state invalid?
I agree, this should either error or keep trying to succeed with some backoff and maximum retries.

crates/bridge-exec/src/ws_client.rs
```diff
@@ -0,0 +1,158 @@
...
+            )),
+        }
+    }
+}
+
+fn make_not_working_error() -> ClientError {
+    ClientError::Transport(BoxError::from("Client is Not Working".to_string()))
+}
+
+/// Implements the `ClientT` trait for `WsClient`.
```
storopoli, line 116:
```suggestion
/// Implements the [`ClientT`] trait for [`WsClient`].
```

crates/bridge-exec/src/ws_client.rs
```diff
@@ -0,0 +1,158 @@
...
+///
+/// This implementation allows `WsClient` to perform JSON-RPC operations,
```
storopoli, line 118:
```suggestion
/// This implementation allows [`WsClient`] to perform JSON-RPC operations,
```

functional-tests/fn_bridge_deposit_seq_unreliable.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+@flexitest.register
+class BridgeDepositSequencerUnreliableTest(testenv.StrataTester):
+    """
+    TODO: Depends on STR-734 operator reassignment, and this can be merged only that is merged
+
```
storopoli, line 35: This should be moved to a comment.

functional-tests/fn_bridge_deposit_seq_unreliable.py
```diff
@@ -0,0 +1,149 @@
...
+        bridge_pk = get_bridge_pubkey(seqrpc)
+        self.debug(f"Bridge pubkey: {bridge_pk}")
+        self.debug("Stopping the sequencer")
+
+        self.make_drt(ctx, el_address, bridge_pk)
+        time.sleep(2)
+
+        # stop sequencer
+        seq.stop()
+        time.sleep(1)
```
storopoli, line 99: Can we avoid `time.sleep`? Maybe use `wait_until`? Is that possible?

bin/bridge-client/src/args.rs
```diff
@@ -71,6 +71,12 @@ pub(crate) struct Cli {
         description = "retry count for the rocksdb database (default: 3)"
     )]
     pub retry_count: Option<u16>,
+
+    #[argh(
+        option,
+        description = "max retries for when rpc server fails during duty polling"
```
storopoli, line 77:
```suggestion
        description = "max retries for when RPC server fails during duty polling"
```

bin/bridge-client/src/errors.rs
```diff
@@ -5,3 +5,12 @@ pub enum InitError {
     #[error("Invalid operation mode, expected: operator(op) or challenger(ch), got: {0}")]
     InvalidMode(String),
 }
+
+#[derive(Debug, Clone, Error)]
+pub enum PollDutyError {
+    #[error("Rpc client: {0}")]
```
storopoli, line 11:
```suggestion
    #[error("RPC client: {0}")]
```

bin/bridge-client/src/errors.rs
```diff
@@ -5,3 +5,12 @@ pub enum InitError {
...
+    RpcError(String),
+
+    #[error("fetching websocket client from pool failed")]
```
storopoli, line 14:
```suggestion
    #[error("fetching WebSocket client from pool failed")]
```

storopoli (changes requested):

crates/bridge-exec/src/ws_client.rs
voidash: Thanks for pointing it out.

crates/bridge-exec/src/errors.rs
voidash: TIL. It should be WebSocket as it refers to protocol rather than web socket which refers to socket used over the internet

storopoli (changes requested):

storopoli (approved):
ACK 91e61df693a1dbe7763f59804c5d046d66ddc41f

functional-tests/tests/fn_bridge_deposit_seq_unreliable.py
```diff
@@ -0,0 +1,71 @@
...
+
+def check_sequencer_down(seqrpc):
+    """
+    Returns True if sequencer RPC is down
+    """
+    try:
+        seqrpc.strata_protocolVersion()
+        return False
+    except RuntimeError:
+        return True
```
storopoli, line 71: This is repeated as it is already included in `utils.py`? Bad rebase?

storopoli (changes requested):

delbonis (approved):
I think this is good now.

storopoli (approved):
ACK 083c1e5d30b8a4286a1fb5771b156857576f24cc

PR #568, opened 2024-12-30
Str 797 l1 block manager

bewakes:
## Description

This PR does the following: 
1. Adds `L1DataManager` which basically hides the `Database` implementation internally.
2. Adds a new `inst_ops_simple` macro that removes some boilerplate that was required in order to define `*DbOps`. Basically, now we need pass in the corresponding db method name to the macro and the functions defined should already exist in the database interface.
3. Refactors `L2DataOps` to use this new macro.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/storage/src/exec.rs
+++ b/crates/storage/src/exec.rs
@@ -112,12 +112,19 @@ macro_rules! inst_ops_simple {
         }
 
         $(
-            fn $iname < $tparam : $tpconstr > (context: &Context<$tparam>, $($aname : $aty),* ) -> DbResult<$ret> {
-                context.db.as_ref(). $iname ( $($aname),* )
-            }
+            inst_ops_ctx_shim!($iname<$tparam: $tpconstr>($($aname: $aty),*) -> $ret);
         )*
     }
 }
 
+macro_rules! inst_ops_ctx_shim {
+    ($iname:ident<$tparam: ident : $tpconstr:tt>($($aname:ident: $aty:ty),*) -> $ret:ty) => {
+        fn $iname < $tparam : $tpconstr > (context: &Context<$tparam>, $($aname : $aty),* ) -> DbResult<$ret> {
+            context.db.as_ref(). $iname ( $($aname),* )
+        }
+    }
+}
+
 pub(crate) use inst_ops;
+pub(crate) use inst_ops_ctx_shim;
 pub(crate) use inst_ops_simple;
--- a/crates/storage/src/exec.rs
+++ b/crates/storage/src/exec.rs
@@ -9,7 +9,7 @@ pub use tracing::*;
 /// Handle for receiving a result from a database operation on another thread.
 pub type DbRecv<T> = tokio::sync::oneshot::Receiver<DbResult<T>>;
 
-macro_rules! inst_ops_common {
+macro_rules! inst_ops {
     {
         ($base:ident, $ctx:ident $(<$($tparam:ident: $tpconstr:tt),+>)?) {
             $($iname:ident($($aname:ident: $aty:ty),*) => $ret:ty;)*
@@ -59,24 +59,7 @@ macro_rules! inst_ops_common {
             pub struct Inner $(<$($tparam: $tpconstr + Sync + Send + 'static),+>)? {
                 ctx: Arc<$ctx $(<$($tparam),+>)?>,
             }
-        }
-    }
-}
 
-macro_rules! inst_ops {
-    {
-        ($base:ident, $ctx:ident $(<$($tparam:ident: $tpconstr:tt),+>)?) {
-            $($iname:ident($($aname:ident: $aty:ty),*) => $ret:ty;)*
-        }
-    } => {
-
-        inst_ops_common!{
-            ($base, $ctx $(<$($tparam: $tpconstr),+>)?) {
-                $( $iname($($aname: $aty),*) => $ret; )*
-            }
-        }
-
-        paste::paste! {
             impl $(<$($tparam: $tpconstr + Sync + Send + 'static),+>)? ShimTrait for Inner $(<$($tparam),+>)? {
                 $(
                     fn [<$iname _blocking>] (&self, $($aname: $aty),*) -> DbResult<$ret> {
@@ -102,22 +85,18 @@ macro_rules! inst_ops {
     }
 }
 
-/// Same as `inst_ops` but calls the methods on context db. The function name must exist
-/// in the database instance
-macro_rules! inst_ops_auto {
+macro_rules! inst_ops_simple {
     {
-        ($base:ident, $ctx:ident $(<$($tparam:ident: $tpconstr:tt),+>)?) {
+        (< $tparam:ident: $tpconstr:tt > => $base:ident) {
             $($iname:ident($($aname:ident: $aty:ty),*) => $ret:ty;)*
         }
     } => {
-        /// Database context for an database operation interface.
-        pub struct Context$(<$($tparam: $tpconstr), +>)? {
-            db: Arc<$($($tparam),+)?>
-
+        pub struct Context<$tparam : $tpconstr> {
+            db: Arc<$tparam>,
         }
 
-        impl$(<$($tparam: $tpconstr + Sync + Send + 'static),+>)? Context$(<$($tparam),+>)? {
-            pub fn new(db: Arc<$($($tparam),+)?>) -> Self {
+        impl<$tparam : $tpconstr + Sync + Send + 'static> Context<$tparam> {
+            pub fn new(db: Arc<$tparam>) -> Self {
                 Self { db }
             }
 
@@ -126,38 +105,19 @@ macro_rules! inst_ops_auto {
             }
         }
 
-        inst_ops_common!{
-            ($base, $ctx $(<$($tparam: $tpconstr),+>)?) {
-                $( $iname($($aname: $aty),*) => $ret; )*
+        inst_ops! {
+            ($base, Context<$tparam : $tpconstr>) {
+                $($iname ($($aname : $aty ),*) => $ret ;)*
             }
         }
 
-        paste::paste! {
-            impl $(<$($tparam: $tpconstr + Sync + Send + 'static),+>)? ShimTrait for Inner $(<$($tparam),+>)? {
-                $(
-                    fn [<$iname _blocking>] (&self, $($aname: $aty),*) -> DbResult<$ret> {
-                        self.ctx.db.$iname($($aname),*)
-                    }
-
-                    fn [<$iname _chan>] (&self, pool: &threadpool::ThreadPool, $($aname: $aty),*) -> DbRecv<$ret> {
-                        let (resp_tx, resp_rx) = tokio::sync::oneshot::channel();
-                        let ctx = self.ctx.clone();
-
-                        pool.execute(move || {
-                            let res = ctx.db.$iname($($aname),*);
-                            if resp_tx.send(res).is_err() {
-                                warn!("failed to send response");
-                            }
-                        });
-
-                        resp_rx
-                    }
-                )*
+        $(
+            fn $iname < $tparam : $tpconstr > (context: &Context<$tparam>, $($aname : $aty),* ) -> DbResult<$ret> {
+                context.db.as_ref(). $iname ( $($aname),* )
             }
-        }
+        )*
     }
 }
 
 pub(crate) use inst_ops;
-pub(crate) use inst_ops_auto;
-pub(crate) use inst_ops_common;
+pub(crate) use inst_ops_simple;
--- a/crates/storage/src/exec.rs
+++ b/crates/storage/src/exec.rs
@@ -59,7 +59,7 @@ where
     }
 }
 
-macro_rules! inst_ops {
+macro_rules! inst_ops_common {
     {
         ($base:ident, $ctx:ident $(<$($tparam:ident: $tpconstr:tt),+>)?) {
             $($iname:ident($($aname:ident: $aty:ty),*) => $ret:ty;)*
@@ -99,7 +99,7 @@ macro_rules! inst_ops {
             }
 
             #[async_trait::async_trait]
-            trait ShimTrait: Sync + Send + 'static {
+            pub trait ShimTrait: Sync + Send + 'static {
                 $(
                     fn [<$iname _blocking>] (&self, $($aname: $aty),*) -> DbResult<$ret>;
                     fn [<$iname _chan>] (&self, pool: &threadpool::ThreadPool, $($aname: $aty),*) -> DbRecv<$ret>;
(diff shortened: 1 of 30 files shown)


crates/storage/src/exec.rs
```diff
@@ -135,4 +135,83 @@ macro_rules! inst_ops {
...
+            }
+
+            pub struct Inner $(<$($tparam: $tpconstr + Sync + Send + 'static),+>)? {
+                ctx: Arc<$ctx $(<$($tparam),+>)?>,
+            }
+
+            impl $(<$($tparam: $tpconstr + Sync + Send + 'static),+>)? ShimTrait for Inner $(<$($tparam),+>)? {
+                $(
+                    fn [<$iname _blocking>] (&self, $($aname: $aty),*) -> DbResult<$ret> {
+                        self.ctx.db.$db_method().$iname($($aname),*)
```
bewakes, line 194: This is where it differs from the original `inst_ops` definition

crates/storage/src/exec.rs
```diff
@@ -135,4 +135,83 @@ macro_rules! inst_ops {
...
+                    }
+
+                    fn [<$iname _chan>] (&self, pool: &threadpool::ThreadPool, $($aname: $aty),*) -> DbRecv<$ret> {
+                        let (resp_tx, resp_rx) = tokio::sync::oneshot::channel();
+                        let ctx = self.ctx.clone();
+
+                        pool.execute(move || {
+                            let res = ctx.db.$db_method().$iname($($aname),*);
```
bewakes, line 202: And here.

crates/storage/src/exec.rs
```diff
@@ -135,4 +152,46 @@ macro_rules! inst_ops {
...
```
delbonis, line 181: This is overly coupled to the patterns that are downstream of the current structure of the `Database` trait and how it's been used in these `Context` types.  This is a problem since we're planning on removing the `Database` trait soon in favor of the specific database traits.

What would be *better* is to write a macro to instantiate the toplevel operation fns when they're just directly calling methods on a trait, where we could pass in an instance of `L2Database` directly, and then invoke that for each of these.  This would make it line up more with the existing macro's usage.

crates/storage/src/exec.rs
```diff
@@ -99,7 +99,7 @@ macro_rules! inst_ops {
             }
 
             #[async_trait::async_trait]
-            trait ShimTrait: Sync + Send + 'static {
+            pub(crate) trait ShimTrait: Sync + Send + 'static {
```
delbonis, line 102: Why does this need to be `pub(crate)`?

crates/storage/src/ops/l1tx_broadcast.rs
```diff
@@ -24,73 +20,13 @@ impl<D: BroadcastDatabase + Sync + Send + 'static> Context<D> {
...
-    bcast_db.get_next_tx_idx()
-}
-
-fn put_tx_entry<D: BroadcastDatabase + Sync + Send + 'static>(
-    context: &Context<D>,
-    txid: Buf32,
-    entry: L1TxEntry,
-) -> DbResult<Option<u64>> {
-    trace!(%txid, "insert_new_tx_entry");
-    assert!(entry.try_to_tx().is_ok(), "invalid tx entry {entry:?}");
```
delbonis, line 84: This is an example of where it might be useful/necessary to have this extra indirection instead of making the system completely all-encompassing but closed-off.

delbonis (changes requested):
I added to the ticket late is that we want to change the L1 database interface to be more like the L2 block database where we look up blocks principally by their ID instead of their height, storing the mapping of height -> blockids separately (and now that I'm thinking about it, having a back-reference to the height in the manifest).  This way there's fewer implicit variables when we're fetching a block (just the scan epoch index, not *also* which entire block we're considering, in the presence of reorgs).

crates/storage/src/exec.rs
bewakes: Yes, agree that this is a bit coupled. I did not fully get what you mean afterwards though.

Another way I can think of is to have a `db()` method in the context that returns the appropriate db. And use `$ctx.db().$iname(..)` inside the macro. This makes it much less coupled and we don't have to pass the db method name in macro invocation. It would be ideal to enforce this via some trait on context but that seems tricky. Now it looks like this:

```rust
impl<D: BroadcastDatabase + Sync + Send + 'static> Context<D> {
   // ...

    // Add this method
    pub fn db(&self) -> &Arc<impl L1BroadcastDatabase> {
        self.db.l1_broadcast_db()
    }
}

inst_ops_auto! {
    // Now this is cleaner as we don't need to pass method name
    (BroadcastDbOps, Context<D: BroadcastDatabase>) {
        get_tx_entry(idx: u64) => Option<L1TxEntry>;
}
```

Corresponding change inisde macro def:

```rust
                   fn [<$iname _blocking>] (&self, $($aname: $aty),*) -> DbResult<$ret> {
                    // HERE                        
                    self.ctx.db().$iname($($aname),*)
                    }

                    fn [<$iname _chan>] (&self, pool: &threadpool::ThreadPool, $($aname: $aty),*) -> DbRecv<$ret> {
                        let (resp_tx, resp_rx) = tokio::sync::oneshot::channel();
                        let ctx = self.ctx.clone();

                        pool.execute(move || {
                            // AND HERE
                            let res = ctx.db().$iname($($aname),*);
                            if resp_tx.send(res).is_err() {
                                warn!("failed to send response");
                            }
                        });

                        resp_rx
                    }
```

bewakes: Also, I think that it will be cleaner if we expect the ops interface to implement the trait methods only and everything else be done on the Manager level.

crates/storage/src/exec.rs
bewakes: Actually it doesn't have to be. Updated.

crates/storage/src/ops/l1tx_broadcast.rs
bewakes: This can and possibly should be done on the manager.

crates/storage/src/exec.rs
delbonis: > I did not fully get what you mean afterwards though.

If it's being annoying, just make the `Context` type be generic over the `L1BroadcastDatabase` (or whatever else).

> I think that it will be cleaner if we expect the ops interface to implement the trait methods only 

Can you expand some more on what this would look like?

The goal I had in mind with this ops interface was to abstract over the low-level database provider trait and present the basic operations it has as both async and blocking varieties and without the extra effort involved in passing generic parameters around.  There's some space in there for indirection in case we need to introduce it but it's meant to be a thin wrapper.

The more authentically Rustic way to do that *would* be to make a proc macro that we use on the trait but that's pretty extra and it doesn't have to be that complicated.

crates/storage/src/ops/l1tx_broadcast.rs
delbonis: If we did it there then we'd need to duplicate these checks.  You can make *an* argument that if we really are leaning into it being a thin wrapper then this wouldn't belong, but then we'd just have to deal with that.  In this context also know we're always interacting with the database in a blocking fashion.

crates/storage/src/exec.rs
bewakes: > Can you expand some more on what this would look like?

I think we are totally on the same page regarding the goals of the Ops interface. It should abstract over the provider trait. That's what I am trying to say, if we are doing just that then we can have those ops functions automatically generated as done in this PR. And yes, ideally via a proc macro on the trait/struct itself. Actually I even thought of pursuing that path, but it would be too much for now.

> > I think that it will be cleaner if we expect the ops interface to implement the trait methods only

So, by this I mean the ops helper functions can be automatically generated because we wouldn't be doing anything extra over there and having something like `inst_ops_auto!` that reduces a lot of boilerplate.

> If it's being annoying, just make the Context type be generic over the L1BroadcastDatabase (or whatever else).

Exposing a `db()` method in the `Context` type should be enough leaving the rest as is. Either way we need to expose `db()` method.

crates/storage/src/ops/l1tx_broadcast.rs
bewakes: > If we did it there then we'd need to duplicate these checks.

I am not sure how, because if ops is just an abstraction over db, then it's not the duty of the abstraction to do the checks, it should be somewhere upwards(manager/handler), isn't it?

bewakes:
>  change the  L1 database interface to be more like the L2 block database

We could do that but is that really necessary? Because the premises are different in that for L1 blocks we don't have to deal with storing the forks of the chain while for L2 blocks we have to and thus it makes sense to index by block height for L1 blocks and by block hash for L2 blocks.

crates/storage/src/exec.rs
delbonis: Another strength of the current approach is that it doesn't make any assumptions about the structure of the types involved, you don't have to have any fields/methods with specific names.

We're debating new design directions with this now and if you had just done it in the established way then we would have moved on by now.

If the problem is the boilerplate then write another macro that wraps the `inst_ops!` invocation and encompasses the *whole* thing, I can do that and get back to you.

crates/storage/src/ops/l1tx_broadcast.rs
delbonis: The point is that it allows us flexibility to add these here if we want to, separation of concerns and all that.

crates/storage/src/exec.rs
delbonis: Something like

```rs

macro_rules! inst_ops_simple {
    {
        ($base:ident, $ctx:ident <$tparam:ident: $tpconstr:tt>) {
            $($iname:ident($($aname:ident: $aty:ty),*) => $ret:ty;)*
        }
    } => {
        pub struct Context<$tparam : $tpconstr> {
            db: Arc<$tparam>,
        }

        impl<$tparam : $tpconstr + Sync + Send + 'static> Context<$tparam> {
            pub fn new(db: Arc<$tparam>) -> Self {
                Self { db }
            }

            pub fn into_ops(self, pool: threadpool::ThreadPool) -> $base {
                $base::new(pool, Arc::new(self))
            }
        }

        inst_ops!($base, Context<$tparam>) {
            $($iname ($($aname : $aty ),*) => $ret ;)*
        }

        $(
            fn $iname < $tparam : $tpconstr > (context: &Context<$tparam>, $($aname : $aty),* ) -> DbResult<$ret> {
                self.db.as_ref(). $iname ( $($aname),* )
            }
        )*
    }
}
```

delbonis: That might also let you get rid of a layer of `Arc` there too.

delbonis:
It makes reasoning about the state of the database a lot easier, so I'd say yes.  This separates tracking the current chain separately from tracking what's in particular blocks.

The key part is this

> This way there's fewer implicit variables [to keep in mental context when writing code using L1 data] when we're fetching a block (just the scan epoch index, not also which entire block we're considering, in the presence of reorgs).

crates/storage/src/exec.rs
bewakes: > We're debating new design directions with this now and if you had just done it in the established way then we would have moved on by now.

I did not intend to start a new debate at all. I just don't want to write all the boilerplate code which can sensibly be solved with a new macro as it is in the latest version of this PR. And this doesn't make `inst_ops!` obsolete, we can use it whenever we desire more flexibility for sure. So it is all good.

> Something like
...

Thank you for the directions on what the new macro might look like. This is somewhat similar to what I've done in this PR. Both make some assumptions about the structure of `Context`, this one expects `.db` and the one in the PR expects `.db()` which I think is more flexible. So I will have the `.db()` method return a ref instead of `Arc`.

In regards to the next comment regarding extra indirection, that's a subjective thing where the checks should be(in ops or in manager). I totally agree that `inst_ops!` is very flexible as it doesn't make any assumptions about the context and it is a commendable construct. For me, the checks make sense on the higher level. So will keep things as they are regarding this.

bewakes:
Could you elaborate on what the implicit variables are?

I think I get what you are trying to say. Referring to blocks by their block hash instead of block numbers gives the natural 1-1 mapping between the index key and block. And also as you mentioned, separate height -> blkids or blkid -> height mappings. And how I see things is that we don't need to keep track of L1 forks. Maybe I am missing something?

And from the updated ticket description:
> store blocks according to their blkid and a table mapping heights to lists of blkids (maybe we could define the first in this line to be the block that’s in the canonical chain)

This seems to add more implicit state inside database, doesn't it?

bewakes:
In any case, Let's keep this PR scoped and do the db structure change as a part of [another ticket](https://alpenlabs.atlassian.net/browse/STR-797).

delbonis (changes requested):
(see reply)

crates/storage/src/exec.rs
bewakes: Yeah, it's a good point that the context itself is defined in the macro. Changed.

bewakes: Couldn't do it the following way because `$iname` and `$tparam` repeat differently. So continued using the `inst_ops_common` macro.

delbonis: Made it work with some tweaks.  The issue was actually that the type constraint wasn't being propagated to the inner `inst_ops!` instantiation properly in what I had sketched out.

I tried to convert a few of the other defs but ran into some issues involving `Vec<u8>` being coerced into `&[u8]` in some places.  This probably isn't worth fixing right now, but the ways to solve it would be either (1) make all the underlying DB interfaces always pass ownership or (2) extend the macros to let the user flag if things should be passed by ref or not.  The issue ultimately is that when we use the async call we have to pass ownership of the args to the closure that runs on the thread pool, so if we want to keep it simple we also have to do it when we do the blocking call, even if it's not strictly necessary.

Doing option 1 is probably the easiest here, but this is an example of where the additional indirection provided by the more verbose version is helpful because it lets us message the interfaces a bit better and make it so not everything has to be *exactly* consistent.

delbonis (approved):

delbonis (dismissed):
Should be good to go now.

Rajil1213 (approved):
Only reviewed the bridge-specific parts. Looks good.

storopoli (changes requested):
In `crates/storage/src/exec.rs`, what is the difference between:

- `inst_ops`
- `inst_ops_simple`
- `inst_ops_ctx_shim`

Can we add docstrings for them? And maybe a module-level docs in `crates/storage/src/exec.rs`

When we had only `inst_ops` was kinda of fine, but now it's getting out of hand.

bewakes:
That's a good point @storopoli, thanks! Added the docstrings.

storopoli (approved):
Thanks for the docstrings.
ACK 3a4be6b07f5d0c647f615572d97d160c7bb945f6

delbonis (dismissed):

delbonis:
aw what's failing now?

bewakes:
Can we merge this @delbonis @storopoli ? All of these were passing and somehow main has failing functional tests. Tried resolving locally but can't replicate at all - they pass!

manishbista28 (approved):

sapinb (approved):
ack [6a4679a](https://github.com/alpenlabs/strata/commit/e9a2cc9df3212c1646026d8f902177509a5bab03)

PR #570, opened 2024-12-30
STR-875 integrate bridge-specific changes

Rajil1213:
## Description

This PR cherry-picks commits from the `bitvm2` branch created during the PoC implementation. This includes the following changes:

* Batch checkpoint now includes the commitment.
* Bridge withdrawal duties are only extracted if they belong to a checkpointed state.
* `docker-compose` for bitcoind now has separate wallets for each of the bridge clients.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The `getBridgeDuties` RPC has been preserved here so that it can be used directly during the bridge implementation in the interim. This may be removed altogether later in favor of extracting bridge duties in the bridge node itself from the checkpointed state queried from the OL client.

## Checklist

## Related Issues

Closes STR-875


bin/prover-client/src/operators/checkpoint.rs
```diff
@@ -62,7 +62,7 @@ impl CheckpointOperator {
     /// Retrieves the latest checkpoint index
     pub async fn fetch_latest_ckp_idx(&self) -> Result<u64, ProvingTaskError> {
         self.cl_client
-            .get_latest_checkpoint_index()
+            .get_latest_checkpoint_index(None)
```
delbonis, line 65: I thought we realized this was only populated when we're running a sequencer and we should use some other construct?  Or is this safe here?

crates/rpc/types/src/types.rs
```diff
@@ -220,6 +222,27 @@ pub struct RawBlockWitness {
...
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct RpcCheckpointCommitmentInfo {
+    /// block where checkpoint was posted
+    pub blockhash: L1BlockId,
+    /// txid of txn for this checkpoint
+    pub txid: Txid,
+    /// wtxid of txn for this checkpoint
+    pub wtxid: Wtxid,
+    // other info
+}
```
delbonis, line 234: For convenience could we also include the L1 block height and tx position too?  It's more than just a "commitment" as it is, it's like a "locator" or "selector" or "ref".

crates/state/src/batch.rs
```diff
@@ -252,3 +261,54 @@ impl BootstrapState {
...
+            batch_info,
+            bootstrap_state,
+        }
+    }
+}
+
+#[derive(
+    Clone, Debug, PartialEq, Eq, Arbitrary, BorshSerialize, BorshDeserialize, Serialize, Deserialize,
+)]
+pub struct CommitmentInfo {
```
delbonis, line 283: See above comment, should reason about what this is really intending to represent beyond just being a container for the data.

crates/state/src/chain_state.rs
```diff
@@ -51,16 +51,16 @@ pub struct Chainstate {
...
-struct HashedChainstate {
-    last_block: Buf32,
-    slot: u64,
-    epoch: u64,
-    l1_state_hash: Buf32,
-    pending_withdraws_hash: Buf32,
-    exec_env_hash: Buf32,
-    operators_hash: Buf32,
-    deposits_hash: Buf32,
+#[derive(BorshSerialize, BorshDeserialize, Clone, Copy)]
```
delbonis, line 54: Maybe this is needed for the proof, but what's this all being added for?  Ideally with SSZ we can completely remove this type.  Should add comment.

docker/bitcoind/entrypoint.sh
delbonis, line 1: If we're going to do this we might as well make it take a list of arbitrary addresses and generate funds to all of them so that we don't make the entrypoint script aware of how we're using it in the bridge environment.

docker/docker-compose.yml
```diff
@@ -117,9 +118,14 @@ services:
       BITCOIND_RPC_USER: rpcuser
       BITCOIND_RPC_PASSWORD: rpcpassword
       BITCOIND_WALLET: default
+      BRIDGE_WALLET_1: bridge_0
+      BRIDGE_WALLET_2: bridge_1
+      BRIDGE_WALLET_3: bridge_2
+      GENERATE_BLOCKS: 4 # if env is set then blocks are generated. the number is number of seconds between L1 blocks / 4
```
delbonis, line 124: See above.

delbonis (changes requested):

bin/prover-client/src/operators/checkpoint.rs
Rajil1213: @sapinb might have some inputs on this.

crates/state/src/chain_state.rs
Rajil1213: `Copy` and `Clone` are mainly for passing it around. Was there a reason these weren't already added before?

As for `BorshDeserialize`, it's for the proofs but also for testing. The bridge proof relies on the chain state so some of the unit tests stores the state in a file and reads off it.

docker/bitcoind/entrypoint.sh
Rajil1213: I get where you're coming from but since this is only used during local testing/runs, I think it's okay to not make it completely generic.

delbonis: Since we're in here already we might as well make it so we don't have to redo it again in the future if we need that.  https://en.wikipedia.org/wiki/Zero_one_infinity_rule

crates/state/src/chain_state.rs
delbonis: They weren't added since it was just meant as an implementation detail for the state hashing.

docker/bitcoind/entrypoint.sh
Rajil1213: I agree but I'd still do it over a separate ticket/PR since it is non-trivial how these addresses will be generated and then accessed. The only way I can think of it to have another script that modifies the `entrypoint.sh` file itself. Here's the ticket for it: https://alpenlabs.atlassian.net/browse/STR-825

Rajil1213: It also depends on how generic we should make it. For example, is it enough to just have a variable for the number of wallets to be generated in the format: `aux_{N}`?

delbonis: Have the envvar be space/commad delimited?

```
echo $PREFUND_ADDRESSES | while read addr; do ... done
```

Rajil1213: Yeah the goal is to have separate wallets for each of the bridge nodes so that their UTXO sets are completely different (for stakes, payments, etc.). So we'll need to first create a wallet and do a `getnewaddress`. In the bridge node, we do `listunspent` on the wallet URL.

delbonis: That shouldn't be particularly hard.  We could add a subcommand to `datatool` to generate BIP-84 addrs for that maybe?

Rajil1213: That might work but I'm not sure. Either way, I think it's best to do it over a separate PR since these are only for local environments until there's a proper way to do integration testing for the bridge.

bin/prover-client/src/operators/checkpoint.rs
sapinb: This is based on the current design, and it this case, the prover does need to be connected to the node running the sequencer directly. (both to get pending checkpoint info and also provide the proof to it) 
I see 2 ways this may be resolved:
1. fullnodes can sync pending checkpoints through p2p
2. post commitments to L1 first, then post proofs separately
WDYT ?

bin/strata-client/src/rpc_server.rs
```diff
@@ -89,6 +90,69 @@ impl<D: Database + Sync + Send + 'static> StrataRpcImpl<D> {
...
+            let toplevel_st = chs_db
+                .get_toplevel_state(idx)?
+                .ok_or(Error::MissingChainstate(idx))?;
+
+            Ok(Arc::new(toplevel_st))
+        })
+        .await?;
+
+        Ok((cs, Some(chs)))
+    }
```
bewakes, line 131: This can also be obtained from status channel, don't need to do db call here.
`self.status_channel.chain_state()`.

bin/strata-client/src/rpc_server.rs
```diff
@@ -89,6 +90,69 @@ impl<D: Database + Sync + Send + 'static> StrataRpcImpl<D> {
...
+            return Ok(None);
+        };
+        let (_, l2_blockidx) = last_checkpoint.batch_info.l2_range;
+
+        // in current implementation, chainstate idx == l2 block idx
+        let idx = l2_blockidx;
+
+        let db = self.database.clone();
+
+        wait_blocking("load_checkpoint_chainstate", move || {
```
bewakes, line 146: same here

functional-tests/fn_sync_bitcoin_reorg.py
```diff
@@ -121,5 +121,5 @@ def check_nth_checkpoint_finalized_on_reorg(
     wait_until(
         lambda: seqrpc.strata_syncStatus()["finalized_block_id"] == to_finalize_blkid,
         error_with="Block not finalized",
-        timeout=10,
+        timeout=120,
```
bewakes, line 124: It should not take 2 minutes to finalize.

bewakes (changes requested):

bin/strata-client/src/rpc_server.rs
Rajil1213: Nice! Will make that change.

bin/strata-client/src/rpc_server.rs
Rajil1213: Does the channel have the latest checkpointed chain state?

functional-tests/fn_sync_bitcoin_reorg.py
Rajil1213: When running this test alone, the 10-second timeout is enough. But when running the entire test suite with `make test-functional`, it is taking much longer.

Rajil1213: I'm not sure if this is the cause but every once in a while, the response `strata_l1status` has an RPC error (although the published inscription count increases):
```
l1 status = {'bitcoin_rpc_connected': True, 'last_rpc_error': "RPC server returned error 'Block heig
ht out of range' (code -8)", 'cur_height': 611, 'cur_tip_blkid': '5aa2423838fd9e2b29d270ec8be655e1d7
3ee3c524c084f96ba035ac8ddf7ea2', 'last_published_txid': '0ca90645e78681b6cad1bad1ec864380ce3464913d5
b004aed021dedf5ab0344', 'published_inscription_count': 25, 'last_update': 1735987576165, 'network': 
'regtest'}
l1 status = {'bitcoin_rpc_connected': True, 'last_rpc_error': "RPC server returned error 'Block heig
ht out of range' (code -8)", 'cur_height': 611, 'cur_tip_blkid': '5aa2423838fd9e2b29d270ec8be655e1d7
3ee3c524c084f96ba035ac8ddf7ea2', 'last_published_txid': '0ca90645e78681b6cad1bad1ec864380ce3464913d5
b004aed021dedf5ab0344', 'published_inscription_count': 26, 'last_update': 1735987576571, 'network': 
'regtest'}
```
Whenever this happens, the test fails.

bin/strata-client/src/rpc_server.rs
Rajil1213: Done in https://github.com/alpenlabs/strata/pull/570/commits/c8316fd992cbddb7a45eb77d95043316fa44130b

bin/strata-client/src/rpc_server.rs
bewakes: Apologies for the late reply, just saw this. But yes, it has the latest checkpointed state.

functional-tests/fn_sync_bitcoin_reorg.py
Rajil1213: Tried up to 90 and it still does not work.

crates/storage/src/exec.rs
```diff
@@ -53,7 +53,7 @@ where
     }
 
     /// Executes the operation directly.
-    #[allow(dead_code)] // FIXME: remove this
+    #[allow(dead_code)] // #FIXME: remove this.
```
storopoli, line 56:
```suggestion
    #[allow(dead_code)] // FIXME: remove this.
```

What?!

storopoli (dismissed):
ACK b6d216781d015610acc524fd494ae3511eaceffa

functional-tests/fn_sync_bitcoin_reorg.py
delbonis: Should we open a ticket to investigate this?

Rajil1213: I think so. Fwiw, I've reverted the change since it's just flaky.

bin/strata-client/src/rpc_server.rs
Rajil1213: Updated it to use the `status_channel`.

storopoli (dismissed):
ACK 3d4fe71444ed058a94a4d80437bb8f9dc7744f1a

bewakes (dismissed):
LGTM

crates/state/src/batch.rs
Rajil1213: Done in https://github.com/alpenlabs/strata/commit/509ada192a80d1ed1b576f791d8d8df0be39bafa.

bin/prover-client/src/operators/checkpoint.rs
Rajil1213: @delbonis any followup thoughts on this?

storopoli (dismissed):
ACK e9c440c4e48929257d2b4a2c668c32d64704ae53

crates/storage/src/exec.rs
```diff
@@ -11,54 +11,47 @@ pub use tracing::*;
...
+///
+/// // fn get_blob_entry<D: Database>(ctx: Context<D>, id: u32) -> DbResult<Option<u32>> {
+/// //     ...
+/// // }
+///
+/// // fn put_blob_entry<D: Database>(ctx: Context<D>, id: Buf32) -> DbResult<()> {
+/// // }
+///
+/// // ... // Other definitions corresponding to above macro invocation
+/// ```
```
storopoli, line 44: Don't comment out code examples, they will not render as code.

Try to setup with hidden setup lines commented with `#` as in (not that you don't need "```rust") https://github.com/alpenlabs/strata/blob/06faa2f907d791d6d8e27a9d49d6a602f1425872/crates/primitives/src/buf.rs?plain=1#L14-L32

If that is too much you can use a plan B of adding `ignore` as explained in https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes

storopoli (changes requested):

crates/storage/src/exec.rs
storopoli: Oh crap. This is the "cherry-pick" PR? Was this merged and you are cherry-picking? (Yes it was).

Cc @bewakes. I'll fix it in a new PR.
This is how it renders:

image

storopoli: #596

Rajil1213: Done!

storopoli (dismissed):
ACK 79bb2ea314cf36198d9087c24014c637007d2b05

prajwolrg (approved):

delbonis (dismissed):

bin/prover-client/src/operators/checkpoint.rs
delbonis: I guess it's fine, we'll be iterating on this later.

sapinb (approved):

Rajil1213:
Since only the docker stuff has changed since the last round of approvals, I think it's okay to merge this one.
Relevant commit: https://github.com/alpenlabs/strata/pull/570/commits/4951ceccfc5a3a470b150b6c4e814140209a720b

PR #582, opened 2025-01-01
STR-887: Hardcrash recovery functional test

voidash:
## Description

Adds a feature flag `debug_utils` and  a`Bailout Manager` for rpc defined shutdowns on various point in block processing 

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

STR-887


crates/common/src/bail_manager.rs
```diff
@@ -0,0 +1,47 @@
...
+/// Represents the context in which the system may need to "bail out."
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum BailContext {
+    /// Bail out after a sync event.
+    SyncEvent,
+    /// Bail out when a new block is processed in FCM.
+    FcmNewBlock,
+    /// Bail out during block signing.
+    SignBlock,
+}
```
delbonis, line 15: Just do this based on strings so it's more straightforward to define and pass in and it's more neutral to the use case.

crates/common/src/bail_manager.rs
delbonis, line 1: `bail_trigger`?

Should rework the imports on this so that the consts don't have to be exported everywhere and the macro can just be used directly.

Although really, it doesn't have to be macro, it can just be a regular fn that takes an arg like:

```rs
fn check_bail_trigger(ctx: &str) {
    if let Some(val) = BAIL_RECEIVER.borrow() {
        if val == ctx {
            process::exit(1);
        }
    }
}
```

I don't *love* the globals but it is only meant for testing and it's a *clean* implementation.

crates/rpc/api/src/lib.rs
```diff
@@ -152,3 +154,13 @@ pub trait StrataSequencerApi {
...
     async fn get_tx_status(&self, txid: HexBytes32) -> RpcResult<Option<L1TxStatus>>;
 }
+
+/// Debug Specific Api
+#[cfg(feature = "debug-utils")]
+#[cfg_attr(all(feature = "debug-utils", not(feature = "client")), rpc(server))]
+#[cfg_attr(all(feature = "debug-utils", feature = "client"), rpc(server, client))]
+pub trait StrataDebugApi {
+    /// for exiting the client based on context
+    #[method(name = "stratadebug_bail")]
```
delbonis, line 164: I think it's actually fine to expose this RPC trait without requiring the feature flag, so keep the feature flag structure simpler.  It's the impl that we only want to enable if it's set.  If the feature isn't set then we can return with an error instead of actually setting up the abort.

Not completely sure if that's the perfect approach but this is always going to be slightly hacky.

In Puru's PR there's a `debug_` namespace that gets set up.  This would probably make sense to be in there as well actually.  If you rebase we can add it to that.

delbonis (changes requested):
Reasonable approach, some minor tweaks.

functional-tests/fn_rpc_bail_out.py
delbonis, line 1: Each of these 3 cases should be their own test.  And the test namespace would be `crash_` since that's really what we're testing rather than testing something about the RPC.

voidash:
https://alpenlabs.atlassian.net/browse/STR-916
This is the ticket that relates to fix for the `crash_advance_consensus_state.py`

crates/common/src/bail_manager.rs
```diff
@@ -0,0 +1,36 @@
+use std::sync::LazyLock;
+
+use tokio::sync::watch;
+
+pub static DUTY_SIGN_BLOCK: &str = "duty_sign_block";
+pub static ADVANCE_CONSENSUS_STATE: &str = "advance_consensus_state";
+pub static SYNC_EVENT: &str = "sync_event";
```
delbonis, line 7: These should have prefixes like `BAIL_` on them so it's clear what they represent.

However, I am not sure these make sense to be here and instead might make more sense to be closer to where they're used.  What do you think?

functional-tests/utils/utils.py
```diff
@@ -579,3 +579,48 @@ def confirm_btc_withdrawal(
     assert (
         btc_balance == original_balance + expected_increase
     ), "BTC balance after withdrawal is not as expected"
+
+
+def handle_bailout(seq, seqrpc, bail_context):
```
delbonis, line 584: This function name is rather vague.  A better name would be something like `check_handle_bailout`.

Would it make more sense to have another base test class that does this repetitive work?  If these all have the same structure we could do something like this to reduce code duplication by a lot:

```py
class CrashTestBase(flexitest.Test):
        ctx.set_env("basic")

        bail_tag = self.get_bail_tag()
        # do the bulk of this function but with bail_tag

    def get_bail_tag(self) -> str:
        raise NotYetImplemented()

class CrashSomethingTest(testenv.CrashTestBase):
        super().__init__(ctx)

    def get_bail_tag(self) -> str:
        return "something"

    # no need for a main function since we just reuse the parent behavior
```

If we want to customize behavior *around* calling `super().main(whatever)` we can override the method but still reuse it, like for `CrashAdvanceConsensusStateTest` which is broken right now.

delbonis (changes requested):
Some duplication in the testing that could be made more idiomatic/reusable, but overall looks good.

crates/common/src/bail_manager.rs
voidash: I think, for consistency and refactoring having them in central place makes it easy to reason. So i am leaving as it is

functional-tests/envs/testenv.py
```diff
@@ -227,6 +227,70 @@ def make_drt(self, el_address, musig_bridge_pk):
...
+    def check_handle_bailout(self):
+        """
+        Handles the bailout process for the given sequencer RPC.
+
+        Raises:
+            AssertionError: If the bailout or chain tip progress fails.
+        """
+        # wait for 2 seconds for chain tip slot to accumulate.
+        # Since the chain tip requirement is not exact, we can sleep here
+        time.sleep(2)
```
storopoli, line 262: This should be our guideline, if you cannot do `wait_until` & Co. you need to add a comment explaining the why of `time.sleep`.

storopoli (approved):
ACK ec6df75f093167edae5f4f0f666418f1c388d918

PR #583, opened 2025-01-01
STR 159: Add Schnorr Precompile

voidash:
## Description
Adds Schnorr precompile at address `0x5400000000000000000000000000000000000002`for verification that takes input in order as 
* public key : 32 bytes
* message hash : 32 bytes
* schnorr signature: 64 bytes

and produces output 0x01 if true and 0x02 if false

### Type of Change

- [ ] Documenation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues
STR-159


crates/util/python-utils/src/schnorr.rs
```diff
@@ -0,0 +1,60 @@
...
+/// # Arguments
+///
+/// * `message` - A string representing the message to sign, encoded in hexadecimal format.
+/// * `secret_key` - A string representing the secret key, encoded in hexadecimal format.
+///
+/// # Returns
+/// - The Schnorr signature, serialized and encoded in hexadecimal format.
+/// - The public key corresponding to the secret key, encoded in hexadecimal format.
+#[pyfunction]
+pub(crate) fn sign_schnorr_sig(message: String, secret_key: String) -> PyResult<(String, String)> {
```
bewakes, line 20: This function exists in `strata-crypto` crate right? Maybe that crate can be used?

crates/util/python-utils/src/schnorr.rs
```diff
@@ -0,0 +1,60 @@
...
+/// * `msg` - A string representing the original message that was signed, encoded in hexadecimal
+///   format.
+/// * `pk` - A string representing the public key corresponding to the signer, encoded in
+///   hexadecimal format.
+///
+/// # Returns
+///
+/// A boolean indicating whether the signature is valid (`true`) or invalid (`false`).
+#[pyfunction]
+pub fn verify_schnorr_sig(sig: String, msg: String, pk: String) -> bool {
```
bewakes, line 51: Same as above

functional-tests/fn_el_schnorr_precompile.py
```diff
@@ -0,0 +1,75 @@
...
+        assert data == "0x01", f"Schnorr verification failed: expected '0x01', got '{data}'."
+
+        invalid_message = hashlib.sha256(b"MakaluStrata").hexdigest()
+        invalid_precompile_input = public_key + invalid_message + signature
+        data = self.call_precompile_and_get_return_data(invalid_precompile_input)
+        assert data == "0x00", f"Schnorr verification failed: expected '0x00', got '{data}'."
+
+        return True
+
+    def call_precompile_and_get_return_data(self, precompile_input: str):
```
bewakes, line 42: Maybe something like this?
```suggestion
    def wait_for_precompile_response(self, precompile_input: str):
```
A function implicitly returns data so the name doesn't need to have "return".

crates/util/python-utils/src/schnorr.rs
voidash: Instead of using function from `strata-crypto`, there is this same logic repeated because
* From the function signature, the parameters have `String` type. Secp library has support for converting to specific type like  `SecretKey` directly from str but if i were to use strata-crypto then i have to first convert the string to Buf type then the output is also Buf type and again convert that to String.We require two extra levels of conversion for the signature signing logic. 
* This is also returning the public key which sign_schnorr_sig from `strata-crypto` doesn't support.

crates/util/python-utils/src/schnorr.rs
voidash: Well this can utilize `strata-crypto` library

crates/util/python-utils/src/schnorr.rs
voidash: I have made changes such that it uses `strata-crypto` library

crates/reth/evm/src/precompiles/schnorr.rs
```diff
@@ -0,0 +1,65 @@
...
+        return Err(SchnorrError::InvalidInput);
+    }
+
+    Ok(SchnorrInput {
+        public_key: Buf32::new(input[0..32].try_into().map_err(SchnorrError::FixedBytes)?),
+        message_hash: Buf32::new(input[32..64].try_into().map_err(SchnorrError::FixedBytes)?),
+        signature: Buf64::new(
+            input[64..128]
+                .try_into()
+                .map_err(SchnorrError::FixedBytes)?,
```
delbonis, line 39: These `FixedBytes` errors will never occur because the slice was sliced properly.  The variant can just be removed.  The only error case we have to consider here is if the input wasn't 128 bytes.

crates/util/python-utils/src/schnorr.rs
delbonis, line 1: Can these take `&str` instead of forcing `String`?  I'm not sure if pyo3 will naturally handle that more efficiently.

*Really* these should take/return Python `bytes` and rely on the caller to convert to/from hex, although maybe that's not the convention that's been established.

crates/util/python-utils/src/schnorr.rs
```diff
@@ -0,0 +1,62 @@
+use std::str::FromStr;
+
+use pyo3::{pyfunction, PyResult};
+use secp256k1::{schnorr::Signature, Keypair, Message, SecretKey, SECP256K1};
+use strata_crypto::{
+    sign_schnorr_sig as schnorr_sig_sign, verify_schnorr_sig as schnorr_sig_verify,
```
delbonis, line 6: Just add a `_inner` suffix instead of reordering the name, so it's more clear what's happening.

crates/reth/evm/src/precompiles/schnorr.rs
```diff
@@ -0,0 +1,65 @@
...
+        ),
+    })
+}
+
+pub fn schnorr_precompile(input: &Bytes, _gas_limit: u64) -> PrecompileResult {
```
delbonis, line 44: `verify_schnorr_precompile`?  What's the naming convention here?

crates/reth/evm/src/precompiles/schnorr.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    #[error("Slice conversion failed: {0}")]
+    FixedBytes(#[from] TryFromSliceError),
+}
+
+fn parse_schnorr_input(input: &Bytes) -> Result<SchnorrInput, SchnorrError> {
+    // validate if the length is 128 bytes or not
+    if input.len() != 128 {
+        return Err(SchnorrError::InvalidInput);
+    }
```
delbonis, line 31: So reading the [Reth impl of ecrecover](https://github.com/bluealloy/revm/blob/6f3d9f89d538896498274e1fb44ffb42f189922a/crates/precompile/src/secp256k1.rs#L16), which is the closest equivalent, it seems like the convention in Ethereum is to actually pad to the length needed (also 128 bytes in their case) and just ignore bytes past the end.

That's not the most narrowly correct solution in my opinion, but it does means that we'll make it so we never have to reject an input without *trying* to verify it.  So we're "defining away" invalid input by defining a sensible behavior for it.

crates/reth/evm/src/precompiles/schnorr.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    let verification_byte = {
+        match verify_schnorr_sig(
+            &schnorr_input.signature,
+            &schnorr_input.message_hash,
+            &schnorr_input.public_key,
+        ) {
+            true => Bytes::from([0x01]),
+            false => Bytes::from([0x00]),
+        }
```
delbonis, line 56: Maybe `Bytes::from([if result { 1u8 } else { 0u8 }])`?  Or even `result as u8` makes sense instead of having an if at all.

delbonis (changes requested):
Reasonable although the input handling could be improved, see comment.

crates/reth/evm/src/precompiles/schnorr.rs
```diff
@@ -0,0 +1,65 @@
...
+use thiserror::Error;
+
+/// 32 Bytes: Public key
+/// 32 Bytes: Message hash
+/// 64 Bytes: Schnorr Signature
+struct SchnorrInput {
+    public_key: Buf32,
+    message_hash: Buf32,
+    signature: Buf64,
+}
```
storopoli, line 16: This is wrong.
You should add that info for the fields and the docstring for the `struct` should explain the `struct` itself.

crates/reth/evm/src/precompiles/schnorr.rs
```diff
@@ -0,0 +1,65 @@
...
+
+#[derive(Debug, Error)]
+enum SchnorrError {
+    #[error("Input not exactly 128 bytes")]
+    InvalidInput,
+
+    #[error("Slice conversion failed: {0}")]
+    FixedBytes(#[from] TryFromSliceError),
+}
```
storopoli, line 25: docstrings

crates/util/python-utils/Cargo.toml
```diff
@@ -13,8 +13,10 @@ name = "strata_utils"
 crate-type = ["cdylib"]
 
 [dependencies]
-shrex = { version = "0.1.0", path = "../shrex", features = ["serde"] }
+strata-primitives.workspace = true
+strata-crypto.workspace = true
 
+shrex = { version = "0.1.0", path = "../shrex", features = ["serde"] }
```
storopoli, line 19: shree is not "external". Should be moved up and globbed together with the `strata-*` crates.

crates/util/python-utils/src/schnorr.rs
storopoli: There isn't a performance difference between `&str` and `String` when using PyO3, as PyO3 handles both types by copying the data to create a new Python string regardless of the input type. PyO3 internally calls Python APIs to create a new Python `str` object from either type.

See:

- https://pyo3.rs/v0.23.3/conversions/tables.html
- https://stackoverflow.com/questions/60512888/how-is-memory-managed-with-str-and-string-native-types-and-pyresultt-when-pa

However, using `&str` can have improvements in the "Rust" side of things.

storopoli: Also regarding to `Bytes`, yes you're right.
That's on me when I created the `python-utils` crate.

We can open an issue and assign it to me, if needed.

crates/util/python-utils/src/schnorr.rs
```diff
@@ -0,0 +1,62 @@
...
+};
+use strata_primitives::buf::{Buf32, Buf64};
+
+/// Signs a message using the Schnorr signature scheme.
+///
+/// This function generates a Schnorr signature for the given message using the provided secret key.
+/// The function returns the serialized signature and the corresponding public key.
```
storopoli, line 13:
```suggestion
/// Returns the serialized signature and the corresponding public key.
```

crates/util/python-utils/src/schnorr.rs
```diff
@@ -0,0 +1,62 @@
...
+use pyo3::{pyfunction, PyResult};
+use secp256k1::{schnorr::Signature, Keypair, Message, SecretKey, SECP256K1};
+use strata_crypto::{
+    sign_schnorr_sig as schnorr_sig_sign, verify_schnorr_sig as schnorr_sig_verify,
+};
+use strata_primitives::buf::{Buf32, Buf64};
+
+/// Signs a message using the Schnorr signature scheme.
+///
+/// This function generates a Schnorr signature for the given message using the provided secret key.
```
storopoli, line 12:
```suggestion
/// Generates a Schnorr signature for the given message using the provided secret key.
```

"Function" is not needed. This is a `fn` anyways.

crates/util/python-utils/src/schnorr.rs
```diff
@@ -0,0 +1,62 @@
...
+/// The function returns the serialized signature and the corresponding public key.
+///
+/// # Arguments
+///
+/// * `message` - A string representing the message to sign, encoded in hexadecimal format.
+/// * `secret_key` - A string representing the secret key, encoded in hexadecimal format.
+///
+/// # Returns
```
storopoli, line 20:
```suggestion
/// # Returns
///
```

empty line before lists

crates/util/python-utils/src/schnorr.rs
```diff
@@ -0,0 +1,62 @@
...
+/// - The Schnorr signature, serialized and encoded in hexadecimal format.
+/// - The public key corresponding to the secret key, encoded in hexadecimal format.
```
storopoli, line 22: why using `-` here and `*` above?
```suggestion
/// * The Schnorr signature, serialized and encoded in hexadecimal format.
/// * The public key corresponding to the secret key, encoded in hexadecimal format.
```

crates/util/python-utils/src/schnorr.rs
```diff
@@ -0,0 +1,62 @@
...
+    Ok((
+        shrex::encode(sig.as_slice()),
+        shrex::encode(&keypair.x_only_public_key().0.serialize()),
+    ))
+}
+
+/// Verifies a Schnorr signature.
+///
+/// This function verifies the authenticity of a Schnorr signature for a given message
+/// and public key.
```
storopoli, line 42:
```suggestion
/// Verifies the authenticity of a Schnorr signature for a given message
/// and public key.
```

functional-tests/fn_el_schnorr_precompile.py
```diff
@@ -0,0 +1,84 @@
+import hashlib
+
+import flexitest
+from strata_utils import sign_schnorr_sig
+from web3 import Web3
+
+import testenv
+from constants import PRECOMPILE_SCHNORR_ADDRESS
+from utils import wait_until_with_value
```
storopoli, line 9: Nice, I had no idea we had a `wait_until_with_value`.
TIL, thanks!

storopoli (changes requested):

storopoli (changes requested):

crates/reth/evm/src/precompiles/schnorr.rs
```diff
@@ -0,0 +1,53 @@
...
+
+fn parse_schnorr_input(input: &Bytes) -> SchnorrInput {
+    if input.len() < 128 {
+        warn!(
+            input_len = %input.len(),
+            "Input length ({}) is below the minimum required threshold of 128 bytes",
+            input.len()
+        );
+    }
```
delbonis, line 27: I would say that since we're defining away the error case case we don't need this warning here.  We don't control how this gets called anyways, so we could have people spamming logs with warnings by sending txs that verify malformed signatures and repeatedly RBFing them.  This also saves us from having to import `tracing` (and figure out how to use it in the proofs, I'm not sure how that'd even work).

We should just have unit tests for this to ensure that message validation fails as we'd expect it to (if the signature doesn't end in zeroes I guess).

I realize I actually contradicted myself in those two comments, I read the reth code afterwards.

delbonis (changes requested):

crates/reth/evm/src/precompiles/schnorr.rs
delbonis: [Disagree with printing a warning](https://github.com/alpenlabs/strata/pull/583#discussion_r1904375540, this isn't something we can control and it's not what's done in the ecrecover design.

crates/reth/evm/src/precompiles/schnorr.rs
```diff
@@ -1,60 +1,46 @@
...
-        public_key: Buf32::new(input[0..32].try_into().map_err(SchnorrError::FixedBytes)?),
-        message_hash: Buf32::new(input[32..64].try_into().map_err(SchnorrError::FixedBytes)?),
-        signature: Buf64::new(
-            input[64..128]
-                .try_into()
-                .map_err(SchnorrError::FixedBytes)?,
+    SchnorrInput {
+        public_key: Buf32::new(input[0..32].try_into().unwrap()),
+        message_hash: Buf32::new(input[32..64].try_into().unwrap()),
+        signature: Buf64::new(input[64..128].try_into().unwrap(),
```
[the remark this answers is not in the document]
storopoli: Then do `expect("infallible")`.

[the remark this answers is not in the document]
storopoli: Or a better message that argues why: `expect("infallible due to YADA YADA YADA")`

bewakes (dismissed):
Looks good for the "python" parts.

storopoli (changes requested):

crates/reth/evm/src/constants.rs
```diff
@@ -5,6 +5,9 @@ use crate::utils::{u256_from, WEI_PER_BTC};
 /// The address for the Bridgeout precompile contract.
 pub const BRIDGEOUT_ADDRESS: Address = address!("5400000000000000000000000000000000000001");
 
+/// The address for the Bridgeout precompile contract.
+pub const SCHNORR_ADDRESS: Address = address!("5400000000000000000000000000000000000002");
+
 /// The fixed withdrawal amount in wei (10 BTC equivalent).
 pub const FIXED_WITHDRAWAL_WEI: U256 = u256_from(10 * WEI_PER_BTC);
```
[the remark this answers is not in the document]
voidash: I am leaving it as it is for now. But i can take that PR after it is created.

crates/reth/evm/src/precompiles/schnorr.rs
```diff
@@ -0,0 +1,114 @@
...
+        &schnorr_input.signature,
+        &schnorr_input.message_hash,
+        &schnorr_input.public_key,
+    );
+    let verification_byte = Bytes::from([result as u8]);
+
+    // currently we can use [ecrecover hack](https://hackmd.io/@nZ-twauPRISEa6G9zg3XRw/SyjJzSLt9)
+    // which costs around ~3000 gas.
+    // setting it as 0, as this requires further discussion
+    let gas_cost = 0;
```
delbonis, line 42: Created ticket [STR-864](https://alpenlabs.atlassian.net/browse/STR-864) for this.

delbonis (approved):

storopoli (approved):
ACK ff5ec5d06f63c8a034f004f1ce56f95db0aa109a

PR #584, opened 2025-01-02
CI - Prover Performance metric fix - 2

MdTeach:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


.github/workflows/prover.yml
```diff
@@ -1,7 +1,8 @@
 name: Run prover guest code checks
 
 on:
-  pull_request:
+  pull_request_target:
+    types: [opened, synchronize, reopened, edited]
```
storopoli, line 5: Can we work around this?
`zizmor` is flagging this as HIGHLY insecure. See https://woodruffw.github.io/zizmor/audits/#remediation_1

storopoli (changes requested):

delbonis (approved):

storopoli (changes requested):
Do you think is possible to save the "mustached" stuff `${{ foo.BAR }}` in an `env` then use that `env` instead in https://github.com/alpenlabs/strata/blob/e2a15f4d8bc1b80bbb95b421817e25d4b038d17f/.github/workflows/prover.yml#L52-L61

We use that in https://github.com/alpenlabs/strata/blob/de7ffbb6a83a18c1010d82ea12209b657230499b/.github/workflows/cron-nightly-rust.yml?plain=1#L48-L57

This will greatly reduce the command injection attack surface.

See https://woodruffw.github.io/zizmor/audits/#template-injection

evgenyzdanovich:
@storopoli @MdTeach I agree this is not perfect in its current form, but I believe it's safe.
 
The potential vulnerability comes from user-defined inputs (such as branch name or PR title, for example). In fact, `zizmor` is happy with this solution, because all the "mustached" params coming from github itself

storopoli (approved):
ACK e2a15f4d8bc1b80bbb95b421817e25d4b038d17f

PR #587, opened 2025-01-05
Check bitcoin block integrity before extracting relevant info

prajwolrg:
## Description

This PR updates the proof logic to check the integrity of the bitcoin block before extracting relevant info from the block. Instead of computing both the merkle root and witness merkle root, it expects inclusion proof of coinbase (where the witness merkle root is committed).

This also introduces two new types of Inclusion Proof: `L1TxProof` and `L1WtxProof`.
- `L1TxProof`: Is meant to be used to verify the inclusion of transaction against the merkle_root in the Bitcoin Header.
- `L1WtxProof`: Is meant to be used to verify the inclusion of transaction against the witness commitment in the coinbase transaction.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/primitives/src/l1.rs
+++ b/crates/primitives/src/l1.rs
@@ -23,7 +23,9 @@ use rand::rngs::OsRng;
 use reth_primitives::revm_primitives::FixedBytes;
 use serde::{de, Deserialize, Deserializer, Serialize};
 
-use crate::{buf::Buf32, constants::HASH_SIZE, errors::ParseError, impl_buf_wrapper};
+use crate::{
+    buf::Buf32, constants::HASH_SIZE, errors::ParseError, impl_buf_wrapper, utils::get_cohashes,
+};
 
 /// ID of an L1 block, usually the hash of its header.
 #[derive(
@@ -105,7 +107,6 @@ impl From<(u64, u32)> for L1TxRef {
     }
 }
 
-/// TODO: This is duplicate with state::l1::L1TxProof
 /// Merkle proof for a TXID within a block.
 // TODO rework this, make it possible to generate proofs, etc.
 #[derive(
@@ -127,6 +128,21 @@ impl L1TxProof {
 
     pub fn position(&self) -> u32 {
         self.position
+    }
+
+    /// Generates an `L1TxProof` for a transaction at the specified index in the list of
+    /// transactions.
+    ///
+    /// This function computes the `txid` (transaction ID) for each transaction in the provided
+    /// slice, calculates the cohashes for the transaction at the given index, and constructs
+    /// the proof.
+    pub fn generate(transactions: &[Transaction], idx: u32) -> Self {
+        let txids = transactions
+            .iter()
+            .map(|tx| tx.compute_txid())
+            .collect::<Vec<_>>();
+        let (cohashes, _txroot) = get_cohashes(&txids, idx);
+        L1TxProof::new(idx, cohashes)
     }
 }
 
--- a/crates/primitives/src/l1.rs
+++ b/crates/primitives/src/l1.rs
@@ -105,8 +105,8 @@ impl From<(u64, u32)> for L1TxRef {
     }
 }
 
-/// A trait for computing some kind of transaction ID (e.g., [`txid`](bitcoin::Txid) or
-/// [`wtxid`](bitcoin::Wtxid)) from a [`Transaction`].
+/// A trait for computing some kind of transaction ID (e.g., [`Txid`](bitcoin::Txid) or
+/// [`Wtxid`](bitcoin::Wtxid)) from a [`Transaction`].
 ///
 /// This trait is designed to be implemented by "marker" types that define how a transaction ID
 /// should be computed. For example, [`TxIdMarker`] invokes [`Transaction::compute_txid`], and
@@ -116,18 +116,18 @@ pub trait TxIdComputable {
     /// Computes the transaction ID for the given transaction.
     ///
     /// The `idx` parameter allows marker types to handle special cases such as the coinbase
-    /// transaction (which has a zero [`wtxid`](bitcoin::Wtxid)) by looking up the transaction
+    /// transaction (which has a zero [`Wtxid`](bitcoin::Wtxid)) by looking up the transaction
     /// index.
     fn compute_id(tx: &Transaction, idx: usize) -> Buf32;
 }
 
-/// Marker type for computing the [`txid`](bitcoin::Txid).
+/// Marker type for computing the [`Txid`](bitcoin::Txid).
 #[derive(
     Clone, Debug, PartialEq, Eq, Arbitrary, BorshSerialize, BorshDeserialize, Serialize, Deserialize,
 )]
 pub struct TxIdMarker;
 
-/// Marker type for computing the [`wtxid`](bitcoin::Wtxid).
+/// Marker type for computing the [`Wtxid`](bitcoin::Wtxid).
 #[derive(
     Clone, Debug, PartialEq, Eq, Arbitrary, BorshSerialize, BorshDeserialize, Serialize, Deserialize,
 )]
--- a/crates/primitives/src/l1.rs
+++ b/crates/primitives/src/l1.rs
@@ -144,6 +144,33 @@ impl L1TxProof {
         let (cohashes, _txroot) = get_cohashes(&txids, idx);
         L1TxProof::new(idx, cohashes)
     }
+
+    /// Computes the merkle root from corresponding given `tx`
+    pub fn compute_root(&self, transaction: &Transaction) -> Buf32 {
+        // `cur_hash` represents the intermediate hash at each step. After all cohashes are
+        // processed `cur_hash` becomes the root hash
+        let mut cur_hash = transaction.compute_txid().to_byte_array();
+
+        let mut pos = self.position();
+        for cohash in self.cohashes() {
+            let mut buf = [0u8; 64];
+            if pos & 1 == 0 {
+                buf[0..32].copy_from_slice(&cur_hash);
+                buf[32..64].copy_from_slice(cohash.as_ref());
+            } else {
+                buf[0..32].copy_from_slice(cohash.as_ref());
+                buf[32..64].copy_from_slice(&cur_hash);
+            }
+            cur_hash = *crate::hash::sha256d(&buf).as_ref();
+            pos >>= 1;
+        }
+        Buf32::from(cur_hash)
+    }
+
+    /// Verifies the inclusion proof of the given transaction against the given root
+    pub fn verify(&self, transaction: &Transaction, root: Buf32) -> bool {
+        self.compute_root(transaction) == root
+    }
 }
 
 /// Includes [`L1BlockManifest`] along with scan rules that it is applied to
@@ -995,12 +1022,17 @@ mod tests {
         Address, Amount, Network, ScriptBuf, TapNodeHash, TxOut, XOnlyPublicKey,
     };
     use rand::{rngs::OsRng, Rng};
-    use strata_test_utils::ArbitraryGenerator;
+    use strata_test_utils::{
+        bitcoin::{get_btc_chain, get_btc_mainnet_block},
+        ArbitraryGenerator,
+    };
 
     use super::{
-        BitcoinAddress, BitcoinAmount, BitcoinTxid, BorshDeserialize, BorshSerialize, XOnlyPk,
+        BitcoinAddress, BitcoinAmount, BitcoinTxid, BorshDeserialize, BorshSerialize, L1TxProof,
+        XOnlyPk,
     };
     use crate::{
+        buf::Buf32,
         errors::ParseError,
         l1::{BitcoinPsbt, BitcoinTxOut, TaprootSpendPath},
     };
@@ -1429,5 +1461,33 @@ mod tests {
             deserialized_txid, txid,
             "original and deserialized txid must be the same"
         );
+    }
+
+    #[test]
+    fn test_l1_tx_proof() {
+        let btc_chain = get_btc_chain();
+        let block = btc_chain.get_block(40321);
+        let merkle_root: Buf32 = block.header.merkle_root.to_byte_array().into();
+        let txs = &block.txdata;
+
+        for (idx, tx) in txs.iter().enumerate() {
+            let proof = L1TxProof::generate(txs, idx as u32);
+            assert!(proof.verify(tx, merkle_root));
+        }
+    }
+
+    #[test]
+    #[ignore]
+    // This test is ignored because it takes ~190s to run. Run with `cargo test --ignored` for
+    // validation.
+    fn test_l1_tx_proof_2() {
+        let block = get_btc_mainnet_block();
+        let merkle_root: Buf32 = block.header.merkle_root.to_byte_array().into();
+        let txs = &block.txdata;
+
+        for (idx, tx) in txs.iter().enumerate() {
+            let proof = L1TxProof::generate(txs, idx as u32);
+            assert!(proof.verify(tx, merkle_root));
+        }
     }
 }
--- a/crates/primitives/src/l1.rs
+++ b/crates/primitives/src/l1.rs
@@ -2,6 +2,7 @@ use std::{
     fmt::Display,
     io::{self, Read, Write},
     iter::Sum,
+    marker::PhantomData,
     ops::Add,
 };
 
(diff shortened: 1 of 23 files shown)


crates/state/src/l1/tx.rs
```diff
@@ -1,15 +1,16 @@
...
 
+use super::L1TxProof;
 use crate::tx::ProtocolOperation;
 
 /// Tx body with a proof.
 #[derive(
     Clone, Debug, BorshSerialize, BorshDeserialize, PartialEq, Eq, Arbitrary, Serialize, Deserialize,
 )]
 pub struct L1Tx {
+    // TODO: verify if we need L1TxProof or L1WtxProof
```
bewakes, line 13: I believe using `L1WtxProof` should suffice because most of the relevant transactions like DA and checkpoint would have data as witness. For the ones that will probably be legacy(deposit txs?) wtxid and txid will be the same.
cc: @delbonis

crates/state/src/l1/inclusion_proof.rs
```diff
@@ -0,0 +1,176 @@
...
+/// A trait for computing some kind of transaction ID (e.g., `txid` or `wtxid`) from a
+/// [`Transaction`].
+///
+/// By implementing this trait for different "marker" types, multiple ID computations can be handled
+/// without duplicating your proofgeneration logic. For instance, `TxId` uses
+/// `Transaction::compute_txid`, while `WtxId` uses `Transaction::compute_wtxid`.
+pub trait TxIdComputer {
+    /// Computes the transaction ID for the given transaction.
+    fn compute_id(tx: &Transaction, idx: usize) -> Buf32;
+}
```
bewakes, line 48: Do you think `HasId` or `IdProvider` would be better than `TxIdComputer`? And the method could also just be `id()`.

bewakes (commented):
Great work on InclusionProof and the traits! Some minor comments. Otherwise looks good on the core parts.

crates/primitives/src/utils.rs
```diff
@@ -1,33 +1,32 @@
...
-/// - `wtxids`: The witness txids slice
-/// - `index`: The index of the txn for which we want the cohashes
+/// - `elements`: A slice of elements (`txids` or `wtxids`) that can be converted into `Buf32`.
+/// - `index`: The index of the transaction for which we want the cohashes.
 ///
 /// # Returns
-/// - A tuple `(Vec<Buf32>, Buf32)` containing the cohashes and the merkle root
+/// - A tuple `(Vec<Buf32>, Buf32)` containing the cohashes and the Merkle root.
 ///
 /// # Panics
```
storopoli, line 18:
```suggestion
/// # Panics
///
```
empty line after headers

crates/primitives/src/utils.rs
```diff
@@ -1,33 +1,32 @@
...
 ///
 /// # Parameters
-/// - `wtxids`: The witness txids slice
-/// - `index`: The index of the txn for which we want the cohashes
+/// - `elements`: A slice of elements (`txids` or `wtxids`) that can be converted into `Buf32`.
+/// - `index`: The index of the transaction for which we want the cohashes.
 ///
 /// # Returns
-/// - A tuple `(Vec<Buf32>, Buf32)` containing the cohashes and the merkle root
+/// - A tuple `(Vec<Buf32>, Buf32)` containing the cohashes and the Merkle root.
```
storopoli, line 16: Also might be interesting to add a `# Notes` explaining or defining precisely what is a "cohash" (maybe use "co-hash")

crates/proof-impl/btc-blockspace/src/block.rs
```diff
@@ -37,86 +50,101 @@ pub fn check_merkle_root(block: &Block) -> bool {
     }
 }
 
+/// Scans the given coinbase transaction for a witness commitment and returns it if found.
+///
+/// This function iterates over the outputs of the provided `coinbase` transaction from the end
+/// towards the beginning, looking for an output whose `script_pubkey` starts with the “magic” bytes
+/// `[0x6a, 0x24, 0xaa, 0x21, 0xa9, 0xed]`. This pattern indicates an `OP_RETURN` with an
+/// embedded witness commitment header. If such an output is found, the function extracts the
+/// following 32 bytes as the witness commitment and returns a `WitnessCommitment`.
```
storopoli, line 59:
```suggestion
/// following 32 bytes as the witness commitment and returns a [`WitnessCommitment`].
```

crates/proof-impl/btc-blockspace/src/block.rs
```diff
@@ -37,86 +50,101 @@ pub fn check_merkle_root(block: &Block) -> bool {
...
+            <[u8; 32]>::try_from(&coinbase.output[pos].script_pubkey.as_bytes()[6..38]).unwrap();
+        Some(WitnessCommitment::from_byte_array(bytes))
+    } else {
+        None
+    }
+}
+
 /// Computes the witness commitment for the block's transaction list.
 ///
 /// Equivalent to [`compute_witness_commitment`](Block::compute_witness_commitment)
```
storopoli, line 82:
```suggestion
/// Equivalent to [`compute_witness_commitment`](Block::compute_witness_commitment).
```

crates/proof-impl/btc-blockspace/src/block.rs
```diff
@@ -37,86 +50,101 @@ pub fn check_merkle_root(block: &Block) -> bool {
...
 
+/// Scans the given coinbase transaction for a witness commitment and returns it if found.
+///
+/// This function iterates over the outputs of the provided `coinbase` transaction from the end
+/// towards the beginning, looking for an output whose `script_pubkey` starts with the “magic” bytes
+/// `[0x6a, 0x24, 0xaa, 0x21, 0xa9, 0xed]`. This pattern indicates an `OP_RETURN` with an
+/// embedded witness commitment header. If such an output is found, the function extracts the
+/// following 32 bytes as the witness commitment and returns a `WitnessCommitment`.
+///
+/// Based on: [rust-bitcoin](https://github.com/rust-bitcoin/rust-bitcoin/blob/b97be3d4974d40cf348b280718d1367b8148d1ba/bitcoin/src/blockdata/block.rs#L190-L210)
```
storopoli, line 61:
```suggestion
/// Based on: [rust-bitcoin](https://github.com/rust-bitcoin/rust-bitcoin/blob/b97be3d4974d40cf348b280718d1367b8148d1ba/bitcoin/src/blockdata/block.rs#L190-L210).
```

crates/state/src/l1/inclusion_proof.rs
storopoli: Or maybe `TxidComputable` (note the `Txid` and not `TxId`).
The reason is that `Computer` means something that can compute things, like a VM or an execution environment etc. `Computable` means that it is a type that we can "compute" a `Txid` out of it.

Also why do we need the `idx`?

Are you trying to get an [`Outpoint`](https://docs.rs/bitcoin/latest/bitcoin/struct.OutPoint.html)?

I don't think we need a trait for this. Whatever you are doing with this trait, e.g.

```rs
fn foo(bar: &impl TxIdComputer)
```

You can do with:

```rs
fn foo(bar: &Into<Buf32>)
fn foo(bar: &Into<Txid>)
```

What am I missing?

crates/state/src/l1/inclusion_proof.rs
```diff
@@ -0,0 +1,176 @@
...
+#[derive(
+    Clone, Debug, PartialEq, Eq, Arbitrary, BorshSerialize, BorshDeserialize, Serialize, Deserialize,
+)]
+pub struct TxId;
+
+/// Marker type for computing the "witness" wtxid.
+#[derive(
+    Clone, Debug, PartialEq, Eq, Arbitrary, BorshSerialize, BorshDeserialize, Serialize, Deserialize,
+)]
+pub struct WtxId;
```
storopoli, line 60: This should be moved to `strata-primitives`.
And we are doing the following convention:

```rs
pub struct BitcoinTxid(Txid)
```

Check this example: https://github.com/alpenlabs/strata/blob/8da0760c8629b8020c425e8e7ce8744dea88721a/crates/primitives/src/l1.rs#L442-L534

and this as well which has a custom `impl Deref`: https://github.com/alpenlabs/strata/blob/8da0760c8629b8020c425e8e7ce8744dea88721a/crates/primitives/src/keys.rs#L10-L37

crates/state/src/l1/tx.rs
storopoli: You can't deposit with legacy addresses right? It's only through P2TR.

storopoli (changes requested):

crates/state/src/l1/tx.rs
prajwolrg: > I believe using `L1WtxProof` should suffice because most of the relevant transactions like DA and checkpoint would have data as witness. For the ones that will probably be legacy(deposit txs?) wtxid and txid will be the same. cc: @delbonis

Yes, the `wtxid` and `txid` will be the same for legacy transactions. However, the `L1TxProof` and `L1WtxProof` will be different since `L1TxProof` is verified against `block.header.merkle_root` but `L1WtxProof` is verified against `witness_commitment_from_coinbase(block.txdata[0])`.

prajwolrg: > You can't deposit with legacy addresses right? It's only through P2TR.

Legacy here means transactions without segwit.

crates/state/src/l1/inclusion_proof.rs
prajwolrg: > Also why do we need the `idx`?

The idx is needed to check if the transaction is coinbase. For coinbase transaction wtxid will be Buf32::zero().

crates/state/src/l1/inclusion_proof.rs
prajwolrg: `TxId` and `WtxId` are just used as marker. It doesn't contain more info.

crates/state/src/l1/inclusion_proof.rs
prajwolrg: With this trait, we are computing `Wtxid` or `Txid` of a transaction based on the marker.

storopoli: Still I think should be moved to `strata-primitives` and a better naming

crates/state/src/l1/inclusion_proof.rs
storopoli: Still it needs to be moved to `strata-primitives`.

crates/state/src/l1/inclusion_proof.rs
prajwolrg: Moved the marker types to `strata-primitives` and renamed the trait to `TxIdComputable`.

crates/proof-impl/btc-blockspace/src/block.rs
```diff
@@ -128,53 +156,87 @@ pub fn check_pow(block: &Header) -> bool {
...
-            block.compute_merkle_root().unwrap(),
-            TxMerkleNode::from_byte_array(*compute_merkle_root(&block).unwrap().as_ref())
-        );
+        let empty_inclusion_proof = L1TxProof::new(0, vec![]);
+        assert!(check_integrity(&block, &empty_inclusion_proof));
     }
 
     #[test]
-    fn test_wtx_root() {
+    #[should_panic]
```
bewakes, line 180: Since `check_integrity` returns a bool, it would be better to explicitly check it returns false rather than expecting a panic to occur which can occur for totally different reason, even though that might not be the case here.

crates/proof-impl/btc-blockspace/src/prover.rs
```diff
@@ -20,10 +24,17 @@ impl ZkVmProver for BtcBlockspaceProver {
     where
         B: ZkVmInputBuilder<'a>,
     {
+        let block = &input.block;
+
+        let inclusion_proof = match witness_commitment_from_coinbase(&block.txdata[0]) {
+            Some(_) => L1TxProof::generate(&block.txdata, 0),
+            None => L1TxProof::new(0, vec![]),
```
bewakes, line 31: The use of `L1TxProof::new(0, vec![])` might not be semantically correct because merkle proof of first txn doesn't necessarily always have zero co-hashes. So perhaps `None` can be used when we don't require inclusion proof for coinbase.

bewakes (changes requested):
Looks great broadly. Some comment.

prajwolrg:
> What's the reasoning for this if we're already trusting bitcoin core with consensus?

We need this mostly in the proof. Since we need to prove that we have scanned all the bitcoin transactions so that we have not missed any deposits (or forced inclusion transaction) in the future. This prevents the malicious sequencer from posting random data on Bitcoin. 
The BitVM2 proof is based on this proof.

crates/primitives/src/l1.rs
```diff
@@ -105,28 +105,47 @@ impl From<(u64, u32)> for L1TxRef {
     }
 }
 
-/// TODO: This is duplicate with state::l1::L1TxProof
-/// Merkle proof for a TXID within a block.
-// TODO rework this, make it possible to generate proofs, etc.
+/// A trait for computing some kind of transaction ID (e.g., [`txid`](bitcoin::Txid) or
+/// [`wtxid`](bitcoin::Wtxid)) from a [`Transaction`].
```
storopoli, line 109:
```suggestion
/// A trait for computing some kind of transaction ID (e.g., [`Txid`](bitcoin::Txid) or
/// [`Wtxid`](bitcoin::Wtxid)) from a [`Transaction`].
```

crates/primitives/src/l1.rs
```diff
@@ -105,28 +105,47 @@ impl From<(u64, u32)> for L1TxRef {
...
+///
+/// This trait is designed to be implemented by "marker" types that define how a transaction ID
+/// should be computed. For example, [`TxIdMarker`] invokes [`Transaction::compute_txid`], and
+/// [`WtxIdMarker`] invokes [`Transaction::compute_wtxid`]. This approach avoids duplicating
+/// inclusion-proof or serialization logic across multiple ID computations.
+pub trait TxIdComputable {
+    /// Computes the transaction ID for the given transaction.
+    ///
+    /// The `idx` parameter allows marker types to handle special cases such as the coinbase
+    /// transaction (which has a zero [`wtxid`](bitcoin::Wtxid)) by looking up the transaction
```
storopoli, line 119:
```suggestion
    /// transaction (which has a zero [`Wtxid`](bitcoin::Wtxid)) by looking up the transaction
```

crates/primitives/src/l1.rs
```diff
@@ -105,28 +105,47 @@ impl From<(u64, u32)> for L1TxRef {
...
+    /// index.
+    fn compute_id(tx: &Transaction, idx: usize) -> Buf32;
+}
+
+/// Marker type for computing the [`txid`](bitcoin::Txid).
```
storopoli, line 124:
```suggestion
/// Marker type for computing the [`Txid`](bitcoin::Txid).
```

crates/primitives/src/l1.rs
```diff
@@ -105,28 +105,47 @@ impl From<(u64, u32)> for L1TxRef {
...
-    position: u32,
-    cohashes: Vec<Buf32>,
-}
+pub struct TxIdMarker;
 
-impl L1TxProof {
-    pub fn new(position: u32, cohashes: Vec<Buf32>) -> Self {
-        Self { position, cohashes }
-    }
+/// Marker type for computing the [`wtxid`](bitcoin::Wtxid).
```
storopoli, line 130:
```suggestion
/// Marker type for computing the [`Wtxid`](bitcoin::Wtxid).
```

crates/primitives/src/utils.rs
```diff
@@ -1,33 +1,36 @@
...
 
-use crate::prelude::Buf32;
+use crate::{hash::sha256d, prelude::Buf32};
 
-/// Generates cohashes for an wtxid in particular index with in given slice of wtxids.
+/// Generates cohashes and computes the Merkle root for a transaction ID at a specific index
+/// within a given slice of elements that can be converted to [`Buf32`].
+///
+/// This function supports any type that implements the [`Into<Buf32>`] trait, such as
+/// [`txids`](bitcoin::Txid) or [`wtxids`](bitcoin::Wtxid).
```
storopoli, line 9:
```suggestion
/// [`Txid`s](bitcoin::Txid) or [`Wtxid`s](bitcoin::Wtxid).
```

crates/primitives/src/utils.rs
```diff
@@ -1,33 +1,36 @@
...
 ///
 /// # Parameters
-/// - `wtxids`: The witness txids slice
-/// - `index`: The index of the txn for which we want the cohashes
+///
+/// - `ids`: A slice of ids ([`txids`](bitcoin::Txid) or [`wtxids`](bitcoin::Wtxid)) that can be
```
storopoli, line 13:
```suggestion
/// - `ids`: A slice of ids ([`Txid`s](bitcoin::Txid) or [`Wtxid`s](bitcoin::Wtxid)) that can be
```

crates/proof-impl/btc-blockspace/src/block.rs
```diff
@@ -37,86 +50,109 @@ pub fn check_merkle_root(block: &Block) -> bool {
     }
 }
 
+/// Scans the given coinbase transaction for a witness commitment and returns it if found.
+///
+/// This function iterates over the outputs of the provided `coinbase` transaction from the end
+/// towards the beginning, looking for an output whose `script_pubkey` starts with the “magic” bytes
```
storopoli, line 56:
```suggestion
/// towards the beginning, looking for an output whose `script_pubkey` starts with the "magic" bytes
```

Normal quotes

crates/state/src/l1/inclusion_proof.rs
```diff
@@ -0,0 +1,150 @@
...
+use serde::{Deserialize, Serialize};
+use strata_primitives::{
+    buf::Buf32,
+    hash::sha256d,
+    l1::{TxIdComputable, TxIdMarker, WtxIdMarker},
+    utils::get_cohashes,
+};
+
+/// A generic proof structure that can handle any kind of transaction ID (e.g.,
+/// [txid](bitcoin::Txid) or [`wtxid`](bitcoin::Wtxid)) by delegating the ID computation to the
```
storopoli, line 15:
```suggestion
/// [`Txid`](bitcoin::Txid) or [`Wtxid`](bitcoin::Wtxid)) by delegating the ID computation to the
```

crates/state/src/l1/inclusion_proof.rs
```diff
@@ -0,0 +1,150 @@
...
+    /// for which this proof is generated.
+    position: u32,
+    /// The intermediate hashes (sometimes called "siblings") needed to reconstruct the Merkle root
+    /// when combined with the target transaction's own ID. These are the Merkle tree nodes at
+    /// each step that pair with the current hash (either on the left or the right) to produce
+    /// the next level of the tree.
+    cohashes: Vec<Buf32>,
+    /// A marker that preserves the association with type `T`, which implements
+    /// [`TxIdComputable`]. This ensures the proof logic depends on the correct
+    /// transaction ID computation ([`txid`](bitcoin::Txid) vs.[`wtxid`](bitcoin::Wtxid)) for the
```
storopoli, line 31:
```suggestion
    /// transaction ID computation ([`Txid`](bitcoin::Txid) vs.[`Wtxid`](bitcoin::Wtxid)) for the
```

crates/state/src/l1/inclusion_proof.rs
```diff
@@ -0,0 +1,150 @@
...
+        Buf32::from(cur_hash)
+    }
+
+    /// Verifies the inclusion proof of the given `transaction` against the provided merkle `root`.
+    pub fn verify(&self, transaction: &Transaction, root: Buf32) -> bool {
+        self.compute_root(transaction) == root
+    }
+}
+
+/// Convenience type alias for the [`txid`](bitcoin::Txid)-based proof.
```
storopoli, line 95:
```suggestion
/// Convenience type alias for the [`Txid`](bitcoin::Txid)-based proof.
```

crates/state/src/l1/inclusion_proof.rs
```diff
@@ -0,0 +1,150 @@
...
+pub type L1TxProof = L1TxInclusionProof<TxIdMarker>;
+
+/// Convenience type alias for the [`wtxid`](bitcoin::Wtxid)-based proof.
```
storopoli, line 98:
```suggestion
/// Convenience type alias for the [`Wtxid`](bitcoin::Wtxid)-based proof.
```

storopoli (changes requested):
More minor nits.

storopoli (approved):
ACK 15e6c69ae220050992c029953f39149bb988a300

PR #590, opened 2025-01-06
STR-950 envelope changes

bewakes:
This PR primarily introduces the following changes:
1. Changes the OP_IF..OP_ENDIF envelope structure. ~It still retains version and size of the payload.~
2. Support of multiple L1 payloads bundled as envelopes in a single reveal transaction.
3. Corresponding db changes. Previously, it had 1-1 mapping between payload intent and L1 transaction, now it is many-to-one.
4. Add a bundler task in btcio such that creates `BundledPayloadEntry` and stores in db, which is ultimately published to bitcoin. Currently, it just does 1-1 mapping like it was before, but the bundling logic can be advanced in future.
5. Renames sequencer db to writer db, because actually it contains data related to payloads and bundles.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


crates/consensus-logic/src/duty/worker.rs
```diff
@@ -410,7 +410,7 @@ fn perform_duty<D: Database, E: ExecEngineCtl>(
             Ok(())
         }
         Duty::CommitBatch(data) => {
-            info!(data = ?data, "commit batch");
+            info!(data = ?data, "commit batch duty");
```
delbonis, line 413: `?data`

But really we should not print the contents here, that's unnecessarily verbose, especially for an `info!` log.

crates/l1tx/src/envelope/builder.rs
```diff
@@ -7,43 +7,61 @@ use bitcoin::{
...
+        .push_slice(PushBytesBuf::from(
+            (payload.data().len() as u32).to_be_bytes(),
+        ));
 
-    trace!(batchdata_size = %envelope_data.data().len(), "Inserting batch data");
-    for chunk in envelope_data.data().chunks(520) {
+    // Insert actual data
+    trace!(batchdata_size = %payload.data().len(), "Inserting batch data");
+    for chunk in payload.data().chunks(520) {
         trace!(size=%chunk.len(), "inserting chunk");
```
delbonis, line 50: This is *way* into the internals.  We just have extensive-enough unit tests it so that we never have to worry about if it's correct or not.

Logs are most useful to have closer to the boundaries of components or for high level components doing significant operations the user might want to know about, where we might have to worry about unexpected interactions we want to go look at later.

crates/l1tx/src/envelope/parser.rs
```diff
@@ -19,75 +24,85 @@ pub enum EnvelopeParseError {
...
+
+    // Parse size
+    let size = next_u32(instructions).ok_or(EnvelopeParseError::InvalidSize)?;
+    // Parse payload
+    let payload = extract_n_bytes(size, instructions)?;
+    Ok(L1Payload::new(payload, ptype))
+}
+
+fn parse_payload_type(bytes: &[u8], params: &RollupParams) -> Option<L1PayloadType> {
+    let str = from_utf8(bytes).ok()?;
```
delbonis, line 88: Instead of trying to parse the bytes as UTF-8 first, interpret the strings we're comparing against as bytes and compare that like `tag_bytes == params.checkpoint_tag.as_bytes()`.

Also, don't like var names that shadow type names, would have called this `tag_str` to also be more explicit.

delbonis (changes requested):
This is a good start but what I would like to see is a new RPC that lets us test this new functionality properly and more explicit functional tests for it.  I don't think enough is sufficiently covered.

This also doesn't get to replicating the visitor-based stuff in the old PR, but that wasn't quite ready yet and we're not going to be in a position with tn1 to be able to take advantage of it yet, so I'm not sure if it's worth trying to work on here.  What do you think?

crates/consensus-logic/src/duty/worker.rs
bewakes: True. Just logged the idx.

crates/l1tx/src/envelope/builder.rs
bewakes: Agreed. Removed the trace here.

crates/l1tx/src/envelope/parser.rs
bewakes: That's a good point. Changed to checkping bytes. Conversion to string was really unnecessary.

bewakes:
> but what I would like to see is a new RPC that lets us test this new functionality properly and more explicit functional tests for it. I don't think enough is sufficiently covered.

I see what you are trying to say here. What we don't have exactly is testing of creation of envelopes and accepting them by the client. However, the unit tests do cover parsing multiple envelopes and getting `ProtocolTxOps` out of a transaction. There are functional tests that do cover parsing a checkpoint(the only protocol op that rollup posts to da now). So I think it is sufficient for what functionalities we have now. We could add tests here by introducing new RPCs, but since we also don't actually have multiple envelopes being created, maybe it makes more sense to have those func tests then?

> This also doesn't get to replicating the visitor-based stuff in the old PR

Yes, Initially I was thinking I would do it here, but since this is getting bigger and [another ticket](https://alpenlabs.atlassian.net/browse/STR-402) also touches on this, it might be better to do it there(or split that one also if necessary).
Also, regarding the visitor stuff, I am not sure if I accurately follow how it is going to help. Here's the summary of what I understand and how I am thinking of implementing on a very high level, the problem we're trying to solve being "ability to store da data in separate db for rollup nodes and just keeping the da commitment in chain/protocol operation".

```rust
enum RawProtocolOp {
  Da(vec<u8>),
  Checkpoint(CheckpointData),
  Deposit(..),
  DepositInfo(..),
}

// Actual operations, which prover cares and is in the chain
enum ProtocolOp {
  Da(vec<u8>), // This only has commitment, others will probably stay the same
  Checkpoint(CheckpointData),
  Deposit(..),
  DepositInfo(..),
}

fn process_raw_op(d: RawProtocolOp, f: impl FnOnce(RawProtocolOp) -> Result<()>) -> Result<L1ProtocolOp> {
   let txop = to_tx_op(&d);
   f(d)?;
}
```

Which will be used  in prover and full node as:
```rust
fn extract_tx_ops_prover(tx: Transasction, params: Params, ..) -> Vec<L1ProtocolOp> {
   let raw_ops = extract_raw_protocol_ops(tx, params);

   raw_ops.map(|r| process_raw_op(r, |_| Ok(()))).collect()
}

fn extract_tx_ops_client(tx: Transasction, params: Params, db: Manager) -> Vec<L1ProtocolOp> {
   let raw_ops = extract_raw_protocol_ops(tx, params);

   raw_ops.map(|r| process_raw_op(r, |d| db.insert(d)).collect()
}
```

crates/l1tx/src/filter.rs
```diff
@@ -154,14 +177,17 @@ mod test {
     }
 
     // Create an envelope transaction. The focus here is to create a tapscript, rather than a
-    // completely valid control block
-    fn create_envelope_tx(rollup_name: String) -> Transaction {
+    // completely valid control block. Includes `n_envelopes` envelopes in the tapscript.
+    fn create_checkpoint_envelope_tx(params: Arc<Params>, n_envelopes: u32) -> Transaction {
```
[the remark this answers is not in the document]
bewakes: Oh yes, this function does not absolutely have to take an `Arc`.

crates/btcio/src/writer/bundler.rs
```diff
@@ -0,0 +1,83 @@
...
+use tokio::time::sleep;
+use tracing::*;
+
+// TODO: get this from config
+const BUNDLE_INTERVAL: u64 = 200; // millis
+
+/// Periodically bundles unbundled intents into payload entries.
+pub(crate) async fn bundler_task(ops: Arc<EnvelopeDataOps>) -> anyhow::Result<()> {
+    let mut last_idx = 0;
+    loop {
```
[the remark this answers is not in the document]
bewakes: Totally. Updated.

crates/l1tx/src/envelope/parser.rs
```diff
@@ -19,75 +22,84 @@ pub enum EnvelopeParseError {
...
+    Ok(payloads)
+}
 
-    let extracted_rollup_name = match (tag, name) {
-        (ROLLUP_NAME_TAG, namebytes) => {
-            String::from_utf8(namebytes.to_vec()).map_err(|_| EnvelopeParseError::InvalidNameValue)
-        }
-        _ => Err(EnvelopeParseError::InvalidNameTag),
-    }?;
+fn parse_l1_payload(
```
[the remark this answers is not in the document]
bewakes: That's a very good point you raised, this is definitely a good candidate for fuzzing, especially we are dealing with data that comes from the bitcoin network and also can be large in size(upto 4MB theoretically).

Doing fuzzing for this, however, requires pretty advanced generation of bitcoin scripts(just relying on random bytes won't probably be sufficient). So I think this might be better done on a separate ticket and possibly post Testnet I. What do you think?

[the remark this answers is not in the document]
storopoli: Yes this will probably be mainnet deliverable... Hopefully...

crates/btcio/src/writer/builder.rs
```diff
@@ -52,7 +52,7 @@ pub enum EnvelopeError {
 // dependencies on `tx-parser`, we include {btcio, feature="strata_test_utils"} , so cyclic
 // dependency doesn't happen
 pub async fn build_envelope_txs<W: WriterRpc>(
-    payload: &L1Payload,
+    payloads: &[L1Payload],
```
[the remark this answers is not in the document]
storopoli: I don't have a hard guideline here. I just raised it up to your attention. Since you're not sure also. Let's keep it

storopoli (approved):
ACK d8f04eebecc177b0fe53805d6c0cb1bb288a70ac

delbonis:
@bewakes What your describing there with that snippet of code is just a janky and incomplete version of what the visitor-based design does.  It doesn't avoid the extra allocs for the DA, for example.  It's also important to separate extracting data from the blocks from writing to the database, so that we can't end up with some dirty intermediate state (like if we were to crash partway through `extract_tx_ops_client` there).

bewakes:
@delbonis, let's continue this discussion in the[ other PR](https://github.com/alpenlabs/strata/pull/618) as that's where this is being implemented.

crates/l1tx/src/envelope/builder.rs
```diff
@@ -7,43 +7,58 @@ use bitcoin::{
...
-        .push_slice(PushBytesBuf::try_from(VERSION_TAG.to_vec())?)
+        .push_slice(tag)
+        // Insert version
         .push_slice(PushBytesBuf::from([version]))
-        .push_slice(PushBytesBuf::try_from(BATCH_DATA_TAG.to_vec())?)
-        .push_int(envelope_data.data().len() as i64);
+        // Insert size
+        .push_slice(PushBytesBuf::from(
+            (payload.data().len() as u32).to_be_bytes(),
+        ));
```
delbonis, line 44: There is no version field or need for a length field.

crates/l1tx/src/envelope/builder.rs
```diff
@@ -7,43 +7,58 @@ use bitcoin::{
...
+fn get_payload_type_tag(
+    payload_type: &L1PayloadType,
+    params: &Params,
+) -> anyhow::Result<PushBytesBuf> {
+    Ok(match *payload_type {
+        L1PayloadType::Checkpoint => {
+            PushBytesBuf::try_from(params.rollup().checkpoint_tag.as_bytes().to_vec())?
+        }
+        L1PayloadType::Da => PushBytesBuf::try_from(params.rollup().da_tag.as_bytes().to_vec())?,
+    })
```
delbonis, line 63: Just make this return `&[u8]` and then convert it to the pushbytesbuf in the caller.

crates/l1tx/src/envelope/parser.rs
```diff
@@ -19,75 +22,84 @@ pub enum EnvelopeParseError {
...
     InvalidVersion,
-    /// Does not have a valid blob tag
-    #[error("Invalid/Missing blob tag")]
-    InvalidBlobTag,
-    /// Does not have a valid blob
-    #[error("Invalid/Missing blob tag")]
-    InvalidBlob,
+    // Does not have a valid size
+    #[error("Invalid/Missing size")]
+    InvalidSize,
```
delbonis, line 33: As before, not necessary.

crates/l1tx/src/envelope/parser.rs
```diff
@@ -19,75 +22,84 @@ pub enum EnvelopeParseError {
...
-    InvalidNameTag,
-    /// Does not have a valid name value
-    #[error("Invalid/Missing value")]
-    InvalidNameValue,
-    // Does not have a valid version tag
-    #[error("Invalid/Missing version tag")]
-    InvalidVersionTag,
+    /// Does not have a valid tag
+    #[error("Invalid/Missing tag")]
+    InvalidTag,
```
delbonis, line 27: Specify it's the "type tag" maybe.

crates/l1tx/src/envelope/parser.rs
```diff
@@ -143,7 +147,7 @@ fn extract_n_bytes(
         Ok(data)
     } else {
         debug!("Extracting {} bytes from instructions", size);
```
delbonis, line 149: Unnecessary log here too.  Just write sufficient unit tests.

crates/rocksdb-store/src/writer/db.rs
```diff
@@ -0,0 +1,206 @@
...
+            Some(id) => self
+                .db
+                .get::<IntentSchema>(&id)?
+                .ok_or_else(|| {
+                    DbError::Other(format!(
+                    "Intent index({idx}) exists but corresponding id does not exist in writer db"
+                ))
+                })
+                .map(Some),
+            None => Ok(None),
```
delbonis, line 83: Can flatten out this match with `let Some(...) =`.

delbonis (changes requested):
We don't need a version field because the type tag is also effectively usable as a version field.

bewakes:
> We don't need a version field because the type tag is also effectively usable as a version field.

Hmm.. How's that going to work? I think it is much cleaner to separate version from the type tag. `strata-da-v1`, `strata-da-v2` looks weird if you are thinking of using versions that way.

bewakes:
@delbonis, As discussed in Core sync meeting, I've removed the version from envelope. This should be good to merge now.

delbonis (approved):
This is good for now.

storopoli (approved):
ACK d7dc0a9ff968b5a2ae538ea1d4a2f2cb62d46fa2

prajwolrg (approved):

PR #595, opened 2025-01-07
Refactor/btcio config and other renamings

bewakes:
## Description

This PR is basically renaming and restructuring of btcio configs and does:
1. Rename `inscription` to `envelope` and related renamings
2. Create a new `BtcIOConfig` struct and includes within `Config` to make it reusable over btcio and also removes redundant `ReaderConfig` and `WriterConfig`.
3. Introduces `WriterContext` for btcio writer tasks, which wraps common items like `Arc<Params>`, `Arc<BtcIOConfig>`, `BitcoinClient`, `StatusChannel`, etc. 
4. Rename crate `tx-parser` to `l1tx` which now also contains enveloper builder logic. Previously it only contained parser logic.
5. Move `state::da_blob` module to `primitives::l1::payload`.
6. Split `starata_primitives::l1` module into multiple ones, re-exporting items to keep imports as they are.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


sapinb (dismissed):

crates/db/src/traits.rs
```diff
@@ -246,27 +246,27 @@ pub trait CheckpointDatabase {
...
 /// A trait encapsulating provider and store traits to interact with the underlying database for
-/// [`BlobEntry`]
+/// [`PayloadEntry`]
 pub trait SequencerDatabase {
-    type BlobDB: BlobDatabase;
+    type L1PayloadDB: L1PayloadDatabase;
 
-    fn blob_db(&self) -> &Arc<Self::BlobDB>;
+    fn payload_db(&self) -> &Arc<Self::L1PayloadDB>;
 }
```
delbonis, line 254: Try to remove this if you can.

crates/config/src/btcio.rs
```diff
@@ -0,0 +1,33 @@
...
+    /// How often to poll btc client
+    pub client_poll_dur_ms: u32,
+    /// How often to invoke the writer
+    pub write_poll_dur_ms: u64,
+    /// How the fees for are determined.
+    // FIXME: This should actually be a part of signer.
+    pub fee_policy: FeePolicy,
+    /// How much amount(in sats) to send to reveal address
+    pub reveal_amount: u64,
+}
```
delbonis, line 14: I think it makes sense to split up writer tx assembly settings from the RPC-related config here.

Also, for consistency would make sense to be `Btcio`.

crates/btcio/src/rpc/traits.rs
delbonis, line 1: Should rename these types to have `Rpc` suffixes.  A name like `Writer` implies that it's the thing *doing* the writing rather than being the methods that are *used by* the writer.

crates/rpc/types/src/types.rs
```diff
@@ -83,7 +83,7 @@ pub struct RpcL1Status {
     pub last_published_txid: Option<Txid>,
 
     /// number of published transactions in current run (commit + reveal pair count as 1)
-    pub published_inscription_count: u64,
+    pub published_envelope_count: u64,
```
delbonis, line 86: Make sure this is correct with respect to envelopes vs envelope bundles or rename it to mean what we want.  Also update doc comment.

crates/state/src/da_blob.rs
delbonis, line 1: Should probably rename this to `da` or `da_payload` or something.  Maybe we should move this to primitives too?

crates/state/src/da_blob.rs
```diff
@@ -83,49 +117,49 @@ impl BlobSpec {
...
     }
 
-    /// Generates the spec from the relevant parts of the blob intent that
-    /// uniquely refers to the blob data.
-    pub fn to_spec(&self) -> BlobSpec {
-        BlobSpec {
+    /// Generates the spec from the relevant parts of the payload intent that
+    /// uniquely refers to the payload data.
+    pub fn to_spec(&self) -> PayloadSpec {
+        PayloadSpec {
```
delbonis, line 160: `::new`?

crates/state/src/exec_env.rs
```diff
@@ -19,7 +19,7 @@ pub struct ExecEnvState {
     /// seen on the corresponding DA layer yet.
     ///
     /// This must always be sorted.
-    waiting_da_blobs: Vec<da_blob::BlobSpec>,
+    waiting_da_blobs: Vec<da_blob::PayloadSpec>,
```
delbonis, line 22: This actually should be kept with the terminology and types relating to about blobs.

crates/state/src/exec_update.rs
```diff
@@ -124,7 +124,7 @@ pub struct UpdateOutput {
     /// DA blobs that we expect to see on L1.  This may be empty, probably is
     /// only set near the end of the range of blocks in a batch since we only
     /// assert these in a per-batch frequency.
-    da_blobs: Vec<da_blob::BlobSpec>,
+    da_blobs: Vec<da_blob::PayloadSpec>,
```
delbonis, line 127: Same here.

crates/storage/src/ops/envelope.rs
```diff
@@ -0,0 +1,84 @@
...
+        None => Ok(None),
+    }
+}
+
+fn get_next_payload_idx<D: SequencerDatabase>(ctx: &Context<D>) -> DbResult<u64> {
+    let payload_db = ctx.db.payload_db();
+    payload_db
+        .get_last_payload_idx()
+        .map(|x| x.map(|i| i + 1).unwrap_or_default())
+}
```
delbonis, line 75: This is kinda jank, we should do something to make this more consistent/safe.

delbonis (changes requested):
Good start.

crates/btcio/src/rpc/traits.rs
bewakes: Yes, that's a good point.

crates/l1tx/src/envelope/builder.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    trace!(batchdata_size = %envelope_data.data().len(), "Inserting batch data");
+    for chunk in envelope_data.data().chunks(520) {
+        trace!(size=%chunk.len(), "inserting chunk");
+        builder = builder.push_slice(PushBytesBuf::try_from(chunk.to_vec())?);
+    }
+    builder = builder.push_opcode(OP_ENDIF);
+
+    Ok(builder.into_script())
+}
```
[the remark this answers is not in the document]
bewakes: Actually I will close 590 and create new one on top of this. And that PR will have the actual inscription structure changes.

crates/l1tx/src/envelope/parser.rs
```diff
@@ -3,18 +3,18 @@ use bitcoin::{
...
-use super::utils::{next_bytes, next_int, next_op};
+use crate::utils::{next_bytes, next_int, next_op};
 
 pub const ROLLUP_NAME_TAG: &[u8] = &[1];
 pub const VERSION_TAG: &[u8] = &[2];
 pub const BATCH_DATA_TAG: &[u8] = &[3];
 
 #[derive(Debug, Error)]
-pub enum InscriptionParseError {
+pub enum EnvelopParseError {
```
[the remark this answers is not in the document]
bewakes: Nice catch!

Also, as I think more of it now, is parsing multiple envelopes actually significant? I know it aims to save bitcoin blockspace, which seems more like an optimization at this point. And it comes with its own nuances, like we need to check and handle when multiple envelopes cause the transaction size to exceed standardness limit.

Having said that, we should definitely include the payload type tags in the envelopes appropriately which will be done in the next PR.

crates/rpc/types/src/types.rs
bewakes: As of this PR, a single commit reveal pair contains a single envelope, so this should be fine. Will change accordingly if needed.

crates/state/src/da_blob.rs
bewakes: Oh yes, this I missed.

crates/db/src/traits.rs
bewakes: This should be done while doing sequencer specific refactors. Removing this here might pull up other unexpected things.

crates/config/src/btcio.rs
bewakes: I think it is fine here. We can do that when we actually need it.

> Also, for consistency would make sense to be Btcio.

Yes indeed.

crates/btcio/src/writer/signer.rs
```diff
@@ -1,39 +1,36 @@
...
+    let (commit, reveal) = build_envelope_txs(&payloadentry.payload, ctx.as_ref()).await?;
 
     let ctxid = commit.compute_txid();
     debug!(commit_txid = ?ctxid, "Signing commit transaction");
-    let signed_commit = client
+    let signed_commit = ctx
+        .client
         .sign_raw_transaction_with_wallet(&commit)
         .await
         .expect("could not sign commit tx")
```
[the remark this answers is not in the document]
bewakes: Added a new enum variant.

crates/consensus-logic/src/duty/worker.rs
```diff
@@ -426,14 +426,15 @@ fn perform_duty<D: Database, E: ExecEngineCtl>(
 
             // serialize and send to l1 writer
 
-            let payload =
+            let payload_data =
                 borsh::to_vec(&signed_checkpoint).map_err(|e| Error::Other(e.to_string()))?;
-            let blob_intent = BlobIntent::new(BlobDest::L1, checkpoint_sighash, payload);
+            let payload = L1Payload::new_checkpoint(payload_data);
+            let blob_intent = PayloadIntent::new(PayloadDest::L1, checkpoint_sighash, payload);
```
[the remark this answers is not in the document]
bewakes: Renamed to `checkpoint_hash`.

crates/state/src/da_blob.rs
bewakes: It is in the same module so doesn't really matter much imo.

crates/state/src/exec_env.rs
bewakes: How are we actually differentiating the terminologies "blob" and "payload"?
Maybe we need to rename as `waiting_da_blobs -> waiting_da_payloads`?

crates/storage/src/ops/envelope.rs
bewakes: Yes. Looks ugly. Changed to 
```rust
Ok(payload_db
        .get_last_payload_idx()?
        .map(|i| i + 1)
        .unwrap_or_default())
```

bewakes:
@delbonis thanks for your comments. Addressed most of the comments, left a few as they are.

crates/btcio/src/writer/context.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use bitcoin::Address;
+use strata_config::btcio::BtcioConfig;
+use strata_primitives::params::Params;
+use strata_status::StatusChannel;
+
+use crate::rpc::traits::WriterRpc;
+
+#[derive(Debug, Clone)]
+pub struct WriterContext<W: WriterRpc> {
```
storopoli, line 11: Docstrings and also for the fields

crates/btcio/src/writer/test_utils.rs
```diff
@@ -23,8 +19,8 @@ pub fn get_db() -> Arc<SequencerDB<RBSeqBlobDb>> {
     Arc::new(SequencerDB::new(seqdb))
 }
 
-/// Returns `Arc` of `InscriptionDataOps` for testing
-pub fn get_inscription_ops() -> Arc<InscriptionDataOps> {
+/// Returns `Arc` of `EnvelopeDataOps` for testing
```
storopoli, line 22:
```suggestion
/// Returns an [`Arc`] of [`EnvelopeDataOps`] for testing
```

crates/config/src/btcio.rs
```diff
@@ -0,0 +1,33 @@
+use serde::Deserialize;
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct BtcioConfig {
```
storopoli, line 4: docstring on struct

crates/config/src/btcio.rs
```diff
@@ -0,0 +1,33 @@
...
+    pub write_poll_dur_ms: u64,
+    /// How the fees for are determined.
+    // FIXME: This should actually be a part of signer.
+    pub fee_policy: FeePolicy,
+    /// How much amount(in sats) to send to reveal address
+    pub reveal_amount: u64,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub enum FeePolicy {
```
storopoli, line 17: docstring on enum

crates/l1tx/src/envelope/parser.rs
```diff
@@ -3,18 +3,18 @@ use bitcoin::{
...
-use super::utils::{next_bytes, next_int, next_op};
+use crate::utils::{next_bytes, next_int, next_op};
 
 pub const ROLLUP_NAME_TAG: &[u8] = &[1];
 pub const VERSION_TAG: &[u8] = &[2];
 pub const BATCH_DATA_TAG: &[u8] = &[3];
 
 #[derive(Debug, Error)]
-pub enum InscriptionParseError {
+pub enum EnvelopeParseError {
```
storopoli, line 17: docstring on enum

crates/primitives/src/l1/block.rs
```diff
@@ -0,0 +1,203 @@
...
+    Arbitrary,
+    BorshSerialize,
+    BorshDeserialize,
+    Deserialize,
+    Serialize,
+)]
+pub struct L1BlockId(Buf32);
+
+impl L1BlockId {
+    /// Computes the blkid from the header buf.  This expensive in proofs and
```
storopoli, line 27:
```suggestion
    /// Computes the [`L1BlockId`] from the header buf. This is expensive in proofs and
```

crates/primitives/src/l1/block.rs
```diff
@@ -0,0 +1,203 @@
...
+    /// should only be done when necessary.
+    pub fn compute_from_header_buf(buf: &[u8]) -> L1BlockId {
+        Self::from(crate::hash::sha256d(buf))
```
storopoli, line 30: import the `crate::hash` module on top.

crates/primitives/src/l1/block.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+}
+
+impl From<(u64, u32)> for L1TxRef {
+    fn from(value: (u64, u32)) -> Self {
+        Self(value.0, value.1)
+    }
+}
+
+/// Includes [`L1BlockManifest`] along with scan rules that it is applied to
```
storopoli, line 88:
```suggestion
/// Includes [`L1BlockManifest`] along with scan rules that it is applied to.
```

crates/primitives/src/l1/block.rs
```diff
@@ -0,0 +1,203 @@
...
+        self.record
+    }
+}
+
+/// Describes an L1 block and associated data that we need to keep around.
+// TODO should we include the block index here?
+#[derive(Clone, Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize, Arbitrary)]
+pub struct L1BlockRecord {
+    /// Block hash/ID, kept here so we don't have to be aware of the hash function
+    /// here.  This is what we use in the MMR.
```
storopoli, line 128:
```suggestion
    /// here. This is what we use in the MMR.
```

crates/primitives/src/l1/block.rs
```diff
@@ -0,0 +1,203 @@
...
+    blockid: L1BlockId,
+
+    /// Block header and whatever additional data we might want to query.
+    header: Vec<u8>,
+
+    /// Merkle root for the transactions in the block.  For Bitcoin, this is
```
storopoli, line 134:
```suggestion
    /// Merkle root for the transactions in the block. For Bitcoin, this is
```

crates/primitives/src/l1/block.rs
```diff
@@ -0,0 +1,203 @@
...
+    /// Current tip block ID as string.
+    pub cur_tip_blkid: String,
+
+    /// Last published txid where L2 blob was present
+    pub last_published_txid: Option<Buf32>,
+
+    /// UNIX millis time of the last time we got a new update from the L1 connector.
+    pub last_update: u64,
+
+    /// number of published transactions in current run (commit + reveal pair count as 1)
```
storopoli, line 201:
```suggestion
    /// Number of published transactions in current run (commit + reveal pair count as 1)
```

crates/primitives/src/l1/payload.rs
```diff
@@ -0,0 +1,166 @@
...
+//!
+//! These types don't care about the *purpose* of the payloads, we only care about what's in them.
+
+use arbitrary::Arbitrary;
+use borsh::{BorshDeserialize, BorshSerialize};
+use num_enum::{IntoPrimitive, TryFromPrimitive};
+use serde::{Deserialize, Serialize};
+
+use crate::buf::Buf32;
+/// DA destination identifier.   This will eventually be used to enable
```
storopoli, line 11:
```suggestion
/// DA destination identifier. This will eventually be used to enable
```
OMG triple spaces?

crates/primitives/src/l1/payload.rs
```diff
@@ -0,0 +1,166 @@
...
+    BorshSerialize,
+    IntoPrimitive,
+    TryFromPrimitive,
+    Serialize,
+    Deserialize,
+)]
+#[borsh(use_discriminant = true)]
+#[repr(u8)]
+pub enum PayloadDest {
+    /// If we expect the DA to be on the L1 chain that we settle to.  This is
```
storopoli, line 32:
```suggestion
    /// If we expect the DA to be on the L1 chain that we settle to. This is
```

crates/primitives/src/l1/payload.rs
```diff
@@ -0,0 +1,166 @@
...
+
+/// Manual `Arbitrary` impl so that we always generate L1 DA if we add future
+/// ones that would work in totally different ways.
+impl<'a> Arbitrary<'a> for PayloadDest {
+    fn arbitrary(_u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
+        Ok(Self::L1)
+    }
+}
+
+/// Summary of a DA payload to be included on a DA layer.  Specifies the target and
```
storopoli, line 45:
```suggestion
/// Summary of a DA payload to be included on a DA layer. Specifies the target and
```

crates/primitives/src/l1/payload.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /// Commitment to the payload.
+    pub fn commitment(&self) -> &Buf32 {
+        &self.commitment
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Arbitrary, BorshDeserialize, BorshSerialize)]
+pub struct L1Payload {
```
storopoli, line 82: docstrings for struct

crates/primitives/src/l1/payload.rs
```diff
@@ -0,0 +1,166 @@
...
+            payload,
+        }
+    }
+
+    /// The target we expect the DA payload to be stored on.
+    pub fn dest(&self) -> PayloadDest {
+        self.dest
+    }
+
+    /// Commitment to the payload, which might be context-specific.  This
```
storopoli, line 146:
```suggestion
    /// Commitment to the payload, which might be context-specific. This
```

example_config.toml
```diff
@@ -14,10 +14,14 @@ network = "regtest"
...
 [sync]
 l1_follow_distance = 6
-max_reorg_depth = 4
-client_poll_dur_ms = 200
 client_checkpoint_interval = 10
 
+[btcio]
+client_poll_dur_ms = 200
+write_poll_dur_ms = 200
+fee_policy = "Smart"
```
storopoli, line 22: Maybe we can do a `#[serde(rename_all = "lowercase")]` in the `enum` above so that this can become:

```suggestion
fee_policy = "smart"
```

See: https://serde.rs/container-attrs.html

storopoli (changes requested):
minor nits

crates/primitives/src/l1/payload.rs
bewakes: I have no idea how these got added.

storopoli (approved):
ACK [628a98d](https://github.com/alpenlabs/strata/pull/595/commits/628a98d11eab18e3eb87b19bdc658182551a9444)

crates/l1tx/src/envelope/parser.rs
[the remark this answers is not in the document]
bewakes: In any case, that change can happen in the next PR which might contain the bundling stuffs.

crates/config/src/btcio.rs
delbonis: You could make an argument that we should even be splitting up the reading and writing into different crates anyways.  These fields really have disjoint uses.

crates/consensus-logic/src/duty/worker.rs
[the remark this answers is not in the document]
delbonis: Still needs to be using the payload hash rather than the checkpoint hash.

crates/l1tx/src/envelope/parser.rs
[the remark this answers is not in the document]
delbonis: How do we proceed here wrt discussion earlier today?

crates/rpc/types/src/types.rs
delbonis: We agreed that the commit-reveal assembly logic in the call today will still handle multiple envelopes, so it's not precise.  Being loose with language in scenarios like this causes chaos when we refactor later.

crates/state/src/da_blob.rs
delbonis: If there was no public facing function then sure, but there is.

crates/state/src/exec_env.rs
delbonis: I mentioned this in the comment where I proposed the renaming.

* Envelopes are a type of container that has a tag and a payload.  DA 
* DA is based around blobs identified by `BlobId`s.  DA blobs can be put into envelopes as the payload.

crates/storage/src/ops/envelope.rs
delbonis: I was referring to how we're juggling the concepts of "next" and "last" here between the layers and doing the +1/-1 math here instead of being consistent throughout.

crates/l1tx/src/envelope/parser.rs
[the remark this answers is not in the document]
bewakes: We leave this as it is and make bundle specific changes on the subsequent PR #590 .

crates/state/src/exec_env.rs
bewakes: I see. But what exactly is being suggested here? The [comment](https://github.com/alpenlabs/strata/pull/518#pullrequestreview-2479342341) suggests to use the term `BlobSpec` -> `PayloadSpec`.

> Maybe we need to rename waiting_da_blobs -> waiting_da_payloads?

Does this make sense then?

crates/storage/src/ops/envelope.rs
bewakes: It does make sense for the underlying db to consistently provide the "last" idx and the layer downstream can use it as it needs. Some layers need the last and some need the next, I can't see a problem here.

crates/state/src/da_blob.rs
bewakes: I've added a `::new` method.

crates/consensus-logic/src/duty/worker.rs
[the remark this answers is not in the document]
bewakes: I see. I missed that the payload now has the type, which also needs to be a part of the hash. Updated accordingly. Nice catch.

storopoli (approved):
ACK e22dd0eaad07165a781c01cf7aceb53eb394c9b0

crates/consensus-logic/src/duty/worker.rs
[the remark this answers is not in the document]
bewakes: But wait, this is causing issues now because the payload contains schnorr signature which is different each time and thus even though the checkpoint is the same, it will give different signature and thus different payload hash, and basically same data is being posted to l1 again and again.

crates/rpc/types/src/types.rs
bewakes: Updated to be published_reveal_txs_count.

storopoli (dismissed):
ACK 0face4f155fa1133bf04fca407ee7a3e5f9ebf25

crates/rpc/types/src/types.rs
storopoli: > Updated to be published_reveal_txs_count.

I don't it. I still see it as `published_envelope_count`

crates/consensus-logic/src/duty/worker.rs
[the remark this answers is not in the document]
delbonis: That's clearly a bug then.  We should only sign it once and store it persistently.  If it's already in the database somewhere we shouldn't resign and republish it differently.

crates/config/src/btcio.rs
delbonis: We should do it now while we're changing things so that we don't have more work for ourselves in the future and our design reflects how it actually works.

crates/rpc/types/src/types.rs
delbonis: @storopoli A reveal tx with multiple envelopes won't correctly result this counter to be incremented by the correct amount.

crates/state/src/exec_env.rs
delbonis: In that case we were using "blob" in a broad sense that also included checkpoint data.  I'm trying to get us to be more deliberate with our language on this.

crates/storage/src/ops/envelope.rs
delbonis: The index of the "last element" of a growing sequence isn't well-defined if there's no elements, but "next" is, it's 0.  With "last" we have to special-case where there's no elements, and return `None` or something, meaning the interface can't simply return an index it has to return an `Option<_>` around the index type.

The lowest-level interface for a store like this should usually be providing the "next" index.

crates/l1tx/src/envelope/parser.rs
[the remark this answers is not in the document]
delbonis: Alright.

crates/consensus-logic/src/duty/worker.rs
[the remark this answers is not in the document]
delbonis: This is also why we get spam in the logs about duplicate writer intents which are now no-longer duplicates.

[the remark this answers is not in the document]
bewakes: > That's clearly a bug then.

I see, it might be. Since the duty system does not persist the duties, there's a chance multiple checkpoint intents are being created. This needs a separate ticket to fix/verify. Here's a new ticket https://alpenlabs.atlassian.net/browse/STR-895

crates/rpc/types/src/types.rs
bewakes: Yes, we need separate counter for the envelopes.

crates/storage/src/ops/envelope.rs
bewakes: That calls for another ticket, there are a lot of places where we've exposed the last inserted id from database.

crates/config/src/btcio.rs
bewakes: We will be splitting writer and most likely the signer, which might even exist together in the upcoming main refactor. So, let's not try to anticipate too many things and bother too much on this split here.

bewakes:
This should be good now. I've created tickets for things that are out of scope.

storopoli (dismissed):
ACK 7584ba835d53ad724545ea9185b90a8e9588676e

storopoli:
Needs rebase for `zizmor` to pass

crates/config/src/btcio.rs
delbonis: However it's undeniable that the reader configuration and the writer/signer configuration should be split up.

I'm not approving this PR until you do this.

crates/primitives/src/l1/payload.rs
```diff
@@ -0,0 +1,174 @@
...
+//!
+//! These types don't care about the *purpose* of the payloads, we only care about what's in them.
+
+use arbitrary::Arbitrary;
+use borsh::{BorshDeserialize, BorshSerialize};
+use num_enum::{IntoPrimitive, TryFromPrimitive};
+use serde::{Deserialize, Serialize};
+
+use crate::{buf::Buf32, hash};
+/// DA destination identifier. This will eventually be used to enable
```
delbonis, line 11: Formatting.

crates/primitives/src/l1/payload.rs
```diff
@@ -0,0 +1,174 @@
...
+    BorshDeserialize,
+    BorshSerialize,
+    IntoPrimitive,
+    TryFromPrimitive,
+    Serialize,
+    Deserialize,
+)]
+#[borsh(use_discriminant = true)]
+#[repr(u8)]
+pub enum BlobDest {
```
delbonis, line 31: Should be `PayloadDest`.

crates/primitives/src/l1/payload.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// The payload that matches the commitment.
+    pub fn payload(&self) -> &L1Payload {
+        &self.payload
+    }
+
+    /// Generates the spec from the relevant parts of the payload intent that
+    /// uniquely refers to the payload data.
+    pub fn to_spec(&self) -> BlobSpec {
+        BlobSpec::new(self.dest, self.commitment)
+    }
```
delbonis, line 173: Should be a new type `PayloadSpec`.

crates/primitives/src/l1/proof.rs
```diff
@@ -0,0 +1,50 @@
...
+)]
+pub struct WtxIdMarker;
+
+impl TxIdComputable for TxIdMarker {
+    fn compute_id(tx: &Transaction, _idx: usize) -> Buf32 {
+        tx.compute_txid().into()
+    }
+}
+
+impl TxIdComputable for WtxIdMarker {
```
delbonis, line 42: What is the rationale for the word "marker" here?  "Selector" is probably more appropriate.

I don't see anywhere this is being used.  What purpose does this serve?  Was this included here as part of a dirty rebase?

crates/rocksdb-store/src/sequencer/db.rs
```diff
@@ -27,8 +27,8 @@ impl RBSeqBlobDb {
     }
 }
 
-impl BlobDatabase for RBSeqBlobDb {
-    fn put_blob_entry(&self, blob_hash: Buf32, blob: BlobEntry) -> DbResult<()> {
+impl L1PayloadDatabase for RBSeqBlobDb {
+    fn put_payload_entry(&self, blob_hash: Buf32, blob: PayloadEntry) -> DbResult<()> {
```
delbonis, line 31: `payload_commitment`, `payload_data`

crates/state/src/exec_env.rs
```diff
@@ -19,7 +19,7 @@ pub struct ExecEnvState {
     /// seen on the corresponding DA layer yet.
     ///
     /// This must always be sorted.
-    waiting_da_blobs: Vec<da_blob::BlobSpec>,
+    waiting_da_blobs: Vec<BlobSpec>,
```
delbonis, line 22: This should stay as `BlobSpec`, correct.  Maybe we have to bring it into this crate?

crates/primitives/src/l1/btc.rs
delbonis, line 1: `types` might be a better name for this?  Or `primitives` even?

crates/primitives/src/l1/block.rs
```diff
@@ -0,0 +1,203 @@
...
+        Self {
+            blockid,
+            txs_root: Buf32(root),
+            header,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default, Arbitrary)]
+pub struct L1Status {
```
delbonis, line 180: This doesn't make much sense to be here along with block structural types, move it to a new `status` module maybe.

delbonis (changes requested):

crates/primitives/src/l1/block.rs
bewakes: Agreed.

crates/primitives/src/l1/btc.rs
bewakes: These are bitcoin related primitives so I think it's better to indicate that in the name.

crates/primitives/src/l1/proof.rs
bewakes: This is probably being used in Prover code.

> Was this included here as part of a dirty rebase?

I don't think so. This was here before, and is a part of this PR because I am splitting `primitives/l1.rs`.

crates/consensus-logic/src/duty/worker.rs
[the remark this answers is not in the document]
bewakes: Yeah, there's something going on here. This might even be fixed in the next PR that I am working on regarding the envelopes.

crates/config/src/btcio.rs
bewakes: Alright, splitting this into `Reader` and `Writer` for now. Will think about splitting `Writer` and `Signer` later on. Sounds good?

crates/primitives/src/l1/payload.rs
bewakes: I've added a new type `PayloadSpec` and kept `BlobSpec` as is. As of now both of these have the same structure.
Also, `BlobSpec` has `dest` which is `PayloadDest`, I am not sure if it needs to be another `BlobDest`, which I think is redundant. If you have opinions on this and needs change, please be exact.

crates/state/src/exec_env.rs
bewakes: I am not sure. As mentioned above, I've created another type `PayloadSpec` in the same module.

crates/primitives/src/l1/block.rs
bewakes: And moved to a new `status` module.

delbonis (approved):
Okay I think this is good now.

storopoli (dismissed):
ACK cc85da76aba3b47328ea75caf2a8b209be8105b9

MdTeach (approved):

prajwolrg (dismissed):

sapinb (approved):

storopoli (approved):
ACK 2eda45d4ea685126e3c6159f47d5ae4b701d09bb

voidash (approved):

PR #600, opened 2025-01-08
STR-936: Support `Address`es and `OP_RETURN`s as withdrawal destinations

storopoli:
## Description

Right now the bridge supports withdraws only to P2TR. The user passes an X-only PK to the bridge withdrawal precompile address.

However we need to support any arbitrary address, along with `OP_RETURN`s as valid withdrawal destinations.

### Type of Change

- [x] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

We are not thoroughly testing that the `OP_RETURN` payload is actually being published on-chain (bitcoin). See https://alpenlabs.atlassian.net/browse/STR-919 for context. But the withdraw would fail otherwise, so it's tested-ish.

## Checklist

## Related Issues

STR-936


crates/primitives/src/errors.rs
```diff
@@ -1,21 +1,35 @@
...
     #[error("not a valid point on the curve: {0}")]
     InvalidPoint(Buf32),
+
+    /// The provided Script is not an `OP_RETURN` script.
+    #[error("not an OP_RETURN script: {0}")]
+    NotOpReturn(ScriptBuf),
+
+    /// Unsupported withdrawal destination.
+    #[error("unsupported withdrawal destination: {0}")]
+    UnsupportedDestination(String),
```
delbonis, line 34: Should clarify these enum variants imo.

crates/reth/evm/src/precompiles/bridge.rs
```diff
@@ -59,7 +54,11 @@ impl<DB: Database> ContextStatefulPrecompile<DB> for BridgeoutPrecompile {
         })?;
 
         // Log the bridge withdrawal intent
-        let evt = WithdrawalIntentEvent { amount, dest_pk };
+        let evt = WithdrawalIntentEvent {
+            amount,
+            // FIXME: How to make this work with a BOSD Descriptor.
+            destination,
```
delbonis, line 60: I would still probably pass it around as plain bytes instead of the encoded representation.  Once we've validated it we don't have to introspect it again until the operator gets it.  (Although we'd probably validate it once more when we have a cleaner break between the EVM EE and the Strata chain, still keeping it as bytes though.)

crates/reth/primitives/src/lib.rs
```diff
@@ -22,6 +23,7 @@ sol! {
         /// Withdrawal amount in sats
         uint64 amount,
         /// 32 bytes pubkey for withdrawal address in L1
-        bytes32 dest_pk,
+        // FIXME: How do I use a BOSD Descriptor here?
+        bytes32 destination,
```
delbonis, line 27: `bytes`, unfortunately

This kinda still is messy, we shouldn't be using logs for this imo.

crates/reth/primitives/src/lib.rs
```diff
@@ -13,7 +13,8 @@ pub struct WithdrawalIntent {
     pub amt: u64,
 
     /// Destination public key for the withdrawal
-    pub dest_pk: B256,
+    // FIXME: How do I use a BOSD Descriptor here?
+    pub destination: B256,
```
delbonis, line 17: `Vec<u8>` still, we can avoid validating it until we actually use it and take advantage of the simpler representation elsewhere.

delbonis (changes requested):
Tweaks to representation.

crates/primitives/src/errors.rs
storopoli: These are not needed. Removed in rebase.

storopoli:
Some tests are failing due to the cache being evil and not refusing to let it go (I think)

crates/reth/primitives/src/lib.rs
storopoli: Done with `byte` thanks!

functional-tests/tests/bridge_withdraw_happy_op_return.py
```diff
@@ -0,0 +1,127 @@
...
+        # Get the output of the tx
+        # OP_RETURN is the second output
+        # FIXME: I have no idea how to do this
+        outputs = last_block["txs"][0].as_dict()["outputs"]
+        op_return_output = outputs[1]
+        self.debug(f"OP_RETURN output: {op_return_output}")
+        op_return_script_type = op_return_output["script_type"]
+        assert op_return_script_type == "nulldata", "OP_RETURN not found"
+
+        return True
```
storopoli, line 120: Any ideas on how to test this guy?

delbonis: Ah we should have an RPC that fetches the pending withdrawal intents and withdrawal assignments.  That's probably the *right way* to do this.  Do you want to do this here or create a new PR for it and rebase?

crates/reth/evm/src/precompiles/bridge.rs
```diff
@@ -33,13 +38,13 @@ impl BridgeoutPrecompile {
...
+        destination_bytes: &Bytes,
         _gas_limit: u64,
         evmctx: &mut revm::InnerEvmContext<DB>,
     ) -> PrecompileResult {
-        // Validate the length of the destination public key
-        let dest_pk = try_into_pubkey(dest_pk_bytes)
-            .map_err(|_| PrecompileError::other("Invalid public key length: expected 32 bytes"))?;
+        // Validate that this is a valid BOSD
+        let destination = try_into_bosd(destination_bytes)?;
+        let destination_bytes = destination.to_bytes().into();
```
delbonis, line 47: Does it make sense to reencode it to bytes like this?  Or could we just copy the calldata bytes after we've validated it?  The current feels safer but it feels like maybe it does an extra copy we wouldn't normally have to do.

functional-tests/tests/bridge_withdraw_happy_op_return.py
storopoli: To be honest, we can disable this test and open a new ticket to add the RPC and fix the withdraw OP_RETURN test.
What do you think?
This is touching a LOT of the codebase. I much prefer to defer the functional test to later than have this in a rebase purgatory.

crates/util/python-utils/src/error.rs
```diff
@@ -50,6 +53,9 @@ impl From<Error> for PyErr {
             Error::BitcoinAddress => {
                 PyErr::new::<PyTypeError, _>("Not a valid bitcoin address".to_owned())
             }
+            Error::OpReturnTooLong => {
+                PyErr::new::<PyTypeError, _>("OP_RETURN bigger than 80 bytes".to_owned())
```
delbonis, line 57: There's a lot of repetitive structure here.  Can we rework this so the match arms all return the `&'static str`s and *then* pass it to the `PyErr` constructor only once?

Also does `TypeError` make sense here?  I think `ValueError` is the right thing to use.  Although maybe this doesn't mean what I think it means.

functional-tests/envs/testenv.py
```diff
@@ -156,6 +157,51 @@ def withdraw(
 
         return l2_tx_hash, tx_receipt, total_gas_used
 
+    def withdraw_op_return(
+        self,
+        ctx: flexitest.RunContext,
+        el_address: str,
+        payload: str,
+    ):
```
delbonis, line 165: Is it possible to avoid duplicating a bunch of this logic with the standard `withdraw` function by consolidating some of it into a more general method?

delbonis (changes requested):
Broad strokes looks good, although as the other comment recognized yeah the testing is kinda messy.

functional-tests/envs/testenv.py
storopoli: Yes, I intend to do this with some arbitrary `descriptor` as a single `withdraw` function. Will refactor.

functional-tests/tests/bridge_withdraw_happy_op_return.py
storopoli: https://alpenlabs.atlassian.net/browse/STR-919

functional-tests/envs/testenv.py
storopoli: Done in b0eef7d99e0f88b31a1a5387f8008991efc5a94e

storopoli:
This PR is ready.

Still haven't figured it out how to func-test the `OP_RETURN` payload withdrawals. I think that they are not going through because we are only doing single withdrawals destinations and `OP_RETURN` with values are not standard transaction. We'll need to try it when we have support to multiple withdrawals destinations in a calldata in the bridge precompile withdrawal.

So, the test is disabled now.

crates/primitives/src/l1/btc.rs
```diff
@@ -739,9 +749,24 @@ impl XOnlyPk {
             network,
         ))
     }
+
+    /// Converts [`XOnlyPk`] to [`Descriptor`].
+    pub fn to_descriptor(self) -> Descriptor {
+        let type_tag = DescriptorType::P2tr.to_u8();
+        Descriptor::from_vec([&[type_tag], self.0.as_bytes()].concat()).expect("infallible")
+    }
```
delbonis, line 757: Surely we can do this without allocating.  Is there not a `Descriptor` constructor that takes the xonly pubkey as an arg directly?

crates/reth/evm/src/utils.rs
```diff
@@ -22,19 +23,29 @@ pub fn wei_to_sats(wei: U256) -> (U256, U256) {
...
-                    amt: evt.amount,
-                    dest_pk: evt.dest_pk,
+                .map(|evt| {
+                    let descriptor = Descriptor::from_bytes(&evt.destination);
+                    match descriptor {
+                        Ok(valid_descriptor) => Some(WithdrawalIntent {
+                            amt: evt.amount,
+                            destination: valid_descriptor,
+                        }),
+                        Err(_) => None,
```
delbonis, line 47: This return type change doesn't seem right.  What sense is there returning incrementally `None` here?  We know the precompile will always only emit properly-formed logs.

Also, should we avoid hardcoding the address?

delbonis (changes requested):
Minor comments.

crates/reth/evm/src/utils.rs
bewakes: I agree. we could probably use `filter_map` here and keep the return type same. The caller doesn't gain anything significant with the `None`s imo(except the count of invalid descriptors).

functional-tests/envs/testenv.py
```diff
@@ -113,27 +113,29 @@ def withdraw(
...
         """
         cfg: RollupConfig = ctx.env.rollup_cfg()
         # D BTC
         deposit_amount = cfg.deposit_amount
-        # Build the p2tr pubkey from the withdraw address
-        change_address_pk = extract_p2tr_pubkey(withdraw_address)
-        self.debug(f"Change Address PK: {change_address_pk}")
+        # Assert is a valid BOSD
+        assert is_valid_bosd(destination), "Invalid BOSD"
+        self.debug(f"Withdrawal Destination: {destination}")
```
bewakes, line 129: I thought we were using `self.info` instead of `self.debug`.

functional-tests/tests/bridge_withdraw_happy_op_return.py
```diff
@@ -0,0 +1,115 @@
...
+        # Those blocks will have only coinbase tx for all the empty blocks.
+        # Block with the withdrawal transfer will have at least two transactions.
+        l1_withdraw_block_height = last_block["height"]
+        self.info(f"withdrawal block height on L1: {l1_withdraw_block_height}")
+
+        return True
+
+        # FIXME: Somehow the block does not have the "hello world"
+        # (OP_RETURN LEN 68656c6c6f20776f726c64)
+        # payload. I think that this will be fixed once we support
```
bewakes, line 95: This sounds weird. How though? because then we would be looking at all the outputs?

functional-tests/tests/bridge_withdraw_happy_op_return.py
```diff
@@ -0,0 +1,115 @@
...
+        # multiple withdraw destinations.
+
+        # Get the output of the tx that is not the conibase tx
+        outputs = last_block["txs"][1].as_dict()["outputs"]
+        # OP_RETURN is the first output
+        op_return_output = outputs[1]
```
bewakes, line 101: If it is first then index 0 instead of 1?
```suggestion
        op_return_output = outputs[0]
```

bewakes (commented):
Looks good broadly. Some minor nits and questions.

functional-tests/envs/testenv.py
storopoli: Maybe, I need to rebase don't know when that directive went into effect.

crates/util/python-utils/src/utils.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/// Converts a string to an `OP_RETURN` BOSD [`Descriptor`].
+#[pyfunction]
+pub fn string_to_opreturn_descriptor(s: &str) -> Result<String, Error> {
+    // Encode the string to hex first
+    let string_bytes = s.as_bytes().to_vec();
+    let op_return_bytes = [vec![0], string_bytes].concat();
+    let descriptor =
+        Descriptor::from_bytes(&op_return_bytes).map_err(|_| Error::OpReturnTooLong)?;
```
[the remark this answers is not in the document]
storopoli: Yeah, good suggestion, moved the discussion to the ticket.

functional-tests/tests/bridge_withdraw_happy_op_return.py
storopoli: To be fair, I desperately need some help and ideas on how to make the `bridge_withdraw_happy_op_return` work 😓

storopoli: Light told me yesterday that any `OP_RETURN` <=80bytes is standard even with amounts tied to it...

crates/primitives/src/l1/btc.rs
```diff
@@ -739,9 +749,24 @@ impl XOnlyPk {
             network,
         ))
     }
+
+    /// Converts [`XOnlyPk`] to [`Descriptor`].
+    pub fn to_descriptor(self) -> Descriptor {
+        let type_tag = DescriptorType::P2tr.to_u8();
+        Descriptor::from_vec([&[type_tag], self.0.as_bytes()].concat()).expect("infallible")
```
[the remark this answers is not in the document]
storopoli: https://github.com/alpenlabs/bitcoin-bosd/pull/49

crates/util/python-utils/src/utils.rs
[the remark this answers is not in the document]
storopoli: https://github.com/alpenlabs/bitcoin-bosd/pull/49

crates/primitives/src/l1/btc.rs
storopoli: Blocked by https://github.com/alpenlabs/bitcoin-bosd/pull/49

functional-tests/tests/bridge_withdraw_happy_op_return.py
storopoli: Rebased and done in 8c2764981f74e2b112e4735852654d2cea780515

functional-tests/envs/testenv.py
storopoli: Still `testenv.py` is filled with `self.debug`. To be honest, `self.info` are really important for maybe CI-hosted functional tests.
Resolving this but we can move this to a discussion in Slack if needed.

crates/reth/evm/src/utils.rs
storopoli: Done in 26d6b54c7cb3bff32de1476c245cb76dcffbf6a0

crates/reth/evm/src/utils.rs
```diff
@@ -22,7 +23,11 @@ pub fn wei_to_sats(wei: U256) -> (U256, U256) {
 }
 
 /// Collects withdrawal intents from bridge-out events in the receipts.
-/// Returns a vector of `WithdrawalIntent`.
+/// Returns a vector of optional [`WithdrawalIntent`]s.
```
bewakes, line 26:
```suggestion
/// Returns a vector of [`WithdrawalIntent`]s.
```

bewakes (changes requested):
Minor typo in the docstring.

crates/util/python-utils/src/utils.rs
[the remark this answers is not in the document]
storopoli: Done in f1bd598db7187b072d6dd784a5ba6253154521f9

crates/primitives/src/l1/btc.rs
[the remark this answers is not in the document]
storopoli: Done in f1bd598db7187b072d6dd784a5ba6253154521f9

crates/primitives/src/l1/btc.rs
storopoli: Done in f1bd598db7187b072d6dd784a5ba6253154521f9

crates/reth/evm/src/utils.rs
storopoli: Done in d0f094cb9b2f8aee51466c6ff3848d99f98730b9

storopoli:
@delbonis and @bewakes should be mergeable now.
Only caveat (see the Notes to reviewers in the PR description):

> We are not thoroughly testing that the `OP_RETURN` payload is actually being published on-chain (bitcoin). See https://alpenlabs.atlassian.net/browse/STR-919 for context. But the withdraw would fail otherwise, so it's tested-ish.

delbonis (approved):

storopoli:
Rebased and bumped to `bitcoin-bosd@0.4.0` which has tighter checks on valid X-only PKs for P2TR descriptors.
Tests are failing because we are blindly using an `Arbitrary` for `Buf32` to try to mock `XOnlyPublicKey`s...
Will fix it...

storopoli:
In d24f4addc62bf6d7f5030bd7d64f47b5ac56018d I've fixed the flakiness of some `strata-consensus-logic` unit tests by doubling the buffer size from `ArbitraryGenerator` from `1024` to `2048` bytes

bin/strata-client/src/extractor.rs
```diff
@@ -517,6 +517,17 @@ mod tests {
         (tx.into(), deposit_request)
     }
 
+    /// Generate a valid [`Buf32`] for [`XOnlyPublicKey`].
+    fn generate_valid_xonly_pk_buf32() -> Buf32 {
+        let mut arb = ArbitraryGenerator::new();
+        loop {
```
bewakes, line 523: At first this loop kind of startled me, but then I realized that the probability is ~0.5. :D

bewakes (approved):
Needs a rebase now.

bin/strata-client/src/extractor.rs
storopoli: Exactly. The probability that a random 32-byte buffer is a valid X-only public key is the probability that x³ +7 is a quadratic residue modulo p. Since each x has about a 50% chance, the probability of it being invalid is approximately 50%.

delbonis (approved):

PR #602, opened 2025-01-09
Fix: large amount of logs from fcm and cpu usage

sapinb:
## Description
Due to watch receiver being cloned on every call, `status_channel.wait_for_client_change()` was returning early on every call, causing the seen CPU spike and large number of fcm logs.
After this, the 100% CPU is not observed, and TRACE logs over 5s is reduced from 445k to 14k.  

Note: Also includes a minor qol change for tab-completing single fn tests.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/status/src/status_manager.rs
+++ b/crates/status/src/status_manager.rs
@@ -121,12 +121,8 @@ impl StatusChannel {
         self.receiver.cl.borrow().clone()
     }
 
-    /// Waits until there's a new client state and returns the client state.
-    pub async fn wait_for_client_change(&self) -> Result<ClientState, RecvError> {
-        let mut s = self.receiver.cl.clone();
-        s.changed().await?;
-        let state = s.borrow().clone();
-        Ok(state)
+    pub fn subscribe_client_state(&self) -> watch::Receiver<ClientState> {
+        self.sender.cl.subscribe()
     }
 
     /// Waits until genesis and returns the client state.
--- a/crates/status/src/status_manager.rs
+++ b/crates/status/src/status_manager.rs
@@ -121,6 +121,7 @@ impl StatusChannel {
         self.receiver.cl.borrow().clone()
     }
 
+    /// Create a subscription to the client state watcher.
     pub fn subscribe_client_state(&self) -> watch::Receiver<ClientState> {
         self.sender.cl.subscribe()
     }
--- a/functional-tests/entry.py
+++ b/functional-tests/entry.py
@@ -21,7 +21,7 @@ def main(argv):
     # Filter the prover test files if not present in argv
     if len(argv) > 1:
         # Run the specific test file passed as the first argument (without .py extension)
-        tests = [str(tst).removesuffix(".py") for tst in argv[1:]]
+        tests = [str(tst).removesuffix(".py").removeprefix("tests/") for tst in argv[1:]]
     else:
         # Run all tests, excluding those containing "prover_", unless explicitly passed in argv
         tests = [test for test in all_tests if "prover_" not in test or test in argv]
--- a/functional-tests/entry.py
+++ b/functional-tests/entry.py
@@ -26,12 +26,12 @@ def main(argv):
         # Run all tests, excluding those containing "prover_", unless explicitly passed in argv
         tests = [test for test in all_tests if "prover_" not in test or test in argv]
 
-    btc_fac = factory.BitcoinFactory([12300 + i for i in range(30)])
-    seq_fac = factory.StrataFactory([12400 + i for i in range(30)])
-    fullnode_fac = factory.FullNodeFactory([12500 + i for i in range(30)])
-    reth_fac = factory.RethFactory([12600 + i for i in range(20 * 3)])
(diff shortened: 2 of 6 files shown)


bewakes (dismissed):

Rajil1213 (approved):

storopoli (changes requested):
A minor nit

functional-tests/tests/bridge_deposit_happy.py
```diff
@@ -21,7 +22,7 @@ class BridgeDepositHappyTest(testenv.StrataTester):
     """
 
     def __init__(self, ctx: flexitest.InitContext):
-        ctx.set_env("basic")
+        ctx.set_env(BasicEnvConfig(101))
```
storopoli, line 25: This change is not needed: https://github.com/alpenlabs/strata/blob/df4b71649f4b90aec289f2deb96b38ab7783eee7/functional-tests/entry.py#L45-L47

storopoli (changes requested):

crates/status/src/status_manager.rs
```diff
@@ -121,12 +121,8 @@ impl StatusChannel {
...
     }
 
-    /// Waits until there's a new client state and returns the client state.
-    pub async fn wait_for_client_change(&self) -> Result<ClientState, RecvError> {
-        let mut s = self.receiver.cl.clone();
-        s.changed().await?;
-        let state = s.borrow().clone();
-        Ok(state)
+    pub fn subscribe_client_state(&self) -> watch::Receiver<ClientState> {
+        self.sender.cl.subscribe()
```
storopoli, line 125: Why you removed the docstring? 😭

functional-tests/tests/bridge_deposit_happy.py
storopoli: Maybe the whole 9349e51362d429e077e929f119f8b200a27ed735 can be dropped?

sapinb: This was to make an independent env, a copy of basic env but only for this test run. Checking a hypothesis that the flaky test is because the shared sequencer instance (with multiple tests doing deposits concurrently) causes this check to fail:
```
assert n_deposits_post == n_deposits_pre + 1, "deposit was not registered"
```
Not relying on deposit counts would be a better solution, but I'm not sure if this can be done currently.

storopoli: Oh hang on! Every time that we do `ctx.set_env("basic")` we are not creating a new env? But just reusing the global one that all tests were using?

WTF!?

Anyways, this is not intuitive at all. Can you add a comment? I can definitely see myself or someone in the future opening a PR and saying "Hey I've refactor the tests (again)" and reverting that change.

crates/status/src/status_manager.rs
sapinb: Added docstring here :)

functional-tests/tests/bridge_deposit_happy.py
sapinb: Added a comment noting this copy is keep it independent. Hopefully the tests are not flaky again ...

functional-tests/entry.py
```diff
@@ -21,17 +21,17 @@ def main(argv):
...
-    fullnode_fac = factory.FullNodeFactory([12500 + i for i in range(30)])
-    reth_fac = factory.RethFactory([12600 + i for i in range(20 * 3)])
-    prover_client_fac = factory.ProverClientFactory([12700 + i for i in range(20 * 3)])
-    bridge_client_fac = factory.BridgeClientFactory([12800 + i for i in range(30)])
+    btc_fac = factory.BitcoinFactory([12300 + i for i in range(100)])
+    seq_fac = factory.StrataFactory([12400 + i for i in range(100)])
+    fullnode_fac = factory.FullNodeFactory([12500 + i for i in range(100)])
+    reth_fac = factory.RethFactory([12600 + i for i in range(100 * 3)])
+    prover_client_fac = factory.ProverClientFactory([12900 + i for i in range(100 * 3)])
+    bridge_client_fac = factory.BridgeClientFactory([13200 + i for i in range(100)])
```
storopoli, line 34: Why are we changing these?

sapinb: not enough ports after adding 2 new envs (all envs share this same range of ports, and freed ones from old tests are not reused. This needs to be improved in flexitest)

bewakes (approved):

functional-tests/entry.py
storopoli: Oh these are the ports! Yeah makes sense :)

functional-tests/tests/bridge_deposit_happy.py
delbonis: @storopoli Yeah the idea with being able to share envs is to make both tests faster by avoiding standing up and tearing down the network for each test *and* to be able to run them in a more realistic environment with other activity happening around it.  Some kinds of tests don't make sense to run in a shared env (like genesis) so we specify their config explicitly.

PR #604, opened 2025-01-09
fix(strata-client): get latest checkpointed state for bridge duties

Rajil1213:
## Description

This PR fixes extraction of bridge duties by only extracting duties from the latest checkpointed (finalized) chain state.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/bin/strata-client/src/rpc_server.rs
+++ b/bin/strata-client/src/rpc_server.rs
@@ -114,8 +114,28 @@ impl<D: Database + Sync + Send + 'static> StrataRpcImpl<D> {
         Ok((cs, chs))
     }
 
-    async fn get_last_checkpoint_chainstate(&self) -> Option<Arc<Chainstate>> {
-        self.status_channel.chain_state().map(Arc::new)
+    async fn get_last_checkpoint_chainstate(&self) -> Result<Option<Arc<Chainstate>>, Error> {
+        let client_state = self.status_channel.client_state();
+
+        let Some(last_checkpoint) = client_state.l1_view().last_finalized_checkpoint() else {
+            return Ok(None);
+        };
+        let (_, l2_blockidx) = last_checkpoint.batch_info.l2_range;
+
+        // in current implementation, chainstate idx == l2 block idx
+        let idx = l2_blockidx;
+
+        let db = self.database.clone();
+
+        wait_blocking("load_checkpoint_chainstate", move || {
+            let chainstate_db = db.chain_state_db();
+            let chainstate = chainstate_db
+                .get_toplevel_state(idx)?
+                .ok_or(Error::MissingChainstate(idx))?;
+
+            Ok(Some(Arc::new(chainstate)))
+        })
+        .await
     }
 }
 
@@ -493,15 +513,18 @@ impl<D: Database + Send + Sync + 'static> StrataApiServer for StrataRpcImpl<D> {
         let deposit_duties = deposit_duties.map(BridgeDuty::from);
 
         // withdrawal duties should only be generated from finalized checkpoint states
-        let withdrawal_duties = self
-            .get_last_checkpoint_chainstate()
-            .await
-            .map(|chainstate| {
-                extract_withdrawal_infos(chainstate.deposits_table())
-                    .map(BridgeDuty::from)
-                    .collect::<Vec<_>>()
-            })
-            .unwrap_or_default();
+        let withdrawal_duties =
+            self.get_last_checkpoint_chainstate()
+                .await
+                .map(|chainstate_opt| {
+                    chainstate_opt
+                        .map(|chainstate| {
+                            extract_withdrawal_infos(chainstate.deposits_table())
+                                .map(BridgeDuty::from)
+                                .collect::<Vec<_>>()
+                        })
+                        .unwrap_or_default()
+                })?;
 
         let mut duties = vec![];
         duties.extend(deposit_duties);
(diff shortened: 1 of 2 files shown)


bin/strata-client/src/rpc_server.rs
```diff
@@ -493,15 +513,18 @@ impl<D: Database + Send + Sync + 'static> StrataApiServer for StrataRpcImpl<D> {
...
+                .await
+                .map(|chainstate_opt| {
+                    chainstate_opt
+                        .map(|chainstate| {
+                            extract_withdrawal_infos(chainstate.deposits_table())
+                                .map(BridgeDuty::from)
+                                .collect::<Vec<_>>()
+                        })
+                        .unwrap_or_default()
+                })?;
```
bewakes, line 527: Would using `and_then` simplify the nesting?

bin/strata-client/src/rpc_server.rs
```diff
@@ -114,8 +114,28 @@ impl<D: Database + Sync + Send + 'static> StrataRpcImpl<D> {
...
+    async fn get_last_checkpoint_chainstate(&self) -> Result<Option<Arc<Chainstate>>, Error> {
+        let client_state = self.status_channel.client_state();
+
+        let Some(last_checkpoint) = client_state.l1_view().last_finalized_checkpoint() else {
+            return Ok(None);
+        };
+        let (_, l2_blockidx) = last_checkpoint.batch_info.l2_range;
+
+        // in current implementation, chainstate idx == l2 block idx
+        let idx = l2_blockidx;
```
bewakes, line 126: I see this is mostly for the comment, but perhaps it isn't actually necessary?

bewakes (commented):
Some nits.

bin/strata-client/src/rpc_server.rs
storopoli: I think that both `map` `and_then` would boil down to same performance.
However if you chain a bunch of `and_then`s instead of `map`s is that `map` takes a `FnOnce(T) -> U` whereas `and_then` takes a `FnOnce(T) -> Option<U>`. With `and_then` you can combine several operations which each may individually fail, whereas `map` applies an infallible transformation.

So if you really want to squeeze out everything, maybe `and_then` would be more efficient in an error situation since it "unwinds" faster?

Rajil1213: I think, performance-wise, they're not much different. The compiler should optimize any subtle differences (happy to be proven wrong though). I am concerned about readability but I don't think that is improved either. Here is what I come up with and I don't like the extra `.clone()` call.

```rust
        // withdrawal duties should only be generated from finalized checkpoint states
        let withdrawal_duties = self
            .get_last_checkpoint_chainstate()
            .await
            .map(|maybe_chainstate| {
                maybe_chainstate.map(|chainstate| chainstate.deposits_table().clone())
            })?
            .map(|v| {
                extract_withdrawal_infos(&v)
                    .map(BridgeDuty::from)
                    .collect::<Vec<_>>()
            })
            .unwrap_or_default();
```

storopoli: Ugh, that's nasty. How that would look like with chained `and_then`. I much prefer than the quadratic nested map (probably it is optimized away AF in `release` mode).

Rajil1213: > Ugh, that's nasty. How that would look like with chained and_then

It'll be the same except that the closures should return wrapped values (`Ok()` or `Some()`).

bewakes: This should work, and possibly looks better?
```rust
let withdrawal_duties = self
      .get_last_checkpoint_chainstate()
      .await?
      .map(|chainstate| {
          extract_withdrawal_infos(chainstate.deposits_table())
              .map(BridgeDuty::from)
              .collect::<Vec<_>>()
      })
      .unwrap_or_default();
```

Rajil1213: Yes, that works. Done in https://github.com/alpenlabs/strata/commit/b7ed029d9ff5be85e8d5512fed22ee84ab4d37bd.

bewakes (approved):
In future we should probably have the status channel provide checkpointed state as well. This looks good for now.

storopoli (dismissed):
ACK b7ed029d9ff5be85e8d5512fed22ee84ab4d37bd

functional-tests/tests/bridge_reassignment.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, ctx: flexitest.InitContext):
         ctx.set_env(
             testenv.BasicEnvConfig(
                 101,
-                n_operators=3,
+                n_operators=2,
```
storopoli, line 25: It works now with 2?

storopoli (commented):
ACK e0032f59391b61d1025f08f8ac49e347b52e26ef

storopoli (approved):
ACK e0032f59391b61d1025f08f8ac49e347b52e26ef

functional-tests/tests/bridge_reassignment.py
Rajil1213: Yes it was fixed in the original PR with that test but the number of operators was not changed.

bewakes (approved):

PR #605, opened 2025-01-09
Refactor/l1tx

prajwolrg:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


bewakes (commented):
Looks good, but I want to confirm this with @delbonis. There must have been a reason why `Vec<u8>` was previously chosen for `L1Tx.tx`.

storopoli:
Are we somehow doing some basic checks as in https://docs.rs/bitcoin/0.32.5/bitcoin/consensus/trait.Decodable.html#impl-Decodable-for-Vec%3CTransaction%3E ?

delbonis:
I'm soft against putting nontrivial structural checks which we "don't own" close to the database like this.  The database layer should be concerned with just ensuring that the general shape of the data is *sane*, not that it's semantically correct.

bin/strata-client/src/extractor.rs
```diff
@@ -54,22 +53,7 @@ pub(super) async fn extract_deposit_requests<L1DB: L1Database>(
         .map_err(RpcServerError::Db)?;
 
     let deposit_info_iter = l1_txs.into_iter().filter_map(move |l1_tx| {
-        let mut raw_tx = l1_tx.tx_data();
-        let tx = Transaction::consensus_decode(&mut raw_tx);
-        if tx.is_err() {
```
delbonis, line 59: So for this we could instead a expose a `.try_parse_tx()` fn that returns a `Result<Transaction, Error>`.

Unrelated but I'm still perplexed as to why this module is in this crate, I know it was explained to me once but it's still wrong.

crates/primitives/src/l1.rs
```diff
@@ -983,6 +983,113 @@ impl<'a> Arbitrary<'a> for TaprootSpendPath {
...
+        reader.read_exact(&mut tx_bytes)?;
+
+        // Now parse those bytes with bitcoin consensus
+        let tx = deserialize(&tx_bytes).map_err(|err| Error::new(ErrorKind::InvalidData, err))?;
+
+        Ok(BitcoinTx(tx))
+    }
+}
+
+impl<'a> arbitrary::Arbitrary<'a> for BitcoinTx {
```
delbonis, line 1027: This could be good to keep on a `RawBitcoinTx` type if it's a wrapper around `Vec<u8>`.

delbonis (changes requested):

bin/strata-client/src/extractor.rs
delbonis: `try_into` probably makes more sense actually, even if it's more implicit.

delbonis (dismissed):

crates/primitives/src/l1.rs
```diff
@@ -983,48 +983,30 @@ impl<'a> Arbitrary<'a> for TaprootSpendPath {
...
-        let tx_len = u32::deserialize_reader(reader)? as usize;
-        let mut tx_bytes = vec![0u8; tx_len];
-        reader.read_exact(&mut tx_bytes)?;
-
-        // Now parse those bytes with bitcoin consensus
-        let tx = deserialize(&tx_bytes).map_err(|err| Error::new(ErrorKind::InvalidData, err))?;
-
-        Ok(BitcoinTx(tx))
+impl TryFrom<&RawBitcoinTx> for Transaction {
+    type Error = bitcoin::consensus::encode::Error;
```
storopoli, line 1003: import this above?

```rs
use bitcoin::consensus
```
then

```suggestion
    type Error = consensus::encode::Error;
```

crates/primitives/src/l1.rs
```diff
@@ -983,48 +983,30 @@ impl<'a> Arbitrary<'a> for TaprootSpendPath {
...
 
-/// Implement BorshSerialize using Bitcoin consensus serialization.
-impl BorshSerialize for BitcoinTx {
-    fn serialize<W: Write>(&self, writer: &mut W) -> std::io::Result<()> {
-        // Use bitcoin's consensus serialization
-        let tx_bytes = serialize(&self.0);
-        BorshSerialize::serialize(&(tx_bytes.len() as u32), writer)?;
-        writer.write_all(&tx_bytes)
+impl TryFrom<RawBitcoinTx> for Transaction {
+    type Error = bitcoin::consensus::encode::Error;
```
storopoli, line 996: import this above?
```rs
use bitcoin::consensus
```
then
```suggestion
    type Error = consensus::encode::Error;
```

crates/primitives/src/l1.rs
```diff
@@ -983,6 +983,95 @@ impl<'a> Arbitrary<'a> for TaprootSpendPath {
...
+        deserialize(&value.0)
+    }
+}
+
+impl<'a> arbitrary::Arbitrary<'a> for RawBitcoinTx {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
+        use bitcoin::{
+            blockdata::transaction::{OutPoint, TxIn, TxOut},
+            Sequence, Transaction,
+        };
```
storopoli, line 1014: Why you are importing this here?
Should be up above.

crates/primitives/src/l1.rs
```diff
@@ -983,6 +983,95 @@ impl<'a> Arbitrary<'a> for TaprootSpendPath {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
+pub struct RawBitcoinTx(Vec<u8>);
```
storopoli, line 987: docstring

storopoli (changes requested):

bewakes (approved):

crates/primitives/src/l1/btc.rs
```diff
@@ -740,6 +741,92 @@ impl XOnlyPk {
...
+        };
+
+        Ok(tx.into())
+    }
+}
+
+/// Represents a raw, byte-encoded Bitcoin transaction with custom [`Arbitrary`] support.  
+/// Provides conversions (via [`TryFrom`]) to and from [`Transaction`].
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
+pub struct RawBitcoinTx(Vec<u8>);
```
storopoli, line 828: This should come before the `impl`s. Can you move it up?

crates/primitives/src/l1/btc.rs
```diff
@@ -740,6 +741,92 @@ impl XOnlyPk {
...
+        deserialize(&value.0)
+    }
+}
+
+impl TryFrom<&RawBitcoinTx> for Transaction {
+    type Error = encode::Error;
+    fn try_from(value: &RawBitcoinTx) -> Result<Self, Self::Error> {
+        deserialize(&value.0)
+    }
+}
```
storopoli, line 762: Can you add trivial tests:

- case when the transaction is valid in bytes and the `TryFrom` succeeds
- case when the transaction is invalid in bytes and the `TryFrom` fails

storopoli (changes requested):

crates/primitives/src/l1/btc.rs
prajwolrg: Yea right! Thanks!

bewakes (approved):

PR #616, opened 2025-01-17
STR-953: `btcio` backport (and some new) strata-bridge RPCs

storopoli:
## Description

Backports `strata-bridge::btcio` RPCs to `strata::btcio`.
Needed to remove `strata-bridge`'s `btcio` and use the `strata::btcio` instead.

- [x] `get_current_timestamp`
- [x] `test_mempool_accept`
- [x] `get_tx_out`
- [x] `submit_package`

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

STR-953

### DIFF
--- a/crates/btcio/src/rpc/client.rs
+++ b/crates/btcio/src/rpc/client.rs
@@ -445,11 +445,14 @@ impl SignerRpc for BitcoinClient {
 #[cfg(test)]
 mod test {
 
-    use bitcoin::{consensus, hashes::Hash, NetworkKind};
+    use bitcoin::{consensus, hashes::Hash, Amount, NetworkKind};
     use strata_common::logging;
 
     use super::*;
-    use crate::test_utils::corepc_node_helpers::{get_bitcoind_and_client, mine_blocks};
+    use crate::{
+        rpc::types::{CreateRawTransactionInput, CreateRawTransactionOutput},
+        test_utils::corepc_node_helpers::{get_bitcoind_and_client, mine_blocks},
+    };
 
     #[tokio::test()]
     async fn client_works() {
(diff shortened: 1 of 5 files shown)


storopoli:
@Rajil1213 there are the only ones missing here apart from the superblocks that we're not backporting.

delbonis (approved):

Rajil1213 (changes requested):
Looks good. But since we're at it, we might want to add a few more RPCs that are relevant to the bridge for the Testnet I impl:

* `submit_package` (for CPFP in core 28.0)
* `gettxout`(to check if a particular utxo exists)

This is not too urgent at the moment though so feel free to create separate tickets for these if there are other tasks of higher priority that you might want to be doing instead.

bewakes (approved):

storopoli:
@bibek thanks for the review. I also added the `submit_package` RPC.

crates/btcio/src/rpc/types.rs
```diff
@@ -187,6 +189,79 @@ pub struct ScriptPubkey {
...
+    /// The transaction id.
+    pub txid: String,
+    /// The wtxid of a different transaction with the same txid but different witness found in the
+    /// mempool.
+    ///
+    /// If set, this means the submitted transaction was ignored.
+    #[serde(rename = "other-wtxid")]
+    pub other_wtxid: Option<String>,
+    /// Sigops-adjusted virtual transaction size.
+    pub vsize: Option<i64>,
```
bewakes, line 302: Could this be just `i64` or `u64`? The docs does not say it is optional.

> `"vsize" : n,                  (numeric) Virtual transaction size as defined in BIP 141.`

storopoli: Makes sense, done in 53af954a5166b1db43a14c87c5105d4d3b1d7f49

bewakes (dismissed):

crates/btcio/src/rpc/client.rs
```diff
@@ -470,10 +536,30 @@ mod test {
...
         assert!(got.as_byte_array().len() == 32);
 
+        // get_tx_out
+        let vout = 0;
+        let got = client
+            .get_tx_out(&tx.compute_txid(), vout, true)
+            .await
+            .unwrap();
+        assert_eq!(got.value, 1.0);
+
```
Rajil1213, line 562: This needs an additional test case for when a UTXO is not present.

crates/btcio/src/rpc/client.rs
```diff
@@ -505,4 +591,97 @@ mod test {
...
+        assert_eq!(expected, got);
+
+        let blocks = mine_blocks(&bitcoind, 101, None).unwrap();
+        let last_block = client.get_block(blocks.first().unwrap()).await.unwrap();
+        let coinbase_tx = last_block.coinbase().unwrap();
+
+        let destination = client.get_new_address().await.unwrap();
+        let change_address = client.get_new_address().await.unwrap();
+        let amount = Amount::from_btc(1.0).unwrap();
+        let change_amount = Amount::from_btc(48.999).unwrap(); // 0.0001 fee
```
Rajil1213, line 619: An additional test case that would be useful is to test out the CPFP mechanism by not having any fees on the parent transaction and having the child tx pay for those fees. This also requires that the transaction be of version 3.

Rajil1213 (changes requested):
Nice work! Requesting a few additional test cases though.

crates/btcio/src/rpc/client.rs
Rajil1213: I do have a test case for this in the CPFP thing that I'm working on for the bridge but that uses the client from `corepc-node`. Please feel free to use any of it.

crates/btcio/src/rpc/client.rs
storopoli: Oh I see. You mean not generated by the internal `bitcoind` wallet?

crates/btcio/src/rpc/client.rs
storopoli: Oh, yes I'll gladly steal!

crates/btcio/src/rpc/client.rs
Rajil1213: It could be just be a UTXO that has already been spent.

storopoli: Oh you mean not in the mempool! Yes got it! Way easier...

crates/btcio/src/rpc/client.rs
```diff
@@ -484,6 +569,13 @@ mod test {
         let got = client.get_utxos().await.unwrap();
         assert_eq!(got.len(), 3);
 
+        // get_tx_out not in the mempool anymore
+        let got = client
+            .get_tx_out(&tx.compute_txid(), 0, false)
```
Rajil1213, line 574: I would have thought this would throw an error since the UTXO does not exist anymore. Is that not the case? If so, when _does_ this call fail? For example, if I want to check whether a UTXO exists, how would I do it? I've seen that calling `gettxout` on core returns null if the UTXO is not found. Since we're deserializing to `GetTxOut`, I'd have expected this to  error. Is there a case where this null output is triggered?

Rajil1213 (changes requested):

crates/btcio/src/rpc/client.rs
storopoli: @Rajil1213 `gettxout` is not a Wallet RPC, but a Blockchain one:

image

Are you confusing with `listunspents`?

storopoli: The UTXO will always exist, either in a node's mempool or in a block, until it's spent.

storopoli: This is analogous to `rust-bitcoin`'s `TxOut` that we add as the outputs to `Transaction`s.

Rajil1213: > Are you confusing with listunspents?

This only returns UTXOs if they belong to your own wallet, doesn't it?

I think `gettxout` is the one we can use to check if an arbitrary UTXO exists on chain.

> The UTXO will always exist, either in a node's mempool or in a block, until it's spent.

Yes. I meant to ask you to include a test that runs `gettxout` against a UTXO that no longer exists i.e., it has been spent in my earlier request.

storopoli: Done in 94cac87e998082d8e97434f0924bbc52be6363ed

crates/btcio/src/rpc/client.rs
storopoli: I honestly don't know what's going on, in 63538a618dfca8c4b1dfe4e88258d04f909e1e05 I've pretty much followed line by line of the 1P1C example in https://bitcoinops.org/en/bitcoin-core-28-wallet-integration-guide/

Rajil1213: Responding here too for visibility. You'll need to send in the prevout in the call to `signrawtransactionwithwallet` since the parent tx is not in the mempool/block. As written right now, the child tx is unsigned.

storopoli: Done in 5994041609c2ec35baf49b3866ace7ff5b7ffa52 and 198ba41fdf3c4740c7998335b72832d58b88f24a

Rajil1213 (approved):

storopoli:
@bewakes and @delbonis just look at the last two commits.
These changes were needed to support One Parent One Child Relay (1P1C), see the first example in https://bitcoinops.org/en/bitcoin-core-28-wallet-integration-guide/

delbonis (approved):

bewakes (approved):

PR #618, opened 2025-01-20
Update L1Tx ref types to include multiple proto ops per tx

bewakes:
## Description

This PR does the following:
1. Adds trait `BlockIndexer` that takes L1 blocks and parses/extracts/processes relevant data and `OpsVisitor` that is to be used by a "`BlockIndexer`" which allows for different processing of relevant info found in transactions.
2. There are two implementations of `OpsVisitor`: `ClientOpsVisitor` and `ProverOpsVisitor` which differ in that `ProverOpsVisitor` is intended to just extract DA commitments from transactions efficiently without actually storing/processing the whole da content.
3. Remove the use of `filter_protocol_op_tx_refs` function whose functionality is better provided by `BlockIndexer` and `OpsVisitor`s.
4. Adds unit test that tests multiple protocol ops(checkpoint and deposit in particular) can be parsed from a single transaction. Previously, only one operation could be parsed from a transaction. Also updates tests accordingly.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

STR-935


crates/l1tx/src/filter.rs
```diff
@@ -15,37 +15,50 @@ use crate::{
...
-pub fn filter_protocol_op_tx_refs(
+pub fn filter_protocol_op_tx_refs<F>(
     block: &Block,
     params: &RollupParams,
     filter_config: &TxFilterConfig,
-) -> Vec<ProtocolOpTxRef> {
+    process_raw: &mut F,
+) -> Vec<ProtocolOpTxRef>
+where
+    F: FnMut(&RawProtocolOperation) -> anyhow::Result<()>,
```
delbonis, line 25: You can write this with `&mut impl FnMut(...) -> ...`

crates/state/src/tx.rs
```diff
@@ -17,7 +17,33 @@ pub enum ProtocolOperation {
     DepositRequest(DepositRequestInfo),
     /// Checkpoint data
     Checkpoint(SignedBatchCheckpoint),
-    // TODO: add other kinds like Proofs and statediffs
+    // TODO: add other kinds like proofs and state diffs
+}
+
+/// Similar to [`ProtocolOperation`] except that this also contains blob data which is not relevant
+/// to chain.
```
delbonis, line 24: I don't see DA blobs here.

Also it *is* relevant, we need to see the blob commitments.

I'm a bit confused about what the purpose of this type is, being distinct from the other one.

crates/state/src/state_op.rs
```diff
@@ -139,8 +139,11 @@ fn apply_op_to_chainstate(op: &StateOp, state: &mut Chainstate) {
 
             // TODO add it to the MMR so we can reference it in the future
             let (header_record, deposit_txs, _) = matured_block.into_parts();
-            for tx in deposit_txs {
-                if let Deposit(deposit_info) = tx.tx().protocol_operation() {
+            for op in deposit_txs
+                .into_iter()
+                .flat_map(|x| x.tx().protocol_ops().to_vec())
```
delbonis, line 144: Can avoid alloc here.

crates/l1tx/src/messages.rs
```diff
@@ -18,20 +18,20 @@ pub enum L1Event {
     GenesisVerificationState(u64, HeaderVerificationState),
 }
 
-/// Core protocol specific transaction. It can be thought of as relevant transactions for the
-/// Protocol
+/// Core protocol specific bitcoin transaction reference. A bitcoin transaction can have multiple
+/// operations relevant to protocol. This is used in the context of [`BlockData`].
 #[derive(Clone, Debug, BorshSerialize, BorshDeserialize)]
 pub struct ProtocolOpTxRef {
```
delbonis, line 24: The name of this type doesn't really accurately describe what it represents at this point.  Perhaps a name like `ProtocolTxEntry` now.

delbonis (changes requested):
Need to rework the data structures a little bit to make them more accurately reflect the design we have now.

Also, see the visitor-oriented approach.  The goal with that was to decouple the "figuring out what to do with protocol ops we get" from the "actually finding the protocol ops" parts so that handling can be different depending on if it's in the proof or in the reader.

crates/state/src/tx.rs
bewakes: > Also it is relevant, we need to see the blob commitments.

Yes, the commitments, but not the actual data/blob which is what I ment in the docstring.

bewakes: > 'm a bit confused about what the purpose of this type is, being distinct from the other one.

So the idea here is that `RawProtocolOp` would contain a variant `Da(Vec<u8>)` whereas `ProtocolOp` would have the counterpart `DaCommitment(Buf32)`. The other variants would be the same. Those are not here right now because we don't actually have DA at the moment.

crates/l1tx/src/messages.rs
bewakes: Good point, the name is not accurately reflective of what it holds. I would go for `ProtocolOpsTxRef` because it still is a transaction ref and contains multiple `ProtocolOp`s now.

crates/state/src/tx.rs
delbonis: Just do the visitor thing.  The approach with the handler function is just an roundabout and inelegant way of doing the same general idea.

crates/l1tx/src/messages.rs
delbonis: I would go for something like `ProtocolTxEntry` since it doesn't refer to a tx on it own like `TxRef` does (as that includes the block height).  This type doesn't make sense to include the block height since it's already being presented as the contents of a block and all of the other entries in the list would have the same block height.

crates/state/src/tx.rs
bewakes: PTAL

delbonis (changes requested):
(Discussed in core sync, leaving review here to continue work.)

bewakes:
Updated the description and made changes.

crates/btcio/src/reader/ops_visitor.rs
```diff
@@ -0,0 +1,316 @@
...
+    batch::SignedBatchCheckpoint,
+    tx::{DepositInfo, DepositRequestInfo, ProtocolOperation},
+};
+
+/// Ops visitor for rollup client. This is intended to get Da commitment as well as process/store da
+/// blob.
+#[derive(Clone, Debug)]
+pub struct ClientOpsVisitor {
+    ops: Vec<ProtocolOperation>,
+    // TODO: Add l1 manager to store da to db
```
delbonis, line 14: This would be increasing coupling.  This should be done as a separate stage *after* we've run the initial pass.

crates/btcio/src/reader/ops_visitor.rs
```diff
@@ -0,0 +1,316 @@
...
+        self.ops
+    }
+
+    fn visit_da<'a>(&mut self, chunks: impl Iterator<Item = &'a [u8]>) {
+        let mut hasher = Sha256::new();
+        for chunk in chunks {
+            hasher.update(chunk);
+        }
+        let hash: [u8; 32] = hasher.finalize().into();
+        // TODO: store da in db
```
delbonis, line 34: So this would push into a list of DA blobs that we store later.

crates/btcio/src/reader/ops_visitor.rs
```diff
@@ -0,0 +1,316 @@
...
+}
+
+impl ClientOpsVisitor {
+    pub fn new() -> Self {
+        Self { ops: Vec::new() }
+    }
+}
+
+impl OpsVisitor for ClientOpsVisitor {
+    fn collect(self) -> Vec<ProtocolOperation> {
```
delbonis, line 24: This would probably be a different type that includes the DA somehow.

crates/l1tx/src/filter/visitor.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    fn index_block(mut self, block: &Block, config: &TxFilterConfig) -> Self
+    where
+        Self: Sized,
+    {
+        for (i, tx) in block.txdata.iter().enumerate() {
+            self.index_tx(i as u32, tx, config);
+        }
+        self
+    }
```
delbonis, line 44: Do we have an idea of what purposes we'd have for overriding this?  It would simplify the interface to have it split out as a separate toplevel function closer to where we have the control logic for the scanning.

crates/l1tx/src/filter/visitor.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    fn collect(self) -> Vec<ProtocolOperation>;
+}
+
+pub trait BlockIndexer {
+    type Output;
+
+    fn collect(self) -> Self::Output;
+
+    fn index_tx(&mut self, txidx: u32, tx: &Transaction, config: &TxFilterConfig);
```
delbonis, line 34: Can you document the design for how these traits work and are expected to be used?  What purpose is there for these being separate traits?  Why does one have a impl-defined return type for `collect` but the other bakes in `Vec<ProtocolOperation>`?  Could these be consolidated into one?

Similarly, why is the `TxFilterConfig` being passed as an arg instead of being a member field that the indexer/visitor has itself.  Maybe we could give it a constructor that takes the filter config instead of passing it each time.

crates/state/src/tx.rs
```diff
@@ -17,7 +20,9 @@ pub enum ProtocolOperation {
     DepositRequest(DepositRequestInfo),
     /// Checkpoint data
     Checkpoint(SignedBatchCheckpoint),
-    // TODO: add other kinds like Proofs and statediffs
+    // Da Commitment
+    DaCommitment(Buf32),
+    // TODO: add other kinds like proofs and state diffs
```
delbonis, line 25: The DA *is* the state diffs.

delbonis (changes requested):
This is a good start but we should still keep iterating on the indexing.   There's a few critiques I lined out.  It feels like there's a lot closer we get closer to an optimal point in the design space.

Something that's a red flag to me is that the visitor types between the reader and client versions are *almost* identical.  This points me to thinking that we could simplify things more here and consolidate *some* logic, but I'm not sure what that would look like.  Perhaps through nesting/composition?

crates/l1tx/src/filter/visitor.rs
bewakes: Oops, totally forgot about the docs.

> Why does one have a impl-defined return type for collect but the other bakes in `Vec<ProtocolOperation>`?

The idea is that being an `OpsVisitor`, it makes more sense to bake `ProtocolOperation` into it. Howerver, for block indexer, it could extract anything from a block and hence that has an associated type. But from your comment above, we want to process da in a separate pass, so this might actually change.


>  Maybe we could give it a constructor that takes the filter config instead of passing it each time.

We could do that and I had thought about it but that feels like enforcing structure of the trait instance. So I made this method to take required context.

crates/l1tx/src/filter/visitor.rs
bewakes: I don't actually see any reason for overriding this as of now. I see what you mean here but this default implementation just looks nicely encapsulated/namespaced in the trait.

crates/l1tx/src/filter/visitor.rs
delbonis: > We could do that and I had thought about it but that feels like enforcing structure of the trait instance. So I made this method to take required context.

Perhaps, but then it makes it appear to be sensible to pass *different* filter configs on successive invocations, which it's *very not*.

crates/btcio/src/reader/tx_indexer.rs
```diff
@@ -1,51 +1,73 @@
...
+        let mut blob = Vec::new();
+
         for chunk in chunks {
             hasher.update(chunk);
+            blob.extend_from_slice(chunk);
         }
         let hash: [u8; 32] = hasher.finalize().into();
-        // TODO: store da in db
-        self.ops.push(ProtocolOperation::DaCommitment(hash.into()));
+        let commitment: Buf32 = hash.into();
```
delbonis, line 42: Abstract this out to use the DA commitment type directly.  Avoid duplication for sensitive things like this, even though there's not much space to mess up.

crates/l1tx/src/filter/indexer.rs
```diff
@@ -23,14 +23,17 @@ pub trait OpsVisitor {
...
 }
 
 pub trait BlockIndexer {
-    type Output;
-
-    fn collect(self) -> Self::Output;
-
     fn index_tx(&mut self, txidx: u32, tx: &Transaction, config: &TxFilterConfig);
 
     fn index_block(mut self, block: &Block, config: &TxFilterConfig) -> Self
```
delbonis, line 39: The `index_tx` fn takes `&mut self` as would be expected, but why does this consume self?

crates/l1tx/src/filter/indexer.rs
```diff
@@ -23,14 +23,17 @@ pub trait OpsVisitor {
...
 
-    fn collect(self) -> Vec<ProtocolOperation>;
+    // Collect data
+    fn collect(
+        self,
+    ) -> (
+        Vec<ProtocolOperation>,
+        Vec<DepositRequestInfo>,
+        Vec<DaEntry>,
+    );
```
delbonis, line 33: This is basically an implicit structure, having a complex tuple like this suggests it could be a struct.

crates/l1tx/src/messages.rs
```diff
@@ -45,6 +46,21 @@ impl ProtocolTxEntry {
...
+pub struct DaEntry {
+    #[allow(unused)]
+    commitment: Buf32,
+    #[allow(unused)]
+    blob: Vec<u8>,
+}
+
+impl DaEntry {
+    pub fn new(commitment: Buf32, blob: Vec<u8>) -> Self {
+        Self { commitment, blob }
```
delbonis, line 60: It doesn't *really* make sense to take both of these args separately, since the commitment can be computed from the blob.

crates/proof-impl/btc-blockspace/src/tx_indexer.rs
```diff
@@ -40,6 +36,16 @@ impl OpsVisitor for ProverOpsVisitor {
...
+
+    fn collect(
+        self,
+    ) -> (
+        Vec<ProtocolOperation>,
+        Vec<strata_state::tx::DepositRequestInfo>,
+        Vec<strata_l1tx::messages::DaEntry>,
+    ) {
+        (self.ops, Vec::new(), Vec::new())
+    }
```
delbonis, line 48: See other comment about implied structs.

delbonis (changes requested):

crates/l1tx/src/filter/visitor.rs
bewakes: I agree, but that applies to every function with arguments that runs on a loop or a series of functions invocations that need to be passed the same object. Having those arguments makes it explicit that the method is supposed to primarily use those to extract information.

crates/l1tx/src/filter/indexer.rs
bewakes: Oops, it doesn't have to.

bewakes: Oh wait, this was done to be able to chain the `.collect()` method. Or do you think it should just return `Self::Output` and get rid of the `collect()` method since we would always be `collect`ing it?

crates/l1tx/src/messages.rs
bewakes: Yes, but shis prevents re-computation of the commitment where it is already computed at the caller.

crates/l1tx/src/filter/indexer.rs
bewakes: Yeah, created a one called `L1TxExtract`.

crates/l1tx/src/messages.rs
delbonis: In contexts where we want the blob contents already it's cheap to hash it.  At least change this to `new_unchecked` to signal that it doesn't validate that they match.

crates/l1tx/src/filter/indexer.rs
```diff
@@ -0,0 +1,111 @@
...
+        &self.dep_reqs
+    }
+
+    pub fn da_entries(&self) -> &[DaEntry] {
+        &self.da_entries
+    }
+}
+
+impl<V: TxIndexer + Clone> BlockIndexer for OpIndexer<V> {
+    fn index_tx(&mut self, txidx: u32, tx: &Transaction, config: &TxFilterConfig) {
```
delbonis, line 79: Once again, this it doesn't make any sense to take the config as an argument.  It only makes sense for it to be a field on the thing that makes the basic calls to the other utility fns that need it for extracting data from txs.

crates/l1tx/src/filter/indexer.rs
```diff
@@ -0,0 +1,111 @@
...
+        let mut visitor = self.visitor.clone();
```
delbonis, line 80: This clone is a red flag.  It should be able to just keep accumulating internally.

crates/l1tx/src/filter/indexer.rs
```diff
@@ -0,0 +1,111 @@
...
+            self.tx_entries.push(entry);
+        }
+
+        self.dep_reqs.extend_from_slice(tx_extract.deposit_reqs());
+        self.da_entries.extend_from_slice(tx_extract.da_entries());
+    }
+
+    fn collect(self) -> L1BlockExtract {
+        L1BlockExtract::new(self.tx_entries, self.dep_reqs, self.da_entries)
+    }
```
delbonis, line 110: This is also screwy because there's a lot of cloning happening.  The final `collect` call could call `collect` on the internal visitor just once before packaging in its own block-level data.

crates/l1tx/src/filter/indexer.rs
delbonis: Chaining isn't useful here, that's not how visitors are typically used.

delbonis (changes requested):
Also I don't think `collect` is the right name here, it makes it look like it's something like an iterator which it's not at all.  Using something like `finalize` or `finish` gives the right intuition.

crates/l1tx/src/filter/indexer.rs
bewakes: I don't think it is a red flag.
Perhaps the name should have been `tx_indexer`. This is an indexer for a single transaction it does accumulate the protocol ops present within a single transaction as you can see with the `visit_*` calls afterwards. The accumulated ops are then constructed into `ProtocolTxEntry`. So, from the perspective of a tx_indexer, it is actually accumulating internally which are collected for every transaction. And from the perspective of a block_indexer, it is also doing the same.

Having said that, if you have _concrete_ suggestions, happy to hear about them.

crates/l1tx/src/filter/indexer.rs
bewakes: Changed a bit to avoid clones.

crates/l1tx/src/filter/indexer.rs
bewakes: The intent with this was to have the API explicitly imply that indexing should happen based on the `TxFilterConfig`. But as I think more about it now, it should be fine even if we bake that into the `OpIndexer`.

crates/l1tx/src/filter/indexer.rs
bewakes: I don't think this has much to do with how visitors work. This totally respects how the visiting and accumulation works under the hood. It's just a matter of choice for finalizing the visitor imo. I prefer

```rust
let data = dataOpsIndexer::new()
    .index_block()
    .finalize();
```
to
```rust
let mut indexer = OpsIndexer::new();
indexer.index_block(&block);
let data = indexer.finalize();
```
We don't intend to do anything else with the indexer after calling `index_block()` except using the finalized data. So I think this chaining is fine.

delbonis: You might prefer it but it's a lot harder to understand the *purpose* of this chaining in context.

It *would* make sense if, by consuming the indexer, it returned the digested output, which then needed some finalization operation applied to it.  This is what's happening with iterator combinators, which wrap inner type with the combinator implementation type that exposes an implementation of `Iterator` based on the inner iterator type which is the next step up the chain.  But that's not what's happening here.

delbonis (approved):

PR #627, opened 2025-01-22
Str 867 update l1 reader worker

bewakes:
## Description

This PR does the following:
1. Use `L1BlockManager` instead of `Database` in reader task.
2. Remove `consensus-logic` dependency in `btcio`(thereby avoiding a possible cyclic dependency) by introducing a trait `EventSubmitter` which provides interface to submit events to csm.
3. Remove `l1_reader` mod from `strata-client` bin and move the init logic to `btcio/reader`.
4. Remove `l1_handler` from `consensus-logic` and rename it to `checkpoint_verification` which now contains only `verify_proof` function. The rest of the l1 persistent logic that was present there has been moved to `btcio/reader`.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist


crates/consensus-logic/src/util.rs
```diff
@@ -0,0 +1,55 @@
...
+        RollupVerifyingKey::Risc0VerifyingKey(vk) => {
+            strata_risc0_adapter::verify_groth16(proof, vk.as_ref(), public_params_raw)
+        }
+        RollupVerifyingKey::SP1VerifyingKey(vk) => {
+            strata_sp1_adapter::verify_groth16(proof, vk.as_ref(), public_params_raw)
+        }
+        // In Native Execution mode, we do not actually generate the proof to verify. Checking
+        // public parameters is sufficient.
+        RollupVerifyingKey::NativeVerifyingKey(_) => Ok(()),
+    }
```
delbonis, line 54: We should probably abstract this out some more so we don't have to do this dispatch everywhere we verify a proof.

delbonis (changes requested):
This might change a fair bit after the chainstate manager PR gets merged.

crates/btcio/src/reader/handler.rs
```diff
@@ -0,0 +1,163 @@
...
+};
+use tracing::*;
+
+use super::query::ReaderContext;
+use crate::rpc::traits::ReaderRpc;
+
+pub(crate) async fn handle_bitcoin_event<R: ReaderRpc, E: EventSubmitter>(
+    event: L1Event,
+    ctx: &ReaderContext<R>,
+    event_submitter: &E,
```
delbonis, line 22: Can do `&impl EventSubmitter`, but it would still be better to use an observer-like system.  This is sortaaa decoupled because there's the trait but this is actually just obscuring a tighter dependency.

crates/btcio/src/reader/handler.rs
```diff
@@ -0,0 +1,163 @@
...
+                .await?;
+            info!(%height, %l1blkid, txs = %num_txs, "wrote L1 block manifest");
+
+            // Write to sync event db if it's something we care about.
+            let blkid: Buf32 = blockdata.block().block_hash().into();
+            let ev = SyncEvent::L1Block(blockdata.block_num(), blkid.into());
+            event_submitter.submit_event_async(ev).await?;
+
+            // Check for da batch and send event accordingly
+            debug!(?height, "Checking for da batch");
```
delbonis, line 70: `%height`

delbonis (changes requested):

crates/btcio/src/reader/handler.rs
bewakes: Oh yes, I could have done `&impl EventSubmitter`.

> it would still be better to use an observer-like system

It would possibly be better, but that should be done in a separate ticket, because that requires broader design discussions.

bewakes: Do you think we can have a better name though?

Rajil1213:
I can see that the `elf`'s are also part of the diff. Is this intentional?

crates/btcio/src/reader/handler.rs
delbonis: The name is fine.

delbonis: Alright I guess since there's more CSM stuff due to be refactored in the mid future we can just leave it as-is.

delbonis:
^That's a good question.

delbonis (approved):
I'm approving this tentatively in case it's intentional.

bewakes:
@delbonis @Rajil1213 actually those elfs are totally unintentional. I was trying to do prover evaluation for some reasons and that generated the elfs. Removed now. Great catch @Rajil1213 !

delbonis (approved):

delbonis:
Force merging since we're all good to go now and don't need to re-review.

PR #633, opened 2025-01-23
STR-949 chainstate manager v2 and minor chainstate restructures

delbonis:
## Description

This is a second attempt at doing the chainstate manager now with some better idea of the context, plus some additional types to support work in the other PR.  I expect some more things will be copied over that are related and minimally invasive.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/consensus-logic/src/duty/worker.rs
+++ b/crates/consensus-logic/src/duty/worker.rs
@@ -1,0 +1,454 @@
+//! Executes duties.
+
+use std::{
+    collections::HashMap,
+    sync::{Arc, RwLock},
+    time::{self},
+};
+
+use strata_btcio::writer::EnvelopeHandle;
+#[cfg(feature = "debug-utils")]
+use strata_common::bail_manager::{check_bail_trigger, BAIL_DUTY_SIGN_BLOCK};
+use strata_crypto::sign_schnorr_sig;
+use strata_db::traits::*;
+use strata_eectl::engine::ExecEngineCtl;
+use strata_primitives::{
+    buf::{Buf32, Buf64},
+    l1::payload::{L1Payload, PayloadDest, PayloadIntent},
+    params::Params,
+};
+use strata_state::{batch::SignedBatchCheckpoint, client_state::ClientState, prelude::*};
+use strata_storage::{L2BlockManager, NodeStorage};
+use strata_tasks::{ShutdownGuard, TaskExecutor};
+use tokio::sync::broadcast;
+use tracing::*;
+
+use super::{
+    block_assembly, extractor,
+    types::{self, Duty, DutyBatch, Identity, IdentityKey},
+};
+use crate::{
+    checkpoint::CheckpointHandle,
+    csm::message::{ClientUpdateNotif, ForkChoiceMessage},
+    duty::checkpoint::check_and_get_batch_checkpoint,
+    errors::Error,
+    sync_manager::SyncManager,
+};
+
+pub fn duty_tracker_task<D: Database>(
+    shutdown: ShutdownGuard,
+    cupdate_rx: broadcast::Receiver<Arc<ClientUpdateNotif>>,
+    batch_queue: broadcast::Sender<DutyBatch>,
+    ident: Identity,
+    database: Arc<D>,
+    storage: Arc<NodeStorage>,
+    params: Arc<Params>,
+) -> Result<(), Error> {
+    let db = database.as_ref();
+    duty_tracker_task_inner(
+        shutdown,
+        cupdate_rx,
+        batch_queue,
+        ident,
+        db,
+        storage.as_ref(),
+        params.as_ref(),
+    )
+}
+
+fn duty_tracker_task_inner(
+    shutdown: ShutdownGuard,
+    mut cupdate_rx: broadcast::Receiver<Arc<ClientUpdateNotif>>,
+    batch_queue: broadcast::Sender<DutyBatch>,
+    ident: Identity,
+    database: &impl Database,
+    storage: &NodeStorage,
+    params: &Params,
+) -> Result<(), Error> {
+    let mut duties_tracker = types::DutyTracker::new_empty();
+
+    let idx = database.client_state_db().get_last_checkpoint_idx()?;
+    let last_checkpoint_state = database.client_state_db().get_state_checkpoint(idx)?;
+    let last_finalized_blk = match last_checkpoint_state {
+        Some(state) => state.sync().map(|sync| *sync.finalized_blkid()),
+        None => None,
+    };
+    duties_tracker.set_finalized_block(last_finalized_blk);
+
+    // TODO: figure out where the l1_tx_filters_commitment is stored
+    // Maybe in the chain state?
+    let rollup_params_commitment = params.rollup().compute_hash();
+
+    loop {
+        if shutdown.should_shutdown() {
+            warn!("received shutdown signal");
+            break;
+        }
+        let update = match cupdate_rx.blocking_recv() {
+            Ok(u) => u,
+            Err(broadcast::error::RecvError::Closed) => {
+                break;
+            }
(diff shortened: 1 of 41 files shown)


crates/primitives/src/epoch.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// Ie. if the terminal blkid is zero then the last slot and epoch number
+    /// are also zero.
+    fn sanity_check(&self) -> bool {
+        if self.is_null() {
+            self.last_slot == 0 && self.epoch == 0
+        } else {
+            self.last_slot != 0
+        }
+    }
+}
```
AaronFeickert, line 94: Why not include this check in the constructor?

delbonis: Maybe I should just get rid of the sanity check altogether.  This special casing isn't well-defined at the moment since it's not actually used in this context.

AaronFeickert: Would it work to avoid the sentinel value and use an `Option` instead? Or does that screw up the commitment call...

delbonis:
Really happy with the general refactoring I did to the CSM logic, this will hopefully make it simpler to make work when rebasing the changes in the other branch.

crates/primitives/src/epoch.rs
delbonis: It'd be nice to always have a well-defined finalized block so that we don't have to do special casing just for the genesis.  Maybe we can rework the interfaces/definitions around this.

delbonis:
Still have to simplify the actual db impls to remove the now-unnecessary chainstate fields.

crates/db/src/traits.rs
```diff
@@ -194,39 +193,43 @@ pub enum BlockStatus {
...
+/// use the unmerged write batches to construct a view of more recent states
+/// than the fully materialized state in-memory.
+///
+/// For now, the full state is just the "toplevel" state that can always be
+/// expected to be of moderate size in memory.
+// TODO maybe rewrite this around storing write batches according to blkid?
 pub trait ChainstateDatabase {
     /// Writes the genesis chainstate at index 0.
-    fn write_genesis_state(&self, toplevel: &Chainstate) -> DbResult<()>;
+    fn write_genesis_state(&self, toplevel: Chainstate) -> DbResult<()>;
```
bewakes, line 211: Won't passing the `Chainstate` instead of `&Chainstate` here incur an extra `clone()` on the caller side?

bewakes (commented):
Great work, looks much cleaner.

crates/test-utils/src/lib.rs
```diff
@@ -7,7 +7,7 @@ pub mod evm_ee;
 pub mod l2;
 
 /// The default buffer size for the `ArbitraryGenerator`.
-const ARB_GEN_LEN: usize = 1024;
+const ARB_GEN_LEN: usize = 16384;
```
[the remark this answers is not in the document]
storopoli: No it does not. It is flaky with `1024`. I had to bump to `2_048` and was still getting flaky tests with `cargo-neatest`. Maybe:

```suggestion
const ARB_GEN_LEN: usize = 4_096;
```

That should eliminate most flakiness. And you should use `cargo-neatest` to test stuff locally.

storopoli (approved):
ACK 164930d99bfad858069446fe7454784f311d123a

crates/test-utils/src/lib.rs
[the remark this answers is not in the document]
delbonis: 16kB of entropy still takes very little time to generate.  A while ago I had it at 4M and that did take longer.  But this we should never realistically end up hitting.

It really should just have a huge buffer and generate the entropy on the fly instead of doing it all upfront.

crates/consensus-logic/src/duty/block_assembly.rs
```diff
@@ -44,20 +45,21 @@ const MAX_L1_ENTRIES_PER_BLOCK: usize = 100;
...
 // TODO pass in the CSM state we're using to assemble this
-pub(super) fn sign_and_store_block<D: Database, E: ExecEngineCtl>(
+pub(super) fn sign_and_store_block(
     slot: u64,
     prev_block_id: L2BlockId,
     l1_state: &LocalL1State,
     ik: &IdentityKey,
-    database: &D,
-    engine: &E,
+    database: &impl Database,
```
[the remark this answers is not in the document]
delbonis: I believe so.

crates/db/src/traits.rs
delbonis: In theory yes but because of how the ops wrapper works this isn't ideal, ultimately because you can't pass a ref between `Future`s.  There should be a ticket to tweak the ops interface to *allow* you to pass by-ref in blocking calls if you want to where it's possible, but take by-value in async calls.

delbonis:
Will make some minor changes in response to these comments and rebase.

crates/primitives/src/epoch.rs
```diff
@@ -0,0 +1,82 @@
...
+impl EpochCommitment {
+    pub fn new(epoch: u64, last_slot: u64, last_blkid: L2BlockId) -> Self {
+        Self {
+            epoch,
+            last_slot,
+            last_blkid,
+        }
+    }
+
+    /// Creates a "null" epoch with
```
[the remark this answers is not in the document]
delbonis: Oops!

delbonis:
Rebased and applied fixes.

crates/test-utils/src/lib.rs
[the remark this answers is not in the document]
bewakes: Its probably fine if it actually doesn't take much time to generate. Just pointing out the existence of `ArbitraryGenerator::new_with_size` and encouraging its usage when we need more entropy for a few test cases that are flaky.

crates/rocksdb-store/src/chain_state/db.rs
```diff
@@ -21,78 +21,59 @@ impl ChainstateDb {
...
-            None => return Err(DbError::OooInsert("Chainstate", idx)),
-        };
-        let post_state = state_op::apply_write_batch_to_chainstate(pre_state, batch);
+        // Make sure we always have a contiguous range of batches.
+        // FIXME this *could* be a race condition / TOCTOU issue, but we're only
+        // going to be writing from a single thread anyways so it should be fine
+        match self.get_last_idx()? {
+            Some(last_idx) => {
+                if idx != last_idx + 1 {
+                    return Err(DbError::OooInsert("Chainstate", idx));
```
bewakes, line 58: This assumes that there can't be updates to a given idx. I am not sure if that is valid, won't we want to update the entry in case of forks?

bewakes (commented):
Writing `WriteBatch` to db seems a bit weird in that it seems to not allow for updating an idx. The rest is good.

storopoli (dismissed):
ACK f10c7ca0597dcb92585da20c2f88d61e506c5092

crates/rocksdb-store/src/chain_state/db.rs
delbonis: Ah this is probably a mistake in translation.

delbonis: Oh wait no this was correct.  If we want to update an idx then we need to explicitly roll it back first.

delbonis:
Okay this should be good to go now.  I don't want to force merge this.

bewakes (approved):

prajwolrg (approved):

crates/consensus-logic/src/duty/worker.rs
sapinb, line 1: unused file likely added during conflict fix.
This was moved inside crates/sequencer

sapinb (approved):

delbonis:
Force merging because we have a few approvals already and we want to keep velocity up.

PR #637, opened 2025-01-29
Split transition information from `BatchInfo` to `BatchTransition`

prajwolrg:
## Description

Currently, the entire BatchInfo (including block heights and L1/L2 states) is proven, even though the block heights are already committed in the state. This leads to redundant data and unnatural edge cases—such as inclusive block ranges requiring small hacks to handle transitions.

This PR addresses those issues by splitting the structure into:

BatchTransition – contains only the essential proof data.
BatchInfo – handles auxiliary bookkeeping.
This separation removes duplication and simplifies handling of batch transitions.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/state/src/batch.rs
+++ b/crates/state/src/batch.rs
@@ -222,12 +222,14 @@ impl BatchTransition {
         }
     }
 
-    /// Creates a [`CheckpointBaseState`] by taking the initial state of the [`BatchTransition`]
+    /// Creates a [`CheckpointBaseStateCommitment`] by taking the initial state of the
+    /// [`BatchTransition`]
     pub fn get_initial_checkpoint_base_state(&self) -> CheckpointBaseStateCommitment {
         CheckpointBaseStateCommitment::new(self.l1_transition.0, self.l2_transition.0)
     }
 
-    /// Creates a [`CheckpointBaseState`] by taking the final state of the [`BatchTransition`]
+    /// Creates a [`CheckpointBaseStateCommitment`] by taking the final state of the
+    /// [`BatchTransition`]
     pub fn get_final_checkpoint_base_state(&self) -> CheckpointBaseStateCommitment {
         CheckpointBaseStateCommitment::new(self.l1_transition.1, self.l2_transition.1)
     }
--- a/crates/state/src/batch.rs
+++ b/crates/state/src/batch.rs
@@ -25,7 +25,7 @@ pub struct BatchCheckpoint {
 
     /// Reference state commitment against which batch transition and corresponding proof is
     /// verified
-    checkpoint_base_state: CheckpointBaseStateCommitment,
+    base_state_commitment: BaseStateCommitment,
 
     /// Proof for the batch obtained from prover manager
     proof: Proof,
@@ -35,13 +35,13 @@ impl BatchCheckpoint {
     pub fn new(
         batch_info: BatchInfo,
         transition: BatchTransition,
-        checkpoint_base_state: CheckpointBaseStateCommitment,
+        base_state_commitment: BaseStateCommitment,
         proof: Proof,
     ) -> Self {
         Self {
             batch_info,
             transition,
-            checkpoint_base_state,
+            base_state_commitment,
             proof,
         }
     }
@@ -54,8 +54,8 @@ impl BatchCheckpoint {
         &self.transition
     }
 
-    pub fn checkpoint_base_state(&self) -> &CheckpointBaseStateCommitment {
-        &self.checkpoint_base_state
+    pub fn base_state_commitment(&self) -> &BaseStateCommitment {
+        &self.base_state_commitment
     }
 
     pub fn proof(&self) -> &Proof {
@@ -69,7 +69,7 @@ impl BatchCheckpoint {
     pub fn get_proof_output(&self) -> CheckpointProofOutput {
         CheckpointProofOutput::new(
             self.batch_transition().clone(),
-            self.checkpoint_base_state().clone(),
+            self.base_state_commitment().clone(),
         )
     }
 
@@ -222,16 +222,16 @@ impl BatchTransition {
         }
     }
 
-    /// Creates a [`CheckpointBaseStateCommitment`] by taking the initial state of the
+    /// Creates a [`BaseStateCommitment`] by taking the initial state of the
     /// [`BatchTransition`]
-    pub fn get_initial_checkpoint_base_state(&self) -> CheckpointBaseStateCommitment {
-        CheckpointBaseStateCommitment::new(self.l1_transition.0, self.l2_transition.0)
+    pub fn get_initial_base_state_commitment(&self) -> BaseStateCommitment {
+        BaseStateCommitment::new(self.l1_transition.0, self.l2_transition.0)
     }
 
-    /// Creates a [`CheckpointBaseStateCommitment`] by taking the final state of the
+    /// Creates a [`BaseStateCommitment`] by taking the final state of the
     /// [`BatchTransition`]
-    pub fn get_final_checkpoint_base_state(&self) -> CheckpointBaseStateCommitment {
-        CheckpointBaseStateCommitment::new(self.l1_transition.1, self.l2_transition.1)
+    pub fn get_final_base_state_commitment(&self) -> BaseStateCommitment {
+        BaseStateCommitment::new(self.l1_transition.1, self.l2_transition.1)
     }
 
     pub fn rollup_params_commitment(&self) -> Buf32 {
@@ -248,12 +248,12 @@ impl BatchTransition {
 #[derive(
     Clone, Debug, PartialEq, Eq, Arbitrary, BorshDeserialize, BorshSerialize, Deserialize, Serialize,
 )]
-pub struct CheckpointBaseStateCommitment {
+pub struct BaseStateCommitment {
     pub initial_l1_state: Buf32,
     pub initial_l2_state: Buf32,
 }
 
-impl CheckpointBaseStateCommitment {
+impl BaseStateCommitment {
     pub fn new(initial_l1_state: Buf32, initial_l2_state: Buf32) -> Self {
         Self {
             initial_l1_state,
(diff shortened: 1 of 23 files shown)


bewakes (dismissed):

crates/state/src/batch.rs
delbonis, line 1: The organization of items in this module is a bit weird.  The `BatchInfo` and `BatchTransition` items are like interleaved.

crates/state/src/batch.rs
```diff
@@ -115,110 +135,86 @@ pub struct BatchInfo {
...
+            l1_range,
+            l2_range,
+            l2_blockid,
+        }
     }
 
-    pub fn rollup_params_commitment(&self) -> Buf32 {
-        self.rollup_params_commitment
+    pub fn idx(&self) -> u64 {
+        self.idx
```
delbonis, line 211: Should more explicitly refer to `epoch`.

crates/state/src/batch.rs
```diff
@@ -115,110 +135,86 @@ pub struct BatchInfo {
...
+        self.rollup_params_commitment
     }
+}
 
-    pub fn final_acc_pow(&self) -> u128 {
-        self.l1_pow_transition.1
+impl BatchInfo {
+    pub fn new(
+        checkpoint_idx: u64,
+        l1_range: (u64, u64),
```
delbonis, line 198: Should this include the terminal L1 blkid?

delbonis (changes requested):
Consider using the `L1BlockCommitment`/`L2BlockCommitment` types that are in the new chainstate manager PR.

crates/state/src/batch.rs
prajwolrg: Not necessarily. Since the information on `BatchInfo` is not verifiable and is only meant to help for debugging use cases, not sure how much helpful that would be here. We can include it if it'll be helpful. CC: @bewakes

crates/state/src/batch.rs
prajwolrg: Yes, they convey the same information. `BatchInfo` is optional to be posted onto Bitcoin but `BatchTransition` is a must. Not sure what to do next here.

prajwolrg:
> Consider using the `L1BlockCommitment`/`L2BlockCommitment` types that are in the new chainstate manager PR.

Yes, makes sense. Will continue on this then once that is merged.

crates/state/src/batch.rs
bewakes: L2 block id is there to avoid possible ambiguities(not in our case though) and avoiding 1 extra db call to fetch the block at given l2 height. Checkpoint does not necessarily need l1/l2 block id, if they can be fetched from height.

However, something to keep in mind: l1 range is currently calculated from client state, while it should in fact be calculated from chainstate. Probably makes sense to fix that in this PR?

crates/state/src/batch.rs
delbonis: I'm just saying the literal ordering of the definitions in the module seems weird.  Unless that's the github diff view getting messed up?

crates/state/src/batch.rs
delbonis: I suppose it's (going to be) included transiently through the blkid.

> l1 range is currently calculated from client state, while it should in fact be calculated from chainstate

I'll be addressing this in the big epoch refactor.  It's a bit more subtle than it seems.

crates/state/src/batch.rs
prajwolrg: the struct is defined as:

```rust
/// Consolidates all information required to describe and verify a batch checkpoint.
/// This includes metadata about the batch, the state transitions, bootstrap info,
/// and the proof itself. The proof verifies that the transition in [`BatchTransition`]
/// is valid for the batch described by [`BatchInfo`].
#[derive(
    Clone, Debug, PartialEq, Eq, Arbitrary, BorshDeserialize, BorshSerialize, Deserialize, Serialize,
)]
pub struct BatchCheckpoint {
    /// Information regarding the current batch checkpoint
    batch_info: BatchInfo,

    /// Transition data for L1 and L2 states, which is verified by the proof.
    transition: BatchTransition,

    /// Bootstrap info based on which the checkpoint transition and proof is verified
    bootstrap: BootstrapState,

    /// Proof for the batch obtained from prover manager
    proof: Proof,
}
```
I don't fully understand what seems weird.

delbonis:
@prajwolrg That PR got merge btw

crates/state/src/batch.rs
delbonis: I'm seeing:

```rs
struct BatchInfo {}
struct BatchTransition {}
impl BatchTransition {}
impl BatchInfo {}
```

prajwolrg: Yup, what do you suggest here exactly?
Having a single struct for `BatchInfo` and `BatchTransition`? I don't get that part.

prajwolrg:
> > Consider using the `L1BlockCommitment`/`L2BlockCommitment` types that are in the new chainstate manager PR.
> 
> Yes, makes sense. Will continue on this then once that is merged.

Right now `L1BlockCommitment` includes the height and the block hash, but in the `BatchTransition` we are verifying commitment of `HeaderVerificationState` which includes some additional info on top of height and block hash. 

How do you suggest I proceed here? Do I include HeaderVerificationState commitment to  L1BlockCommitment? Or leave it out as it is?

prajwolrg:
> > > Consider using the `L1BlockCommitment`/`L2BlockCommitment` types that are in the new chainstate manager PR.
> > 
> > 
> > Yes, makes sense. Will continue on this then once that is merged.
> 
> Right now `L1BlockCommitment` includes the height and the block hash, but in the `BatchTransition` we are verifying commitment of `HeaderVerificationState` which includes some additional info on top of height and block hash.
> 
> How do you suggest I proceed here? Do I include HeaderVerificationState commitment to L1BlockCommitment? Or leave it out as it is?

For `BatchInfo`, `L1BlockCommitment` and `L2BlockCommitment` has been used.

crates/state/src/batch.rs
delbonis: Like,

```rs
struct BatchInfo {}
impl BatchInfo {}
struct BatchTransition {}
impl BatchTransition {}
```

crates/state/src/batch.rs
```diff
@@ -102,178 +128,156 @@ impl From<SignedBatchCheckpoint> for BatchCheckpoint {
...
     /// L1 height range(inclusive) the checkpoint covers
-    pub l1_range: (u64, u64),
+    pub l1_range: (L1BlockCommitment, L1BlockCommitment),
 
     /// L2 height range(inclusive) the checkpoint covers
-    pub l2_range: (u64, u64),
+    pub l2_range: (L2BlockCommitment, L2BlockCommitment),
 
+    /// The last L2 block upto which this checkpoint covers since the previous checkpoint
+    pub l2_blockid: L2BlockId,
```
delbonis, line 147: This is no longer necessary to have here as a separate field since it's in `l2_range.1`.

crates/state/src/batch.rs
```diff
@@ -102,178 +128,156 @@ impl From<SignedBatchCheckpoint> for BatchCheckpoint {
...
     pub l1_transition: (Buf32, Buf32),
 
     /// The inclusive hash range of `Chainstate` for L2 blocks.
-    /// This represents the transition of L2 state from the starting state to the
-    /// ending state. The state root is computed via
-    /// [`super::chain_state::Chainstate::compute_state_root`].
+    ///
+    /// Represents a transition from the starting L2 state to the ending L2 state.
+    /// The state root is computed via [`super::chain_state::Chainstate::compute_state_root`].
     pub l2_transition: (Buf32, Buf32),
```
delbonis, line 166: Oh I was misunderstanding this the whole time.

For now I don't think we need to explicitly state these L2 state roots here since they're transitively committed to by the blkids in the block commitments in the other struct.  Or do these represent something else.

In the *future* we might commit to the "outer post-state" here (right now that *is* what's put in the epoch terminal block), but we can revisit that later.

crates/state/src/batch.rs
```diff
@@ -102,178 +128,156 @@ impl From<SignedBatchCheckpoint> for BatchCheckpoint {
...
+/// NOTE/TODO: This state serves as the starting point for verifying a checkpoint proof. If we move
+/// towards a strict mode where we prove each checkpoint recursively, this should be replaced with
+/// `GenesisState`.
 #[derive(
     Clone, Debug, PartialEq, Eq, Arbitrary, BorshDeserialize, BorshSerialize, Deserialize, Serialize,
 )]
-pub struct BootstrapState {
-    pub start_l1_height: u64,
-    // TODO is this a blkid?
+pub struct CheckpointBaseState {
```
delbonis, line 253: Maybe should have `Commitment` in the name?  Since it's not actually the expanded state it's just a commitment to it.

delbonis (changes requested):
Okay so I was somewhat confused by the type/field names being vague here.  The documentation should have clarified it but the names themselves led me in the wrong direction initially.

crates/state/src/batch.rs
prajwolrg: Right now there's a slight inconsistency between how states are recorded in `BatchInfo` versus `BatchTransition`. For example, in `BatchInfo`, if you have an l1_range of (1, 5), the next range will be (6, 10). In this context, the block IDs (and the corresponding state root it commits) are recorded after the block is applied. However, in `BatchTransition`, the state roots are captured before the block is applied, leading to a slight mismatch.

MdTeach (approved):

bewakes (approved):
I would go for some concise name for `CheckpointBaseStateCommitment`, something like `BaseStateCommitment` which should be clear enough given that the struct has docstring. But the rest is good.

crates/state/src/batch.rs
delbonis: Then would it make sense to change the framing to be based on "prev checkpoint tip" and "new checkpoint tip".  And make it half open in the reverse direction?

> However, in BatchTransition, the state roots are captured before the block is applied, leading to a slight mismatch.

This seems like a weird special case.  What is the reason for this?

MdTeach (approved):

crates/state/src/batch.rs
prajwolrg: For L2 it could have been based entirely on `L2BlockId` since it already commits the `state_root`. But since `L1BlockId` doesn't commit to the L1State, it's done this way for the consistency between L1 and L2.

delbonis (approved):
I think this can be merged right?

delbonis:
Merging to unblock stuff.

PR #417, opened 2024-10-16
feat: sync missing el blocks from client db

sapinb:
## Description
If reth db is missing blocks expected by the client chainstate, sync the missing blocks.

### Type of Change

-   [x] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Documentation update
-   [ ] Refactor
-   [ ] New or updated tests

## Checklist

-   [ ] I have performed a self-review of my code.
-   [ ] I have commented my code where necessary.
-   [ ] I have updated the documentation if needed.
-   [ ] My changes do not introduce new warnings.
-   [ ] I have added tests that prove my changes are effective or that my feature works.
-   [ ] New and existing tests pass with my changes.

PR #653, opened 2025-02-05
Fix sequencer crash tests

sapinb:
## Description
* fix for crash_advance_consensus_state
  * apply any extra blocks from db before startup in fcm
* fix for sync_event_new_tip crash
  * sync missing EL blocks from CL chainstate (port from #417)
  * in fcm init, get chain state height from chain state db instead of sync state
* move crash_* tests to own group

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist


bin/strata-client/src/el_sync.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    // last idx of chainstate whose corresponding block is present in el
+    let sync_from_idx = find_last_match((earliest_idx, latest_idx), |idx| {
+        let Some(chain_state) = chainstate_manager.get_toplevel_chainstate_blocking(idx)? else {
+            anyhow::bail!(format!("Missing chain state idx: {}", idx));
+        };
+
+        let block_id = chain_state.chain_tip_blkid();
+
+        Ok(engine.check_block_exists(*block_id)?)
```
bewakes, line 27: Now as I think of it, is it possible to first fetch the latest block from EL and then binary search on the db to find the corresponding block and sync after that? This would reduce the engine call to a single call. Something like:
```rust
let eng_block_id = engine.get_latest_block();
let sync_from_idx = find_last_match((earliest, latest), |idx| {
   let chain_state = manager.get_toplevel_state(...);
   let block_id = chain_state.chain_tip_blkid();
   Ok(block_id==eng_block_id)
});
```

crates/consensus-logic/src/fork_choice_manager.rs
```diff
@@ -232,6 +234,36 @@ pub fn tracker_task<D: Database, E: ExecEngineCtl>(
...
+                    engine.as_ref(),
+                    &csm_ctl,
+                    &status_channel,
+                )?;
+            }
+            slot += 1;
+        }
+        info!("completed check for unprocessed l2blocks");
+    }
+
```
bewakes, line 266: nit: this could be separated into a function.

functional-tests/tests/el_sync_from_chainstate.py
```diff
@@ -0,0 +1,125 @@
+import flexitest
+from web3 import Web3
+
+from envs import testenv
+from utils import (
+    wait_until,
+)
```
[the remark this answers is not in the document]
sapinb: well .. the linter has not complained yet 🤷

bin/strata-client/src/el_sync.rs
```diff
@@ -0,0 +1,103 @@
...
+pub fn sync_chainstate_to_el(
+    storage: &NodeStorage,
+    engine: &impl ExecEngineCtl,
+) -> anyhow::Result<()> {
+    let chainstate_manager = storage.chainstate();
+    let l2_block_manager = storage.l2();
+    let earliest_idx = chainstate_manager.get_earliest_write_idx_blocking()?;
+    let latest_idx = chainstate_manager.get_last_write_idx_blocking()?;
+
+    debug!(?earliest_idx, ?latest_idx, "search for last known idx");
```
delbonis, line 17: These can be printed with `%` instead of `?`.

bin/strata-client/src/el_sync.rs
```diff
@@ -0,0 +1,103 @@
...
+
+        Ok(engine.check_block_exists(*block_id)?)
+    })?
+    .map(|idx| idx + 1) // sync from next index
+    .unwrap_or(0); // sync from genesis
+
+    debug!(?sync_from_idx, "last known index in EL");
+
+    for idx in sync_from_idx..=latest_idx {
+        debug!(?idx, "Syncing chainstate");
```
delbonis, line 35: Same here.

bin/strata-client/src/el_sync.rs
```diff
@@ -0,0 +1,103 @@
...
+        let Some(chain_state) = chainstate_manager.get_toplevel_chainstate_blocking(idx)? else {
+            anyhow::bail!(format!("Missing chain state idx: {}", idx));
```
delbonis, line 37: Consider making this a real `Error` type instead of using `bail!`.

bin/strata-client/src/el_sync.rs
```diff
@@ -0,0 +1,103 @@
...
+        };
+
+        let block_id = chain_state.chain_tip_blkid();
+
+        let Some(l2block) = l2_block_manager.get_block_data_blocking(block_id)? else {
+            anyhow::bail!(format!("Missing L2 block idx: {}", block_id));
```
delbonis, line 43: Same here.

bin/strata-client/src/el_sync.rs
```diff
@@ -0,0 +1,103 @@
...
+        engine.submit_payload(payload)?;
+        engine.update_head_block(*block_id)?;
+    }
+
+    Ok(())
+}
+
+fn find_last_match<F>(range: (u64, u64), predicate: F) -> anyhow::Result<Option<u64>>
+where
+    F: Fn(u64) -> anyhow::Result<bool>,
```
delbonis, line 57: This fn arg could also be written more cleanly with `impl Fn(u64) -> anyhow::Result<bool>`.

bin/strata-client/src/main.rs
```diff
@@ -284,8 +286,8 @@ fn do_startup_checks(
         }
         Ok(false) => {
             // Current chain tip tip block is not known by the EL.
-            // TODO: Try to sync EL using existing block payloads from DB.
-            anyhow::bail!("missing expected evm block, block_id = {}", chain_tip);
+            warn!("missing expected evm block, block_id = {}", chain_tip);
```
delbonis, line 289: `warn!(%chain_tip, "missing expected EVM block");`

delbonis, line 289: `warn!(%chain_tip, "missing expected EVM block");`

bin/strata-sequencer-client/src/duty_fetcher.rs
```diff
@@ -25,7 +25,7 @@ where
             }
         };
 
-        debug!("got {} duties", duties.len());
+        info!("got {} duties", duties.len());
```
delbonis, line 28: `info!(count = %duties.len(), "got new duties");`

delbonis (changes requested):
Something got really weird with the github review and some comments got submitted into multiple separate reviews without me intending to, quality software.

But the actual comment I wanted to make was that there's a few places here where the term `idx` is being used in a way that could be vague.  It would be better to use a more explicit term like `slot` if possible.

bin/strata-client/src/el_sync.rs
sapinb: @bewakes This is a good, but we need to index ELBlockid -> L2Blockid in our db to be able to map the result from reth to the blockid used by chain state. I'll create a ticket to do this optimization later.

bin/strata-client/src/el_sync.rs
sapinb: updated

bin/strata-client/src/el_sync.rs
sapinb: updated

bin/strata-client/src/main.rs
sapinb: updated

bin/strata-sequencer-client/src/duty_fetcher.rs
sapinb: updated

delbonis (approved):
This looks good but it seems that `bridge_deposit_seq_unreliable` failed in CI.  Is that a flaky test?  It says that a wait timed out so I wonder if we could be adjusting the tolerances there or if there's something slowing down startup?

storopoli (approved):
ACK ce2df6175133f42cfc13b04b03b9ff7bb5c9c678

The log availability in 7101d91c31a3e10307309d95f0809ceb2ec79f7f IS AWESOME! THANKS!

storopoli:
Amazing `cargo-audit` has a new vulnerability: https://rustsec.org/advisories/RUSTSEC-2025-0006

storopoli (approved):
ACK 6a25e9caeb88f3eab07168bf333fc2ac19ca5de4

sapinb:
converting this to draft while fixing functional test issues

sapinb:
@storopoli @delbonis @Rajil1213  inspecting the logs didnt help much. Found that the second DRT was not being processed into DT in the test, but only when run in the ci. Might be some obscure timing issue in the bridge duty system. But all this is getting replaced anyway so not going to dig more into this. 
Using an independent env for the test fixes it anyway.

delbonis (approved):

delbonis:
Yeah I feel like crash tests should be in their own envs anyways.

delbonis:
Force merging to keep things rolling.

PR #657, opened 2025-02-06
Remove blocks as CSM input

delbonis:
## Description

The goal of this PR (which is related to the big STF refactors I'm slowly merging in) is to remove the CSM tracking of the chain tip and center the FCM as the source of truth on that.  The CSM will only have L1 messages as inputs and will be mainly concerned about checkpoints and L1 finalization.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

These are all sorta related.

* [STR-696](https://alpenlabs.atlassian.net/browse/STR-696)
* [STR-827](https://alpenlabs.atlassian.net/browse/STR-827)
* [STR-929](https://alpenlabs.atlassian.net/browse/STR-929)

[STR-696]: https://alpenlabs.atlassian.net/browse/STR-696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


delbonis:
Since there's some loose ends in the bookkeeping without the CSM being aware of the chain tip, I'm extending this refactor to change some things about how checkpoints are tracked in the CSM and how it works more generally.

* Sync events are primarily just updates to the L1 tip and genesis.  Checkpoints are no longer separate messages.  The CSM does all validation, deriving checkpoints from L1 manifests.
* The "actual" `ClientState`, which tracks and is used to validate checkpoints, is this new `InternalState` structure corresponding to L1 blocks by height.  These are a pure function of the previous `InternalState` and an input L1 block.  When we get a new block we create a new instance, when it reorgs we discard invalid ones.
  * The instance corresponding to the chain tip is the source of truth on what the last "confirmed" checkpoint is, as of the chain tip.  The instance that corresponds to whatever the "buried" L1 block is determines what the finalized checkpoint is.
  * We keep in the `ClientState` a list of the last several `InternalState`s so we can access them easily.
* Some refactoring relating to how we validate checkpoints.
* L1 block assembly ignores the client state more, fetches the L1 blocks we include directly from the L1 database manager.

In the short term, I need some eyes on this to make sure I'm not missing anything important.  There's a LOT of `TODO`s, mostly to remove code that's no longer relevant, or reconnect some pieces that had to be temporarily disconnected/stubbed.  I'm running into some issues where functions have multiple responsibilities or struct/field names are ambiguous, so I have to reverse engineer some things from context and it's likely I may be subtly misinterpreting some things.

The plan (in a future PR) is to eventually the `InternalState` type out and onto disk, and change the CSM logic so most of the bookkeeping that happens persistently by way of processing sequential `SyncEvent`s just happens on the fly in memory.  This approach is much more durable.

crates/state/src/client_state.rs
```diff
@@ -44,9 +44,15 @@ pub struct ClientState {
...
 
+    /// The depth at which we accept blocks to be finalized.
+    pub(super) finalization_depth: u64,
+
     /// Hash of verification state at `genesis_l1_height`. The hash is computed via
     /// [`super::l1::HeaderVerificationState::compute_hash`]
     pub(super) genesis_l1_verification_state_hash: Option<Buf32>,
+
+    /// Internal states according to each block height.
+    pub(crate) int_states: BTreeMap<u64, InternalState>,
```
bewakes, line 55: Looks like  `StateQueue` can be used here to deal with `PartialEq` issue that you mentioned earlier.

delbonis: Yeah it'd probably be fine to use that actually.

delbonis: Did.

delbonis:
Finished monster rebase.

delbonis:
PR is not done yet but I've got most of the basic functionality working again and I don't anticipate any major changes, so I'm marking this as "ready for review".

Several tests are still failing but it's *probably* just silly things that *should* be straightforward to address after going through a full run of the test suite and seeing what's upset.  There's also a few loose ends to tie up like readding some checks that I removed/simplified for testing and a bunch of commented code blocks in places that can probably be removed now that I don't need to reference then anymore.  And of course fixing lints.

crates/rocksdb-store/src/checkpoint/db.rs
```diff
@@ -23,21 +24,80 @@ impl RBCheckpointDB {
...
+        // This is kinda nontrivial so we don't want concurrent writes to
+        // clobber each other, so we do it in a transaction.
+        //
+        // That would probably never happen, but better safe than sorry!
+        self.db
+            .with_optimistic_txn(
+                rockbound::TransactionRetry::Count(self.ops.retry_count),
+                |txn| {
+                    let mut summaries: Vec<EpochSummary> = txn
+                        .get::<EpochSummarySchema>(&epoch_idx)?
```
sapinb, line 41: Need to use `.get_for_update()` instead of `.get()`, otherwise its not really in the transaction.
Now it feels kinda stupid to keep both .get() and .get_for_update() methods here on the txn, just adds to confusion.

crates/rocksdb-store/src/checkpoint/schemas.rs
```diff
@@ -1,8 +1,20 @@
...
 
 define_table_with_seek_key_codec!(
     /// A table to store idx -> `CheckpointEntry` mapping
     (CheckpointSchema) u64 => CheckpointEntry
 );
+
+define_table_with_seek_key_codec!(
+    /// Table mapping epoch indexes to the list of summaries in that index.
+    // This is weird because I wanted to avoid too much duplication and
+    // `EpochCommitment` doesn't implement `KeyCodec`.
```
sapinb, line 14: It only needs to implement `BorshSerialize` and `BorshDeserialize` which `EpochCommitment` already implements. This should work:

```rust
define_table_with_default_codec!(
    /// Table mapping epoch indexes to the list of summaries in that index.
    (EpochSummarySchema) EpochCommitment => Vec<EpochSummary>
);
```

delbonis:
I've discovered these things:

* `sync_from_rpc`: the Strata node is syncing but the follower EL client isn't actually getting the EL block, I haven't investigated further
* `sync_bitcoin_reorg`: doing the same weird behavior that `sync_l1_finalization` was doing, needs to be transformed in the same way, should be easy
* `el_genesis` works, `forward_tx` works, but `el_block_generation` doesn't, which is weird since it ALSO seems like in `sync_from_rpc` we *were* producing EL blocks, not sure what's happening there

crates/rocksdb-store/src/checkpoint/schemas.rs
delbonis: Wait but we do need to fetch all of the possible summaries stored for an epoch *index*.  A commitment can only map to a single summary.  I did have the idea of iterating from a fake epoch that's like `(i, 0, 0)` and seeking until we encountered `(i+1, _, _)`, but I'm not sure if it's worth revisiting that part right now and if it should be split out into another ticket to simplify the schema and reduce the duplication.

crates/consensus-logic/src/l1_handler.rs
sapinb, line 1: @delbonis this file is unused, likely from a rebase. Is there anything relevant here that is not moved to `crates/btcio/src/reader/handler.rs` ?

crates/btcio/src/reader/handler.rs
```diff
@@ -67,29 +73,33 @@ async fn handle_blockdata<R: ReaderRpc>(
...
-    sync_evs.push(SyncEvent::L1Block(blockdata.block_num(), blkid.into()));
-
-    // Check for checkpoint and create event accordingly
+    // TODO remove all this outdated stuff
+    /*
+        // Check for checkpoint and create event accordingly
     debug!(%height, "Checking for checkpoints in l1 block");
-    let checkpoints = check_for_commitments(&blockdata, *seq_pubkey);
-    debug!(?checkpoints, "Received checkpoints");
+        let checkpoints = check_for_commitments(&blockdata, *seq_pubkey);
```
sapinb, line 94: sequencer signature check is not being performed at any point in the new flow

crates/btcio/src/reader/query.rs
```diff
@@ -152,12 +153,44 @@ fn handle_poll_error(err: &anyhow::Error, status_updates: &mut Vec<L1StatusUpdat
     }
 }
 
+/// Reverts the reader state to the height where the last checkpoint is finalized.
+async fn handle_new_filter_rule<R: ReaderRpc>(
```
sapinb, line 157: function is unused. likely from rebase

crates/btcio/src/reader/query.rs
```diff
@@ -152,12 +153,44 @@ fn handle_poll_error(err: &anyhow::Error, status_updates: &mut Vec<L1StatusUpdat
...
+}
+
 /// Checks for epoch changes and any L1 reverts necessary. Internally updates reader's state.
 fn check_epoch_change<R: ReaderRpc>(
     ctx: &ReaderContext<R>,
     state: &mut ReaderState,
 ) -> anyhow::Result<Option<L1Event>> {
-    let new_epoch = ctx.status_channel.epoch().unwrap_or(0);
+    // FIXME something here might have gotten borked with the bookkeeping here
+    // in a rebase, let me (trey) know if there's something wrong here
```
storopoli, line 191: Leaving this comment as a placeholder to tackle this FIXME

crates/btcio/src/reader/query.rs
```diff
@@ -358,30 +395,48 @@ async fn process_block<R: ReaderRpc>(
...
+            .send(L1Event::GenesisVerificationState(
+                height,
+                l1_verification_state,
+            ))
+            .await
+        {
+            error!("failed to submit L1 block event, did the persistence task crash?");
+            return Err(e.into());
+        }
+    }*/
```
storopoli, line 427: remove?

crates/btcio/src/reader/query.rs
```diff
@@ -358,30 +395,48 @@ async fn process_block<R: ReaderRpc>(
...
-    if height == genesis_threshold {
+    /*if height == genesis_threshold {
         info!(%height, %genesis_ht, "time for genesis");
         let l1_verification_state =
             get_verification_state(ctx.client.as_ref(), genesis_ht + 1, &get_btc_params()).await?;
-        let ev = L1Event::GenesisVerificationState(height, l1_verification_state);
+        let ev = L1Event::GenesisVerificationState(block_commitment, l1_verification_state);
         l1_events.push(ev);
-    }
+    }*/
```
storopoli, line 406: remove?

crates/crypto/src/lib.rs
```diff
@@ -1,68 +1,2 @@
...
-        let mod_msg = Buf32::from(mod_msg);
-        assert!(!verify_schnorr_sig(&sig, &mod_msg, &pk));
-
-        let sig = sign_schnorr_sig(&mod_msg, &sk);
-        let res = verify_schnorr_sig(&sig, &mod_msg, &pk);
-        assert!(res);
-    }
-}
+// FIXME this stub had to be moved to make a bodgy refactor work
+pub use strata_primitives::crypto::*;
```
storopoli, line 2: This crate does not have any purpose now. Should be deleted and anything that uses it should now use `primitives:crypto`.

crates/rpc/types/src/types.rs
```diff
@@ -126,25 +127,46 @@ impl Default for RpcL1Status {
...
 pub struct RpcClientStatus {
     /// Blockchain tip.
+    // TODO remove this since the CSM doesn't track this anymore, we're pulling it in indirectly
     #[serde(with = "hex::serde")]
+    #[deprecated(note = "no longer tracked by client state")]
     pub chain_tip: [u8; 32],
 
     /// L1 chain tip slot.
+    // TODO remove this since the CSM doesn't track this anymore, we're pulling it in indirectly
+    #[deprecated(note = "no longer tracked by client state")]
```
storopoli, line 141: We should remove all these deprecated RPCs for testnet 1.

crates/sequencer/Cargo.toml
```diff
@@ -27,7 +27,6 @@ tracing.workspace = true
 strata-test-utils.workspace = true
 
 [lints]
-rust.missing_docs = "warn"
```
storopoli, line 30: Please don't remove this lint? Use `clippy::allow` instead.

crates/sequencer/src/checkpoint/expiry.rs
```diff
@@ -11,6 +11,7 @@ use tracing::{debug, error, warn};
 
 use crate::checkpoint::CheckpointHandle;
 
+// FIXME WHAT DO THESE FIELDS REPRESENT????????
```
storopoli, line 14: See what missing docs do? 😂

storopoli: Now you're left to resort to gitblame Spanish Inquisition.

storopoli (changes requested):
Few nits.
Added a bunch of placeholders for important TODOs that can be deferred later with the premise that they are being properly tracked in issues.

crates/crypto/src/lib.rs
delbonis: I would *like* for it to still have a purpose.  I had to move a bunch of state types into the primitives crate but I didn't *want* to.  It was easier than redoing the whole type hierarchy.  I am going to create a ticket for redoing the crates.

crates/rpc/types/src/types.rs
delbonis: Yes definitely.  I'll create tickets for removing these.  I didn't want to expand the scope further by going and actually removing all the uses so I just added the deprecated attr to make noise.

crates/sequencer/Cargo.toml
delbonis: it's a little excessive

crates/sequencer/src/checkpoint/expiry.rs
delbonis: fineeeee :P

crates/consensus-logic/src/l1_handler.rs
delbonis: There shouldn't be.  There was definitely stuff that got clobbered by the rebase.

crates/btcio/src/reader/handler.rs
delbonis: I want the query and persistence tasks consolidated more as well.  Like this `L1Event` thing is a hangover that we should ideally be able to get rid of.

crates/consensus-logic/src/l1_handler.rs
delbonis: I want the query and persistence tasks themselves consolidated at some point as well. Like this L1Event thing is a hangover that we should ideally be able to get rid of.  The recent PR mainly just moved the handler modules into the btcio crate, which was only half of it.

bin/strata-client/src/main.rs
```diff
@@ -91,6 +91,7 @@ fn main_inner(args: Args) -> anyhow::Result<()> {
         .build()
         .expect("init: build rt");
     let task_manager = TaskManager::new(runtime.handle().clone());
+    //strata_tasks::set_panic_hook(); // only if necessary
```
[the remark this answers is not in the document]
delbonis: It's really just during troubleshooting, will update.

delbonis:
More updates:

* `client_restart`: Ended up getting stuck in `wait_for_proof_with_timeout` while waiting for dependencies.  Had to kill the rollup client and prover client in htop to get the test to abort.  The normal timeout is 1hr which is a little high to be waiting for locally.  This is far from what I'm comfortable troubleshooting, can @MdTeach maybe take a look?  I think it actually makes sense to remove anything specifically testing prover logic from this test since it's really just supposed to be testing that we can restart and continue to make progress.
* A few more tests are just failing because we called the client too soon before RPC came up.  I don't know exactly why startup is slower now but surely we can fix this by adding a wait in the startup procedure.
* `btcio_read`: There was a bug that caused us to try to include L1 blocks that we didn't have in the epoch and checkpoint.  This seems to be fine *now* but I uncovered an off by 1 bug with the L1 block accounting in the chainstate that I think we just never noticed before.  I don't know for sure what exactly was messing up the bookkeeping causing the error.  That was really weird.
* `el_*`: I have no idea why these EL tests are failing, the logs show nothing is amiss, it's just not doing anything.  Please I need someone to take a look at these.  I don't even know where to start. :pray: 
* `bridge_*`: I think there was something screwed up with the way we were putting manifests into blocks that is making these broken.  Going to look at it more tomorrow.

crates/consensus-logic/src/fork_choice_manager.rs
```diff
@@ -352,178 +385,197 @@ fn wait_for_fcm_event(
...
+    // Apply the reorg.
+    match apply_tip_update(tip_update, fcm_state) {
+        Ok(()) => {
+            info!(%tip_blkid, "new chain tip");
+
+            // Also this is the point at which we update the engine head and
+            // safe blocks.  These are technically different, but in practice
+            // they mean the same thing and we have to update them both anyways.
+            engine.update_head_block(tip_blkid)?;
+            engine.update_safe_block(tip_blkid)?;
```
sapinb, line 526: @delbonis it should be enough to do just update_safe_block here. engine api requires head block to always be set in the requests, so update safe block call will also send the same value in head block anyway.

crates/sequencer/src/block_template/block_assembly.rs
```diff
@@ -106,17 +106,17 @@ fn prepare_l1_segment(
     let cur_real_l1_height = l1man
         .get_chain_tip()?
         .expect("blockasm: should have L1 blocks by now");
-    let target_height = cur_real_l1_height - 1; // -1 to give some buffer
+    let target_height = cur_real_l1_height - 1; // -1 to give some buffer for very short reorgs
```
sapinb, line 109: We can just use the `l1_reorg_safe_depth` param as the buffer here, so we dont care about reorgs in fcm and chain state.
```rust
target_height = cur_real_l1_height - params.l1_reorg_safe_depth;
```

delbonis: Yeah I was thinking about switching to that.  I'm just wasn't sure if it'd break timings.

prajwolrg:
Everything seems to pass now. FINALLYYY!!

prajwolrg (approved):
utACK

sapinb (approved):
LGTM

crates/rpc/types/src/types.rs
delbonis: Created as STR-1087.

bin/strata-client/src/rpc_server.rs
```diff
@@ -96,22 +101,24 @@ impl StrataRpcImpl {
...
             return Ok((cs, None));
         }
 
-        let chs = self.status_channel.chain_state().map(Arc::new);
+        let chs = self.status_channel.get_cur_tip_chainstate().clone();
 
         Ok((cs, chs))
     }
 
+    // TODO remove this RPC, we aren't supposed to be exposing this
```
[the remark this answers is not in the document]
delbonis: Created as STR-1087.

storopoli (approved):
ACK 140a6212ec06725f38caf8b4a2030355179fadf5

PR #661, opened 2025-02-11
STR-497 Investigate using Zeroize in bridge/sequencer

krsnapaudel:
## Description
Zeroize `Keypair` and Sequencer `IdentityData`.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

[STR-1074](https://alpenlabs.atlassian.net/browse/STR-1074)

[STR-1074]: https://alpenlabs.atlassian.net/browse/STR-1074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


storopoli (changes requested):
> I am not confident about `exec_handler.erase_keypair()` getting called or `IdentityData` getting dropped.

Then create a test similar to https://github.com/alpenlabs/strata/blob/e3b21079a533a2806883c94580dade4b31293cbe/crates/primitives/src/keys.rs?plain=1#L93-L128

krsnapaudel:
@storopoli: I tried to write a test:

`    fn test_zeroize_on_drop() {
        let master = Xpriv::new_master(Network::Regtest, &[2u8; 32]).unwrap();
        let keys = Arc::new(SequencerKeys::new(&master).unwrap());
        let keys_clone = Arc::clone(&keys);

        // Store original values
        let master_chaincode = *keys_clone.master_xpriv().chain_code.as_bytes();
        let derived_chaincode = *keys_clone.derived_xpriv().chain_code.as_bytes();

        // Verify data exists
        assert_ne!(master_chaincode, [0u8; 32]);
        assert_ne!(derived_chaincode, [0u8; 32]);

        fn dummy_fn(keys: Arc<SequencerKeys>) {
            println!("got keys {}", keys.master);
        }
        // Create and drop our test wrapper
        {
            dummy_fn(keys_clone);
        }

        // NOTE: SecretKey::non_secure_erase writes `1`s to the memory.
        assert_eq!(keys.master_xpriv().private_key.secret_bytes(), [1u8; 32]);
        assert_eq!(keys.derived_xpriv().private_key.secret_bytes(), [1u8; 32]);
    }`

This is not going to work because Arc has a reference to shared_keys. If we try to drop that reference, then we cannot use for assert. This I think happens with or without Arc. Correct me if I got this wrong.

storopoli:
I guess you're having the same dopamine hit that I had when I coded the `ZeroizableXpriv` tests!
Seems that with [d4da127](https://github.com/alpenlabs/strata/pull/661/commits/d4da12787d8d5014b71d7fe99ee456bc339d6739) the thing works no?

bin/bridge-client/src/modes/operator/bootstrap.rs
```diff
@@ -197,5 +200,11 @@ pub(crate) async fn bootstrap(args: Cli) -> anyhow::Result<()> {
     // They are supposed to run indefinitely in most cases
     tokio::try_join!(rpc_task, duty_task)?;
 
+    // Attempt to erase keypair from ExecHandler and ExecHandler.sig_manager
+    if let Ok(mut handler) = Arc::try_unwrap(shared_exec_handler) {
+        handler.erase_keypair(); // Ensure keypair is erased before dropping
+        handler.sig_manager.erase_keypair();
+    }
+
```
AaronFeickert, line 208: This will not run if the previous macro fails.

crates/sequencer/src/duty/types.rs
```diff
@@ -364,3 +365,16 @@ impl IdentityData {
...
+        self.key.zeroize(); // Zeroize the private key
+    }
+}
+
+impl Drop for IdentityData {
+    fn drop(&mut self) {
+        #[cfg(feature = "zeroize")]
+        self.zeroize();
+    }
+}
```
AaronFeickert, line 380: Why not derive `Zeroize` and `ZeroizeOnDrop` directly? This is more robust against future struct changes.

AaronFeickert (changes requested):

delbonis (commented):
Instead of manually calling `non_secure_erase`, would it make sense to use [`ZeroizeOnDrop`](https://docs.rs/zeroize/latest/zeroize/derive.ZeroizeOnDrop.html) to give a `Drop` impl?  It should be safe to add it on `IdentityData` for sure.

delbonis:
Oh wait you are actually doing that!  I don't know how I missed that.

crates/sequencer/src/duty/types.rs
delbonis: If I had to guess maybe it's because the feature flag is optional?

AaronFeickert: The derives can be conditional on the flag if needed.

bin/bridge-client/src/modes/operator/bootstrap.rs
krsnapaudel: Good catch. I have updated the code.

crates/sequencer/src/duty/types.rs
krsnapaudel: I have removed the impls and added Zeroize and ZeroizeOnDrop to derive.

bin/strata-sequencer-client/src/main.rs
```diff
@@ -70,13 +72,21 @@ fn main_inner(args: Args) -> Result<()> {
...
+        duty_executor_worker(rpc, duty_rx, handle.clone(), idata_clone),
     );
 
     task_manager.start_signal_listeners();
-    task_manager.monitor(Some(Duration::from_millis(SHUTDOWN_TIMEOUT_MS)))?;
+    let result = task_manager.monitor(Some(Duration::from_millis(SHUTDOWN_TIMEOUT_MS)));
 
-    Ok(())
+    // Zeroize idata
+    if let Ok(mut idata) = Arc::try_unwrap(idata) {
```
bewakes, line 82: Maybe we won't be doing `Arc::try_unwrap` for `idata` in other places but as per the docs, it is strongly recommended to use `Arc::into_inner` if the `Err` variant is not used(like in this case), to prevent potential race conditions and having to lose the value.

https://doc.rust-lang.org/std/sync/struct.Arc.html#method.try_unwrap

bewakes (commented):
Minor comment on `Arc::try_unwrap`. Looks great otherwise.

krsnapaudel:
@bewakes and I discussed whether the following code is actually necessary considering the objects derive/implement `ZeroizeOnDrop`. For tests, we do need this to verify `zeroize` worked. In `bridge-client/src/modes/operator/bootstrap.rs` and `strata-sequencer-client/src/main.rs`, this might not be necessary.

```
// Attempt to erase keypair from ExecHandler and ExecHandler.sig_manager
    if let Some(mut handler) = Arc::into_inner(shared_exec_handler) {
        handler.erase_keypair(); // Ensure keypair is erased before dropping
        handler.sig_manager.erase_keypair();
    }
```

bin/strata-sequencer-client/src/main.rs
AaronFeickert: TIL!

delbonis:
It's *really hard* to test this since Rust makes it hard to access data that's been `Drop`ped.  But we can kinda hack around that because what we really want to test is the *impl* itself.

Something like this in a test should work:

```rs
fn is_all_zero<T>(p: *const T) -> bool {
   let len = std::mem::size_of::<T>()
   let buf = std::slice::from_raw_parts::<_, u8>(p as *const u8, len); // maybe needs some better casting
   buf.iter().all(|v| v == 0)
}

let data: Arc<IdentityData> = ...;
let raw = data.as_ptr();
assert!(!is_all_zero(raw));
std::ptr::drop_in_place(data.as_ptr() as *mut IdentityData);
assert!(is_all_zero(raw));
```

This actually *shouldn't* even invalidate the `IdentityData`!  It would just be a value that'd never normally be constructed.

AaronFeickert:
I've always considered zeroizing to be a best-effort attempt, as noted by the subtleties in the [documentation](https://docs.rs/zeroize/latest/zeroize/). So I wouldn't necessarily go too crazy on complex tests, except for perhaps very custom implementations that don't use vanilla `zeroize` functionality.

crates/sequencer/src/duty/types.rs
```diff
@@ -351,7 +352,7 @@ pub enum IdentityKey {
 ///
 /// This is really just a stub that we should replace
 /// with real cryptographic signatures and putting keys in the rollup params.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Zeroize, ZeroizeOnDrop)]
```
AaronFeickert, line 355: Not strictly needed since the fields are `ZeroizeOnDrop`. But keeping it is probably fine as defense in depth.

crates/sequencer/src/duty/types.rs
```diff
@@ -307,7 +308,7 @@ impl StateUpdate {
 }
 
 /// Describes an identity that might be assigned duties.
-#[derive(Clone, Debug, BorshDeserialize, BorshSerialize)]
+#[derive(Clone, Debug, BorshDeserialize, BorshSerialize, Zeroize, ZeroizeOnDrop)]
```
AaronFeickert, line 311: Can you confirm that this `Buf32` is in fact secret data?

crates/key-derivation/src/sequencer.rs
```diff
@@ -187,4 +181,37 @@ mod tests {
...
+        assert_ne!(derived_chaincode, [0u8; 32]);
+
+        fn bar(keys_clone: Arc<SequencerKeys>) {
+            println!("got keys {}", keys_clone.master);
+        }
+
+        fn foo(keys: Arc<SequencerKeys>) {
+            println!("got keys {}", keys.master);
+            bar(keys.clone());
+        }
```
AaronFeickert, line 207: These will clog up the logs. Are they necessary outside of initial debugging of the test?

crates/key-derivation/src/sequencer.rs
```diff
@@ -56,14 +56,6 @@ impl SequencerKeys {
...
 }
 
-// Manual Drop implementation to zeroize keys on drop.
-impl Drop for SequencerKeys {
-    fn drop(&mut self) {
-        #[cfg(feature = "zeroize")]
-        self.zeroize();
-    }
-}
-
```
AaronFeickert, line 66: Why was this removed? It means this is no longer `ZeroizeOnDrop`, so the marker trait is a lie!

crates/bridge-exec/src/handler.rs
```diff
@@ -111,6 +111,11 @@ where
         Ok(())
     }
 
+    /// Attempts to erase the secret within the keypair.
+    pub fn erase_keypair(&mut self) {
+        self.keypair.non_secure_erase();
+    }
```
AaronFeickert, line 117: Might it make more sense to put this into a `Drop` implementation, so the caller doesn't need to remember to call it?

crates/bridge-sig-manager/src/manager.rs
```diff
@@ -126,6 +126,11 @@ impl SignatureManager {
         entry.ok_or(BridgeSigError::TransactionNotFound)
     }
 
+    /// Attempts to erase the secret within the keypair.
+    pub fn erase_keypair(&mut self) {
+        self.keypair.non_secure_erase();
+    }
```
AaronFeickert, line 132: Might it make more sense to put this into a `Drop` implementation, so the caller doesn't need to remember to call it?

AaronFeickert (changes requested):

crates/sequencer/src/duty/types.rs
krsnapaudel: When I remove them, I see this: 

the method `zeroize_or_on_drop` exists for mutable reference `&mut IdentityKey`, but its trait bounds were not satisfied
the following trait bounds were not satisfied:
`duty::types::IdentityKey: zeroize::Zeroize`
which is required by `duty::types::IdentityKey: zeroize::__internal::AssertZeroize`
`duty::types::IdentityKey: zeroize::ZeroizeOnDrop`
which is required by `&&mut duty::types::IdentityKey: zeroize::__internal::AssertZeroizeOnDrop`
`&mut duty::types::IdentityKey: zeroize::Zeroize`
which is required by `&mut duty::types::IdentityKey: zeroize::__internal::AssertZeroize`

crates/sequencer/src/duty/types.rs
krsnapaudel: From `IdentityData`:
```
pub struct IdentityData {
    /// Unique identifying info.
    pub ident: Identity,
    /// Signing key.
    pub key: IdentityKey,
}
```
So this is not the secret key. Removing them leads to compiler complaints (similar to the previous comment) because we are not implementing our own Zeroize and Drop.

crates/key-derivation/src/sequencer.rs
krsnapaudel: No. They were meant for debugging test runs.

crates/key-derivation/src/sequencer.rs
krsnapaudel: My understanding was L131-132 means `zeroize()` will be called automatically on drop.
```
#[cfg(feature = "zeroize")]
impl ZeroizeOnDrop for SequencerKeys {}
```
Again I am a rust newbie. I discussed this with @bewakes. Any guidance will be appreciated.

crates/sequencer/Cargo.toml
```diff
@@ -36,3 +37,4 @@ rustdoc.all = "warn"
 
 [features]
 debug-utils = ["strata-common/debug-utils"]
+default = ["zeroize"]
```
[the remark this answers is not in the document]
krsnapaudel: I think I added this in the beginning to make compiler happy. Removing.

crates/bridge-exec/src/handler.rs
krsnapaudel: I have put this call inside `impl Drop`. This means we cannot call the following anymore:
```
// Attempt to erase keypair from ExecHandler and ExecHandler.sig_manager
    if let Some(mut handler) = Arc::into_inner(shared_exec_handler) {
        handler.erase_keypair(); // Ensure keypair is erased before dropping
        handler.sig_manager.erase_keypair();
    }
```
I suppose `self.keypair.non_secure_erase();` will get called for ExecHandler and SignatureManager on drop. So we should be fine.

crates/bridge-sig-manager/src/manager.rs
krsnapaudel: Same response as in the previous comment.

krsnapaudel:
Based on discussions with @bewakes and @Rajil1213, relying on automatic drop in `bridge-client/src/modes/operator/bootstrap.rs` and `strata-sequencer-client/src/main.rs`.

bin/strata-sequencer-client/src/main.rs
```diff
@@ -52,7 +52,8 @@ fn main_inner(args: Args) -> Result<()> {
     init_logging(handle);
 
     let config = get_config(args.clone())?;
-    let idata = load_seqkey(&config.sequencer_key)?;
+    let idata = Arc::new(load_seqkey(&config.sequencer_key)?);
+    let idata_clone = Arc::clone(&idata);
```
delbonis, line 56: Can just do `idata.clone()`.

bin/bridge-client/src/modes/operator/bootstrap.rs
```diff
@@ -158,8 +158,11 @@ pub(crate) async fn bootstrap(args: Cli) -> anyhow::Result<()> {
         msg_polling_interval,
     };
 
+    let shared_exec_handler = Arc::new(exec_handler);
+    let exec_handler_clone = shared_exec_handler.clone();
+
     let task_manager = TaskManager {
-        exec_handler: Arc::new(exec_handler),
+        exec_handler: exec_handler_clone,
```
delbonis, line 165: Can just do `shared_exec_handler.clone()`.

Conventionally, you don't need to add the `shared_` prefix either fwiw.

bin/strata-sequencer-client/src/duty_executor.rs
```diff
@@ -67,16 +67,18 @@ where
...
     R: StrataSequencerApiClient + Send + Sync,
 {
     let duty_id = duty.id();
     debug!(%duty_id, ?duty, "handle_duty");
     let duty_result = match duty.clone() {
-        Duty::SignBlock(duty) => handle_sign_block_duty(rpc, duty, duty_id, idata).await,
-        Duty::CommitBatch(duty) => handle_commit_batch_duty(rpc, duty, duty_id, idata).await,
+        Duty::SignBlock(duty) => handle_sign_block_duty(rpc, duty, duty_id, idata.clone()).await,
+        Duty::CommitBatch(duty) => {
+            handle_commit_batch_duty(rpc, duty, duty_id, idata.clone()).await
```
delbonis, line 80: *Possibly* these calls should actually take `&IdentityData` to avoid having to clone the arc.  I am not sure if they go on to move it into another thread (so actually *need* to own it) so I might be wrong.

delbonis (dismissed):
Should use that snippet I sent earlier to verify that `ZeroizeOnDrop` is behaving as expected for at least some types.

crates/key-derivation/src/sequencer.rs
AaronFeickert: No worries! The `ZeroizeOnDrop` trait is just a marker trait when not used in a `derive`. You still need to do a `Drop` implementation that performs the intended zeroization. You don't need to implement the marker trait yourself if the struct in question derives it (it's done for you), which isn't the case here.

crates/bridge-exec/src/handler.rs
```diff
@@ -344,3 +344,53 @@ where
         )
     }
 }
+
+impl<TxBuildContext> Drop for ExecHandler<TxBuildContext>
+where
+    TxBuildContext: BuildContext + Sync + Send,
+{
+    fn drop(&mut self) {
+        // TODO erase keypair inside sig_manager, but sig_manager.keypair is not public.
```
AaronFeickert, line 353: This is already handled.

crates/sequencer/src/duty/types.rs
AaronFeickert: Ah sure, because `IdentityData` contains it as a field. Looking more closely, it's probably easier to remove the derives from both `Identity` and `IdentityData`. The reasoning would be that `Identity` contains no secrets, and the `IdentityKey` contained within `IdentityData` is already `ZeroizeOnDrop`; so when `IdentityData` goes away, so does the key.

AaronFeickert (changes requested):

krsnapaudel:
> Should use that snippet I sent earlier to verify that `ZeroizeOnDrop` is behaving as expected for at least some types.

Apologies @delbonis. On my own, I could not make your code compile. With ChatGPT's help, I did :)

AaronFeickert (changes requested):
I think there may have been some wires crossed from my last review comments. My recommendation is to keep the derived zeroization traits for `IdentityKey` in place, since it represents secret data. This will allow you to remove the derived traits from `IdentityData` and `Identity`, but still have the underlying `IdentityKey` zeroized safely. The overall idea would be to keep zeroization traits as tightly coupled to secret data as possible, which means that higher-level structs don't have to worry about this.

crates/key-derivation/src/sequencer.rs
```diff
@@ -187,4 +189,37 @@ mod tests {
             ChildNumber::Hardened { index } => assert_eq!(index, 0),
         }
     }
+
+    #[test]
+    #[cfg(feature = "zeroize")]
+    fn test_zeroize_idata() {
```
storopoli, line 195: Why this is called `idata` and there's no "`idata`" being tested?

crates/key-derivation/src/sequencer.rs
```diff
@@ -139,9 +131,19 @@ impl Zeroize for SequencerKeys {
 #[cfg(feature = "zeroize")]
 impl ZeroizeOnDrop for SequencerKeys {}
 
+// Manual Drop implementation to zeroize keys on drop.
+impl Drop for SequencerKeys {
+    fn drop(&mut self) {
+        #[cfg(feature = "zeroize")]
+        self.zeroize();
+    }
+}
```
storopoli, line 140: Just a check: is the feature `zeroize` default on `key-derivation` crate?

storopoli (changes requested):

crates/key-derivation/src/sequencer.rs
krsnapaudel: Yes it seems.

crates/key-derivation/src/sequencer.rs
krsnapaudel: Updated the test.

storopoli:
I still think that we should do a similar approach with `ZeroizableXpriv` https://github.com/alpenlabs/strata/blob/42592a9f3ce1ca53c7ed27df090868617493b411/crates/primitives/src/keys.rs#L10-L13 for `ZeroizableKeypair` (with the custom `Deref`/`DerefMut`) and then just replace that everywhere we use `Keypair`.

No need to mess around with tokio or `Arc`s.
Once keypairs are out of scope, BAM they are "zeroed".

storopoli (changes requested):
Now we need to:

1. Replace `ZeroizableKeypair` whenever we use `Keypair` everywhere.
2. We still need to handle the `IdentityData` that contains secret data, but hey it is not `Copy`: https://github.com/alpenlabs/strata/blob/9c08becfaaf9055073b2f579cb60a329e398cd7d/crates/sequencer/src/duty/types.rs#L343-L361 Then you can trivially implement derive `Zeroize`/`ZeroizeOnDrop`. Maybe not, but worth trying.

crates/primitives/src/keys.rs
```diff
@@ -54,6 +54,53 @@ impl Zeroize for ZeroizableXpriv {
...
+impl Zeroize for ZeroizableKeypair {
+    #[inline]
+    fn zeroize(&mut self) {
+        self.0.non_secure_erase();
+    }
+}
+
+#[cfg(feature = "zeroize")]
+impl ZeroizeOnDrop for ZeroizableKeypair {}
+
```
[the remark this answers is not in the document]
AaronFeickert: Do you mean in the case where the struct changes? Or if upstream borks something internally?

crates/primitives/src/keys.rs
AaronFeickert, line 1: The role of the `zeroize` feature here is unclear to me. The module itself is not feature gated, the struct is, some of its implementation isn't, and some of its traits are. This seems really brittle for no reason.

crates/primitives/src/keys.rs
```diff
@@ -54,6 +54,53 @@ impl Zeroize for ZeroizableXpriv {
...
 
+/// A zeroizable on [`Drop`] wrapper around [`Keypair`].
+#[cfg(feature = "zeroize")]
+#[derive(Clone, PartialEq, Eq)]
+pub struct ZeroizableKeypair(Keypair);
+
+impl ZeroizableKeypair {
+    /// Create a new [`ZeroizableKeypair`] from an [`Keypair`].
+    ///
+    /// This should take ownership of the `Keypair` since it is zeroized on drop.
```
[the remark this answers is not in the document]
AaronFeickert: Why not link it?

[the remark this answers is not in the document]
krsnapaudel: Not sure about @AaronFeickert's comments.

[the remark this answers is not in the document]
AaronFeickert: You can directly [link](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html) to items in docstrings.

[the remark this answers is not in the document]
krsnapaudel: I realized linking works for structs, not variable names. Is this correct. Then we should probably leave these variable names as they are.

crates/primitives/src/keys.rs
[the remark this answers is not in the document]
storopoli: upstream borks

crates/primitives/src/keys.rs
storopoli: Yes, agreed.

The feature is default so it's good https://github.com/alpenlabs/strata/blob/9862d4bb2f01099513523045665b4540a96c565a/crates/primitives/Cargo.toml#L35-L36

@krsnapaudel can you please add a `#[cfg(feature = "zeroize")]` in ? https://github.com/alpenlabs/strata/blob/9862d4bb2f01099513523045665b4540a96c565a/crates/primitives/src/lib.rs#L17

Then we can remove all feature-gates inside the `key.rs` what you think Aaron?

storopoli (changes requested):

crates/sequencer/src/duty/types.rs
```diff
@@ -365,3 +366,57 @@ impl IdentityData {
...
+        }
+
+        foo(idata.clone());
+
+        // Drop the last reference, triggering `ZeroizeOnDrop`
+        drop(idata);
+
+        // Now check if the memory has been zeroized
+        assert!(is_all_zero(raw_ptr), "contents {:?}", raw_ptr);
+    }
```
storopoli, line 421: @krsnapaudel You're not dropping the `Arc` that holds the `IdentityData` hence the thing is not getting the drop check and not also being zeroized.

Try to use the same approach as done in `ZeroizableXpriv` tests: https://github.com/alpenlabs/strata/blob/9862d4bb2f01099513523045665b4540a96c565a/crates/primitives/src/keys.rs#L143-L178

storopoli (changes requested):

crates/sequencer/src/duty/types.rs
krsnapaudel: Sure, but why does the test pass locally. If it's Arc not getting dropped, then it should fail. The reason to use Arc was to be able to use Trey's code.

crates/primitives/src/keys.rs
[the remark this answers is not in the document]
AaronFeickert: I suppose if the intent is a variable reference, fine as is!

crates/primitives/src/keys.rs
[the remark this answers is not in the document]
AaronFeickert: One option for guarding against struct changes in manual zeroization that I like is destructuring. It's not foolproof, but it's definitely better on net.

crates/primitives/src/keys.rs
AaronFeickert: That is, the feature doesn't appear to be strictly additive with respect to the rest of the codebase.

crates/primitives/src/keys.rs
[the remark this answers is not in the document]
krsnapaudel: @storopoli There is a test for `ZeroizableKeypair`.

crates/sequencer/src/duty/types.rs
krsnapaudel: @storopoli : Updated the test. It's passing now both locally and here.

storopoli (dismissed):
ACK 9f0e0e912f324fa12ae43e84734c0740c2615a5f
Nice work!

crates/key-derivation/src/sequencer.rs
```diff
@@ -139,9 +131,15 @@ impl Zeroize for SequencerKeys {
 #[cfg(feature = "zeroize")]
 impl ZeroizeOnDrop for SequencerKeys {}
 
+// Manual Drop implementation to zeroize keys on drop.
+impl Drop for SequencerKeys {
+    fn drop(&mut self) {
+        self.zeroize();
```
AaronFeickert, line 137: This won't work without the feature enabled. (But I still question the need for the feature _at all_.)

AaronFeickert (changes requested):

crates/key-derivation/src/sequencer.rs
krsnapaudel: Removed the feature for `zeroize` from `key-derivation` as well.

AaronFeickert (approved):

storopoli (approved):
ACK 6f696fae8230a23c14fdf1ca178e3cac2e632659

krsnapaudel:
@delbonis : I tried to stick with your code to test `zeroize`. In the end, had to switch to @storopoli's example code because the test passed locally, but not in CI. Not sure why. Anyway, would you like to take a final look?

bewakes (approved):

PR #669, opened 2025-02-13
Update config and args

bewakes:
## Description
This PR does the following:
1. Remove a lot of config args from strata-client's `Args` struct. It now only accepts `--datadir`, `--config`, `--rollup-params`, `--rpc-host`, `--rpc-port` and generic override param `-o`.
2. The usage of `-o` is something like `./strata-client --config configfile.toml -o bitcoind.rpc_user=user -o btcio.reader.client_poll_dur_ms=1000 -o exec.reth.rpc_url=http://reth`
3. Anything passed via `-o` will override the contents from config file and the default values.
4. Add `factory/config.py` module corresponding to `strata_config::Config` with default values set and conversion to toml.

**Note** that the previous args passing no longer works and this has impacts on deployment too. Possibly need to update docker stuffs as well.
### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist


bin/strata-client/src/args.rs
```diff
@@ -1,171 +1,222 @@
...
 
-    #[argh(option, description = "bitcoind RPC user")]
-    pub bitcoind_user: Option<String>,
+    #[argh(option, description = "rpc port")]
+    pub rpc_port: Option<u16>,
 
-    #[argh(option, description = "bitcoind RPC password")]
-    pub bitcoind_password: Option<String>,
+    #[argh(option, short = 'o', description = "generic config overrides")]
+    pub overrides: Vec<String>,
```
storopoli, line 56: Can you add docstrings for these fields?

bin/strata-client/src/args.rs
```diff
@@ -1,171 +1,222 @@
...
+}
 
-    /// Max retries for Bitcoin RPC calls.
-    #[argh(option, description = "max retries for bitcoin RPC (default: 3)")]
-    pub bitcoind_retry_count: Option<u8>,
+impl Args {
+    /// Overrides config. First overrides with the generic overrides passed via -o and then
```
storopoli, line 60:
```suggestion
    /// Overrides config. First overrides with the generic overrides passed via `-o` and then
```

bin/strata-client/src/args.rs
```diff
@@ -1,171 +1,222 @@
...
+    /// overrides the result with a couple of commonly passed args.
```
storopoli, line 61: I don't see "a couple" here. This is not that informative. This means that it will override by chunks of pairs of configurations (the "couple")?
Maybe do `s/a couple/some`.

bin/strata-client/src/args.rs
```diff
@@ -1,171 +1,222 @@
...
+        let mut json_config = to_value(&mut *config).expect("Config json serialization failed");
 
-    #[argh(option, short = 'n', description = "L1 network to run on")]
-    pub network: Option<Network>,
+        for (path, val) in parse_overrides(&self.overrides)?.iter() {
+            apply_override(path, val, &mut json_config)?;
+        }
+        // Convert back to json.
+        *config =
+            from_value(json_config).expect("Should be able to create Config from serde json Value");
```
storopoli, line 94: Maybe we should use `anyhow` here instead of `expect()`?

bin/strata-client/src/args.rs
```diff
@@ -1,171 +1,222 @@
...
-        option,
-        description = "max interval between bitcoin RPC retries in ms (default: 1000)"
-    )]
-    pub bitcoind_retry_interval: Option<u64>,
+    /// Override config using the generic overrides. The idea here is to first convert the
+    /// [`Config`] to json and then apply the overrides by splitting the key by a dot.
+    fn override_generic(&self, config: &mut Config) -> anyhow::Result<bool> {
+        let original = config.clone();
+        // Convert config as json
+        let mut json_config = to_value(&mut *config).expect("Config json serialization failed");
```
storopoli, line 87: Maybe we should use `anyhow` here instead of `expect()`?

bin/strata-client/src/args.rs
```diff
@@ -1,171 +1,222 @@
...
+            apply_override(other, str_value, &mut config[key])?;
+        }
+        [] => return Err(anyhow!("Invalid override path")),
+    };
+    Ok(())
+}
 
-    #[argh(option, description = "reth authrpc host:port")]
-    pub reth_authrpc: Option<String>,
+/// Parses a string `T`. If parsing fails, attempts to parse it again by converting it to JSON
```
storopoli, line 129:
```suggestion
/// Parses a string into a `T`. If parsing fails, attempts to parse it again by converting it to JSON
```

The `T` is the type that you're parsing the string into.

crates/config/src/config.rs
```diff
@@ -1,45 +1,54 @@
...
-
-#[derive(Debug, Clone, Deserialize)]
-#[serde(untagged)]
-pub enum ClientMode {
-    FullNode(FullNodeConfig),
-    Sequencer(SequencerConfig),
-}
+const DEFAULT_RPC_PORT: u16 = 8542;
+const DEFAULT_P2P_PORT: u16 = 8543;
+const DEFAULT_DATADIR: &str = "strata-data";
```
storopoli, line 10: Docstrings to at least say which configs these belongs to?
From the context, the `strata-client` full node?

crates/config/src/config.rs
```diff
@@ -1,45 +1,54 @@
...
+    pub p2p_port: u16,
+
+    pub sync_endpoint: Option<String>,
+
     pub l2_blocks_fetch_limit: u64,
+
+    #[serde(default = "default_datadir")]
     pub datadir: PathBuf,
+
     pub db_retry_count: u16,
```
storopoli, line 30: Docstrings also in these fields please?

storopoli (changes requested):
minor nits.
These are awesome, thanks! Nice work.

bin/strata-client/src/args.rs
```diff
@@ -1,171 +1,230 @@
...
+    let mut result = Vec::new();
+    for item in overrides {
+        let (key, value) = item
+            .split_once("=")
+            .ok_or(anyhow!("Invalid override: must be in 'key=value' format"))?;
+        let path: Vec<_> = key.split(".").map(|x| x.to_string()).collect();
+        result.push((path, value.to_string()));
+    }
+    Ok(result)
+}
```
delbonis, line 121: Functional programming heuristic: The way this is written, it's not possible to parse just a single override string.  You always have to parse a *list of* override strings.  The body of the loop is where the hard part is happening, and then we don't need a function just to transform elements of a list.

bin/strata-client/src/args.rs
```diff
@@ -1,171 +1,230 @@
...
-    pub sequencer_rpc: Option<String>,
+/// Apply override to config.
+fn apply_override(path: &[String], str_value: &str, config: &mut Value) -> anyhow::Result<()> {
+    match path {
+        [key] => {
+            config[key] = parse_value(str_value)?;
+        }
+        [key, other @ ..] => {
+            apply_override(other, str_value, &mut config[key])?;
+        }
```
delbonis, line 131: The `config[key]` bit panics if `config` isn't a json dict.  The snippet I sent gives a safer way to handle this that gives smarter error handling.

bin/strata-client/src/args.rs
```diff
@@ -1,171 +1,230 @@
...
-    pub reth_authrpc: Option<String>,
+/// Parses a string into a `T`. If parsing fails, attempts to parse it again by converting it to
+/// JSON string i.e. surrounds the string with double quotes.
+/// If the second deserialization attempt also fails, it returns the error from the first attempt.
+fn parse_value<T: DeserializeOwned>(str_value: &str) -> Result<T, serde_json::Error> {
+    match serde_json::from_str(str_value) {
+        Ok(v) => Ok(v),
+        Err(e) => match serde_json::from_str::<T>(&format!("\"{str_value}\"")) {
+            Ok(v) => Ok(v),
+            Err(_) => Err(e),
```
delbonis, line 145: This is messy because it's trying to parse all possible json values.  Don't try to, constrain behavior.  It's only reasonable right now to support bools, ints, and strings.  When the user passes something they expect to be a string then it implicitly needs to be a valid "the chars between the quotes" of a json string token, so it'd break unexpectedly if they pass something that has a `\` in it, like maybe someone running Windows would.

bin/strata-client/src/args.rs
```diff
@@ -1,171 +1,230 @@
...
+            overridden = true
+        }
+        if let Some(rpc_host) = &self.rpc_host {
+            config.client.rpc_host = rpc_host.to_string();
+            overridden = true
+        }
+        if let Some(rpc_port) = &self.rpc_port {
+            config.client.rpc_port = *rpc_port;
+            overridden = true
+        }
```
delbonis, line 86: These can be lowered to just being overrides instead of having to maintain additional special casing for these.

```rs
fn extract_cmdline_override_strs(args: &Args) -> anyhow::Result<Vec<Override>> {
    let mut overrides = Vec::new();

    if let Some(dd) = &self.datadir {
        overrides.push(Override::new("client.datadir", Value::String(dd.to_owned()));
    }

    // the rest

    Ok(overrides)
}
```

So really you'd have some `.get_overrides()` fn that just returns a list of overrides, the ones explicitly added and then the ones generated from the other options like `-d` and `-p`.  Also I think this is missing a few options.

This pattern also makes it easy when we want to parse values from the env, it's all just the override system.

bin/strata-client/src/args.rs
```diff
@@ -1,171 +1,230 @@
...
+        let mut json_config = to_value(&mut *config).context("Config json serialization failed")?;
 
-    #[argh(option, short = 'n', description = "L1 network to run on")]
-    pub network: Option<Network>,
+        for (path, val) in parse_overrides(&self.overrides)?.iter() {
+            apply_override(path, val, &mut json_config)?;
+        }
+        // Convert back to json.
+        *config = from_value(json_config)
+            .context("Should be able to create Config from serde json Value")?;
```
delbonis, line 102: These contexts are for adding more information about what the program was trying to do when the fn was called, not just literally what went wrong if there were to be an error.  That would be the job of the inner error type.

functional-tests/factory/factory.py
```diff
@@ -188,26 +202,43 @@ def create_fullnode(
...
+            "-o", f"bitcoind.rpc_password={bitcoind_config["bitcoind_pass"]}",
+            "-o", "bitcoind.network=regtest",
+
+            # reth
+            "-o", f"exec.reth.rpc_url={reth_config["reth_socket"]}",
+            "-o", f"exec.reth.secret={reth_config["reth_secret_path"]}",
+            "--sequencer"
+
+            # client
+            "-o", f"client.sync_endpoint={sequencer_rpc}",
```
delbonis, line 238: This is good to see here, but really we should be writing all of these to the config file instead of just building the whole configuration as overrides (and same in above case).  It's also less likely to silently break if we use the nice types you also added.

functional-tests/factory/config.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+@dataclass
+class Config:
+    client: ClientConfig
+    bitcoind: BitcoindConfig
+    btcio: BtcioConfig
+    sync: SyncConfig
+    exec: ExecConfig
+    relayer: RelayerConfig
```
delbonis, line 84: Can't we also use `field(default_factory=...)` for these?

delbonis (changes requested):
As discussed in DMs, the round trip through json and the weird special casing of strings is a problem.  This forces that the config be well-formed before applying the overrides, which would breaks use cases like where the user has an incomplete config wants to "finish" it by setting overrides at run time.  Maybe that specific scenario isn't particularly common, but it makes it generally more flexible if we apply the overrides on the toml parsing first before trying to interpret it concretely.

crates/config/src/config.rs
```diff
@@ -5,28 +5,38 @@ use serde::{Deserialize, Serialize};
 
 use crate::{bridge::RelayerConfig, btcio::BtcioConfig};
 
+/// Default values for some of the [`ClientConfig`] attributes.
 const DEFAULT_RPC_PORT: u16 = 8542;
 const DEFAULT_P2P_PORT: u16 = 8543;
 const DEFAULT_DATADIR: &str = "strata-data";
```
storopoli, line 11: This docstring only applies to `DEFAULT_RPC_PORT`.
Can you please add for the ones?

storopoli (dismissed):
ACK 7e15fca7b38b30a0f37e2b3e469defdf9f3f4264

crates/config/src/config.rs
bewakes: Oops..

bin/strata-client/src/args.rs
bewakes: Truee.

functional-tests/factory/factory.py
bewakes: Yeah good point. Changed.

functional-tests/factory/config.py
bewakes: We can totally! changed.

bewakes:
@delbonis , made the changes. Seems like I was wrong about using json would let us dynamically parse stuffs. Removed all the json manipulation now.

bin/strata-client/src/args.rs
```diff
@@ -1,171 +1,232 @@
...
+            overrides.push(format!("client.rpc_host={}", rpc_host));
+        }
+        if let Some(rpc_port) = &self.rpc_port {
+            overrides.push(format!("client.rpc_port={}", rpc_port));
+        }
 
-    /// Max retries for Bitcoin RPC calls.
-    #[argh(option, description = "max retries for bitcoin RPC (default: 3)")]
-    pub bitcoind_retry_count: Option<u8>,
+        overrides
```
delbonis, line 87: You can do `format!("foo.bar={bar}")` instead of having to put it at the end.

delbonis (changes requested):

bin/strata-client/src/args.rs
```diff
@@ -1,171 +1,232 @@
...
+        let (path, val) = res?;
+        apply_override(&path, &val, table)?;
+    }
 
-    #[argh(option, description = "sequencer rpc host:port")]
-    pub sequencer_rpc: Option<String>,
+    toml::Value::Table(table.clone())
+        .try_into()
+        .map_err(|_| ConfigError::ConfigNotParseable)
+}
```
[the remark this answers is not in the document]
bewakes: Yeah, this makes sense. Agree that the `apply_overrides` is a bit weird. changed.

bin/strata-client/src/args.rs
```diff
@@ -74,65 +73,54 @@ impl Args {
...
+            table.insert(path.to_string(), value);
             Ok(())
         }
-        [key, other @ ..] => {
+        Some((key, rest)) => {
             if let Some(t) = table.get_mut(key).and_then(|v| v.as_table_mut()) {
-                apply_override(other, str_value, t)
+                apply_override(rest, value, t)
             } else if table.contains_key(key) {
                 Err(ConfigError::TraversePrimitiveAt(key.to_string()))
```
delbonis: Actually I know this is what I named it in the code I sent but it's not strictly correct since it might be an array.  Maybe `TraverseNonTableAt`?  I'm not sure.  Maybe this doesn't matter.

bin/strata-client/src/errors.rs
```diff
@@ -43,13 +39,7 @@ pub enum ConfigError {
     #[error("can't traverse into primitive: {0}")]
```
delbonis, line 39: But regarding other comment, it would make sense for the error *message* to be more correct here, it's plausible someone might think a path like `foo.7.bar` is valid for trying to traverse into index 7 of a toml list.

delbonis: maybe `tried to traverse into non-table key: {0}`?

bin/strata-client/src/args.rs
bewakes: Yes, I too think that `TraverseNonTableAt` is more precise.

bin/strata-client/src/args.rs
```diff
@@ -74,65 +73,54 @@ impl Args {
     /// Overrides passed directly as args and not as overrides.
     fn get_direct_overrides(&self) -> Vec<String> {
         let mut overrides = Vec::new();
+        if self.sequencer {
+            overrides.push("client.is_sequencer=true".to_string());
+        }
         if let Some(datadir) = &self.datadir {
-            overrides.push(format!("client.datadir={}", datadir.to_string_lossy()));
+            overrides.push(format!("client.datadir={datadir:?}"));
```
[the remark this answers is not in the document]
bewakes: Yep, used `.to.str()`.

sapinb (approved):

PR #672, opened 2025-02-15
Use HeaderVerificationState on Chainstate

prajwolrg:
## Description

- Consistent `DepositIntent` type
- Handle reorg on `HeaderVerificationState`
- Make `HeaderVerificationState` part of `ChainState`

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This will be used in the next PR to verify that all the deposits on EL were originated on L1, as part of the proof statement.

## Checklist


storopoli:
Can `dest_ident ` not be a `Vec<u8>` in and be a strong typed field?

https://github.com/alpenlabs/strata/blob/d8a19921f556024af84a6c1c4a8cb2068951b1d6/crates/state/src/bridge_ops.rs#L67-L74

crates/test-utils/src/l2.rs
```diff
@@ -78,8 +81,8 @@ pub fn gen_params_with_seed(seed: u64) -> Params {
             da_tag: "strata-da".to_string(),
             block_time: 1000,
             cred_rule: block_credential::CredRule::Unchecked,
-            horizon_l1_height: 40318,
-            genesis_l1_height: 40320, // we have mainnet blocks from this height test-utils
+            horizon_l1_height: 40319,
+            genesis_l1_height: 40321, // we have mainnet blocks from this height test-utils
```
storopoli, line 85: We had an off-by-one error all this time then?

storopoli (approved):
Very nice! Thanks!
Minor nits, but already approving.

crates/test-utils/src/l2.rs
prajwolrg: Thanks for the catch. No we didn't have one-by-error.
I was working to handle the reorg support with the blocks at epoch boundaries, and had changed that. Reverted back again by adding tests elsewhere.

delbonis:
@storopoli @prajwolrg That would break when we try to support other execution envs than the EVM (or I guess something else that supports non-disambiguated 20 byte addrs).  Right now it's fine to just have operators ignore deposit requests with non-20-byte addrs.

PR #674, opened 2025-02-17
STR-985: Make `ArbitraryGenerator` generate more entropy on the fly

storopoli:
## Description

The default was 65kb, which was even called for `u64`, e.g. `BitcoinAmount`.
So let's make `ArbitraryGenerator` reasonable where the default is 1kb and calling `new_with_size()` for bigger sizes.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

STR-985

### DIFF
--- a/bin/prover-client/src/task_tracker.rs
+++ b/bin/prover-client/src/task_tracker.rs
@@ -221,7 +221,7 @@ mod tests {
     fn gen_task_with_deps(n: u64) -> (ProofKey, Vec<ProofKey>) {
         let mut deps = Vec::with_capacity(n as usize);
         let host = ProofZkVm::Native;
-        let mut gen = ArbitraryGenerator::new();
+        let mut gen = ArbitraryGenerator::new_with_size(2_048);
 
         let start: L1BlockId = gen.generate();
         let end: L1BlockId = gen.generate();
--- a/bin/strata-client/src/extractor.rs
+++ b/bin/strata-client/src/extractor.rs
@@ -599,7 +599,7 @@ mod tests {
         let empty_deposits = empty_chain_state.deposits_table_mut();
         let mut deposits_table = DepositsTable::new_empty();
 
-        let mut arb = ArbitraryGenerator::new();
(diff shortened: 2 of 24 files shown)


delbonis:
The way the ticket was written, we can keep the comfortable buffer sizes (or even increase the default), and it would simply only generate the precise amount of randomness requested when asked for it.

> We should keep track of how much entropy is being asked for and generate it on the fly into a preallocated buffer.  This buffer won’t be expanded (as that could break lifetimes), but we’ll only generate the randomness as needed.

^This is really the right way to do it, which I should have done when I reworked it last.  Sure we can keep `new_with_size`, but that would just change the size of the backing buffer.

bewakes (commented):
- One thing that crossed my mind is that if we want dynamic size estimation, we could probably use `std::mem::size_of::<T>()`. But this wouldn't work for structs that contain heap allocated items like `Vec`. I really wish there was some way around this, because otherwise we have to try hard to estimate the size of the resulting object.
- Another thing is that why are we opting for exact sizes instead of bit operations(like `1 << 12`) while using `new_with_size()`. The latter seems easier to reason about just by looking at it and we can also easily control how many bits/bytes we are allocating.

storopoli:
@delbonis so the solution is doing the reverse of what I did?
Keep the big default and move everything to `new()` and not `new_with_size()`?

delbonis:
@bewakes I agree it would be better if it worked like that.  But the `Arbitrary` interface demands borrowed slices so you can generate types like `struct Foo<'a>(&'a [u8]);`.  Also yeah it might have been slightly simpler to define the default size with `1 << n`.

@storopoli It's still good practice to constrain it where it's known to be safe so we use less memory.  But this would be a more complete solution.

delbonis:
What's the status of this?

storopoli:
No bandwidth :/

storopoli:
Closing this for now and putting the JIRA ticket on hold.

PR #673, opened 2025-02-17
BIP-43-compatible Derivation Paths

storopoli:
## Description

We need to respect some protected ranges (0-19,999) in the [BIP-43](https://github.com/bitcoin/bips/blob/master/bip-0043.mediawiki) for the first (purpose) child number.

This also follows the spec of the bridge's Secret Service.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [x] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [x] New and existing tests pass with my changes.

PR #675, opened 2025-02-18
fix: `key-derivation` and `datatool`

storopoli:
## Description

In #673, we made the operator's derived p2p messages and MuSig2 signing keys as full hardened paths.
This broke the `key-derivation` and `datatool` crates that expected them to be non-hardened paths, hence it also broke the functional tests with respect to the bridge.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/bin/datatool/src/args.rs
+++ b/bin/datatool/src/args.rs
@@ -53,18 +53,11 @@ pub(crate) struct SubcXpriv {
 #[argh(
     subcommand,
     name = "genseqpubkey",
-    description = "generates a sequencer pubkey from a master xpriv"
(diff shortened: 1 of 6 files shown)


storopoli:
Ideally we would go forward with this PR. But it would need to do changes in the `bridge-client`, and a bunch of other things.
Since the Secret Service crate in the `strata-bridge` already has the correct spec with the hardened paths, eventually we'll remove the bridge code from `strata`.

Closing this in favor of #676

PR #679, opened 2025-02-21
`HeaderVerificationState` improvements

prajwolrg:
## Description

- Update next block target calculation to use `rust-bitcoin` instead of custom implementation. This also removes the dependency of `ethnum` crate which was based on previous implementation.
- Return errors instead of panicking on update to the state.
- Support reorg handling

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Although we do not need reorg support for TN1, it might be needed in the future.

## Checklist

### DIFF
--- a/crates/btcio/src/reader/query.rs
+++ b/crates/btcio/src/reader/query.rs
@@ -5,7 +5,7 @@ use std::{
 };
 
 use anyhow::bail;
-use bitcoin::{Block, BlockHash};
+use bitcoin::{params::Params as BtcParams, Block, BlockHash, CompactTarget};
 use secp256k1::XOnlyPublicKey;
 use strata_config::btcio::ReaderConfig;
 use strata_l1tx::{
@@ -15,8 +15,8 @@ use strata_l1tx::{
 use strata_primitives::{
     block_credential::CredRule,
     l1::{
-        get_btc_params, get_difficulty_adjustment_start_height, BtcParams, HeaderVerificationState,
-        L1BlockCommitment, L1BlockId, TimestampStore,
+        get_relative_difficulty_adjustment_height, EpochTimestamps, HeaderVerificationState,
+        L1BlockCommitment, L1BlockId, TimestampStore, TIMESTAMPS_FOR_MEDIAN,
     },
     params::Params,
 };
@@ -366,13 +366,44 @@ async fn process_block<R: ReaderRpc>(
     status_updates.push(L1StatusUpdate::CurHeight(height));
     status_updates.push(L1StatusUpdate::CurTip(l1blkid.to_string()));
 
-    let l1_verification_state =
-        fetch_verification_state(ctx.client.as_ref(), height, &get_btc_params()).await?;
+    let l1_verification_state = fetch_verification_state(
+        ctx.client.as_ref(),
+        height,
+        ctx.params.rollup.l1_reorg_safe_depth,
+    )
+    .await?;
 
     let block_ev = L1Event::BlockData(block_data, state.epoch(), l1_verification_state);
     let l1_events = vec![block_ev];
 
     Ok((l1_events, l1blkid))
+}
+
+/// Retrieves the timestamps for a specified number of blocks starting from the given block height,
+/// moving backwards. For each block from `height` down to `height - count + 1`, it fetches the
+/// block’s timestamp. If a block height is less than 1 (i.e. there is no block), it inserts a
+/// placeholder value of 0. The resulting vector is then reversed so that timestamps are returned in
+/// ascending order (oldest first).
+async fn fetch_block_timestamps_ascending(
+    client: &impl ReaderRpc,
+    height: u64,
+    count: usize,
+) -> anyhow::Result<Vec<u32>> {
+    let mut timestamps = Vec::with_capacity(count);
+
+    for i in 0..count {
+        let current_height = height.saturating_sub(i as u64);
+        // If we've gone past block 1, push 0 as a placeholder.
+        if current_height < 1 {
+            timestamps.push(0);
+        } else {
+            let header = client.get_block_header_at(current_height).await?;
+            timestamps.push(header.time);
+        }
+    }
+
+    timestamps.reverse();
+    Ok(timestamps)
 }
 
 /// Returns the [`HeaderVerificationState`] after applying the given block height. This state can be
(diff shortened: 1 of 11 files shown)


Rajil1213 (approved):

crates/primitives/src/l1/timestamp_store.rs
```diff
@@ -2,17 +2,22 @@ use arbitrary::Arbitrary;
...
 
-/// The number of timestamps stored in the buffer.
-/// We use a fixed-size array of 11 elements because, according to Bitcoin consensus rules,
-/// we need to check that a block's timestamp is not lower than the median of the last eleven
-/// blocks' timestamps.
-const N: usize = 11;
+/// The number of timestamps used for calculating the median.
+/// According to Bitcoin consensus rules, we need to check that a block's timestamp
+/// is not lower than the median of the last eleven blocks' timestamps.
+pub const N: usize = 11;
```
sapinb, line 8: Can we call it something better than `N` ? Characters are not at a premium here 😆 .
Especially as we are exporting this outside this file.

crates/primitives/src/l1/timestamp_store.rs
```diff
@@ -2,17 +2,22 @@ use arbitrary::Arbitrary;
...
 
 /// The middle index for selecting the median timestamp.
 /// Since N is odd, the median is the element at index 5 (the 6th element)
 /// after the timestamps are sorted.
-const MID: usize = 5;
+pub const MID: usize = 5;
```
sapinb, line 13: Same here. If this is exported outside, let the name be descriptive enough.

crates/btcio/src/reader/query.rs
```diff
@@ -387,57 +420,80 @@ async fn process_block<R: ReaderRpc>(
...
+    // This index indicates where the next timestamp would be inserted.
+    let ring_buffer_head = (block_height as usize) % total_timestamp_count;
+    let timestamp_history = TimestampStore::new_with_head(&timestamps, ring_buffer_head);
+
+    // Compute the block ID for the verified block.
+    let block_id: L1BlockId = block_header.block_hash().into();
+
+    // Fetch the target for the next block
+    let next_block_target = client
+        .get_block_header_at(block_height + 1)
```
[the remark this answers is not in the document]
prajwolrg: My assumptions was wrong. Have updated the code to support this case.

MdTeach (dismissed):

Rajil1213 (approved):

crates/btcio/src/reader/query.rs
```diff
@@ -598,28 +663,68 @@ mod test {
...
+        assert_eq!(header_vs.next_block_target, new_header_vs.next_block_target);
+        assert_eq!(header_vs.epoch_timestamps, new_header_vs.epoch_timestamps);
+        assert_eq!(
+            header_vs.block_timestamp_history,
+            new_header_vs.block_timestamp_history
+        );
+        assert_eq!(
+            header_vs.last_verified_block,
+            new_header_vs.last_verified_block
+        );
```
bewakes, line 728: Since `HeaderVerificationState` implements `Eq` and seems like all the fields are being checked for equality, maybe it makes sense to directly do `assert_eq!(header_vs, new_header_vs)`?

bewakes: Just noticed that `last_verified_block` is checked twice and `accumulated_pow` is not checked.

crates/btcio/src/reader/query.rs
```diff
@@ -387,57 +420,89 @@ async fn process_block<R: ReaderRpc>(
...
-    let head = (height + 1) as usize % N;
-    let last_11_blocks_timestamps = TimestampStore::new_with_head(timestamps, head);
-
-    let l1_blkid: L1BlockId = vh.block_hash().into();
+            current_epoch_start_height
+        };
+    let previous_epoch_start_header = client
+        .get_block_header_at(previous_epoch_start_height)
+        .await
+        .unwrap_or(current_epoch_start_header);
```
sapinb, line 449: `.unwrap_or(current_epoch_start_header);` 
Wont single network error here make the state logically inconsistent ?
Shouldnt this just error out if it cant get it (or retry network errors ? )

crates/primitives/src/l1/header_verification.rs
```diff
@@ -32,277 +31,395 @@ use crate::{buf::Buf32, hash};
...
+
+        for header in &headers {
+            verification_state
+                .check_and_update_full(header, &MAINNET)
+                .unwrap();
+        }
+        let before_vs = verification_state.clone();
+
+        verification_state
+            .reorg(&headers, &headers, &MAINNET)
```
bewakes, line 381: Seems like this checking valid changes when the same headers are passed to reorg. Should this not be testing for different reorg headers?

crates/btcio/src/reader/query.rs
prajwolrg: Yes, this is wrong. Nice catch!

bewakes (commented):
Looks good mostly. A couple of queries though.

crates/btcio/src/reader/query.rs
```diff
@@ -366,15 +366,48 @@ async fn process_block<R: ReaderRpc>(
...
     Ok((l1_events, l1blkid))
 }
 
+/// Retrieves the timestamps of a specified number of blocks from a given height in descending
+/// order.
+///
+/// This fetches timestamps of `height`, `height-1`, `height-2`, …, down to `height - count + 1` but
+/// returns them in ascending order so that the oldest timestamp is first and the newest is last.
+///
+/// If the provided `height` is less than 1, returns a vector containing a single timestamp of 0.
```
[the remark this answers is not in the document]
prajwolrg: This was done to also handle the case where the `height` is less than `count`. But yes, this feels weird. I have updated the docstring to do a better job.

crates/btcio/src/reader/query.rs
prajwolrg: Yes, this was avoided earlier because the `fetch_verification_state` always set `accumulated_pow` to zero. 
Removed `accumulated_pow` as a whole for now since we do not need it anywhere else. So now can be directly checked.

crates/primitives/src/l1/header_verification.rs
prajwolrg: We can pass a different reorg header, but I couldn't find the ones for mainnet.

bewakes (approved):

MdTeach (approved):

sapinb (approved):
Reorg related logic will not be used in current context. Otherwise looks good.

Rajil1213 (approved):

crates/btcio/src/reader/query.rs
```diff
@@ -387,57 +418,87 @@ async fn process_block<R: ReaderRpc>(
...
+        };
+    let previous_epoch_start_header = client
+        .get_block_header_at(previous_epoch_start_height)
+        .await?;
+
+    // Fetch the block header at the height
+    let block_header = client.get_block_header_at(block_height).await?;
+
+    // Increase the count to include additional timestamps to safely cover potential reorg depths.
+    let total_timestamp_count = TIMESTAMPS_FOR_MEDIAN + l1_reorg_safe_depth as usize;
```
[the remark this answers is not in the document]
prajwolrg: Created a [spike ticket](https://alpenlabs.atlassian.net/browse/STR-1100) to track this and resolving it for now.

PR #683, opened 2025-02-22
Str 1071 strata cicd

purusang:
## Description
This pr introduces continuous deployment to strata-client, strata-sequencer-client and strata-prover.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


storopoli (approved):
ACK 4a33b91529c25c844e7a2f97085cf37e37ddce07

Nice job! Thanks for being patient with zizmor and the security lints.

voidash (approved):

.github/workflows/cd.yml
```diff
@@ -0,0 +1,215 @@
...
+# changed programs. It then updates the Helm values with the new Docker image tags and
+# commits and pushes the changes to the deployments repo.
+
+name: Continuous Deployment
+on:
+  pull_request:
+    types:
+      - closed
+    branches:
+      - main
```
sapinb, line 12: will this workflow run if I close a PR without merging it ?
ie. just create a PR to main and close it without merging ?

sapinb (changes requested):

.github/workflows/cd.yml
sapinb: on:
  push:
    branches
      - main

Wont this be more appropriate ? Was there any issue when doing this?

.github/workflows/cd.yml
```diff
@@ -0,0 +1,215 @@
...
+        required: true
+        default: 'staging'
+        type: choice
+        options:
+          - staging
+          # an extra option for running the workflow manually in a separate cluster
+          - staging1
+          # currently we only desire to support cd to staging(s), production is only
+          # for future reference
+          - production
```
sapinb, line 27: lets not add invalid options here and ensure arbitrary values cannot be added

.github/workflows/cd.yml
purusang: > will this workflow run if I close a PR without merging it ?

It wont.

purusang: > Wont this be more appropriate ? Was there any issue when doing this?

No issue at all. It was because we were following pr integrations all the time.

purusang: @sapinb what do you think on @Rajil1213 's proposal on triggering only on release tags?

.github/workflows/cd.yml
```diff
@@ -0,0 +1,215 @@
...
+            strata-reth:
+              - 'bin/strata-reth/**'
+            prover-client:
+              - 'bin/prover-client/**'
+      - name: Export change detection outputs
+        env:
+          STRATA_CLIENT_CHANGED_OUTPUT: ${{ steps.filter.outputs.strata-client }}
+          STRATA_SEQUENCER_CHANGED_OUTPUT: ${{ steps.filter.outputs.strata-sequencer-client }}
+          STRATA_RETH_CHANGED_OUTPUT: ${{ steps.filter.outputs.strata-reth }}
+          PROVER_CLIENT_CHANGED_OUTPUT: ${{ steps.filter.outputs.prover-client }}
```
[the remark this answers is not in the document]
purusang: Agree. So we should dockerize all the binaries unconditionally.

.github/workflows/cd.yml
sapinb: @Rajil1213 We are approaching next release so we will also need tag based deployment for releases/* branch. But for main branch, going for continuous deployment (ie. on every push to main) is ok. This is going to be deployed to staging where we do want to check that things work before we do actual releases.

delbonis (dismissed):
This is still a manual trigger right?

purusang:
> This is still a manual trigger right?

This is both. On-push-main or manual dispatch.

storopoli (approved):
ACK fd3d088a611d47f8beeb213a05570c2e89b9d234

MdTeach (approved):

PR #684, opened 2025-02-24
Use `HeaderVerificationState` in `Chainstate`

prajwolrg:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/state/src/l1/view.rs
+++ b/crates/state/src/l1/view.rs
@@ -1,6 +1,8 @@
 use arbitrary::Arbitrary;
 use borsh::{BorshDeserialize, BorshSerialize};
-use strata_primitives::l1::{L1BlockCommitment, L1BlockId, L1HeaderRecord};
+use strata_primitives::l1::{
+    HeaderVerificationState, L1BlockCommitment, L1BlockId, L1HeaderRecord,
+};
 
 /// Describes state relating to the CL's view of L1.  Updated by entries in the
 /// L1 segment of CL blocks.
@@ -18,6 +20,9 @@ pub struct L1ViewState {
 
     /// The "safe" L1 block header.  This block is the last block inserted into the L1 MMR.
     pub(crate) safe_block_header: L1HeaderRecord,
+
+    /// State against which the new L1 block header are verified
+    pub(crate) header_vs: HeaderVerificationState,
 }
 
 impl L1ViewState {
@@ -26,12 +31,14 @@ impl L1ViewState {
         horizon_height: u64,
         genesis_height: u64,
         genesis_trigger_block: L1HeaderRecord,
+        header_vs: HeaderVerificationState,
     ) -> Self {
         Self {
             horizon_height,
             genesis_height,
             safe_block_height: genesis_height,
             safe_block_header: genesis_trigger_block,
+            header_vs,
         }
     }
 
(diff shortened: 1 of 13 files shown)


bewakes (commented):
This is great that now we have long needed l1's continuity + pow validation in chain transition.

Nevertheless, I am still inclined towards not storing genesis verification state(VS) in manifests at all if we don't need to store verification state for every l1 manifest. Using manifest as a store **only** for the initial verification state seems a bit non-ideal to me. Sure, it doesn't pose any problems: if it is None then it is not genesis block, if it is Some then it is genesis l1 block. But that does leave room for interpretation for what's there in the manifest database.

I am not entirely against this, but I think more ideal would be to have a special L1Event or something similar that would initialize the VS in Chainstate without having to hack around making a field as Optional in the database.

cc: @delbonis

crates/state/src/l1/view.rs
```diff
@@ -26,12 +31,14 @@ impl L1ViewState {
...
         genesis_height: u64,
         genesis_trigger_block: L1HeaderRecord,
+        header_vs: HeaderVerificationState,
     ) -> Self {
         Self {
             horizon_height,
             genesis_height,
             safe_block_height: genesis_height,
             safe_block_header: genesis_trigger_block,
+            header_vs,
```
sapinb, line 41: Can header verification state be stored as part of epoch summaries instead of directly in chainstate ?
This is only used / updated at epoch boundaries and is tied up to each epoch. I think it would be better to not save a copy of this on every chainstate entry.

sapinb (approved):

sapinb:
> This is great that now we have long needed l1's continuity + pow validation in chain transition.
> 
> Nevertheless, I am still inclined towards not storing genesis verification state(VS) in manifests at all if we don't need to store verification state for every l1 manifest. Using manifest as a store **only** for the initial verification state seems a bit non-ideal to me. Sure, it doesn't pose any problems: if it is None then it is not genesis block, if it is Some then it is genesis l1 block. But that does leave room for interpretation for what's there in the manifest database.
> 
> I am not entirely against this, but I think more ideal would be to have a special L1Event or something similar that would initialize the VS in Chainstate without having to hack around making a field as Optional in the database.
> 
> cc: @delbonis

My opinion is to avoid adding an event that will only ever be triggered once. I agree the optional field is not the best option, but a special one-off event isnt either.

Or could we have a `L1BlockProvider` instead, that will integrate fetching from db OR from remote if unavailable in db ?
Would potentially making rpc calls during client state transition be acceptable ?

bewakes:
> My opinion is to avoid adding an event that will only ever be triggered once

True that it will be triggered only once, but it is a significant event and just by looking at the data types it becomes clear what we are trying to do. Indicating some special case by an `Option` or an empty array implicitly in database just feels like something that might backfire in the future.

>  Would potentially making rpc calls during client state transition be acceptable ?

That would indeed simplify things. Doing this is probably better than what we are doing in the db in the PR. But again, the idea is to not have client transition to deal with bitcoin.
As you suggested, having an `L1BlockProvider` is also a great idea. A minor concern is that we will have now two `Arc`s of `L1Db`: one in `NodeStorage` and one in the `L1BlockProvider`.

Another thing that we might go for is initializing the L1Reader with scan start height that's just early enough to include the block needed for `HeaderVerificationState`, similar to what you mentioned.

sapinb (dismissed):

bewakes (dismissed):

bewakes (dismissed):

PR #688, opened 2025-02-25
feat(btcio): `getrawtransaction` verbosities zero and one

storopoli:
## Description

- Adds `btcio`'s capabilities to `getrawtransactions` with verbosities zero and one.
- Proper types.
- Tests.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Needed by the bridge.

## Checklist

### DIFF
--- a/crates/btcio/src/rpc/types.rs
+++ b/crates/btcio/src/rpc/types.rs
@@ -5,7 +5,7 @@ use bitcoin::{
     address::{self, NetworkUnchecked},
     block::Header,
     consensus::{self, encode},
-    Address, Amount, Block, BlockHash, SignedAmount, Transaction, Txid,
+    Address, Amount, Block, BlockHash, SignedAmount, Transaction, Txid, Wtxid,
 };
 use serde::{
     de::{self, IntoDeserializer, Visitor},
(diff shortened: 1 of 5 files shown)


Rajil1213 (dismissed):
I know that these function names correspond to what bitcoin-core calls them but I really would like more descriptive names (instead of having to check what the meaning of each verbosity value is). Other than that, the PR looks good. Thank you for getting to this so quickly!

crates/btcio/src/rpc/types.rs
```diff
@@ -165,6 +165,42 @@ pub struct GetBlockVerbosityOne {
...
+}
+
+/// Result of JSON-RPC method `getrawtransaction` with verbosity set to 1.
+///
+/// Method call: `getrawtransaction "txid" ( verbosity )`
+#[derive(Clone, PartialEq, Eq, Debug, Deserialize, Serialize)]
+pub struct GetRawTransactionVerbosityOne {
+    pub in_active_chain: Option<bool>,
+    #[serde(deserialize_with = "deserialize_tx")]
+    pub hex: Transaction,
```
bewakes, line 191: Having a `Transaction` indicated by the name `hex` is probably misleading. I would suggest using `serde(rename = "hex")` along with the `deserialize_with` if it makes sense. Or we can possibly leave hex as `String` as is if there's a need to closely represent the bitcoin's rpc.

bewakes (changes requested):

crates/btcio/src/rpc/types.rs
Rajil1213: Agreed. LSPs will tell you that it's a Transaction but I recently got burnt by this. So, I'd prefer the renaming approach too.

storopoli: Good point Bibek! Will change, thanks for the suggestion.

storopoli: Done in 40cc6d36d16b6b8809af996ff42790e281374129

storopoli:
Bump @Rajil1213

Rajil1213 (approved):
I thought this had already been merged. *sighs*

PR #705, opened 2025-03-03
STR-482 improved master xpriv resolution

delbonis:
## Description

This just reworks how we resolve the master xpriv to clean it up a bit and add very loud errors in the right places.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

* STR-482

### DIFF
--- a/bin/bridge-client/src/modes/operator/bootstrap.rs
+++ b/bin/bridge-client/src/modes/operator/bootstrap.rs
@@ -1,6 +1,6 @@
 //! Module to bootstrap the operator node by hooking up all the required services.
 
-use std::{path::PathBuf, sync::Arc, time::Duration};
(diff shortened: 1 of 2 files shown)


storopoli (commented):
Would be really nice if you could in this PR fix that these should all be hardened:
https://github.com/alpenlabs/strata/blob/de3a4ffafe8b679f73de1cfd23b0cebae81549f6/crates/primitives/src/constants.rs#L88
https://github.com/alpenlabs/strata/blob/de3a4ffafe8b679f73de1cfd23b0cebae81549f6/crates/primitives/src/constants.rs#L103

But the datatool explodes if we do this because it converts xpub -> xpub and that fails if is hardened.
This is the bridge spec that we're using, all paths should be hardened.

PR #706, opened 2025-03-03
Full chainstate da

sapinb:
## Description

V3

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


crates/l1tx/src/utils.rs
```diff
@@ -94,3 +94,16 @@ pub fn get_operator_wallet_pks(params: &RollupParams) -> Vec<Buf32> {
...
+    let mut push_data = PushBytesBuf::new();
+    push_data
+        .extend_from_slice(data)
+        .expect("data should be within limit");
+
+    Builder::new()
+        .push_opcode(OP_RETURN)
+        .push_slice(push_data)
+        .into_script()
+}
```
delbonis, line 109: What is the purpose of this function?  Why is it in this crate?

crates/primitives/src/batch.rs
```diff
@@ -210,6 +219,29 @@ impl Checkpoint {
...
+
+impl CheckpointSidecar {
+    pub fn new(chainstate: Vec<u8>) -> Self {
+        Self { chainstate }
+    }
+
+    pub fn chainstate(&self) -> &[u8] {
+        &self.chainstate
+    }
 }
```
delbonis, line 245: I suggested the name `LazyChainstateBlob` in [this ticket](https://alpenlabs.atlassian.net/browse/STR-1055) that I shared earlier.

crates/primitives/src/errors.rs
```diff
@@ -28,4 +28,9 @@ pub enum ParseError {
     /// Converting from an unsupported [`Address`](bitcoin::Address) type for a [`Buf32`].
     #[error("only taproot addresses are supported but found {0:?}")]
     UnsupportedAddress(Option<AddressType>),
+
+    /// Could not get a network address from descriptor
+    /// Using String error as [`bitcoin_bosd::DescriptorError`] does not impl Clone
+    #[error("descriptor error: {0}")]
```
delbonis, line 34: Saying `error` in this error message is redundant.  Should just be `descriptor: {0}` or `bosd: {0}`.

crates/primitives/src/l1/btc.rs
```diff
@@ -871,6 +878,53 @@ impl<'a> arbitrary::Arbitrary<'a> for RawBitcoinTx {
...
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        // Generate arbitrary script
+        let script_len = usize::arbitrary(u)? % 100; // Limit script length
+        let script_bytes = u.bytes(script_len)?;
+        let script = ScriptBuf::from(script_bytes.to_vec());
+
+        Ok(Self(script))
+    }
+}
 #[cfg(test)]
```
delbonis, line 928: Formatting.

crates/l1tx/src/filter/types.rs
```diff
@@ -1,20 +1,85 @@
...
 #[derive(Clone, Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
 pub struct EnvelopeTags {
     pub checkpoint_tag: String,
     pub da_tag: String,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub struct ExpectedWithdrawalFulfillment {
+    /// withdrawal address scriptbuf
+    pub destination: BitcoinScriptBuf,
```
delbonis, line 31: Why can't this just be a `Vec<u8>` here instead of needing a wrapper for `ScriptBuf` and all of the bloat along with that?  Instances of this type are always well-formed since we generate them deterministically from our chainstate.  We're only using this for comparisons with `ScriptBuf`s, which we look at as plain byte slices anyways.

crates/l1tx/src/filter/types.rs
```diff
@@ -1,20 +1,85 @@
...
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.destination.cmp(&other.destination)
+    }
+}
+
+impl PartialOrd for ExpectedWithdrawalFulfillment {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
```
delbonis, line 51: This is not a compliant implementation of these traits.  When doing the binary search, instead use the `_by` or `_by_key` variants and *then* check the other fields.

crates/l1tx/src/filter/types.rs
```diff
@@ -1,20 +1,85 @@
...
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.output.cmp(&other.output)
+    }
+}
+
+impl PartialOrd for DepositSpendConfig {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
```
delbonis, line 81: Again this isn't a compliant implementation, see the prev comment.

crates/l1tx/src/filter/types.rs
```diff
@@ -1,20 +1,85 @@
...
+    /// index of deposit entry
+    pub deposit_idx: u32,
+    /// utxo for this deposit
+    pub output: OutputRef,
+}
+
+impl DepositSpendConfig {
+    pub fn from_outpoint(outpoint: OutPoint) -> Self {
+        Self {
+            deposit_idx: 0,
```
delbonis, line 64: When does this make sense to use like this?  If it's for a test then add `#[cfg(test)]` or just construct it with a `new` fn.

crates/l1tx/src/filter/types.rs
```diff
@@ -31,7 +96,10 @@ pub struct TxFilterConfig {
     pub expected_blobs: SortedVec<Buf32>,
 
     /// For deposits that might be spent from.
-    pub expected_outpoints: SortedVec<Outpoint>,
+    pub expected_outpoints: SortedVec<DepositSpendConfig>,
+
+    /// For withdrawal fulfillment transactions sent by bridge operator.
+    pub expected_withdrawal_fulfillments: SortedVec<ExpectedWithdrawalFulfillment>,
```
delbonis, line 102: Ah I think I see why those `Ord` impls were there.  These probably should not be `SortedVec`s then.  Or maybe we should define a `SortedMap`/`SortedLookupTable` type that's parametric over some trait impl like:

```rs
trait TableSchema {
    type Item;
    type Key: Eq + PartialEq + Ord + PartialOrd;
    fn get_key(i: &Self::Item) -> &Self::Key;
}
```

And then lets us do searches/lookups based on that `get_key` fn.

We could also theoretically redefine the `DepositsTable` based on this *in theory*, but I wouldn't do that here.

crates/primitives/src/sorted_vec.rs
```diff
@@ -41,12 +41,26 @@ impl<T: Ord + Clone> SortedVec<T> {
...
 
     /// Perform binary search on the vector
-    pub fn binary_search(&self, value: &T) -> Result<usize, usize> {
-        self.inner.binary_search(value)
+    pub fn binary_search(&self, value: &T) -> Option<&T> {
+        self.inner
+            .binary_search(value)
+            .ok()
+            .map(|idx| &self.inner[idx])
+    }
```
delbonis, line 53: As a sanity check, this fn doesn't *really* make sense.  It's just returning a ref to a different instance of the same type.  We'd *expect* this to be literally equivalent based on the relationships implied by the `Eq` and `Ord` impls, but we're not.  Instead (for this to be useful) we'd construct a different dummy instance of `T`.

Using `_by_key` is the interpretation here that really makes sense since it doesn't rely on this weird implication.

crates/primitives/src/sorted_vec.rs
```diff
@@ -41,12 +41,26 @@ impl<T: Ord + Clone> SortedVec<T> {
...
+
+    pub fn binary_search_by_key<F, B>(&self, b: &B, mut f: F) -> Option<&T>
+    where
+        F: FnMut(&T) -> &B,
+        B: Ord,
+    {
+        self.inner
+            .binary_search_by(|k| f(k).cmp(b))
+            .ok()
+            .map(|idx| &self.inner[idx])
```
delbonis, line 63: I am a little bit perplexed as to why this doesn't also call out to `.binary_search_by_key`.

crates/chaintsn/src/transition.rs
```diff
@@ -395,6 +424,11 @@ fn process_deposit_updates(
                 }
             }
 
+            DepositState::Executing(_) => {
+                // dont reassign executing withdrawals as front payment has been done.
+                // nothing else to do here for now
+            }
```
delbonis, line 430: Should probably be `Fulfilled` since it's more clear what that means.

crates/state/src/bridge_state.rs
```diff
@@ -465,4 +476,32 @@ impl WithdrawOutput {
...
+
+#[derive(Clone, Debug, Eq, PartialEq, BorshDeserialize, BorshSerialize, Serialize, Deserialize)]
+pub struct ExecutionState {
+    /// The index of the operator that has fronted the funds for the withdrawal,
+    /// and who will be reimbursed by the bridge notaries.
+    assignee: OperatorIdx,
+
+    /// Actual amount sent in withdrawal
+    amt: BitcoinAmount,
+}
```
delbonis, line 493: Does it maybe make sense for this to include the fulfillment txid?

Also as in the other comment, this would be `FulfilledState` which implies that it's waiting for reclamation.

crates/primitives/src/l1/btc.rs
```diff
@@ -871,6 +878,53 @@ impl<'a> arbitrary::Arbitrary<'a> for RawBitcoinTx {
...
+        BorshSerialize::serialize(&(script_bytes.len() as u64), writer)?;
+        writer.write_all(&script_bytes)?;
+        Ok(())
+    }
+}
+
+// Implement BorshDeserialize for BitcoinScriptBuf
+impl BorshDeserialize for BitcoinScriptBuf {
+    fn deserialize_reader<R: Read>(reader: &mut R) -> std::io::Result<Self> {
+        let script_len = u64::deserialize_reader(reader)? as usize;
```
delbonis, line 909: Maybe should be u32 since we would never normally have a scriptbuf that large.

crates/l1tx/src/filter/withdrawal_fulfillment.rs
```diff
@@ -0,0 +1,254 @@
...
+    let frontpayment_txout = tx.output.first()?;
+    let metadata_txout = tx.output.get(1)?;
+    if !metadata_txout.script_pubkey.is_op_return() {
+        return None;
+    }
+
+    // 2. Check withdrawal is to an address we expect
+    let withdrawal = filter_conf
+        .expected_withdrawal_fulfillments
+        .binary_search_by_key(&frontpayment_txout.script_pubkey, |expected| {
```
delbonis, line 31: The way this currently works would break if there's two withdrawals with the same dest spk.  We have to include the deposit idx in the op_return payload anyways, so we should be doing the lookup based on that.

crates/l1tx/src/filter/withdrawal_fulfillment.rs
```diff
@@ -0,0 +1,254 @@
...
+            expected.destination.inner()
+        })?;
+
+    // 3. Ensure amount is equal to the expected amount
+    let actual_amount_sats = frontpayment_txout.value.to_sat();
+    if actual_amount_sats < withdrawal.amount {
```
delbonis, line 37: This is actually a somewhat thought-provoking condition.  What if it's *more* than the requested amount?

crates/l1tx/src/filter/withdrawal_fulfillment.rs
```diff
@@ -0,0 +1,254 @@
...
+        return None;
+    }
+
+    // 4. Ensure it has correct metadata of the assigned operator.
+    let expected_metadata_script =
+        create_opreturn_metadata(withdrawal.operator_idx, withdrawal.deposit_idx);
+    if metadata_txout.script_pubkey != expected_metadata_script {
+        return None;
+    }
```
delbonis, line 46: This check wouldn't happen like this.  We should be parsing it from the tx upfront and then our other checks are dependent on what we get out of this.  It also means we fail fast if we encounter an irrelevant tx with an op_return in outidx 1 without having to do the slower path of looking it up in the table.

crates/db/src/types.rs
```diff
@@ -215,8 +216,14 @@ impl CheckpointEntry {
...
-    pub fn new_pending_proof(info: BatchInfo, transition: (Buf32, Buf32)) -> Self {
-        let checkpoint = Checkpoint::new(info, transition, Proof::default());
+    pub fn new_pending_proof(
+        info: BatchInfo,
+        transition: (Buf32, Buf32),
+        chainstate: &Chainstate,
+    ) -> Self {
+        let sidecar =
+            CheckpointSidecar::new(borsh::to_vec(chainstate).expect("serialize chainstate"));
+        let checkpoint = Checkpoint::new(info, transition, Proof::default(), sidecar);
```
delbonis, line 226: Can we supply the chainstate on the fly when we're constructing the final payload to publish instead of including it in this database?  Since it's not involved in the proving process.

crates/l1tx/src/filter/mod.rs
```diff
@@ -79,6 +81,21 @@ fn parse_checkpoint_envelopes<'a>(
...
 
+/// Parse transaction and filter out any deposits that have been spent.
+fn parse_deposit_spends<'a>(
+    tx: &'a Transaction,
+    filter_conf: &'a TxFilterConfig,
+) -> impl Iterator<Item = DepositSpendInfo> + 'a {
+    tx.input.iter().filter_map(|txin| {
+        filter_conf
+            .expected_outpoints
+            .binary_search_by_key(&txin.previous_output, |config| config.output.outpoint())
```
delbonis, line 92: Leaving a note here on this since I would expect that it'd be reworked again as well.

delbonis (changes requested):
On the issue of doing the lookups in the filter config, it makes more sense to instead expose these checks as higher level fns on the filter config type itself instead of having the caller access the field directly.  As usual, exposing implementation details is messy, and in this case we'd be expecting the caller to understand the internal invariants/expectations.

So have fns like `fn get_expected_fulfillment_info(&self, idx: u32) -> Option<&ExpectedFulfillmentInfo>` and such.

crates/l1tx/src/utils.rs
sapinb: The metadata in OP_RETURN needs to be compared in one transaction. Instead of parsing the ScriptBuf, its easier to just build the expected one and compare it. This is a helper I copied over from strata-bridge that builds the OP_RETURN.

crates/l1tx/src/filter/types.rs
sapinb: The SortedVec<> being used here requires the items to be Ord at the top level. Unless not use this and just normal Vec<> with sort_by ?

crates/db/src/types.rs
sapinb: Not without separating 2 types for the checkpoint posted in da and one stored in db

crates/l1tx/src/filter/withdrawal_fulfillment.rs
bewakes: Can't we just simply consider it fulfilled if an operator pays more than needed?

crates/l1tx/src/filter/types.rs
```diff
@@ -62,11 +129,60 @@ impl TxFilterConfig {
...
+                            // TODO: This uses FIXED OPERATOR FEE for TN1
+                            amount: output.amt().to_sat().saturating_sub(OPERATOR_FEE.to_sat()),
+                            operator_idx: dispatched_state.assignee(),
+                            deposit_idx: deposit.idx(),
+                        }
+                    });
+                Some(expected)
+            }
+            _ => None,
+        })
```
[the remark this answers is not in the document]
bewakes: The comment on the `DispatchCommand` says 
> At the moment, the underlying `Vec` of [`WithdrawOutput`] always has a single element.

So we can just assume there will be a single element in the outputs for now.

crates/primitives/src/batch.rs
bewakes: Why `Lazy` though? Also, something like `CheckpointSidecar` or `CheckpointExtra` indicates that we are not restricted to posting chainstate only. This seems descriptive enough to me.

crates/l1tx/src/filter/withdrawal_fulfillment.rs
bewakes: We have a check for the deposit idx and operator idx just after that, so this should be fine and just serves as redundant check.

crates/l1tx/src/filter/types.rs
bewakes: I've now added a new trait 
```rust 
pub trait HasKey<K: Ord> {
   fn get_key(&self) -> K 
}
```

And a new struct `SortedVecWithKey<K: Ord, T: HasKey<K>>` which doesn't require `PartialOrd` to be implemented on complex types

crates/primitives/src/batch.rs
delbonis: To emphasize that this is *not* how we intend to do it long term.  It'll look a lot different.

crates/l1tx/src/filter/types.rs
```diff
@@ -62,11 +129,60 @@ impl TxFilterConfig {
...
+        self.expected_outpoints = chainstate
+            .deposits_table()
+            .deposits()
+            .map(|deposit| DepositSpendConfig {
+                deposit_idx: deposit.idx(),
+                output: *deposit.output(),
+            })
+            .collect::<Vec<_>>()
+            .into();
+    }
```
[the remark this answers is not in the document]
delbonis: I understand this intention, but then we should shift to storing the shared key in the chainstate then, constructed at genesis.  This aligns with the idea that was discussed previously of caching the musig keys in the chainstate, and keeps it being a pure function.

crates/l1tx/src/filter/types.rs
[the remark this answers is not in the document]
delbonis: This is still incorrect behavior.  That's an implementation detail that you're depending on, creating an implicit coupling between this code which consumes the commands and the code that constructs them.  If it's impossible to make work then we should instead add an assertion that checks that there's only one.

crates/l1tx/src/utils.rs
delbonis: We can directly compare the script bytes as slices (with `.as_bytes()`) instead of forcing allocations here.

But in any case, why does is it called `op_return_nonce`?  Why "nonce"?  At least it should be a verb.

crates/state/src/bridge_state.rs
```diff
@@ -347,6 +347,14 @@ impl DepositEntry {
     }
 }
 
+#[cfg(feature = "test_utils")]
+impl DepositEntry {
+    pub fn with_state(mut self, state: DepositState) -> Self {
+        self.state = state;
+        self
+    }
+}
```
delbonis, line 356: Why not just `.set_state`?

crates/state/src/client_state.rs
```diff
@@ -348,14 +348,16 @@ impl InternalState {
...
 pub struct CheckpointL1Ref {
     pub block_height: u64,
+    pub blockid: L1BlockId,
     pub txid: Buf32,
     pub wtxid: Buf32,
 }
 
 impl CheckpointL1Ref {
-    pub fn new(block_height: u64, txid: Buf32, wtxid: Buf32) -> Self {
+    pub fn new(block_height: u64, blockid: L1BlockId, txid: Buf32, wtxid: Buf32) -> Self {
```
delbonis, line 357: Use `L1BlockCommitment` if possible.  This is what it's for.

crates/state/src/state_op.rs
```diff
@@ -289,45 +292,41 @@ impl StateCache {
...
+            .get_deposit_mut(deposit_idx)
+            .expect("stateop: missing deposit idx");
+
+        if !matches!(deposit_ent.deposit_state(), DepositState::Fulfilled(_)) {
+            // TODO: handle this better after TN1 bridge is integrated
+            warn!("stateop: deposit spent at unexpected state");
+        }
+
+        deposit_ent.set_state(DepositState::Reimbursed);
+    }
```
delbonis, line 329: We're shifting towards simplifying this interface.  These checks can be done outside `StateCache`.  We can just have a `.set_deposit_state` fn.

crates/storage/src/managers/chainstate.rs
```diff
@@ -5,19 +5,26 @@ use std::sync::Arc;
...
     pub fn new<D: Database + Sync + Send + 'static>(pool: ThreadPool, db: Arc<D>) -> Self {
         let ops = ops::chainstate::Context::new(db.chain_state_db().clone()).into_ops(pool);
         let wb_cache = cache::CacheTable::new(64.try_into().unwrap());
-        Self { ops, wb_cache }
+        let (bcast_tx, _) = broadcast::channel(1024);
+        Self {
+            ops,
+            wb_cache,
+            bcast_tx,
+        }
```
delbonis, line 27: This isn't really *not* the place for this.  We should really be using a higher level construct for this.  This means we're using something fairly low in the stack to signal logic much higher in the stack.

crates/btcio/src/reader/utils.rs
```diff
@@ -23,46 +22,40 @@ pub(crate) fn find_checkpoint(entries: &[RelevantTxEntry]) -> Option<&SignedChec
...
     }
-    debug!(
-        ?l2blk_comm,
-        "Found checkpoint in L1 block, waiting for corresponding processed terminal L2 block"
-    );
 
-    let mut rx = l2mgr.subscribe_to_valid_block_updates();
+    debug!(?epoch, "epoch chainstate not found, waiting");
+
+    let mut rx = chainstate_manager.subscribe_chainstate_updates();
```
delbonis, line 41: This should be pulling from the status channel.

Does it really make sense for this to even pull chainstate updates or just the CSM updates and do lookups into the chainstate from that?  We only update the filter configuration in response to new L1 checkpoints being included.

crates/btcio/src/reader/utils.rs
```diff
@@ -23,46 +22,40 @@ pub(crate) fn find_checkpoint(entries: &[RelevantTxEntry]) -> Option<&SignedChec
...
-            (h, blkid) => {
-                debug!(%h, %blkid, "Received valid block update");
+                let chainstate = chainstate_manager
+                    .get_toplevel_chainstate_async(slot)
+                    .await?
+                    .expect("chainstate should be found in db");
+
+                return Ok(chainstate);
             }
+            _ => {}
```
delbonis, line 58: Just use an `if`, this `match` obscures what's happening.

crates/l1tx/src/filter/checkpoint.rs
```diff
@@ -0,0 +1,162 @@
...
+        borsh::from_slice::<Chainstate>(signed_checkpoint.checkpoint().sidecar().chainstate())
+    {
+        warn!(?err, "invalid chainstate in checkpoint");
+        eprintln!("invalid chainstate in checkpoint: {}", err);
+
+        return None;
+    }
+
+    Some(signed_checkpoint)
+}
```
delbonis, line 74: A concept in PL theory (which borrows from linguistics) is the distinction between "syntax" and "semantics".  Syntax refers to the structure of the way a language is written.  The semantics is how that written language is used to convey meaning.

"Parsing" describes looking at a sequence of tokens (or phonemes, or bytes) and extracting the structure.  "Semantics" involves more advanced concepts like the type system.  It would be nice if we could have a data structure to describe a written program where it was impossible to construct instances that weren't well-typed.  But this would involve *incredibly* complicated data structure and parsing logic because there's multiple layers of interpretation that would have to be checked in a single step.

This is an example of where we're running into an issue where because we're trying to do two things at once, and the name `parse_checkpoint` hints at this.  This function is trying to do two things: (1) parse the checkpoint structure from the data, and (2) verify the checkpoint.  And the issue we're running into is that we need to have conditional compilation to turn off the checkpoint verification when we're in testing and won't be constructing *semantically* valid checkpoints (when we should still be ensuring that they're *syntactically* valid).  This pattern should be a huge red flag!  Move verifying the checkpoint out towards where have more context for if we want to properly verify it or not.  Perhaps we could have an additional flag in `TxFilterConfig` for if we're in a test env and would want to skip that check.  Do we know for sure that we actually want to verify these checkpoints here?

Also,

* if for some reason the conditional complication *was* the right way to do it, you can have `#[cfg(not(feature = "test_utils"))]` on specifically the `if` block verifying the signature, so there only has to be one toplevel fn def
* comments go before attributes
* don't forget to remove the `eprintln!`s

delbonis (changes requested):
I spent some more time going through this since I had more time and identified a few more subtle issues.  I understand this is a pressing issue but these subtleties are very important so we want to get it right.

crates/primitives/src/batch.rs
bewakes: To be honest, naming is hard, and very very subjective. For me `Lazy` by _no means_ implies that "we don't intend to do this long term". For me `Lazy` means something that is deferred until it is actually needed.

Let's not please debate over petty names.

crates/primitives/src/sorted_vec.rs
```diff
@@ -41,14 +135,25 @@ impl<T: Ord + Clone> SortedVec<T> {
...
+    /// Perform binary search by key on the vector
+    pub fn binary_search_by_key<B, F>(&self, value: &B, f: F) -> Result<usize, usize>
+    where
+        F: FnMut(&T) -> B,
+        B: Ord,
+    {
+        self.inner.binary_search_by_key(value, f)
+    }
     /// Returns the number of elements in the [`SortedVec`].
     pub fn len(&self) -> usize {
```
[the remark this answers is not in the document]
bewakes: This was not originally meant to be used like a `HashMap`, instead like a vector with sortedness enforced by the type. That is why the interfaces are similar to a `Vec`. If we intend to use something like a `HashMap` then the `Table` you mentioned in the other comment makes more sense. Here in this `SortedVec*`, `.get()` should just take in `usize` instead of key.

crates/storage/src/managers/chainstate.rs
bewakes: I don't think this is true. The hierarchy is  "Database" > "Ops" > "Manager", ignoring rocksdb and rockbound.

Everything higher on the stack interacts with the `Manager`. So, I too think that the manager is very much qualified to notify the interacting components about the database changes, that also whenever one subscribes.

crates/btcio/src/reader/utils.rs
bewakes: Oh yes, that must have been stuck there with the refactor.

crates/l1tx/src/filter/checkpoint.rs
bewakes: I think it makes sense to omit the sig verification here since we are verifying the checkpoint fully in chain transition anyways.

crates/state/src/bridge_state.rs
bewakes: Possibly to have a nicer chaining.

crates/l1tx/src/filter/checkpoint.rs
bewakes: We actually need it here to ensure that what we've written in the database is something that can be trusted.

>  It would be nice if we could have a data structure to describe a written program where it was impossible to construct instances that weren't well-typed.

Totally agree on that. We should strive for this as we move along.

>  Perhaps we could have an additional flag in TxFilterConfig for if we're in a test env and would want to skip that check

I wouldn't do it this way because this will bubble out to the client config or env vars and we shouldn't have unit test specific flags exposed that way. Such things should be done via feature flags.

crates/storage/src/managers/chainstate.rs
delbonis: It's a layering violation.  The source of truth that writes new chainstates (the FCM) is what those systems should really be listening to.  Components generally shouldn't be sending signals *up* the stack, especially when they go into different verticals.

crates/l1tx/src/filter/checkpoint.rs
delbonis: So another reason is this is bad is because feature flags are supposed to be strictly *additive* in their behavior.  This is hard to define, but it means that enabling a feature flag shouldn't be *disabling* behavior like this.

However, after going through it with Sapin, it seems this works fine with this flagging removed.  The `verify_signed_checkpoint_sig` is smart enough to always return true in the `Unchecked` case, which is what we're using in tests anyways.  So we removed it and this works out.

crates/btcio/src/reader/utils.rs
delbonis: This was resolved.

crates/l1tx/src/filter/types.rs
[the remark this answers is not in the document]
delbonis: This will be resolved later.

crates/l1tx/src/filter/types.rs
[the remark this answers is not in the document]
delbonis: Fixed the incorrect behavior.

crates/l1tx/src/utils.rs
delbonis: Reworked to no longer need these checks.

crates/primitives/src/sorted_vec.rs
```diff
@@ -1,37 +1,131 @@
...
+        self.clone()
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
-pub struct SortedVec<T> {
+pub struct SortedVecWithKey<K: Ord, T: HasKey<K>> {
     inner: Vec<T>,
+    _phantom: PhantomData<K>,
 }
```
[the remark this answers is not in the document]
delbonis: Fixed.

crates/primitives/src/sorted_vec.rs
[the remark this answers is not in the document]
delbonis: Fixed.

crates/state/src/state_op.rs
delbonis: Moving on, doesn't matter.

crates/primitives/src/sorted_vec.rs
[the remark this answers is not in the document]
delbonis: Implemented.

crates/l1tx/src/filter/types.rs
delbonis: Not important, moved on.

crates/primitives/src/sorted_vec.rs
delbonis: Fixed.

delbonis:
Okay we have everything looking good to go.  That one test failing is expected.  Force merging to unblock.

PR #703, opened 2025-03-03
Str 1111 fix node sync

bewakes:
## Description

This PR does the following:
1. Adds a debug util that allows a worker task to pause by calling an RPC. The module is `strata_common::worker_pause_manager`.
2. Adds a functional test that tests correct operation if a fullnode's l2 sync is lagging behind its l1 sync.
3. In L1 reader, when a checkpoint is received, it checks if the l2 block referred by the checkpoint is actually present in L2 database, waits otherwise. In doing so, moves `engine.finalize_block` call from csm worker to fcm worker as it makes more sense for EE finalization to be in FCM
4. Adds `safe_block_hash` and `finalized_block_hash` to ExecEngine's state case, because if not there's a chance that stale `finalized_block_hash` is passed during `forkChoiceUpdatedV2` call. 

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have commented my code where necessary.
- [ ] I have updated the documentation if needed.
- [ ] My changes do not introduce new warnings.
- [ ] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [ ] New and existing tests pass with my changes.

PR #715, opened 2025-03-06
Str 1111 fullnode sync finalization fix

sapinb:
## Description
Implementation of #706 invalidated the fix done in #703 regarding fullnode sync and finalization of L2 chain when L1 chain sync is multiple epochs ahead of L2 chain sync. 
This PR fixes the issue by only applying finalization on epochs which are known to be synced on both L1 and L2 chains.

- [X] finalize latest common epoch
- [X] init FCM state correctly
- [x] init L2 Sync worker state correctly

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

depends on checkpoints being correctly saved, fix made in #718

### DIFF
--- a/crates/consensus-logic/src/fork_choice_manager.rs
+++ b/crates/consensus-logic/src/fork_choice_manager.rs
@@ -504,7 +504,7 @@ fn process_fc_message(
             let status = if ok {
                 // check if any pending blocks can be finalized
                 if let Err(err) = handle_epoch_finalization(fcm_state, engine) {
-                    error!(?err, "failed to finalize epoch");
+                    error!(%err, "failed to finalize epoch");
                 }
 
                 // Update status.
@@ -950,7 +950,7 @@ fn handle_new_client_state(
 
     let finalized_epoch_changed = match handle_epoch_finalization(fcm_state, engine) {
         Err(err) => {
-            error!(?err, "failed to finalize epoch");
+            error!(%err, "failed to finalize epoch");
             false
         }
         Ok(Some(finalized_epoch)) if finalized_epoch == new_fin_epoch => {
--- a/crates/consensus-logic/src/fork_choice_manager.rs
+++ b/crates/consensus-logic/src/fork_choice_manager.rs
@@ -338,6 +338,8 @@ pub fn tracker_task<E: ExecEngineCtl>(
 
     handle_unprocessed_blocks(&mut fcm, &storage, engine.as_ref(), &status_channel)?;
 
+    status_channel.set_fcm_initialized();
+
     if let Err(e) = forkchoice_manager_task_inner(
         &shutdown,
         handle,
--- a/crates/consensus-logic/src/fork_choice_manager.rs
+++ b/crates/consensus-logic/src/fork_choice_manager.rs
@@ -338,8 +338,6 @@ pub fn tracker_task<E: ExecEngineCtl>(
 
     handle_unprocessed_blocks(&mut fcm, &storage, engine.as_ref(), &status_channel)?;
 
-    status_channel.set_fcm_initialized();
-
     if let Err(e) = forkchoice_manager_task_inner(
         &shutdown,
         handle,
@@ -430,7 +428,9 @@ fn forkchoice_manager_task_inner<E: ExecEngineCtl>(
             FcmEvent::NewFcmMsg(m) => {
                 process_fc_message(m, &mut fcm_state, engine, &status_channel)
             }
-            FcmEvent::NewStateUpdate(st) => handle_new_client_state(&mut fcm_state, st, engine),
+            FcmEvent::NewStateUpdate(st) => {
+                handle_new_client_state(&mut fcm_state, st, engine, &status_channel)
+            }
             FcmEvent::Abort => break,
         }?;
     }
@@ -502,6 +502,11 @@ fn process_fc_message(
             };
 
             let status = if ok {
+                // check if any pending blocks can be finalized
+                if let Err(err) = handle_epoch_finalization(fcm_state, engine) {
+                    error!(?err, "failed to finalize epoch");
+                }
+
                 // Update status.
                 let status = ChainSyncStatus {
                     tip: fcm_state.cur_best_block,
@@ -523,10 +528,6 @@ fn process_fc_message(
             };
 
             fcm_state.set_block_status(&blkid, status)?;
-            // check if any pending blocks can be finalized
-            if let Err(err) = handle_epoch_finalization(fcm_state, engine) {
-                error!(?err, "failed to finalize epoch");
-            }
         }
     }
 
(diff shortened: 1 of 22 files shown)


crates/consensus-logic/src/fork_choice_manager.rs
```diff
@@ -857,35 +935,89 @@ fn handle_new_client_state(
...
-        return Ok(());
+    info!(?new_fin_epoch, "got new finalized block");
+    fcm_state.attach_epoch_pending_finalization(new_fin_epoch);
+
+    // Update the new state.
+    fcm_state.cur_csm_state = Arc::new(cs);
+
+    let finalized_epoch_changed = match handle_epoch_finalization(fcm_state, engine) {
+        Err(err) => {
+            error!(?err, "failed to finalize epoch");
```
delbonis, line 953: Should use % for this.

crates/consensus-logic/src/fork_choice_manager.rs
```diff
@@ -857,35 +935,89 @@ fn handle_new_client_state(
...
+            false
+        }
+        Ok(Some(finalized_epoch)) if finalized_epoch == new_fin_epoch => {
+            debug!(?finalized_epoch, "finalized latest epoch");
+            true
+        }
+        Ok(Some(finalized_epoch)) => {
+            debug!(?finalized_epoch, "finalized earlier epoch");
+            true
+        }
```
delbonis, line 963: Under what conditions would this happen?

crates/consensus-logic/src/fork_choice_manager.rs
```diff
@@ -857,35 +935,89 @@ fn handle_new_client_state(
...
+            tip: fcm_state.cur_best_block,
+            prev_epoch: *fcm_state.get_chainstate_prev_epoch(),
+            finalized_epoch: *fcm_state.chain_tracker.finalized_epoch(),
+            // FIXME this is a bit convoluted, could this be simpler?
+            safe_l1: fcm_state.cur_chainstate.l1_view().get_safe_block(),
+        };
+
+        let update = ChainSyncStatusUpdate::new(status, fcm_state.cur_chainstate.clone());
+        trace!("publishing new chainstate");
+        status_channel.update_chain_sync_status(update);
```
delbonis, line 983: This should be done outside of this call so we do it unconditionally regardless of the epoch advancing.  Otherwise wouldn't we'd emit two events for a single chain tip update (which would confuse things)?  Is there some reason we can't?

functional-tests/tests/sync/sync_fullnode_l2_lag_restart.py
```diff
@@ -0,0 +1,112 @@
...
+        wait_until_with_value(
+            lambda: (
+                fnrpc.strata_clientStatus()["tip_l1_block"]["height"],
+                seqrpc.strata_clientStatus()["tip_l1_block"]["height"],
+            ),
+            lambda v: v[0] >= checkpt_l1_blk_height,
+            error_with="Fullnode L1 sync did not catch upto buried checkpoint",
+            timeout=10,
+            debug=True,
+        )
```
delbonis, line 80: Why does this also make the seqrpc call if it only checks the fnrpc call?

crates/consensus-logic/src/fork_choice_manager.rs
```diff
@@ -429,6 +502,11 @@ fn process_fc_message(
             };
 
             let status = if ok {
+                // check if any pending blocks can be finalized
+                if let Err(err) = handle_epoch_finalization(fcm_state, engine) {
+                    error!(?err, "failed to finalize epoch");
+                }
```
delbonis, line 508: Also % for this too.

crates/consensus-logic/src/fork_choice_manager.rs
```diff
@@ -857,35 +935,89 @@ fn handle_new_client_state(
...
     }
 
+    Ok(())
+}
+
+/// Tries to finalize pending epochs
+fn handle_epoch_finalization(
+    fcm_state: &mut ForkChoiceManager,
+    engine: &impl ExecEngineCtl,
+) -> anyhow::Result<Option<EpochCommitment>> {
```
delbonis, line 993: By "finalize pending epochs" we mean "update database entries, update the execution engine, clean up orphan tips" right?  This doc could use some more elaboration since it assumes the reader knows what's involved here.

crates/consensus-logic/src/fork_choice_manager.rs
```diff
@@ -146,6 +150,43 @@ impl ForkChoiceManager {
...
+            return false;
+        }
+
+        self.epochs_pending_finalization.push_back(epoch);
+
+        true
+    }
+
+    fn find_latest_finalizable_epoch(&self) -> Option<(usize, &EpochCommitment)> {
+        // the latest epoch which we have processed and is safe to finalize
```
delbonis, line 178: How do we define "safe to finalize"?

delbonis (changes requested):
Nothing structurally wrong with this.  The main issues are about clarity of comments/logs.  There's also several logs that could be converted over to `%` printing instead of `?`.

The only maybe substantive question is if we really need a queue of epochs pending finalization or if we can just keep track of the last checkpointed one?  Then when we get the epoch terminal we could check to see if the checkpoint entry exists in the db (and is finalized) and then just immediately wrap it up.  That would reduce the in-memory bookkeeping and probably make it more crash tolerant.

functional-tests/tests/sync/sync_fullnode_l2_lag_restart.py
sapinb: Mainly for debugging. Its helpful to see where both tips are in the logs.

crates/consensus-logic/src/fork_choice_manager.rs
sapinb: this is mainly for debugging purposes. One of the epochs "pending finalization" but not the latest epoch added got finalized during this call to handle_epoch_finalization(). 
ie. the L1 chain being read 1+ epochs ahead of the L2 chain (during initial sync of a new fullnode) and the L2 chain is slowly catching up.

crates/consensus-logic/src/fork_choice_manager.rs
sapinb: added more detailed explanation in the docstring

crates/consensus-logic/src/fork_choice_manager.rs
sapinb: done.

crates/consensus-logic/src/fork_choice_manager.rs
sapinb: done.

crates/consensus-logic/src/fork_choice_manager.rs
sapinb: L2 sync worker needs notification on change of "finalized" epoch in FCM's chain tracker,to keep in sync.
  
The finalized_epoch in `ChainSyncStatus` is also updated independent of the chainstate updates, so doing this is needed if this info is to be kept in sync.
If there is other code that is under the strict requirement that chainsync updates cannot happen twice with the same chainstate (which I dont believe there is ?) This should be ok .

delbonis (approved):

delbonis:
Forced merged because rules are confused.

PR #718, opened 2025-03-06
Make fullnode return non-empty checkpoint reference in get_checkpoint_info RPC

bewakes:
## Description
Resolves: https://alpenlabs.atlassian.net/browse/STR-1110

This PR does the following:
1. When a checkpoint is seen in L1, if it is a full node, it creates a new `CheckpointEntry` with appropriate status. This lets fullnode serve checkpoint info correctly in the `getCheckpointInfo` RPC. For sequencer, this was always working.
2. Remove the epoch l2 headers length check while creating checkpoint. We don't have a better way to handle this now. It used to return error if the length exceeded some value(5000).

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/consensus-logic/src/csm/worker.rs
+++ b/crates/consensus-logic/src/csm/worker.rs
@@ -384,7 +384,7 @@ fn apply_action(
             let mut ckpt_entry = match ckpt_db.get_checkpoint_blocking(epoch)? {
                 Some(c) => c,
                 None => {
-                    warn!(%epoch, "missing checkpoint we wanted to set the state of, creating new entry");
+                    info!(%epoch, "creating new checkpoint entry since the database does not have one");
 
                     CheckpointEntry::new(
                         checkpoint,
--- a/crates/consensus-logic/src/csm/worker.rs
+++ b/crates/consensus-logic/src/csm/worker.rs
@@ -2,7 +2,7 @@
 
 use std::{sync::Arc, thread};
 
-use strata_db::types::CheckpointConfStatus;
+use strata_db::types::{CheckpointConfStatus, CheckpointEntry, CheckpointProvingStatus};
 use strata_eectl::engine::ExecEngineCtl;
 use strata_primitives::prelude::*;
 use strata_state::{
@@ -376,12 +376,22 @@ fn apply_action(
 
         // Update checkpoint entry in database to mark it as included in L1.
         SyncAction::UpdateCheckpointInclusion {
-            epoch,
+            checkpoint,
             l1_reference,
         } => {
-            let Some(mut ckpt_entry) = ckpt_db.get_checkpoint_blocking(epoch)? else {
-                warn!(%epoch, "missing checkpoint we wanted to set the state of, ignoring");
-                return Ok(());
+            let epoch = checkpoint.batch_info().epoch();
+
+            let mut ckpt_entry = match ckpt_db.get_checkpoint_blocking(epoch)? {
+                Some(c) => c,
+                None => {
+                    warn!(%epoch, "missing checkpoint we wanted to set the state of, creating new entry");
+
+                    CheckpointEntry::new(
+                        checkpoint,
+                        CheckpointProvingStatus::ProofReady,
+                        CheckpointConfStatus::Pending,
+                    )
+                }
             };
 
             ckpt_entry.confirmation_status = CheckpointConfStatus::Confirmed(l1_reference);
--- a/crates/sequencer/src/checkpoint/worker.rs
+++ b/crates/sequencer/src/checkpoint/worker.rs
@@ -320,9 +320,10 @@ fn fetch_epoch_l2_headers(
     //
     // The break conditions are a little weird so we use a bare `loop`.
     loop {
-        if headers.len() >= limit {
-            return Err(Error::MalformedEpoch(summary.get_epoch_commitment()).into());
-        }
+        // TODO: we need some way to limit L2 blocks in an epoch, we can't just error like this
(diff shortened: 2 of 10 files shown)


crates/sequencer/src/checkpoint/worker.rs
```diff
@@ -320,9 +320,10 @@ fn fetch_epoch_l2_headers(
     //
     // The break conditions are a little weird so we use a bare `loop`.
     loop {
-        if headers.len() >= limit {
-            return Err(Error::MalformedEpoch(summary.get_epoch_commitment()).into());
-        }
+        // TODO: we need some way to limit L2 blocks in an epoch, we can't just error like this
+        // if headers.len() >= limit {
```
sapinb, line 324: can we also remove the unused limit variable ?
`#[allow(unused)]` 😞

functional-tests/tests/sync/sync_from_rpc.py
```diff
@@ -62,3 +63,14 @@ def main(self, ctx: flexitest.RunContext):
...
         )
         assert seq_el_hash == fn_el_hash, "EL blocks don't match"
+
+        # Check fullnode sees same checkpoint reference as sequencer
+        wait_until(
+            lambda: seqrpc.strata_getCheckpointInfo(0)["confirmation_status"] != "pending",
+            error_with="Did not see checkpoint in l1",
+        )
+        fn_checkpt_info = fnrpc.strata_getCheckpointInfo(0)
+        sq_checkpt_info = seqrpc.strata_getCheckpointInfo(0)
```
sapinb, line 73: can we wait and check till checkpoint 1 or 2 is synced correctly? Just to cover more ground with this test.

sapinb (dismissed):

sapinb:
can we disable this failing fn test `sync_fullnode_l2_lag` ? I'll enable it when its fixed in #715

sapinb (approved):

functional-tests/tests/sync/sync_from_rpc.py
```diff
@@ -62,3 +63,16 @@ def main(self, ctx: flexitest.RunContext):
...
         )
         assert seq_el_hash == fn_el_hash, "EL blocks don't match"
+
+        # Check fullnode sees same checkpoint reference as sequencer
+        epoch = 1
+        wait_until(
+            lambda: seqrpc.strata_getCheckpointInfo(epoch)["confirmation_status"] != "pending",
+            error_with="Did not see checkpoint in l1",
+            timeout=15,
+        )
```
delbonis, line 73: Should have a dedicated utils fn for this.  This is rather ad-hoc.  See the others I added recently.

crates/consensus-logic/src/csm/worker.rs
```diff
@@ -376,12 +376,22 @@ fn apply_action(
...
         } => {
-            let Some(mut ckpt_entry) = ckpt_db.get_checkpoint_blocking(epoch)? else {
-                warn!(%epoch, "missing checkpoint we wanted to set the state of, ignoring");
-                return Ok(());
+            let epoch = checkpoint.batch_info().epoch();
+
+            let mut ckpt_entry = match ckpt_db.get_checkpoint_blocking(epoch)? {
+                Some(c) => c,
+                None => {
+                    warn!(%epoch, "missing checkpoint we wanted to set the state of, creating new entry");
```
delbonis, line 387: I don't really think this is a warning to be honest.

Also, what should we do in the case of reorgs?  This won't be an issue here but maybe we should create a ticket if there isn't a simple and obviously correct way to deal with it.

crates/consensus-logic/src/csm/worker.rs
```diff
@@ -376,12 +376,22 @@ fn apply_action(
...
+
+                    CheckpointEntry::new(
+                        checkpoint,
+                        CheckpointProvingStatus::ProofReady,
+                        CheckpointConfStatus::Pending,
+                    )
+                }
             };
 
             ckpt_entry.confirmation_status = CheckpointConfStatus::Confirmed(l1_reference);
```
delbonis, line 397: When we're constructing a new one, can't we preemptively set it as `Confirmed` here?  I suppose it doesn't really matter.

delbonis (changes requested):
This is a good improvement to handle that oversight and ensure more correct behavior.

But I still kindaaa feel like the checkpoint entry types could be split out more because this table still feels like it's serving multiple roles.  I guess as a rough comparison, we don't have a specially-tagged version of a block entry that we store on disk while we're constructing it before it's broadcast.  I don't think this is actionable here yet but it's something to keep in mind.

functional-tests/tests/sync/sync_from_rpc.py
bewakes: Yes, used the one there's already there.

crates/consensus-logic/src/csm/worker.rs
bewakes: Yes, this isn't a warning anymore. Corrected.

In case of reorgs, we should probably clear the corresponding database checkpoint entries(and possibly other entries which might be related). Just reverting L1 manifests and corresponding tx entries won't be enough.

crates/consensus-logic/src/csm/worker.rs
bewakes: Yeah, it doesn't matter. I just didn't want to re-set the same status twice.

sapinb (approved):

PR #719, opened 2025-03-06
trigger for CD updated to use manual trigger only

purusang:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -4,10 +4,6 @@
 
 name: Continuous Deployment
 on:
-  push:
-    branches:
-      - main
-
   workflow_dispatch:
     inputs:
       environment:


storopoli (changes requested):
Why comment and not remove?

storopoli (approved):
ACK 6597357894b0345052158179537c5c5e8f67312b

maybe we should add a trigger that runs when we tag versions?

```yaml
on:
  push:
    tags:
      - v1.**
```

See: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs

purusang:
> ACK [6597357](https://github.com/alpenlabs/strata/commit/6597357894b0345052158179537c5c5e8f67312b)
> 
> maybe we should add a trigger that runs when we tag versions?
> 
> ```yaml
> on:
>   push:
>     tags:
>       - v1.**
> ```
> 
> See: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs

Yeah that's some thing good to have. But I'm still working on it actively in deployment repo's side. So might make some further changes and add this feature with other changes or when we are ready for releases.

PR #722, opened 2025-03-07
STR-521 BIP44 derivation path for Strata EVM address

krsnapaudel:
## Description

First draft.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/bin/strata-cli/src/constants.rs
+++ b/bin/strata-cli/src/constants.rs
@@ -1,7 +1,10 @@
 use std::time::Duration;
 
 use alloy::consensus::constants::ETH_TO_WEI;
-use bdk_wallet::bitcoin::{bip32::ChildNumber, Amount, Network};
+use bdk_wallet::bitcoin::{
+    bip32::{ChildNumber, DerivationPath},
+    Amount, Network,
+};
 
 /// Number of blocks after bridge in transaction confirmation that the recovery path can be spent.
 pub const RECOVER_DELAY: u32 = 1008;
--- a/bin/strata-cli/src/constants.rs
+++ b/bin/strata-cli/src/constants.rs
@@ -1,7 +1,7 @@
 use std::time::Duration;
 
 use alloy::consensus::constants::ETH_TO_WEI;
-use bdk_wallet::bitcoin::{Amount, Network};
+use bdk_wallet::bitcoin::{bip32::ChildNumber, Amount, Network};
 
 /// Number of blocks after bridge in transaction confirmation that the recovery path can be spent.
 pub const RECOVER_DELAY: u32 = 1008;
@@ -38,27 +38,18 @@ pub const SIGNET_BLOCK_TIME: Duration = Duration::from_secs(30);
 pub const BRIDGE_MUSIG2_PUBKEY: &str =
     "14ced579c6a92533fa68ccc16da93b41073993cfc6cc982320645d8e9a63ee65";
 
-/// BIP44 purpose index for HD wallets.
+/// Strata CLI [`DerivationPath`] for Strata EVM wallet
 ///
-/// These should be _hardened_ [`ChildNumber`].
-pub const BIP44_HD_WALLET_IDX: u32 = 44;
-
-/// BIP44 coin type index to indicate Testnet.
-///
-/// These should be _hardened_ [`ChildNumber`].
-pub const BIP44_ETHEREUM_MAINNET_IDX: u32 = 60;
-
-/// BIP44 account index for user wallets.
-///
-/// These should be _hardened_ [`ChildNumber`].
-pub const BIP44_USER_ACCOUNT_IDX: u32 = 0;
-
-/// BIP44 change index for receiving (external) addresses.
-///
-/// These should be a normal [`ChildNumber`].
-pub const BIP44_RECEIVING_ADDRESS_IDX: u32 = 0;
-
-/// BIP44 address index.
-///
-/// These should be a normal [`ChildNumber`].
-pub const BIP44_ADDRESS_IDX: u32 = 0;
+/// This corresponds to the path: `m/44'/60'/0'/0/0`.
+pub const BIP44_STRATA_EVM_WALLET_PATH: &[ChildNumber] = &[
+    // Purpose index for HD wallets based on BIP44.
+    ChildNumber::Hardened { index: 44 },
+    // Coin type index for Ethereum mainnet
+    ChildNumber::Hardened { index: 60 },
+    // account index for user wallets.
+    ChildNumber::Hardened { index: 0 },
+    // Change index for receiving (external) addresses.
+    ChildNumber::Normal { index: 0 },
+    // Address index.
+    ChildNumber::Normal { index: 0 },
+];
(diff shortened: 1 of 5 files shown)


bin/strata-cli/src/seed.rs
```diff
@@ -92,14 +99,28 @@ impl Seed {
...
+            ChildNumber::from_hardened_idx(BIP44_TESTNET_IDX).expect("valid hardened child number"),
+            ChildNumber::from_hardened_idx(BIP44_USER_WALLET_IDX)
+                .expect("valid hardened child number"),
+            ChildNumber::from_normal_idx(BIP44_RECEIVING_ADDRESS_IDX)
+                .expect("valid normal child number"),
+            ChildNumber::from_str("*").expect("valid normal child number"),
+        ]);
+
+        // Create the master key from the existing seed
+        let master_key = Xpriv::new_master(Network::Testnet, self.0.as_ref()).unwrap();
```
storopoli, line 117: So the path is `m/44'/1'/0'/0'/0` for the main account?

storopoli (commented):
As we discussed offline, EVM uses an account-based blockchain. Hence we cannot do `*` like we did in paths for the Bitcoin wallet.
Every new index from the "base path" is a new account, where `0` is the primary account and anything higher than that are alternative accounts from the same master key.

It's fine to keep as `0` for now, but eventually we want to expose that to be parametrizable to give support for user's alternative accounts from the same master key.

delbonis (changes requested):

bin/strata-cli/src/seed.rs
storopoli: Shouldn't this be `m/44'/60'/0'/0/0`?

krsnapaudel: @storopoli : Yes, based on the updated ticket description

bin/strata-cli/src/constants.rs
```diff
@@ -37,3 +37,28 @@ pub const SIGNET_BLOCK_TIME: Duration = Duration::from_secs(30);
 
 pub const BRIDGE_MUSIG2_PUBKEY: &str =
     "14ced579c6a92533fa68ccc16da93b41073993cfc6cc982320645d8e9a63ee65";
+
+/// BIP44 purpose index for HD wallets.
+///
+/// These should be _hardened_ [`ChildNumber`].
+pub const BIP44_HD_WALLET_IDX: u32 = 44;
+
+/// BIP44 coin type index to indicate Testnet.
```
barakshani, line 46:
```suggestions
/// BIP44 coin type index to indicate EVM Mainnet.
```
60 actually indicates mainnet, which we also use for the testnet (as others also do, and we want to remain compatible on the wallet level).

storopoli (dismissed):
ACK 96568c17f21fb9ead30200aadb25d373ac66c18a

But don't merge yet, @delbonis might have some reservation on the new `const` naming.

storopoli:
@Zk2u can you check this PR? The context is in STR-521. Asking you since you were the original implementor of the EVM Wallet in `strata-cli` and there might be something that you might catch here.

storopoli:
> This should be marked as a breaking change and a new version of the CLI as it's not compatible with the previous wallet

@krsnapaudel can you bump the `strata-cli`'s `Cargo.toml` to `0.2.0`?

bin/strata-cli/src/seed.rs
```diff
@@ -91,15 +98,30 @@ impl Seed {
...
+                .expect("valid hardened child number"),
+            ChildNumber::from_hardened_idx(BIP44_USER_ACCOUNT_IDX)
+                .expect("valid hardened child number"),
+            ChildNumber::from_normal_idx(BIP44_RECEIVING_ADDRESS_IDX)
+                .expect("valid normal child number"),
+            ChildNumber::from_normal_idx(BIP44_ADDRESS_IDX).expect("valid normal child number"),
+        ]);
+
+        // Create the master key from the existing seed
+        let master_key = Xpriv::new_master(Network::Testnet, self.0.as_ref()).unwrap();
```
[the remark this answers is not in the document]
krsnapaudel: pub const DEFAULT_NETWORK: Network = Network::Signet;
This probably is not right for StrataWallet.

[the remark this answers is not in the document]
storopoli: Yes, in general this is "not right for EVM `StrataWallet`".
But BIP-32 is a `bitcoin` code:

https://github.com/alpenlabs/strata/blob/a9b458548094b9b74f5827502dc0a3fa845be6fe/bin/strata-cli/src/seed.rs#L8

It can correct parse into a valid extended private key for a EVM-based wallet.

HOWEVER, the network choice changes the extended private key serialized data, as explained in [BIP-32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#serialization-format): 
image

Given that the spec for [MetaMask](https://support.metamask.io/configure/wallet/importing-a-seed-phrase-from-another-wallet-software-derivation-path/) always hardcode mainnet (while [Frame](https://github.com/floating/frame/blob/dac4378979fe1f490f4d0bf141dc19c201d2cb58/main/signers/Signer/derive.ts#L33-L38) does not), and given the popularity of MetaMask as an EVM wallet; we should, IMO, for EVM-based wallets always hardcode `Network::Bitcoin` (mainnet).
If we do this, there should be a `// NOTE: Bla Bla Blah` comment explaining why we hardcoded `Network::Bitcoin` right on top of `Network::Bitcoin`.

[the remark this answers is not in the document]
storopoli: @Zk2u agreed, but that's what happens when you not only shove a bitcoin standard into EVM, but also depends on some external dep (MetaMask) to also have the same point of view and implementation as you...

bin/strata-cli/src/seed.rs
```diff
@@ -100,8 +100,11 @@ impl Seed {
     pub fn get_strata_wallet(&self) -> EthereumWallet {
         let derivation_path = DerivationPath::master().extend(BIP44_STRATA_EVM_WALLET_PATH);
 
-        // Create the master key from the existing seed
-        let master_key = Xpriv::new_master(Network::Testnet, self.0.as_ref()).expect("valid xpriv");
+        // Network choice affects how extended public and private keys are serialized. See
+        // https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#serialization-format.
+        // Given the popularity of MetaMask, we their example (they always hardcode mainnet) and
+        // hardcode Network::Bitcoin (mainnet) for EVM-based wallet.
```
storopoli, line 106:
```suggestion
        // Given the popularity of MetaMask, we follow their example (they always hardcode mainnet) and
        // hardcode Network::Bitcoin (mainnet) for EVM-based wallet.
```
Maybe the action was missing?

storopoli (dismissed):
ACK 317c71bbb856378a0b2a75d99da0b58c19b99bb2

delbonis (approved):
Yeah I'm really sad that this is how the keyderiv works but I suppose it's unavoidable if we want to avoid surprises with the most widely used wallet.

Also, do we expect all those CI jobs to be failing?

storopoli:
> Also, do we expect all those CI jobs to be failing?

No, @krsnapaudel can you please check what's going on?
Only the `cargo-audit` because of https://osv.dev/vulnerability/RUSTSEC-2024-0437 should be failing.

krsnapaudel:
@storopoli Apparently, accepting commit suggestion for GH can lead to formatting issues.

storopoli (approved):
ACK 026598ca44c24466c678a73f3ebed9cb14d94d0f

PR #725, opened 2025-03-10
`strata-cli`: fixes for testnet 1

storopoli:
## Description

According to the [specs parameterization](https://www.notion.so/Detailed-Strata-Bridge-Design-149901ba000f80a4b38adb5cfb522d94#149901ba000f80248d26e3175887bd64):

- Recovery delay is 144 blocks (1 day)
- Fees are 1k satoshi
- Finality depth (reorg) is 6 blocks

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This came up in https://github.com/alpenlabs/docs/pull/1#discussion_r1987772614

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [x] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [x] New and existing tests pass with my changes.

## Related Issues

None

PR #726, opened 2025-03-10
Update RECOVER_DELAY per Testnet I requirements

barakshani:
## Description

The timelock on the DRT for the user to take back their funds should be 144 blocks, as per the main requirements.
See also https://github.com/alpenlabs/docs/pull/1#discussion_r1987772614

**This should not be merged before we double check that in all other places we change this as well.**


storopoli:
@barakshani can we close this in favor of #725 (it does the same changes and also bumps the version to `0.2.0` from `0.1.0`)

PR #730, opened 2025-03-11
Fix/checkpoint verification in CL STF

prajwolrg:
### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [x] Dependency Update

## Checklist

### DIFF
--- a/crates/chaintsn/src/transition.rs
+++ b/crates/chaintsn/src/transition.rs
@@ -6,6 +6,7 @@ use std::{cmp::max, collections::HashMap};
 
 use bitcoin::{block::Header, consensus, params::Params, OutPoint, Transaction};
 use rand_core::{RngCore, SeedableRng};
+use strata_crypto::groth16_verifier::verify_rollup_groth16_proof_receipt;
 use strata_primitives::{
     batch::SignedCheckpoint,
     epoch::EpochCommitment,
@@ -16,6 +17,7 @@ use strata_primitives::{
     params::RollupParams,
 };
 use strata_state::{
+    batch::verify_signed_checkpoint_sig,
     block::L1Segment,
     bridge_ops::{DepositIntent, WithdrawalIntent},
     bridge_state::{DepositState, DispatchCommand, WithdrawOutput},
@@ -86,7 +88,7 @@ fn compute_init_slot_rng(state: &StateCache) -> SlotRng {
 
 /// Update our view of the L1 state, playing out downstream changes from that.
 ///
-/// Returns if there was an update processed.
+/// Returns true if there epoch needs to be updated.
 fn process_l1_view_update(
     state: &mut StateCache,
     l1seg: &L1Segment,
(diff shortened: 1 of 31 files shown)


crates/chaintsn/src/transition.rs
```diff
@@ -158,15 +178,28 @@ fn process_l1_checkpoint(
...
     let ckpt = signed_ckpt.checkpoint(); // inner data
 
+    let receipt = ckpt.construct_receipt();
+    if !params.proof_publish_mode.allow_empty() {
+        assert!(
+            verify_rollup_groth16_proof_receipt(&receipt, &params.rollup_vk).is_ok(),
+            "posted checkpoint proof verification failed"
+        )
+    }
+
```
bewakes, line 202: Should we actually assert here directly? I think it would be better to return Result or and Option. If we have multiple checkpoints, only one of them valid, then returning result/option would allow us to try to verify checkpoint until it passes. If none pass then we can probably panic or assert.

bewakes: I also think we should check if the proof is the expected proof i.e. proof we are expecting for the next epoch.

crates/chaintsn/src/transition.rs
```diff
@@ -86,51 +88,69 @@ fn compute_init_slot_rng(state: &StateCache) -> SlotRng {
...
+    // epoch
+    if prev_finalized_epoch.is_null() {
+        return Ok(true);
+    }
+
+    // For all other non-genesis batch, we need to check that the new finalized epoch has been
+    // updated when processing L1Checkpoint
+    let new_finalized_epoch = state.state().finalized_epoch();
 
+    if new_finalized_epoch.epoch() > prev_finalized_epoch.epoch() {
```
bewakes, line 136: the only concern with this check is that, this implies we can have l1 segment without checkpoint, which is not true in our case because we only create a l1 segment if we see a valid previous checkpoint on L1. In practice, this probably won't cause issues, but seems like this kind of violates the theoretical invariant.

bewakes (commented):
A couple of comments.

bewakes (approved):

MdTeach (approved):

sapinb (approved):

crates/chaintsn/src/transition.rs
```diff
@@ -158,14 +179,38 @@ fn process_l1_checkpoint(
...
+        // If the proof is empty but empty proofs are not allowed, this will fail.
+        if !params.proof_publish_mode.allow_empty() {
+            error!(%ckpt_epoch, "Invalid checkpoint: Empty proof");
+            return Err(TsnError::InvalidProof);
+        }
+    } else {
+        // Otherwise, verify the non-empty proof.
+        verify_rollup_groth16_proof_receipt(&receipt, &params.rollup_vk)
+            .map_err(|e| TsnError::InvalidProof)?;
+    }
```
delbonis, line 213: I just realized that this returning an error means if a sequencer publishes an invalid checkpoint then we can *never* produce a new block after we receive it (without rolling back L1), since this will always fail the state transition.  We should just disregard it (and print an error/warning).

PR #756, opened 2025-03-30
Reorder address validation for faucet (strata-cli)

pranjalpokharel7:
## Description

Validate address before solving PoW challenge in faucet command (strata-cli). 

**Issue**
Screenshot 2025-03-28 at 18 08 04

\
**Fixed** 
Screenshot 2025-03-30 at 10 54 46

**Updated Fixed Screenshot**
Screenshot 2025-03-31 at 08 44 02

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/bin/strata-cli/src/cmd/faucet.rs
+++ b/bin/strata-cli/src/cmd/faucet.rs
@@ -70,12 +70,31 @@ impl fmt::Display for Chain {
 }
 
 pub async fn faucet(args: FaucetArgs, seed: Seed, settings: Settings) {
-    println!("Fetching challenge from faucet");
-    let client = reqwest::Client::new();
-    let base = Url::from_str(&settings.faucet_endpoint).expect("valid url");
-
     #[cfg(feature = "strata_faucet")]
     let network_type = net_type_or_exit(&args.network_type);
+
+    #[cfg(feature = "strata_faucet")]
+    let (address, claim) = match network_type {
+        NetworkType::Signet => (
+            resolve_signet_address(&args, &seed, &settings).to_string(),
+            "claim_l1",
+        ),
+        NetworkType::Strata => (
+            resolve_strata_address(&args, &seed, &settings).to_string(),
+            "claim_l2",
+        ),
+    };
+
+    #[cfg(not(feature = "strata_faucet"))]
+    let (address, claim) = (
+        resolve_signet_address(&args, &seed, &settings).to_string(),
+        "claim_l1",
+    );
+
+    println!("Fetching challenge from faucet");
+
+    let client = reqwest::Client::new();
+    let base = Url::from_str(&settings.faucet_endpoint).expect("valid url");
 
     #[cfg(feature = "strata_faucet")]
     let endpoint = {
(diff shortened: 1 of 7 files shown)


storopoli (changes requested):
Is there a way to not panic?
Would be much better to ask the user to retry entering the address, right?

bin/strata-cli/src/cmd/faucet.rs
```diff
@@ -45,47 +45,8 @@ pub async fn faucet(args: FaucetArgs, seed: Seed, settings: Settings) {
     #[cfg(feature = "strata_faucet")]
```
barakshani, line 45: I'd assume the cli message 2 lines above should also be displayed only if we passed the address verification, as this challenge is probably the POW challenge we have in the message below.

Feel free to add an alternative message in the lines of "Making a faucet request" or "Requesting BTC from the faucet" so the user gets some acknowledgment.

bin/strata-cli/src/cmd/faucet.rs
```diff
@@ -146,15 +95,60 @@ pub async fn faucet(args: FaucetArgs, seed: Seed, settings: Settings) {
...
+
+    #[cfg(not(feature = "strata_faucet"))]
+    println!("Claiming to signet address {}", address);
+
+    let url = format!(
+        "{base}{}/{}/{}",
+        claim,
+        encode(&solution.to_le_bytes()),
+        address
+    );
```
barakshani, line 151: A minor comment, but I'd say it make more sense to have this before the proof-of-work code (currently, line 100) as this refers to a previous context.

bin/strata-cli/src/cmd/faucet.rs
pranjalpokharel7: Makes sense, will make the necessary changes.

krsnapaudel:
@pranjalpokharel7 : Your code will have conflicts with [my recent PR](https://github.com/alpenlabs/strata/pull/755). Let me know if you need help fixing conflicts.

delbonis (dismissed):

bin/strata-cli/src/cmd/faucet.rs
```diff
@@ -70,13 +70,76 @@ impl fmt::Display for Chain {
...
+                    address
+                        .require_network(settings.network)
+                        .expect("wrong bitcoin network")
+                }
+            };
+            (address.to_string(), "claim_l1")
+        }
+        NetworkType::Strata => {
+            let l2w = StrataWallet::new(&seed, &settings.strata_endpoint).unwrap();
+            // they said EVMs were advanced 👁️👁️
```
storopoli, line 102: This comment is not needed anymore it was a hot take on alloy's huge trait bounds on generics.

bin/strata-cli/src/cmd/faucet.rs
```diff
@@ -70,13 +70,76 @@ impl fmt::Display for Chain {
 }
 
 pub async fn faucet(args: FaucetArgs, seed: Seed, settings: Settings) {
+    #[cfg(feature = "strata_faucet")]
```
storopoli, line 73: Does it make sense to just `#[cfg(feature = "strata_faucet")]` the function `faucet` instead?

storopoli (changes requested):

storopoli (changes requested):
Check my previous review

bin/strata-cli/src/cmd/faucet.rs
pranjalpokharel7: Should we write two versions of the function in that case? Or will this feature simply be enabled for all standard builds? From the devnet docs, it's made clear that the faucet deposit will only work for signet, the CLI will not remain consistent in that case (feature is not enabled during devnet CLI build).

pranjalpokharel7: Did you mean something like this (two versions of the faucet function)? 
```rust
#[cfg(feature = "strata_faucet")]
pub async fn faucet(...) {}

#[cfg(not(feature = "strata_faucet"))]
pub async fn faucet(...) {}
```

storopoli: No, I thought that a bunch of `#[cfg(feature = "strata_faucet")]` inside the function was for everything in the whole function.
But no right?
Ok fine.

pranjalpokharel7: I've refactored the function so that it is more readable.

pranjalpokharel7:
@krsnapaudel is the faucet unusable for strata address as of now? I tried the command on the `main` branch, it's failing due to an unexpected EOF error. Should I look into this as well? 

Screenshot 2025-03-31 at 16 32 30

krsnapaudel:
@pranjalpokharel7 : You should use the config from this PR. Note, the PoW difficulty is too high to actually claim tokens.
https://github.com/alpenlabs/docs/pull/7/files

pranjalpokharel7:
@barakshani @krsnapaudel regarding ticket [STR-1205](https://alpenlabs.atlassian.net/browse/STR-1205)

I'm working on avoiding panics for incorrect password in the CLI, and I am opting for the first option to solve the issue. Let me know if I should solve this some other way. We can handle this,

1. (Selected) At the calling site (bin/strata-cli/src/main.rs), which will preserve the logic of original `load_or_create` function,
```rust
async fn main() {
// ...

    let seed = seed::load_or_create(&persister).unwrap_or_else(|e| {
        let narrowed = e.narrow::<IncorrectPassword, _>();
        match narrowed {
            Ok(_) => std::process::exit(1), // system exit, surrounding code is a bit clunky
            Err(_) => panic!("{:?}", narrowed.unwrap()),
        }
    });    
}
```


2. (Alternative) Or within the `load_or_create` function (`bin/strata-cli/src/seed.rs`), which simplifies things but changes the "library" code itself. If we are also not returning the IncorrectPassword error then we should consequently remove it from the codebase (unused elsewhere).
```rust
            Err(e) => {
                let narrowed = e.narrow::<aes_gcm_siv::Error, _>();
                if let Ok(_aes_error) = narrowed {
                    println!("Incorrect password");
                    std::process::exit(1); // exit here only, which means error type is not propagated back
                }

                Err(narrowed.unwrap_err().broaden())
            }
```

krsnapaudel:
@pranjalpokharel7 I think we should not panic. So propagate the error and make the error message as user-friendly as possible.

bin/strata-cli/src/cmd/faucet.rs
krsnapaudel: @pranjalpokharel7 : Based on discussions with @Zk2u and @bewakes, you could take out the feature `strata_faucet`. Hopefully this makes the change cleaner.

pranjalpokharel7: Removed the `strata_faucet` feature in this latest commit - [b7878b9](https://github.com/alpenlabs/strata/pull/756/commits/b7878b90ecb571b70ed9ed00625c0a54524b8a0f).

barakshani:
test: `STR-1205`

storopoli (approved):
ACK 4180e01ce82c46900cb5f0f1ab36fa8b87b7877a

BUT
Before merging can you show a screenshot of the new error messages and the wallet-cli not panicking?

pranjalpokharel7:
Invalid Address -
Screenshot 2025-04-02 at 14 31 22

Incorrect Password (agree that the error message itself isn't anything special, but that way we could avoid panics with minimum amount of changes) -
Screenshot 2025-04-02 at 14 32 50

PR #772, opened 2025-04-14
Error handling in strata CLI

krsnapaudel:
## Description

A number of errors in Strata CLI lead to panics. An attempt to prevent panics in almost all cases. Not part of a JIRA task.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/bin/strata-cli/src/errors.rs
+++ b/bin/strata-cli/src/errors.rs
@@ -1,0 +1,52 @@
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum CliError {
+    #[error("Invalid input: {0}")]
+    UserInput(#[from] UserInputError),
+
+    #[error("Internal error: {0}")]
+    Internal(#[source] anyhow::Error),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum UserInputError {
+    #[error("unsupported language. Use --help to see available options.")]
+    UnsupportedLanguage,
+
+    #[error("unsupported network. Use --help to see available options.")]
+    UnsupportedNetwork,
+
+    #[error("wrong network for address")]
+    WrongNetwork,
+
+    #[error("invalid signet address. Must be a valid Bitcoin address.")]
+    InvalidSignetAddress,
+
+    #[error("invalid Strata address. Must be a valid EVM address.")]
+    InvalidStrataAddress,
+
+    #[error("either signet or Strata address must be provided")]
+    MissingTargetAddress,
+}
+
+#[derive(Debug, Error)]
+pub enum WalletError {
+    #[error("wallet not initialized")]
+    NotInitialized,
+
+    #[error("failed to load wallet")]
+    LoadFailed,
+
+    #[error("failed to scan wallet")]
+    ScanFailed,
+
+    #[error("failed to sync wallet")]
+    SyncFailed,
+
+    #[error("failed to save wallet state")]
+    PersistFailed,
+
+    #[error("invalid wallet descriptor")]
+    InvalidWalletDescriptor,
+}
--- a/bin/strata-cli/src/net_type.rs
+++ b/bin/strata-cli/src/net_type.rs
@@ -1,5 +1,7 @@
 use std::{fmt, str::FromStr};
 
+use crate::errors::{CliError, UserInputError};
+
 /// Represents a type of network, either Alpen's signet or Strata
 #[non_exhaustive]
 #[derive(PartialEq, Eq, Debug, Clone)]
@@ -8,17 +10,14 @@ pub enum NetworkType {
     Strata,
 }
 
-/// Attempted to parse a string into [`NetworkType`] but the input was invalid.
-pub struct InvalidNetworkType;
-
 impl FromStr for NetworkType {
-    type Err = InvalidNetworkType;
+    type Err = CliError;
 
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
+    fn from_str(s: &str) -> Result<Self, CliError> {
         match s.to_lowercase().as_str() {
             "signet" => Ok(Self::Signet),
             "strata" => Ok(Self::Strata),
-            _ => Err(InvalidNetworkType),
+            _ => Err(CliError::UserInput(UserInputError::UnsupportedNetwork)),
         }
     }
 }
@@ -34,12 +33,6 @@ impl fmt::Display for NetworkType {
 }
 
 /// Parses `val` as a [`NetworkType`]. Prints error message and exits if invalid.
-pub fn net_type_or_exit(val: &str) -> NetworkType {
-    match NetworkType::from_str(val) {
-        Ok(t) => t,
-        Err(InvalidNetworkType) => {
-            println!("Invalid network type. Must be signet or strata");
-            std::process::exit(1)
-        }
-    }
+pub fn parse_net_type(val: &str) -> Result<NetworkType, CliError> {
+    val.parse()
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13500,6 +13500,7 @@ version = "0.3.0"
 dependencies = [
  "aes-gcm-siv",
  "alloy",
+ "anyhow",
  "argh",
  "argon2",
  "async-trait",
@@ -13523,6 +13524,7 @@ dependencies = [
  "sled",
  "strata-primitives",
  "terrors",
+ "thiserror 2.0.11",
  "tokio",
  "zeroize",
  "zxcvbn",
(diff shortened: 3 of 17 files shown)


krsnapaudel:
I found another panic from signet. Working on it.

bewakes:
Great. This is a simple categorization of errors into two high level classes: UserInput, and Internal(which users can't do much about).

However, All of the `Internal` errors are just enclosing anyhow errors, which is quite opaque. I think it would be better to have a separate enum variant for each of those, just in case we want to handle some of those internal errors differently in the future rather than just displaying and exiting. I have something like this in mind:

```rust
enum InternalError {
  #error("Failed to load wallet")
  WalletLoadFailed,

  #error("Failed to fetch balance")
  BalanceFetchFailed,
 
  ...
}
```
And later we can assign error code to each of these variants. Something like `Wallet Load failed: Error code -12212`

Also, perhaps we can get rid of the `CliError` and directly just have `UserError` and the above variant to avoid redundant nesting. The top level `CliError` does not seem to provide much value right now because every error is cli error because the tool is cli.

What do you think?

bewakes (changes requested):
Having a lot of `.map_err(|e| OneOf::new(...format!("{e:?}")))` seems too much work. There's some pattern here. How about creating a higher order function that simplifies this? Something like:

```rust
/// For internal errors
fn make_internal_err<E, F>(f: F) -> impl FnOnce(E) -> CliError
where
    E: std::fmt::Debug,
    F: FnOnce(String) -> InternalError,
{
    move |e| OneOf::new(f(format!("{e:?}")))
}

/// For user errors
fn make_user_err<E, F>(f: F) -> impl FnOnce(E) -> CliError
where
    E: std::fmt::Debug,
    F: FnOnce(String) -> UserError,
{
    move |e| OneOf::new(f(format!("{e:?}")))
}
```
Which can then be used as say
```rust
let mut new_pw =
        Password::read(true).map_err(make_internal_err(InternalError::ReadPassword))?;
```
instead of having to do:
```rust
let mut new_pw = Password::read(true)
        .map_err(|e| OneOf::new(InternalError::ReadPassword(format!("{e:?}"))))?;
```
Which I think is easier to write and hides away the string formatting.

krsnapaudel:
@Zk2u : Let's have a call to discuss. Maybe I am unsure how to go about this in the right way.

bin/strata-cli/src/errors.rs
```diff
@@ -0,0 +1,89 @@
...
+    move |e| {
+        DisplayedError::InternalError(
+            String::from_str(msg.as_ref()).expect("infallible"),
+            Box::new(e),
+        )
+    }
+}
+
+pub(crate) trait DisplayableError {
+    type T;
```
delbonis, line 63: This associated type should be `Output`.  Just `T` as an associated type is ambiguous.

I'm a bit confused on how this is intended to be used, when considering the names here.  For a comparison, see `anyhow`'s `context` and `with_context` functions.  It seems like this is doing something similar.

bin/strata-cli/src/errors.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn user_error(self, msg: impl AsRef<str>) -> Result<Self::T, DisplayedError>;
+    fn internal_error(self, msg: impl AsRef<str>) -> Result<Self::T, DisplayedError>;
+}
+
+impl<T, E: std::fmt::Debug + 'static> DisplayableError for Result<T, E> {
+    type T = T;
+
+    fn user_error(self, msg: impl AsRef<str>) -> Result<Self::T, DisplayedError> {
+        self.map_err(|_| DisplayedError::UserError(msg.as_ref().to_string()))
```
delbonis, line 72: Shouldn't we still want to print something about the the underlying error somewhere?

bin/strata-cli/src/net_type.rs
```diff
@@ -8,17 +10,17 @@ pub enum NetworkType {
...
+    type Err = DisplayedError;
 
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
+    fn from_str(s: &str) -> Result<Self, DisplayedError> {
         match s.to_lowercase().as_str() {
             "signet" => Ok(Self::Signet),
             "strata" => Ok(Self::Strata),
-            _ => Err(InvalidNetworkType),
+            _ => Err(user_error(format!(
+                "Unsupported network: '{}'. Must be `signet` or `strata`.",
```
delbonis, line 21: Can be `{s}`.

bin/strata-cli/src/signet.rs
```diff
@@ -163,25 +163,30 @@ pub async fn sync_wallet(
...
-            }
-            WalletUpdate::SpkScan(update) => {
-                wallet.apply_update(update).expect("update to connect")
-            }
+            WalletUpdate::SpkSync(update) => wallet
+                .apply_update(update)
+                .map_err(|e| UpdateError::from(Box::new(e) as BoxedErr))?,
+            WalletUpdate::SpkScan(update) => wallet
+                .apply_update(update)
+                .map_err(|e| UpdateError::from(Box::new(e) as BoxedErr))?,
```
delbonis, line 177: Here you could have another shim.

```rs
impl UpdateError {
  fn from_err<E: foobar + Sync + Send>(e: E) -> Self {
    Self::from(Box::new(e) as BoxedError)
  }
}
```

So that you could just call it as `.map_err(UpdateError::from_err)`.

bin/strata-cli/src/signet/backend.rs
```diff
@@ -26,26 +26,40 @@ use tokio::sync::mpsc::UnboundedSender;
...
-
         impl From<BoxedErr> for $name {
             fn from(err: BoxedErr) -> Self {
                 Self(err)
             }
         }
+
+        impl std::fmt::Display for $name {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                write!(f, "{:?}", self.0)
```
delbonis, line 46: Should use `{}` formatting in here since it's a `Display` impl no?

But actually, since we're just calling out to the inner type, it's slightly better to do `std::fmt::Display::fmt(&self.0, f)` (or `Debug`) since it avoids the macro expansion and any indirection for the formatter that would happen there.

delbonis (changes requested):
Looks pretty good, some things could be more idiomatic.  Sorry for the delay.

bin/strata-cli/src/signet/backend.rs
storopoli: It is not good to do `Debug` stuff, i.e. `{:?}` in `Display` impls

bin/alpen-cli/src/cmd/drain.rs
```diff
@@ -43,24 +50,55 @@ pub async fn drain(
...
+        .map(|a| {
+            AlpenAddress::from_str(&a).map_err(|_| {
+                DisplayedError::UserError(
+                    format!(
+                        "Invalid Alpen address '{}'. Must be an EVM-compatible address",
+                        a
+                    ),
+                    Box::new(InvalidAlpenAddress),
+                )
+            })
```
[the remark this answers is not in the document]
krsnapaudel: @Zk2u If this is the case we also don't need UnsupportedMnemonicLanguage and MissingTargetAddress that you added. I will remove those as well.

bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -127,23 +142,26 @@ pub async fn deposit(
         builder.add_recipient(bridge_in_address.script_pubkey(), BRIDGE_IN_AMOUNT);
         builder.add_data(&op_return_data);
         builder.fee_rate(fee_rate);
-        builder.finish().expect("valid psbt")
+        builder
+            .finish()
+            .internal_error("Failed to create bridge transaction")?
```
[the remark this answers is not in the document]
krsnapaudel: Actually not.
`Internal error: Failed to create bridge transaction: CoinSelection(InsufficientFunds { needed: 1000001122 SAT, available: 0 SAT })`

delbonis (dismissed):
Seems pretty good now.

bin/alpen-cli/src/cmd/deposit.rs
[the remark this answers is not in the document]
krsnapaudel: I have marked it a user error. Let me know if that's correct.

storopoli (approved):
ACK 9e7b9a1f51b03c97ea70c06141d3697d90e5b53c

Getting fancy with `@` sigil bind operator in `match` statements...

delbonis (approved):

PR #782, opened 2025-04-25
l1tx: Add deposit transaction signature validation logic

bewakes:
## Description

This PR introduces an extra check to the parsed deposit transaction. It now also checks that the deposit transaction is not only being spent to N-of-N operators with specified amount but is also being spent by N-of-N operators.

This is necessary because anyone can create a deposit transaction that spends the amount to the operators. But since this won't have any Deposit request transaction, and hence no deposit index assigned, operators can't withdraw from this using the protocol.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/l1tx/Cargo.toml
+++ b/crates/l1tx/Cargo.toml
@@ -16,7 +16,6 @@ thiserror.workspace = true
 tracing.workspace = true
 
 [dev-dependencies]
-hex.workspace = true
 strata-btcio = { workspace = true, features = ["test_utils"] }
 strata-state = { workspace = true, features = ["test_utils"] }
 strata-test-utils.workspace = true
--- a/crates/l1tx/Cargo.toml
+++ b/crates/l1tx/Cargo.toml
@@ -11,6 +11,7 @@ anyhow.workspace = true
 bitcoin.workspace = true
 borsh.workspace = true
 musig2.workspace = true
+secp256k1.workspace = true
 thiserror.workspace = true
 tracing.workspace = true
 
--- a/crates/l1tx/Cargo.toml
+++ b/crates/l1tx/Cargo.toml
@@ -16,6 +16,7 @@ thiserror.workspace = true
 tracing.workspace = true
 
 [dev-dependencies]
+hex.workspace = true
 strata-btcio = { workspace = true, features = ["test_utils"] }
 strata-state = { workspace = true, features = ["test_utils"] }
 strata-test-utils.workspace = true
--- a/crates/l1tx/src/deposit/deposit_tx.rs
+++ b/crates/l1tx/src/deposit/deposit_tx.rs
@@ -2,7 +2,6 @@
 
 use bitcoin::{
     hashes::Hash,
-    key::TapTweak,
     opcodes::all::OP_RETURN,
     sighash::{Prevouts, SighashCache},
     Amount, OutPoint, ScriptBuf, TapNodeHash, TapSighashType, Transaction, TxOut, XOnlyPublicKey,
@@ -10,7 +9,7 @@ use bitcoin::{
 use secp256k1::{schnorr::Signature, Message};
 use strata_primitives::{
     buf::Buf32,
-    l1::{DepositInfo, OutputRef, XOnlyPk},
+    l1::{DepositInfo, OutputRef},
     prelude::DepositTxParams,
 };
 
@@ -80,7 +79,7 @@ fn validate_deposit_signature(
     let schnorr_sig = Signature::from_slice(sig_bytes.get(..64)?).unwrap();
 
     // Parse the internal pubkey and merkle root
-    let internal_pubkey = XOnlyPk::from_address(&dep_config.address).ok()?;
+    let internal_pubkey = dep_config.operators_pubkey;
     let merkle_root: TapNodeHash = TapNodeHash::from_byte_array(*tag_data.tapscript_root.as_ref());
 
     let int_key = XOnlyPublicKey::from_slice(internal_pubkey.inner().as_bytes()).unwrap();
@@ -102,12 +101,8 @@ fn validate_deposit_signature(
     // Prepare the message for signature verification
     let msg = Message::from_digest(*sighash.as_byte_array());
 
-    // Compute the tweaked output key
-    let (output_key, _) = int_key.tap_tweak(secp, Some(merkle_root));
-
     // Verify the Schnorr signature
-    secp.verify_schnorr(&schnorr_sig, &msg, &output_key.to_inner())
-        .ok()
+    secp.verify_schnorr(&schnorr_sig, &msg, &int_key).ok()
 }
 
 struct DepositTag<'buf> {
@@ -205,7 +200,10 @@ mod tests {
         script::{Builder, PushBytesBuf},
         Network,
     };
-    use strata_primitives::{l1::BitcoinAddress, params::DepositTxParams};
+    use strata_primitives::{
+        l1::{BitcoinAddress, XOnlyPk},
+        params::DepositTxParams,
+    };
 
     use crate::deposit::{
         deposit_tx::{
@@ -217,11 +215,13 @@ mod tests {
     const ADDRESS: &str = "bcrt1p729l9680ht3zf7uhl6pgdrlhfp9r29cwajr5jk3k05fer62763fscz0w4s";
 
     fn dummy_config() -> DepositTxParams {
+        let addr = BitcoinAddress::parse(ADDRESS, Network::Regtest).unwrap();
         DepositTxParams {
             magic_bytes: MAGIC_BYTES.to_vec(),
             address_length: 20,
             deposit_amount: 10,
-            address: BitcoinAddress::parse(ADDRESS, Network::Regtest).unwrap(),
+            address: addr.clone(),
+            operators_pubkey: XOnlyPk::from_address(&addr).unwrap(),
         }
     }
 
--- a/crates/l1tx/src/deposit/deposit_tx.rs
+++ b/crates/l1tx/src/deposit/deposit_tx.rs
@@ -140,7 +140,7 @@ fn parse_tag_script<'a>(
         return Err(DepositParseError::TagOversized);
     }
 
-    parse_tag(data, config)
+    parse_tag(data, &config.magic_bytes, config.address_length)
 }
 
 /// Parses the script buffer which has the following structure:
@@ -148,10 +148,10 @@ fn parse_tag_script<'a>(
 /// sats_amt(8 bytes)]
 fn parse_tag<'b>(
     buf: &'b [u8],
-    config: &DepositTxParams,
+    magic_bytes: &[u8],
+    addr_len: u8,
 ) -> Result<DepositTag<'b>, DepositParseError> {
     // data has expected magic bytes
-    let magic_bytes = &config.magic_bytes;
     let magic_len = magic_bytes.len();
 
     if buf.len() < magic_len + DEPOSIT_IDX_LEN + SATS_AMOUNT_LEN + TAKEBACK_HASH_LEN {
@@ -172,7 +172,7 @@ fn parse_tag<'b>(
         ee_takeback_amt.split_at(ee_takeback_amt.len() - SATS_AMOUNT_LEN - TAKEBACK_HASH_LEN);
 
     // Check dest_buf len
-    if dest_buf.len() != config.address_length as usize {
+    if dest_buf.len() != addr_len as usize {
         return Err(DepositParseError::InvalidDestLen(dest_buf.len() as u8));
     }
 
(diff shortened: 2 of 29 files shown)


crates/l1tx/src/deposit/deposit_tx.rs
```diff
@@ -49,9 +61,60 @@ pub fn extract_deposit_info(tx: &Transaction, config: &DepositTxParams) -> Optio
...
+
+    // --- Extract and validate input signature
+    let input = tx.input[0].clone();
+    let sig_bytes = &input.witness[0];
+    let schnorr_sig = Signature::from_slice(&sig_bytes[..64]).unwrap(); // TODO: enforce length?
+
+    // --- Parse the internal pubkey and merkle root
+    let internal_pubkey = XOnlyPk::from_address(&dep_config.address).ok()?;
+    let mut hash_bytes = [0; 32];
+    hash_bytes.copy_from_slice(tag_data.tapscript_root.as_bytes());
```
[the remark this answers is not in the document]
bewakes: This is because `TapNodeHash::from_bytes_array` expects a `[u8; 32]`.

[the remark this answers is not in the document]
Rajil1213: Does deref work?

```
TapNodeHash::from_bytes_array(*hash_bytes);
```

crates/l1tx/src/deposit/deposit_tx.rs
```diff
@@ -49,9 +61,60 @@ pub fn extract_deposit_info(tx: &Transaction, config: &DepositTxParams) -> Optio
...
+    let merkle_root: TapNodeHash = TapNodeHash::from_byte_array(hash_bytes);
+
+    let int_key = XOnlyPublicKey::from_slice(internal_pubkey.inner().as_bytes()).unwrap();
+    // --- Compute the tweaked output key
+    let (output_key, _) = int_key.tap_tweak(&secp, Some(merkle_root));
```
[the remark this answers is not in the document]
storopoli: Yes it needs to be tweaked.
If (from above):

```rs
let internal_pubkey = XOnlyPk::from_address(&dep_config.address).ok()?;
```

is the N-of-N bridge pubkey then the internal key needs to be tweaked by the merkle root hash to derive the P2TR address (i.e. the external key).

delbonis (changes requested):
Agree with Rajil's comments.

Surely *all* of the logic being done in `validate_deposit_signature` is already being done somewhere else and doesn't need to be reimplemented, isn't it?  Should we be importing a strata-bridge repo crate for this?

storopoli:
> Agree with Rajil's comments.
> 
> Surely _all_ of the logic being done in `validate_deposit_signature` is already being done somewhere else and doesn't need to be reimplemented, isn't it? Should we be importing a strata-bridge repo crate for this?

You can use the [`DepositInfo::compute_spend_infos`](https://github.com/alpenlabs/strata-bridge/blob/e6b2b141e91f49dae9fb386f881b40c8e47e89cf/crates/primitives/src/deposit.rs#L150-L212) in `strata-bridge-primitives` crate.

crates/l1tx/src/deposit/deposit_tx.rs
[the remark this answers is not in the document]
Rajil1213: Isn't that done already when constructing the `ScriptBuf` in the line below? Either way, it looks like the `output_key` is not being used.

Rajil1213:
> You can use the [DepositInfo::compute_spend_infos](https://github.com/alpenlabs/strata-bridge/blob/e6b2b141e91f49dae9fb386f881b40c8e47e89cf/crates/primitives/src/deposit.rs#L150-L212) in strata-bridge-primitives crate.

No, that function validates the Deposit Request and not the Deposit Transaction. There is nothing in the bridge that actually checks if a given transaction is a valid Deposit because we can directly check for the Txid (constructed off of the Deposit Request Transaction).

crates/l1tx/src/deposit/deposit_tx.rs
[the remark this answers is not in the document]
bewakes: Deref worked like this:

```rust
let merkle_root: TapNodeHash = TapNodeHash::from_byte_array(*tag_data.tapscript_root.as_ref());
``` 
thanks!

crates/l1tx/src/deposit/deposit_tx.rs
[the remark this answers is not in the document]
bewakes: We do need the tweaked `output_key` as well, but only when we are verifying schnorr down. Generating script pubkey doesn't need the tweaked key.

crates/l1tx/src/deposit/deposit_tx.rs
```diff
@@ -20,6 +20,10 @@ use crate::{
     utils::{next_bytes, next_op},
 };
 
+const TAKEBACK_HASH_LEN: usize = 32;
```
storopoli, line 22: Use https://docs.rs/bitcoin/latest/bitcoin/taproot/constant.TAPROOT_CONTROL_NODE_SIZE.html

crates/l1tx/src/deposit/deposit_tx.rs
```diff
@@ -20,6 +20,10 @@ use crate::{
...
+const SATS_AMOUNT_LEN: usize = 8;
```
storopoli, line 23: Use `std::mem::sizeof::<u64>()`: https://doc.rust-lang.org/std/mem/fn.size_of.html

crates/l1tx/src/deposit/deposit_tx.rs
```diff
@@ -20,6 +20,10 @@ use crate::{
...
+const DEPOSIT_IDX_LEN: usize = 4;
```
storopoli, line 24: Use std::mem::sizeof::<u32>(): https://doc.rust-lang.org/std/mem/fn.size_of.html

storopoli (changes requested):

bewakes:
I've ignored the multiple ops testcases for now.

Rajil1213:
> @Rajil1213 do you think it is fine to expect there's another input that contains checkpoint/da data as witness in a transaction that contains deposit?

Apologies for the late response. From a bridge perspective, it does not make much practical sense for a transaction to be both a deposit _and_ a checkpoint/da because they are posted by different entities via different independent processes. At least for the moment, I don't see how/why a the bridge operators would also be responsible for posting a checkpoint transaction.

Rajil1213 (changes requested):
Looks good but I still don't see a more comprehensive test that specifically checks for the feature that is being added in this PR namely, a transaction that only sends funds to the multisig address (with the correct metadata) will fail validation if it has not been signed off by the bridge operators, and inversely, a transaction that is signed off by all the bridge operators should be successfully validated. It could be that this test exists and I missed it, in which case, could you point me towards that test?

In case the test does not indeed exist, I think it is a good idea to have a sample of a valid/invalid deposit transaction somewhere in this codebase. This could be inside a file inside the `test-data` directory or could be created on the fly. I'll be happy to help out in creating either one.

crates/l1tx/src/deposit/deposit_tx.rs
```diff
@@ -49,9 +65,53 @@ pub fn extract_deposit_info(tx: &Transaction, config: &DepositTxParams) -> Optio
...
+    let prevout = Prevouts::All(&utxos);
+    let sighash = SighashCache::new(tx)
+        .taproot_key_spend_signature_hash(0, &prevout, TapSighashType::Default)
+        .unwrap();
+
+    // Prepare the message for signature verification
+    let msg = Message::from_digest(*sighash.as_byte_array());
+
+    // Verify the Schnorr signature
+    secp.verify_schnorr(&schnorr_sig, &msg, &int_key).ok()
```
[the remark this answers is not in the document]
bewakes: It shouldn't be. We sign with the internal private key right, not the tweaked one.

crates/l1tx/src/deposit/deposit_tx.rs
```diff
@@ -76,81 +136,239 @@ fn parse_tag_script<'a>(
...
+        let script = Builder::new()
+            .push_opcode(OP_RETURN)
+            .push_slice(PushBytesBuf::try_from(valid_payload).unwrap())
+            .into_script();
+        let config = dummy_config();
 
-        assert!(out.is_some());
-        let out = out.unwrap();
+        // Might still fail inside parse_tag (e.g., InvalidMagic), but must NOT fail for
+        // MissingTag/NoData/TagOversized
```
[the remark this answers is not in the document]
bewakes: Yes, because `parse_tag_script` does:
1. check if valid opreturn script
2. parse opreturn data.

This test just checks for the first case and hence only makes sure errors from the first case don't occur.

bewakes:
@Rajil1213 thank you for pointing the tests out and for offering help!

Actually the tests for valid deposit and when opreturn is empty are already there. You can check `crates/l1tx/filter/mod.rs`.
I've now also added two more cases: one for tapnode hash mismatch and the other for signing with different internal key.

crates/l1tx/src/deposit/deposit_tx.rs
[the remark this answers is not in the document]
Rajil1213: I think we sign with the tweaked key. It might be useful to perform the test on an actual deposit transaction (on bitcoin) to confirm this.

Rajil1213 (changes requested):
While we're here, can we also change this line to use `config.deposit_amount` instead of the hardcoded `BRIDGE_DENOMINATION`?

https://github.com/alpenlabs/strata/blob/2ddac56677f55cf864bcbac431e86b62f97cfc8b/crates/l1tx/src/deposit/deposit_tx.rs#L35-L39

crates/l1tx/src/deposit/deposit_tx.rs
```diff
@@ -76,81 +136,239 @@ fn parse_tag_script<'a>(
...
 ) -> Result<DepositTag<'b>, DepositParseError> {
     // data has expected magic bytes
-    let magic_bytes = &config.magic_bytes;
     let magic_len = magic_bytes.len();
 
-    // Do some math to make sure there is a magic.
-    let exp_min_len = magic_len + 4;
-    if buf.len() < exp_min_len {
-        return Err(DepositParseError::InvalidMagic);
+    if buf.len() < magic_len + DEPOSIT_IDX_LEN + SATS_AMOUNT_LEN + TAKEBACK_HASH_LEN {
```
Rajil1213, line 153: This check should also include the EE address length.

```suggestion
    if buf.len() < magic_len + DEPOSIT_IDX_LEN + addr_len as usize + SATS_AMOUNT_LEN + TAKEBACK_HASH_LEN {
```

crates/test-utils/src/bitcoin.rs
```diff
@@ -74,12 +97,27 @@ pub fn create_test_deposit_tx(
...
+        .unwrap();
+
+    let msg = Message::from_digest(*sighash.as_ref());
+
+    // Sign the sighash
+    let sig = secp.sign_schnorr(&msg, keypair);
+
+    tx.input[0].witness.push(sig.as_ref());
+
+    tx
```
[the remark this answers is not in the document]
bewakes: That is a fair concern. However, whether a transaction follows the consensus rule in bitcoin is something that tests in btcio filter care less about. They are more concerned about the transaction containing data in a specific format. Having said that, I don't mean txs in unit tests should not follow the consensus rules, but as long as we are creating structures(like op return and script pubkey in this case) properly, the unit tests should be fine.

What you are pointing to here probably resides on the functional test side, which I thought were already there and should have covered the consensus testings you mentioned. But I see no bridge related functional tests - they must have been removed as a part of splitting out the bridge code. Not sure how those ends can be tested.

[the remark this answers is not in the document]
Rajil1213: I partly agree. For the sake of completeness, there are two things that need to be tested, namely whether the deposit transaction is:

1. _structurally_ valid (single witness, right OP_RETURN data, etc.), and
2. _cryptographically_ valid (witness is correct such that the DRT can be spent).

I agree that we don't need a consensus-valid transaction to check the first case and the current test is indeed already doing that.

However, I strongly recommend also checking the latter. As it stands right now, the test constructs a transaction, computes a sighash, signs it with a key and then, reverse-engineers that process in the test by reconstructing the same sighash and verifying the signature against the public key counterpart. At no point does it actually check whether the sighash construction is valid or whether the right key is being used (tweaked/untweaked). Given the opaque nature of these cryptographic constructs, I think these deserve to be tested more rigorously.

One way to do this is to use `corepc-node` to spin up a bitcoin-core instance, construct and broadcast a DRT and then, construct and broadcast the DT that spends this DRT. For the purposes of this test, the DRT does not need to be completely valid -- the OL only cares about the hash of the script so the script itself can just be an `OP_TRUE`. While this is a much better test, I understand that it might be too much of a hassle. An alternative is to have a test vector for the deposit transaction. To that end, here is a consensus-valid deposit transaction:

```
020000000001012bcf5b9855dd3ba8dcf73eaa08587350f5aa8b5e2254027a26a4a10810c5d1f50000000000fdffffff0200e1f505000000002251209accbdba14ffa7ca9d636ec63beee4c00c9b9504e4fc4f71a779491907ff1ef10000000000000000486a467374726174610000000070997970c51812dc3a010c7d01b50e0d17dc79c8b8604298f5b77e6b863aac2a613adf755ef12a7f17dbe8a22d9947907e1ff2900000000005f767a0014151fc51a6332aac566a93f99c4e3dd85ad21d90f408c319980ebb3283b0a4d802f1c5abd73a76a2d84aad951d957e58d8d5f4342b4a0289bfb66ea278c23a98480100000000
```

This has been signed off by 3 operators with the following x-only public keys:

```
0xb49092f76d06f8002e0b7f1c63b5058db23fd4465b4f6954b53e1f352a04754d
0x1e62d54af30569fd7269c14b6766f74d85ea00c911c4e1a423d4ba2ae4c34dc4
0xa4d869ccd09c470f8f86d3f1b0997fa2695933aaea001875b9db145ae9c1f4ba
```

I'm fine with either approach as long as we are testing on an "actual" deposit transaction.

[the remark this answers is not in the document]
bewakes: I see your point regarding the reverse engineering and I agree. As for the ways to do these, I think it's better to have test vectors as it is much easier and tests what needs to be tested. Similarly, it is also agnostic of DRT's existence and how it is created. Since strata does not care about DRT(it might store it for other purposes, but protocol does not require it), it's best to not include its construction even in the tests.

I will add a new test case for this.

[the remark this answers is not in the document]
delbonis: Is there a reason that we would want to re-check consensus validity rules here?

crates/l1tx/src/deposit/deposit_tx.rs
```diff
@@ -49,9 +63,53 @@ pub fn extract_deposit_info(tx: &Transaction, config: &DepositTxParams) -> Optio
...
+    let prevout = Prevouts::All(&utxos);
+    let sighash = SighashCache::new(tx)
+        .taproot_key_spend_signature_hash(0, &prevout, TapSighashType::Default)
+        .unwrap();
+
+    // Prepare the message for signature verification
+    let msg = Message::from_digest(*sighash.as_byte_array());
+
+    // Verify the Schnorr signature
+    secp.verify_schnorr(&schnorr_sig, &msg, &int_key).ok()
```
delbonis, line 104: Yeah as discussed this is trying to reimplement bitcoin's transction validation logic.

There exists a [`Transaction::verify`](https://docs.rs/bitcoin/latest/bitcoin/struct.Transaction.html#method.verify) function that we can probably use to achieve the result we're looking for in checking the witness.  Manually reimplementing taproot validation is *not* the way to go here because there's bound to be a nontrivial xor of the sets of admitted transactions between this and the widely agreed upon algorithm due to all the special casing and stuff.

Then after doing the checks we can just pull out the pubkeys on the involved `TxOut`s and whatever else we might want to extract to do our own additional checks.

delbonis (changes requested):

crates/l1tx/src/deposit/deposit_tx.rs
delbonis: Also, I *believe* this code would panic on
* an OP_TRUE output being spent with an empty witness (not likely, but a miner could do it easily)
* a witness where the first witness stack element isn't a valid signature (think like a hash preimage reveal but we use `0xff`*64 as the preimage)

crates/test-utils/src/bitcoin.rs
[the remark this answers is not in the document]
Rajil1213: > Is there a reason that we would want to re-check consensus validity rules here?

This is only in context of the test. The actual code does not need to check full consensus-validity since it will only ever receive valid transactions -- it only needs to check if the witness in the transaction corresponds to the aggregated bridge pubkey.

In the test, I want to make sure that we are testing against a transaction that actually follows all the consensus rules.

crates/l1tx/src/deposit/deposit_tx.rs
Rajil1213: > There exists a [Transaction::verify](https://docs.rs/bitcoin/latest/bitcoin/struct.Transaction.html#method.verify) function that we can probably use to achieve the result we're looking for in...
> Then after doing the checks we can just pull out the pubkeys on the involved TxOuts and whatever else we might want to extract to do our own additional checks.

Nice. I didn't know this existed. It does not really help us though -- we still need to compute the sighash in order to verify the signature against the aggregated bridge pubkey. And as far as I can tell, that computation is what you're trying to avoid and not the schnorr verification.

You're right about the two extra panic cases that we should handle here.

bewakes: Using `Transaction::verify` is a great suggestion! For a moment I was happy to discover it. However, it does not support Taproot or Tapscript at the moment, so it can't be used.

I have added some checks for the possible panics that might occur.

crates/test-utils/src/bitcoin.rs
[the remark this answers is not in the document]
bewakes: There shouldn't actually be a reason to re-check the consensus validity rules here. What @Rajil1213  is saying is that the way deposit transaction was created for the tests looks brittle because there are aspects to schnorr signature and signing ceremony that we might not be doing correctly. In particular, the tests doing the exact reverse of what's being done in the actual validation function. So it would be nice to check this works with the actual transaction.
And in fact he is right, I was signing the test transaction using the untweaked key while it should have been done by the tweaked key.

crates/l1tx/src/filter/mod.rs
```diff
@@ -230,4 +235,60 @@ mod test {
...
+        // Configure and update rollup params by setting the custom operator keys
+        let mut params = gen_params();
+
+        params.rollup.operator_config = OperatorConfig::Static(op_pubkeys);
+
+        // Derive filter config and update magic bytes and deposit amount that the above transaction
+        // has.
+        let mut filterconf = TxFilterConfig::derive_from(params.rollup()).unwrap();
+        filterconf.deposit_config.magic_bytes = "strata".bytes().collect();
+        filterconf.deposit_config.deposit_amount = 100000000;
```
storopoli, line 266: So many ZER00000000000000000000000ES!

```suggestion
        filterconf.deposit_config.deposit_amount = 100_000_000;
```

storopoli (changes requested):
minor nit

crates/l1tx/src/filter/mod.rs
bewakes: Oh yes. Almost killed a puppy :D . Updated.

storopoli (approved):
ACK 4240c9b7dc9d0460fe613e528fde2717f6cf9e1a

crates/l1tx/src/deposit/deposit_tx.rs
```diff
@@ -74,14 +75,24 @@ fn validate_deposit_signature(
...
+    // Check if witness is present.
+    if input.witness.is_empty() {
+        return None;
+    }
+    let sig_witness = &input.witness[0];
+    if sig_witness.len() < 64 {
+        return None;
+    }
+    let sig_bytes = &sig_witness[..64];
+    let schnorr_sig = Signature::from_slice(sig_bytes).unwrap();
```
delbonis, line 88: This still has the `0xff`*64 panic I mentioned.  The `.get` slicing was fine before, the issue is that you were indexing (instead of `.get()`) and calling `.unwrap()`.

delbonis (changes requested):
I really hate that we have to reimpl this but if this is the only way then eh I suppose.

crates/l1tx/src/deposit/deposit_tx.rs
storopoli: Also, please don't use magic numbers like `64`. You can use constants such as https://docs.rs/bitcoin/latest/bitcoin/key/constants/constant.SCHNORR_SIGNATURE_SIZE.html

crates/l1tx/src/deposit/deposit_tx.rs
bewakes: No it doesn't in the `bitcoin-rust` version that we use. It uses `bitcoinconsensus==0.105` under the hood which does not support taproot. `bitcoin-rust` has the updated `bitcoinconsensus` to `0.106` which supports taproot in their main. But I don't want to use that git revision as a dependency. It seems to introduce some breaking changes.

crates/l1tx/src/deposit/deposit_tx.rs
bewakes: Fixed.

crates/l1tx/src/deposit/deposit_tx.rs
Rajil1213: Even if it did support taproot (which it will in the not-too-distant future), how would it help us? We'd still need to compute the message (sighash) in order to verify the signature against the bridge pubkey, don't we?

If it is indeed possible to get away without having to compute the sighash, then we should definitely leave a FIXME comment here somewhere.

crates/l1tx/Cargo.toml
```diff
@@ -16,6 +16,7 @@ thiserror.workspace = true
 tracing.workspace = true
 
 [dev-dependencies]
+hex.workspace = true
```
storopoli, line 19: But you don't need that. You have `bitcoin` in your dependencies and you can use `use bitcoin::hex_conservative as hex`

See the docs here: https://docs.rs/hex-conservative/0.2.0/hex_conservative/index.html

storopoli (changes requested):

storopoli (dismissed):
ACK 3f63989b92a813b501507ecd0dc4fc8a75af3a9d

crates/l1tx/src/deposit/deposit_tx.rs
bewakes: I don't think we need to compute the sighash. Once we have `TxOut`, which we need to compute from the agg pubkey and tapnode hash, for each input(which is just 1 in the case of deposit), it should be enough to verify the transaction. As far as I know, the sighash info is in the signature itself(if it is not the default sighash). Once sighash type is known, the consensus can validate the transaction.

Left a FIXME.

Rajil1213: Oh I see. Yeah that should work. When we do make that fix, we should check the witness size as well because there's a highly unlikely but possible case where the depositer spends the funds into the bridge address with the right metadata structure after whatever the take back period is. Checking the witness size ensures that a keypath spend has been used which necessarily means that it was spent by the bridge operators.

delbonis: > But I don't want to use that git revision as a dependency.

I'll create a Jira ticket.

PR #798, opened 2025-05-02
Str 1172 ignore checkpoints with invalid proofs

irnb:
## Description

* Functional Test for STR-1172:
* Fix proof verification logic to correctly reject empty proofs when strict proof publishing mode is enabled.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

STR-1172


crates/consensus-logic/src/checkpoint_verification.rs
```diff
@@ -112,5 +113,11 @@ pub fn verify_proof(
         return Ok(());
     }
 
+    if !accept_empty_proof && is_empty_proof {
+        return Err(ZkVmError::ProofVerificationError(
+            "Empty proof received, which is not acceptable in strict proof mode".to_string(),
```
storopoli, line 118: Can we have also an active voice recommendation here.
Something like "check the environmental variable FOO" or "use BAR" etc?

functional-tests/tests/cl_ignore_checkpoint_with_invalid_proof.py
```diff
@@ -0,0 +1,214 @@
...
+              for epoch 4 to L1 again.
+            - The full node and sequencer 2 should ignore the checkpoint with invalid proof,
+              not finalize epoch 4 based on it, and wait for a valid checkpoint.
+            - Start the strict prover again.
+            - Both the full node and sequencer 2 should continue working normally,
+              generate a valid proof, and finalize epoch 4 (and subsequently epoch 5)
+              with the valid checkpoint.
+            - After finalizing epoch 4 with the valid proof, resubmit the previous
+              invalid proof data for epoch 4.
+            - The full node and sequencer 2 should ignore this resubmission and continue
```
voidash, line 53: It's nice of you to include the steps. However steps are explanatory in the code below too.

voidash (commented):
Some nits

functional-tests/tests/fullnode_ignore_checkpoint_with_invalid_proof.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+
+        seq_fast = ctx.get_service("seq_node_fast")
+        prover_fast = ctx.get_service("prover_client_fast")
+
+        # this fullnode has a strict proof policy but connected to the fast sequencer
+        fullnode = ctx.get_service("fullnode")
+
+        seq_strict = ctx.get_service("seq_node_strict")
+        prover_strict = ctx.get_service("prover_client_strict")
```
voidash, line 43:
```suggestion
        seq_fast = ctx.get_service("seq_node_fast")
        prover_fast = ctx.get_service("prover_client_fast")
        seq_strict = ctx.get_service("seq_node_strict")
        prover_strict = ctx.get_service("prover_client_strict")
        
               # this fullnode has a strict proof policy but connected to the fast sequencer
        fullnode = ctx.get_service("fullnode")
```

functional-tests/tests/cl_ignore_checkpoint_with_invalid_proof.py
barakshani: I find this helpful.

functional-tests/envs/testenv.py
```diff
@@ -383,6 +384,226 @@ def init(self, ctx: flexitest.EnvContext) -> flexitest.LiveEnv:
         return BasicLiveEnv(svcs, bridge_pk, rollup_cfg)
 
 
+class IgnoreCheckpointWithInvalidProofEnvConfig(flexitest.EnvConfig):
```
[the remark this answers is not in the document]
irnb: Thanks for suggesting the name, definitely a better fit! 🔥 

As for the second suggestion, would you prefer we address it in this PR, or polish it in future changes?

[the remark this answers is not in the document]
voidash: I think it's small enough that we can address this in this PR. It's your call

functional-tests/tests/cl_ignore_checkpoint_with_invalid_proof.py
irnb: I also think having these kinds of test step comments helps reviewers quickly understand the test flow and makes it easier to catch any unintended behavior.

That said, I realize having large comments directly in the test files isn’t the cleanest approach and comes with some trade-offs (though I’m personally fine with it.)

Maybe we could move this type of explanation to the documentation instead.

wdyt?

functional-tests/tests/fullnode_ignore_checkpoint_with_invalid_proof.py
irnb: Thanks for the review! 😄
I’ve addressed most of the comments.

functional-tests/tests/cl_ignore_checkpoint_with_invalid_proof.py
```diff
@@ -0,0 +1,213 @@
...
+
+        ## Start strict sequencer related services
+        prover_strict.start()
+        seq_strict.start()
+        fullnode.start()
+
+        fullnode_rpc = fullnode.create_rpc()
+
+        ## Wait for strict sequencer to finalize epoch 3
+        wait_until_epoch_finalized(seqrpc_strict, 3, timeout=60)
```
barakshani, line 150: What happens here? Do we re-generate the first 3 epochs/batches?

crates/consensus-logic/src/checkpoint_verification.rs
```diff
@@ -112,5 +113,114 @@ pub fn verify_proof(
...
+            matches!(result, Err(ZkVmError::ProofVerificationError(msg)) if msg.contains("Empty proof received for checkpoint") && msg.contains("which is not allowed in strict proof mode"))
+        );
+    }
+
+    #[test]
+    fn test_empty_proof_on_timeout_mode_with_non_native_vk() {
+        let (checkpoint, mut rollup_params) = get_test_input();
+
+        // Ensure the mode is Timeout for this test
+        rollup_params.proof_publish_mode = ProofPublishMode::Timeout(1000);
```
storopoli, line 190:
```suggestion
        rollup_params.proof_publish_mode = ProofPublishMode::Timeout(1_000);
```

crates/consensus-logic/src/checkpoint_verification.rs
```diff
@@ -112,5 +113,114 @@ pub fn verify_proof(
...
+
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn test_empty_proof_on_timeout_mode_with_native_vk() {
+        let (checkpoint, mut rollup_params) = get_test_input();
+
+        // Ensure the mode is Timeout for this test
+        rollup_params.proof_publish_mode = ProofPublishMode::Timeout(1000);
```
storopoli, line 211:
```suggestion
        rollup_params.proof_publish_mode = ProofPublishMode::Timeout(1_000);
```

functional-tests/envs/testenv.py
```diff
@@ -383,6 +384,225 @@ def init(self, ctx: flexitest.EnvContext) -> flexitest.LiveEnv:
         return BasicLiveEnv(svcs, bridge_pk, rollup_cfg)
 
 
+class DualSequencerMixedPolicyEnvConfig(flexitest.EnvConfig):
```
storopoli, line 387: Please add docstrings?
What is dual sequencer for example? And why do we need that?

MdTeach (approved):

functional-tests/tests/cl_ignore_checkpoint_with_invalid_proof.py
voidash: I guess we can leave it then

functional-tests/tests/cl_ignore_checkpoint_with_invalid_proof.py
irnb: Yes, the new strict sequencer still generates three batches, but because the timestamps and L1 block numbers have changed, they don’t match the first three batches from the fast sequencer.

I’ve disabled this test for two reasons: 
   * our current Python Bitcoin library doesn’t support Taproot witness parsing.
   * we don’t yet have an L1 syncing mechanism. ( regarding to calculating valid public inputs)
   
A strict sequencer (and strict full node) should simply ignore any invalid dummy-checkpoint L1 transaction whether it fails due to bad public inputs or an invalid proof—and continue producing valid checkpoints normally, even if the invalid checkpoint carries a valid signature.

Given these limitations, the test is incomplete and may contain incorrect logic, so I’ve disabled it for now and plan to revisit it later. If you’d prefer, we can delete it entirely. When we eventually design L1 syncing, we should make sure to cover this scenario.

barakshani: > Yes, the new strict sequencer still generates three batches

So the _only_ reason we have the first sequencer is that we (get to epoch 4 and) post a mock proof to bitcoin, so later we can reuse that tx and resubmit the da blob?

irnb: exactly.

irnb: Oh nice idea 🔥 

Given that we need a valid signature for our checkpoint, I’m afraid the current `functional-test` capabilities won’t allow us; each run generates a new and random Schnorr key for signing. 
To support this, we’d need to add a new flow in the `functional-test` to start the sequencer with preconfigured credentials (btw, it’s should be a quick fix).

@voidash, please correct me if I’m mistaken.

Since this test is disabled for now, do you think it’s worth investing more time in it at the moment?

functional-tests/tests/cl_ignore_checkpoint_with_invalid_proof.py
storopoli: Yes please leave these. The only downside that I see is that it might get outdated, so we need to be diligent with PR reviews!

storopoli (approved):
ACK ce822b5c06d216d50fa958de3b55b435f4b93715

delbonis (approved):

voidash (approved):

PR #800, opened 2025-05-05
STR-1308: EVM state diff(snapshot) prototype

evgenyzdanovich:
## Description

A prototype impl of EVM DA using state "snapshot" (without compression).

What's meant by "snapshot" is: current impl of EVM DA does not use **diffs**, but rather **sets** a new state every time (without diffing currently).
 
The impl also:
- defines broad structs and mechanism for capturing state related pieces.
- contains a tested implementation of state reconstruction based only on batch snapshots (diffs in the future).

P.S. Once [basic DA diff library](https://github.com/alpenlabs/strata/pull/807) is settled, an additional lift to use diffs should be somewhat minimal. 

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist


crates/reth/statediff/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+            contracts: value.contracts,
+        }
+    }
+}
+
+/// A representation of several [`BlockStateDiff`] ready to be serialized and posted on DA.
+///
+/// N.B. "original" counterparts are needed to correctly construct the [`BatchStateDiff`]
+/// for the range of blocks and not include changes for the keys whose values
+/// were changed inside batch, but end up having the same value as before the batch.
```
delbonis, line 30: This seems like it's not just a diff then?  We should have some builder type that remembers the original values and tracks modifications we're making, comparing against the originals, and updates an inner diff structure we extract at the end?

crates/reth/statediff/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+            }
+
+            let original_acc_storage = self.original_storage_slots.entry(*addr).or_default();
+            let cur_acc_storage = self.storage_slots.entry(*addr).or_default();
+
+            for (key, value) in &acc.storage {
+                // Update original account storage if seen for the first time in the batch.
+                if !original_acc_storage.contains_key(key) {
+                    original_acc_storage.insert(*key, value.original_value());
+                }
```
delbonis, line 87: There's this weirdness here when we merge these diffs because we have to carry along the extra originals information inside the diff.  What happens if the two diffs have different originals?  That would probably be a bug, but this feels weird to me.

delbonis (changes requested):
I'm not sure what from this design comes out of how the exex API is designed.  Including tracking in the `BatchStateDiff` structure feels undesirable, see comments

I know this is a prototype and that's why we're using serde for it, but that seems like a *huge* overhead.  At least adding `#[serde(rename = ...)]` to fields to make them encode as single chars would slim this down and skipping serializing entries in `AccountInfo` if they're unchanged.

Should we also include in the prototype a checker/applicator that applies the diff on top of a previous state and ensures it matches a final state?  It would be nice to have that since that's how I imagine the flow in the proof would be more like, instead of recomputing the diff while checking the blocks and checking it matches a hash passed.

crates/reth/statediff/src/lib.rs
evgenyzdanovich: Well, agree. Introduced a thin builder type that does pretty much the same.

crates/reth/statediff/src/lib.rs
evgenyzdanovich: I mean, this weirdness, IIUC, is imposed by the exex interface. If that happens - it pretty much means there's a bug in there.

evgenyzdanovich:
@delbonis @MdTeach added some logic to reconstruct the state root based off of the batch diffs. 
Please note - the logic is very raw and not yet tested, but feel free to take a look and leave suggestions.

I'm working on proper testing of reconstructing logic - both unit and esp fntests.

crates/reth/db/src/rocksdb/schema.rs
```diff
@@ -8,3 +8,13 @@ define_table_with_seek_key_codec!(
...
 );
+
+define_table_with_seek_key_codec!(
+    /// store of block state diff data. Data stored as serialized bytes for directly serving in rpc
+    (BlockStateDiffSchema) B256 => Vec<u8>
+);
+
+define_table_with_seek_key_codec!(
+    /// block number => hash mapping for easier testing.
+    (BlockNumberByHash) u64 => Vec<u8>
```
delbonis, line 19: Shouldn't this be `u64 => B256`?  And wouldn't it be called `BlockHashByNumber`?  I'm not sure of the real schema here.

crates/reth/statediff/src/state.rs
```diff
@@ -0,0 +1,126 @@
...
+    storage_trie: HashMap<Address, MptNode>,
+}
+
+use thiserror::Error as ThisError;
+
+#[derive(ThisError, Debug)]
+/// A concrete error that may happen during state reconstruction.
+pub enum StateError {
+    #[error("Mpt Error: {0}")]
+    MptError(#[from] strata_mpt::Error),
```
delbonis, line 21: In the main strata repo we've been doing this as

```
/// docs
#[derive(Debug, ThisError)]
pub enum StateError {
    #[error("mpt: {0}")]
    ...
```

not sure what the proper style in here *ought* be.

crates/reth/statediff/src/state.rs
```diff
@@ -0,0 +1,126 @@
+use std::collections::HashMap;
+
+use revm_primitives::{alloy_primitives::Address, B256, U256};
+use strata_mpt::{keccak, MptNode, StateAccount, EMPTY_ROOT};
+
+use crate::BatchStateDiff;
+
+/// An (in-memory) representation of the EVM state reconstructed only from [`BatchStateDiff`].
+#[derive(Clone, Default, Debug)]
+pub struct StateReconstructed {
```
delbonis, line 10: `ReconstructedState`?

Also the order of items in this module is weird and split up into different parts.

crates/reth/statediff/src/state.rs
```diff
@@ -0,0 +1,126 @@
...
+
+        // Adjust accounts based on the accounts into and its slots.
+        for (address, account_info) in batch_diff.accounts {
+            let acc_info_trie_path = keccak(address);
+            if account_info.is_none() {
+                // Account was actually destructed.
+                self.state_trie.delete(&acc_info_trie_path)?;
+                continue;
+            }
+            let account_info = account_info.unwrap();
```
delbonis, line 53: Could write this as

```rs
let Some(account_info) = account_info else {
    // was deleted
    ...
    continue;
};

...
```

crates/reth/rpc/src/rpc.rs
```diff
@@ -42,4 +43,38 @@ where
...
         res.map_err(to_jsonrpsee_error("Failed fetching witness"))
     }
+
+    fn get_block_state_diff(&self, block_hash: B256) -> RpcResult<Option<BlockStateDiff>> {
+        self.db
+            .get_state_diff_by_hash(block_hash)
+            .map_err(to_jsonrpsee_error("Failed fetching block state diff"))
+    }
+
+    fn get_state_root_by_diffs(&self, block_number: u64) -> RpcResult<Option<B256>> {
```
delbonis, line 53: I would call this `_via_diffs`.  Saying "by" here implies that the diffs are a key we're looking up the state root by.

delbonis (changes requested):
Somewhat more minor changes.

I am curious about how this will lead into DA based state reconstruction?  Do we have an alternate set of stages for that?  I'm not sure what infra there is.

crates/mpt/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+edition = "2021"
+name = "strata-mpt"
+version = "0.1.0"
```
storopoli, line 4: Can we add a description? I cannot keep count on how many TLAs we have so far... What is MPT?

crates/reth/exex/src/state_diff_exex.rs
```diff
@@ -0,0 +1,75 @@
...
+                .put_state_diff(block_hash, block_number, &state_diff)
+            {
+                error!(?err, ?block_hash);
+                break;
+            }
+
+            finished_height = Some(BlockNumHash::new(block_number, block_hash))
+        }
+
+        Ok(finished_height)
```
storopoli, line 57: If the db operation has an `Error` then the function merrily returns an `Ok(Some(T))` instead of `Error`, is that really intended? Because the state diff is not in the DB and the function clearly succeed?

storopoli (changes requested):

crates/reth/db/src/rocksdb/schema.rs
evgenyzdanovich: I'm not sure why, but having `B256` as a value (instead of `Vec<u8>`) doesn't work with macro.

crates/reth/rpc/src/rpc.rs
evgenyzdanovich: done

crates/reth/statediff/src/state.rs
evgenyzdanovich: done. Also made some effort to put structure into this new crate and move pieces into moreless logical parts.

crates/reth/statediff/src/state.rs
evgenyzdanovich: Sure, changed for consistency.

crates/reth/statediff/src/state.rs
evgenyzdanovich: oh yeah, my c++ roots sometimes show up in different places, but I do agree it's much more rustean way. thanks!

crates/reth/exex/src/state_diff_exex.rs
evgenyzdanovich: Yeah, that's the intent. Could your expand why it can be problematic or propose a better way please?

crates/mpt/Cargo.toml
evgenyzdanovich: Sure, no problem. It's Merkle Patricia Trie.

evgenyzdanovich:
> I am curious about how this will lead into DA based sync? Do we have an alternate set of processing stages for that? I'm not sure what infra there is.

@delbonis I have taken a deep look at it, but it seems we gotta build a set of processing stages (given that we have plans to allow DA based sync)

crates/reth/exex/src/state_diff_exex.rs
delbonis: I feel like we don't want to lose these payloads.

crates/reth/db/src/rocksdb/schema.rs
delbonis: Oh is B256 not serializable according to their codec types?

crates/mpt/Cargo.toml
storopoli: Oh now it makes total sense. Yeah MPTs of course... But the description is very welcomed.

delbonis (approved):
I think all the major issues I looked at earlier have been dealt with so I'm approving this now, but I'm not sure what the next steps are and if those maybe make sense to do here and now.  Do you have an idea of what else is needed here?

delbonis:
Like obviously sync is a next step but that feels like we'd want it to be a different PR.  And/or making a more advanced RPC control interface.

bewakes (approved):
Just got a chance to look closely into this today. Seems like a lot of things are already in place. Great work!

crates/reth/db/src/rocksdb/schema.rs
evgenyzdanovich: Well, it seems like it's not serializable.

crates/reth/exex/src/state_diff_exex.rs
evgenyzdanovich: well, that's exactly what's being achieved here. We send the event `ExExEvent::FinishedHeight(finished_height)` only for the block height that has been committed to the db.

crates/reth/statediff/src/state.rs
evgenyzdanovich: @storopoli Well, the original comment from Trey was more about whether to put comment before `#[derive]` or after, it's **not** about error handling in IIUC.

storopoli: Yeah I was probably very tired. This is obvious. Sorry about this.
My original intention was to give awareness to the fact that we don't have a proper guidelines on how to do error structs, enums etc with `thiserror` in our guidelines.
I am resolving this for now.

storopoli (approved):
ACK 1edbb1d3b7f516708961567ead18b325be82c1c4

barakshani:
LGTM!

prajwolrg (approved):

PR #802, opened 2025-05-13
(STR-1349): Do not include deposit request in protocol ops

bewakes:
Fixes: https://alpenlabs.atlassian.net/browse/STR-1349
## Description

This PR does the following:
1. Removes including DRTs to protocol ops in btcio transaction indexer. This was inconsistent with how prover indexed DRTs. Interesting but prover did not indexed DRT while btcio indexer did.
2. Adds indexing withdrawal fulfillment and deposit spent to prover indexer. btcio indexer already had this.
3. Refactored tests to reuse similar logic between prover and btcio tx indexers. This comprises the large portion of this PR.
4. Add corresponding new tests.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/btcio/src/reader/tx_indexer.rs
+++ b/crates/btcio/src/reader/tx_indexer.rs
@@ -47,7 +47,6 @@ impl TxVisitor for ReaderTxVisitorImpl {
     }
 
     fn visit_deposit_request(&mut self, dr: DepositRequestInfo) {
-        self.ops.push(ProtocolOperation::DepositRequest(dr.clone()));
         self.deposit_requests.push(dr);
     }
 
@@ -156,6 +155,8 @@ mod test {
         }
     }
 
+    // Ignored because deposit request is not included as ops
+    #[ignore]
     #[test]
     fn test_index_txs_deposit_request() {
         let params = gen_params();
--- a/crates/btcio/src/reader/tx_indexer.rs
+++ b/crates/btcio/src/reader/tx_indexer.rs
@@ -82,6 +82,7 @@ mod test {
         test_index_deposit_request_with_visitor, test_index_deposit_with_visitor,
         test_index_multiple_deposits_with_visitor, test_index_no_deposit_with_visitor,
         test_index_tx_with_multiple_ops_with_visitor,
+        test_index_withdrawal_fulfillment_with_visitor,
     };
 
     use crate::reader::tx_indexer::ReaderTxVisitorImpl;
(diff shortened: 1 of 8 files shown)


prajwolrg (commented):
Right now there's a separate impl of `TxVisitor` for the Prover [here](https://github.com/alpenlabs/strata/blob/main/crates/proof-impl/btc-blockspace/src/tx_indexer.rs). Should we also remove this as part of this PR?

crates/btcio/src/reader/tx_indexer.rs
```diff
@@ -156,6 +155,8 @@ mod test {
         }
     }
 
+    // Ignored because deposit request is not included as ops
+    #[ignore]
     #[test]
```
storopoli, line 160:
```suggestion
    #[test]
    #[ignore = "Ignored because deposit request is not included as ops"]
```
Ref: https://doc.rust-lang.org/reference/attributes/testing.html#the-ignore-attribute

storopoli (changes requested):

Rajil1213 (changes requested):
I guess this works too but why not remove `crates/l1tx/src/deposit/deposit_request.rs` entirely?

bewakes:
> Right now there's a separate impl of TxVisitor for the Prover [here](https://github.com/alpenlabs/strata/blob/main/crates/proof-impl/btc-blockspace/src/tx_indexer.rs). Should we also remove this as part of this PR?

@prajwolrg prover already removed that code and that's why there was inconsistency between prover's protocol ops and client's protocol ops.

> I guess this works too but why not remove crates/l1tx/src/deposit/deposit_request.rs entirely?

True, we can also do that. However, client might want to store DTRs for it's own bookeeping so it's better to keep it.

crates/test-utils/src/tx_indexer.rs
```diff
@@ -0,0 +1,350 @@
...
+
+    for (i, info) in tx_entries.iter().flat_map(ops_extractor).enumerate() {
+        if let ProtocolOperation::Deposit(deposit_info) = info {
+            assert_eq!(
+                deposit_info.address,
+                if i == 0 {
+                    dest_addr1.clone()
+                } else {
+                    dest_addr2.clone()
+                },
```
delbonis, line 87: Can avoid clones here, just do `&deposit_info.address`.  Not really an efficiency thing but a cleanliness thing.

crates/test-utils/src/tx_indexer.rs
```diff
@@ -0,0 +1,350 @@
...
+            panic!("test: expected DepositInfo for transaction {i}");
+        }
+    }
+
+    tx_entries
+}
+
+pub fn test_index_no_deposit_with_visitor<V>(
+    visitor: impl Fn() -> V,
+    _: fn(&Indexed<V::Output>) -> Vec<ProtocolOperation>,
```
delbonis, line 110: Probably should be `impl Fn` so that it's more general, we don't care about it being just a fn ptr.  Same in the other test fns here.

delbonis (changes requested):
Glad to see a bunch of this code removed.  Just some cleanliness changes.

crates/l1tx/src/utils.rs
```diff
@@ -139,4 +142,36 @@ pub mod test_utils {
...
+    pub fn get_filter_config_from_deposit_entries(
+        params: Params,
+        deps: Vec<DepositEntry>,
+    ) -> TxFilterConfig {
+        let mut filterconfig = TxFilterConfig::derive_from(params.rollup()).unwrap();
+        // Watch all withdrawals that have been ordered.
+        let exp_fulfillments = deps
+            .iter()
+            .flat_map(conv_deposit_to_fulfillment)
+            .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
bewakes: Nope, not actually. :) Updated.

crates/test-utils/src/tx_indexer.rs
bewakes: That's a good point. Updated.

delbonis (approved):

MdTeach (dismissed):

storopoli (dismissed):
ACK 8b4788f4bbba62f823e79a3dceab88d5b138f3ff

Rajil1213 (approved):

Rajil1213:
Looks like CI is stuck on this PR. I'm closing and re-opening this ticket to retrigger it. 

(CI was stuck in one of the checks in an earlier commit which I've since canceled but it's still stuck).

Rajil1213 (approved):

PR #805, opened 2025-05-14
STR-1090 Encapsulate OL execution in new ChainExecutor

delbonis:
## Description

This PR revolves around two new pieces, the chain executor and the exec worker.  The chain executor encapsulates the mid-level logic involved in executing blocks and is like a wrapper around the low-level chaintsn crate's work.  The exec worker is a service that will manage the chainstate on disk and provide a coherent interface of high level operations to it.  The FCM and whatever other future sync workers will call it via a opaque handle.

We achieve this by providing more structure to how the state transition function accesses state and auxiliary inputs.  There are a few new traits that we will be converting existing code over to use for accessing state and other entries on-disk like manifests.  These establish data flows that will eventually be used when we have DA, ledger accounts, ASM updates, etc.  In the meantime, we do have to make some provisions for legacy issues when we move more of the new pieces into place.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


crates/state/src/context.rs
delbonis, line 1: Is there a reason this needed to be moved to the state crate?  It's really a part of the chain exec logic.  A stumbling block we had before was that we were trying to make these context traits overly general (like `L2Header`) and it became cumbersome to work with (to me), so I was putting it closer to where it would be used.

prajwolrg: My bad. I don't know why I ported it over there. Fixed it.

crates/chainexec/src/executor.rs
```diff
@@ -0,0 +1,95 @@
...
+        "chainexec: nonsensical post-state epoch (pre={pre_state_epoch}, post={post_state_epoch})"
+    );
+
+    // Verify state root matches.
+    // TODO move this check somewhere else where we have more context
+    /*
+        if *header.state_root() != computed_sr {
+            warn!(block_sr = %header.state_root(), %computed_sr, "state root mismatch");
+            Err(Error::StateRootMismatch)?
+    }*/
```
delbonis, line 108: Leaving a note on this so I remember to go back and make sure we do this in the right places.

crates/chainexec/src/validation_util.rs
```diff
@@ -0,0 +1,30 @@
...
+    prelude::*,
+};
+use tracing::*;
+
+/// Considers if the block is plausibly valid and if we should attach it to the
+/// pending unfinalized blocks tree.  The block is assumed to already be
+/// structurally consistent.
+// TODO remove FCM arg from this
+#[allow(dead_code)]
+pub fn check_new_block(_blkid: &L2BlockId, block: &L2Block, params: &RollupParams) -> bool {
```
delbonis: `// TODO rework this fn, make it make more sense with the limited scope`

crates/chain-worker/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "strata-chain-worker"
+version = "0.1.0"
+edition = "2024"
+
```
storopoli, line 5:
```suggestion

[lints]
workspace = true

```

crates/chainexec/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "strata-chainexec"
+version = "0.1.0"
+edition = "2024"
+
```
storopoli, line 5:
```suggestion

[lints]
workspace = true

```

storopoli (changes requested):

crates/consensus-logic/src/sync_manager.rs
delbonis, line 1: Correct me if I'm wrong but these two new "spawn" fns are actually the base fns in the new threads, not spawn fns which would spawn a new task and return to the caller.

crates/consensus-logic/src/sync_manager.rs
```diff
@@ -134,3 +172,74 @@ pub fn start_sync_tasks<E: ExecEngineCtl + Sync + Send + 'static>(
...
+    Ok(())
+}
+
+fn spawn_chain_worker(
+    shutdown: ShutdownGuard,
+    handle: Handle,
+    storage: Arc<NodeStorage>,
+    status_channel: StatusChannel,
+    params: Arc<Params>,
+    exect_ctl_handle: ExecCtlHandle,
```
delbonis: Typo?

crates/status/src/status_manager.rs
```diff
@@ -189,6 +189,15 @@ impl StatusChannel {
         Ok(state.clone())
     }
 
+    /// Waits until declared and returns the client state where genesis was triggered.
+    pub async fn wait_until_declared_epoch_is_some(&self) -> Result<ClientState, RecvError> {
```
delbonis: A `RecvError` here just means that the sender side was dropped without sending a value, it's not an exceptional condition, we should either gobble the error and return `None` or convert it to our own error type with a context-aware meaning.

crates/consensus-logic/src/sync_manager.rs
```diff
@@ -134,3 +172,74 @@ pub fn start_sync_tasks<E: ExecEngineCtl + Sync + Send + 'static>(
...
+    let chain_exec = ChainExecutor::new(params.rollup().clone());
+    let shared = Arc::new(Mutex::new(WorkerShared::default()));
+    let state = strata_chain_worker::init_worker_state(
+        shared.clone(),
+        context,
+        chain_exec,
+        exect_ctl_handle,
+        cur_tip,
+    )?;
+    let input = ChainWorkerInput::new(shared.clone(), chain_msg_rx);
```
delbonis: There's a lot of init logic happening here that belongs in the chain worker modules.  Like the `WorkerShared` struct is solidly an implementation detail that should not even be exported.  The ideal way to construct these would have been to have some function that spawns the worker in a new thread and returns the `ChainWorkerHandle`.  Then this handle would be passed to the FCM when setting that up.

I realize that maybe there's some constraints on needing to wait for genesis to happen that makes that more complicated.  It would be better to encapsulate this complexity and shifting it down into the worker task somehow, instead of exposing it.

crates/consensus-logic/src/sync_manager.rs
```diff
@@ -134,3 +172,74 @@ pub fn start_sync_tasks<E: ExecEngineCtl + Sync + Send + 'static>(
...
+        ),
+    };
+
+    let blkid = *cur_tip.blkid();
+    info!(%blkid, "starting exec worker");
+
+    let exec_env_id = ();
+    let state = ExecWorkerState::new(engine, exec_env_id, cur_tip, cur_tip);
+    let ctx = ExecWorkerCtx::new(storage.l2().clone());
+    exec_worker_task(shutdown, state, exec_rx, &ctx)?;
```
delbonis: I am not as sure of the ideal way for this task to work, but I believe there's a similar version of this exposing of the init process here as well.  I had not fully completed this exec worker client task, but I believe it should resemble the chain worker task where the caller just uses the opaque `ExecWorkerHandle`.

crates/consensus-logic/src/sync_manager.rs
```diff
@@ -4,20 +4,32 @@
...
+    chain_worker_context::ChainWorkerCtx,
     csm::{
         ctl::CsmController,
         message::{ClientUpdateNotif, CsmMessage, ForkChoiceMessage},
-        worker,
+        worker::{self},
     },
-    fork_choice_manager,
+    exec_worker_context::ExecWorkerCtx,
+    fork_choice_manager::{self},
```
delbonis: These imports didn't get simplified for some reason.

bin/strata-client/src/el_sync.rs
delbonis, line 1: I am getting more uncomfortable with this very non-trivial sync impl existing in the main binary crate, we should probably move this elsewhere.  Can we make parts of it part of the exec client worker, maybe needing some new message types?  That feels like the most appropriate place for it.

crates/consensus-logic/src/sync_manager.rs
```diff
@@ -83,6 +95,8 @@ pub fn start_sync_tasks<E: ExecEngineCtl + Sync + Send + 'static>(
     // Create channels.
     let (fcm_tx, fcm_rx) = mpsc::channel::<ForkChoiceMessage>(64);
     let (csm_tx, csm_rx) = mpsc::channel::<CsmMessage>(64);
+    let (chain_msg_tx, chain_msg_rx) = mpsc::channel::<ChainWorkerMessage>(64);
+    let (exec_tx, exec_rx) = strata_eectl::handle::make_handle_pair();
```
delbonis: See other comments.  I know this pattern exists literally in the same space but we can try to avoid perpetuating it.

bin/strata-client/src/el_sync.rs
```diff
@@ -28,43 +24,58 @@ pub(crate) fn sync_chainstate_to_el(
...
+    debug!(%tip_block, "L2 tip block");
 
-    info!(%earliest_idx, %latest_idx, "search for last known idx");
+    let latest_header = l2_block_manager
+        .get_block_data_blocking(&tip_block)?
+        .ok_or(DbError::MissingL2Block(tip_block))?
+        .header()
+        .clone();
+    let latest_idx = latest_header.slot();
+    info!(%latest_idx, "search for last known idx");
```
delbonis: `searching`?  and is talking about like an EL block number/slot?

delbonis (commented):
@prajwolrg Okay I've finally been able to sit down and review this.

There's a lot of good stuff put together here but the main remaining issue is leaking initialization logic impl details from the worker tasks.  It's one of the smells we do currently have with other tasks that I would like to avoid making worse.

bin/strata-client/src/el_sync.rs
prajwolrg: yes

crates/status/src/status_manager.rs
prajwolrg: Good catch! This was meant to be a temporary fix. I had forgotten to remove it.

crates/consensus-logic/src/sync_manager.rs
prajwolrg: Yes, you're correct. The name seems to be misleading.

bin/strata-client/src/el_sync.rs
prajwolrg: Moved it to `exec_worker_task`.

crates/chain-worker/src/worker.rs
```diff
@@ -0,0 +1,362 @@
...
+    let cur_tip = wait_for_genesis_and_resolve_tip(&status_channel, &handle)?;
+
+    let blkid = *cur_tip.blkid();
+    info!(%blkid, "starting chain worker");
+
+    // Create the chain executor
+    let chain_exec = ChainExecutor::new(params.rollup().clone());
+
+    // Initialize shared state and worker state
+    let shared = Arc::new(Mutex::new(WorkerShared::default()));
```
delbonis: This shared struct needs to be shared with the handle.  Currently we're constructing two separate instances of it.  That might have been my mistake when I was reworking some of this.

Also, the name `spawn_..._internal` still doesn't make sense, it's not a spawn function.  It'd probably be better to call *this* one `worker_task` and the other fn `worker_task_inner`.

crates/consensus-logic/src/sync_manager.rs
prajwolrg: Updated this.

crates/eectl/src/sync.rs
```diff
@@ -0,0 +1,115 @@
...
+        assert_eq!(last_bundle.header().slot(), tip_idx);
+    }
+
+    for bundle in bundles_to_sync {
+        let tip_blockid = bundle.block().header().get_blockid();
+        let payload = ExecPayloadData::from_l2_block_bundle(&bundle);
+        debug!(?payload, "Submitting payload to engine");
+        engine.submit_payload(payload)?;
+        engine.update_safe_block(tip_blockid)?;
+    }
```
delbonis: This kinda circumvents what I'm trying to set up with the new exec worker context trait by directly fetching blocks.  We can fetch the `ExecPayloadData` from the context trait instead of reconstructing it out of the block.  I don't see a reason we'd have to go block-by-block.  Is there some reason I can't see?

crates/consensus-logic/src/sync_manager.rs
```diff
@@ -134,3 +151,51 @@ pub fn start_sync_tasks<E: ExecEngineCtl + Sync + Send + 'static>(
         status_channel,
     })
 }
+
+fn spawn_exec_worker_2<E: ExecEngineCtl + Sync + Send + 'static>(
```
delbonis: _2?

crates/db/src/traits.rs
```diff
@@ -177,7 +182,7 @@ pub enum BlockStatus {
 /// For now, the full state is just the "toplevel" state that can always be
 /// expected to be of moderate size in memory.
 // TODO maybe rewrite this around storing write batches according to blkid?
-pub trait ChainstateDatabase {
+pub trait OldChainstateDatabase {
```
delbonis: We should remove this trait and all the stuff built on top of it, since we have supersceded it.

crates/eectl/src/sync.rs
prajwolrg: AFAIU, we have to go block-by-block to get to the current EVM EE State. Since we don't have a way to go to the current EE State using DA, we need to execute all the transactions in EE to get to the latest state. 

We can fetch `ExecPayloadData` directly, but we would still require getting L2BlockManager to get all the blockids for `ExecPayloadData`.

crates/consensus-logic/src/exec_worker_context.rs
```diff
@@ -36,7 +36,19 @@ impl ExecWorkerContext for ExecWorkerCtx {
...
-    fn get_cur_tip(&self) -> EngineResult<L2BlockCommitment> {
+    fn fetch_parent(&self, block: &L2BlockCommitment) -> EngineResult<L2BlockCommitment> {
+        let parent_blk = self
+            .l2man
+            .get_block_data_blocking(block.blkid())?
+            .ok_or(DbError::MissingL2Block(*block.blkid()))?;
+        let parent_blkid = parent_blk.header().parent();
+        // REVIEW: Check if simply doing -1 is enough or we want to fetch the parent block again and
+        // get the slot from there in case the slot are empty
+        let parent_slot = block.slot() - 1;
```
delbonis: We should be able to just fetch the header.

prajwolrg (approved):

delbonis:
Merging since all green and reviews resolved.

PR #806, opened 2025-05-15
Str 1289 asm base

prajwolrg:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


crates/asm/src/state.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub pow_state: HeaderVerificationState,
+
+    /// Events emitted by subprotocols, keyed by L1 block:
+    /// - The outer Vec is ordered by block ID.
+    /// - Each inner Vec contains tuples of `(subprotocol_id, event_hash)`.
+    ///
+    /// `subprotocol_id` (u8) identifies which subprotocol emitted the event,
+    /// and `Buf32` is the protocol‐computed hash of the event payload.
+    // TODO: Eventually this will use an MMR for minimal, non‐linear‐growing on‐chain state.
+    pub events: Vec<(L1BlockId, Vec<(u8, Buf32)>)>,
```
delbonis, line 45: This isn't needed in the ASM state.  We only care about headers and PoW in the chain view state.

crates/asm/src/subprotocol/mod.rs
delbonis, line 1: Each subprotocol should be its own crate so that we don't introduce unintentional interdependencies within the system.

crates/asm/src/subprotocol/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn id(&self) -> u8 {
+        Self::VERSION
+    }
+
+    /// Process the L1Block and extracts all the relevant information from L1 for the subprotocol
+    ///
+    /// Update it's own state and output a list of InterProtoMsg addressed to other subprotocols
+    fn process_block(&mut self, _block: &Block) -> Vec<(u8, InterProtoMsg)> {
+        vec![]
+    }
```
delbonis, line 35: We don't need a `process_block` function that takes the whole block, that introduces more complexity since we'd be parsing data out of the block in multiple places.  The subprotocol only sees txs that have its subprotocol ID.

Also, default impls are intended more for cases where it's a trivial extension of the existing logic like:

```rs
fn try_get_something(&self) -> Result<Option<T>, Error>;

fn expect_get_something(&self) -> T {
    self.try_get_something().unwrap().unwrap()
}
```

Or for how `clone_from` works.

crates/asm/src/subprotocol/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// 1-byte subprotocol identifier / version tag (matches SPS-50).
+    const VERSION: u8;
+
+    /// Returns the identifier of the subprotocol for this section.
+    ///
+    /// This ID corresponds to the version or namespace of the subprotocol whose
+    /// state is serialized in this section.
+    fn id(&self) -> u8 {
+        Self::VERSION
+    }
```
delbonis, line 28: This is redundant, isn't it?

crates/asm/src/subprotocol/core.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    /// Summary of the last checkpoint that was successfully verified.
+    /// New proofs are checked against this epoch summary.
+    verified_checkpoint: EpochSummary,
+
+    /// The L1 block ID up to which the `verified_checkpoint` covers.
+    last_checkpoint_ref: L1BlockId,
+
+    /// Public key of the batch producer authorized to submit checkpoint proofs.
+    batch_producer_pubkey: Buf32,
```
delbonis, line 35: I guess should be `sequencer_pubkey` or `checkpoint_signing_pubkey`.

Really we could use more sophisticated signalling here.

crates/asm/src/subprotocol/core.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    /// Administrator public key. Holds the authority to update the
+    /// `batch_producer_pubkey` via a Bitcoin transaction.
+    administrator: Buf32,
+
+    /// Public key of the consensus manager, responsible for higher‐level
+    /// governance and protocol parameter updates.
+    consensus_manager: Buf32,
```
delbonis, line 43: This is premature, we haven't specced any of this out yet.

crates/asm/src/subprotocol/core.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+impl TryFrom<&SectionState> for CoreASMState {
+    type Error = ASMError;
+
+    fn try_from(section: &SectionState) -> Result<Self, Self::Error> {
+        CoreASMState::try_from_slice(&section.data)
+            .map_err(|e| ASMError::Deserialization(section.id(), e))
+    }
+}
```
delbonis, line 71: This `to_section_state` fn and `TryFrom` impl should be generic on the `Subprotocol`.  The `State` being an associated type lets us avoid having to do any of this boilerplate here.

crates/asm/src/msg.rs
```diff
@@ -0,0 +1,23 @@
...
+}
+
+/// Describes all the messages that can originate when processing ProtocolOps by a subprotocol that
+/// needs to be consumed by other subprotocols
+#[derive(Clone, Debug, Eq, PartialEq, BorshSerialize, BorshDeserialize, Serialize, Deserialize)]
+pub enum InterProtoMsg {
+    /// This message originates from the Core OL Subprotocol and is meant to be passed to the
+    /// Bridge Subprotocol
+    Withdrawal(WithdrawalIntent),
+}
```
delbonis, line 23: This is introducing coupling by making the general ASM infrastructure implicitly dependent on the subprotocols that we intend to use it with.  We can really leverage the type system here to avoid this.  See the discussion about `dyn` for an imperfect but better solution.

Define some kind of `SubprotoMsgRouter` type that abstracts over this low level bookkeeping, and then pass in a `&mut` into the process function instead of making it return the messages.  It allows for more abstraction over the types.

crates/asm/src/state.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub fn subprotocols(&self) -> Vec<impl Subprotocol> {
+        let mut protocols = Vec::with_capacity(self.sections.len());
+        for section in &self.sections {
+            // Ignore any sections that fail to deserialize
+            if let Ok(proto) = section.subprotocol() {
+                protocols.push(proto);
+            }
+        }
+        protocols
+    }
```
delbonis, line 74: Using `impl Subprotocol` here means that this function gets specialized to a single type.  It would only valid for this ever to ever have a single value.  However, this can very easily break because we might have two subprotocols that when serialized coincidentally can deserialize into each other and get completely garbage data!

crates/asm/src/state.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    /// Attempts to deserialize this section into its corresponding subprotocol state machine.
+    ///
+    /// Uses the section’s `subprotocol_id` to select and construct the appropriate
+    /// `Subprotocol` implementation via `TryFrom<SectionState>`.
+    pub fn subprotocol(&self) -> Result<impl Subprotocol, ASMError> {
+        match self.id() {
+            CORE_SUBPROTOCOL_ID => CoreASMState::try_from(self),
+            _ => Err(ASMError::InvalidSubprotocol(self.id())),
+        }
```
delbonis, line 94: This not having a verb-y name implies that it's a simple accessor, but it's actually doing a fallible deserialization.  So on its own it should be `to_subproto_state`.

This is actually the source of the issue in the `subprotocols` function.  It's not manifesting here because there only is a single impl for `Subprotocol` that exists in the body of the function, but it would break if you tried to implement the bridge subproto.

It would make more sense for this to be defined as `fn try_parse_subproto_state<T: Subprotocol>(&self) -> Result<T::State, Error>` and *only* attempt deserializing as that one subproto's state type.

crates/asm/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "asm"
```
delbonis, line 2: Maybe crate names would be like `strata-asm-common`, `strata-asm-proto-core`, `strata-asm-proto-bridge-v1`, `strata-asm-impl`?

delbonis (changes requested):

crates/asm/src/state.rs
prajwolrg: Should this be part of OL subprotocol state? Because for the validation of OL STF, we need to verify that OL has applied all the relevant L1 information like deposits in the correct way. The extraction of this relevant information is done as part of ASM STF. Since OL STF can include range of L1Blocks and lags behind in L1Reference, we need to support verifying that the relevant L1 information that was used by OL, is indeed the one that were extracted as part of ASM STF.

crates/asm/stf/src/tx_filter.rs
```diff
@@ -0,0 +1,65 @@
...
+/// `Transaction`.
+///
+/// The SPS-50 header MUST be encoded as an `OP_RETURN` in output index 0, with payload:
+/// ```text
+/// [0..4]   ASCII magic “ALPN”
+/// [4]      subprotocol type (u8)
+/// [5]      tx type (u8)
+/// [6..]    auxiliary data (ignored here)
+/// ```
+fn parse_sps50_header(tx: &Transaction) -> Option<(SubprotocolId, u8)> {
```
delbonis, line 16: We should put this in a separate crate (probably in the common repo) with other libraries to do all of the transaction parsing logic.  Probably not `l1tx` since that's the current ad-hoc system.  As-is it either means that we'd have encoding and decoding separated (bad for repo organization) the bridge repo would have to strongly depend on this ASM-related crate in the chain client repo (bad for dependency management).

crates/asm/subprotocols/bridge-v1/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+impl From<BridgeV1StateStub> for SectionState {
+    fn from(state: BridgeV1StateStub) -> Self {
+        let data =
+            borsh::to_vec(&state).expect("BorshSerialize on BridgeV1StateStub should never fail");
+        SectionState {
+            id: BRIDGE_V1_SUBPROTOCOL_ID,
+            data,
+        }
+    }
+}
```
delbonis, line 44: Having `from_section` as part of the trait and the `from` not is asymmetrical, which is a red flag.  If we just define some common codec system (you *could* use the borsh traits here, but also see the `Codec` type I defined in the OL DA PR) and add it as an associated type constraint on `State` then this becomes much cleaner to deal with.

delbonis (changes requested):

crates/asm/stf/src/subprotocols.rs
```diff
@@ -0,0 +1,16 @@
...
+/// Unknown protocols are simply skipped.
+pub(crate) fn parse_subprotocols(sections: &[SectionState]) -> Vec<Box<dyn Subprotocol>> {
+    sections
+        .iter()
+        .filter_map(|sec| match sec.id {
+            CORE_SUBPROTOCOL_ID => CoreASMState::from_section(sec).ok(),
+            BRIDGE_V1_SUBPROTOCOL_ID => BridgeV1StateStub::from_section(sec).ok(),
+            _ => None,
+        })
+        .collect()
```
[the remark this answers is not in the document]
delbonis: That code there probably won't directly solve our problems, but I can imagine it *might* get us part of the way there.

crates/asm/src/state.rs
prajwolrg: As part of OL STF, does it references both the L1Reference and the ASM State? i.e. Does subprotocols have ability to introspect arbitrary history of both L1Headers and ASM State?

crates/asm/stf/src/subprotocols.rs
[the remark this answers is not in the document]
prajwolrg: I don't understand how it might get us part of the way there. I understand we can use enum of subprotocols if we want to avoid `dyn` and maybe add some macros to make things less verbose with using enums, but I don't understand how it might work with `SubprotocolStack`.

crates/asm/stf/src/tx_filter.rs
prajwolrg: Agree! Adding a TODO and leaving it as is for now.

crates/asm/subprotocols/bridge-v1/src/lib.rs
prajwolrg: Thanks, made things symmetrical by adding `try_from_section` and `to_section` on the trait.

When using the associated type on the Subprotocol trait, the trait becomes dyn-incompatible, so haven't used that in the trait. Not sure how should I proceed forward.

crates/asm/stf/src/tx_filter.rs
```diff
@@ -0,0 +1,65 @@
...
+///
+/// Transactions that lack a valid SPS-50 header (wrong magic, not OP_RETURN in output[0],
+/// or too-short payload) are filtered out.
+pub(crate) fn group_txs_by_subprotocol<T>(
+    transactions: T,
+) -> HashMap<SubprotocolId, Vec<Transaction>>
+where
+    T: IntoIterator<Item = Transaction>,
+{
+    let mut map: HashMap<SubprotocolId, Vec<Transaction>> = HashMap::new();
```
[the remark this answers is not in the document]
prajwolrg: I used `HashMap` to avoid sorting.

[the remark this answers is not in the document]
delbonis: It's fewer allocations since you don't need to allocate to construct the hash tree nodes.  But I suppose it's fine since I don't expect to have more than a few entries here.

crates/asm/subprotocols/bridge-v1/src/lib.rs
delbonis: The `Subprotocol` type is not the state type, there's no need to ever dyn it.

delbonis:
I'll make a commit later to reflect what's in the notion doc.

delbonis:
Okay take a look over on that.  It does rely on `dyn` a bit more than I wish it did but the number of vtable calls is linear in the number of inter-proto messages being sent, which we expect to actually be pretty uncommon in practice.

prajwolrg:
@delbonis I tried to implement `asm_stf` function with the new changes.

crates/asm/src/state.rs
prajwolrg: I still don't understand.

When processing L1Block N, all the subprotocols get access to:
1. ASM State N-1
2. Arbitrary access to L1BlockId

> If we just check that the checkpoint attests to a correct recent ASM state then we don't have to do anything like this, the checkpoint proof checks that. We can completely black-box it.

When creating the checkpoint proof, it references a recent ASM State (say K), but when the checkpoint proof is settled on Bitcoin, it might settle on L1Block K+5. In that case how will the checkpoint proof be verified?

delbonis (changes requested):
I have a quick idea I want to try here to maybe avoid an extra clone.

crates/asm/src/state.rs
prajwolrg: We decided to store a single hash of all the logs we got after processing all the subprotocols.

crates/asm/common/src/state.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub chain_view: ChainViewState,
+
+    /// States for each subprotocol section, sorted by Subprotocol Version/ID.
+    pub sections: Vec<SectionState>,
+}
+
+impl AnchorState {
+    /// Gets a section by protocol ID by doing a linear scan.
+    pub fn find_section(&self, id: SubprotocolId) -> Option<&SectionState> {
+        self.sections.iter().find(|s| s.id == id)
```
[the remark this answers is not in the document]
prajwolrg: For few protocols, sanity checks might be more expensive than the linear search. Leaving it as is for now.

crates/asm/stf/src/transition.rs
```diff
@@ -0,0 +1,58 @@
...
+
+/// Computes the next AnchorState by applying the Anchor State Machine (ASM) state transition
+/// function (STF) to the given previous state and new L1 block.
+pub fn asm_stf<S: AsmSpec>(pre_state: AnchorState, block: Block) -> AnchorState {
+    let mut pow_state = pre_state.chain_view.pow_state.clone();
+
+    // 1. Validate and update PoW header continuity for the new block
+    pow_state
+        .check_and_update_continuity(&block.header, &Params::MAINNET)
+        .expect("header doesn't follow the consensus rules");
```
sapinb, line 34: `asm_stf` is also going to run in csm besides prover, cannot panic on invalid input. Unless we expect this to have redundant checks in caller and here both ?

delbonis: For this yeah, since the panic is input-dependent, we should error more gracefully.

crates/asm/common/src/subprotocol.rs
delbonis, line 1: These tx input related types should be moved to a new module since they're getting kinda large.  Also the `struct` blocks and `impl` blocks are misordered.

crates/asm/common/src/subprotocol.rs
```diff
@@ -0,0 +1,129 @@
...
+/// 1. Byte 0: subprotocol-defined transaction type
+/// 2. Bytes 1…: auxiliary payload (type-specific)
+#[derive(Debug)]
+pub struct Sps50TagPayload<'p> {
+    /// The transaction type as defined by the SPS-50 subprotocol.
+    tx_type: u8,
+
+    /// The remaining, type-specific payload for this transaction.
+    auxiliary_data: &'p [u8],
+}
```
delbonis, line 88: We can just call this `tag` and `TagPayload`.  Or maybe "protocol tag"?

crates/asm/common/src/subprotocol.rs
```diff
@@ -0,0 +1,129 @@
...
+    /// Gets the inner transaction.
+    pub fn tx(&self) -> &Transaction {
+        self.tx
+    }
+
+    /// Returns a reference to the parsed SPS-50 payload for this transaction,
+    /// which contains the subprotocol-specific transaction type and auxiliary data.
+    pub fn sps50_payload(&self) -> &Sps50TagPayload<'t> {
+        &self.sps_50_payload
+    }
```
delbonis, line 128: As above, `tag`.

crates/asm/stf/src/transition.rs
```diff
@@ -0,0 +1,58 @@
...
+    S::call_subprotocols(&mut loader_stage);
+
+    // 4. PROCESS: feed each subprotocol its slice of txs
+    let mut process_stage =
+        ProcessStage::new(all_relevant_transactions, loader_stage.into_handler());
+    S::call_subprotocols(&mut process_stage);
+
+    // 5. FINISH: let each subprotocol process its buffered interproto messages
+    let mut finish_stage = FinishStage::new(process_stage.into_handler());
+    S::call_subprotocols(&mut finish_stage);
```
delbonis, line 50: Instead of having to do a `into_handler` each time, maybe we should have it on the outside and pass in a `&mut ref` to it to the stage?

crates/asm/stf/src/stage.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+impl FinishStage {
+    pub(crate) fn new(handler: HandlerRelayer) -> Self {
+        let sections = Vec::new();
+        Self { handler, sections }
+    }
+
+    pub(crate) fn into_sections(self) -> Vec<SectionState> {
+        self.sections
```
delbonis, line 87: We have to assert that these are sorted by subprotocol when we return them.

crates/asm/common/src/state.rs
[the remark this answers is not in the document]
delbonis: We're doing the linear search multiple times, but we'd only do the sanity checks once, and we could even do them *while deserializing* if that was a concern.  But yes, n is unlikely to ever be >3.

delbonis (changes requested):
Mostly just cleanup at this point.

crates/asm/stf/src/transition.rs
```diff
@@ -36,18 +37,19 @@ pub fn asm_stf<S: AsmSpec>(pre_state: AnchorState, block: Block) -> AnchorState
...
-        ProcessStage::new(all_relevant_transactions, loader_stage.into_handler());
-    S::call_subprotocols(&mut process_stage);
+    let mut process_stage = ProcessStage::new(all_relevant_transactions);
+    S::call_subprotocols(&mut process_stage, &mut manager);
 
     // 5. FINISH: let each subprotocol process its buffered interproto messages
-    let mut finish_stage = FinishStage::new(process_stage.into_handler());
-    S::call_subprotocols(&mut finish_stage);
+    let mut finish_stage = FinishStage::new();
+    S::call_subprotocols(&mut finish_stage, &mut manager);
```
delbonis: Ahhh noo!

I meant

```rs
struct FinishStage<'m> {
  manager: &'m mut SubprotoManager,
  // ...
}

// so you can do
let mut manager = ...;

let mut loader_stage = LoaderStage::new(&mut manager);
S::call_subprotocols(...);

let mut process_stage = ProcessStage::new(&mut manager);
S::call_subprotocols(...);

let mut finish_stage = FinishStage::new(&mut manager);
S::call_subprotocols(...);
```

This hides the manager from the spec.  Maybe there's some case where passing the manager in would be useful?  I'm not sure.

And then you could make constructing the final state be done in part by the manager at the end instead of rolling it into a stage?  Not sure if that makes it cleaner or not.

delbonis (changes requested):
I should have been more clear, my mistake.

bewakes (commented):
This is good! The ASM stf with staged processing is neat. A few more comments.

crates/asm/common/src/state.rs
```diff
@@ -0,0 +1,78 @@
...
+/// receiving protocol transactions at L1 and updating its storage. A zk-SNARK proof
+/// attests that the transition from the previous ASM state to the new state
+/// was performed correctly on the given L1 block.
+#[derive(Clone, Debug)]
+pub struct AnchorState {
+    /// The current view of the L1 chain required for state transitions.
+    pub chain_view: ChainViewState,
+
+    /// States for each subprotocol section, sorted by Subprotocol Version/ID.
+    pub sections: Vec<SectionState>,
```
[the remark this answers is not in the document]
prajwolrg: I don't really have much opinion here. We can implement that as well. As of now things everything works fine even though they are not sorted. CC: @delbonis.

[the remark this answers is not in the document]
bewakes: It's fine if the vector doesn't need to be sorted. Just pointing out that if it needs to be sorted there's `SortedVec` in `strata_primitives` which might fit the use case.

crates/asm/common/src/subprotocol.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// Message type that we receive messages from other subprotocols using.
+    type Msg: Clone + Any;
+
+    /// Constructs a new state to use if the ASM does not have an instance of it.
+    fn init() -> Self::State;
+
+    /// Process the transactions and extract all the relevant information from L1 for the
+    /// subprotocol
+    ///
+    /// Update it's own state and output a list of InterProtoMsg addressed to other subprotocols
```
bewakes, line 40: A little confusion here. When you say output a list of `InterProtoMsg`, it indicates that this method is supposed to return `InterProtoMsg`. As I understand it, the msgs are processed by the relayer right?

prajwolrg: Yes, updated the doc.

crates/asm/common/src/state.rs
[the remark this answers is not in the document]
delbonis: But like this is fine and we'll be converting it to SSZ when that tooling is ready, where we wouldn't be able to use those types anyways.

delbonis (approved):

crates/asm/common/src/error.rs
```diff
@@ -1,14 +1,29 @@
+use std::fmt::{Debug, Display};
+
 use strata_primitives::l1::L1VerificationError;
 use thiserror::Error;
 
 use crate::SubprotocolId;
 
+/// A generic “expected vs actual” error.
+#[derive(Debug, Error)]
+#[error("expected {expected}, found {actual}")]
```
delbonis, line 10: Oh you can do this???

prajwolrg: Oh yes. ChatGPT suggested this.

PR #809, opened 2025-05-19
Str 1275 exp backoff for engine calls

irnb:
## Description
Today any *engine* connection failure inside block execution leads to:

1. The block being tagged **Invalid**.
2. The Sync crate advancing past that height (see Issue 1).

A transient network hiccup should trigger **retry with back-off** or, after `N` failures, a *graceful shutdown*, **not** a permanent Invalid flag. 

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

**Why is the retry_with_backoff + graceful shutdown strategy sufficient?**
[Notion Link](
https://www.notion.so/Sync-Crate-Analysis-1f6901ba000f80418a35ed25eda463b7?pvs=4#1f9901ba000f8081b7c9fbaeb91f0c8e)

## Checklist

## Related Issues

STR-1275

### DIFF
--- a/crates/sequencer/src/block_template/block_assembly.rs
+++ b/crates/sequencer/src/block_template/block_assembly.rs
@@ -1,5 +1,8 @@
 use std::{thread, time};
 
+use strata_common::retry::{
+    policies::ExponentialBackoff, retry_with_backoff, DEFAULT_ENGINE_CALL_MAX_RETRIES,
+};
 use strata_consensus_logic::checkpoint_verification;
 use strata_db::DbError;
 use strata_eectl::{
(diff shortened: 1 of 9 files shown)


crates/sequencer/src/block_template/block_assembly.rs
```diff
@@ -311,6 +311,8 @@ fn prepare_exec_data<E: ExecEngineCtl>(
         remaining_gas_limit,
     );
 
+    // TODO: should we use retry with backoff in here too?
+    // IMO yes, but we don't need graceful shutdown for it
```
irnb, line 315: @bewakes WDYT? I don’t think we need a dedicated graceful-shutdown path here. prepare_payload only submits a payload-creation task to the execution client—if it fails, no real harm done; we’ll just retry in the next iteration. And if the issue escalates, the fork-choice manager already has graceful-shutdown logic to handle it.

bewakes: Yes, I agree. There's no harm done if this submission fails. So yeah, we don't necessarily need a graceful shutdown here. But retry with backoff would still be a good safety net here.

delbonis (approved):

PR #814, opened 2025-05-21
STR-1401 faucet signet leads to 404

krsnapaudel:
## Description

In the case of signet, extra / between base url and `claim_l1` was causing the 404 error.
Claim to signet address was failing because of "strata" to "alpen" renaming in alpen CLI. The fix is part of [this PR](https://github.com/alpenlabs/faucet-api/pull/69)

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/bin/alpen-cli/src/cmd/faucet.rs
+++ b/bin/alpen-cli/src/cmd/faucet.rs
@@ -181,7 +181,7 @@ pub async fn faucet(
     let url = format!(
         "{base}{}/{}/{}",
         claim,
-        encode(&solution.to_be_bytes()),
+        encode(&solution.to_le_bytes()),
         address
     );
     let res = client
--- a/bin/alpen-cli/src/cmd/faucet.rs
+++ b/bin/alpen-cli/src/cmd/faucet.rs
@@ -220,10 +220,8 @@ fn pow_valid(mut hasher: Sha256, difficulty: u8, solution: Solution) -> bool {
 
 /// Ensures that the URL has a trailing slash.
 fn ensure_trailing_slash(mut url: Url) -> Url {
-    if !url.path().ends_with('/') {
-        let new_path = format!("{}/", url.path().trim_end_matches('/'));
-        url.set_path(&new_path);
-    }
+    let new_path = format!("{}/", url.path().trim_end_matches('/'));
+    url.set_path(&new_path);
     url
 }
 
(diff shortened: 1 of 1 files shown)


storopoli (changes requested):
We should canonicalize the `base` URL to remove trailing `/`.
If it ends with `/` we should remove it and then all additions to that URL will start with a `/`, e.g. `/foo/bar/baz`.

barakshani (approved):

krsnapaudel:
> We should canonicalize the `base` URL to remove trailing `/`. If it ends with `/` we should remove it and then all additions to that URL will start with a `/`, e.g. `/foo/bar/baz`.

The default behavior of 
```
let base = Url::from_str(&settings.faucet_endpoint)
        .user_error("Invalid faucet endopoint. Check the config file")?;
```
seems to add a slash at the end. Is there a problem with not removing and adjusting our paths?

krsnapaudel:
> We should canonicalize the `base` URL to remove trailing `/`. If it ends with `/` we should remove it and then all additions to that URL will start with a `/`, e.g. `/foo/bar/baz`.

I have added to code to add '/' if necessary.

bin/alpen-cli/src/cmd/faucet.rs
```diff
@@ -215,3 +217,32 @@ fn pow_valid(mut hasher: Sha256, difficulty: u8, solution: Solution) -> bool {
...
+        let fixed = ensure_trailing_slash(url);
+        assert_eq!(fixed.as_str(), "https://example.com/");
+    }
+
+    #[test]
+    fn leaves_trailing_slash_when_present() {
+        let url = Url::parse("https://example.com/").unwrap();
+        let fixed = ensure_trailing_slash(url);
+        assert_eq!(fixed.as_str(), "https://example.com/");
+    }
```
storopoli, line 248: Nice with the tests!
Can we add one with `https://example.com//` to see if it passes?

storopoli (changes requested):

bin/alpen-cli/src/cmd/faucet.rs
krsnapaudel: Good catch. It did not. I have fixed `ensure_trailing_slash`.

storopoli (approved):
ACK 8d635a6ca13de4bfa5407bfd5ad8977646b09d88

This needs to be backported to `releases/0.2.0`

PR #817, opened 2025-05-22
(reader/query): Verify checkpoint before txfilter update

bewakes:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


bewakes:
I am a bit iffy about checking the chainstate finalized info for checking if checkpoint was accepted. Didn't find other minimal way to do this though.

crates/btcio/src/reader/query.rs
```diff
@@ -279,7 +283,56 @@ async fn poll_for_new_blocks<R: Reader>(
...
+        // Wait for some change to happen
+        tokio::select! {
+            _ = rx.changed() => {}
+            // NOTE: what timeout to wait for here?
+            _ = tokio::time::sleep(timeout_duration) => {
+                error!("Timed out waiting for chainstate update.");
+                return Err(anyhow!("reader:query:wait_for_checkpoint_accept: Timed out waiting for chainstate update"));
+            }
+        }
+    }
```
delbonis, line 335: Does it make sense to error out in the loop this way?

Make sure that this usage of `watch` is correct.  Perhaps `.wait_for` might be more applicable?  You could wrap a call to that in the timeout and probably not need the loop around all of this.

Also should be a more structured error type instead of `anyhow!`.

crates/btcio/src/reader/query.rs
```diff
@@ -279,7 +283,56 @@ async fn poll_for_new_blocks<R: Reader>(
...
+async fn wait_until_checkpoint_accepted_or_rejected<R: Reader>(
+    ctx: &ReaderContext<R>,
+    checkpoint: &SignedCheckpoint,
+) -> anyhow::Result<bool> {
+    let batch_info = checkpoint.checkpoint().batch_info();
+    let l2_commt = checkpoint.checkpoint().batch_info().l2_range.1;
+    let ckpt_epoch = batch_info.epoch();
+
+    let s = info_span!("wait_checkpoint_accepted", %ckpt_epoch, last_blkid=%l2_commt.blkid());
+    let _ = s.enter();
```
delbonis, line 301: Careful, `.enter()` interacts poorly with async code due to how future polling works.  The preferred way to do this is to wrap a future like `.instrument(span)`.

See: https://docs.rs/tracing/latest/tracing/span/struct.EnteredSpan.html#in-asynchronous-code

If you're unlucky then you end up entering the span many times and it breaks the logging output because you overflow buffers and spans go unclosed or something weird.

crates/btcio/src/reader/query.rs
```diff
@@ -258,16 +252,26 @@ async fn poll_for_new_blocks<R: Reader>(
...
                 if let Some(checkpt) = find_checkpoint_in_events(&evs) {
                     // if we have a checkpoint in this block, update filterconfig based on this
-                    let chainstate = borsh::from_slice(checkpt.checkpoint().sidecar().chainstate())
-                        .expect("deserialize chainstate");
+                    let accepted = wait_until_checkpoint_accepted_or_rejected(ctx, checkpt).await?;
+
+                    if accepted {
+                        let chainstate =
+                            borsh::from_slice(checkpt.checkpoint().sidecar().chainstate())
+                                .expect("deserialize chainstate");
```
delbonis, line 269: This shouldn't panic if it's malformed, even if the proof is (somehow) valid.

While we're here we should probably make this more organized, so that here it feels like you're just directly (but fallibly) taking the decoded sidecar field(s) out of the checkpoint structure, instead of needing to call borsh yourself here (especially since we'd like to move away from borsh and this removes one more place where we depend on it directly).

delbonis (changes requested):
Structurally I think this looks good, but there's a couple of points that could be cleaned up.

crates/btcio/src/reader/query.rs
bewakes: Using `.wait_for` is not suitable here because we not only need to wait for something to happen but also exit when something else happens. In this particular case we are waiting for checkpoint X at a height to be accepted but we should also wait until some other checkpoint at the height gets accepted and return boolean accordingly.

> Also should be a more structured error type instead of anyhow!.
Not sure if it needs to be more structured since this is not a library module and we just exit the client when the reader fails for some reason.

crates/btcio/src/reader/query.rs
bewakes: Ohh, didn't know that. Thanks.

> (ask me how I know 🙃)

Okay I'll bite. How do you know? 🙃

crates/primitives/src/batch.rs
```diff
@@ -222,6 +222,11 @@ impl Checkpoint {
     pub fn sidecar(&self) -> &CheckpointSidecar {
         &self.sidecar
     }
+
+    /// Deserialize sidecar data as chainstate. Returns `None` if cannot be deserialized.
+    pub fn sidecar_as_chainstate<T: BorshDeserialize>(&self) -> Option<T> {
+        borsh::from_slice(self.sidecar().chainstate()).ok()
+    }
```
delbonis, line 229: This `as` infix [implies](https://stackoverflow.com/questions/72996050/when-to-use-as-vs-to-vs-into-in-rust) that it's doing a trivial transformation and infallibly return the value.  In this case it's also not a verb name, so it implies it's a field accessor like `.sidecar()` does.

Not only is the function fallible, it's also gobbling the error which could be useful in a test or a consumer.  It should be named something like `try_parse_sidecar` and return a `Result` (not an anyhow result).

It's accessing the `.chainstate()` field, so I would expect that it'd return a `Chainstate`, but really it's relying on the caller to decide the type it gets parsed as.  This is weird because there's should only ever be one valid type to parse this as.  What we could do is add a def like `pub type SidecarChainstateData = Chainstate;` to declare upfront what type this field should be decoded as, and use that in the return type.  So we can change that later and any consumers that depend on it can manage those breakages locally.

crates/btcio/src/reader/query.rs
```diff
@@ -279,7 +299,116 @@ async fn poll_for_new_blocks<R: Reader>(
...
+                error!("Status channel closed unexpectedly: {}", e);
+                return Err(anyhow!("reader:query:wait_for_checkpoint_accept: Status channel closed: {}", e));
+            }
+        }
+        _ = tokio::time::sleep(config.chainstate_update_timeout) => {
+            error!("Timed out waiting for chainstate update after {:?}", config.chainstate_update_timeout);
+            return Err(anyhow!(
+                "reader:query:wait_for_checkpoint_accept: Timed out waiting for chainstate update after {:?}s",
+                config.chainstate_update_timeout
+            ));
```
delbonis, line 408: Structured errors instead of `anyhow!`.

Also, the status channel returns an error [if-and-only-if the sender is closed, meaning the upstream task is shutting down](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.changed).  It's not an error for us, we should just interpret it as a signal that we should exit too.

crates/btcio/src/reader/query.rs
```diff
@@ -279,7 +299,116 @@ async fn poll_for_new_blocks<R: Reader>(
...
+                warn!("Checkpoint not accepted: different block id for same slot finalized");
+                return Ok(false);
+            } else if l2_commt.slot() < tip_commitment.slot() {
+                // We've been looking for stale checkpoint to be accepted.
+                warn!("Checkpoint not accepted: higher slot finalized than the slot waited for");
+                return Ok(false);
+            }
+        }
+
+        wait_for_chainstate_change(config, &mut rx).await?;
```
delbonis, line 387: This control flow is really convoluted and indirect still.

The call to `wait_for_chainstate_change` has the information figure out what condition it is, but it discards it and we only use it as a way for us to signal to do another iteration on the loop.  Then in the next iteration we go and perform the checks to decide if we should either propagate up the timeout, get the value back out of the watch channel a second time and do checks, or just wait again.

It should return some thing like

```rs
enum QueryEvent {
  Update(ChainSyncStatusUpdate),
  Timeout,
  Exit // or make it return `Some(QueryEvent)`
}
```

and then pattern match on that to figure out what to do, then just wait on the function again.

crates/btcio/src/reader/query.rs
```diff
@@ -258,16 +258,36 @@ async fn poll_for_new_blocks<R: Reader>(
...
+                    if accepted {
+                        if let Some(chainstate) = checkpt.checkpoint().sidecar_as_chainstate() {
+                            state
+                                .filter_config_mut()
+                                .update_from_chainstate(&chainstate);
+                        } else {
+                            anyhow::bail!("Could not extract chainstate from checkpoint");
+                        }
+                    } else {
+                        warn!("Checkpoint did not get accepted");
```
delbonis, line 289: Should this really be a warning?  It's fine to have some lag.

crates/btcio/src/reader/query.rs
```diff
@@ -258,16 +258,36 @@ async fn poll_for_new_blocks<R: Reader>(
...
+                        .instrument(span)
+                        .await?;
+
+                    if accepted {
+                        if let Some(chainstate) = checkpt.checkpoint().sidecar_as_chainstate() {
+                            state
+                                .filter_config_mut()
+                                .update_from_chainstate(&chainstate);
+                        } else {
+                            anyhow::bail!("Could not extract chainstate from checkpoint");
```
delbonis, line 286: If `sidecar_as_chainstate` returned an error result then we wouldn't have to re-recognize the `None` condition and construct a new error.  We'd probably want to wrap it with additional context, but we could do that with `.map_err`.

delbonis (changes requested):

crates/btcio/src/reader/query.rs
delbonis: Well basically that.  And then the trace viewer I was using locked up for several seconds while updating the UI to display a trace that had several hundred (at least) nested spans, and then also displayed some error about a corrupted trace.

PR #821, opened 2025-05-22
Fix OL ignoring withdrawal fulfillments

delbonis:
## Description

See [STR-1415](https://alpenlabs.atlassian.net/browse/STR-1415).

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Unsure if this actually fixes the issue.

## Checklist

### DIFF
--- a/crates/chaintsn/src/errors.rs
+++ b/crates/chaintsn/src/errors.rs
@@ -55,6 +55,9 @@ pub enum OpError {
     #[error("invalid proof")]
     InvalidProof,
 
+    #[error("op referenced non-existent deposit {0}")]
+    UnknownDeposit(u32),
+
     /// Used to discard checkpoints we aren't looking for.
     #[error("operation does not advance the finalized epoch")]
     EpochNotExtend,
--- a/crates/chaintsn/src/transition.rs
+++ b/crates/chaintsn/src/transition.rs
@@ -162,16 +162,19 @@ fn process_l1_block(
     block_mf: &L1BlockManifest,
     params: &RollupParams,
 ) -> Result<(), TsnError> {
+    let blkid = block_mf.blkid();
+
     // Just iterate through every tx's operation and call out to the handlers for that.
     for tx in block_mf.txs() {
-        let in_blkid = block_mf.blkid();
         for op in tx.protocol_ops() {
             // Try to process it, log a warning if there's an error.
             if let Err(e) = process_proto_op(state, block_mf, op, params) {
-                warn!(?op, %in_blkid, %e, "invalid protocol operation");
+                warn!(?op, in_blkid = %blkid, %e, "invalid protocol operation");
             }
         }
     }
+
+    debug!(%blkid, "processed block manifest");
 
     Ok(())
 }
@@ -184,18 +187,28 @@ fn process_proto_op(
 ) -> Result<(), OpError> {
     match &op {
         ProtocolOperation::Checkpoint(ckpt) => {
+            let epoch = ckpt.checkpoint().batch_info().epoch();
+            debug!(%epoch, "processing checkpoint proto-op");
             process_l1_checkpoint(state, block_mf, ckpt, params)?;
         }
 
         ProtocolOperation::Deposit(info) => {
+            let deposit_idx = info.deposit_idx;
+            debug!(%deposit_idx, "processing deposit proto-op");
             process_l1_deposit(state, block_mf, info)?;
         }
 
         ProtocolOperation::WithdrawalFulfillment(info) => {
+            let deposit_idx = info.deposit_idx;
+            let txid = &info.txid;
+            debug!(%deposit_idx, %txid, "processing withdrawal fulfillment proto-op");
             process_withdrawal_fulfillment(state, info)?;
         }
 
         ProtocolOperation::DepositSpent(info) => {
+            let deposit_idx = info.deposit_idx;
+            let txid = &info.deposit_idx;
+            debug!(%deposit_idx, %txid, "processing reimbursement proto-op");
             process_deposit_spent(state, info)?;
         }
 
@@ -294,6 +307,10 @@ fn process_withdrawal_fulfillment(
     state: &mut StateCache,
     info: &WithdrawalFulfillmentInfo,
 ) -> Result<(), OpError> {
+    if !state.check_deposit_exists(info.deposit_idx) {
+        return Err(OpError::UnknownDeposit(info.deposit_idx));
+    }
+
     state.mark_deposit_fulfilled(info);
     Ok(())
 }
@@ -301,6 +318,9 @@ fn process_withdrawal_fulfillment(
 /// Locked deposit on L1 has been spent.
 fn process_deposit_spent(state: &mut StateCache, info: &DepositSpendInfo) -> Result<(), OpError> {
     // Currently, we are not tracking how this was spent, only that it was.
+    if !state.check_deposit_exists(info.deposit_idx) {
+        return Err(OpError::UnknownDeposit(info.deposit_idx));
+    }
 
     state.mark_deposit_reimbursed(info.deposit_idx);
     Ok(())
--- a/crates/state/src/state_op.rs
+++ b/crates/state/src/state_op.rs
@@ -302,7 +302,20 @@ impl StateCache {
         };
     }
 
-    /// Updates the deposit state to Fulfilled.
+    /// Returns if the deposit with some idx exists or not.
+    pub fn check_deposit_exists(&self, deposit_idx: u32) -> bool {
+        self.state()
+            .deposits_table()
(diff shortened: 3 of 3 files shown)


crates/l1tx/src/filter/withdrawal_fulfillment.rs
```diff
@@ -110,6 +152,9 @@ mod test {
     use super::*;
     use crate::filter::types::{conv_deposit_to_fulfillment, OPERATOR_FEE};
 
+    // First 4 bytes of "strata"?  I am not sure what this is supposed to be.
+    const MAGIC: [u8; 4] = [b's', b't', b'r', b'a'];
```
barakshani, line 156: Bridge currently uses `alpenstrata`

storopoli: Maybe we should move https://github.com/alpenlabs/alpen/blob/b978d34155796136340224dd5f927c48a83d0f9c/crates/util/python-utils/src/constants.rs#L5-L6
to `strata-test-utils`?

storopoli: and use it here

crates/l1tx/src/filter/withdrawal_fulfillment.rs
```diff
@@ -121,16 +166,18 @@ mod test {
     }
 
     fn create_opreturn_metadata(
```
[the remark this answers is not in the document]
delbonis: We really gotta figure out [this thing](https://alpenlabs.atlassian.net/browse/STR-1056) with the test utils crate.  I'm not sure the best solution here but it makes iterative testing really painful since it has to rebuild a ton of stuff.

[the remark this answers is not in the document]
delbonis: But in any case I moved it.

crates/l1tx/src/deposit/deposit_tx.rs
```diff
@@ -234,10 +234,12 @@ mod tests {
         },
         error::DepositParseError,
     };
-    const MAGIC_BYTES: &[u8] = &[1, 2, 3, 4, 5];
+
+    const MAGIC_BYTES: &[u8] = &[1, 2, 3, 4];
     const ADDRESS: &str = "bcrt1p729l9680ht3zf7uhl6pgdrlhfp9r29cwajr5jk3k05fer62763fscz0w4s";
 
     fn dummy_config() -> DepositTxParams {
+        assert_eq!(MAGIC_BYTES.len(), 4, "test: magic not for bytes");
```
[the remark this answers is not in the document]
delbonis: oops!

crates/l1tx/src/filter/withdrawal_fulfillment.rs
storopoli, line 156: This is what I am planning to use in the bridge. Finishing the PR...
```suggestion
    // NOTE: magic bytes must be EXACTLY 4 bytes
    const MAGIC: [u8; 4] = b"alp\0";
```

bin/strata-client/src/main.rs
```diff
@@ -78,6 +78,14 @@ fn main_inner(args: Args) -> anyhow::Result<()> {
     let params = resolve_and_validate_params(args.rollup_params.as_deref(), &config)
         .map_err(anyhow::Error::from)?;
 
+    // Sanity check to ensure we don't d
```
barakshani, line 81: we don't...?

delbonis:
Okay I am not sure how to resolve this conflict, it seems like some nontrivial change from I'm not sure where. @storopoli

crates/l1tx/src/filter/withdrawal_fulfillment.rs
Rajil1213: Wait, if we're hardcoding this value, why not just hard-code it as per the spec (i.e., `alpn`)? https://www.notion.so/SPS-50-L1-transaction-header-and-interpretation-1c5901ba000f80608cf8fdd9781666db?pvs=4

storopoli:
> Okay I am not sure how to resolve this conflict, it seems like some nontrivial change from I'm not sure where. @storopoli

Wait, is this already on `main`? Because your target branch is `releases/0.2.0`. Is this intentional?

storopoli:
> @storopoli Should I be merging into main first and _then_ backport?

Yes!

crates/l1tx/src/filter/withdrawal_fulfillment.rs
storopoli: We need to have a decision on this. I like `alpn` instead of `alp0` (all lowercase) but we need to make it consistent in `alpen` and `strata-bridge ` codebases.

Yay or nay for `alpn`?

Cc @barakshani

barakshani: we literally have months to bikeshed on this. let's merge this thing

bin/strata-client/src/errors.rs
```diff
@@ -10,6 +10,9 @@ pub enum InitError {
     #[error("missing init client state")]
     MissingInitClientState,
 
+    #[error("rollup name too long")]
```
bewakes, line 13: nit: The error message is not coherent with the enum variant. `InvalidRollupName` sounds like can occur even when the name is not long but does not match the expected one.

crates/l1tx/src/filter/types.rs
```diff
@@ -55,6 +55,15 @@ impl TxFilterConfig {
             generate_taproot_address(&operator_wallet_pks, rollup_params.network)?;
 
         let rollup_name = rollup_params.rollup_name.clone();
+
+        // Make sure the magic bytes is always of length 4.
+        // TODO replace this with a more generic system
+        let magic_bytes = rollup_name
+            .bytes()
+            .chain(std::iter::repeat(0))
+            .take(4)
```
bewakes, line 64: This `4` appears in multiple places. Might help to use a `const`.

bewakes (commented):
minor nits

storopoli:
@delbonis I've rebased this and set the target branch to `main`.
Let's first get the fix into `main` then we backport the squash commit that lands on `main` into `releases-0.2.0`.

As a heads up here `MAGIC_BYTES` is `"ALPN"` whereas in `strata-bridge` already merged is `"alpn"` (lowercase). Either we change here to lowercase or we merge this as UPPERCASE and I open a PR in `strata-bridge` to `s/alpn/ALPN/` all over the place.

barakshani:
> > > As a heads up here MAGIC_BYTES is "ALPN"
> > 
> > 
> > Don't we take this value from rollup name?
> 
> we have `tag` `rollup_name` `da-tag` all of these are header bytes that identify an `OP_RETURN`, hence `MAGIC_BYTES`.

I meant, if this is an externally provided parameter, why do we need to change something in this PR?

crates/asm/common/src/tx.rs
```diff
@@ -1,6 +1,6 @@
 use bitcoin::Transaction;
 
-/// A parsed SPS-50 tag payload (excluding the “ALPN” magic and subprotocol ID),
```
delbonis, line 3: This implies that the spec says the magic is `ALPN`.  The spec does not prescribe a specific magic value.

crates/asm/stf/src/tx_filter.rs
```diff
@@ -9,7 +9,7 @@ use strata_asm_common::{SubprotocolId, TagPayload, TxInput};
 ///
 /// The SPS-50 header MUST be encoded as an `OP_RETURN` in output index 0, with payload:
 /// ```text
-/// [0..4]   ASCII magic “ALPN”
+/// [0..4]   ASCII magic "ALPN"
```
delbonis, line 12: as above

crates/util/python-utils/src/constants.rs
```diff
@@ -3,7 +3,7 @@ use std::sync::LazyLock;
 use bdk_wallet::bitcoin::{Amount, Network};
 
 /// Magic bytes to add to the metadata output in transactions to help identify them.
-pub(crate) const MAGIC_BYTES: &[u8; 11] = b"alpenstrata";
+pub(crate) const MAGIC_BYTES: &[u8; 4] = b"ALPN";
```
delbonis, line 6: is this the thing we need to fix?

delbonis:
@storopoli This PR doesn't specify any value for the magic.  It reads it from the `rollup_name` field.

delbonis:
Oh

>  It would be nice to have the same value in both repos.

I actually think it's better if it's different so that we can make sure our code is reading from params consistently.  We've had issues in the past where things were hardcoded that shouldn't have been.

bewakes (approved):

PR #834, opened 2025-05-28
STR-767: EEST integration

evgenyzdanovich:
## Description

https://eest.ethereum.org/main/

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/.github/workflows/cron-eest.yml
+++ b/.github/workflows/cron-eest.yml
@@ -51,7 +51,12 @@ jobs:
           ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
           ECR_REPOSITORY_PREFIX: ${{ vars.ECR_REPOSITORY_PREFIX }}
       run: |
-        BINARIES=("strata-client", "strata-sequencer-client", "alpen-reth", "prover-client")
+        BINARIES=(
+          "strata-client"
+          "strata-sequencer-client"
+          "alpen-reth"
+          "prover-client"
+        )
         
         for IMAGE in "${BINARIES[@]}"; do
           ECR_REPOSITORY="${ECR_REPOSITORY_PREFIX}/${IMAGE}"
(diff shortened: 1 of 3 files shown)


.github/workflows/cron-eest.yml
```diff
@@ -0,0 +1,138 @@
...
+#  schedule:
+#    - cron: '0 3 * * 0'  # Weekly: Sunday 03:00 UTC
+#  workflow_dispatch:      # Manual trigger (optional)
+
+# TODO: clean this up.
+on:
+  pull_request:
+  merge_group:
+  push:
+    branches: [main]
```
storopoli: Adding a blocker here, since this workflow takes at least 1h (you've cancelled before termination):

image

Is this intended to run weekly?

By the way TOTALLY OFFTOPIC, but you can embed easter eggs in these cron hours.
See https://github.com/alpenlabs/alpen/blob/e1095761316f75ccec79279994a819a0a0f4ccbe/.github/workflows/cron-nightly-rust.yml#L5

Ref: https://en.wikipedia.org/wiki/Taxicab_number

.github/workflows/cron-eest.yml
```diff
@@ -0,0 +1,138 @@
...
+
+env:
+  CI: false
+  COMMIT: ${{ github.sha }}
+
+permissions:
+  contents: read
+  pull-requests: read
```
storopoli: You give global empty permissions, then you add then manually to the `jobs` per-use basis.
```suggestion
permissions: {}
```

.github/workflows/cron-eest.yml
evgenyzdanovich: changed to run weekly, yeah. Previously, I was testing it for some time.

.github/workflows/cron-eest.yml
evgenyzdanovich: Well, there's only one job and the permissions to read aren't "dangerous", the repo is public anyways

storopoli: Yes, but this is totally different from all the other CI jobs.
I think we should make this in the same template as others.
It is also what `zizmor` suggest in their docs: see here in the `after` tab https://docs.zizmor.sh/audits/#remediation_4

evgenyzdanovich: Yeah, you convinced me, thanks. Done!

storopoli (approved):
ACK b43196574d742824263211a9f40fcbcc642451b8

evgenyzdanovich:
@sapinb @purusang PTAL

purusang (approved):
Just a curiosity. Why do we pull all the images from ecr whereas build one for the prover only?

PR #856, opened 2025-06-05
Feat(ASM): draft impl of core subprotocol

irnb:
## Description

OL core ASM subprotocol that handles checkpointing and related functionality. (SPS-62)

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

STR-1333

### DIFF
--- a/crates/asm/common/src/msg.rs
+++ b/crates/asm/common/src/msg.rs
@@ -4,7 +4,7 @@ use std::any::Any;
 
 use borsh::{BorshDeserialize, BorshSerialize};
 
-use crate::{BRIDGE_SUBPROTOCOL_ID, SubprotocolId};
+use crate::SubprotocolId;
 
 /// Generic wrapper around a inter-proto msg.
 pub trait InterprotoMsg: Any + 'static {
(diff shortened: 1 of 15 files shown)


crates/asm/common/src/msg.rs
```diff
@@ -32,6 +34,36 @@ impl<const ID: SubprotocolId> InterprotoMsg for NullMsg<ID> {
...
+    pub data: Vec<u8>,
+    /// Nonce to prevent replay attacks?
+    pub nonce: u64,
+}
+
+/// Withdrawal messages to be sent to Bridge subprotocol
+#[derive(Clone, Debug)]
+pub struct WithdrawalMsg {
+    pub withdrawals: Vec<L2ToL1Msg>,
+}
```
delbonis: This is more than just withdrawals.  It should be a generic messages container a la SPS-msg-fmt.

crates/asm/common/src/msg.rs
```diff
@@ -32,6 +34,36 @@ impl<const ID: SubprotocolId> InterprotoMsg for NullMsg<ID> {
...
+pub struct L2ToL1Msg {
+    /// Destination address on L1 (Bitcoin)
+    pub dest_address: Vec<u8>,
+    /// Amount to withdraw in satoshis
+    pub amount: u64,
+    /// Additional data payload for the withdrawal
+    pub data: Vec<u8>,
+    /// Nonce to prevent replay attacks?
+    pub nonce: u64,
+}
```
delbonis: A few things:

* dest address is irrelevant here, what if we're just sending a message to the upgrade subprotocol?
* we don't need an amount here, we're "a layer down" from where we want to treat quantities as a preserved thing
* nonce isn't really useful, since we are effectively using the epoch as a sequence number
* I mentioned this in a few other places, but `L2ToL1Msg` is a really bad name because it's too similar to `L1ToL2Msg` , which means it interrupts your flow and you have to sanity check to make sure you read it right.  You also have to toggle your shift state 8 times when you're writing it which makes it really annoying to type.
* This is really part of the checkpoint structure, so it should be closer to that in the `state` crate.  This `msg` module is for messages *between* subprotocols, strictly within a ASM STF invocation.

crates/asm/common/src/subprotocol.rs
```diff
@@ -28,11 +28,17 @@ pub trait Subprotocol: 'static {
...
+    type GenesisConfig: Any + BorshDeserialize + BorshSerialize;
+
     /// Message type that we receive messages from other subprotocols using.
     type Msg: Clone + Any;
 
     /// Constructs a new state to use if the ASM does not have an instance of it.
-    fn init() -> Self::State;
+    /// # Arguments
+    /// * `genesis_config` - Configuration parameters for initializing the state
+    fn init(genesis_config: Self::GenesisConfig) -> Self::State;
```
delbonis: Should take a ref so that we don't have to clone it each time.

crates/asm/common/src/subprotocol_genesis.rs
```diff
@@ -0,0 +1,102 @@
...
+
+/// Registry for managing genesis state for all subprotocols.
+///
+/// This registry stores serialized genesis state that are used
+/// when initializing subprotocol states. The state are keyed
+/// by subprotocol ID and stored in serialized form to avoid type dependencies.
+#[derive(Debug, Clone, Default, BorshSerialize, BorshDeserialize)]
+pub struct GenesisConfigRegistry {
+    /// Map of subprotocol ID to serialized genesis state
+    configs: BTreeMap<SubprotocolId, Vec<u8>>,
```
delbonis, line 21: I can see why you did this this way.  Since if there's shared information across subprotocols then we can have it in another field.  But as it's written, we could just extract the `Vec<u8>` for the subprotocol and pass the ref to the `init` fn, preventing from reading data for other subprotocols.  So *as-is* it could be simplified and made safer, but maybe we don't want to.

crates/asm/subprotocols/core/src/checkpoint_zk_verifier.rs
delbonis, line 1: All of this should be in a shared crate.  I think this is introducing a 3rd separate place where we implement checkpoint verification, and differently!  Perhaps we should wait to finish this until we specify the checkpoint structure and verification.

crates/asm/subprotocols/core/src/error.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// Invalid ZK proof
+    #[error("Invalid ZK Proof")]
+    InvalidProof,
+}
+
+impl From<borsh::io::Error> for CoreError {
+    fn from(e: borsh::io::Error) -> Self {
+        CoreError::TxParsingError(e.to_string())
+    }
+}
```
delbonis: Is `TxParsingError` the right variant here?  It's coming from borsh.  If there's a contextual reason then I would think it shouldn't be a `From` impl, which implies that all borsh errors are tx parsing errors.

crates/asm/subprotocols/core/src/lib.rs
```diff
@@ -30,6 +67,28 @@ pub struct CoreOLState {
...
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize)]
+pub struct CoreGenesisConfig {
+    /// The initial checkpoint verifying key for zk-SNARK proof verification
+    pub checkpoint_vk: VerifyingKey,
+
+    /// The initial verified checkpoint state (usually genesis checkpoint)
+    pub initial_checkpoint: EpochSummary,
+
+    /// The initial L1 block reference for the checkpoint
+    pub initial_l1_ref: L1BlockId,
```
delbonis, line 86: "initial" or "current"?

crates/asm/subprotocols/core/src/lib.rs
```diff
@@ -44,17 +103,76 @@ impl Subprotocol for OLCoreSubproto {
...
+            // `tracing` crate. We need a proper logging mechanism to identify which
+            // transaction processing failed and why. For now, we print errors to stderr
+            // as a temporary solution.
+            //
+            // We can't propagate errors to upper layers when transaction processing fails because
+            // invalidating and rejecting transactions is normal and expected behavior. We don't
+            // want to halt the entire block processing because of a single invalid transaction.
+            if let Err(e) = result {
+                let txid = tx.tx().compute_txid();
+                eprintln!("Error processing transaction (txid: {txid:?}): {e:?}");
```
delbonis, line 150: Should be `warn` probably.  And use log params!

delbonis (changes requested):
good start

crates/asm/subprotocols/core/src/lib.rs
irnb: By "warn properly" do you mean a proper message log or the warn! macro from the tracing crate?

```
// TODO: Implement proper logging approach
// Since this code also runs as a part of zkVM guest program, we cannot use the
// `tracing` crate. We need a proper logging mechanism to identify which
// transaction processing failed and why. For now, we print errors to stderr
// as a temporary solution.
```

Is it OK to use tracing method in pure zkvm function (guest program)?

crates/asm/subprotocols/core/src/lib.rs
irnb: Since it's part of CoreGenesisConfig, it's initial, because if we provide GenesisConfig that means we don't have state and we are in the subprotocol genesis phase.

crates/asm/subprotocols/core/src/checkpoint_zk_verifier.rs
irnb: I agree

crates/asm/common/src/msg.rs
```diff
@@ -32,39 +34,215 @@ impl<const ID: SubprotocolId> InterprotoMsg for NullMsg<ID> {
...
+            MessageError::BufferTooShort => write!(f, "buffer too short"),
+            MessageError::TypeOutOfBounds(ty) => write!(f, "type {ty:#x} out of bounds"),
+            MessageError::NonMinimalEncoding(ty) => {
+                write!(f, "non-minimal encoding for type {ty:#x}")
+            }
+        }
+    }
+}
+
+impl std::error::Error for MessageError {}
```
delbonis, line 190: Can use `thiserror::Error` instead of doing this manually.  Unless there's some reason you refrained from it?

crates/asm/common/src/msg.rs
delbonis, line 1: A lot of this can use the message fmt types from strata-common instead of having to reimpl it.

PR #825, opened 2025-05-23
STR-1184 & STR-1377: Bump reth to v1.3.1 and migrate to RSP

MdTeach:
## Description
* Upgrade reth from `v1.1.5` -> `v1.3.1`
* Use [RSP](https://github.com/succinctlabs/rsp) for AlpenEE proving
* Refactor Strata -> Alpen for Evm components

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [x] Dependency Update

## Notes to Reviewers
* This implementation currently burns the base fee instead of transferring it to the `BASEFEE_ADDRESS`.
The issue is being tracked in [STR-1442](https://alpenlabs.atlassian.net/browse/STR-1442).
* Uses a [fork of rsp ](https://github.com/alpenlabs/rsp/tree/alpen/rsp)because the original library does not expose a public method to instantiate ClientExecutor. The changes in the fork are minimal. Need to reach out to the sp1 team or send a pull request to rsp.

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have commented my code where necessary.
- [ ] I have updated the documentation if needed.
- [ ] My changes do not introduce new warnings.
- [ ] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [ ] New and existing tests pass with my changes.

## Related Issues

[STR-1442]: https://alpenlabs.atlassian.net/browse/STR-1442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #857, opened 2025-06-05
feat(zkvm): replace anyhow::Error with typed errors in EVM proof implementation

storopoli:
## Description

This is the last bit of `anyhow` untyped errors that should be moved to typed errors in STR-315.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

STR-315

### DIFF
--- a/crates/proof-impl/evm-ee-stf/src/error.rs
+++ b/crates/proof-impl/evm-ee-stf/src/error.rs
@@ -1,0 +1,75 @@
+//! Error types for EVM Execution Environment State Transition Function (EE-STF) proof
+//! implementation.
+
+use alloy_primitives::Address;
+use revm_primitives::alloy_primitives::U256;
+use thiserror::Error;
+
+/// EVM EE-STF proof implementation operations errors.
+#[derive(Error, Debug)]
+pub enum EvmEeStfError {
+    /// Database operations error.
+    #[error("Database error: {0}")]
+    Database(#[from] DatabaseError),
+
+    /// Merkle Patricia Trie operations error.
+    #[error("MPT error: {0}")]
+    Mpt(#[from] strata_mpt::Error),
+
+    /// Account operations error.
+    #[error("Account error: {0}")]
+    Account(#[from] AccountError),
+
+    /// Validation error.
+    #[error("Validation error: {message}")]
+    Validation { message: String },
+
+    /// Block processing error.
+    #[error("Block processing error: {message}")]
+    BlockProcessing { message: String },
+}
+
+/// Database operations errors.
+#[derive(Error, Debug)]
+pub enum DatabaseError {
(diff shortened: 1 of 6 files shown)


crates/proof-impl/evm-ee-stf/src/error.rs
```diff
@@ -0,0 +1,75 @@
...
+    #[error("Account error: {0}")]
+    Account(#[from] AccountError),
+
+    /// Validation error.
+    #[error("Validation error: {message}")]
+    Validation { message: String },
+
+    /// Block processing error.
+    #[error("Block processing error: {message}")]
+    BlockProcessing { message: String },
```
delbonis: Saying `error` is redundant in all of these messages, fwiw.

crates/proof-impl/evm-ee-stf/src/error.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/// Database operations errors.
+#[derive(Error, Debug)]
+pub enum DatabaseError {
+    /// Account not found.
+    #[error("Account not found: {address}")]
+    AccountNotFound { address: Address },
```
delbonis, line 37: Something like this I usually quote it instead of using the colon, like you're doing for `StorageSlotNotFound`.  The colon implies that the thing that comes after it is somehow the cause somehow (which I guess is true here in a metaphorical sense).

delbonis (changes requested):
minor nits

crates/proof-impl/evm-ee-stf/src/error.rs
storopoli: Ok done in 80eb4da21e464fbf014294c9987c649388947c5d

crates/proof-impl/evm-ee-stf/src/error.rs
storopoli: I don't follow 100%, you mean what I did in 80eb4da21e464fbf014294c9987c649388947c5d?

storopoli:
> Given that, I think we should hold this PR for now and revisit it after PR #825 is merged.

Moving to draft

crates/proof-impl/evm-ee-stf/src/error.rs
delbonis: Yes, although I always do single quotes here.

delbonis: It's fine for now, there's a separate ticket to make the error message style consistent.

storopoli:
Actually this is not needed anymore since #825 removed the `anyhow` deps that this PR was doing typed error. It also removed a bunch of stuff that this PR added typed errors

PR #859, opened 2025-06-05
ci: sync rust toolchain with `rust-toolchain.toml` file

storopoli:
## Description

CI jobs needs to specify a `toolchain` since `dtolnay/rust-toolchain` does not support `rust-toolchain.toml` (see https://github.com/dtolnay/rust-toolchain/issues/133).
This is annoying since we have manually sync the rust nightly version across multiple files, and also the damn thing is automatically updated every day 1.

This PR adds the feature for CI jobs to extract the nightly version from the `rust-toolchain.toml` using regex magic 🧙🏻 and saves the output in a ENV variable that can be used in further steps by the CI job to know which nightly version to use.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update
- [x] CI improvements

## Checklist

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -28,19 +28,37 @@ env:
 permissions: {}
 
 jobs:
+  extract-rust-version:
+    name: Extract Rust toolchain version
+    runs-on: ubuntu-latest
+    outputs:
+      rust-version: ${{ steps.extract.outputs.rust-version }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+        with:
+          persist-credentials: false
+
+      - name: Extract Rust toolchain version
+        id: extract
+        uses: ./.github/actions/extract-rust-version # zizmor: ignore[unpinned-uses]
+
   update:
     name: Update
     runs-on: ubuntu-latest
+    needs: extract-rust-version
     permissions:
       contents: write # Needed to create commits
       pull-requests: write # Needed to create a PR
     steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
         with:
           persist-credentials: false
+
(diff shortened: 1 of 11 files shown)


Rajil1213 (changes requested):
The `prepare` job should be called `extract-rust-version` and moved to a separate action file under `.github/actions` (similar to the `cleanup` action). Is there a reason why that approach was not taken?

.github/workflows/dependencies.yml
```diff
@@ -28,19 +28,37 @@ env:
...
+  extract-rust-version:
+    name: Extract Rust toolchain version
+    runs-on: ubuntu-latest
+    outputs:
+      rust-version: ${{ steps.extract.outputs.rust-version }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+        with:
+          persist-credentials: false
```
Rajil1213, line 40: Can we not do these in the composite action?

storopoli: I don't follow. What is "the composite action"?

Rajil1213: The one in the `actions` directory.

Rajil1213: Essentially, I want the `extract-rust-version` action to output the version and make it available to other jobs/steps. It should do whatever is necessary to achieve that including checking out the repository. This makes the rest of the workflows a lot cleaner. If this is not something that github CI supports, then I guess it's fine.

storopoli: Ok done in c274447facac022e8f0e634245a1d70fa78824b0

storopoli:
I think GitHub actions is having issues today. Let's close and open again

.github/workflows/dependencies.yml
storopoli: Reverting it Rajil. It does not work in see https://github.com/alpenlabs/alpen/actions/runs/15474728868/job/43569505493#step:2:1

It only worked for cleanup action because it did not depend on the repo being available in the CI runner, since it only deletes stuff from the CI runner machine.

Rajil1213: That's unfortunate. 😞

PR #864, opened 2025-06-06
Str 1455 upgrade subprotocol

prajwolrg:
## Description

This is an implementation of basic flow in the Upgrade Subprotocol based on the PRD. 

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


crates/asm/subprotocols/upgrade/src/state.rs
```diff
@@ -0,0 +1,113 @@
...
+    pub fn tick_scheduled(&mut self) -> Vec<ScheduledUpgrade> {
+        for s in &mut self.scheduled {
+            s.decrement_blocks_remaining();
+        }
+        let (ready, rest): (Vec<_>, Vec<_>) = std::mem::take(&mut self.scheduled)
+            .into_iter()
+            .partition(|u| u.blocks_remaining() == 0);
+        self.scheduled = rest;
+        ready
+    }
```
delbonis: As mentioned in the spec docs, this is the wrong way to approach this.  We should just be treating it as an list of operations ordered by an activation height field (instead of a "remaining blocks" field), so that we don't have to touch any entries that aren't going to be acted upon yet.

crates/asm/subprotocols/upgrade/src/subprotocol.rs
```diff
@@ -0,0 +1,80 @@
...
+        // Process each transaction based on its type
+        for tx in txs {
+            match tx.tag().tx_type() {
+                CANCEL_TX_TYPE => {
+                    let _ = handle_cancel_tx(state, tx);
+                }
+                ENACT_TX_TYPE => {
+                    let _ = handle_enactment_tx(state, tx);
+                }
+                _ => {
```
delbonis: Should be `UPDATE_TX_TYPE`.

crates/asm/subprotocols/upgrade/src/subprotocol.rs
```diff
@@ -0,0 +1,80 @@
...
+
+        // Process each transaction based on its type
+        for tx in txs {
+            match tx.tag().tx_type() {
+                CANCEL_TX_TYPE => {
+                    let _ = handle_cancel_tx(state, tx);
+                }
+                ENACT_TX_TYPE => {
+                    let _ = handle_enactment_tx(state, tx);
+                }
```
delbonis: Shouldn't be matching on the tx_type field like this.  We should have a function that parses the relevant operation data out of the tx *first* and then pattern match on that to handle the operation, instead of making parsing and handling the same operation.

crates/asm/subprotocols/upgrade/src/multisig/msg.rs
```diff
@@ -0,0 +1,62 @@
...
+
+/// A high‐level multisig operation that participants can propose.
+#[derive(Debug, Clone, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
+pub enum MultisigOp {
+    /// Cancel a pending action
+    Cancel(CancelAction),
+    /// Execute a committed action
+    Enact(EnactAction),
+    /// Propose an upgrade
+    Upgrade(UpgradeAction),
```
delbonis: We should not be thinking of these as *multisig* ops.  We should be separating the authorization verification logic from the action implementation logic.  These are just different kinds of operations that the administration authority can issue.

Also, should be consistent about if we're using the term "action" and "operation" here.  I don't think there's a particular reason to use one term or the other, but we should just be consistent.

crates/asm/subprotocols/upgrade/src/multisig/msg.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+/// A multisig payload comprising an operation plus a nonce, ready for hashing and signing.
+#[derive(Debug, Clone, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct MultisigPayload {
+    op: MultisigOp,
+    nonce: u64,
```
delbonis: If this works like ethereum nonces then actually it should be "seqno", imo.

crates/asm/subprotocols/upgrade/src/state.rs
```diff
@@ -0,0 +1,113 @@
...
+    queued: Vec<QueuedUpgrade>,
+
+    /// Actions that have completed their waiting period and can be
+    /// enacted by EnactmentTx, but can no longer be cancelled.
+    committed: Vec<CommittedUpgrade>,
+
+    /// Actions that will be executed automatically without requiring
+    /// an EnactmentTx transaction.
+    scheduled: Vec<ScheduledUpgrade>,
+}
```
delbonis: There's a lot of shared structure here (and in accessor functions) that makes me think that it could be abstracted into some kind of container that better captures this.

Maybe something like:

```rs
struct HeightQueue<T> {
  entries: Vec<HeightQueueEntry<T>>
}

struct HeightQueueEntry<T> {
  id: u32,
  ready_height: u32,
  data: T,
}
```

And then you'd just have the wrapper fns to handle moving entries around according to the policy, similar to what you already have.

crates/asm/subprotocols/upgrade/src/txs/updates/id.rs
```diff
@@ -0,0 +1,27 @@
+use borsh::{BorshDeserialize, BorshSerialize};
+use strata_primitives::buf::Buf32;
+
+/// A unique identifier for an update
+#[derive(Debug, Clone, Copy, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct UpdateId(Buf32);
```
delbonis: What's the rationale for having this be a function of the update data?  We can establish a consistent global ordering here (at least after txs are accepted in a block) so I think it might make more sense to have this just be based on a sequential ordering and use a u32.

crates/asm/subprotocols/upgrade/src/txs/updates/operator.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use crate::{crypto::PubKey, error::UpgradeError};
+
+/// An update to the Bridge Operator Set:
+/// - removes the specified `old_members`
+/// - adds the specified `new_members`
+#[derive(Debug, Clone, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct OperatorSetUpdate {
+    new_members: Vec<PubKey>,
+    old_members: Vec<PubKey>,
```
delbonis: As mentioned in other doc, members being removed should be `u16`s.

crates/asm/subprotocols/upgrade/src/roles.rs
```diff
@@ -0,0 +1,46 @@
...
+    StrataAdmin,
+    /// The multisig authority that has the exclusive ability to change the
+    /// VerifyingKey of the OL STF.
+    StrataConsensusManager,
+}
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
+pub enum StrataProof {
+    ASM,
+    OlStf,
```
delbonis: Should be `ProofType` and variants should be like `Asm`, `OrchCheckpoint`.

crates/asm/subprotocols/upgrade/src/multisig/authority.rs
```diff
@@ -0,0 +1,95 @@
...
+/// nonce.
+#[derive(Debug, Clone, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct MultisigAuthority {
+    /// The role of this multisignature authority.
+    role: Role,
+    /// The public keys of all grant-holders authorized to sign.
+    config: MultisigConfig,
+    /// Nonce for the multisig configuration.
+    /// This is used to prevent replay attacks
+    nonce: u64,
```
delbonis: Probably should be seqno, as in the other comment.

crates/asm/subprotocols/upgrade/src/multisig/config.rs
```diff
@@ -0,0 +1,209 @@
...
+
+/// Configuration for a multisignature authority:
+/// who can sign (`keys`) and how many of them must sign (`threshold`).
+#[derive(Debug, Clone, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct MultisigConfig {
+    /// The public keys of all grant-holders authorized to sign.
+    pub keys: Vec<PubKey>,
+    /// The minimum number of keys that must sign to approve an action.
+    pub threshold: u8,
+}
```
delbonis: We should make this another step more generic.  Have an enum that's `AuthorityConfig`, and this a `SchnorrMultisigAuthConfig` type.  Wherever there's a space to put a "version field" it's probably a good idea.  If at the very least so we can switch to a new sig type if we have to.

That being said, that might make the multisig update types a bit more complicated.

crates/asm/subprotocols/upgrade/src/multisig/config.rs
```diff
@@ -0,0 +1,209 @@
...
+
+impl MultisigConfig {
+    /// Create a new config; panics if threshold == 0 or > keys.len().
+    pub fn new(keys: Vec<PubKey>, threshold: u8) -> Self {
+        assert!(!keys.is_empty(), "keys cannot be empty");
+        let max = keys.len() as u8;
+        assert!(
+            (1..=max).contains(&threshold),
+            "threshold must be between 1 and number of keys"
+        );
```
delbonis: Should have a `# Panics` section in the doc comment here.

crates/asm/subprotocols/upgrade/src/multisig/config.rs
```diff
@@ -0,0 +1,209 @@
...
+                min_required,
+            });
+        }
+
+        Ok(())
+    }
+
+    pub fn apply(&mut self, update: &MultisigConfigUpdate) {
+        // Remove specified old members
+        self.keys.retain(|key| !update.old_members().contains(key));
```
delbonis: I think this makes it n^2.  If we assert these lists are always sorted then we can do a more efficient merge-and-remove pass with both this and the new entries.  Do you think that's feasible here?

crates/asm/subprotocols/upgrade/src/upgrades/delayed.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use crate::txs::updates::{UpgradeAction, id::UpdateId};
+
+/// A time-delayed upgrade action with different delay semantics
+#[derive(Debug, Clone, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct DelayedUpgrade<T> {
+    id: UpdateId,
+    action: UpgradeAction,
+    blocks_remaining: u64,
+    _marker: std::marker::PhantomData<T>,
```
delbonis: What does this generic parameter mean here?  What does it represent?

crates/asm/subprotocols/upgrade/src/upgrades/delayed.rs
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    // Getters
+    pub fn id(&self) -> &UpdateId {
+        &self.id
+    }
+    pub fn action(&self) -> &UpgradeAction {
+        &self.action
+    }
+    pub fn blocks_remaining(&self) -> u64 {
```
delbonis: formatting!

crates/asm/subprotocols/upgrade/src/upgrades/queued.rs
```diff
@@ -0,0 +1,30 @@
...
+    upgrades::delayed::DelayedUpgrade,
+};
+
+#[derive(Debug, Clone, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct QueueDelay;
+
+pub type QueuedUpgrade = DelayedUpgrade<QueueDelay>;
+
+pub const ASM_VK_QUEUE_DELAY: u64 = 12_960;
+pub const OL_STF_VK_QUEUE_DELAY: u64 = 4_320;
```
delbonis: These should be in some common high-level constants.rs file, or maybe somehow pulled in from some `UpgradeParams`.  I haven't thought about how that data would be passed around.

crates/asm/subprotocols/upgrade/src/upgrades/scheduled.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[derive(Debug, Clone, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct ExecutionDelay;
+
+pub type ScheduledUpgrade = DelayedUpgrade<ExecutionDelay>;
+
+pub const MULTISIG_CONFIG_UPDATE_ENACTMENT_DELAY: u64 = 2_016;
+pub const OPERATOR_UPDATE_ENACTMENT_DELAY: u64 = 2_016;
+pub const SEQUENCER_UPDATE_ENACTMENT_DELAY: u64 = 2_016;
+pub const VK_UPDATE_ENACTMENT_DELAY: u64 = 144;
```
delbonis: See other comment on constants.

crates/asm/subprotocols/upgrade/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+
+pub mod actions;
+pub mod crypto;
+pub mod error;
+pub mod multisig;
+pub mod roles;
+pub mod state;
+pub mod subprotocol;
+pub mod txs;
+pub mod upgrades;
```
delbonis: I've come around on how crate exports should be structured.  We should only be exporting the items that actually should be exported, like the subprotocol itself, tx data types, etc.  And we *shouldn't* export modules that aren't themselves some major subsystem we expect people to depend on the internals of.

delbonis (changes requested):
A few more general points:

* There's a lot of chunks of code here that seem like they're "almost duplicated" across the different kinds of updates.  It's not quite as much as it seemed in an earlier version of the specs, but it's still there.
* Most of these msg/tx data types should be in their own crate so that the crates that are involved in signing/generating them don't have to depend on all this ASM infra in order to construct them.
* Following that, we also need some utilities to construct these upgrade txs.  Maybe not going as far as to integrate it into the datatool, but if we have something to decode and verify a thing then we need something to sign and encode that thing too.  Since we might find out that we've defined the types in a way that's impossible to construct. :)

crates/asm/subprotocols/upgrade/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+[dependencies]
+borsh.workspace = true
+strata-asm-common.workspace = true
+strata-primitives.workspace = true
+thiserror.workspace = true
+zkaleido.workspace = true
+
+[lints]
+workspace = true
```
storopoli, line 16: THANK YOU!

crates/asm/subprotocols/upgrade/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+name = "strata-asm-proto-upgrade"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+borsh.workspace = true
+strata-asm-common.workspace = true
+strata-primitives.workspace = true
+thiserror.workspace = true
+zkaleido.workspace = true
```
storopoli: We are breaking the deps up into strata on top and others below
```suggestion
strata-asm-common.workspace = true
strata-primitives.workspace = true
zkaleido.workspace = true

borsh.workspace = true
thiserror.workspace = true
```

storopoli (commented):
Just reviewed the `Cargo.toml`s.

crates/asm/subprotocols/upgrade/src/upgrades/delayed.rs
prajwolrg: This is to have distinction between `QueuedUpgrades` and `ScheduledUpgrades`.

crates/asm/subprotocols/upgrade/src/state.rs
prajwolrg: I use `DelayedUpgrade<T>` to handle this in some portion. `QueuedUpgrade` and `ScheduledUpgrade` are instantiation of the type.

crates/asm/subprotocols/upgrade/src/upgrades/delayed.rs
delbonis: This sorta feels like it's misusing generic types.  It'd make more sense to newtype it a la `struct QueuedUpdate(DelayedUpgrade);`.

crates/asm/subprotocols/upgrade/src/state.rs
delbonis: There's a lot of implicit invariants / duplicated logic here to handle the amorphous `Vec`s.  Encapsulation is key!

crates/asm/txs/upgrade/src/crypto/multisig_config.rs
delbonis, line 1: Need to make sure it's clear this is *all* about schnorr.

delbonis (changes requested):

PR #866, opened 2025-06-09
Update fullnode docker

bewakes:
This PR updates the docker config for full node. This is what will be used in the docs for users to operate a full node.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


sapinb:
@bewakes why is the sequencer config removed ? The other 2 docker-compose files still use it.
Also, I recommend keeping the helper scripts. Those are useful when running/testing locally.

bewakes:
Yes, those should not have been removed to maintain compatibility with the existing config. I reverted those changes.

docker/docker-compose-fullnode.yml
delbonis, line 1: Having both build and image directives in this seems wrong.  We shouldn't make this dependent on particular images existing in AWS ECR.

bewakes: You know, actually it is not wrong. This gives us flexibility to use the local build when we want to. For the end users, they can pull the images instead of having to build locally.
However to separate things out, we can have a dedicated compose file for users to pull the official images. Since we already have `docker-compose.yml` that supports local builds, this file `docker-compose-fullnode.yml` could just have the image option and not build option.

bewakes: Updated.

bewakes:
Once this is merged, I'll create a backport for `releases/0.2.0`.

docker/docker-compose-fullnode.yml
delbonis: We definitely should not commingle the different use-cases into a single compose file.

sapinb (approved):

PR #867, opened 2025-06-09
STR-1461: Enhance fntests with keep-alive execution mode.

evgenyzdanovich:
## Description

Oftentimes, during development and local testing, I find myself:
- taking a random fntest,
- substituting it with indefinite while true: pass
- running it and using it as a keep-alive mechanism (to have the local env spin up).

Thus, I propose a small enhancement to `fntests` to make it easy and lightweight mechanism to play around and test stuff locally.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This is **the first change** in the series of upcoming small cleanup/improvements as part of **STR-1461**.
Given it's quite isolated and independent thing, I squeezed it out in a separate PR.

## Checklist

### DIFF
--- a/functional-tests/tests/keepalive_stub_test.py
+++ b/functional-tests/tests/keepalive_stub_test.py
@@ -1,0 +1,20 @@
+import time
+
+import flexitest
+
+from envs import testenv
+
+
+@flexitest.register
+class KeepAliveEnvMockTest(testenv.StrataTester):
+    """
+    A dynamically populated mock test for the keep-alive mode.
+    """
+
+    def __init__(self, ctx: flexitest.InitContext):
+        ctx.set_env("{ENV}")
+
+    def main(self, _ctx: flexitest.RunContext):
+        while True:
+            print("running fn-tests in keep-alive mode")
+            time.sleep(60)
--- a/functional-tests/README.md
+++ b/functional-tests/README.md
@@ -110,3 +110,25 @@ PROVER_TEST=1 ./run_test.sh -g prover
 ```
 
 The test harness script will be extended with more functionality as we need it.
+
+
+## Keep-alive env setup
+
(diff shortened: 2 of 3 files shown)


functional-tests/tests/keepalive_stub_test.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@flexitest.register
+class KeepAliveEnvMockTest(testenv.StrataTester):
+    """
+    A dynamically populated mock test for the keep-alive mode.
+    """
+
+    def __init__(self, ctx: flexitest.InitContext):
+        ctx.set_env("{ENV}")
```
delbonis, line 15: huh?

delbonis (approved):

delbonis:
@evgenyzdanovich Yeah I have recognized some desire for this in the past but I wasn't quite sure the best way to go about it.  Really the way to implement it is to have a mode where the test runtime stands up a specific env and then just sit and wait for `KeyboardInterrupt` before just shutting it down.  Because it's not really *running a test* like this approach is.  This works fine but when I have some time maybe I'll go and implement that.

functional-tests/tests/keepalive_stub_test.py
evgenyzdanovich: dynamically injecting the env name from what's passed in CLI.

you can check `load_keepalive_mock_test`.
I understand its not ideal, just hope it's not too hacky:)

evgenyzdanovich:
> is to have a mode where the test runtime stands up a specific env and then just sit and wait for KeyboardInterrupt before just shutting it down

Agree, that's what `flexitest` would ideally have in the future :) Until that, we can live with this little hack

PR #872, opened 2025-06-12
Replace time.sleep with wait_until where relevant

bewakes:
## Description

This PR replaces the ad-hoc `time.sleep`s with `wait_until*` which is more intentional and easier to reason about. Not all the instances are replaced, there are a couple of places where this replacement is not feasible.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/functional-tests/envs/testenv.py
+++ b/functional-tests/envs/testenv.py
@@ -401,9 +401,12 @@ class HubNetworkEnvConfig(flexitest.EnvConfig):
         }
 
         # Wait until clients are ready
-        seq_waiter = StrataWaiter(sequencer.create_rpc(), None, timeout=30, interval=2)
+        import logging
+
+        logger = logging.getLogger(__name__)
+        seq_waiter = StrataWaiter(sequencer.create_rpc(), logger, timeout=30, interval=2)
         seq_waiter.wait_until_client_ready()
-        fn_waiter = StrataWaiter(fullnode.create_rpc(), None, timeout=30, interval=2)
+        fn_waiter = StrataWaiter(fullnode.create_rpc(), logger, timeout=30, interval=2)
         fn_waiter.wait_until_client_ready()
         # TODO: add others like prover, reth, btc
 
--- a/functional-tests/tests/prover/prover_checkpoint_latest.py
+++ b/functional-tests/tests/prover/prover_checkpoint_latest.py
@@ -1,5 +1,3 @@
-import time
-
 import flexitest
 
 from envs import testenv
--- a/functional-tests/utils/wait/__init__.py
+++ b/functional-tests/utils/wait/__init__.py
@@ -6,7 +6,7 @@ Usage:
 
     # Create waiters with RPC clients
     strata_waiter = StrataWaiter(seqrpc, logger, timeout=30)
-    reth_waiter = RethWaiter(rethrpc, timeout=10)
+    reth_waiter = RethWaiter(rethrpc, logger, timeout=10)
     prover_waiter = ProverWaiter(prover_rpc, logger, timeout=300)
 
     # Use waiter methods
--- a/functional-tests/utils/wait/base.py
+++ b/functional-tests/utils/wait/base.py
@@ -1,20 +1,21 @@
 import logging
-from dataclasses import dataclass
-from typing import Generic, TypeVar
+from typing import Protocol
 
 from utils.utils import wait_until, wait_until_with_value
 
-T = TypeVar("T")
 
+class WaiterProtocol(Protocol):
+    """Protocol defining the interface that waiter classes must implement."""
 
-@dataclass
-class BaseWaiter(Generic[T]):
-    inner: T
+    timeout: int
+    interval: float
     logger: logging.Logger
-    timeout: int = 10
-    interval: float = 0.5
 
-    def _wait_until(self, *args, timeout=None, step=None, **kwargs):
+
+class WaiterMixin:
+    """Mixin providing wait utilities for classes implementing WaiterProtocol."""
+
+    def _wait_until(self: WaiterProtocol, *args, timeout=None, step=None, **kwargs):
         return wait_until(
             *args,
             timeout=timeout or self.timeout,
@@ -22,10 +23,22 @@ class BaseWaiter(Generic[T]):
             **kwargs,
         )
 
-    def _wait_until_with_value(self, *args, timeout=None, step=None, **kwargs):
+    def _wait_until_with_value(self: WaiterProtocol, *args, timeout=None, step=None, **kwargs):
         return wait_until_with_value(
             *args,
             timeout=timeout or self.timeout,
             step=step or self.interval,
             **kwargs,
         )
+
+
+class RpcWaiter(WaiterMixin):
+    """Waiter base class that waits on a particular RPC"""
+
+    def __init__(
+        self, rpc_client, logger: logging.Logger, timeout: int = 20, interval: float = 0.5
+    ):
+        self.rpc_client = rpc_client
+        self.logger = logger
+        self.timeout = timeout
+        self.interval = interval
--- a/functional-tests/utils/wait/prover.py
+++ b/functional-tests/utils/wait/prover.py
@@ -1,16 +1,9 @@
 import time
-from dataclasses import dataclass
-from typing import Any
 
-from utils.wait.base import BaseWaiter
+from utils.wait.base import RpcWaiter
 
 
-@dataclass
-class ProverWaiter(BaseWaiter[Any]):
-    """
-    Wrapper for encapsulating and waiting prover related operations
-    """
-
+class ProverWaiter(RpcWaiter):
     def wait_for_proof_completion(self, task_id: str, timeout: int | None = None) -> bool:
         """
         Waits for a proof task to complete/fail within a specified timeout period.
@@ -35,7 +28,7 @@ class ProverWaiter(BaseWaiter[Any]):
 
         while True:
             # Fetch the proof status
-            proof_status = self.inner.dev_strata_getTaskStatus(task_id)
+            proof_status = self.rpc_client.dev_strata_getTaskStatus(task_id)
             assert proof_status is not None
             self.logger.info(f"Got the proof status {proof_status}")
 
@@ -62,7 +55,7 @@ class ProverWaiter(BaseWaiter[Any]):
 
         timeout = timeout or self.timeout
         self._wait_until(
-            lambda: self.inner.dev_strata_getReport() is not None,
+            lambda: self.rpc_client.dev_strata_getReport() is not None,
             error_with="Prover did not start on time",
             timeout=timeout,
             step=self.interval,
--- a/functional-tests/utils/wait/reth.py
+++ b/functional-tests/utils/wait/reth.py
@@ -1,19 +1,11 @@
-from dataclasses import dataclass
-from typing import Any
-
-from utils.wait.base import BaseWaiter
+from utils.wait.base import RpcWaiter
 
 
-@dataclass
-class RethWaiter(BaseWaiter[Any]):
-    """
-    Wrapper for encapsulating and waiting reth related rpcs
-    """
-
+class RethWaiter(RpcWaiter):
     def wait_until_eth_block_exceeds(self, height, message: str | None = None):
         message = message or f"Timeout: waiting for block at height {height}"
         return self._wait_until_with_value(
-            lambda: int(self.inner.eth_blockNumber(), 16),
+            lambda: int(self.rpc_client.eth_blockNumber(), 16),
             lambda value: value > height,
             error_with=message,
             timeout=self.timeout,
@@ -25,7 +17,7 @@ class RethWaiter(BaseWaiter[Any]):
         Waits until eth block number reaches at least the specified height.
         """
         return self._wait_until_with_value(
-            lambda: int(self.inner.eth_blockNumber(), 16),
+            lambda: int(self.rpc_client.eth_blockNumber(), 16),
             lambda value: value >= height,
             error_with=message or f"Timeout: waiting for block height {height}",
             timeout=self.timeout,
@@ -36,11 +28,11 @@ class RethWaiter(BaseWaiter[Any]):
         """
         Get the current block number from reth RPC.
         """
-        return int(self.inner.eth_blockNumber(), 16)
+        return int(self.rpc_client.eth_blockNumber(), 16)
 
     def wait_until_state_diff_at_blockhash(self, blockhash, timeout: None | int = None):
         return self._wait_until_with_value(
-            lambda: self.inner.strataee_getBlockStateDiff(blockhash),
+            lambda: self.rpc_client.strataee_getBlockStateDiff(blockhash),
             lambda value: value is not None,
             error_with="Finding non empty statediff for blockhash {blockhash} timed out",
             timeout=timeout or self.timeout,
@@ -49,7 +41,7 @@ class RethWaiter(BaseWaiter[Any]):
     def wait_until_block_witness_at_blockhash(self, blockhash, timeout: None | int = None):
         return self._wait_until_with_value(
             # TODO: parameterize True
-            lambda: self.inner.strataee_getBlockWitness(blockhash, True),
+            lambda: self.rpc_client.strataee_getBlockWitness(blockhash, True),
             lambda value: value is not None,
             error_with="Finding non empty witness for blockhash {blockhash} timed out",
             timeout=timeout or self.timeout,
--- a/functional-tests/tests/prover/prover_cl_dispatch.py
+++ b/functional-tests/tests/prover/prover_cl_dispatch.py
@@ -33,7 +33,6 @@ class ProverClientTest(testenv.StrataTestBase):
         reth_waiter = self.create_reth_waiter(reth.create_rpc())
         reth_waiter.wait_until_eth_block_at_least(20)
 
-
         # Dispatch the prover task
         start_block_id = cl_slot_to_block_id(seqrpc, CL_PROVER_PARAMS["start_block"])
         start_block_commitment = {"slot": CL_PROVER_PARAMS["start_block"], "blkid": start_block_id}
--- a/functional-tests/utils/wait/strata.py
+++ b/functional-tests/utils/wait/strata.py
@@ -1,16 +1,10 @@
-from dataclasses import dataclass
 from typing import Any
 
 from factory.seqrpc import RpcError
-from utils.wait.base import BaseWaiter
+from utils.wait.base import RpcWaiter
 
 
-@dataclass
-class StrataWaiter(BaseWaiter[Any]):
-    """
-    Wrapper for encapsulating and waiting strata related rpcs
-    """
-
+class StrataWaiter(RpcWaiter):
     def wait_until_genesis(self, message: str | None = None):
         """
         Waits until we see genesis. That is to say, that `strata_syncStatus`
@@ -22,7 +16,7 @@ class StrataWaiter(BaseWaiter[Any]):
         def _check_genesis():
             try:
                 # This should raise if we're before genesis.
-                ss = self.inner.strata_syncStatus()
+                ss = self.rpc_client.strata_syncStatus()
                 self.logger.info(
                     f"after genesis, tip is slot {ss['tip_height']} blkid {ss['tip_block_id']}"
                 )
@@ -53,15 +47,15 @@ class StrataWaiter(BaseWaiter[Any]):
         self.logger.info(f"waiting for epoch {epoch}")
 
         def _query():
-            status = self.inner.strata_syncStatus()
+            status = self.rpc_client.strata_syncStatus()
             self.logger.debug(f"checked status {status}")
-            commitments = self.inner.strata_getEpochCommitments(epoch)
+            commitments = self.rpc_client.strata_getEpochCommitments(epoch)
             if len(commitments) > 0:
                 comm = commitments[0]
                 self.logger.info(
                     f"now at epoch {epoch}, slot {comm['last_slot']}, blkid {comm['last_blkid']}"
                 )
-                return self.inner.strata_getEpochSummary(
+                return self.rpc_client.strata_getEpochSummary(
                     epoch,
                     comm["last_slot"],
                     comm["last_blkid"],
@@ -89,10 +83,10 @@ class StrataWaiter(BaseWaiter[Any]):
 
         Returns the new epoch number.
         """
-        init_epoch = self.inner.strata_syncStatus()["cur_epoch"]
+        init_epoch = self.rpc_client.strata_syncStatus()["cur_epoch"]
 
         def _query():
-            ss = self.inner.strata_syncStatus()
+            ss = self.rpc_client.strata_syncStatus()
             self.logger.info(f"waiting for next epoch, ss {ss}")
             return ss["cur_epoch"]
 
@@ -122,7 +116,7 @@ class StrataWaiter(BaseWaiter[Any]):
         """
 
         def _check():
-            cs = self.inner.strata_clientStatus()
+            cs = self.rpc_client.strata_clientStatus()
             l1_height = cs["tip_l1_block"]["height"]
             conf_epoch = cs["confirmed_epoch"]
             self.logger.info(f"confirmed epoch as of {l1_height}: {conf_epoch}")
@@ -141,7 +135,7 @@ class StrataWaiter(BaseWaiter[Any]):
         Waits until strata chain tip exceeds the given height.
         """
         return self._wait_until_with_value(
-            lambda: self.inner.strata_syncStatus(),
+            lambda: self.rpc_client.strata_syncStatus(),
             lambda stat: stat["tip_height"] > height,
             error_with=msg or "Timeout: expected number of blocks are not being created",
             timeout=timeout,
@@ -160,7 +154,7 @@ class StrataWaiter(BaseWaiter[Any]):
         """
 
         def _check():
-            cs = self.inner.strata_clientStatus()
+            cs = self.rpc_client.strata_clientStatus()
             l1_height = cs["tip_l1_block"]["height"]
             fin_epoch = cs["finalized_epoch"]
             self.logger.info(f"finalized epoch as of {l1_height}: {fin_epoch}")
@@ -186,7 +180,7 @@ class StrataWaiter(BaseWaiter[Any]):
         message = message or "Strata client did not start on time"
 
         self._wait_until(
-            lambda: self.inner.strata_protocolVersion() is not None,
+            lambda: self.rpc_client.strata_protocolVersion() is not None,
             error_with=message,
             timeout=timeout,
             step=interval,
@@ -205,7 +199,7 @@ class StrataWaiter(BaseWaiter[Any]):
         """
 
         def _check():
-            ss = self.inner.strata_syncStatus()
+            ss = self.rpc_client.strata_syncStatus()
             slot = ss["tip_height"]  # TODO rename to tip_slot
             of_epoch = ss["observed_finalized_epoch"]
             self.logger.info(f"observed final epoch as of L2 slot {slot}: {of_epoch}")
@@ -234,7 +228,7 @@ class StrataWaiter(BaseWaiter[Any]):
         """
 
         def _check():
-            ss = self.inner.strata_syncStatus()
+            ss = self.rpc_client.strata_syncStatus()
             slot = ss["tip_height"]  # TODO rename to slot
             epoch = ss["cur_epoch"]
             view_l1 = ss["safe_l1_block"]["height"]
@@ -262,7 +256,7 @@ class StrataWaiter(BaseWaiter[Any]):
         message = message or "L1 reader did not catch up with bitcoin network"
 
         return self._wait_until_with_value(
-            lambda: self.inner.strata_l1status(),
+            lambda: self.rpc_client.strata_l1status(),
             lambda value: value["cur_height"] >= height,
             error_with=message,
             timeout=timeout,
@@ -282,7 +276,7 @@ class StrataWaiter(BaseWaiter[Any]):
         message = message or "Blocks not generated"
 
         return self._wait_until_with_value(
-            lambda: self.inner.strata_getRecentBlockHeaders(height),
+            lambda: self.rpc_client.strata_getRecentBlockHeaders(height),
             lambda value: value is not None,
             error_with=message,
             timeout=timeout,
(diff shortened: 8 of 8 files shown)


functional-tests/mixins/bridge_mixin.py
```diff
@@ -181,13 +181,16 @@ def make_drt(self, el_address, musig_bridge_pk):
...
         time.sleep(1)
 
         # time to mature DRT
         self.btcrpc.proxy.generatetoaddress(6, seq_addr)
+        # FIXME: what are we waiting for here?
         time.sleep(3)
 
         # time to mature DT
         self.btcrpc.proxy.generatetoaddress(6, seq_addr)
+        # FIXME: what are we waiting for here?
```
bewakes: Not sure what the waiting condition should be here.

delbonis: Transactions being included in blocks.  Need some way to wait on that.

bewakes: `generatetoaddress` should include the txs in the block. Perhaps this is waiting for something else, like bridge operator or strata client seeing it as included?

functional-tests/tests/load/basic_load.py
```diff
@@ -17,8 +16,12 @@ def main(self, ctx: flexitest.RunContext):
         prover_client_rpc = prover_client.create_rpc()
         rethrpc = reth.create_rpc()
 
-        # Wait for some blocks with transactions to be generated.
-        time.sleep(30)
+        # Wait for a couple of blocks(3) with transactions to be generated.
```
evgenyzdanovich: The original intent **was to have 20-30 blocks generated** (while sleeping) and filled with transactions. That's a special kind of env where transactions are spawned automatically


So to preserve this idea, I'd 

```

wait_until(
            lambda: int(rethrpc.eth_blockNumber(), base=16) > 25,
            error_with="Blocks not generated",
            timeout=30,
        )
        
```

functional-tests/tests/el_stateroot_from_state_diffs.py
```diff
@@ -13,6 +13,7 @@ def __init__(self, ctx: flexitest.InitContext):
     def main(self, ctx: flexitest.RunContext):
         reth = ctx.get_service("reth")
         rethrpc = reth.create_rpc()
+        # FIXME: what is this waiting for?
```
evgenyzdanovich: Short answer: to produce some blocks. 

You see, this test runs against a special load env, which mean that transactions are produced automatically and blocks are not empty. This sleep basically allow for the blocks to be produced and calc state root from state diffs.

**TL;DR: With regards to your task - here you can fetch the latest block number at the beginning of the test and wait until the chain produces 20 more blocks from than point onwards**

evgenyzdanovich (commented):
Look good, great work!

Not approving just yet because of two reasons:
- one test is consistently failing. I briefly took a look and found no obvious reason (and the test passes for me locally on this branch). Prover logs are clean as well. Ideally, we gotta figure out what's happening
- the PR is in draft, so I assume the work is still being done here

functional-tests/tests/prover/prover_checkpoint_latest.py
```diff
@@ -13,6 +15,7 @@ def main(self, ctx: flexitest.RunContext):
         prover_client = ctx.get_service("prover_client")
         prover_client_rpc = prover_client.create_rpc()
 
+        time.sleep(1)
```
bewakes: @evgenyzdanovich , I added this sleep here and it seems to pass. Actually I should have added this to `prover_checkpoint_runner.py` because that's where I removed this `time.sleep(1)`. But anyway this passes.

The `wait_until` just below this statement should ensure that prover client is ready, isn't it? There's something I can't figure out. :)

functional-tests/tests/prover/prover_checkpoint_runner.py
```diff
@@ -41,7 +41,6 @@ def main(self, ctx: flexitest.RunContext):
 
         # Wait until the prover client reports readiness
         # TODO this should be done in the env init
-        time.sleep(1)
```
bewakes, line 44: Here @evgenyzdanovich, I had removed the sleep from here after which it started failing and apparently adding it to another test case passes.

functional-tests/tests/el_stateroot_from_state_diffs.py
```diff
@@ -13,9 +12,16 @@ def __init__(self, ctx: flexitest.InitContext):
...
-        time.sleep(20)
 
-        block = int(rethrpc.eth_blockNumber(), base=16)
+        # Get initial block number and wait for 20 more blocks to be generated
+        initial_block = get_latest_eth_block_number(rethrpc)
+        wait_until(
+            lambda: get_latest_eth_block_number(rethrpc) >= initial_block + 20,
+            error_with="Timeout: 20 blocks were not generated",
+            timeout=60,
+        )
```
delbonis: This should probably be refactored out into the utils library.

functional-tests/tests/btcio/btcio_read.py
```diff
@@ -31,30 +27,27 @@ def main(self, ctx: flexitest.RunContext):
...
 
-        time.sleep(3)
-
         received_block = btcrpc.getblock(btcrpc.proxy.getbestblockhash())
-        l1stat = seqrpc.strata_l1status()
+        l1stat = wait_until_with_value(
+            lambda: seqrpc.strata_l1status(),
+            lambda value: value["cur_height"] >= received_block["height"],
+            error_with="L1 reader did not catch up with bitcoin network",
+        )
```
delbonis: This should definitely be refactored out into the utils library, as a "wait until the strata node reaches the bitcoin node as of when this function was called" function.

functional-tests/tests/btcio/btcio_read.py
```diff
@@ -31,30 +27,27 @@ def main(self, ctx: flexitest.RunContext):
...
         # generate 2 more btc blocks
         generate_n_blocks(btcrpc, 2)
-        time.sleep(MAX_HORIZON_POLL_INTERVAL_SECS * 2)
 
-        next_l1stat = seqrpc.strata_l1status()
+        next_l1stat = wait_until_with_value(
+            lambda: seqrpc.strata_l1status(),
+            lambda value: value["cur_height"] >= cur_l1_height + 2,
+            error_with="L1 reader did not catch up with bitcoin network",
+        )
```
delbonis: See above previous comment.  You could add a optional parameter to also add some offset to the computed target height.

functional-tests/tests/el_block_generation.py
```diff
@@ -23,9 +23,11 @@ def main(self, ctx: flexitest.RunContext):
...
 
         for _ in range(5):
-            time.sleep(3)
-            cur_blocknum = int(rethrpc.eth_blockNumber(), 16)
+            cur_blocknum = wait_until_with_value(
+                lambda: int(rethrpc.eth_blockNumber(), 16),
+                lambda value, last_blk=last_blocknum: value > last_blk,
+                error_with="Timeout: seem to not be making progress",
+                timeout=3,
+            )
```
delbonis: Should be refactored out.

functional-tests/tests/el_block_state_diff.py
```diff
@@ -33,12 +32,12 @@ def main(self, ctx: flexitest.RunContext):
...
-        # Get the state diff data
-        state_diff_data = rethrpc.strataee_getBlockStateDiff(blockhash)
-        assert state_diff_data is not None, "non empty state diff"
-
+        state_diff_data = wait_until_with_value(
+            lambda: rethrpc.strataee_getBlockStateDiff(blockhash),
+            lambda value: value is not None,
+            error_with="Finding non empty statediff for blockhash {blockhash} timed out",
+            timeout=2,
+        )
```
delbonis: Could maybe be refactored out.

functional-tests/tests/el_block_witness_data.py
```diff
@@ -33,12 +32,12 @@ def main(self, ctx: flexitest.RunContext):
...
-        # Get the witness data
-        witness_data = rethrpc.strataee_getBlockWitness(blockhash, True)
-        assert witness_data is not None, "non empty witness"
-
+        witness_data = wait_until_with_value(
+            lambda: rethrpc.strataee_getBlockWitness(blockhash, True),
+            lambda value: value is not None,
+            error_with="Finding non empty witness for blockhash {blockhash} timed out",
+            timeout=2,
+        )
```
delbonis: Also maybe.

functional-tests/tests/el_sync_from_chainstate.py
delbonis, line 1: Definitely a lot of repetitive structure here that can be refactored out.

functional-tests/tests/load/basic_load.py
```diff
@@ -17,10 +16,14 @@ def main(self, ctx: flexitest.RunContext):
...
         rethrpc = reth.create_rpc()
 
-        # Wait for some blocks with transactions to be generated.
-        time.sleep(30)
+        # Wait for a some blocks with transactions to be generated.
+        wait_until(
+            lambda: eth_block_number(rethrpc) > 25,
+            error_with="Blocks not generated",
+            timeout=30,
+        )
```
delbonis: This has similar structure as in the other test.

functional-tests/tests/prover/prover_checkpoint_runner.py
```diff
@@ -52,14 +51,14 @@ def main(self, ctx: flexitest.RunContext):
...
+        def consecutive_checkpoints_verified():
             ckpt_idx = sequencer_rpc.strata_getLatestCheckpointIndex(True)
             logging.info(f"cur checkpoint idx: {ckpt_idx}")
             return ckpt_idx == PROVER_CHECKPOINT_SETTINGS["CONSECUTIVE_PROOFS_REQUIRED"]
 
         # Wait until the required number of consecutive checkpoint proofs are generated and verified
         wait_until(
-            _ck1,
+            consecutive_checkpoints_verified,
             timeout=PROVER_CHECKPOINT_SETTINGS["PROVER_TIMEOUT_SECONDS"],
```
delbonis: Really should avoid depending on globals like this, pass it through the context.

functional-tests/tests/prover/prover_el_deposit_withdraw.py
```diff
@@ -150,7 +151,7 @@ def test_checkpoint(self, l1_block, l2_block, prover_client_rpc, rethrpc):
         # Given that L1 blocks are happening more frequent that L2, it's safe
         # to assert only L2 latest block.
         wait_until(
-            lambda: int(rethrpc.eth_blockNumber(), base=16) > l2[1],
+            lambda: get_latest_eth_block_number(rethrpc) > l2[1],
             timeout=60,
         )
```
delbonis: Similar repetitive structure, I think.

functional-tests/tests/rpc_exec_update.py
```diff
@@ -15,9 +14,12 @@ def main(self, ctx: flexitest.RunContext):
...
         seqrpc = seq.create_rpc()
-        time.sleep(2)
-
-        recent_blks = seqrpc.strata_getRecentBlockHeaders(1)
+        recent_blks = wait_until_with_value(
+            lambda: seqrpc.strata_getRecentBlockHeaders(1),
+            lambda value: value is not None,
+            error_with="Blocks not generated",
+            timeout=2,
+        )
```
delbonis: I think there should already be a fn for this?  Not sure.  But really this should be phrased as a "wait for next block" fn.

functional-tests/tests/rpc_recent_blocks.py
```diff
@@ -18,14 +17,11 @@ def main(self, ctx: flexitest.RunContext):
...
-            if blk is None:
-                counter += 1
-                time.sleep(1)
-            else:
-                break
+        wait_until(
+            lambda: seqrpc.strata_getHeadersAtIdx(NUM_BLOCKS_TO_RECEIVE) is not None,
+            error_with=f"Expected block {NUM_BLOCKS_TO_RECEIVE} not generated",
+            timeout=20,
+        )
```
delbonis: There's some slight nuance here because the constant is phrased as if it's expecting to get n blocks, but actually we're just checking to see if the nth block has been generated.  If this test runs later in the sequence and we're on block 100, then this won't be checking what we expect it to check, which I would assume to be "block 100+10 has been generated".

delbonis (commented):
A few things:

I see this new `wait_until_strata_client_ready` fn that's called in many tests.  This should really be done in the env init instead of the tests, since if that fails it's a fault of the env not the test.  And if we do the same thing the same at the start of every test, then that's a sign that there's redundancy that could be unified.  We could also have a helper for the crash tests that is like a "restart and wait until healthy" fn, to unify it into one step.

The other thing, while we're here, is that we can see in the utils module that there's a ton of similar functions that take similar sets of arguments.  I am not sure how to structure this, but it really feels like we could define some types that unify related kinds of waits (maybe like `BitcoinWaiter`, `StrataWaiter`, `RethWaiter`, but also some kind of `ReaderWaiter` that takes both the Bitcoin and Strata RPCs, and maybe something similar for the Strata+Reth interfacing, if needed), so that we don't have to import each fn explicitly, pass in the RPCs each time we call them, and maybe they can share some defaults like the timeout or polling frequency.

functional-tests/tests/el_sync_from_chainstate.py
bewakes: Yep!

functional-tests/tests/rpc_exec_update.py
bewakes: No there wasn't. Created one.

functional-tests/tests/prover/prover_checkpoint_latest.py
evgenyzdanovich: Well, I have a wild hypothesis here. The problem **may be** that the witness not being generated quick enough (for the latest block)  and thus the checkpoint simply fails. Adding a sleep actually **may** make a difference here. 

If that hypothesis is correct, we can come up with some way to wait on that...

functional-tests/tests/prover/prover_checkpoint_runner.py
evgenyzdanovich: same here, the same hypothesis applies here.

Ideally, I'd like to see the logs of the `prover-client` that's related to running certain test OR **having at least the time at which a certain test has started** (this way I can at least time-correlate the prover logs to the test). Right now the logs are a bit messy.

evgenyzdanovich (approved):
Looks good to me in the current form. 

I'm not sure how if you would like to address the last comments from me and dig very deep into prover tests, thus I aint gonna block this PR either.

delbonis (changes requested):
Still waiting on the simplification.

Alternatively, instead of having some "waiter" family of types, I think it might make more sense to put these wait functions on the env types, maybe as a mixin?.

evgenyzdanovich (approved):

functional-tests/tests/btcio/btcio_read_reorg.py
```diff
@@ -18,33 +15,33 @@ def __init__(self, ctx: flexitest.InitContext):
...
-        to_be_invalid_block = seqrpc.strata_getL1blockHash(invalidate_height)
+        to_be_invalid_block = seq_rpc.strata_getL1blockHash(invalidate_height)
         # Wait for at least 1 block to be added after invalidating `REORG_DEPTH` blocks.
-        time.sleep(BLOCK_GENERATION_INTERVAL_SECS * 1 + SEQ_SLACK_TIME_SECS)
-        block_from_invalidated_height = seqrpc.strata_getL1blockHash(invalidate_height + 1)
+        block_from_invalidated_height = wait_until_with_value(
+            lambda: seq_rpc.strata_getL1blockHash(invalidate_height + 1),
+            lambda value: value is not None,
+            error_with="L1 Block not produced in time",
+        )
```
delbonis: This needs to be brought into the waiter.

functional-tests/tests/cl_block_witness.py
```diff
@@ -17,18 +15,13 @@ def main(self, ctx: flexitest.RunContext):
...
         witness_1 = self.get_witness(seqrpc, 1)
         assert witness_1 is not None
 
-        time.sleep(1)
-        witness_2 = self.get_witness(seqrpc, 2)
-        assert witness_2 is not None
+        wait_until(
+            lambda: self.get_witness(seqrpc, 2) is not None,
+            error_with="Failed to get cl witness in time",
+        )
```
delbonis, line 24: This needs to be brought into the waiter.

functional-tests/tests/el_stateroot_from_state_diffs.py
```diff
@@ -13,9 +13,16 @@ def __init__(self, ctx: flexitest.InitContext):
...
         rethrpc = reth.create_rpc()
-        time.sleep(20)
+        reth_waiter = RethWaiter(rethrpc, timeout=60)
+
+        # Get initial block number and wait for 20 more blocks to be generated
+        initial_block = get_latest_eth_block_number(rethrpc)
+        _ = reth_waiter.wait_until_eth_block_exceeds(
+            initial_block + 20,
+            message="Timeout: 20 blocks were not generated",
+        )
```
delbonis, line 22: This needs to be brought into the waiter.

functional-tests/tests/prover/prover_checkpoint_runner.py
```diff
@@ -38,29 +44,28 @@ def main(self, ctx: flexitest.RunContext):
...
-            return ckpt_idx == PROVER_CHECKPOINT_SETTINGS["CONSECUTIVE_PROOFS_REQUIRED"]
+            return ckpt_idx == self.checkpoint_settings.consecutive_proofs_required
 
         # Wait until the required number of consecutive checkpoint proofs are generated and verified
         wait_until(
-            _ck1,
-            timeout=PROVER_CHECKPOINT_SETTINGS["PROVER_TIMEOUT_SECONDS"],
+            consecutive_checkpoints_verified,
+            timeout=self.checkpoint_settings.prover_timeout_seconds,
         )
```
delbonis, line 68: This needs to be brought into the waiter.

functional-tests/tests/prover/prover_checkpoint_runner.py
```diff
@@ -38,29 +44,28 @@ def main(self, ctx: flexitest.RunContext):
...
         sequencer_rpc = sequencer.create_rpc()
+        seq_waiter = StrataWaiter(sequencer_rpc, self.logger, timeout=60)
 
         # Wait until the prover client reports readiness
-        # TODO this should be done in the env init
-        time.sleep(1)
         wait_until(
             lambda: prover_rpc.dev_strata_getReport() is not None,
             error_with="Prover did not start on time",
         )
```
delbonis, line 52: Also this.

functional-tests/tests/prover/prover_checkpoint_runner.py
```diff
@@ -73,11 +78,11 @@ def _ck1():
...
-            return ckpt_idx == PROVER_CHECKPOINT_SETTINGS["CONSECUTIVE_PROOFS_REQUIRED"] * 2
+            return ckpt_idx == self.checkpoint_settings.consecutive_proofs_required
 
         # Wait until another portion of consecutive proofs are generated and verified
         # after the restart of the sequencer.
         wait_until(
             _ck2,
-            timeout=PROVER_CHECKPOINT_SETTINGS["PROVER_TIMEOUT_SECONDS"],
+            timeout=self.checkpoint_settings.prover_timeout_seconds,
         )
```
delbonis, line 87: And this. :P

functional-tests/tests/rpc_el_inactive.py
```diff
@@ -55,41 +52,34 @@ def main(self, ctx: flexitest.RunContext):
...
 
         # check if new blocks are being created again
         cur_slot = seqrpc.strata_clientStatus()["chain_tip_slot"]
 
-        wait_until(
-            lambda: seqrpc.strata_clientStatus()["chain_tip_slot"] > cur_slot,
-            error_with="New blocks are not being created",
+        seq_waiter.wait_until_chain_tip_exceeds(
+            cur_slot,
+            msg="New blocks are not being created",
```
delbonis, line 78: We don't really need to pass in this message arg each time.  We can make a message like this a default parameter.

This is actually somewhat incorrect.  It says that new blocks aren't being created, but it's possible that new blocks *are* being created, but not fast enough to reach the target before the deadline.  We should correct this (in the waiter) and add maybe add the slots in the exception message.

functional-tests/tests/rpc_recent_blocks.py
```diff
@@ -18,21 +17,18 @@ def main(self, ctx: flexitest.RunContext):
...
-            else:
-                break
-
-        recent_blks = seqrpc.strata_getRecentBlockHeaders(NUM_BLOCKS_TO_RECEIVE)
-        assert len(recent_blks) == NUM_BLOCKS_TO_RECEIVE
+        wait_until(
+            lambda: seqrpc.strata_getHeadersAtIdx(EXPECTED_L2_BLOCKS) is not None,
+            error_with=f"Expected block {EXPECTED_L2_BLOCKS} not generated",
+            timeout=20,
+        )
```
delbonis: This really needs to be moved into the waiter!

functional-tests/tests/sync/sync_genesis.py
```diff
@@ -17,20 +17,23 @@ def main(self, ctx: flexitest.RunContext):
...
         last_slot = 0
         for _ in range(5):
-            time.sleep(3)
-            stat = seqrpc.strata_syncStatus()
+            stat = wait_until_with_value(
+                lambda: seqrpc.strata_syncStatus(),
+                lambda value, last_slt=last_slot: value["tip_height"] > last_slt,
+                error_with="seem not to be making progress",
+                timeout=3,
+            )
```
delbonis: This really really needs to be moved into the waiter!

functional-tests/utils/wait/__init__.py
```diff
@@ -0,0 +1,27 @@
+"""
+Wait module for functional tests.
+
+This module provides waiter classes for different components:
+- RethWaiter: Ethereum/Reth RPC waiting operations
+- StrataWaiter: Strata L2 chain waiting operations
+- ProverWaiter: Proof generation waiting operations
```
delbonis: Stating this here is redundant isn't it?  We have the types imported, they have docs don't they?  I am not quite sure of the conventions on pydoc.

bewakes:
> Part of the issue with maintaining tests is because we all have a tendency to discount them because they're "just" tests, but it's just as important to maintain test infra as it is to maintain the main featureful code as projects grow.

Couldn't agree more!

Here, I've now made some more changes. PTAL

functional-tests/utils/wait/__init__.py
bewakes: Cleaned it up a bit

functional-tests/utils/wait/base.py
```diff
@@ -0,0 +1,18 @@
...
+@dataclass
+class BaseWaiter(ABC):
+    rpc: Any
+    logger: Logger
+    timeout: int = 10
+    interval: float = 0.5
+
+    def __post_init__(self):
+        self.wait_until = wait_until
+        self.wait_until_with_value = wait_until_with_value
```
delbonis: What's happening here?  I'm not sure how it makes sense for these to be local fns like this.

Shouldn't we be doing something like

```py
def _wait_until(self, ...):
  utils.wait_until(..., timeout=self.timeout, interval=self.interval)
```

instead?

functional-tests/utils/wait/base.py
```diff
@@ -0,0 +1,18 @@
...
+from logging import Logger
+from typing import Any
+
+from utils.utils import wait_until, wait_until_with_value
+
+
+@dataclass
+class BaseWaiter(ABC):
+    rpc: Any
+    logger: Logger
```
delbonis: It's not useful to declare this `rpc` field here, since it's not any RPC in particular, *and* it becomes odd in a waiter with two RPC handles (which one should be the super field and which should be a new field?).

functional-tests/utils/wait/prover.py
```diff
@@ -0,0 +1,70 @@
...
+from utils.wait.base import BaseWaiter
+
+
+@dataclass
+class ProverWaiter(BaseWaiter):
+    """
+    Wrapper for encapsulating and waiting prover related operations
+    """
+
+    message: str = "Timeout: waiting for proof completion"
```
delbonis: Why is this a field instead of a const?

functional-tests/utils/wait/reth.py
```diff
@@ -0,0 +1,56 @@
...
+from utils.wait.base import BaseWaiter
+
+
+@dataclass
+class RethWaiter(BaseWaiter):
+    """
+    Wrapper for encapsulating and waiting reth related rpcs
+    """
+
+    message: str = "Timeout: waiting for blocks"
```
delbonis: (see other comment)

functional-tests/envs/testenv.py
```diff
@@ -28,6 +29,16 @@ def premain(self, ctx: flexitest.RunContext):
...
+        self.logger = logger
+
+    def create_reth_waiter(self, reth_rpc, timeout=20, interval=0.5, **kwargs) -> RethWaiter:
+        return RethWaiter(reth_rpc, self.logger, timeout, interval, **kwargs)
+
+    def create_strata_waiter(self, strata_rpc, timeout=20, interval=0.5, **kwargs) -> StrataWaiter:
+        return StrataWaiter(strata_rpc, self.logger, timeout, interval, **kwargs)
+
+    def create_prover_waiter(self, prover_rpc, timeout=20, interval=0.5, **kwargs) -> ProverWaiter:
+        return ProverWaiter(prover_rpc, self.logger, timeout, interval, **kwargs)
```
delbonis, line 41: These waiters aren't really being provided by the *test case*, they're more derived from the *test env* being passed into the `main` fn, since that's what provides us the RPCs that they're principally wrappers around.  So these fns make more sense to be on the env like I mentioned, taking the logger passed in or using the default logger fns.

functional-tests/utils/wait/prover.py
bewakes: This wasn't even being used. The original idea was to be able to provide a default message during waiter initialization. But that isn't really needed.

functional-tests/utils/wait/base.py
bewakes: The functions wouldn't be local, they are accessible from every child class. I was trying to not pass around the args. But since the `timeout or self.timeout` and `interval or self.interval` pattern is very common, I have eventually defined it this way. :)

functional-tests/utils/wait/base.py
bewakes: It is in fact useful, only that the name might be misleading. This allows for cleanly inheriting this base class. If we don't have this "rpc" here and try to create one for each child class, then there's issues with default arguments. Like you can't do:

```python
@dataclass
class Base:
  b: str
  c: int = 10

# This is error, because non default param `a` defined after default param `c`
@dataclass
class Child(Base):
  a: str # Error
```
We really want to pass the rpc as the first parameter and the rest afterwards, possibly defaulted.

This can be worked around by having custom `__init__`, but that's ugly and doesn't give benefits of dataclass.

---

So I've now changed the name to be `inner` instead of `rpc`, retaining it as the first argument and also making `BaseWaiter` generic. Later, if we need two rpcs, then we can define a class for that and pass it to the constructor.

```python
@dataclass
class BaseWaiter[T]:
  inner: T
  logger: Logger
  ...  # other defaults

# If we want multiple rpcs then
@dataclass
class MultiRPC:
  rpc_a: Any
  rpc_b: Any
  rpc_c: Any

@dataclass
class MultiWaiter(BaseWaiter[MultiRPC]):
   def wait_until_something(self):
      self.wait_until(self.inner.rpc_a.is_ready)
```

We can't fully benefit from this because `service.create_rpc()` is not typed, but this should give us all the flexibility and ergonomics.

functional-tests/envs/testenv.py
bewakes: I can understand your point. However, it also makes (more)sense for them to be on the test case for these reasons:
1. While it's true that we are waiting on rpcs provided by environment right now, it doesn't always have to be the case. We might be waiting for/wrapping something else not coming from the env.
2. This gives us nicer ergonomics. We already have `self.logger`, we don't have to pass it around.
3. We get better customization per test.

delbonis: 1. We will always be waiting on "something" to happen in the env.
2. It gives us nicer ergonomics at the cost of ergonomics.  We have to fetch the service from the env and create the RPC handle from the service.
3. We can still do this with helper fns in the test modules.

functional-tests/utils/wait/base.py
delbonis: >We really want to pass the rpc as the first parameter and the rest afterwards, possibly defaulted.

Is it really important that the RPC is the first parameter?  Especially if we're having factory-like methods that init these objects?  You *can* put `@dataclass` on derived classes and it passes through the kwargs for the base's fields.

I would argue that this probably could go without being a dataclass, since it's not just a dumb data container.  A field being defined by the base class but not used in that class is minor red flag.  It's over-specializing the base class, as I mentioned above.

bewakes: I agree. The class is more complicated than necessary. I didn't want to repeat the `__init__` definitions everywhere. But we can always use a base class for that! :D 

I have now changed the base waiter to be a mixin that exposes `wait_until`s and created a base class to wrap a single RPC to wait on.

functional-tests/envs/testenv.py
bewakes: For 2, we do need to create rpc handles as they are not only used by the waiters, but by the test logic as well. So, that shouldn't be a problem.
I think this is good. We should now be working on more important things.

bewakes:
What should we do with the two failing tests(`prover_checkpoint_latest` and `procer_cl_dispatch`)? Those are failing in main as well.

I think we should just create a new ticket to fix those and not get this PR through.

functional-tests/utils/wait/base.py
delbonis: > I have now changed the base waiter to be a mixin that exposes wait_untils and created a base class to wrap a single RPC to wait on.

Looking at the new design, this overcomplicating it now.  `WaiterMixin` is easier to misuse.  I will fix it later.

functional-tests/envs/testenv.py
delbonis: Furthermore, putting these fns in `StrataTestBase` makes the implication that certain types of services are always expected to be present in the env it's being called with.  Especially relevant with the prover service as that's only present in a few tests, but will also apply to reth in the near/mid future.

bewakes:
@delbonis let's merge this PR now. The failing tests need to be fixed in a separate ticket: they exist in main branch as well.

Please feel free to fix things later in another PR as you mentioned in the comment.

PR #873, opened 2025-06-12
STR-1301: Adjust log level to info in case of RPC errors in prover client.

evgenyzdanovich:
## Description

Adjust log level to info in case of RPC errors in prover client.
Such errors are somewhat expected - `exex`'s witness generation sometimes lags behind the block production.
Changing the level to info makes more sense to not pollute logs with false positives.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


storopoli:
> @storopoli it's a transient error (**that gets retried**), there's little-to-no benefit of having our logs polluted with errors, if it's not something critical. Every such error message triggers our alerting system.
> 
> > then we need to handle properly the lag.
> 
> fixing/optimizing exex is a whole new task (quite complicated) that we gonna handle in the future.

I understand everything. But, still, from first principles, I disagree. I'm not gonna block this but I won't be ACK'ing. I am dismissing my review.

MdTeach (approved):

PR #876, opened 2025-06-12
Fix MMR audit reports

delbonis:
## Description

This PR addresses:

* STR-1406
* STR-1408
* STR-1409

As well as several general cleanup to various things.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


crates/util/mmr/src/hasher.rs
```diff
@@ -1,22 +1,121 @@
...
+    fn zero_hash() -> Self::Hash {
+        <Self::Hash as MerkleHash>::zero()
+    }
 }
 
+/// Generic impl over [`Digest`] impls, where hash is `[u8; 32]`.
+// TODO make this generic over other hash types
 impl<D: Digest> MerkleHasher for D {
-    type Item = Hash;
+    type Hash = Hash32;
```
delbonis: A thing missing here is that we should make this impl a little safer.  It assumes the digest always uses 32 byte outputs, but that's not always true for any `Digest`, obviously.  Need some clever way to address this.

crates/util/mmr/src/hasher.rs
```diff
@@ -1,22 +1,121 @@
...
+        // isn't actually relevant for security at all, it's just good practice,
+        // so we can avoid pulling in additional dependencies.  This is
+        // primarily used when verifying a root computed from a proof against a
+        // trusted one, and presumably the party that might be probing us
+        // already knows what the known-good root is if they're giving us a
+        // proof against it.
+        let mut acc: u32 = 0;
+        for i in 0..LEN {
+            acc += (a[i] ^ b[i]) as u32;
+        }
```
bewakes, line 55: We don't need to use sum and `u32`s if we use bitwise OR which should be faster.
```suggestion
        let mut acc: u8 = 0;
        for i in 0..LEN {
            acc |= (a[i] ^ b[i]);
        }
```

crates/util/mmr/src/hasher.rs
```diff
@@ -1,22 +1,121 @@
...
 
+/// 32 byte hash.
+pub type Hash32 = [u8; 32];
+
+type Tag = [u8; 64];
+
+lazy_static! {
+    static ref NODE_TAG_PREFIX: Tag = make_tag(b"node");
+    static ref LEAF_TAG_PREFIX: Tag = make_tag(b"leaf");
+}
```
bewakes: Is there a specific reason why the tag is of length 64? It seems that if this for domain separation a couple bytes(or even single) should be sufficient.

But perhaps this is to be agnostic of the hash function which might not guarantee the "avalanche effect" as sha family. Is that true?

bewakes (commented):
Looks good overall. Some suggestions and questions.

Cargo.toml
```diff
@@ -258,6 +258,7 @@ http = "1.0.0"
 hyper = "0.14.25"
 jsonrpsee = "0.24"
 jsonrpsee-types = "0.24"
+lazy_static = "1.5"
```
Zk2u: Unless we're targeting an MSRV of <1.70.0, you can do this with `std::cell::OnceCell` or `std::sync::OnceLock`

crates/util/mmr/Cargo.toml
```diff
@@ -7,10 +7,8 @@ version = "0.3.0-alpha.1"
 workspace = true
 
 [dependencies]
-arbitrary.workspace = true
 borsh.workspace = true
 digest.workspace = true
+lazy_static.workspace = true
```
Zk2u: Unless we're targeting an MSRV of <1.70.0, you can do this with `std::cell::OnceCell` or `std::sync::OnceLock`

crates/util/mmr/src/hasher.rs
```diff
@@ -1,22 +1,121 @@
 use digest::{generic_array::GenericArray, Digest};
+use lazy_static::lazy_static;
+use sha2::Sha256;
 
-pub type Hash = [u8; 32];
+/// 20 byte hash.
+pub type Hash20 = [u8; 20];
```
Zk2u: This should probably be a `pub type Hash<const N: usize> = [u8; N];`

crates/util/mmr/src/hasher.rs
Zk2u: Yes the report says that the C implementation of doing this appends a `b"node"` or a `b"leaf"` before finalising the sha256 hash to prevent collisions, I think copying that is better here. In that case we don't even need `std::sync::OnceLock` or lazy statics.

crates/util/mmr/src/hasher.rs
Zk2u: Annoyingly, digest::Digest only has a fn [output_size](https://docs.rs/digest/latest/digest/trait.Digest.html#tymethod.output_size)() -> [usize](https://doc.rust-lang.org/nightly/std/primitive.usize.html); which isn't a const fn or an associated type so i'm not sure if this is possible :thinking:

crates/util/mmr/src/hasher.rs
```diff
@@ -1,22 +1,121 @@
...
-    /// combines the left and Right nodes to form a single Node
-    fn hash_node(left: Hash, right: Hash) -> Hash;
+    /// Hash value.
+    type Hash: MerkleHash;
+
+    /// Hashes an arbitrary message as leaf data to compute a leaf hash.
+    fn hash_leaf(buf: &[u8]) -> Self::Hash;
+
+    /// Hashes a node's left and right children to compute the node's hash.
+    fn hash_node(left: Self::Hash, right: Self::Hash) -> Self::Hash;
```
Zk2u, line 74: This should be a `fn hash(buf: &[u8]) -> Self::Hash;`, or even better an incremental API. Then in a different struct, you should have these hash_leaf and hash_node functions which wrap the hash (or hasher) function of this underlying Hasher implementation with something that takes the inputs you've written above, and appends `b"node"` or `b"hash"`. A struct is used instead of a trait to force correct logic, as long as the hasher works.

crates/util/mmr/src/lib.rs
```diff
@@ -4,53 +4,48 @@ pub mod hasher;
...
+    /// phantom data for hasher
+    _pd: PhantomData<MH>,
 }
 
-impl<H: MerkleHasher + Clone> MerkleMr<H> {
-    #[allow(unused)]
-    pub(crate) fn new(cap_log2: usize) -> Self {
+impl<MH: MerkleHasher + Clone> MerkleMr64<MH> {
+    /// Constructs a new MMR with some capacity.
+    pub fn new(cap_log2: usize) -> Self {
```
Zk2u: It should be clearer what this cap_log2 is. Another comment to explain and give example would be great

edit: i know what it is because i've touched this code before, but someone unfamiliar might not

crates/util/mmr/src/lib.rs
```diff
@@ -62,78 +57,106 @@ impl<H: MerkleHasher + Clone> MerkleMr<H> {
...
+        let peaks_len = self.peaks.len() as u64;
+
+        // To prevent overflow in this case.
+        if peaks_len == u64::BITS as u64 {
+            return u64::MAX;
+        }
+
+        // There's probably some bithacky thing to make this simpler.
+        (1 << peaks_len) - 1
+    }
```
Zk2u: they both will have the same bytecode but this is nicer
```suggestion
    pub fn max_capacity(&self) -> u64 {
        match self.peaks.len() as u64 {
           0 => 0,
           peaks_len => u64::MAX >> (64 - peaks_len),
       }
    }
```

Zk2u: maybe when const expressions stabilise we can make this all generic instead

crates/util/mmr/src/lib.rs
```diff
@@ -62,78 +57,106 @@ impl<H: MerkleHasher + Clone> MerkleMr<H> {
...
         if !self.num.is_power_of_two() && self.num != 1 {
             return Err(MerkleError::NotPowerOfTwo);
         }
 
         Ok(self.peaks[(self.num.ilog2()) as usize])
     }
 
-    #[allow(unused)]
-    pub(crate) fn add_leaf_updating_proof(
+    /// Adds a new leaf, consuming a proof and returning an updated copy of the
```
Zk2u: not consuming, it takes by reference

crates/util/mmr/src/lib.rs
```diff
@@ -186,34 +208,29 @@ impl<H: MerkleHasher + Clone> MerkleMr<H> {
         }
     }
 
-    #[allow(unused)]
-    pub(crate) fn add_leaf_updating_proof_list(
+    pub fn add_leaf_updating_proof_list(
         &mut self,
```
Zk2u, line 231: docs on behaviour

crates/util/mmr/src/hasher.rs
delbonis: I'm worried this will allow LLVM to realize it can vectorize this loop (which we rather don't want).

crates/util/mmr/src/hasher.rs
delbonis: @bewakes To align with the SHA-256 block size.  The same reason that taproot's tagged hashing does this same thing.

crates/util/mmr/src/hasher.rs
storopoli: > I'm worried this will allow LLVM to realize it can vectorize this loop (which we rather don't want).

Tell the compiler what to do then: https://nnethercote.github.io/perf-book/inlining.html

delbonis: I'm just going to leave it as-is.

bewakes: > I'm worried this will allow LLVM to realize it can vectorize this loop (which we rather don't want).

That's a good point, LLVM can do that. However, the u32 accumulator also _doesn't_ guarantee that LLVM won't optimize the loop. If we _really_ want to do constant time equality check then let's do this properly even if that means having a dependency. Because otherwise, we will just have false sense of security while calling `eq_ct` which might backfire in the future. If we don't need `eq_ct`, I think it makes more sense to just have `eq` rather than something not guaranteed.

crates/util/mmr/src/hasher.rs
delbonis: Reworked these types even more.

crates/util/mmr/Cargo.toml
delbonis: Fixed, actually using `LazyCell`.

crates/util/mmr/src/hasher.rs
delbonis: So this is actually a convenience fn, it doesn't really have a reason to exist since the MMR treats leafs as being already hashed.  I'm just including it here so we can dismiss the audit comment.

crates/util/mmr/src/hasher.rs
delbonis: Fixed with new type.

delbonis:
Force merging since it's just waiting on fntests now.

PR #878, opened 2025-06-12
Remove download feature from corepc, update bitcoind version in CI

pranjalpokharel7:
## Description

Remove download feature from `corepc-node` package since it is expected that the `bitcoind` executable is installed by default on target testing machine (package is only use for testing currently). This will help us speed up local development by avoiding any unnecessary re-downloads.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [x] Dependency Update

## Checklist

### DIFF
--- a/.github/workflows/functional.yml
+++ b/.github/workflows/functional.yml
@@ -87,16 +87,16 @@ jobs:
 
       - name: Install bitcoind
         env:
-          BITCOIND_VERSION: "28.1"
+          BITCOIND_VERSION: "29.0"
           BITCOIND_ARCH: "x86_64-linux-gnu"
-          SHASUM: "07f77afd326639145b9ba9562912b2ad2ccec47b8a305bd075b4f4cb127b7ed7"
         run: |
           curl -fsSLO --proto "=https" --tlsv1.2 "https://bitcoincore.org/bin/bitcoin-core-$BITCOIND_VERSION/bitcoin-$BITCOIND_VERSION-$BITCOIND_ARCH.tar.gz"
-          sha256sum -c <<< "$SHASUM bitcoin-$BITCOIND_VERSION-$BITCOIND_ARCH.tar.gz"
+          curl -fsSLO --proto "=https" --tlsv1.2 "https://bitcoincore.org/bin/bitcoin-core-$BITCOIND_VERSION/SHA256SUMS"
+          sha256sum --ignore-missing --check SHA256SUMS
           tar xzf "bitcoin-$BITCOIND_VERSION-$BITCOIND_ARCH.tar.gz"
           sudo install -m 0755 -t /usr/local/bin bitcoin-"$BITCOIND_VERSION"/bin/*
           bitcoind --version
-          rm -rf "bitcoin-$BITCOIND_VERSION" "bitcoin-$BITCOIND_VERSION-$BITCOIND_ARCH.tar.gz"
+          rm -rf SHA256SUMS "bitcoin-$BITCOIND_VERSION" "bitcoin-$BITCOIND_VERSION-$BITCOIND_ARCH.tar.gz"
 
       - name: Set up Python
         uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
--- a/crates/btcio/Cargo.toml
+++ b/crates/btcio/Cargo.toml
@@ -38,7 +38,8 @@ strata-state = { workspace = true, features = ["test_utils"] }
 strata-status.workspace = true
 strata-test-utils.workspace = true
 
-corepc-node = { version = "0.8.0", features = ["29_0", "download"] }
+# update bitcoind version in CI to match corepc-node feature
+corepc-node = { version = "0.8.0", features = ["29_0"] }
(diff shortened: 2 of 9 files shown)


.github/workflows/functional.yml
```diff
@@ -87,7 +87,7 @@ jobs:
 
       - name: Install bitcoind
         env:
-          BITCOIND_VERSION: "28.1"
+          BITCOIND_VERSION: "29.0"
           BITCOIND_ARCH: "x86_64-linux-gnu"
           SHASUM: "07f77afd326639145b9ba9562912b2ad2ccec47b8a305bd075b4f4cb127b7ed7"
```
storopoli: You need to change this as well

storopoli: You also need to add the `Install bitcoind` step (with the v29 version and proper hash) to `unit.yml` for the unit tests. They need to have `bitcoind` in their `PATH`.

storopoli (changes requested):

.github/workflows/functional.yml
pranjalpokharel7: Do you mean the SHASUM? The standard way `bitcoincore.org` recommends to verify the checksums is to download a `SHA256SUMS` file and run the check as such (see linux verification instructions under https://bitcoincore.org/en/download/) - 
```bash
# download bitcoincore
curl -fsSLO --proto "=https" --tlsv1.2 "https://bitcoincore.org/bin/bitcoin-core-$BITCOIND_VERSION/bitcoin-$BITCOIND_VERSION-$BITCOIND_ARCH.tar.gz"

# download sha256sums file for core version
curl -fsSLO --proto "=https" --tlsv1.2 "https://bitcoincore.org/bin/bitcoin-core-$BITCOIND_VERSION/SHA256SUMS"

# check sums
sha256sum --ignore-missing --check SHA256SUMS
```
Is there a specific reason why we're specifying SHASUM variable manually? This way it is more forward compatible.

storopoli: No, you are downloading v29 trying to SHA-256 verify against a hash that was set to v28. You will succeed only in 1 out of $2^{256}$ chances with 100% probability.
Update the variable `SHASUM` to the proper v29 value 
CleanShot 2025-06-13 at 07 28 51@2x

storopoli: Ok, I think you might have solved in https://github.com/alpenlabs/alpen/pull/878/commits/0c6e35dcfd807234af54644516bc110ec8b54b22

storopoli (approved):
ACK 0c6e35dcfd807234af54644516bc110ec8b54b22

storopoli:
Needs to be backported to `releases/0.2.0`

Rajil1213 (dismissed):
Can we also add a note in the `Cargo.toml` file to update the version in CI when we update the bitcoind version there?

storopoli:
> Looks like I was too late. We can add that note later.

No we're not, removed from the queue.

storopoli (changes requested):
We also need to update the documentation here for `bitcoind` https://github.com/alpenlabs/alpen/blob/17975f403b391ae5daf7022fa361c65476a54276/CONTRIBUTING.md?plain=1#L77-L87

You can copy pasta what @Rajil1213 has done in `strata-bridge`: https://github.com/alpenlabs/strata-bridge/pull/157/commits/845c194e4839546d180782f67deb77990486d974

with the caveat that we need `bitcoind` to run unit and functional tests.

crates/btcio/Cargo.toml
```diff
@@ -38,7 +38,8 @@ strata-state = { workspace = true, features = ["test_utils"] }
 strata-status.workspace = true
 strata-test-utils.workspace = true
 
-corepc-node = { version = "0.8.0", features = ["29_0", "download"] }
+# update bitcoind version in CI when changing the corepc-node feature (protocol number)
```
storopoli: WTF is "protocol number"? You mean the bitcoin version?
```suggestion
# update bitcoind version in CI to match corepc-node feature
```

crates/util/python-utils/Cargo.toml
```diff
@@ -38,6 +38,9 @@ strata-common.workspace = true
 
 anyhow.workspace = true
 bitcoind-async-client.workspace = true
-corepc-node = { version = "0.8.0", features = ["29_0", "download"] }
+
+# update bitcoind version in CI when changing the corepc-node feature (protocol number)
```
storopoli: WTF is "protocol number"? You mean the bitcoin version?
```suggestion
# update bitcoind version in CI to match corepc-node feature
```

storopoli (changes requested):

crates/util/python-utils/Cargo.toml
pranjalpokharel7: Oops 😓

pranjalpokharel7: Corrected ✅

storopoli (approved):
ACK 5446eae2823f933169fce96c1d63103cc8440c81

PR #879, opened 2025-06-12
STR-1407 Fix cache locking

delbonis:
## Description

There's a possible race-like condition that can occur when there's an error making a database fetch.  This PR reworks the locking pattern to address this, as well as improve efficiency of cache access, and other minor improvements.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Please verify I didn't bungle anything.

## Checklist

### DIFF
--- a/crates/storage/src/cache.rs
+++ b/crates/storage/src/cache.rs
@@ -326,13 +326,12 @@ mod concurrent_tests {
         let results: Vec<_> = futures::future::join_all(tasks).await;
 
         // All tasks should fail with same error
-        // TODO: fix this, this is not the case
         for result in results {
             assert!(matches!(result.unwrap(), Err(DbError::Busy)));
         }
 
         // Only one fetch should occur even on error
-        assert_eq!(fetch_count.load(Ordering::SeqCst), 1);
+        assert_eq!(fetch_count.load(Ordering::SeqCst), 5);
 
         // Cache should remain empty
         assert_eq!(cache.get_len(), 0);
@@ -429,12 +428,7 @@ mod concurrent_tests {
             let fetch_count = fetch_count.clone();
             tokio::spawn(async move {
                 cache
-                    .get_or_fetch(&42, || {
-                        let (tx, rx) = tokio::sync::oneshot::channel();
-                        fetch_count.fetch_add(1, Ordering::SeqCst);
-                        tx.send(Ok(100)).expect("send value");
-                        rx
-                    })
+                    .get_or_fetch(&42, helper_fetch_fn(fetch_count, Ok(100)))
                     .await
             })
         };
@@ -463,6 +457,8 @@ mod concurrent_tests {
         // One should get the cached value from the other
         assert!(async_val == 100 || async_val == 200);
         assert!(blocking_val == 100 || blocking_val == 200);
+        // Both values should be the same
+        assert_eq!(async_val, blocking_val);
 
         // Only one fetch should occur
         assert_eq!(fetch_count.load(Ordering::SeqCst), 1);
@@ -471,23 +467,18 @@ mod concurrent_tests {
     #[tokio::test(flavor = "multi_thread", worker_threads = 8)]
     async fn test_high_concurrency_stress() {
         let cache = Arc::new(CacheTable::<u64, u64>::new(50.try_into().unwrap()));
-        let total_fetches = Arc::new(AtomicUsize::new(0));
+        let fetch_count = Arc::new(AtomicUsize::new(0));
 
         // 100 concurrent operations on 10 different keys
         let tasks: Vec<_> = (0..100)
             .map(|i| {
                 let cache = cache.clone();
-                let total_fetches = total_fetches.clone();
+                let fetch_count = fetch_count.clone();
                 let key = (i % 10) as u64;
 
                 tokio::spawn(async move {
                     cache
-                        .get_or_fetch(&key, || {
-                            let (tx, rx) = tokio::sync::oneshot::channel();
-                            total_fetches.fetch_add(1, Ordering::SeqCst);
-                            tx.send(Ok(key * 100)).expect("send value");
-                            rx
-                        })
+                        .get_or_fetch(&key, helper_fetch_fn(fetch_count, Ok(key * 100)))
                         .await
                 })
             })
@@ -501,7 +492,7 @@ mod concurrent_tests {
         }
 
         // Should have exactly 10 fetches (one per unique key)
-        assert_eq!(total_fetches.load(Ordering::SeqCst), 10);
+        assert_eq!(fetch_count.load(Ordering::SeqCst), 10);
 
         // Cache should contain all 10 keys
         assert_eq!(cache.get_len(), 10);


delbonis:
Someone please verify that this properly addresses the concern.

crates/storage/src/cache.rs
```diff
@@ -166,35 +166,51 @@ impl<K: Clone + Eq + Hash, V: Clone> CacheTable<K, V> {
...
+
             Ok(Err(e)) => {
-                error!(?e, "failed to make database fetch");
+                // Important ordering for the locks.
+                let mut cache_lock = self.cache.lock();
+                trace!("re-acquired cache lock");
                 *slot_lock = SlotState::Error;
-                self.purge(k);
-                return Err(e);
+                cache_lock.pop(k);
```
storopoli: Something like
```suggestion
                if Arc::ptr_eq(&cache_lock.peek(&k).unwrap(), &slot) {
                    cache_lock.pop(k);
                }
```

crates/storage/src/cache.rs
```diff
@@ -166,35 +166,51 @@ impl<K: Clone + Eq + Hash, V: Clone> CacheTable<K, V> {
...
-                error!("database fetch aborted");
-                self.purge(k);
-                return Err(DbError::WorkerFailedStrangely);
-            }
-        };
+                // Important ordering for the locks.
+                let mut cache_lock = self.cache.lock();
+                trace!("re-acquired cache lock");
+                *slot_lock = SlotState::Error;
+                cache_lock.pop(k);
```
storopoli: something like:
```suggestion
                if Arc::ptr_eq(&cache_lock.peek(&k).unwrap(), &slot) {
                    cache_lock.pop(k);
                }
```

crates/storage/src/cache.rs
```diff
@@ -220,33 +236,63 @@ impl<K: Clone + Eq + Hash, V: Clone> CacheTable<K, V> {
...
+
             Err(e) => {
-                warn!(?e, "failed to make database fetch");
+                // Important ordering for the locks.
+                let mut cache_lock = self.cache.lock();
+                trace!("re-acquired cache lock");
                 *slot_lock = SlotState::Error;
-                self.purge(k);
-                return Err(e);
+                cache_lock.pop(k);
```
storopoli:
```suggestion
                if Arc::ptr_eq(&cache_lock.peek(&k).unwrap(), &slot) {
                    cache_lock.pop(k);
                }
```

delbonis:
I've did change to use the `ptr_eq` comparison as a separate function.

bewakes:
Here, I asked claude to generate test for our desired behaviour. In particular look at the `test_purge_during_fetch` test. That seems alright to me, but what do you guys think?

https://claude.ai/public/artifacts/93515022-dbbd-462a-b05b-aef1e81e9edb

delbonis:
@bewakes That does not compile:

```
error[E0277]: `*mut ()` cannot be sent between threads safely
   --> crates/storage/src/cache.rs:475:26
    |
475 |               let handle = tokio::spawn(async move {
    |                            ^            ---------- within this `{async block@crates/storage/src/cache.rs:475:39: 475:49}`
    |  __________________________|
    | |
476 | |                 let key = i as u64;
477 | |                 let result = cache_clone
478 | |                     .get_or_fetch(&key, || {
...   |
490 | |                 (key, result)
491 | |             });
    | |______________^ `*mut ()` cannot be sent between threads safely
    |
    = help: within `{async block@crates/storage/src/cache.rs:475:39: 475:49}`, the trait `Send` is not implemented for `*mut ()`
note: required because it appears within the type `GuardNoSend`
   --> /home/trey/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lock_api-0.4.12/src/lib.rs:103:12
    |
103 | pub struct GuardNoSend(*mut ());
    |            ^^^^^^^^^^^
    = note: required because it appears within the type `(&mut LruCache<u64, Arc<tokio::sync::RwLock<SlotState<String>>>>, GuardNoSend)`


```

I think it's trying to send a mutex guard across threads.

delbonis:
Okay I see, since we're holding the guard across yield points it makes the `get_or_fetch` call `!Send`.  So I changed the calls to `tokio::spawn` to `tokio::task::spawn_local`, which makes it compile.  However, we need to do something else here since all of the tests that completed failed with this error:

```
thread 'cache::ai_tests::test_cache_eviction_during_fetch' panicked at /home/trey/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.45.0/src/task/local.rs:418:29:
`spawn_local` called from outside of a `task::LocalSet` or LocalRuntime
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

and then it seems like the remaining tests have not completed.  These used the blocking/non-async interface, and these seem to hang.

I am not sure how to troubleshoot this anymore.  I am not sure if this is a deadlock or a busy spin or what.

delbonis:
I have committed what I have now.

bewakes:
@delbonis could you have a look at the latest state of cache module? It is much cleaner and easier to reason about now, but I am feeling a bit weird with the blocking calls.

crates/storage/src/cache.rs
```diff
@@ -1,252 +1,114 @@
...
+        tokio::task::block_in_place(move || {
+            tokio::runtime::Handle::current().block_on(async {
+                self.get_or_fetch(key, || {
+                    let (tx, rx) = tokio::sync::oneshot::channel();
+                    let result = fetch_fn();
+                    let _ = tx.send(result);
+                    rx
+                })
+                .await
+            })
```
bewakes: This is what I feel a bit weird about. Works though.

crates/storage/src/cache.rs
```diff
@@ -298,8 +160,8 @@ mod tests {
         assert_eq!(len, 0);
     }
 
-    #[test]
-    fn test_basic_blocking() {
+    #[tokio::test(flavor = "multi_thread")]
```
bewakes, line 199: Had to do this because of `block_on`. A part of the "weirdness"

crates/storage/src/cache.rs
```diff
@@ -1,252 +1,142 @@
-//! Generic cache utility for what we're inserting into the database.
```
storopoli, line 1: Can you please re-add the module-level docs. You can change it for whatever you want them to be. I just want to have module-level documentation everywhere possible.

crates/storage/src/cache.rs
```diff
@@ -1,252 +1,142 @@
...
-        *slot_lock = SlotState::Ready(res.clone());
-        if complete_tx.send(res.clone()).is_err() {
-            // This happens if there was no waiters, which is normal, leaving it
-            // here if we need to debug it.
-            //warn!("failed to notify waiting cache readers");
-        }
-
-        Ok(res)
+        tokio::task::block_in_place(move || {
+            tokio::runtime::Handle::current().block_on(async {
```
storopoli: Can we have imports for these at the top?
```rs
use foo;
use bar;
```

storopoli (changes requested):

crates/storage/src/cache.rs
```diff
@@ -1,252 +1,150 @@
...
 pub(crate) struct CacheTable<K, V> {
-    cache: Mutex<lru::LruCache<K, CacheSlot<V>>>,
+    /// The actual cache.
+    cache: RwLock<LruCache<K, V>>,
+
+    /// One mutex per key to prevent duplicate fetches.
+    /// Created when inserting, cleaned up when the entry is purged.
+    /// Not sure if trying to clean this up after fetch complete is worthwhile. Seems to add a
+    /// bunch of complex logic whereas keeping it won't have much memory footprint.
+    fetch_mutexes: RwLock<HashMap<K, Arc<Mutex<()>>>>,
```
delbonis, line 24: This is a bit of a red flag.  The broadcast channel was to allow subsequent queries for the same key to reuse the freshly fetched value from the cache.  This approach changes the flow quite a bit.

crates/storage/src/cache.rs
```diff
@@ -1,252 +1,150 @@
...
-            Ok(Err(e)) => {
-                error!(?e, "failed to make database fetch");
-                *slot_lock = SlotState::Error;
-                self.purge(k);
-                return Err(e);
+        // After acquiring the lock, check for the value
+        if let Some(value) = self.get(key) {
+            // Value present, no need to call fetch_fn
+            return Ok(value);
+        }
```
delbonis, line 114: This requires taking a lock on the LRU cache again, doubling the lock contention, which is not ideal.

crates/storage/src/cache.rs
```diff
@@ -1,252 +1,150 @@
...
+        block_in_place(move || {
+            runtime::Handle::current().block_on(async {
+                self.get_or_fetch(key, || {
+                    let (tx, rx) = oneshot::channel();
+                    let _ = tx.send(fetch_fn());
+                    rx
+                })
+                .await
+            })
+        })
```
delbonis, line 147: This assumes that there is a runtime context on the current thread.  This means that it does not work on main or on threads started with `thread::spawn`, which is all of them (unless you do a `Handle::enter` call).

The locking primitives have blocking functions.  This is so that you can use them without being in a runtime context.  That's what the original version did.

crates/storage/src/cache.rs
delbonis: See comment about runtime context.

delbonis (commented):
@bewakes This completely changes the implementation of the cache table, which invalidates the audit findings, so we'd probably want to do another audit on this new implementation.  I had tried to just apply the changes to the implementation that the audit requested, to the best of my understanding.  The remaining work in this PR was just to write correct tests, avoiding fundamentally rewriting the implementation.

crates/storage/src/cache.rs
bewakes: It changes the flow, but the behavior is still the same.

crates/storage/src/cache.rs
bewakes: Could we try to create a temporary runtime if there's no current runtime?

crates/storage/src/cache.rs
delbonis: The behavior is not the same.  The locking pattern is different (more contention) and it introduces this new dummy lock structure.

crates/storage/src/cache.rs
delbonis: No, that's not the correct behavior since that involves a setting up [a *ton* of bookkeeping (kernel calls, allocations)](https://docs.rs/tokio/latest/src/tokio/runtime/scheduler/current_thread/mod.rs.html) to set up the runtime and make the call, only to tear it down immediately after that call returns.

The correct behavior is to use the functions designed for this use-case.  This also goes a long way to ensure that we avoid *any* additional context switches in un-contended contexts.

PR #886, opened 2025-06-17
add startup check for genesis config mismatch

sapinb:
## Description
Add startup check for early detection of genesis config mismatch between params.json and reth.

PR to release: #887 

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/bin/strata-client/src/main.rs
+++ b/bin/strata-client/src/main.rs
@@ -255,7 +255,7 @@ fn do_startup_checks(
     // Ensure reth and strata are running on same params
     let genesis_block = make_genesis_block(params);
     let genesis_check_res = retry_with_backoff(
-        "engine_submit_payload",
+        "engine_check_block_exists",
         DEFAULT_ENGINE_CALL_MAX_RETRIES,
         &ExponentialBackoff::default(),
         || engine.check_block_exists(L2BlockRef::Ref(&genesis_block)),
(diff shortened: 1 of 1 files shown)


crates/eectl/src/engine.rs
```diff
@@ -46,7 +46,13 @@ pub trait ExecEngineCtl {
     /// Check if a block exists on the chain.
     /// If this returns true, it should be safe to use this id
     /// in any of update_*_block methods, submit_payload and prepare_payload
-    fn check_block_exists(&self, id: L2BlockId) -> EngineResult<bool>;
+    fn check_block_exists<'a>(&self, id_ref: L2BlockRef<'a>) -> EngineResult<bool>;
+}
+
+#[derive(Debug)]
+pub enum L2BlockRef<'a> {
```
bewakes, line 53: Was this enum required regardless of this latest check with EL or is this introduced explicitly for this task? I mean we should be able to compute block id of genesis block, right? or is that not feasible?

delbonis (dismissed):
I probably would have split this into two separate functions but ig it's fine.

crates/eectl/src/engine.rs
sapinb: During the first startup, db does not have genesis block, so using id will not work. Adding the ref for this single case.

bewakes (approved):

PR #898, opened 2025-06-22
STR-1438 Database manipulation tool

krsnapaudel:
## Description

First draft to explore what we can do with database traits.

It would be good to determine what functionality, new APIs and commands we need before going further.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


bin/alpen-dbtool/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+edition = "2021"
+name = "alpen-dbtool"
+version = "0.3.0-alpha.1"
+
+[[bin]]
+name = "alpen-dbtool"
+path = "src/main.rs"
+
```
storopoli: Add the workspace lints:

```suggestion

[lints]
workspace = true
```

barakshani:
As we talked in person, I recommend focusing on the first 2 scenarios:
1. we have some sealed l2 blocks in our database, but we fail to prove the corresponding batch. In this case the minimum that we need is to easily fetch the entire batch of transactions and (after the problem is found, say a problematic tx) revert until the latest correct block. Other data that we most likely need in this case is access to the previous state -- a state snapshot seems to be supported in both [rocksDB](https://github.com/facebook/rocksdb/wiki/Snapshot) and [sled](https://github.com/bonsairobo/sled-snapshots).
2. a batch is sealed but due to a bug it does not comply with our sealing conditions, for example right now we seal batches at 1b gas. If these conditions become consensus rule, then other nodes won't accept the batch valid and we'd have to revert it, similar to above.
3. In a similar manner, though not very likely, the above can happen with an EVM block in case we have some incompatibility with RETH and REVM. I would also check with the EE/Prover team if there is any chance that some unsupported tx gets into the mempool (this means that we would try to include it in a block, but couldn't as it's not supported). Unsupported in this case can account to different versions of EVM. We're using some wrappers and we manually updating versions, so perhaps the mempool is a place where we can cause incompatibility. In this case we'd need some access to the mempool.

bin/alpen-dbtool/src/cli.rs
```diff
@@ -0,0 +1,80 @@
...
+
+/// Alpen DB tool – offline database & chain‑maintenance utility.
+#[derive(Parser, Debug)]
+#[command(
+    name = "alpen-dbtool",
+    version = crate_version!(),
+    about = "Inspect, repair and roll back an Alpen node’s database while the node is offline.",
+    propagate_version = true,
+    disable_help_subcommand = true
+)]
```
delbonis: Please use `argh` as most of the rest of the project does, unless there's some very strong reason to.  Clap takes a long time to compile and has much of the same features we actually use.

Also, unless someone else told you otherwise and I'm just wrong, we should probably still call it `strata-dbtool`.  Otherwise it won't make any sense why we're calling it `alpen-dbtool` to interact with `strata-client`'s database.

bin/alpen-dbtool/src/cmd/chainstate.rs
```diff
@@ -0,0 +1,142 @@
...
+    println!("Finalized epoch: {:?}", top_level_state.finalized_epoch());
+    println!("L1 view: {:?}", top_level_state.l1_view());
+    println!("Deposits table: {:?}", top_level_state.deposits_table());
+
+    Ok(())
+}
+
+/// Reset the chainstate to a specific L2 block.
+pub(crate) fn reset_chainstate(
+    db: Arc<CommonDb>,
```
delbonis: Usually a function like this (and the other fns here) would be expected to take a `&CommonDb` to avoid the clone in `get_chainstate`.

However, all of these functions and across the other subcommands should actually take `impl ChainstateDb` (actually, `&impl ...`) and similar in order to be generic over the database instead of demanding it be the RocksDB backend impl.  So then we'd only explicitly state the concrete impl when we're setting it up in main.

bin/alpen-dbtool/src/cmd/l1.rs
```diff
@@ -0,0 +1,171 @@
...
+        l1_block_manifest.epoch()
+    );
+
+    // Print relevant transactions
+    for tx in l1_block_manifest.txs().iter() {
+        for proto_op in tx.protocol_ops().iter() {
+            match proto_op {
+                ProtocolOperation::Checkpoint(signed_checkpoint) => {
+                    println!(
+                        "checkpoint commitment: {:?}",
```
delbonis: Probably makes sense to include the txid and tx index where each of these protocol operations are.  Either putting it on the same line or indenting the proto op data beneath it (in case there's multiple proto ops in the same tx).

bin/alpen-dbtool/src/cmd/l1.rs
```diff
@@ -0,0 +1,171 @@
...
+                        signed_checkpoint.checkpoint().commitment()
+                    );
+                }
+                ProtocolOperation::DaCommitment(da_commitment) => {
+                    println!("DA commitment: {da_commitment:?}");
+                }
+                ProtocolOperation::WithdrawalFulfillment(wf_info) => {
+                    println!("checkpoint commitment: {wf_info:?}");
```
delbonis: checkpoint commitment -> withdrawal fulfillment?

bin/alpen-dbtool/src/errors.rs
delbonis, line 1: I've seen code like this in another one of the cli tools.  I wonder if we could put this in some `strata-cli-common` crate?

delbonis (changes requested):
This is a really good start!

Cargo.toml
```diff
@@ -246,7 +246,7 @@ borsh = { version = "1.5.0", features = ["derive"] }
 bytes = "1.6.0"
 cfg-if = "1.0.0"
 chrono = "0.4.38"
-clap = { version = "4.5", features = ["derive", "wrap_help", "cargo", "color"] }
+clap = "4"
```
storopoli: Can you please add the full version of the dependency, i.e. `X.Y.Z`?

storopoli (changes requested):

bin/strata-dbtool/src/cmd/chainstate.rs
```diff
@@ -0,0 +1,141 @@
...
+pub(crate) struct GetChainstateArgs {
+    /// chainstate write index; defaults to the latest
+    #[argh(positional)]
+    pub(crate) write_idx: Option<u64>,
+}
+
+/// Resets the chainstate to a specific L2 block
+#[derive(FromArgs, Debug)]
+#[argh(subcommand, name = "reset-chainstate")]
+pub(crate) struct ResetChainstateArgs {
```
barakshani: would "revert" be a better term for this?

barakshani:
> We cannot cover reth database here. Same for L1 broadcaster db.

Why not? I'm also in the opinion that we cannot revert the chainstate if we don't revert the reth db, at least right now as they are tightly coupled.

krsnapaudel:
> > We cannot cover reth database here. Same for L1 broadcaster db.
> 
> Why not? I'm also in the opinion that we cannot revert the chainstate if we don't revert the reth db, at least right now as they are tightly coupled.

My understanding is that the database api mentioned by @delbonis in the ticket does not include reth db. Maybe @delbonis can clarify.

bin/alpen-dbtool/src/cmd/l1.rs
krsnapaudel: Thanks for catching the typo.

delbonis:
Yeah @krsnapaudel the Reth database is a separate database.

@barakshani The databases are not tightly coupled, it should be *possible* to do it independently.  However I am not sure the ways this could break if there's major inconsistencies.  I am not sure what the Reth tooling is like, this is really more of a @MdTeach question *I think*.

I can only reliably speak about the strata client database.

barakshani:
> Yeah @krsnapaudel the Reth database is a separate database.
> 
> @barakshani The databases are not tightly coupled, it should be _possible_ to do it independently. However I am not sure the ways this could break if there's major inconsistencies. I am not sure what the Reth tooling is like, this is really more of a @MdTeach question _I think_.
> 
> I can only reliably speak about the strata client database.

@delbonis I meant that db's should be consistent, as blocks are coupled. So if we revert the OL chain, we should also revert reth blocks, otherwise we'll get the same EL blocks.

Turns out this already exists for reth: https://reth.rs/cli/reth/stage/unwind/.

storopoli:
@krsnapaudel this needs rebasing

storopoli:
> > @krsnapaudel this needs rebasing
> 
> Yes. I have rewritten history and created commits based on your tips.

Amazing! This is MUCH MUCH better! Congrats!

We still need to solve the conflicts. So you might need to "fast-forward" your local `main` branch and then rebase this PR branch on top of that and solve the conflicts.

delbonis:
Seems like the unit tests are failing because `strata-dbtool` isn't being built?

krsnapaudel:
@delbonis, @sapinb : While functional tests for revert do pass, we are not touching checkpoint db, L1 write db or L2 broadcaster db. Should we try to add apis to update them here or target in a follow up PR?

delbonis (approved):

delbonis:
Remaining comments all resolved, merging.

PR #899, opened 2025-06-23
STR-1506: fix EF tests docker setup 

evgenyzdanovich:
## Description

Fix EF tests against docker setup:
- enhance init_keys.sh to allow passing params to datatool
- create and use a separate prover config in the setup
- use `prover-client` from `releases/0.2.0` (rather than from main)
- some nits

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/.github/workflows/staging-eest.yml
+++ b/.github/workflows/staging-eest.yml
@@ -6,7 +6,6 @@ on:
   workflow_dispatch:      # Manual trigger (optional)
 
 env:
-  #CI: false
   COMMIT: ${{ github.sha }}
 
 permissions: {}
@@ -86,6 +85,7 @@ jobs:
 
     - name: Build params
       run: |
+        # TODO: consider switching to uv instead of pip3.
         # Install base58
         pip3 install base58
 


delbonis (approved):
Sorry for the delay.  No issues, approving.

evgenyzdanovich:
@delbonis apologies, I'd want to add one small thingy. Please check one last commit and re-approve.

evgenyzdanovich:
@storopoli @purusang @voidash Could you guys please take a look? Thanks!

delbonis (dismissed):

.github/workflows/staging-eest.yml
```diff
@@ -1,29 +1,32 @@
...
 
 on:
   schedule:
-    - cron: '0 3 * * 0'  # Weekly: Sunday 03:00 UTC
+    - cron: '0 15 * * *'  # Daily at 15:00 UTC
   workflow_dispatch:      # Manual trigger (optional)
 
 env:
-  CI: false
+  #CI: false
```
storopoli: 1. What this does?
2. Do we need to leave it commented out or can we remove it? Remember the YAGNI acronym.

.github/workflows/staging-eest.yml
```diff
@@ -65,6 +68,22 @@ jobs:
...
+        # Build prover
+        docker build \
+          --build-arg PROVER_FEATURES="default" \
+          -t "strata-prover-client:latest" \
+          -f "docker/prover-client/Dockerfile" .
+
+    - name: Switch to main and do stuff
+      run: |
+        # Required because compose yml and test_ee_proof live on main so far.
+        git checkout main
```
storopoli, line 84: When we backport this do we need to update this on `main`? and remove the `git checkout main`?

.github/workflows/staging-eest.yml
```diff
@@ -65,6 +68,22 @@ jobs:
...
+
     - name: Build params
       run: |
         # Install base58
```
storopoli, line 89: Can be added as a TODO, but would be really nice to move away from `pip` (and `poetry` in the future) to use `uv` like we do with `zizmor`: https://github.com/alpenlabs/alpen/blob/1d1ef5567ded5ff923410e84e1bad3693f0e3196/.github/workflows/zizmor.yml?plain=1#L19-L23

.github/workflows/staging-eest.yml
evgenyzdanovich: Yes, removed. Unsure exactly what it does, but `CI: false` only used in `cd.yml` (which this file was forked from initially). I assume its not needed here.

.github/workflows/staging-eest.yml
evgenyzdanovich: we don't need to backport it - `checkout`ing main basically allows it to not be backported. All the binaries are taken from `releases/020`, the only parts from main are: usage of this compose and test script.

delbonis (approved):

.github/workflows/staging-eest.yml
evgenyzdanovich: left a TODO, thanks

.github/workflows/staging-eest.yml
```diff
@@ -65,32 +67,44 @@ jobs:
...
-        cd docker && ./init-keys.sh ../target/release/strata-datatool
+        # Generate params: rollup_name := alpn, block_time := 1000
+        cd docker && ./init-keys.sh ../target/release/strata-datatool -n alpn -t 1
         cd configs
 
+        # TODO: remove once it's merged to main (init-keys are modified in the current PR).
         # Overwrite the block time for faster blocks.
         jq '.block_time = 1000' params.json > tmp.json && mv tmp.json params.json
+        # Overwrite the rollup_name so the sequencer is happy.
+        jq '.rollup_name = "alpn"' params.json > tmp.json && mv tmp.json params.json
```
purusang, line 103: Do no think this will be needed after changes in `init-keys.sh` in this pr.

purusang (changes requested):

.github/workflows/staging-eest.yml
evgenyzdanovich: Yeah, I agree. the whole block with `jq` commands won't be needed, but the changes to `init-keys.sh` are made in this current PR, while the workflow takes the version from main (so it's kinda _cyclic dep on changes_ in a way) :) Once changes to init-keys are merged, we can remove `jq`

This fact is also described in the comment at line 99:
`# TODO: remove once it's merged to main (init-keys are modified in the current PR).`

purusang (approved):

PR #906, opened 2025-06-30
STR-1407 Fix cache locking (v2)

delbonis:
## Description

There's a possible race-like condition that can occur when there's an error making a database fetch. This PR reworks the locking pattern to address this, as well as improve efficiency of cache access, and other minor improvements.

This is a partial reset from the [old PR](https://github.com/alpenlabs/alpen/pull/879), but also converting from parking_lot mutexes to tokio ones and naming the branch properly.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


delbonis:
Pending needing to write tests that match the below direction, without substantially changing the implementation unless absolutely necessary.

> A race condition exists between marking a cache slot as errored and purging it. Because the global cache mutex is dropped before error handling, another thread can insert a new slot under the same key before purge(k) runs, causing the new slot to be inadvertently removed. solution:
>     * In `get_or_fetch` (async and blocking variants), hold the cache mutex while marking a slot `Error` and before removing it, or verify the slot identity with `Arc::ptr_eq` before purging.  **(done)**
>     * Add a concurrent test that spawns two tasks fetching the same key, one of which fails, to ensure the new slot is preserved.

crates/storage/src/cache.rs
```diff
@@ -87,14 +86,30 @@ impl<K: Clone + Eq + Hash, V: Clone> CacheTable<K, V> {
     /// Gets the number of elements in the cache.
     // TODO replace this with an atomic we update after every op
     #[allow(dead_code)]
```
storopoli, line 48:
```suggestion
    #[expect(dead_code)]
```

storopoli: Or give it a reason https://doc.rust-lang.org/reference/attributes/diagnostics.html#lint-reasons

crates/storage/src/cache.rs
```diff
@@ -87,14 +86,30 @@ impl<K: Clone + Eq + Hash, V: Clone> CacheTable<K, V> {
...
-    pub(crate) fn get_len(&self) -> usize {
-        let cache = self.cache.lock();
+    pub(crate) async fn get_len_async(&self) -> usize {
+        let cache = self.cache.lock().await;
+        cache.len()
+    }
+
+    /// Gets the number of elements in the cache.
+    // TODO replace this with an atomic we update after every op
+    #[allow(dead_code)]
```
storopoli, line 56:
```suggestion
    #[expect(dead_code)]
```

Or give it a reason https://doc.rust-lang.org/reference/attributes/diagnostics.html#lint-reasons

crates/storage/src/cache.rs
storopoli: the comment can be moved to the reason

crates/storage/src/cache.rs
storopoli: the comment can be moved to the reason

crates/storage/src/cache.rs
```diff
@@ -87,14 +86,30 @@ impl<K: Clone + Eq + Hash, V: Clone> CacheTable<K, V> {
...
+    pub(crate) fn get_len_blocking(&self) -> usize {
+        let cache = self.cache.blocking_lock();
         cache.len()
     }
 
     /// Removes the entry for a particular cache entry.
-    pub(crate) fn purge(&self, k: &K) {
-        let mut cache = self.cache.lock();
+    #[allow(dead_code)]
```
storopoli: same

crates/storage/src/cache.rs
```diff
@@ -87,14 +86,30 @@ impl<K: Clone + Eq + Hash, V: Clone> CacheTable<K, V> {
...
+    pub(crate) async fn purge_async(&self, k: &K) {
+        let mut cache = self.cache.lock().await;
+        cache.pop(k);
+    }
+
+    /// Removes the entry for a particular cache entry.
+    #[allow(dead_code)]
```
storopoli: same

crates/storage/src/cache.rs
```diff
@@ -127,74 +142,97 @@ impl<K: Clone + Eq + Hash, V: Clone> CacheTable<K, V> {
     /// This might remove slots that are in the process of being filled.  Those
     /// operations will complete, but we won't retain those values.
     #[allow(dead_code)]
```
storopoli: same

crates/storage/src/cache.rs
```diff
@@ -127,74 +142,97 @@ impl<K: Clone + Eq + Hash, V: Clone> CacheTable<K, V> {
...
-        let mut cache = self.cache.lock();
+    pub(crate) fn blocking_clear(&self) -> usize {
+        let mut cache = self.cache.blocking_lock();
         let len = cache.len();
         cache.clear();
         len
     }
 
     /// Inserts an entry into the table, dropping the previous value.
     #[allow(dead_code)]
```
storopoli: same

crates/storage/src/cache.rs
```diff
@@ -127,74 +142,97 @@ impl<K: Clone + Eq + Hash, V: Clone> CacheTable<K, V> {
...
-    pub(crate) fn insert(&self, k: K, v: V) {
+    pub(crate) async fn insert_async(&self, k: K, v: V) {
+        let slot = Arc::new(RwLock::new(SlotState::Ready(v)));
+        self.cache.lock().await.put(k, slot);
+    }
+
+    /// Inserts an entry into the table, dropping the previous value.
+    #[allow(dead_code)]
```
storopoli: same

crates/storage/src/cache.rs
```diff
@@ -208,46 +246,91 @@ impl<K: Clone + Eq + Hash, V: Clone> CacheTable<K, V> {
...
+            // In this (likely) case, there's no readers, so we can keep the
+            // value and avoid making an additional clone.
+            *slot_state = SlotState::Ready(vv);
         }
+    }
+}
 
-        Ok(res)
+/// Convenience function to safely remove a cache slot key from the cache, iff
+/// it matches an expected value.  Returns if it did the removal.
```
storopoli:
```suggestion
/// it matches an expected value. Returns if it did the removal.
```

"Trey was here"

crates/storage/src/cache.rs
```diff
@@ -127,74 +142,97 @@ impl<K: Clone + Eq + Hash, V: Clone> CacheTable<K, V> {
...
+            error!("database fetch aborted");
+        }
+
+        if let Ok(Err(e)) = fetch_res.as_ref() {
+            error!(%e, "failed to make database fetch");
+        }
+
+        // And then re-acquire the lock on the slot before handling the result.
+        let mut slot_guard = slot.write().await;
+        trace!("re-acquired slot lock");
```
storopoli: aren't we missing a `trace!("acquired slot lock")` before this?

sapinb:
Updates:
1. A deadlock between `slot_guard` rwlock read and write introduced in earlier commits of the PR is resolved.
2. An existing issue where concurrent reads on same cache slot would return inconsistent error responses is resolved.
    -  all concurrent reads on same slot will fail with correct `DbError` variant if the db read fails (instead of only first one failing with correct `DbError` variant and all other reads failing with `DbError::WorkerFailedStrangely`)
3. **A narrow window for race condition between db read error and cache slot clear still exists.** But with these changes, this scenario will return the previously received `DbError` variant instead of `DbError::CacheLoadFail`. This should make it functionally indistinguishable from 2, resolvable with same retry logic as 2 as required and thus harmless.

This could use some further optimization to reduce cloning, but deferring that for future. 

@delbonis @storopoli

crates/storage/src/cache.rs
sapinb: reduced this to 2 instances from 7.

crates/storage/src/cache.rs
sapinb: added trace for read and write lock acquisition

PR #912, opened 2025-07-02
Simplify checkpoint verification logic

prajwolrg:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


bin/datatool/src/util.rs
```diff
@@ -110,19 +110,28 @@ fn resolve_rollup_vk() -> RollupVerifyingKey {
     #[cfg(all(feature = "sp1-builder", not(feature = "risc0-builder")))]
     {
         use strata_sp1_guest_builder::GUEST_CHECKPOINT_VK_HASH_STR;
+        use zkaleido_sp1_groth16_verifier::SP1Groth16Verifier;
         let vk_buf32: Buf32 = GUEST_CHECKPOINT_VK_HASH_STR
             .parse()
             .expect("invalid sp1 checkpoint verifier key hash");
-        RollupVerifyingKey::SP1VerifyingKey(vk_buf32)
+        let sp1_verifier =
+            SP1Groth16Verifier::load(&sp1_verifier::GROTH16_VK_BYTES, vk_buf32.0).unwrap();
```
MdTeach: NIT: Use .expect("...") with a helpful error message instead of .unwrap() for better error context.

bin/datatool/src/util.rs
```diff
@@ -110,19 +110,28 @@ fn resolve_rollup_vk() -> RollupVerifyingKey {
...
-        let vk_buf32 = vk_u8.into();
-        RollupVerifyingKey::Risc0VerifyingKey(vk_buf32)
+        use zkaleido_risc0_groth16_verifier::Risc0Groth16Verifier;
+        let risc0_verifier = Risc0Groth16Verifier::load(
+            risc0_groth16::verifying_key(),
+            risc0_zkvm::BN254_IDENTITY_CONTROL_ID,
+            risc0_zkvm::ALLOWED_CONTROL_ROOT,
+            Digest::new(GUEST_RISC0_CHECKPOINT_ID),
+        )
+        .unwrap();
```
MdTeach: NIT: Use .expect("...") with a helpful error message instead of .unwrap() for better error context.

crates/consensus-logic/src/checkpoint_verification.rs
```diff
@@ -82,50 +82,28 @@ pub fn verify_proof(
...
-    if !allow_empty && is_empty_proof {
-        return Err(ZkVmError::ProofVerificationError(format!(
-            "Empty proof received for checkpoint {checkpoint_idx}, which is not allowed in strict proof mode. \
-            Check `proof_publish_mode` in rollup_params; set it to a non-strict mode (e.g., `timeout`) to accept empty proofs."
-        )));
+    if expected_public_output != actual_public_output {
+        dbg!(actual_public_output, expected_public_output);
+        return Err(ZkVmError::ProofVerificationError(
+            "Public output mismatch during proof verification".to_string(),
+        ));
```
delbonis, line 98: Should make a ticket to eliminate this duplication in our serialized proof structure.  The proof that we put on-chain should not have this receipt structure, we would construct this receipt from the fields in the checkpoint which we already just compare the receipt to.

Since we're changing the proof structure to support the ASM-based flow, we can probably wait to do that and maybe do it as part of the integration work there.

crates/test-utils/data/sp1_rollup_vk.json
delbonis, line 1: Is there no compact binary representation of this we could be using instead?

delbonis (approved):
This is a big improvement.  I wish we could go farther but I didn't quite realize the limitations.  If you have any ideas on going further (like refactoring/simplifying how `verify_rollup_groth16_proof_receipt` works) then feel free to keep going, but I am happy to accept this as-is.

crates/test-utils/data/sp1_rollup_vk.json
prajwolrg: We can do a binary representation. I opted for the json version, so all the fields of the VK are visible in case anyone wants to look into it.

crates/consensus-logic/src/checkpoint_verification.rs
prajwolrg: Created the [ticket](https://alpenlabs.atlassian.net/browse/STR-1553) to track this.

mukeshdroid (approved):
ACK d721e7a for the crypto owned files.

PR #923, opened 2025-07-09
ASM MohoProgram Impl

prajwolrg:
## Description

- Implements the MohoProgram trait to wrap ASM STF.
- Defines AuxBundle as input for ASM STF. 
- Implements preprocessing for ASM STF

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


crates/asm/common/src/subprotocol.rs
```diff
@@ -29,15 +29,43 @@ pub trait Subprotocol: 'static {
...
     /// subprotocol.
     ///
     /// Updates the subprotocol’s internal state and collects any resulting `InterprotoMsg` and
     /// `Log` on the provided `MsgRelayer`.
-    fn process_txs(state: &mut Self::State, txs: &[TxInput<'_>], relayer: &mut impl MsgRelayer);
+    fn process_txs(
+        state: &mut Self::State,
+        txs: &[TxInput<'_>],
+        anchor_pre: &AnchorState,
+        aux_inputs: &[Self::AuxInput],
```
delbonis, line 69: I think we can make this so it's just one aux input.  If the subprotocol requires more than one then we can just have it define that internally.  It simplifies the parsing from the encoded values.

Also, if we're saying that it makes sense to pass the pre state here, I wonder if it instead makes sense to have some kind of context struct that we can expand further without having to add new args.

crates/asm/common/src/subprotocol.rs
```diff
@@ -90,3 +123,18 @@ pub trait SubprotoHandler {
...
+/// Implementers should record requests (e.g. “I need header for block 123”)
+/// and then fetch all of them in batch before `process_txs` runs.
+///
+/// # Notes
+///
+/// If any requested auxiliary data is malformed or cannot be provided,
+/// the subsequent `process_txs` call will panic.
+pub trait AuxInputCollector {
+    /// Request the L1 header for a specific block number.
+    fn request_header(&mut self, block_number: u64);
```
delbonis: I feel like this should be handled out of band instead of trying to bring it in.  It's mixing layers of abstraction since this is talking about concrete concepts like headers but the aux structures are provided as blobs.

I'm imagining that, since the CSM would have access to this data, it can look at the block, see from the txs that it's going to require certain data like historical headers proofs, and then pre-prepares the aux input for that.

crates/asm/moho-incremental/src/input.rs
```diff
@@ -0,0 +1,96 @@
...
+            .prev_blockhash
+            .to_raw_hash()
+            .to_byte_array();
+        StateReference::new(parent_ref)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+/// A wrapper around Bitcoin's `Block` to provide Borsh (de)serialization.
+pub struct L1Block(pub Block);
```
delbonis: Comments before attributes.

Also, I thought we circled back to the thinking that the STF itself should take a narrower "manifest"-like container that just has things like header, height, list of tagged txs (which we could even still pass by ref!), etc. instead of the full block.

Edit: Ah shit, I'm probably confusing different concepts here.

crates/asm/moho-incremental/src/program.rs
```diff
@@ -0,0 +1,44 @@
+use moho_runtime_interface::MohoProgram;
+use moho_types::{InnerStateCommitment, StateReference};
+use strata_asm_common::AnchorState;
+use strata_asm_stf::{StrataAsmSpec, asm_stf};
+use strata_primitives::hash::compute_borsh_hash;
+
+use crate::input::AsmStepInput;
+
+#[derive(Debug)]
+pub struct AsmMohoIncremetal;
```
delbonis: Should probably be named like `AsmStfProgram`.  This isn't a proof itself.

crates/asm/moho-incremental/src/input.rs
prajwolrg: In that case, we would need to create a separate proof that proves:
1. We correctly extracted all the tagged txs. 
2. The block txdata is consistent with the block header.

From the past experience we know that making one large proof is efficient than splitting the same proof statements into multiple proofs and verifying one proof in another.

crates/asm/moho-incremental/src/program.rs
prajwolrg: Should I also rename the crate?

crates/asm/common/src/subprotocol.rs
prajwolrg: Makes sense!

crates/asm/common/src/subprotocol.rs
prajwolrg: The number of AuxInput depends on the transactions. If the subprotocol has a transaction that requires AuxInput and if multiple of those transactions are present in a block, we need multiple aux input. So, making it a slice of AuxInput made sense to me.

Yea we can create a context struct to encapsulate anchor_pre and aux_inputs.

crates/asm/moho-program/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+moho-types.workspace = true
+strata-asm-common.workspace = true
+strata-asm-stf.workspace = true
+strata-primitives.workspace = true
+
+[dev-dependencies]
+strata-test-utils.workspace = true
+
+[lints]
+workspace = true
```
storopoli: Move up to be consistent with other `Cargo.toml`s in the workspace.

storopoli (changes requested):

crates/asm/common/src/subprotocol.rs
delbonis: > The number of AuxInput depends on the transactions.

It only needs to be dependent on the number of subprotocols.  If there's some information particular to each transaction that the subprotocol cares about, then it can impose that when it defines its aux input structure.

crates/asm/moho-incremental/src/input.rs
delbonis: We would still do all of that in the same proof, that part is fine.

I was trying to say that the `asm_stf` function should not take `&Block`, but should take a manifest-like structure with the protocol txs pre-extracted, possibly even as `TxInput`s.

So maybe

```rs
struct BlockData<'b> {
  header: &'b Header,
  protocol_txs: &'b [TxInput<'b>]
}
```

I am feeling a conflict between being able to reuse the reader code which would generate the new manifests (which means it'd want to just own the contents normally) and the ASM STF code which would want to optimize for avoiding clones.

crates/asm/moho-incremental/src/program.rs
delbonis: Maybe not?  If you do then like `moho-program-impl` would be best, I think.

crates/asm/common/src/log.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    /// Verification key update for the rollup state transition function.
+    OlStfUpdate(OlStfUpdate),
+
+    /// Verification key update for the execution environment state transition.
+    AsmStfUpdate(AsmStfUpdate),
+
+    /// Export state update event carrying export entry data.
+    NewExportEntry(NewExportEntry),
+}
```
delbonis: ASM logs messages are just bytes.  This enum is unnecessary.

crates/asm/common/src/log.rs
```diff
@@ -0,0 +1,77 @@
...
+
+/// Details for an export state update event.
+#[derive(Debug, Clone)]
+pub struct NewExportEntry {
+    /// Export container ID.
+    pub container_id: u16,
+    /// Export entry data.
+    pub entry_data: ExportEntry,
+}
```
delbonis: I think we need messages to delete entries too.  Unless we want to emit all of the exports on every block and regenerate it entirely?  We would also need messages to update the shared container payload too.

crates/asm/moho-program/src/program.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    fn update_export_state(export_state: &mut ExportState, output: &Self::StepOutput) {
+        // Iterate through each AsmLog; if we find an NewExportEntry, add it to ExportState
+        for log in output {
+            if let AsmLog::NewExportEntry(export) = log {
+                export_state.add_entry(export.container_id, export.entry_data.clone());
+            }
+        }
+    }
```
delbonis: We should have some more sophisticated solution here than just adding a new return value to `process_transition`.  It's sloppy.  A simple approach would be a `MohoProgramOutput<P: MohoProgram>` struct which has these as fields, as that gives us more structure.

I am not sure we even necessarily need to have the outputs exposed in this interface.  We could bottle it up entirely, process the outputs as part of the `process_transition` function, and have the changes we want to make to the moho state exposed more declaratively in this `MohoProgramOutput` struct.

crates/asm/stf/src/transition.rs
```diff
@@ -29,7 +31,11 @@ impl AsmSpec for StrataAsmSpec {
 
 /// Computes the next AnchorState by applying the Anchor State Machine (ASM) state transition
 /// function (STF) to the given previous state and new L1 block.
-pub fn asm_stf<S: AsmSpec>(pre_state: AnchorState, block: &Block) -> Result<AnchorState, AsmError> {
+pub fn asm_stf<S: AsmSpec>(
+    pre_state: &AnchorState,
+    block: &Block,
+    aux: &AuxBundle,
+) -> AsmResult<(AnchorState, Vec<AsmLog>)> {
```
delbonis: This is messy for the same reasons as above.  See the `BlockExecutionOutput` struct in the chain exec PR as a similar scenario where we can put more structure here.

crates/asm/moho-incremental/src/input.rs
prajwolrg: In that case, we would need a separate function to generate BlockData from Block. Let me try to do that.

crates/asm/common/src/log.rs
prajwolrg: I was thinking about this, and I think we can replace the entries in the export container by some accumulator (similar to MMR accumulator for L1 headers). We only add to the accumulator after all the pre-conditions have been fulfilled. That means in our case we don't add `BridgeAssignment` information in the ExportEntry, we instead add the `BridgeFulfillment` information. The criteria for the successful Fulfillment can be defined by the BridgeSubprotocol. This will also simplify the bridge proof. The bridge proof then only requires showing the merkle proof of inclusion of BridgeFulfillment. This will also limit the size of the MohoState.

crates/asm/common/src/log.rs
prajwolrg: I don't understand why ASM log messages needs to be bytes and can't be enum.

crates/asm/common/src/log.rs
delbonis: I'm not sure I follow, can you elaborate?  Would that still be safe when reassigning operators?

prajwolrg: > Would that still be safe when reassigning operators?

Yes, because we add ExportEntry only after the operator has fulfilled the withdrawal request "properly" - right amounts, to right users, within the deadline. If any of that is not fulfilled, it is not added to export entry. When we add to ExportEntry there should be no need to reassign anything.

delbonis: Oh that's interesting, then we could make the structure more limited.  Can you develop this more in another doc so we can reason about how it would work more fully?

delbonis (approved):
Approving draft impl based on discussion.

delbonis:
Force merging because issues addressed and failed CI job is just codecov.

PR #933, opened 2025-07-15
Allow el reorg

sapinb:
## Description
Fixes for 2 error cases:
1. when strata client reorgs, alpen-reth client should follow
  a. use reth fork with required changes to allow block generation at fork of ancestor of canonical chain
2. fix missing db entries after strata-client restart during FCM processing (after `handle_new_block`) 

NOTE: The change in this linked reth fork does not need to be upstreamed. An equivalent feature has already been merged into reth 1.5.0. This release is using to reth 1.1.5, thus requiring the fork. This fork does not need to be maintained post testnet 1 lifetime. 

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -132,46 +132,46 @@ revm-primitives = { version = "15.1.0", features = [
 ], default-features = false }
 
 # reth itself:
-reth = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-basic-payload-builder = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-chain-state = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-chainspec = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-cli = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-cli-commands = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-cli-util = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-db = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-errors = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-ethereum-forks = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-ethereum-payload-builder = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-evm = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-evm-ethereum = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-exex = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-ipc = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-network-api = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-node-api = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
-reth-node-builder = { git = "https://github.com/paradigmxyz/reth.git", rev = "v1.1.5" }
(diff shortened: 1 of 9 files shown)


bin/strata-client/src/el_sync.rs
```diff
@@ -70,6 +70,55 @@ pub fn sync_chainstate_to_el(
     Ok(())
 }
 
+/// Sync missing or invalid L2 block status using chainstate.
+///
+/// TODO: retry on network errors
+pub fn sync_chainstate_l2_status(storage: &NodeStorage) -> Result<(), Error> {
```
barakshani, line 74: This function is run only during restart. Does it make sense to run it regularly?

barakshani (dismissed):
Let's deploy from this branch and see if it fixes the issue before merging to main. Thanks for the thorough investigation of the issue!

Cargo.toml
```diff
@@ -132,46 +132,46 @@ revm-primitives = { version = "15.1.0", features = [
...
+reth-rpc-eth-api = { git = "https://github.com/alpenlabs/reth.git", rev = "e4940d6d09067bad" }
+reth-rpc-eth-types = { git = "https://github.com/alpenlabs/reth.git", rev = "e4940d6d09067bad", default-features = false }
+reth-rpc-layer = { git = "https://github.com/alpenlabs/reth.git", rev = "e4940d6d09067bad" }
+reth-rpc-server-types = { git = "https://github.com/alpenlabs/reth.git", rev = "e4940d6d09067bad" }
+reth-rpc-types-compat = { git = "https://github.com/alpenlabs/reth.git", rev = "e4940d6d09067bad" }
+reth-tasks = { git = "https://github.com/alpenlabs/reth.git", rev = "e4940d6d09067bad" }
+reth-transaction-pool = { git = "https://github.com/alpenlabs/reth.git", rev = "e4940d6d09067bad" }
+reth-trie = { git = "https://github.com/alpenlabs/reth.git", rev = "e4940d6d09067bad" }
+reth-trie-common = { git = "https://github.com/alpenlabs/reth.git", rev = "e4940d6d09067bad" }
+reth-trie-db = { git = "https://github.com/alpenlabs/reth.git", rev = "e4940d6d09067bad" }
```
storopoli, line 174: If this gets through. Can we try to upstream this patch to `reth`?
What was the change that you had to do? I could not find in the PR description.

bin/strata-client/src/el_sync.rs
```diff
@@ -70,6 +70,55 @@ pub fn sync_chainstate_to_el(
     Ok(())
 }
 
+/// Sync missing or invalid L2 block status using chainstate.
+///
+/// TODO: retry on network errors
```
storopoli: These two lines should be comments not docstrings

```suggestion
//
// TODO: retry on network errors
```

storopoli (commented):
some comments

Cargo.toml
sapinb: This change does not need to be upstreamed. An equivalent feature has already been merged into reth 1.5.0. This release is using to reth 1.1.5, thus requiring the fork. This fork does not need to be maintained post testnet 1 lifetime.

storopoli: Amazing! Music to my ears. Cc @MdTeach

storopoli: Can you please add

> This change does not need to be upstreamed. An equivalent feature has already been merged into reth 1.5.0. This release is using to reth 1.1.5, thus requiring the fork. This fork does not need to be maintained post testnet 1 lifetime.

Somewhere in the PR description?
This needs to be documented and easily accessible if needed in the future. Once I resolve this thread, it will be buried forever.

barakshani: This is a hotfix for the latest outage we had. Coincidentally, we are working on upgrading to reth 1.5, so the change here will be supported out-of-the-box. But for Testent I, we ship with reth 1.1.5, so we need this fix in.

storopoli (approved):
ACK 3756f926ee48c1ee23894ae2ef7839764a5e6fc5

bin/strata-client/src/el_sync.rs
sapinb: This is run as part of startup checks so it is only run once per start.
This issue should only be present in the final block, but checking everything for sanity. It only uses local db access and no network requests, and the find_last_match is a binary search, so the total overhead to startup time is minimal.

bin/strata-client/src/el_sync.rs
sapinb: updated,

PR #934, opened 2025-07-15
fix: change all taproot sighashes to `Default` and better parsing

storopoli:
## Description

1. Normalize all `TapSighashesType`s to `Default` since we can save 1 byte in all transactions.

2. Make the `validate_deposit_signature` in `strata-l1tx` robust to Taproot Signatures. This is done in pair with the bridge since a `taproot::Signature` can have 64 or 65 bytes depending on having a sighash byte or not.

    From the [BIP-341](https://github.com/bitcoin/bips/blob/83ac8427e7f81cead035728b9c1d925aceddf0d0/bip-0341.mediawiki?plain=1#L139), taproot signatures can have 64 or 65 bytes:

    > A Taproot signature is a 64-byte Schnorr signature, as defined in BIP340, with the sighash byte appended in the usual Bitcoin fashion. This sighash byte is optional. If omitted, the resulting signatures are 64 bytes, and a `SIGHASH_DEFAULT` mode is implied.

    This is in the BIP for "Taproot: SegWit version 1 spending rules" (BIP-341) hence we should make our parsing rules robust to that.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Merge together with https://github.com/alpenlabs/strata-bridge/pull/217.

Needs to be backported to `releases/0.2.0`.

## Checklist

## Related Issues

STR-1581

### DIFF
--- a/crates/l1tx/src/deposit/deposit_tx.rs
+++ b/crates/l1tx/src/deposit/deposit_tx.rs
@@ -5,10 +5,10 @@ use bitcoin::{
     key::TapTweak,
     opcodes::all::OP_RETURN,
     sighash::{Prevouts, SighashCache},
-    taproot::TAPROOT_CONTROL_NODE_SIZE,
-    Amount, OutPoint, ScriptBuf, TapNodeHash, TapSighashType, Transaction, TxOut, XOnlyPublicKey,
+    taproot::{self, TAPROOT_CONTROL_NODE_SIZE},
+    Amount, OutPoint, ScriptBuf, TapNodeHash, Transaction, TxOut, XOnlyPublicKey,
 };
-use secp256k1::{constants::SCHNORR_SIGNATURE_SIZE, schnorr::Signature, Message};
+use secp256k1::Message;
 use strata_primitives::{
     buf::Buf32,
     l1::{DepositInfo, OutputRef},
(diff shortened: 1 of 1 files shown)


crates/l1tx/src/deposit/deposit_tx.rs
```diff
@@ -92,10 +92,28 @@ fn validate_deposit_signature(
         return None;
     }
     let sig_witness = &input.witness[0];
-    if sig_witness.len() < SCHNORR_SIGNATURE_SIZE {
-        return None;
-    }
-    let sig_bytes = &sig_witness[..SCHNORR_SIGNATURE_SIZE];
+
+    // Handle both 64-byte (SIGHASH_DEFAULT) and 65-byte (explicit sighash) signatures.
```
Rajil1213: Use [`taproot:Signature::from_slice`](https://docs.rs/bitcoin/latest/bitcoin/taproot/struct.Signature.html#method.from_slice) instead.

Rajil1213 (changes requested):
Parsing should use the method from `rust-bitcoin` instead.

As this makes the OL parsing of deposit signature agnostic to the sighash type, this should be merged and deployed before the corresponding PR in [`strata-bridge`](https://github.com/alpenlabs/strata-bridge/pull/217).

crates/l1tx/src/deposit/deposit_tx.rs
storopoli: Done in 1ef8e0c8bcabd2fc2f23a8fe98bb95944be29ebc

AaronFeickert:
Reviewed this with @storopoli.

PR #942, opened 2025-07-18
cd.yml supports different docker repos for main and releases/0.2.0 branch

voidash:
### Type of Change

- [x] CI/CD Change

## Related Issues

STR-1539


purusang (approved):
If we are planning staging-eest workflow to run in main only then its fine.

PR #948, opened 2025-07-22
STR-1593 fix l1 reader sync event crash

sapinb:
## Description
- [x] CSM should propagate error to task manager to trigger shutdown/restart
- [x] prevent db inconsistency between l1 and sync_event dbs due to crash during btcio reader db writes

Note: in btcio reader, setting an l1 block as canonical and posting its corresponding sync event needs to be atomic in current implementation. As a quick fix, these operations have been put in a single db transaction, but this breaks the separation between l1 and sync_event dbs. 
Need to better handle cross database (and also cross system) atomicity, or ensure the code does not require atomicity in regards to persistence.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/btcio/src/reader/handler.rs
+++ b/crates/btcio/src/reader/handler.rs
@@ -1,5 +1,6 @@
 use bitcoin::{consensus::serialize, hashes::Hash, Block};
 use bitcoind_async_client::traits::Reader;
+use strata_common::check_bail_trigger;
 use strata_primitives::{
     buf::Buf32,
     l1::{
@@ -74,7 +75,7 @@ async fn handle_blockdata<R: Reader>(
 
     storage.l1().put_block_data_async(manifest).await?;
 
-    strata_common::check_bail_trigger("btcio_pre_l1_write");
+    check_bail_trigger("btcio_pre_l1_write");
 
     // Create a sync event if it's something we care about.
     let blkid: Buf32 = blockdata.block().block_hash().into();
@@ -86,7 +87,7 @@ async fn handle_blockdata<R: Reader>(
         .await?;
     info!(%height, %l1blockid, txs = %num_txs, sync_ev = %sync_event_idx, "wrote L1 block manifest");
 
-    strata_common::check_bail_trigger("btcio_post_l1_write");
+    check_bail_trigger("btcio_post_l1_write");
 
     sync_evs.push(sync_event_idx);
 
--- a/crates/btcio/src/reader/handler.rs
+++ b/crates/btcio/src/reader/handler.rs
@@ -25,12 +25,14 @@ pub(crate) async fn handle_bitcoin_event<R: Reader>(
             // L1 reorgs will be handled in L2 STF, we just have to reflect
             // what the client is telling us in the database.
             let height = block.height();
-            ctx.storage
+            let sync_event = SyncEvent::L1Revert(block);
+            let sync_event_idx = ctx
+                .storage
                 .l1()
-                .revert_canonical_chain_async(height)
+                .revert_canonical_chain_and_write_sync_event_async(height, sync_event)
                 .await?;
             debug!(%height, "reverted L1 block database");
-            vec![SyncEvent::L1Revert(block)]
+            vec![sync_event_idx]
         }
 
         L1Event::BlockData(blockdata, epoch, hvs) => {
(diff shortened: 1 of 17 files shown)


crates/consensus-logic/src/csm/ctl.rs
```diff
@@ -60,4 +60,15 @@ impl EventSubmitter for CsmController {
...
+
+    async fn submit_event_idx_async(&self, ev_idx: u64) -> anyhow::Result<()> {
+        let msg = CsmMessage::EventInput(ev_idx);
+        if self.csm_tx.send(msg).await.is_err() {
+            warn!(%ev_idx, "sync event receiver closed when submitting");
+        } else {
+            trace!(%ev_idx, "sent csm event input");
+        }
+
+        Ok(())
```
storopoli: This does not need to be a `Result` return type. There's no way this can return the `Error` variant.

```suggestion
    async fn submit_event_idx_async(&self, ev_idx: u64) {
        let msg = CsmMessage::EventInput(ev_idx);
        if self.csm_tx.send(msg).await.is_err() {
            warn!(%ev_idx, "sync event receiver closed when submitting");
        } else {
            trace!(%ev_idx, "sent csm event input");
        }
```

crates/btcio/src/reader/handler.rs
```diff
@@ -71,16 +73,22 @@ async fn handle_blockdata<R: Reader>(
     let l1blockid = *manifest.blkid();
 
     storage.l1().put_block_data_async(manifest).await?;
-    storage
-        .l1()
-        .extend_canonical_chain_async(&l1blockid)
-        .await?;
-    info!(%height, %l1blockid, txs = %num_txs, "wrote L1 block manifest");
+
+    strata_common::check_bail_trigger("btcio_pre_l1_write");
```
storopoli: Can you please import the function?

crates/btcio/src/reader/handler.rs
```diff
@@ -71,16 +73,22 @@ async fn handle_blockdata<R: Reader>(
...
     let block_commitment = L1BlockCommitment::new(height, blkid.into());
-    sync_evs.push(SyncEvent::L1Block(block_commitment));
+    let sync_event = SyncEvent::L1Block(block_commitment);
+    let sync_event_idx = storage
+        .l1()
+        .extend_canonical_chain_and_write_sync_event_async(&l1blockid, sync_event)
+        .await?;
+    info!(%height, %l1blockid, txs = %num_txs, sync_ev = %sync_event_idx, "wrote L1 block manifest");
+
+    strata_common::check_bail_trigger("btcio_post_l1_write");
```
storopoli: Can you please import the function?

storopoli (changes requested):
Minor nits.

Why this is not being pushed to `main` first, then backported to `releases/0.2.0`?

crates/consensus-logic/src/csm/ctl.rs
sapinb: updated.

crates/btcio/src/reader/handler.rs
sapinb: updated.

crates/btcio/src/reader/handler.rs
sapinb: updated.

storopoli (approved):
ACK 9ae26271f4de1489c7d6725f5aff1650cd6614a5

delbonis:
Oh since this is being merged directly into release then ig we can accept the debt here since it doesn't impact ongoing work.

functional-tests/mixins/seq_crash_mixin.py
```diff
@@ -22,14 +26,22 @@ def premain(self, ctx: flexitest.RunContext):
...
         assert protocol_version is not None, "Sequencer RPC inactive"
 
-    def handle_bail(self, bail_tag: Callable[[], str], **kwargs) -> int:
+    @abstractmethod
+    def get_recovery_metric(self) -> T:
+        """
+        Abstract method to get a metric value used to determine crash recovery.
+        Derived classes should override this to return their desired recovery metric.
+        """
+        pass
```
delbonis, line 35: Technically this should raise `NotImplementedError`.

delbonis (approved):

PR #950, opened 2025-07-23
feat(asm): implement core subprotocol with proof verification

irnb:
## Description
ASM Core Subprotocol 

## Next Steps 

- [ ] Create tasks for implementing `placeholder` methods when requirements are finalized
- [ ] Test subprotocol with real blocks and transactions in integration environment
- [ ]  ssz  in ASM 

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

You can search for `[PLACE_HOLDER]` in the code to identify areas pending design, spec, or consensus on a solution.

## Checklist

## Related Issues
 
STR-1333


crates/asm/common/src/constants.rs
```diff
@@ -0,0 +1,15 @@
...
+/// and must match the `subprotocol_id` checked in `SectionState::subprotocol()`.
+pub const CORE_SUBPROTOCOL_ID: SubprotocolId = 1;
+
+/// The unique identifier for the BridgeV1 subprotocol within the Anchor State Machine.
+pub const BRIDGE_SUBPROTOCOL_ID: SubprotocolId = 2;
+
+/// Core Subprotocol Transaction type identifiers
+pub const OL_STF_CHECKPOINT_TX_TYPE: TxType = 1;
+pub const FORCED_INCLUSION_TX_TYPE: TxType = 2;
+pub const EE_UPGRADE_TX_TYPE: TxType = 3;
```
prajwolrg: I'm not sure if we should define all the constants of all the subprotocols in a single file in the common crate. I can definitely see the value in this, but I'm not sure if this is the right approach.

crates/asm/common/src/msg.rs
```diff
@@ -32,12 +37,142 @@ impl<const ID: SubprotocolId> InterprotoMsg for NullMsg<ID> {
...
+///
+/// We store the type and body separately for Borsh compatibility while
+/// maintaining full SPS-msg-fmt compliance through encoding/decoding.
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize, PartialEq, Eq)]
+pub struct OLToASMMessage {
+    /// Message type identifier
+    ty: TypeId,
+    /// Message body
+    body: Vec<u8>,
+}
```
prajwolrg: Are there benefits to using this approach than just creating a new type like in https://github.com/alpenlabs/alpen/blob/main/crates/asm/common/src/log.rs#L39 ? This seems like a better approach, but I'm not sure.

crates/asm/common/src/subprotocol_genesis.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// * `config` - The genesis configuration to register
+    pub fn register<T: BorshSerialize>(
+        &mut self,
+        id: SubprotocolId,
+        config: &T,
+    ) -> Result<(), AsmError> {
+        let serialized = borsh::to_vec(config).map_err(|e| AsmError::Serialization(id, e))?;
+        self.configs.insert(id, serialized);
+        Ok(())
+    }
```
prajwolrg: We can make the function generic on Subprotocol instead. `config` would then be `Subprotocol::GenesisConfig`.

crates/asm/moho-program/src/program.rs
```diff
@@ -42,7 +42,10 @@ impl MohoProgram for AsmStfProgram {
             aux_input: &input.aux_bundle,
         };
 
-        asm_stf::<StrataAsmSpec>(pre_state, stf_input).unwrap()
+        // TODO: Evaluate whether genesis_registry is needed during preprocessing
+        // pass empty `genesis_registry` for now
```
prajwolrg: I think this is required. It will be exactly similar to the processing phase.

crates/asm/common/src/subprotocol.rs
```diff
@@ -39,29 +75,60 @@ pub trait Subprotocol: 'static {
...
+    ///   deserialized into Self::GenesisConfig before use. Each subprotocol can decide whether to
+    ///   accept None (empty genesis) or require proper configuration.
     ///
-    /// # Parameters
-    /// - `state`    – current subprotocol state snapshot (read-only)
-    /// - `txs`      – list of new L1 transactions to examine
-    /// - `collector`– hook for registering requests for auxiliary inputs
+    /// # Returns
+    /// The initialized state or an error if deserialization/initialization fails
+    fn init(genesis_config_data: Option<&[u8]>) -> Result<Self::State, AsmError>;
```
prajwolrg: Does it make sense for the init function to take `Self::GenesisConfig` as the argument? We can set it to be None, in case it is not required by the subprotocols.

crates/asm/stf/src/preprocess.rs
```diff
@@ -62,9 +66,14 @@ pub fn pre_process_asm<'b, S: AsmSpec>(
     let mut manager = SubprotoManager::new();
 
     // 3. LOAD: Initialize each subprotocol in the subproto manager.
-    // We use empty auxpayload in the loader stage as no auxiliary data is needed during loading.
+    // We use empty aux_payload in the loader stage as no auxiliary data is needed during loading.
     let aux = BTreeMap::new();
-    let mut loader_stage = SubprotoLoaderStage::new(pre_state, &mut manager, &aux);
+
+    // TODO: Evaluate whether genesis_registry is needed during preprocessing
+    // pass empty `genesis_registry` for now
```
prajwolrg: Yes, it is needed.

crates/asm/subprotocols/core/src/handlers/checkpoint.rs
```diff
@@ -0,0 +1,119 @@
...
+use crate::{CoreOLState, error::*, messages, parsing, types, verification};
+
+/// Handles OL STF checkpoint transactions according to the specification
+///
+/// This function implements the complete checkpoint verification workflow:
+///
+/// 1. **Extract and validate** the signed checkpoint from transaction data
+/// 2. **Verify signature** using the current sequencer public key
+/// 3. **Verify ASM zk-SNARK proof** using the current verifying key
+/// 4. **Construct expected public parameters** from local ASM state
```
prajwolrg: Core subprotocol is not verifying the ASM STF Proof. It's verifying the OL STF Proof.

crates/asm/subprotocols/core/src/lib.rs
```diff
@@ -44,34 +66,77 @@ impl Subprotocol for OLCoreSubproto {
...
-        todo!()
+        for tx in txs {
+            let result = handlers::route_transaction(state, tx, anchor_pre, aux_inputs, relayer);
+
+            // TODO: Implement proper logging approach
+            // Since this code also runs as a part of zkVM guest program, we cannot use the
+            // `tracing` crate. We need a proper logging mechanism to identify which
+            // transaction processing failed and why. For now, we print errors to stderr
+            // as a temporary solution.
+            //
```
prajwolrg: I was thinking about this and I think we can redefine the tracing crate macros for the zkVM guest programs, so it becomes a noop something like:

```rust
// 1) When *not* building for zkvm, just pull in the real tracing macros:
#[cfg(not(target_os = "zkvm"))]
pub use tracing::{error, warn, info, debug, trace};

// 2) When building for zkvm, define your own zero-cost shim macros:
#[cfg(target_os = "zkvm")]
#[macro_export]
macro_rules! error { ($($arg:tt)*) => {{ let _ = format_args!($($arg)*); }} }

#[cfg(target_os = "zkvm")]
#[macro_export]
macro_rules! warn  { ($($arg:tt)*) => {{ let _ = format_args!($($arg)*); }} }

#[cfg(target_os = "zkvm")]
#[macro_export]
macro_rules! info  { ($($arg:tt)*) => {{ let _ = format_args!($($arg)*); }} }

#[cfg(target_os = "zkvm")]
#[macro_export]
macro_rules! debug { ($($arg:tt)*) => {{ let _ = format_args!($($arg)*); }} }

#[cfg(target_os = "zkvm")]
#[macro_export]
macro_rules! trace { ($($arg:tt)*) => {{ let _ = format_args!($($arg)*); }} }
```

crates/asm/subprotocols/core/src/messages/l1_to_l2.rs
```diff
@@ -0,0 +1,107 @@
...
+/// * `end_height` - Ending L1 block height for message range
+/// * `commitment_hash` - Hash commitment to the message range
+///
+/// # Returns
+/// Result indicating if the message range is valid
+pub(crate) fn validate_l1_to_l2_messages(
+    _start_height: u64,
+    _end_height: u64,
+    _commitment_hash: &Buf32,
+) -> Result<()> {
```
prajwolrg: I think this is where we would require AuxInput and AnchorState.

crates/asm/subprotocols/core/src/parsing/inscription.rs
```diff
@@ -0,0 +1,75 @@
...
+/// The extracted data from the inscription envelope or parsing error
+pub(crate) fn parse_inscription_envelope(tx: &TxInputRef<'_>) -> Result<Vec<u8>> {
+    // [PLACE_HOLDER]
+    // TODO: Implement proper Bitcoin inscription parsing
+    // This should:
+    // 1. Parse the transaction witness stack
+    // 2. Extract inscription envelope data
+    // 3. Validate inscription format
+    // 4. Return the embedded checkpoint data
+
```
prajwolrg: Parsing of inscription is done as part of `strata-l1tx`. Can we use that?

prajwolrg: It uses `TxFilterConfig` right now, but we might be able to get away with it.

crates/asm/subprotocols/core/src/types.rs
```diff
@@ -0,0 +1,130 @@
...
+/// This structure contains all necessary parameters to properly initialize
+/// the Core subprotocol state.
+///
+/// This struct sharing the same fields as CoreOLState but i create this
+/// separately to avoid confusion (for now).
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize)]
+pub struct CoreGenesisConfig {
+    /// The initial checkpoint verifying key for zk-SNARK proof verification
+    /// Stored as serialized bytes for Borsh compatibility.
+    pub checkpoint_vk_bytes: Vec<u8>,
```
prajwolrg: zkaleido defines the `VerifyingKey` type, that is just a wrapper around Vec<u8>. We can use that here.

prajwolrg (changes requested):
Good work! Also really great docstrings!

crates/asm/common/src/constants.rs
delbonis: I agree, they are not part of the base ASM system, they're part of the subprotocol crates.

crates/asm/common/src/msg.rs
delbonis: Agree on this.  Also the name `OLToASMMessage` is also really verbose.  As I said before, we should have a single term to describe this.

The ty/body formatting is also something that should not be presented here.  The OL logs are opaque, but we format them according to the msgfmt spec.

crates/asm/common/src/constants.rs
delbonis: EE upgrades are not done through ASM.  EEs are an OL concept and upgrades are managed on OL.

crates/asm/common/src/subprotocol.rs
delbonis: I agree, it should be deserialized by the wrapper and passed in as a concrete type.

crates/asm/common/src/subprotocol.rs
```diff
@@ -146,7 +220,7 @@ pub trait SubprotoHandler {
 ///
 /// # Parameters
 ///
-/// - `data`: an opaque byte slice whose meaning is defined entirely by the collector’s
+/// - `data`: an opaque byte slice whose meaning is defined entirely by the collector's
 ///   implementation.
```
delbonis: I'm confused on what this is referring to now.

crates/asm/stf/src/stage.rs
```diff
@@ -36,9 +41,20 @@ impl Stage for SubprotoLoaderStage<'_, '_> {
...
+            // State not found in the anchor state, which occurs in two scenarios:
+            // 1. During genesis block processing, before any state initialization
+            // 2. When introducing a new subprotocol to an existing chain
+            // In either case, we must initialize a fresh state from the provided configuration in
+            // genesis_registry
+            None => {
+                // Try to get genesis config data from registry
+                let genesis_config_data = self.genesis_registry.get_raw(S::ID);
+
+                S::init(genesis_config_data).expect("asm: failed to initialize subprotocol state")
```
delbonis: This is where we'd parse it according to `S::GenesisData`.

crates/asm/subprotocols/core/src/handlers/checkpoint.rs
```diff
@@ -0,0 +1,119 @@
...
+///                   (e.g. withdrawal, etc.)
+///
+///
+/// # Security Notes
+///
+/// - Proof public parameters should constructed from our own state, not sequencer input
+/// - All state transitions are validated for proper progression
+/// - Proof verification uses verifying key from state
+/// - L1→L2 message commitments are verified against expected range
+pub(crate) fn handle(
```
delbonis: What is being `handle`d?

crates/asm/subprotocols/core/src/handlers/checkpoint.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    // 8. Validate L1→L2 Message Range using the rolling hash
+    let prev_l1_height = state.verified_checkpoint.new_l1().height();
+    let new_l1_height = checkpoint.batch_info().final_l1_block().height();
+    let expected_commitment = &public_params.l1_to_l2_msgs_range_commitment_hash;
+    messages::l1_to_l2::validate_l1_to_l2_messages(
+        prev_l1_height,
+        new_l1_height,
+        expected_commitment,
+    )?;
```
delbonis: This checkpoint validation logic needs to be thought about more explicitly.  See draft spec doc in DM.

For example, the checkpoint structure needs to be heavily reworked.  The receipt structure should not have been included, we construct the receipt from the current state and the other fields on the checkpoint.  We should be treating the proof like a signature that's passed along with the checkpoint data to authenticate it, not treating a proof-with-receipt as its own structure we compare against the rest of the checkpoint data.  This is just how it was written originally, but we need to rethink it.

crates/asm/common/src/msg.rs
```diff
@@ -32,12 +37,142 @@ impl<const ID: SubprotocolId> InterprotoMsg for NullMsg<ID> {
...
+            target_subprotocol,
+            messages,
+        }
+    }
+}
+
+impl InterprotoMsg for MessagesContainer {
+    fn id(&self) -> SubprotocolId {
+        self.target_subprotocol
+    }
```
delbonis: We don't directly pass these OL logs as inter-proto messages.  The bridge subprotocol should export a opaque enum type for the messages it expects to receive from other subprotocols.

crates/asm/subprotocols/core/src/handlers/checkpoint.rs
```diff
@@ -0,0 +1,119 @@
...
+        let mut bridge_messages = Vec::new();
+        for ol_msg in &public_params.l2_to_l1_msgs {
+            if let Ok(decoded) = ol_msg.decode() {
+                bridge_messages.push(decoded);
+            }
+        }
+
+        if !bridge_messages.is_empty() {
+            let container =
+                MessagesContainer::with_messages(BRIDGE_SUBPROTOCOL_ID, bridge_messages);
```
delbonis: The way this is written assumes that all logs are bridge messages.  We should parse the OL messages and if they are withdrawal intent messages then we send deposit intent interproto messages to the bridge subprotocol.

Unfortunately we still have to define what withdrawal intent OL messages look like, should DM about this.

crates/asm/subprotocols/core/src/handlers/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+use crate::{CoreOLState, OLCoreSubproto, error::*};
+
+/// Routes transactions to appropriate handlers based on transaction type
+pub(crate) fn route_transaction(
+    state: &mut CoreOLState,
+    tx: &TxInputRef<'_>,
+    _anchor_pre: &AnchorState,
+    _aux_inputs: &[<OLCoreSubproto as Subprotocol>::AuxInput],
+    relayer: &mut impl MsgRelayer,
+) -> Result<()> {
```
delbonis: Stylistic: don't really like functions that do stuff being in mod.rs/lib.rs files like this.

crates/asm/subprotocols/core/src/handlers/forced_inclusion.rs
delbonis, line 1: This can just be removed.  We probably won't be handling forced incls via the checkpoint subprotocol.

crates/asm/subprotocols/core/src/messages/l1_to_l2.rs
delbonis: Yeah and this is going to be somewhat hard to write without some of the surrounding infra.  It might make sense to either leave this stubbed off (and just trust the checkpoints include the messages) or to just not check the aux data against the ASM history.

crates/asm/subprotocols/core/src/types.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    /// The initial verified checkpoint state (usually genesis checkpoint)
+    pub initial_checkpoint: EpochSummary,
+
+    /// The initial L1 block reference for the checkpoint
+    pub initial_l1_ref: L1BlockId,
```
delbonis: We'd pick the height ahead of genesis actually happening, so we can't know this.

crates/asm/subprotocols/core/src/types.rs
```diff
@@ -0,0 +1,130 @@
...
+/// This struct sharing the same fields as CoreOLState but i create this
+/// separately to avoid confusion (for now).
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize)]
+pub struct CoreGenesisConfig {
+    /// The initial checkpoint verifying key for zk-SNARK proof verification
+    /// Stored as serialized bytes for Borsh compatibility.
+    pub checkpoint_vk_bytes: Vec<u8>,
+
+    /// The initial verified checkpoint state (usually genesis checkpoint)
+    pub initial_checkpoint: EpochSummary,
```
delbonis: We'd probably want to scope this more narrowly to just be the genesis blkid.  Everything else is either unknowable until the ASM genesis L1 block *or* is just 0.

delbonis (changes requested):
Couple things:

* These crates shouldn't do any of the heavy lifting for doing checkpoint validation.  This should be in a separate crate that's implemented according to the spec, and then this just calls into that.
* I've been kinda trying to rename the "OL core" subprotocol to "OL checkpoint", since "core" is a bit ambiguous and calling it "checkpoint" keeps is responsibilities narrowly defined.

crates/asm/common/src/msg.rs
```diff
@@ -32,12 +37,129 @@ impl<const ID: SubprotocolId> InterprotoMsg for NullMsg<ID> {
...
+    pub fn encode(&self) -> Result<Vec<u8>, AsmError> {
+        let mut result = Vec::new();
+        strata_msg_fmt::try_encode_into_buf(
+            self.0.ty(),
+            self.0.body().iter().copied(),
+            &mut result,
+        )
+        .map_err(AsmError::from)?;
+        Ok(result)
+    }
```
delbonis: The [`Msg` trait](https://github.com/alpenlabs/strata-common/blob/main/crates/msg-fmt/src/lib.rs#L37) has a `to_vec` function that's basically this, and cannot error.

crates/asm/common/src/msg.rs
```diff
@@ -32,12 +37,129 @@ impl<const ID: SubprotocolId> InterprotoMsg for NullMsg<ID> {
...
+
+    /// Returns the raw encoded message bytes (alias for encode)
+    pub fn encoded(&self) -> Result<Vec<u8>, AsmError> {
+        self.encode()
+    }
```
delbonis: Not sure the purpose of this fn?

crates/asm/common/src/msg.rs
```diff
@@ -32,12 +37,129 @@ impl<const ID: SubprotocolId> InterprotoMsg for NullMsg<ID> {
...
+}
+
+impl BorshSerialize for Message {
+    fn serialize<W: std::io::Write>(&self, writer: &mut W) -> std::io::Result<()> {
+        // Serialize as (ty, body) tuple for Borsh compatibility
+        (self.0.ty(), self.0.body().to_vec()).serialize(writer)
+    }
+}
+
+impl BorshDeserialize for Message {
```
delbonis, line 98: If the purpose of this wrapper type was for borsh support then we could implement these traits on the upstream type.  However, in any case where we don't actually have to parse the messages, it's better to treat them as opaque u8 vecs.  So all serialization should be with byte vecs if it's not *explictly* required to be a valid messages.  It's technically permitted for "logs" to not be formatted as valid "messages".

crates/asm/common/src/msg.rs
```diff
@@ -32,12 +37,129 @@ impl<const ID: SubprotocolId> InterprotoMsg for NullMsg<ID> {
...
+pub type L2ToL1Msg = Message;
+
+/// Generic container for multiple messages following SPS-msg-fmt
+#[derive(Clone, Debug)]
+pub struct MessagesContainer {
+    /// The target subprotocol ID
+    pub target_subprotocol: SubprotocolId,
+    /// Messages using strata-msg-fmt
+    pub messages: Vec<OwnedMsg>,
+}
```
delbonis: Should probably remove this to avoid confusion.

crates/asm/common/src/subprotocol_genesis.rs
delbonis, line 1: This module could probably just be called `genesis`.

crates/asm/logs/src/constants.rs
```diff
@@ -1,20 +1,21 @@
...
+macro_rules! define_ids {
+    ($type:ty, $fn_name:ident, $($name:ident = $value:expr),* $(,)?) => {
         $(
-            pub const $name: TypeId = $value;
+            pub const $name: $type = $value;
         )*
 
         /// Get all defined type IDs as an array
-        pub const fn all_type_ids() -> &'static [TypeId] {
+        pub const fn $fn_name() -> &'static [$type] {
```
delbonis: Can't this be

```rs
pub const LOG_TYPE_IDS: &[$type] = &[$($name),*];
```

instead of having to have the wrapping fn?

crates/asm/stf/src/stage.rs
```diff
@@ -36,9 +41,26 @@ impl Stage for SubprotoLoaderStage<'_, '_> {
...
+            // In either case, we must initialize a fresh state from the provided configuration in
+            // genesis_registry
+            None => {
+                // Deserialize genesis config from registry, or use default if not found
+                let genesis_config: S::GenesisConfig =
+                    self.genesis_registry.get(S::ID).unwrap_or_else(|| {
+                        // For subprotocols that use () as GenesisConfig, this will work
+                        // For other subprotocols, they should provide config in registry
+                        borsh::from_slice(&[])
+                            .expect("asm: subprotocol requires genesis config but none provided")
```
delbonis: This would panic if a subprotocol wasn't specified to have a genesis value but *does* have a non-`()` state type.  I feel like this is fine?  Not sure what we should do about that.

crates/asm/subprotocols/core/src/constants.rs
```diff
@@ -0,0 +1,53 @@
...
+
+// Define all subprotocol IDs
+define_ids! {SubprotocolId, all_subprotocol_ids,
+    CORE_SUBPROTOCOL_ID = 1,
+    BRIDGE_SUBPROTOCOL_ID = 2,
+}
+
+// Define all transaction type IDs
+define_ids! {TxType, all_core_tx_type_ids,
+    OL_STF_CHECKPOINT_TX_TYPE = 1,
```
delbonis, line 26: We'd want to export this def from where we define the transaction types, which would be in the strata-common repo.

crates/asm/subprotocols/core/src/messages/mod.rs
```diff
@@ -0,0 +1,9 @@
+//! Message processing for the Core subprotocol
+//!
+//! This module handles L1↔L2 message processing, validation, and forwarding.
+
+pub(crate) mod l1_to_l2;
+pub(crate) mod l2_to_l1;
```
delbonis: Probably shouldn't be pub(crate) if the other items being exported are pub(crate).

crates/asm/subprotocols/core/src/parsing/mod.rs
```diff
@@ -0,0 +1,6 @@
+//! This module handles checkpoint data extraction.
+
+pub(crate) mod checkpoint;
```
delbonis: Same as in other comment.

crates/asm/subprotocols/core/src/verification/mod.rs
```diff
@@ -0,0 +1,11 @@
+//! Verification logic for the Core subprotocol
+//!
+//! This module contains all verification functionality including signature verification,
+//! proof verification, and state transition validation.
+
+pub(crate) mod proof;
+pub(crate) mod signature;
+pub(crate) mod state_transition;
+
```
delbonis: Same as other comment.

crates/asm/subprotocols/core/src/types.rs
```diff
@@ -0,0 +1,115 @@
...
+
+impl CoreGenesisConfig {
+    /// Create a new genesis config with the given rollup verifying key
+    pub fn new(
+        checkpoint_vk: &VerifyingKey,
+        genesis_l1_block: L1BlockCommitment,
+        sequencer_pubkey: Buf32,
+    ) -> Result<Self> {
+        Ok(Self {
+            checkpoint_vk: checkpoint_vk.clone(),
```
delbonis: `checkpoint_vk` be passed by value.

storopoli:
Labeling this as "P2 - Nice to have" since it is T2.
Once we are release T1, we might revisit the label.

crates/asm/subprotocols/core/src/constants.rs
irnb: I added a `TODO` comment above this line as a reminder to fix it when we export from the strata_common repo.

PR #956, opened 2025-07-26
feat(asm): implement bridge subprotocol

prajwolrg:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


crates/asm/subprotocols/bridge-v1/src/errors.rs
```diff
@@ -0,0 +1,123 @@
...
+        expected: BitcoinAmount,
+        actual: BitcoinAmount,
+    },
+
+    /// Withdrawal destination doesn't match assignment destination
+    #[error("Withdrawal destination mismatch: expected {expected}, got {actual}")]
+    DestinationMismatch {
+        expected: ScriptBuf,
+        actual: ScriptBuf,
+    },
```
delbonis: All of these error messages should be formatted as `thing wrong (expected {expected}, got {got})`.

So for example: `operator mismatch (expected {expected}, got {got})`

Also in the other error types.

crates/asm/subprotocols/bridge-v1/src/lib.rs
```diff
@@ -1,65 +1,27 @@
...
+//!
+//! The main entry point is [`subprotocol::BridgeV1Subproto`] which implements the `Subprotocol`
+//! trait for integration with the Anchor State Machine.
+
+pub mod constants;
+pub mod errors;
+pub mod msgs;
+pub mod state;
+pub mod subprotocol;
+pub mod txs;
```
delbonis: I've been more recently leaning more towards wanting to not expose the internal module structure in crates unless there's a good reason to for understanability.  We should make these non-`pub` and then only export the items that are actually needed in the public interface.

crates/asm/subprotocols/bridge-v1/src/txs/withdrawal_fulfillment/create.rs
delbonis, line 1: Seems like this should all not be in the in this crate, doesn't this at least partially exist in the l1tx crate?

crates/asm/subprotocols/bridge-v1/src/txs/withdrawal_fulfillment/parse.rs
```diff
@@ -0,0 +1,172 @@
...
+};
+
+use crate::{
+    errors::WithdrawalParseError,
+    txs::withdrawal_fulfillment::USER_WITHDRAWAL_FULFILLMENT_OUTPUT_INDEX,
+};
+
+const OPERATOR_IDX_SIZE: usize = std::mem::size_of::<OperatorIdx>();
+const DEPOSIT_IDX_SIZE: usize = std::mem::size_of::<u32>();
+const DEPOSIT_TXID_SIZE: usize = std::mem::size_of::<Txid>();
```
delbonis: This technically works but it's referring to a different concept.  We don't define these in terms of their in-memory sizes, we define them as being 4/4/32 bytes, so we should just write that explicitly.

crates/asm/subprotocols/bridge-v1/src/txs/withdrawal_fulfillment/parse.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    let mut offset = 0;
+    let operator_idx_bytes = &withdrawal_auxdata[offset..offset + OPERATOR_IDX_SIZE];
+
+    offset += OPERATOR_IDX_SIZE;
+    let deposit_idx_bytes = &withdrawal_auxdata[offset..offset + DEPOSIT_IDX_SIZE];
+
+    offset += DEPOSIT_IDX_SIZE;
+    let deposit_txid_bytes = &withdrawal_auxdata[offset..offset + DEPOSIT_TXID_SIZE];
```
delbonis: Since these are all constants we can just compute the offsets statically and then copy the bytes out statically.

crates/asm/subprotocols/bridge-v1/src/txs/withdrawal_fulfillment/parse.rs
```diff
@@ -0,0 +1,172 @@
...
+    );
+
+    let deposit_idx = u32::from_be_bytes(
+        deposit_idx_bytes
+            .try_into()
+            .expect("deposit index is exactly 4 bytes"),
+    );
+
+    let deposit_txid: Txid =
+        encode::deserialize(deposit_txid_bytes).expect("deposit txid is exactly 32 bytes");
```
delbonis: I usually do

```rs
let mut operator_idx_buf = [0u8; 4];
operator_idx_buf.copy_from_slice(...);
let operator_idx = u32::from_be_bytes(operator_idx_buf);
```

so we don't have to do this fake `.expect` error handling which would be optimized out anyways.

crates/asm/subprotocols/bridge-v1/src/subprotocol.rs
```diff
@@ -0,0 +1,221 @@
...
+        relayer: &mut impl MsgRelayer,
+    ) {
+        // Process each transaction based on its type
+        for tx in txs {
+            match tx.tag().tx_type() {
+                DEPOSIT_TX_TYPE => Self::process_deposit_tx(state, tx),
+                WITHDRAWAL_TX_TYPE => Self::process_withdrawal_tx(state, tx, relayer),
+                _ => continue, // Ignore unsupported transaction types
+            }
+        }
```
delbonis: This should parse the tx into an enum and then dispatch on the enum.  The current structure couples parsing and handling into the same pass, which makes testing more of a rat's nest.

crates/asm/subprotocols/bridge-v1/src/msgs.rs
```diff
@@ -0,0 +1,9 @@
+use borsh::{BorshDeserialize, BorshSerialize};
+
+use crate::state::withdrawal::WithdrawalCommand;
+
+/// Message type that we receive messages from other subprotocols using.
+#[derive(Clone, Debug, Eq, PartialEq, BorshDeserialize, BorshSerialize)]
+pub enum BridgeIncomingMsg {
+    ProcessWithdrawal(WithdrawalCommand),
```
delbonis: `DispatchWithdrawal` might be a better name.

crates/asm/subprotocols/bridge-v1/src/subprotocol.rs
```diff
@@ -0,0 +1,221 @@
...
+        );
+    }
+
+    /// Processes a withdrawal fulfillment transaction with error logging.
+    fn process_withdrawal_tx(
+        state: &mut BridgeV1State,
+        tx: &TxInputRef<'_>,
+        relayer: &mut impl MsgRelayer,
+    ) {
+        let withdrawal_info = match extract_withdrawal_info(tx) {
```
delbonis: These two functions being structured the way they are is a good illustration of what I meant in the other comment.  They're called `process_deposit_tx` and `process_withdrawal_tx`, but we haven't actually validated that they're structured like those txs.  We see the tx tag and say "we should process this tx expecting it to be a foo tx", but we haven't validated that it does plausibly look like a foo tx by the time we're dispatching it down to the handler function.  Ideally, we should extract all of the relevant data for validation out of the tx while we're parsing it, and then we can just have these handler functions just operate on that instead of having to look at the tx in detail.

I realize this is trying to be a direct adaption of what we've already written, so this is partially coming from that legacy design, but this is the direction we should try to be moving in.

crates/asm/subprotocols/bridge-v1/src/txs/deposit/create.rs
```diff
@@ -0,0 +1,321 @@
...
+    sighash::{Prevouts, SighashCache, TapSighashType},
+};
+use musig2::{FirstRound, KeyAggContext, SecNonceSpices};
+use rand::{RngCore, rngs::OsRng};
+use strata_l1tx::utils::generate_agg_pubkey;
+use strata_primitives::buf::Buf32;
+
+use crate::txs::deposit::parse::DepositInfo;
+
+pub const TEST_MAGIC_BYTES: &[u8; 4] = b"ALPN";
```
delbonis: Should be defined in a test module probably.

delbonis (changes requested):
Good start putting everything together, but should clean up component boundaries a little more.

crates/asm/subprotocols/bridge-v1/src/txs/withdrawal_fulfillment/create.rs
prajwolrg: No, the withdrawal fulfillment tx doesn't exist in the l1tx crate. Right now we don't parse withdrawal fulfillment tx in strata at all. It is only handled in the bridge.

crates/asm/subprotocols/bridge-v1/src/txs/withdrawal_fulfillment/parse.rs
prajwolrg: Yea, right! Thanks!

crates/asm/subprotocols/bridge-v1/src/txs/deposit/create.rs
prajwolrg: I opted out of that because we would have to pass in magic bytes as argument in all the tests. This  module is already gated with `cfg(test)`.

crates/asm/subprotocols/bridge-v1/src/txs/withdrawal_fulfillment/create.rs
prajwolrg: Yes, it does. 

There are two major differences:
1. The first output is not SPS-50 compatible. All the metadata is in the last output.
2. The extraction logic requires `TxFilterConfig`, which doesn't make much sense now.

I think for the time being it's fine to have two different implementation.

crates/asm/subprotocols/bridge-v1/src/errors.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+#[derive(Debug, Error)]
+pub enum WithdrawalCommandError {
+    /// No unassigned deposits are available for processing
+    #[error("No unassigned deposits available for withdrawal command processing")]
+    NoUnassignedDeposits,
```
Rajil1213, line 153: This error feels panic-worthy. Is there any case where this is possible without breaking the peg?

crates/asm/subprotocols/bridge-v1/src/lib.rs
```diff
@@ -1,65 +1,27 @@
...
+//! and withdrawal processing between Bitcoin L1 and the orchestration layer.
+//!
+//! # Architecture
+//!
+//! The bridge consists of several key components:
+//!
+//! - **Operators**: Entities that process withdrawals and maintain bridge security
+//! - **Deposits**: Bitcoin UTXOs locked to N/N multisig operator addresses
+//! - **Assignments**: Task assignments linking deposits to specific operators
+//! - **Withdrawals**: Commands for operators to release funds back to Bitcoin
```
Rajil1213:
```suggestion
//! - **Withdrawals**: Commands for operators to release funds from the multisig.
```

The funds never really leave Bitcoin.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -0,0 +1,727 @@
...
+    /// This operator fronts the funds for the withdrawal and will be
+    /// reimbursed by the bridge notaries upon successful execution.
+    current_assignee: OperatorIdx,
+
+    /// List of operators who were previously assigned to this withdrawal.
+    ///
+    /// When a withdrawal is reassigned, the current assignee is moved to this
+    /// list before a new operator is selected. This prevents reassigning to
+    /// operators who have already failed to execute the withdrawal.
+    previous_assignees: Vec<OperatorIdx>,
```
Rajil1213, line 49: NOTE: this is technically new logic (in contradiction to the claim in the PR description).

Ultimately, this change is better because it will prevent cycle lengths that are smaller than the total operator count.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -0,0 +1,727 @@
...
+            previous_assignees: Vec::new(),
+            exec_deadline,
+        })
+    }
+
+    /// Returns the deposit index associated with this assignment.
+    ///
+    /// # Returns
+    ///
+    /// The deposit index as [`u32`].
```
Rajil1213: This feels redundant. The same for the rest of the accessors with a `Returns` section.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -0,0 +1,727 @@
...
+/// - **`exec_deadline`** - Bitcoin block height deadline for execution
+///
+/// # Execution Deadline
+///
+/// The `exec_deadline` represents the latest Bitcoin block height at which
+/// the withdrawal can be executed. If a checkpoint is processed after this
+/// height and the withdrawal hasn't been completed, the assignment becomes invalid.
+#[derive(
+    Clone, Debug, Eq, PartialEq, BorshDeserialize, BorshSerialize, Serialize, Deserialize, Arbitrary,
+)]
```
Rajil1213, line 30: Feature Request: it might be useful to implement `PartialOrd`, `Ord` for this type which would then allow us to use [`SortedVec`](https://github.com/alpenlabs/alpen/blob/dae809cb7154f5174e2dbfa394fbb34e3f3b46a0/crates/primitives/src/sorted_vec.rs#L19) in the `AssignmentTable`.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -0,0 +1,727 @@
...
+/// - Creating new assignments with optimized insertion
+/// - Looking up assignments by deposit index
+/// - Filtering assignments by operator or expiration status
+/// - Removing completed or expired assignments
+#[derive(Clone, Debug, Eq, PartialEq, BorshDeserialize, BorshSerialize)]
+pub struct AssignmentTable {
+    /// Vector of assignment entries, sorted by deposit index.
+    ///
+    /// **Invariant**: MUST be sorted by `AssignmentEntry::deposit_idx` field.
+    assignments: Vec<AssignmentEntry>,
```
Rajil1213: We can enforce this invariant more neatly by implementing `PartialOrd` on `AssignmentEntry` and using `SortedVec` here.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -0,0 +1,727 @@
...
+
+impl AssignmentTable {
+    /// Creates a new empty assignment table.
+    ///
+    /// Initializes the table with no assignments, ready for operator
+    /// assignment management.
+    ///
+    /// # Returns
+    ///
+    /// A new empty [`AssignmentTable`].
```
Rajil1213: This and the rest of the doc comments on the accessors are redundant.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -0,0 +1,727 @@
...
+
+    /// Reassigns all expired assignments to new randomly selected operators.
+    ///
+    /// Iterates through all assignments and reassigns those whose execution deadlines
+    /// have passed (current height >= exec_deadline). Each expired assignment is
+    /// reassigned using the provided seed for deterministic random operator selection.
+    ///
+    /// This method handles bulk reassignment of expired assignments, ensuring that
+    /// withdrawals don't get stuck due to unresponsive operators. If any individual
+    /// reassignment fails (e.g., no eligible operators), the entire operation fails
```
Rajil1213, line 354: I think it bears repeating here what `eligible` means.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -0,0 +1,461 @@
...
+/// - Once assigned, indices are never reused (no deletion support)
+#[derive(Clone, Debug, Eq, PartialEq, BorshDeserialize, BorshSerialize)]
+pub struct OperatorTable {
+    /// Next unassigned operator index for new registrations.
+    next_idx: OperatorIdx,
+
+    /// Vector of registered operators, sorted by operator index.
+    ///
+    /// **Invariant**: MUST be sorted by `OperatorEntry::idx` field.
+    operators: Vec<OperatorEntry>,
```
Rajil1213: Another instance where using a `SortedVec` can help maintain this invariant more cleanly.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -0,0 +1,461 @@
...
+    /// Updates the multisig membership status for multiple operators, inserts new operators,
+    /// and recalculates the aggregated key.
+    ///
+    /// This is the central method for updating multisig membership. All other methods that modify
+    /// the multisig set should call this method internally to ensure the aggregated key is
+    /// correctly recalculated.
+    ///
+    /// # Parameters
+    ///
+    /// - `updates` - Slice of (operator_index, is_in_multisig) pairs for existing operators
```
Rajil1213, line 255: This should note that `updates` can override `inserts`.

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -0,0 +1,912 @@
...
+        let actual_txid = withdrawal_info.deposit_txid.clone();
+        if expected_txid != actual_txid {
+            return Err(WithdrawalValidationError::DepositTxidMismatch {
+                expected: expected_txid,
+                actual: actual_txid,
+            });
+        }
+
+        // Validate withdrawal amount against assignment command
+        let expected_amount = assignment.withdrawal_command().amt();
```
Rajil1213: BLOCKER: This isn't accounting for operator fees, is it?

crates/asm/subprotocols/bridge-v1/src/state/withdrawal.rs
```diff
@@ -0,0 +1,157 @@
...
+/// operators to construct the appropriate Bitcoin transaction output.
+///
+/// # Bitcoin Descriptors
+///
+/// The destination uses Bitcoin Output Script Descriptors (BOSD) which provide
+/// a standardized way to specify Bitcoin addresses and locking conditions.
+#[derive(
+    Clone, Debug, Eq, PartialEq, BorshDeserialize, BorshSerialize, Serialize, Deserialize, Arbitrary,
+)]
+#[serde(rename_all = "snake_case")]
```
Rajil1213: Nit: This is redundant, right?

crates/asm/subprotocols/bridge-v1/src/txs/deposit/create.rs
```diff
@@ -0,0 +1,321 @@
...
+        .map(|sk| {
+            let pk = PublicKey::from_secret_key(&secp, sk);
+            let (x_only, parity) = pk.x_only_public_key();
+            let adjusted_pk =
+                PublicKey::from_x_only_public_key(x_only, bitcoin::secp256k1::Parity::Even);
+            let adjusted_sk = if parity == bitcoin::secp256k1::Parity::Odd {
+                sk.negate()
+            } else {
+                *sk
+            };
```
Rajil1213: Use the [`EvenSecretKey`](https://github.com/alpenlabs/alpen/blob/dae809cb7154f5174e2dbfa394fbb34e3f3b46a0/crates/primitives/src/crypto.rs#L57) type instead for better type safety.

crates/asm/subprotocols/bridge-v1/src/txs/deposit/create.rs
Rajil1213: Not just this constant but looks like this entire module is only used in tests. It might be worth cleaning it up so as to not mix actual library code with testing code.

crates/asm/subprotocols/bridge-v1/src/txs/deposit/create.rs
```diff
@@ -0,0 +1,321 @@
...
+/// signature validation against the aggregated operator public key.
+///
+/// # Arguments
+/// - `deposit_info`: The deposit information containing index, amount, destination, and tapscript
+///   root
+/// - `operators_privkeys`: Slice of private keys corresponding to the operator public keys
+///
+/// # Returns
+/// The properly formatted and signed Bitcoin transaction
+pub fn create_test_deposit_tx(
```
Rajil1213: Instead of redoing the logic of creating the deposit transaction, you should just use a test vector like in the `l1tx` test. It should be the bridge that cares about the internal logic of creating a deposit transaction, the ASM (afaict) only cares about parsing it.

https://github.com/alpenlabs/alpen/blob/dae809cb7154f5174e2dbfa394fbb34e3f3b46a0/crates/l1tx/src/filter/mod.rs#L239-L242

crates/asm/subprotocols/bridge-v1/src/txs/deposit/parse.rs
```diff
@@ -0,0 +1,311 @@
...
+
+    /// The destination address for the deposit.
+    pub address: Vec<u8>,
+
+    /// The outpoint of the deposit transaction.
+    pub outpoint: OutputRef,
+
+    /// The tapnode hash (merkle root) from the Deposit Request Transaction (DRT) being spent.
+    ///
+    /// This value is extracted from the auxiliary data and represents the merkle root of the
```
Rajil1213, line 39:
```suggestion
    /// This value is extracted from the auxiliary data and represents the merkle proof of the
```

crates/asm/subprotocols/bridge-v1/src/txs/deposit/create.rs
Rajil1213: My bad. Looks like the `create` module is already gated with `cfg(test)`.

Rajil1213 (changes requested):
Agreed with Trey's review on mixing the validation and processing logic for the different transactions. Other than that, I've left some nits and commentary but there aren't any major blockers from me.

I do want to flag that this PR has such a large surface and touches so many critical parts that it makes me nervous. If the commits are atomic enough, we should consider not squashing these commits so that we can bisect more easily should there arise any issues in the future.

crates/asm/subprotocols/bridge-v1/src/subprotocol.rs
prajwolrg: Make sense. Tried to do it as part of [937a665](https://github.com/alpenlabs/alpen/pull/956/commits/937a6654a278f560e2846990b4de746d04d8119b).

crates/asm/subprotocols/bridge-v1/src/subprotocol.rs
prajwolrg: Implemented it as part of [937a665](https://github.com/alpenlabs/alpen/pull/956/commits/937a6654a278f560e2846990b4de746d04d8119b)

crates/asm/subprotocols/bridge-v1/src/txs/deposit/create.rs
prajwolrg: This is just for testing purpose. 

With SPS-50, we require OP_RETURN to be the first output, but the test vector has it as the second output, so couldn't use the test vector directly.

crates/asm/subprotocols/bridge-v1/src/txs/deposit/create.rs
prajwolrg: Thanks you. I didn't realize we had this already.

crates/asm/subprotocols/bridge-v1/src/state/withdrawal.rs
prajwolrg: Yes. thanks.

crates/asm/subprotocols/bridge-v1/src/errors.rs
delbonis: I agree this is a little weird.  I need to think about how we want to handle errors here, since a error bubbling up means we can't produce a proof, but maybe we actually do want to capture and handle them somehow so we can manually intervene later.  Should we set a "something unexpected happened" flag and allow manual intervention on an accelerated timeline?

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
delbonis: I agree it's fine to make some minor algorithmic changes if it improves integration.

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
prajwolrg: Yes. Thanks for pointing it out. Implemented in [3b84a2e](https://github.com/alpenlabs/alpen/pull/956/commits/3b84a2ea2b3d92cd4bcf39bb5cbe2382aa1c72ba)

crates/asm/subprotocols/bridge-v1/src/errors.rs
prajwolrg: Yes, this breaks the peg. 

Not sure what the best approach to handle here is. We don't want to panic if there is a way to recover from this with upgrades. But if there is no way to recover from the situation, i.e. the peg has already been broken it also doesn't make sense to upgrade. So panic might be a good option.

This is really tricky. This situation would likely arise only if the peg has already been broken.

crates/l1tx/src/utils.rs
```diff
@@ -58,23 +58,35 @@ pub fn next_u32(instructions: &mut Instructions<'_>) -> Option<u32> {
...
+///
+/// # Panics
+///
+/// Panics if any key in the iterator is not a valid x-only public key.
+pub fn generate_agg_pubkey<'k>(
+    keys: impl Iterator<Item = &'k Buf32>,
+) -> anyhow::Result<XOnlyPublicKey> {
+    let keys = keys.map(|op| {
         PublicKey::from_x_only_public_key(
             XOnlyPublicKey::from_slice(op.as_ref()).expect("slice not an x-only public key"),
```
bewakes: Since we are already returning a result, I think it's better to return error if not x-only public key rather than panicing.

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -0,0 +1,897 @@
...
+    deposits: DepositsTable,
+
+    /// Table of operator assignments for withdrawal processing.
+    assignments: AssignmentTable,
+
+    /// The amount of bitcoin expected to be locked in the N/N multisig.
+    denomination: BitcoinAmount,
+
+    /// The duration (in blocks) for assignment execution deadlines.
+    deadline_duration: u64,
```
bewakes, line 52: Seems like we expect these to change during the lifetime of a bridge version, right? because otherwise I wouldn't expect these here.

crates/asm/subprotocols/bridge-v1/src/subprotocol.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// 2. **Peg Mechanism Failure**: The bridge's peg to Bitcoin has been compromised, potentially
+    ///    due to operator collusion or critical implementation bugs
+    ///
+    /// Both conditions represent unrecoverable protocol violations where continued operation
+    /// poses significant risk of fund loss.
+    fn process_msgs(state: &mut Self::State, msgs: &[Self::Msg]) {
+        for msg in msgs {
+            match msg {
+                BridgeIncomingMsg::DispatchWithdrawal(withdrawal_cmd) => {
+                    // TODO: Pass actual L1BlockId instead of placeholder
```
bewakes, line 149: why can't we get a `L1BlockId` here? does it need interface changes?

crates/asm/subprotocols/bridge-v1/src/txs/withdrawal_fulfillment/parse.rs
```diff
@@ -0,0 +1,188 @@
...
+        let parser = ParseConfig::new(*TEST_MAGIC_BYTES);
+        let tag_data = parser.try_parse_tx(&tx).expect("Should parse transaction");
+        let tx_input_ref = TxInputRef::new(&tx, tag_data);
+
+        // Extract withdrawal info using the actual parser
+        let extracted_info = extract_withdrawal_info(&tx_input_ref)
+            .expect("Should successfully extract withdrawal info");
+
+        assert_eq!(extracted_info, info);
+    }
```
bewakes, line 191: We can/should test for all the invalid cases as well.

crates/asm/subprotocols/bridge-v1/src/txs/handler.rs
```diff
@@ -0,0 +1,34 @@
...
+
+            // FIXME: This is a placeholder for the actual container ID logic.
+            let container_id = 0; // Replace with actual logic to determine container ID
+            let withdrawal_processed_log =
+                NewExportEntry::new(container_id, unlock.to_export_entry());
+            relayer.emit_log(AsmLogEntry::from_log(&withdrawal_processed_log).expect("FIXME:"));
+
+            Ok(())
+        }
+    }
```
bewakes, line 47: I suggest adding tests for emitted logs, if not done for the function calling this function. It might also make more sense to return logs here and emit from the caller. Makes testing easier.

crates/asm/subprotocols/bridge-v1/src/txs/deposit/create.rs
```diff
@@ -0,0 +1,314 @@
+//! Test utilities for creating deposit transactions.
```
bewakes: Looks like a util function. It might make more sense to keep it inside a `test_utils` module.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -0,0 +1,442 @@
...
+            panic!(
+                "Cannot create operator table with empty entries - at least one operator is required"
+            );
+        }
+        let agg_operator_key = generate_agg_pubkey(entries.iter().map(|o| o.wallet_pk()))
+            .unwrap()
+            .into();
+        Self {
+            next_idx: entries.len() as OperatorIdx,
+            operators: SortedVec::new_unchecked(
```
bewakes, line 178: Why unchecked? Are we sure this will be sorted?

bewakes: I see it is mentioned sequential indices in the docstrings. But I would rely more on types and this doesn't seem to create overhead sorting the operators list which won't be very large.

bewakes (commented):
Awesome work! Some suggestions and queries.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
prajwolrg: Since we enumerate over the list and directly use the enumeration, we can be sure it sorted. Are you suggesting we create a `SortedVec` first and insert each one individually?

crates/asm/subprotocols/bridge-v1/src/subprotocol.rs
prajwolrg: Yes, will do that later after this PR separately.

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
prajwolrg: I don't think they will be updated very soon. But putting it in the state allows it to be updated directly without updating the VerifyingKey of the whole ASM.

crates/asm/subprotocols/bridge-v1/src/txs/handler.rs
prajwolrg: Instead of doing the tests for emitted logs here, it makes sense to do this test outside of each subprotocol and in the ASM STF level, to see if the expected logs are emitted. I think we can add those tests later.

delbonis: I actually agree we should have narrower tests here that don't require running the whole system.  Splitting parsing and handling gets us part of the way there.  Don't think this is urgent though, since this is still in flux.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
delbonis: There *is* a constructor fn to force a sort, but this is probably fine as-is (maybe add a debug assert?).

crates/asm/subprotocols/bridge-v1/src/errors.rs
prajwolrg: Added panic in case of failure to assign deposit [fa69618](https://github.com/alpenlabs/alpen/pull/956/commits/fa69618a20eb1ebf72e4e8fb5f7b8ab1c85fd5cb)

delbonis (approved):

crates/asm/subprotocols/bridge-v1/src/errors.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// This indicates the transaction was not signed by the expected operator set.
+    #[error("Deposit signature validation failed: {reason}")]
+    InvalidSignature { reason: String },
+
+    /// The deposit amount does not match the expected amount for this bridge configuration.
+    #[error("Invalid deposit amount")]
+    MismatchDepositAmount(Mismatch<u64>),
+
+    /// A deposit with this index already exists in the deposits table.
+    /// This should not occur since deposit indices are guaranteed unique by the N/N multisig.
```
barakshani, line 87: I'm not sure that I follow the explanation here: the bridge operator _should_ take care of it, but I'm not sure it is "guaranteed".

crates/asm/subprotocols/bridge-v1/src/errors.rs
```diff
@@ -0,0 +1,144 @@
...
+    #[error("Deposit index {0} already exists in deposits table")]
+    DepositIdxAlreadyExists(u32),
+
+    /// Cannot create deposit entry with empty operators list.
+    /// Each deposit must have at least one notary operator.
+    #[error("Cannot create deposit entry with empty operators.")]
```
barakshani, line 93: [note to self] need to understand what this is about

crates/asm/subprotocols/bridge-v1/src/errors.rs
```diff
@@ -0,0 +1,144 @@
...
+    EmptyOperators,
+}
+
+#[derive(Debug, Error)]
+pub enum WithdrawalParseError {
```
barakshani, line 102: I have some trouble with this error, as well as the previous one for deposits. Basically everyone can send this kind of tx, so if something doesn't work, it does not necessarily mean a system error. I'm not sure if returning an error here is the correct behaviour. It's definitely good for testing period, as well for alerts, but there's the conceptual level in which we (I suppose) don't want to emit errors for non-error events as well as the option that someone will spam the system with invalid withdrawal requests.
I can't really make my mind on this, so discussion on the topic is welcome.

crates/asm/subprotocols/bridge-v1/src/errors.rs
barakshani: > Should we set a "something unexpected happened" flag and allow manual intervention on an accelerated timeline?

I'm kinda in favour. I can already imagine this incident on mainnet because of a bug : )   : (

crates/asm/subprotocols/bridge-v1/src/lib.rs
```diff
@@ -1,65 +1,32 @@
-//! BridgeV1 Subprotocol
-use borsh::{BorshDeserialize, BorshSerialize};
-use strata_asm_common::{
-    AnchorState, AsmError, AuxInputCollector, MsgRelayer, NullMsg, Subprotocol, SubprotocolId,
-    TxInputRef,
-};
+//! Bridge V1 Subprotocol
+//!
+//! This crate implements the Strata-Bridge subprotocol.
```
barakshani:
```suggestion
//! This crate implements the Strata bridge subprotocol.
```
This is the bridge subprotocol of Strata, i.e. Strata (bridge subprotocol).

crates/asm/subprotocols/bridge-v1/src/lib.rs
```diff
@@ -1,65 +1,32 @@
...
+//!
+//! The bridge manages Bitcoin deposits, operator assignments,
```
barakshani:
```suggestion
//! The bridge manages Bitcoin deposits, withdrawal assignments,
```
Not sure if that is what you meant.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -0,0 +1,665 @@
+//! Operator Assignment Management
+//!
+//! This module contains types and tables for managing operator assignments to deposits.
+//! Assignments link specific deposits to operators who are responsible for processing
+//! withdrawal requests within specified deadlines.
```
```suggestion
//! Assignments link specific deposit UTXOs to operators who are responsible for processing
//! the corresponding withdrawal requests within specified deadlines.

File: crates/asm/subprotocols/bridge-v1/src/state/assignment.rs [hunk 38]
```diff
@@ -0,0 +1,665 @@
...
+    bridge::{BitcoinBlockHeight, OperatorIdx},
+    buf::Buf32,
+    l1::{BitcoinAmount, BitcoinTxid, L1BlockId},
+    sorted_vec::SortedVec,
+};
+
+use super::withdrawal::WithdrawalCommand;
+use crate::{errors::WithdrawalCommandError, state::deposit::DepositEntry};
+
+/// Assignment entry linking a deposit to an operator for withdrawal processing.
```
```suggestion
/// Assignment entry linking a deposit UTXO to an operator for withdrawal processing.

File: crates/asm/subprotocols/bridge-v1/src/state/assignment.rs [hunk 39]
```diff
@@ -0,0 +1,665 @@
...
+///
+/// Each assignment represents a task assigned to a specific operator to process
+/// a withdrawal from a particular deposit.
```
```suggestion
/// Each assignment represents a task, assigned to a specific operator, to process
/// a withdrawal of a particular deposit UTXO.

File: crates/asm/subprotocols/bridge-v1/src/state/assignment.rs [hunk 40]
```diff
@@ -0,0 +1,665 @@
...
+    /// Deposit entry that has been assigned
+    deposit_entry: DepositEntry,
+
+    /// Withdrawal command specifying outputs and amounts.
+    withdrawal_cmd: WithdrawalCommand,
+
+    /// Index of the operator currently assigned to execute this withdrawal.
+    ///
+    /// This operator fronts the funds for the withdrawal and will be
+    /// reimbursed by the bridge notaries upon successful execution.
```
barakshani: I'm not sure what's the value in writing this, but we should also keep in mind that we have cooperative path. The interplay between the assignee and the cooperative path is yet to be determined.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
barakshani: I disagree in general (punt those to a tech-debt ticket, or implement in a separate twin PR), though my objection is soft.

In this particular case, we don't have any particular design for re-assignments or concrete product requirements, hence we cannot conclude this is an improvement or even desired.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -0,0 +1,665 @@
...
+    ///
+    /// When a withdrawal is reassigned, the current assignee is moved to this
+    /// list before a new operator is selected. This prevents reassigning to
+    /// operators who have already failed to execute the withdrawal.
+    previous_assignees: Vec<OperatorIdx>,
+
+    /// Bitcoin block height deadline for withdrawal execution.
+    ///
+    /// The withdrawal must be executed before this block height.
+    /// If a checkpoint is processed at or after this height without
```
barakshani: what checkpoint? the assigning checkpoint?

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -0,0 +1,665 @@
...
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.deposit_entry.cmp(&other.deposit_entry)
+    }
+}
+
+/// Filters and returns eligible operators for assignment or reassignment.
+///
+/// Returns a list of operators who meet all eligibility criteria:
+/// - Must be part of the deposit's notary operator set
+/// - Must not have previously been assigned to this withdrawal (prevents reassignment to failed
```
barakshani, line 74: this is new (as in https://github.com/alpenlabs/alpen/pull/956/files#r2242050781), right?

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -0,0 +1,665 @@
...
+///   operators)
+/// - Must be currently active in the network
```
barakshani, line 76: what does this mean?

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -0,0 +1,665 @@
...
+    ///
+    /// - `Ok(AssignmentEntry)` - A new assignment entry with randomly selected operator
+    /// - `Err(WithdrawalCommandError::NoEligibleOperators)` - If no eligible operators are
+    ///   available
+    pub fn create_with_random_assignment(
+        deposit_entry: DepositEntry,
+        withdrawal_cmd: WithdrawalCommand,
+        exec_deadline: BitcoinBlockHeight,
+        current_active_operators: &[OperatorIdx],
+        seed: L1BlockId,
```
barakshani, line 126: We need some check on this L1 block. If there are no constrains than the sequencer (or whoever makes the assignment) can grind on L1 blocks until assigned to a favourite operator. I couldn't find where we call this function, but somewhere we should have this constraint + we should prove correct execution.
If needed, that is not everything is implemented by this PR, let's open a tech-debt ticket for this.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -0,0 +1,665 @@
...
+    /// let current_height = BitcoinBlockHeight::from(1000);
+    /// let active_operators = vec![1, 2, 3];
+    /// let seed = L1BlockId::from([0u8; 32]);
+    ///
+    /// table.reassign_expired_assignments(current_height, &active_operators, seed)?;
+    /// ```
+    pub fn reassign_expired_assignments(
+        &mut self,
+        operator_fee: BitcoinAmount,
+        current_height: BitcoinBlockHeight,
```
barakshani, line 381: this value is mellable -- do we prove this somehow?

crates/asm/subprotocols/bridge-v1/src/errors.rs
prajwolrg: Yes, we rely on 1/N honesty bridge assumption for this. 

This error is emitted in case there is a duplicate somehow.

crates/asm/subprotocols/bridge-v1/src/errors.rs
prajwolrg: Yes, you're correct, anyone can send it and it can result in an error. But it's different on how we handle those error. 

We generally log the error and continue with other transactions, there's no system error that causes any panics. The error I think makes sense, because it is indeed an error when parsing, or validating.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
prajwolrg: Yes, the docstring is confusing. Let me update it.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
prajwolrg: Yes, this is new.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
prajwolrg: This is defining the criteria to select a valid operator for a withdrawal assignment. I understand this is new and is not specified in the PRD, but I think it makes sense.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
prajwolrg: There will be a check on the L1Block. It'll be passed as an argument when we process the subprotocol. The ASM STF is responsible for checking it's a valid L1Block and is following all the consensus rules of bitcoin.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
prajwolrg: I don't understand. The values are part of the BridgeState, and we can only use the right values.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
prajwolrg: This is true regardless of the cooperative path.

I have updated the docstring to be more precise.

bewakes (dismissed):

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -0,0 +1,897 @@
...
+    /// - The deposit index already exists in the deposits table
+    pub fn process_deposit_tx(
+        &mut self,
+        tx: &Transaction,
+        info: &DepositInfo,
+    ) -> Result<(), DepositValidationError> {
+        // Validate the deposit first
+        self.validate_deposit(tx, info)?;
+
+        let notary_operators = self.operators().current_multisig_indices();
```
barakshani: where's the code for `current_multisig_indices()`?

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -0,0 +1,897 @@
+//! Bridge state types.
+//!
+//! This just implements a very simple n-of-n multisig bridge.  It will be
```
barakshani: what do you mean by "implements n-of-n bridge"? where does the n-of-n come into play in this code?

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -0,0 +1,897 @@
...
+        if self.deposits().get_deposit(info.deposit_idx).is_some() {
+            return Err(DepositValidationError::DepositIdxAlreadyExists(
+                info.deposit_idx,
+            ));
+        }
+
+        Ok(())
+    }
+
+    /// Creates a withdrawal assignment by selecting an unassigned deposit and assigning it to an
```
barakshani:
```suggestion
    /// Creates a withdrawal assignment by selecting an unassigned deposit UTXO and assigning it to an

File: crates/asm/subprotocols/bridge-v1/src/state/bridge.rs [hunk 49]
```diff
@@ -0,0 +1,897 @@
...
+    /// operator.
+    ///
+    /// This function handles incoming withdrawal commands by:
```
barakshani, line 205: Is "withdrawal command" different than withdrawal request? Any reason to use this terminology (over the term we used so far)?

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -0,0 +1,897 @@
...
+    /// 1. Finding a deposit that has not been assigned yet
```
barakshani:
```suggestion
    /// 1. Finding a deposit UTXO that has not been assigned yet

File: crates/asm/subprotocols/bridge-v1/src/state/bridge.rs [hunk 51]
```diff
@@ -0,0 +1,897 @@
...
+    /// 2. Randomly selecting an operator from the current multisig set that is also a notary
+    ///    operator for that deposit
+    /// 3. Creating an assignment linking the deposit to the selected operator with a deadline
```
barakshani: utxo

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -0,0 +1,897 @@
...
+    ///    calculated from the current block height plus the configured deadline duration
+    ///
+    /// # Parameters
+    ///
+    /// - `withdrawal_cmd` - The withdrawal command specifying outputs and amounts
+    /// - `l1_block_id` - The L1 block ID used as seed for random operator selection
+    /// - `current_block_height` - The current Bitcoin block height for deadline calculation
```
barakshani, line 216: why are these different?

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -0,0 +1,897 @@
...
+    ///
+    /// - `Ok(())` - If the withdrawal assignment was successfully created
+    /// - `Err(WithdrawalCommandError)` - If no suitable deposit/operator combination could be found
+    ///
+    /// # Errors
+    ///
+    /// Returns error if:
+    /// - No unassigned deposits are available
+    /// - No current multisig operators are notary operators for any unassigned deposit
+    /// - The deposit for the unassigned index is not found
```
barakshani: utxo X 2

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -0,0 +1,897 @@
...
+        self.assignments.reassign_expired_assignments(
+            self.operator_fee,
+            current_block_height,
+            &self.operators().current_multisig_indices(),
+            *l1_block_id,
+        )
+    }
+
+    /// Processes a withdrawal fulfillment transaction by validating it, removing the deposit, and
+    /// removing the assignment.
```
barakshani: removing from where? pls explain in the text as below

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -0,0 +1,897 @@
...
+
+        Ok(OperatorClaimUnlock {
+            withdrawal_txid: tx.compute_txid().into(),
+            deposit_txid: removed_assignment.deposit_txid(),
+            deposit_idx: removed_assignment.deposit_idx(),
+            operator_idx: withdrawal_info.operator_idx,
+        })
+    }
+
+    /// Validates the parsed withdrawal it against assignment information.
```
barakshani: please rephrase

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -0,0 +1,897 @@
...
+    /// - `Ok(())` - If the withdrawal is valid according to assignment information
+    /// - `Err(WithdrawalValidationError)` - If validation fails for any reason
+    ///
+    /// # Errors
+    ///
+    /// Returns error if:
+    /// - No assignment exists for the referenced deposit
+    /// - The operator doesn't match the assigned operator
+    /// - The withdrawal specifications don't match the assignment
+    fn validate_withdrawal(
```
barakshani: This is the bridge withdrawal after a fulfillment (and claim), right? why do we even needs this logic on the OL?

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -0,0 +1,897 @@
...
+        &self,
+        withdrawal_info: &WithdrawalFulfillmentInfo,
+    ) -> Result<(), WithdrawalValidationError> {
+        let deposit_idx = withdrawal_info.deposit_idx;
+
+        // Check if an assignment exists for this deposit
+        let assignment = self
+            .assignments
+            .get_assignment(deposit_idx)
```
barakshani, line 334: didn't we remove this information upon `process_withdrawal_fulfillment_tx()`?

crates/asm/subprotocols/bridge-v1/src/state/deposit.rs
```diff
@@ -0,0 +1,397 @@
...
+    /// Bitcoin UTXO reference (transaction hash + output index).
+    output: OutputRef,
+
+    /// Historical set of operators that formed the N/N multisig for this deposit.
+    ///
+    /// This preserves the specific operators who controlled the multisig when the
+    /// deposit was locked, since the active operator set may change over time.
+    /// Any one honest operator from this set can process user withdrawals.
+    ///
+    /// TODO: Convert this to a windowed bitmap for better memory efficiency.
```
barakshani, line 59: create a tech-debt ticket?

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
prajwolrg: It's in [operator.rs](https://github.com/alpenlabs/alpen/blob/STR-1317-bridge-asm/crates/asm/subprotocols/bridge-v1/src/state/operator.rs#L242)

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -0,0 +1,431 @@
...
+/// # Key Separation Design
+///
+/// The two separate keys allow for different cryptographic schemes:
+/// - Message signing can use a different mechanism than Bitcoin transactions
+/// - Currently, only `wallet_pk` is actively used for signatures
+///
+/// # Bitcoin Compatibility
+///
+/// The `wallet_pk` follows [BIP 340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#design)
+/// standards, corresponding to a [`PublicKey`](bitcoin::secp256k1::PublicKey) with even parity
```
barakshani:
```suggestion
/// The `wallet_pk` follows [BIP 340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#design), 
///corresponding to a [`PublicKey`](bitcoin::secp256k1::PublicKey) with even parity
```
or "the [bip340] standard".

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -0,0 +1,431 @@
...
+/// for compatibility with Bitcoin's Taproot and MuSig2 implementations.
+#[derive(
+    Clone, Debug, Eq, PartialEq, Hash, BorshDeserialize, BorshSerialize, Serialize, Deserialize,
+)]
+pub struct OperatorEntry {
+    /// Global operator index.
+    idx: OperatorIdx,
+
+    /// Pubkey used to verify signed messages from the operator.
```
barakshani:
```suggestion
    /// Public key used to verify signed messages from the operator.

File: crates/asm/subprotocols/bridge-v1/src/state/operator.rs [hunk 61]
```diff
@@ -0,0 +1,431 @@
...
+    signing_pk: Buf32,
+
+    /// Wallet pubkey used to compute MuSig2 pubkey from a set of operators.
```
barakshani:
```suggestion
    /// Wallet public key used to compute MuSig2 public key from a set of operators.

File: crates/asm/subprotocols/bridge-v1/src/state/bridge.rs (hunk 47 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/bridge-v1/src/state/bridge.rs]: Forgot to remove this. Thanks for catching it.

File: crates/asm/subprotocols/bridge-v1/src/state/operator.rs [hunk 62]
```diff
@@ -0,0 +1,431 @@
...
+    wallet_pk: Buf32,
+
+    /// Whether this operator is part of the current N/N multisig set.
```
barakshani, line 44: what is current? I thought that each deposit entry has its own operator list. Do we have, in addition, some global "current" active operator list?

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -0,0 +1,431 @@
...
+    ///
+    /// Reference to the signing public key as [`Buf32`].
+    pub fn signing_pk(&self) -> &Buf32 {
+        &self.signing_pk
+    }
+
+    /// Returns the wallet public key for Bitcoin transactions.
+    ///
+    /// This key is used in MuSig2 aggregation for Bitcoin transaction signatures
+    /// and follows BIP 340 standards for Taproot compatibility.
```
barakshani:
```suggestion
    /// and follows BIP 340 for Taproot compatibility.

File: crates/asm/subprotocols/bridge-v1/src/state/bridge.rs (hunk 49 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/bridge-v1/src/state/bridge.rs:205]: I don't it's only a very subtle difference. Withdrawal request makes sense from the perspective of the user. Withdrawal command makes sense from the perspective of the Bridge subprotocol because the command comes from the Checkpoint subprotocol.

File: crates/asm/subprotocols/bridge-v1/src/state/operator.rs (hunk 62 above)
Comment (barakshani) [reply crates/asm/subprotocols/bridge-v1/src/state/operator.rs:44]: ok, this is used to know who is eligible to be assigned a withdrawal?

File: crates/asm/subprotocols/bridge-v1/src/state/bridge.rs (hunk 56 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/bridge-v1/src/state/bridge.rs]: No. This is the front payment transaction (withdrawal fulfillment transaction).

File: crates/asm/subprotocols/bridge-v1/src/state/bridge.rs (hunk 57 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/bridge-v1/src/state/bridge.rs:334]: Yes, we remove it for `process_withdrawal_fulfillment_tx`. This is where we are validating if the withdrawal_fulfillment_tx is valid. 

Moving the validation logic earlier in the code, so it's easier to read.

File: crates/asm/subprotocols/bridge-v1/src/state/bridge.rs (hunk 56 above)
Comment (barakshani) [reply crates/asm/subprotocols/bridge-v1/src/state/bridge.rs]: let's call it `validate_withdrawal_fulfillment` then

File: crates/asm/subprotocols/bridge-v1/src/state/bridge.rs (hunk 57 above)
Comment (barakshani) [reply crates/asm/subprotocols/bridge-v1/src/state/bridge.rs:334]: it's fine, I just thought it's in different context, as in https://github.com/alpenlabs/alpen/pull/956#discussion_r2251104075.

File: crates/asm/subprotocols/bridge-v1/src/state/operator.rs [hunk 64]
```diff
@@ -0,0 +1,431 @@
...
+        self.operators
+            .as_slice()
+            .binary_search_by_key(&idx, |e| e.idx)
+            .ok()
+            .map(|i| &self.operators.as_slice()[i])
+    }
+
+    /// Returns indices of operators in the current N/N multisig.
+    ///
+    /// Only returns indices of operators where `is_in_current_multisig` is `true`.
```
barakshani, line 236: double checking: if `operator_1` is in N/N multisig but not in current operator list, it will not be included here?

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
prajwolrg: I'm not sure if I understand fully what you mean here. But I've tried to accommodate this.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
prajwolrg: yes.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
prajwolrg: yes. although we do not support adding/removing operators, the code is here to allow for that.

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
barakshani: The question was why we need two argument for the bitcoin block, which I believe you now consolidated.

crates/asm/subprotocols/bridge-v1/src/subprotocol.rs
```diff
@@ -0,0 +1,160 @@
...
+            Ok(reassigned_deposits) => {
+                info!(
+                    count = reassigned_deposits.len(),
+                    deposits = ?reassigned_deposits,
+                    "Successfully reassigned expired assignments"
+                );
+            }
+            Err(e) => {
+                // PANIC: Failure to reassign expired assignments indicates a violation of the
+                // bridge's fundamental 1/N honesty assumption. This means no honest operators
```
barakshani:
```suggestion
                // bridge's fundamental 1/N honesty assumption. This means no operators

File: crates/asm/subprotocols/bridge-v1/src/state/bridge.rs [hunk 66]
```diff
@@ -0,0 +1,888 @@
...
+        // Verify the deposit amount matches the bridge's expected amount
+        if info.amt.to_sat() != self.denomination.to_sat() {
+            return Err(DepositValidationError::MismatchDepositAmount(Mismatch {
+                expected: self.denomination.to_sat(),
+                got: info.amt.to_sat(),
+            }));
+        }
+
+        // Validate the DRT spending signature against the aggregated operator key
+        validate_drt_spending_signature(
```
barakshani, line 144: I believe should be the first check. If it fails we should not return an error, as it's not the bridge operators that processed the DRT. Any subsequent failure should return an error.

crates/asm/subprotocols/bridge-v1/src/errors.rs
barakshani: I'd try to find a softer language, but up to you

crates/asm/subprotocols/bridge-v1/src/errors.rs
barakshani: @Rajil1213 @storopoli I'd be happy to hear your opinion on this. Doesn't have to be on this particular case, but in general.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
barakshani: Actually I believe we had some discussion with product on this (outside of the PRD), and this is aligned.

barakshani (approved):
I checked overall compatibility with the spec. Nice work!

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
prajwolrg: I don't have a strong opinion on the order of the checks. Understanding your rationale will be helpful for me here.

Any of the validation failure should return an appropriate error. It's how we handle the error that we need to be careful about. Returning early on any error, I understand is the idiomatic way to handle things.

barakshani: the rationale is not really about the logical order, it's related to the other topic of what should be an error and what not.
So I think that if we are sure the bridge sent the deposit tx but something goes wrong (say same index or incorrect amount), we should return an error. If it wasn't the bridge that sent the tx, I'm just not sure error is the right way to exit.
That's why I wanted the first check to be "is this deposit sent by the bridge operators or not?".

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
Rajil1213: > ...hence we cannot conclude this is an improvement or even desired.

This is a slightly different context. This _is_ an improvement in the sense that it reimplements the current logic in a slightly better way. In the current implementation, it is completely random, so it is possible that we end up with the assignments being in the order: `0, 1, 0, 1` (i.e., we have a cycle of 2  even if there are 3 operators). In this new approach, we'll not get a cycle that is less than 3 i.e, we will not reassign to the same operator twice until all operators have been used. I believe this is a better approach.

crates/asm/subprotocols/bridge-v1/src/errors.rs
Rajil1213: It depends on the kind of interface we are developing.

If the interface accepts a stricter structure and not just any transaction, then it makes sense for this to error in certain cases for example, if someone sent a withdrawal/deposit with the right metadata in the OP_RETURN but still failed some validation. What we do with that error is a different matter altogether.

On the other hand, if the interface accepts any arbitrary transaction, then it should only return a boolean value indicating whether it satisfies the predicate we are interested in.

A third case is where we accept an arbitrary transaction and then if valid, construct a new structure that is isomorphic to the original transaction. If the transaction is invalid, we return `None` instead. This is the approach we are taking in the bridge.

If I'm not mistaken, in this particular case, this error is emitted by the bridge subprotocol machinery where I think it's fine to error out since it is expected that the transaction is valid and so we should error out if the transaction fails to meet those expectations.

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
Rajil1213: I think this is following the order of "simpler test first". Checking the validity of the signature is more complicated in terms of CPU cycles (real-world impact should be minimal). So if the transaction's output or metadata does not even match the expected format/values, why go to the trouble of performing that computation?

crates/asm/subprotocols/bridge-v1/src/errors.rs
storopoli: I agree with @Rajil1213.

For the API, it makes sense to emit an error, which does not mean that we'll eventually crash and burn.

I think we should have the current approach where we error in general for anything protocol related and we decide in the consumer side of the code how the program behaves when encountering particular errors.

delbonis (approved):

PR #958, opened 2025-07-29
[DO NOT MERGE]: Str 1602 typed sled

bewakes:
**Update**: _The library introduced in this PR needs to go to a new repository so this should not be merged. It is here for reviews and to use the library in this repository._

## Description

This PR introduces a new crate `typed-sled` which is basically a typed wrapper to `sled` similar to `rockbound` that we have been using.
The ultimate goal is to have this crate as a standalone library. For now it is in the repo to be able to iterate fast on porting the rocksdb impls to sled impls.

For more details on the task see https://alpenlabs.atlassian.net/browse/STR-1602.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/typed-sled/examples/basic_usage.rs
+++ b/crates/typed-sled/examples/basic_usage.rs
@@ -1,9 +1,7 @@
 use borsh::{BorshDeserialize, BorshSerialize};
 use dashmap as _;
 use thiserror as _;
-use typed_sled::{
-    CodecError, KeyCodec, Schema, SledDb, SledTree, TreeName, ValueCodec, error::Result,
-};
+use typed_sled::{CodecError, Schema, SledDb, SledTree, TreeName, ValueCodec, error::Result};
 
 #[derive(BorshSerialize, BorshDeserialize, Debug, Clone)]
 struct User {
@@ -14,25 +12,6 @@ struct User {
 
 #[derive(Debug)]
 struct UserSchema;
-
-impl KeyCodec<UserSchema> for u32 {
-    fn encode_key(&self) -> typed_sled::CodecResult<Vec<u8>> {
-        Ok(self.to_be_bytes().to_vec())
-    }
-
-    fn decode_key(buf: &[u8]) -> typed_sled::CodecResult<Self> {
-        if buf.len() != 4 {
-            return Err(CodecError::InvalidKeyLength {
-                schema: UserSchema::TREE_NAME.0,
-                expected: 4,
-                actual: buf.len(),
-            });
-        }
-        let mut bytes = [0; 4];
-        bytes.copy_from_slice(buf);
-        Ok(u32::from_be_bytes(bytes))
-    }
-}
 
 impl ValueCodec<UserSchema> for User {
     fn encode_value(&self) -> typed_sled::CodecResult<Vec<u8>> {
--- a/crates/typed-sled/src/codec.rs
+++ b/crates/typed-sled/src/codec.rs
@@ -52,3 +52,29 @@ pub trait ValueCodec<S: Schema>: Sized {
     /// Decodes the value from bytes.
     fn decode_value(buf: &[u8]) -> CodecResult<Self>;
 }
+
+macro_rules! derive_key_codec_for_integers {
+    ($($int:ty), *) => {
+        $(impl<T: Schema> KeyCodec<T> for $int {
+            fn encode_key(&self) -> CodecResult<Vec<u8>> {
+                Ok(self.to_be_bytes().into())
+            }
+
+            fn decode_key(buf: &[u8]) -> CodecResult<Self> {
+                const SIZE: usize = std::mem::size_of::<$int>();
+                if buf.len() != SIZE {
+                    return Err(CodecError::InvalidKeyLength {
+                        schema: T::TREE_NAME.0,
+                        expected: SIZE,
+                        actual: buf.len(),
+                    });
+                }
+                let mut bytes = [0u8; SIZE];
+                bytes.copy_from_slice(buf);
+                Ok(<$int>::from_be_bytes(bytes))
+            }
+        })*
+    };
+}
+
+derive_key_codec_for_integers!(u8, u16, u32, u64, u128, i8, i16, i32, i64, i128);
--- a/crates/typed-sled/src/lib.rs
+++ b/crates/typed-sled/src/lib.rs
@@ -17,7 +17,7 @@
 //!
 //! ```rust,no_run
 //! use borsh::{BorshDeserialize, BorshSerialize};
-//! use typed_sled::{CodecError, KeyCodec, Schema, SledDb, TreeName, ValueCodec, error::Result};
+//! use typed_sled::{CodecError, Schema, SledDb, TreeName, ValueCodec, error::Result};
 //!
 //! #[derive(BorshSerialize, BorshDeserialize, Debug)]
 //! struct User {
@@ -32,24 +32,6 @@
 //!     const TREE_NAME: TreeName = TreeName("users");
 //!     type Key = u32;
 //!     type Value = User;
-//! }
-//!
-//! impl KeyCodec<UserSchema> for u32 {
-//!     fn encode_key(&self) -> typed_sled::CodecResult<Vec<u8>> {
-//!         Ok(self.to_be_bytes().to_vec())
-//!     }
-//!     fn decode_key(buf: &[u8]) -> typed_sled::CodecResult<Self> {
-//!         if buf.len() != 4 {
-//!             return Err(CodecError::InvalidKeyLength {
-//!                 schema: UserSchema::TREE_NAME.0,
-//!                 expected: 4,
-//!                 actual: buf.len(),
-//!             });
-//!         }
-//!         let mut bytes = [0; 4];
-//!         bytes.copy_from_slice(buf);
-//!         Ok(u32::from_be_bytes(bytes))
-//!     }
 //! }
 //!
 //! impl ValueCodec<UserSchema> for User {
--- a/crates/typed-sled/src/test_utils.rs
+++ b/crates/typed-sled/src/test_utils.rs
@@ -5,18 +5,18 @@
 
 use borsh::{BorshDeserialize, BorshSerialize};
 
-use crate::{CodecError, CodecResult, KeyCodec, Schema, SledDb, TreeName, ValueCodec};
+use crate::{CodecError, CodecResult, Schema, SledDb, TreeName, ValueCodec};
 
 /// Common test value type used across all tests.
 #[derive(BorshSerialize, BorshDeserialize, Debug, Clone, PartialEq)]
-pub struct TestValue {
+pub(crate) struct TestValue {
     pub id: u32,
     pub name: String,
 }
 
 impl TestValue {
     /// Creates a test value for "Alice"
-    pub fn alice() -> Self {
+    pub(crate) fn alice() -> Self {
         Self {
             id: 1,
             name: "Alice".to_string(),
@@ -24,7 +24,7 @@ impl TestValue {
     }
 
     /// Creates a test value for "Bob"
-    pub fn bob() -> Self {
+    pub(crate) fn bob() -> Self {
         Self {
             id: 2,
             name: "Bob".to_string(),
@@ -32,7 +32,7 @@ impl TestValue {
     }
 
     /// Creates a test value for "Charlie"
-    pub fn charlie() -> Self {
+    pub(crate) fn charlie() -> Self {
         Self {
             id: 3,
             name: "Charlie".to_string(),
@@ -40,7 +40,7 @@ impl TestValue {
     }
 
     /// Creates a test value with a generated name based on id
-    pub fn new_with_name(id: u32) -> Self {
+    pub(crate) fn new_with_name(id: u32) -> Self {
         Self {
             id,
             name: format!("Item {}", id),
@@ -48,7 +48,7 @@ impl TestValue {
     }
 
     /// Creates a test value with custom id and name
-    pub fn new(id: u32, name: &str) -> Self {
+    pub(crate) fn new(id: u32, name: &str) -> Self {
         Self {
             id,
             name: name.to_string(),
@@ -84,30 +84,6 @@ impl Schema for TestSchema3 {
     const TREE_NAME: TreeName = TreeName("test3");
     type Key = u32;
     type Value = TestValue;
-}
-
-// Generic codec implementations that work for any schema with Key=u32, Value=TestValue
-
-impl<S> KeyCodec<S> for u32
-where
-    S: Schema<Key = u32, Value = TestValue>,
-{
-    fn encode_key(&self) -> CodecResult<Vec<u8>> {
-        Ok(self.to_be_bytes().to_vec())
-    }
-
-    fn decode_key(buf: &[u8]) -> CodecResult<Self> {
-        if buf.len() != 4 {
-            return Err(CodecError::InvalidKeyLength {
-                schema: S::TREE_NAME.0,
-                expected: 4,
-                actual: buf.len(),
-            });
-        }
-        let mut bytes = [0; 4];
-        bytes.copy_from_slice(buf);
-        Ok(u32::from_be_bytes(bytes))
-    }
 }
 
 impl<S> ValueCodec<S> for TestValue


delbonis:
This is very interesting, but do you think it would be better in a separate repo?

delbonis:
Something to consider here is that sled trees are not 1:1 with RocksDB column families.  For example, you can only do transactions within a tree, not across them.  We *do* have scenarios where we have two kinds of data structures we want to manipulate within the context of a transaction (like for example, a new block and the list of blkids at a block height).  With the current design that's a bit tricky, since it more closely couples the typed schema with the tree, as in `struct SledTree<S: Schema>`.

I would suggest introducing a `const PREFIX: u8` associated const on the schema trait, which is what's suggested by sled developers to distinguish different types of entries within a tree.  Then you can shift down to simply have a `impl<S: Schema> SledTree { ... }` without making the `SledTree` type be parametric over the schema.  And then you'd do the same thing for the batch types.

Also it seems like you're manually expanding out what the thiserror macros do in the codec module.  Is there some reason for that?

bewakes:
> Also it seems like you're manually expanding out what the thiserror macros do in the codec module. Is there some reason for that?

Not really actually. This is something I missed to refactor. Changed to use thiserror now.

bewakes:
One interesting setback though, in sled, there's atomic `compare_and_swap` for trees. But that cannot be used inside transaction because both handle atomicity differently.

Ran into this while trying to convert this rockbound impl for broadcast db:
```rust
impl L1BroadcastDatabase for L1BroadcastDb {
    fn put_tx_entry(&self, txid: Buf32, txentry: L1TxEntry) -> DbResult<Option<u64>> {
        self.db
            .with_optimistic_txn(
                TransactionRetry::Count(self.ops.retry_count),
                |txn| -> Result<Option<u64>, anyhow::Error> {
                    if txn.get::<BcastL1TxSchema>(&txid)?.is_none() {
                        let idx = get_next_id::<BcastL1TxIdSchema, DB>(txn)?;
                        txn.put::<BcastL1TxIdSchema>(&idx, &txid)?;
                        txn.put::<BcastL1TxSchema>(&txid, &txentry)?;
                        Ok(Some(idx))
                    } else {
                        txn.put::<BcastL1TxSchema>(&txid, &txentry)?;
                        Ok(None)
                    }
                },
            )
            .map_err(|e| DbError::TransactionError(e.to_string()))
    }
    ...
}
```
to typed- sled impl:
```rust
impl L1BroadcastDatabase for L1BroadcastDBSled {
    fn put_tx_entry(&self, txid: Buf32, txentry: L1TxEntry) -> DbResult<Option<u64>> {
        let next = self.get_next_idx()?;
        if self.tx_tree.contains_key(&txid)? {
            self.tx_tree.insert(&txid, &txentry)?;
            return Ok(None);
        }
        // We expect there is no entry for next idx.
        // TODO: we might need to be able to retry here.
        self.tx_id_tree.compare_and_swap(next, None, Some(txid))?;
        Ok(Some(next))
    }
    ...
}
```
I think this translation works in this particular case, yet to discover other cases.

cc: @sapinb

crates/typed-sled/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "typed-sled"
+version = "0.1.0"
+edition = "2024"
+
```
storopoli, line 5: Everybody forgets this.
I am thinking of adding a CI check triggering by any changes or creation of `Cargo.toml`s and if it does not contain the workspace lints it fails.

```suggestion
edition = "2024"

[workspace]
lints = true
```

storopoli (changes requested):

bewakes:
Okay, this should be good to go now. See #979 for the library usage.

crates/typed-sled/examples/basic_usage.rs
```diff
@@ -0,0 +1,108 @@
+use std::sync::Arc;
+
+use borsh::{BorshDeserialize, BorshSerialize};
+use dashmap as _;
+use thiserror as _;
```
storopoli, line 3: Why do you need this in the example?

functional-tests/entry.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 from gevent import monkey
 
-# This is important for locust to work with flexitest.
+# This important for locust to work with flexitest.
```
storopoli: I don't think you want to do this change

crates/typed-sled/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+[lints]
+workspace = true
+
+[dependencies]
+dashmap = "6.1.0"
+sled = "0.34.7"
```
storopoli, line 11: Can you consolidate this in the root `Cargo.toml` and then update this `Cargo.toml` and the one in `bin/alpen-cli` to use the `workspace` as well: https://github.com/alpenlabs/alpen/blob/fcae36d817f7195ae34ad1e616c507f793e46cdd/bin/alpen-cli/Cargo.toml#L45

crates/typed-sled/.gitignore
```diff
@@ -0,0 +1 @@
+/target
```
storopoli, line 1: Why do we need this? Can't we update the https://github.com/alpenlabs/alpen/blob/fcae36d817f7195ae34ad1e616c507f793e46cdd/.gitignore#L62

to be:

```
**/target/
``

File: crates/typed-sled/src/lib.rs [hunk 6]
```diff
@@ -0,0 +1,105 @@
+//! # typed-sled
+//!
+//! A type-safe wrapper around the sled embedded database.
+//!
+//! This library provides a schema-based approach to working with sled,
```
storopoli:
```suggestion
//! This library provides a schema-based approach to working with [`sled`],
```

storopoli: Amazing library-level docs. Makes you wanna frame it 😃

crates/typed-sled/src/lib.rs
```diff
@@ -0,0 +1,105 @@
+//! # typed-sled
+//!
+//! A type-safe wrapper around the sled embedded database.
```
storopoli:
```suggestion
//! A type-safe wrapper around the [`sled`] embedded database.
```

crates/typed-sled/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+//! ## Features
+//!
+//! - **Type Safety**: Schema-based table definitions with associated key/value types
+//! - **Serialization**: Flexible codec system for efficient binary encoding
+//! - **Transactions**: Multi-table atomic operations
+//! - **Error Handling**: Comprehensive error types with proper error chaining
+//!
+//! ## Example
+//!
+//! ```rust,no_run
```
storopoli, line 18: Why `no_run`?
This might get obsolete soon and there will be no errors or warnings to tell you so.
What do we need to make it not `no_run`?

storopoli (changes requested):
This is really good.
Some small requested changes.
I'd like also someone who has a background knowledge about our current `{alpen,strata}-db*` codebases to take a look and review the specifics of the implementation.

crates/typed-sled/src/codec.rs
```diff
@@ -0,0 +1,54 @@
...
+    /// Encodes the key into bytes.
+    fn encode_key(&self) -> CodecResult<Vec<u8>>;
+    /// Decodes the key from bytes.
+    fn decode_key(buf: &[u8]) -> CodecResult<Self>;
+}
+
+/// Trait for encoding and decoding values for a specific schema.
+pub trait ValueCodec<S: Schema>: Sized {
+    /// Encodes the value into bytes.
+    fn encode_value(&self) -> CodecResult<Vec<u8>>;
```
delbonis, line 51: I am not necessarily suggesting we change this, but why are we using two different traits for this?  I realize that's what rockbound does, but is there a reason to directly copy this pattern?  I can see one reason being that we don't want to automatically impl borsh on keys because it uses little endian integers and that would break iteration.

Also, this should be encoding to a sled `IVec` to avoid allocations for small keys (like integers).

crates/typed-sled/src/schema.rs
```diff
@@ -0,0 +1,28 @@
+use std::fmt::Debug;
+
+use crate::codec::{KeyCodec, ValueCodec};
+
+/// A type-safe wrapper for tree names.
+#[derive(Debug, Hash, Eq, PartialEq)]
+pub struct TreeName(pub &'static str);
```
delbonis, line 7: *Should* this be pub?

crates/typed-sled/src/tree.rs
```diff
@@ -0,0 +1,582 @@
...
+
+/// Decodes a raw key-value pair into typed schema types.
+fn decode_pair<S: Schema>((k, v): (IVec, IVec)) -> Result<(S::Key, S::Value)> {
+    let key = S::Key::decode_key(&k)?;
+    let value = S::Value::decode_value(&v)?;
+    Ok((key, value))
+}
+
+/// Converts a typed key bound to a raw byte bound.
+fn key_bound<S: Schema>(k: Bound<&S::Key>) -> Result<Bound<Vec<u8>>> {
```
delbonis, line 18: I *think* this should also be an `IVec`.

crates/typed-sled/examples/basic_usage.rs
```diff
@@ -0,0 +1,108 @@
...
+struct User {
+    id: u32,
+    name: String,
+    email: String,
+}
+
+#[derive(Debug)]
+struct UserSchema;
+
+impl KeyCodec<UserSchema> for u32 {
```
delbonis: Might as well define a macro to impl this for arbitrary integer types and then impl it on all of them.

Also maybe borsh/serde shim macros for `ValueCodec`?

crates/typed-sled/src/codec.rs
```diff
@@ -0,0 +1,54 @@
...
+        #[source]
+        source: Box<dyn std::error::Error>,
+    },
+    /// I/O error during codec operations.
+    #[error("I/O error")]
+    IO(#[from] std::io::Error),
+
+    /// Other
+    #[error("Other error: {0}")]
+    Other(String),
```
delbonis: For consistency:

* `InvalidKeyLength`: `invalid key length in schema '{schema}' (expected {expected}, actual {actual})`
* `SerializationFailed`/`DeserializationFailed`: `failed to {,de}serialize schema '{schema}' value: {source}` (not sure if you can format the source error like that)
* `Io`: `io: {0)`
* `Other`: `{0}`

crates/typed-sled/src/db.rs
```diff
@@ -0,0 +1,245 @@
...
+    tree::SledTree,
+};
+
+/// A type-safe wrapper around sled database with schema-based tree management.
+#[derive(Debug)]
+pub struct SledDb {
+    /// Mapping of treenames to sled tree.
+    inner_trees: DashMap<TreeName, Arc<Tree>>,
+    /// The actual sled db.
+    inner_db: Arc<Db>,
```
delbonis: Couple of comments:

* Calling it `SledDb` is a bit weird when the base type isn't `SledDb`, but we're also adding new behavior.  The [kv crate](https://docs.rs/kv/latest/kv/) called it `Store`.  (It's fine we're not using this crate fwiw, it seems unmaintained and doesn't give us everything we'd want from it.)
* `Db` already implements `Clone`, no need to arc it again.
* Why are we putting it in an arc anyways if we're not implementing `Clone` ourselves?
* Do we need to store a table of trees here when we can obtain new instances by calling `open_tree` again?  This would alleviate the desire to import dashmap.

crates/typed-sled/src/error.rs
```diff
@@ -0,0 +1,242 @@
...
+    #[error("Database error: {0}")]
+    SledError(#[from] SledError),
+
+    /// Sled transaction error
+    #[error("Transaction error: {0}")]
+    TransactionError(#[from] UnabortableTransactionError),
+
+    /// CAS error
+    #[error("CAS error: {0}")]
+    CASError(#[from] CompareAndSwapError),
```
delbonis: Similar to the codec error type, we can make this more terse *and* be more specific at the same time:

* `codec: {0}`
* `sled: {0}`
* `sled tx: {0}`
* `sled cas: {0}`

It's almost never necessary to say there's was an error in an error message, it's implied by it being an error message.

storopoli:
> Again would like to raise that this should be its own repo to alleviate project management concerns.

Sorry I have not seen the review history.
Yes! I agree with @delbonis. What's stopping us into moving this to its own repo?
Use the https://github.com/alpenlabs/rust-template-workspace for workspace repos or https://github.com/alpenlabs/rust-template for standalone repos and you'll be 99% good to go!

crates/typed-sled/examples/basic_usage.rs
bewakes: I needed to do this to make clippy happy. For some reason, `cargo clippy` was complaining that these libraries in `Cargo.toml` were unused in the example.

functional-tests/entry.py
bewakes: Some vim typo in normal mode :D

crates/typed-sled/.gitignore
bewakes: Yes, but since this is supposed to go in the separate repo, I wanted to make this crate as independent to workspace as possible.

crates/typed-sled/src/lib.rs
bewakes: The reason it has `no_run` is because there's database file being opened in the `main()` function in the example.  Basically a side effect. Are you saying `no_run` might get obsolete soon or the example code?

I am not sure if we should try to make it not `no_run` but to make it so, we need to get rid of opening the database file.

crates/typed-sled/src/schema.rs
bewakes: Yes, it will be used in the macro to generate the Schemas in the `db-store-sled` crate.

crates/typed-sled/src/codec.rs
bewakes: Agree. thanks.

crates/typed-sled/src/error.rs
bewakes: totally :D

crates/typed-sled/examples/basic_usage.rs
bewakes: I think that should be done in the crate using this library. Doing so would also make this crate not having borsh/serde as deps.

crates/typed-sled/src/db.rs
bewakes: - Do you think `SledStore` would be a better name? It would be ideal to have almost identical interfaces to sled's `Db` and call this `SledDb`, but I just didn't want to put those extra efforts for what we need it for now. Definitely can add identical interfaces in future.
- Yes right, the `Db` and `Tree` clones in sled are already cheaper so no need to wrap them in Arc.
- Same as above.
- We can get new instances by calling `open_tree` but that's one non-trivial function invocation which I want to avoid. In the definition of `open_tree` there's a couple of rw locks happening which I think is better to avoid. I agree this comes up with `dashmap` dependency but I think it is worth it.

bewakes: I've removed the redundant `Arc` wrapping.

crates/typed-sled/src/codec.rs
bewakes: Yes, the main reason I think this patterns is good is that it allows us to have different encodings of same type within a schema.

I am trying to avoid `IVec` as much as possible to let the library users work on simpler types. Also, the conversion to `IVec` from `Vec<u8>` is pretty cheap and for the other way around we can work with slices cheaply and only copy when necessary.

crates/typed-sled/src/tree.rs
bewakes: Here also, same reasoning as above.

bewakes:
> What's stopping us into moving this to its own repo?

@storopoli , the only reason I am having this crate in this repo is to have faster iteration while working on the `db-store-sled` crate that uses this crate. Now that the `db-store-sled` crate is almost complete, which means this library has the necessary features. Also, I could get better reviews on the crate.
Now I will create a new repo and  close this PR.

bewakes:
Closing this as I've now created a separate repo for this crate: https://github.com/alpenlabs/typed-sled

PR #959, opened 2025-07-29
fix: gas price handling when from_rpc is true

mdqst:
## Description

fixed the backwards logic for `from_rpc`.
now when `from_rpc` is true, the script actually fetches the current gas price from the network instead of using a hardcoded 1 gwei.
this should prevent transactions from getting stuck due to wrong gas pricing.

### Type of Change

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
* [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ] Documentation update
* [ ] Refactor
* [ ] New or updated tests
* [ ] Dependency Update

## Notes to Reviewers

just a quick fix on the gas price logic - nothing else changed. should be smooth to review.

## Checklist

* [x] I have performed a self-review of my code.
* [x] I have commented my code where necessary.
* [ ] I have updated the documentation if needed.
* [x] My changes do not introduce new warnings.
* [ ] I have added (where necessary) tests that prove my changes are effective or that my feature works.
* [ ] New and existing tests pass with my changes.

### DIFF
--- a/functional-tests/utils/transaction.py
+++ b/functional-tests/utils/transaction.py
@@ -69,7 +69,7 @@ class TransactionBuilder(_TransactionFaucet):
             tx.setdefault("accessList", [{"address": tx["to"], "storageKeys": []}])
 
             tx.setdefault(
-                "gasPrice", self.w3.eth.gas_price if from_rpc else self.w3.to_wei("1", "gwei")
+                "gasPrice", self.w3.eth.gas_price if not from_rpc else self.w3.to_wei("1", "gwei"))
             )
 
         elif tx_type == TransactionType.EIP1559:


evgenyzdanovich:
@mdqst I don't think the fix is accurate. 

Currently (at head), the logic is correct: `if from_rpc then we query w3.eth.gas_price` which is actually **an eth_gasPrice RPC that returns the actual gas_price**. 
The fix you propose does the opposite: if not from_rpc

evgenyzdanovich (changes requested):
@mdqst I don't think the fix is accurate. 

Currently (at head), the logic is correct: `if from_rpc then we query w3.eth.gas_price` which is actually **an eth_gasPrice RPC using web3 lib that returns the actual gas_price**. By default, gas_price is one gwei that's usually enough for testing on a local setup.

The fix you propose does the opposite: if not from_rpc, we fetch gas_price from eth_gasPrice, which seems counterintuitive

PR #960, opened 2025-07-30
fix(cli): correct recovery pk in opreturn

Zk2u:
## Description

Corrects an issue in the CLI where an incorrect public key was being supplied into the opreturn.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -78,20 +71,14 @@ pub async fn deposit(
         alpen_address.to_string().cyan(),
     );
 
-    println!(
-        "Recovery address: {}",
-        recovery_address.to_string().yellow()
-    );
-
-    let (secret_key, _) = SECP256K1.generate_keypair(&mut OsRng);
+    let (secret_key, public_key) = SECP256K1.generate_keypair(&mut OsRng);
```
Rajil1213: Why not log the recovery public key here?

Rajil1213 (commented):
Thanks for the quick fix.

The changes look good but has this been tested against staging/prod?

bin/alpen-cli/src/cmd/deposit.rs
Zk2u: It's random so not much point logging it TBH...

Rajil1213: I mean so was the recovery address which _was_ logged.

storopoli: Yes, but we save that "random" in the `DescriptorRecovery` database and it is REALLY helpful to have it logged since this PR fixes something that was wrong with the recovery public key.

bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -132,7 +119,7 @@ pub async fn deposit(
     // <recovery_address_pk>
     // <alpen_address>
     let magic_bytes = settings.magic_bytes.as_bytes();
-    let recovery_address_pk_bytes = recovery_address_pk.serialize();
+    let recovery_address_pk_bytes = public_key.serialize();
```
storopoli: NACK!
Before we were serializing an `XOnlyPublicKey` without the parity byte, `{02,03}`.
Now you're serializing a `PublicKey` which has the parity byte.
Just call the [`XOnlyPublicKey::from(public_key)`](https://docs.rs/bitcoin/latest/bitcoin/struct.XOnlyPublicKey.html#impl-From%3CPublicKey%3E-for-XOnlyPublicKey)

storopoli (changes requested):
There's a footgun here.

Also why the hell are we using `bdk_wallet` if we are deferring key generation to `secp256k1` and `OsRng` and saving it to a custom made database `DescriptorRecovery`. We are only using it to generate the transaction. Might as well use the new [`bdk_tx`](https://lib.rs/crates/bdk_tx) crate.

crates/primitives/src/crypto.rs
```diff
@@ -79,6 +79,13 @@ impl From<SecretKey> for EvenSecretKey {
     }
 }
 
+pub fn even_kp((sk, pk): (SecretKey, PublicKey)) -> (SecretKey, PublicKey) {
+    match (sk, pk) {
+        (sk, pk) if pk.x_only_public_key().1 == Parity::Odd => (sk.negate(), pk.negate(SECP256K1)),
+        (sk, pk) => (sk, pk),
+    }
+}
+
```
storopoli: Why not do a `From<Keypair>` instead?

bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -211,15 +207,13 @@ fn bridge_in_descriptor(
...
+) -> DescriptorTemplateOut {
+    bdk_wallet::descriptor!(
         tr(bridge_pubkey,
             and_v(v:pk(private_key),older(recover_delay))
         )
     )
-    .expect("valid descriptor");
-
-    Ok(desc)
+    .expect("valid descriptor")
```
storopoli, line 216: This will panic. Or do we have safeguards when we call this function?
Shouldn't we do a proper error here?

Zk2u: This will only panic if the descriptor is invalid. This descriptor we know is always valid

crates/primitives/src/crypto.rs
Zk2u: Because I don't see an ergonomics boost from adding another newtype

bin/alpen-cli/src/cmd/deposit.rs
storopoli: Ok solving this.

crates/primitives/src/crypto.rs
storopoli: We already have new types for even secret keys: https://github.com/alpenlabs/alpen/blob/00737933fb000a37f61801afc4461020fa29df4a/crates/primitives/src/crypto.rs#L55-L81

Why not `EvenKeypair`?
Anyways, at least add docstrings since it is a new public function in `primitives` and `crypto` related?

storopoli: Other people might find useful since `EvenSecretKey` is used in `alpen` at 
CleanShot 2025-07-30 at 10 58 48@2x

and in `strata-bridge` in 
CleanShot 2025-07-30 at 10 59 16@2x

Zk2u: I'll add an EvenPublicKey and do some additions, but I want to keep this fn. Will doc it too.

storopoli: That works!

PR #970, opened 2025-08-02
chore: fix misspell in error message

gap-editor:
correct form is "does" because L1Header is singular


MdTeach:
Thanks—I’ll consolidate these

PR #974, opened 2025-08-04
fix: correct retry logic to execute operation exactly max_retries times

GarmashAlex:
Fix logical error in retry_with_backoff function where operation was executed
max_retries + 1 times instead of max_retries times.

Changes:
- Change loop range from 0..=max_retries to 0..max_retries
- Update retry condition from attempt < max_retries to attempt < max_retries - 1
- Replace unreachable!() with proper final attempt handling
- Update tests to match corrected logic
- Clarify documentation to specify total executions will be max_retries

This ensures that when max_retries = 2, the operation is executed exactly
2 times instead of 3 times, making the behavior consistent with the
parameter name and documentation.


storopoli:
This is a no-op:
CleanShot 2025-08-04 at 11 14 41@2x

Thanks, but we're closing this.

PR #977, opened 2025-08-04
STR-1635 Service framework

delbonis:
## Description

This PR adds a new crate for a service framework library as laid out in the ticket.  This does not include the integration to adapt existing services to use the new framework.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Still working on this, want to be able to directly expose command worker services.

## Checklist


delbonis:
Something I'm thinking about here is adjusting the `Service` type to not require specifying the input, just the message type.  This would require changing a few minor things but would improve the ergonomics especially for testing.

crates/service/src/adapters.rs
```diff
@@ -0,0 +1,361 @@
+#![allow(missing_debug_implementations)] // none of these make sense to be debug
```
storopoli:
```suggestion
#![allow(missing_debug_implementations, reason = "one of these make sense to be debug")]
```

crates/service/src/adapters.rs
```diff
@@ -0,0 +1,361 @@
...
+#![allow(clippy::manual_async_fn)] // clippy is just wrong about this, the types don't work
```
storopoli:
```suggestion
#![allow(clippy::manual_async_fn, reason = "clippy is just wrong about this, the types don't work")]
```

crates/service/src/adapters.rs
```diff
@@ -0,0 +1,361 @@
...
+
+use std::{any::Any, collections::*, fmt::Debug, future::Future, sync::Arc};
+
+use futures::stream::{Stream, StreamExt};
+use tokio::sync::{mpsc, Mutex};
+
+use super::*;
```
storopoli: I don't like this pattern, it is error-prone by shadowing. Can we do the regular `use crate::foo`?

crates/service/src/types.rs
```diff
@@ -0,0 +1,89 @@
...
+
+/// Response from handling an input.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum Response {
+    /// Normal case, should continue.
+    Continue,
+
+    /// Service should exit early.
+    ShouldExit,
+}
```
storopoli, line 15: We should have a `Retry` here to handle this TODO: https://github.com/alpenlabs/alpen/blob/d33444305b753a5606a557969f3a1b252f1aac8e/crates/service/src/async_worker.rs#L44-L51

Check the [`retry` module in the `algebra` crate in `strata-bridge`](https://github.com/alpenlabs/strata-bridge/blob/main/crates/algebra/src/retry.rs).
It is a FP approach but you can get valuable insights.
We are thinking of refactoring it to a standalone repo/crate.

crates/service/src/async_worker.rs
```diff
@@ -0,0 +1,64 @@
...
+            Err(e) => {
+                // TODO support optional retry
+                error!(?input, %e, "failed to process message");
+                break;
+            }
+        };
+
+        // Update the status.
+        let status = S::get_status(state);
+        let _ = status_tx.send(status);
```
storopoli: Status updates after every message could be expensive for high-frequency services. We should consider batching or rate-limiting updates. Can be incorporated in the TODO with a more specific action rather than "more sophisticated".

crates/service/src/async_worker.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    futures::select! {
+        _ = exit_fut => (),
+        res = wkr_fut => res?,
+    };
+
+    Ok(())
+}
+
+async fn worker_task_inner<S: AsyncService, I>(
```
storopoli, line 45: We don't have a graceful shutdown for the `async_worker` like we do for the `sync_worker`: https://github.com/alpenlabs/alpen/blob/d33444305b753a5606a557969f3a1b252f1aac8e/crates/service/src/sync_worker.rs#L40-L44

This can be something straight out of `strata-tasks`'s `ShutdownGuard`: https://github.com/alpenlabs/alpen/blob/d33444305b753a5606a557969f3a1b252f1aac8e/crates/tasks/src/shutdown.rs#L54-L80
or adapted if needed.

crates/service/src/types.rs
delbonis: Actually retryability would be dependent on error conditions, so it wouldn't be a `Retry` variant on this.  If we return `Ok` then we treat that input as having been processed successfully.  But yes that crate looks interesting.

I wasn't quite sure what to do with this `Response` concept.  I knew I needed some way to communicate this concept but it feels a little stubby.

crates/service/src/async_worker.rs
delbonis: The services this is planned to be used for already do per-input status updates.  The listener uses watch channels so rapid inputs will just be discarded, we won't block, and the consumers are expected to be okay with theoretically missing updates (only caring about the latest value).

crates/service/src/async_worker.rs
delbonis: I'm not sure how that graceful shutdown is supposed to work in an async context, since we'd be dropping the future.  I can add a cleanup call to the async service crate.  I also wanted to add init calls anyways.

crates/service/src/types.rs
delbonis: Actually, I'm going to hold off on implementing retries in this PR in the interest of keeping it conservative.  We don't have a strong need for natively supporting it right now, so I'm just going to move on and create a ticket for it for later.

crates/service/src/async_worker.rs
delbonis: Added a `before_shutdown` handler so that we can perform some logic in async tasks on shutdown.

storopoli: That's good. Thanks

PR #979, opened 2025-08-06
Str 1233 sled impl

bewakes:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


crates/db-store-sled/src/chain_state/schemas.rs
```diff
@@ -0,0 +1,31 @@
...
+define_table_with_seek_key_codec!(
+    /// Table to store state instance data.
+    (StateInstanceSchema) StateInstanceId => StateInstanceEntry
+);
+// TODO: copied over from db-store-rocksdb. needs to go to somewhere common.
+/// Describes the entry for a state in the database.
+#[derive(Clone, Debug, Eq, PartialEq, BorshDeserialize, BorshSerialize)]
+pub(crate) struct StateInstanceEntry {
+    pub(crate) toplevel_state: Chainstate,
+}
```
delbonis: I think it's probably fine to go ahead and add a `strata-db-store-comon` crate?

crates/db-store-sled/src/lib.rs
delbonis, line 1: Couple of points:

* We should try to avoid replicating the same patterns as we have in the RocksDB crate.  It's fine to streamline all the init.  It should also exist in its own module to avoid cluttering the crate root.  Re-export whatever is needed from there.
* I also want to shift towards *not* exporting internal module structure in a crate unless it helps with organization.  So all the database types should not be `pub` and we should `pub use` just the important types themselves, if we even really need that (since we have `DatabaseBackend` now).

crates/db-store-sled/src/macros.rs
```diff
@@ -0,0 +1,125 @@
...
+                let bincode_options = ::bincode::options()
+                    .with_fixint_encoding()
+                    .with_big_endian();
+
+                bincode_options.serialize(self)
+                    .context("Failed to serialize key")
+                    .map_err(|err| ::typed_sled::codec::CodecError::SerializationFailed {
+                        schema: $table_name::tree_name(),
+                        source: err.into(),
+                    })
```
delbonis: Is bincode the only reasonable choice here?

The need for this might be reduced if we do a blanket `KeyCodec` impl for the integer types and respect `Ord` and whatnot.

crates/db-store-sled/src/macros.rs
```diff
@@ -0,0 +1,125 @@
...
+            }
+        }
+
+        impl_borsh_value_codec!($table_name, $value);
+    };
+}
+
+#[macro_export]
+macro_rules! impl_borsh_value_codec {
```
delbonis, line 63: Oh this is what I was referencing with one of the points I made about having a macro in the other crate for shimming to borsh.

crates/db-store-sled/src/macros.rs
```diff
@@ -0,0 +1,125 @@
...
+    ($table_name:ident, $value:ty) => {
+        impl ::typed_sled::codec::ValueCodec<$table_name> for $value {
+            fn encode_value(
+                &self,
+            ) -> ::std::result::Result<::std::vec::Vec<u8>, ::typed_sled::codec::CodecError> {
+                ::borsh::to_vec(self).map_err(|err| {
```
delbonis, line 69: If the trait uses `IVec` then this might be slightly tricky.  We'd want to avoid forcing an alloc if we can make the buffer stack allocated in the first place, but `IVec` doesn't support `Write` so we can't do it directly.  We'd need like an `IVecBuffer` type that starts as an array and then converts itself to a vec when it realizes that we're writing more than we can stack alloc in an `IVec.

crates/db-store-sled/src/sync_event/schemas.rs
delbonis, line 1: Maybe hold off on merging this until after the CSM refactor PR merges so that we can just drop this entirely.

crates/db-store-sled/src/broadcaster/db.rs
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+}
+
+impl L1BroadcastDatabase for L1BroadcastDBSled {
+    fn put_tx_entry(&self, txid: Buf32, txentry: L1TxEntry) -> DbResult<Option<u64>> {
+        let next = self.get_next_idx()?;
+
+        let nxt = (&self.tx_tree, &self.tx_id_tree)
+            .transaction(|(txtree, txidtree)| {
```
delbonis: Is this not a race condition?

crates/db-store-sled/src/chain_state/db.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn clone_inst(&self, id: StateInstanceId) -> DbResult<StateInstanceId> {
+        let entry = self
+            .state_tree
+            .get(&id)?
+            .ok_or(DbError::MissingStateInstance)?;
+        let next_id = self.next_state_id_or_zero()?;
+        if next_id == 0 {
+            return Err(DbError::MissingStateInstance);
+        }
+        self.state_tree.insert(&next_id, &entry)?;
```
delbonis: Please make sure this is not a race condition!

crates/db-store-sled/src/chain_state/db.rs
```diff
@@ -0,0 +1,149 @@
...
+                .get(wb_id)?
+                .ok_or(DbError::MissingWriteBatch(*wb_id))?;
+            last_wb = Some(wb);
+        }
+
+        // Applying the last write batch is really simple.
+        if let Some(last_wb) = last_wb {
+            let entry = StateInstanceEntry::new(last_wb.into_toplevel());
+            self.state_tree.insert(&state_id, &entry)?;
+        }
```
delbonis: This should be done all in a transaction.

functional-tests/entry.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 from gevent import monkey
 
-# This is important for locust to work with flexitest.
+# This important for locust to work with flexitest.
```
storopoli: Not needed

storopoli (changes requested):
We should moved `typed-sled` to it's own repo.
Discussed in https://github.com/alpenlabs/alpen/pull/958#issuecomment-3160878497

crates/db-store-sled/src/broadcaster/db.rs
bewakes: Not actually because since we read the key by doing a `.get(nxt)` check inside the transaction, if during insert the existing value is different from when we read, the transaction rolls back. In order to retry, I've now used `transaction_with_retry` method.

crates/db-store-sled/src/chain_state/db.rs
bewakes: In this case, a CAS will do because we are inserting on the state tree only for `state_id`. Updated accordingly.

crates/db-store-sled/src/broadcaster/db.rs
bewakes: Getting a valid next_id is not super clean at the moment though. We need to fetch and see if it is None. This is because, rollback only occurs when the key we read is modified when we write to it in a transaction.

crates/db-store-sled/src/chain_state/db.rs
bewakes: Updated to have this in a transaction. Same applies with the next id as above comment.

crates/db-store-sled/src/macros.rs
bewakes: Yeah, there's no reason to use bincode. I did blanket impls for integer types and got rid of bincode.

crates/db-store-sled/src/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+            l2_db,
+            sync_event_db,
+            client_state_db,
+            chain_state_db,
+            checkpoint_db,
+            writer_db,
+            prover_db,
+            broadcast_db,
+        }
+    }
```
delbonis: This should take the underlying Sled database handle and construct each of the typed database types from it.  I realize this was probably copying the pattern from RocksDB but that should separately be updated.

crates/db-store-sled/src/macros.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+#[macro_export]
+macro_rules! define_sled_database {
+    (
+        $(#[$meta:meta])*
+        pub struct $db_name:ident {
+            $($field:ident: $schema:ty),* $(,)?
+        }
+        $(, config: $config_usage:ident)?
```
delbonis: This macro is probably a good idea, but this syntax is weird.  Not sure what the *best* way to improve it would be.

Also I don't think we need to `#[macro_export]` this, do we?

crates/db-store-sled/src/prover/db.rs
```diff
@@ -0,0 +1,70 @@
...
+use zkaleido::ProofReceiptWithMetadata;
+
+use super::schemas::{ProofDepsSchema, ProofSchema};
+use crate::define_sled_database;
+
+define_sled_database!(
+    pub struct ProofDBSled {
+        proof_tree: ProofSchema,
+        proof_deps_tree: ProofDepsSchema,
+    }, config: _config
```
delbonis: Specifically this part is weird and it's not obvious what it's doing or why we need it from the syntax.

crates/db-store-sled/src/macros.rs
```diff
@@ -0,0 +1,134 @@
...
+macro_rules! sled_db_test_setup {
+    ($db_type:ty, $test_macro:ident) => {
+        fn setup_db() -> $db_type {
+            let db = sled::Config::new().temporary(true).open().unwrap();
+            let sled_db = typed_sled::SledDb::new(db).unwrap();
+            let config = $crate::SledDbConfig::test();
+            <$db_type>::new(sled_db.into(), config).unwrap()
+        }
+
+        $test_macro!(setup_db());
```
delbonis, line 101: This is crazy, I didn't realize the macro expander would let you do this.

crates/db/src/errors.rs
```diff
@@ -78,19 +78,19 @@ pub enum DbError {
     #[error("tried to revert to index {0} above current tip {1}")]
     RevertAboveCurrent(u64, u64),
 
-    #[error("IO Error (rocksdb)")]
+    #[error("IO Error")]
     IoError,
```
delbonis: While we're here, we should probably expose the error message here if possible.

crates/reth/db/src/sled/db.rs
```diff
@@ -79,34 +85,49 @@ impl<DB: SchemaDBOperations> StateDiffProvider for WitnessDB<DB> {
...
+    ) -> DbResult<()> {
+        (&self.block_hash_by_number_tree, &self.state_diff_tree)
+            .transaction(
+                |(bht, sdt)| -> ConflictableTransactionResult<(), typed_sled::error::Error> {
+                    bht.insert(&block_number, &block_hash.to_vec())?;
+                    let serialized = match bincode::serialize(state_diff) {
+                        Ok(data) => data,
+                        Err(err) => {
+                            return Err(sled::transaction::ConflictableTransactionError::Abort(
+                                sled::Error::Unsupported(format!("Serialization failed: {}", err))
```
delbonis: Should clean up these imports somehow.

crates/db-store-sled/src/lib.rs
bewakes: Yes, that's true. This is the result of copying the pattern from rocksdb. Cleaned it now.

crates/db-store-sled/src/macros.rs
bewakes: The macro does reduce a lot of boilerplate. I asked claude to define one for me and it got this definition. It's probably fine because this closely resembles how one would define a struct, except for the optional config param.

crates/db/src/errors.rs
bewakes: Updated the variant to be `IoError(String)`, although it is not used anywhere atm.

crates/reth/db/src/sled/db.rs
bewakes: I've cleaned up for non `Error` imports. For `Error`s, I think we should qualify them explicitly as otherwise it might cause confusion with other Error types.

crates/db-store-sled/src/chain_state/schemas.rs
bewakes: Probably not yet, now that the rocksdb crate does not exist.

crates/db-store-sled/src/lib.rs
bewakes: Agree. And I've got rid of those patterns.

crates/db-store-sled/src/macros.rs
bewakes: I've now cleaned up the macro definition and made it simpler.

Cargo.toml
```diff
@@ -315,14 +318,14 @@ reqwest = { version = "0.12.7", default-features = false, features = [
...
 ] }
-rockbound = { git = "https://github.com/alpenlabs/rockbound", rev = "v2.0.1-alpen.3" }
 secp256k1 = { version = "0.29.1", features = ["global-context", "std"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.94"
 serde_with = "3.3.0"
 sha2 = "0.10"
 shrex = { version = "1", features = ["serde"] }
 shrex_macros = "1"
+sled = "0.34.7"
```
storopoli, line 334: How much of a lift would be to move this to the `1.0.0-alpha.foo` releases?

EDIT: probably we would need to change this on `typed-sled` as well :/

storopoli (dismissed):
ACK d6502d2f852297f38a22efa6462dd5a78523c239
But let's wait for @delbonis

crates/db/src/errors.rs
delbonis: Oh then just remove it??

Cargo.toml
delbonis: Yeah let's worry about that in a future PR.

delbonis (changes requested):
I should have realized this sooner, but the ticket said

> The deliverable here is to make a new crate like the rocksdb-store crate called sled-store that has impls for each of the database traits defined in db.  In another ticket we will set up the flagging to build one or the other backend.

Because I wanted to support *both* backends until we validated that Sled worked as expected and didn't have some unexpected performance regressions.  I wanted to take this incrementally.  This is why the dbtool has a db type arg in the first place.

But it too late for that, so I will tentatively approve this after the unresolved review comments have been addressed.  We can re-add the RocksDB backend (perhaps with some minor cleanups) in another PR.

delbonis:
@bewakes It's an easy performance fix for short keys, that's why Sled uses it.

storopoli (approved):
ACK 88c9f91aeb84235477c68ea7ef3d8db30cdb183f

crates/db-store-sled/src/macros.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+#[macro_export]
+macro_rules! impl_borsh_value_codec {
+    ($table_name:ident, $value:ty) => {
+        impl ::typed_sled::codec::ValueCodec<$table_name> for $value {
+            fn encode_value(
+                &self,
+            ) -> ::std::result::Result<::std::vec::Vec<u8>, ::typed_sled::codec::CodecError> {
+                ::borsh::to_vec(self).map_err(|err| {
```
delbonis, line 69: Bumping [this comment](https://github.com/alpenlabs/alpen/pull/979/files#r2257704978) about `IVec`.

delbonis (changes requested):
As mentioned in DM: it would be less disruptive to un-remove the RockDB backend in this PR, especially since [this PR](https://github.com/alpenlabs/alpen/pull/1019) that depends on it has been opened.

storopoli:
> it would be less disruptive to un-remove the RockDB backend in this PR, especially since [this PR](https://github.com/alpenlabs/alpen/pull/1019)

No, the idea is to remove it.
The PR will be changed to `sled` once this hits `main`.
I am for now benchmarking `RocksDB` for later comparisons with `sled`.
I'll mark #1019 as a draft to not send the wrong message.

crates/db-store-sled/src/macros.rs
bewakes: Oops, that comment is gone now :(. Could you share what your concrete suggestions regarding `IVec` are? I don't understand why we need to know about the existence of `IVec` at this level.

storopoli:
> Like, that wasn't what the ticket described indicated. The goal was to have feature flags as a transitional measure. I did say that "okay well so much for that", but then I decided against it because it would be overly disruptive.
> 
> (From STR-1233:)
> 
> > The deliverable here is to make a new crate like the rocksdb-store crate called sled-store that has impls for each of the database traits defined in db.  In another ticket we will set up the flagging to build one or the other backend.

@delbonis, thanks for flagging this.

@bewakes what do you think about bringing back `rocksdb` by removing commit 8731a92239ed972f9d6c8bf6a83a6d6113121671 and feature-gating `rocksdb` making sure that it is non-default and we make `sled` default?

bewakes:
@storopoli We can bring back rocksdb for sure which requires some updates related to changs on db traits(which probably shouldn't be that much). It's just that I feel like we don't benefit much from maintaining two database impls. For every changes in db traits, we need to reflect that on both impls.

Cargo.toml
```diff
@@ -120,6 +121,7 @@ strata-config = { path = "crates/config" }
 strata-consensus-logic = { path = "crates/consensus-logic" }
 strata-crypto = { path = "crates/crypto", default-features = false }
 strata-db = { path = "crates/db" }
+strata-db-store-rocksdb = { path = "crates/db-store-rocksdb" }
```
storopoli, line 124: shouldn't this be `optional`?

Cargo.toml
```diff
@@ -321,6 +323,7 @@ reqwest = { version = "0.12.7", default-features = false, features = [
   "zstd",
   "json",
 ] }
+rockbound = { git = "https://github.com/alpenlabs/rockbound", rev = "v2.0.1-alpen.3" }
```
storopoli: shouldn't this be `optional`

bin/alpen-reth/src/init_db.rs
```diff
@@ -0,0 +1,52 @@
...
+use typed_sled as _;
+#[cfg(feature = "rocksdb")]
+use {alpen_reth_db::rocksdb::WitnessDB as RocksWitnessDB, rockbound::OptimisticTransactionDB};
+#[cfg(all(feature = "sled", not(feature = "rocksdb")))]
+use {alpen_reth_db::sled::WitnessDB as SledWitnessDB, typed_sled::SledDb};
+
+// Type aliases for witness database
+#[cfg(all(feature = "sled", not(feature = "rocksdb")))]
+pub(crate) type WitnessDB = SledWitnessDB;
+#[cfg(feature = "rocksdb")]
```
storopoli: You should also guard here for `not(feature = "sled")`

bin/alpen-reth/src/init_db.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    let typed_sled =
+        SledDb::new(sled_db).map_err(|e| eyre!("Failed to create typed sled db: {}", e))?;
+
+    let witness_db = WitnessDB::new(Arc::new(typed_sled))
+        .map_err(|e| eyre!("Failed to create witness db: {}", e))?;
+    Ok(Arc::new(witness_db))
+}
+
+#[cfg(feature = "rocksdb")]
```
storopoli: You should also guard here for `not(feature = "sled")`

bin/strata-client/src/init_db.rs
```diff
@@ -0,0 +1,40 @@
+use std::{path::Path, sync::Arc};
+
+#[cfg(feature = "rocksdb")]
```
storopoli: You should also guard here for `not(feature = "sled")`

bin/strata-client/src/init_db.rs
```diff
@@ -0,0 +1,40 @@
...
+use strata_db_store_rocksdb::{
+    init_rocksdb_backend, open_rocksdb_database, DbOpsConfig, RocksDbBackend, ROCKSDB_NAME,
+};
+// Consume the sled dependency when rocksdb is active to avoid unused crate warnings
+#[cfg(feature = "rocksdb")]
```
storopoli: You should also guard here for `not(feature = "sled")`

bin/alpen-reth/src/init_db.rs
```diff
@@ -0,0 +1,52 @@
+use std::{fs, path::Path, sync::Arc};
+
+use eyre::{eyre, Context, Result};
+// Consume the sled dependencies when rocksdb is active to avoid unused crate warnings
+#[cfg(feature = "rocksdb")]
```
storopoli: You should also guard here for `not(feature = "sled")`

bin/alpen-reth/src/init_db.rs
```diff
@@ -0,0 +1,52 @@
...
+use sled as _;
+#[cfg(feature = "rocksdb")]
+use typed_sled as _;
+#[cfg(feature = "rocksdb")]
```
storopoli: You should also guard here for `not(feature = "sled")`

bin/strata-client/src/init_db.rs
```diff
@@ -0,0 +1,40 @@
...
+use strata_db_store_sled as _;
+#[cfg(all(feature = "sled", not(feature = "rocksdb")))]
+use strata_db_store_sled::{
+    init_core_dbs, open_sled_database, SledBackend, SledDbConfig, SLED_NAME,
+};
+
+// Type aliases for database backends
+#[cfg(all(feature = "sled", not(feature = "rocksdb")))]
+pub(crate) type DatabaseImpl = SledBackend;
+#[cfg(feature = "rocksdb")]
```
storopoli: You should also guard here for `not(feature = "sled")`

bin/strata-client/src/init_db.rs
```diff
@@ -0,0 +1,40 @@
...
+) -> anyhow::Result<Arc<DatabaseImpl>> {
+    #[cfg(all(feature = "sled", not(feature = "rocksdb")))]
+    {
+        let sled_db = open_sled_database(datadir, SLED_NAME)?;
+        let retry_delay_ms = 200u64;
+        let db_config = SledDbConfig::new_with_constant_backoff(db_retry_count, retry_delay_ms);
+        Ok(init_core_dbs(sled_db.clone(), db_config.clone())?)
+    }
+
+    #[cfg(feature = "rocksdb")]
```
storopoli: You should also guard here for `not(feature = "sled")`

bin/strata-dbtool/src/db.rs
```diff
@@ -1,30 +1,59 @@
 use std::{path::Path, sync::Arc};
 
 use strata_cli_common::errors::{DisplayableError, DisplayedError};
-use strata_db::traits::DatabaseBackend;
+#[cfg(feature = "rocksdb")]
+use strata_db_store_rocksdb::{
+    init_rocksdb_backend, open_rocksdb_database, DbOpsConfig, RocksDbBackend, ROCKSDB_NAME,
+};
+// Consume the sled dependency when rocksdb is active to avoid unused crate warnings
+#[cfg(feature = "rocksdb")]
```
storopoli: You should also guard here for `not(feature = "sled")`

bin/strata-dbtool/src/db.rs
```diff
@@ -1,30 +1,59 @@
...
             "sled" => Ok(DbType::Sled),
-            other => Err(format!("unknown db type {other}").into()),
+            #[cfg(feature = "rocksdb")]
+            "rocksdb" | "rocks" => Ok(DbType::Rocksdb),
+            other => {
+                let available_types = [
+                    #[cfg(all(feature = "sled", not(feature = "rocksdb")))]
+                    "sled",
+                    #[cfg(feature = "rocksdb")]
+                    "rocksdb",
```
storopoli: You should also guard here for `not(feature = "sled")`

bin/strata-dbtool/src/db.rs
```diff
@@ -1,30 +1,59 @@
...
+                Err(format!("unknown db type '{other}', available types: {available_types}").into())
+            }
         }
     }
 }
 
+// Type alias for database backend
+#[cfg(all(feature = "sled", not(feature = "rocksdb")))]
+type DatabaseImpl = SledBackend;
+#[cfg(feature = "rocksdb")]
```
storopoli: You should also guard here for `not(feature = "sled")`

bin/strata-dbtool/src/db.rs
```diff
@@ -34,6 +63,16 @@ pub(crate) fn open_database(
                 .internal_error("Could not open sled backend")
                 .map(Arc::new)?;
 
+            Ok(backend)
+        }
+        #[cfg(feature = "rocksdb")]
```
storopoli: You should also guard here for `not(feature = "sled")`

crates/reth/db/src/lib.rs
storopoli, line 1: You should also guard here for `not(feature = "sled")`

delbonis:
On the feature flags, I think it's safe to just do the checks as `#[cfg(feature = "...")]` and not include the `and(not(feature = "..."))`.  Including both would cause things to break horribly *anyways*, and we can force it by having a line like this in binaries, maybe exported as a macro in the db crate?

```rs
#[cfg(and(feature = "sled", feature = "rocksdb")]
compile_error!("multiple database backends configured");
```

bin/strata-dbtool/src/db.rs
storopoli: Done in f8c9a6ffedcbe1e4b1741dafcb1b9b1845fdcb40

PR #989, opened 2025-08-15
Update rollup params with magic bytes and remove horizon height

prajwolrg:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


bin/datatool/src/util.rs
```diff
@@ -283,8 +284,29 @@ fn exec_genparams(cmd: SubcParams, ctx: &mut CmdContext) -> anyhow::Result<()> {
...
+        let name_bytes = name_str.as_bytes();
+        if name_bytes.len() != 4 {
+            return Err(anyhow::anyhow!(
+                "Name must be exactly 4 bytes long, got {}",
+                name_bytes.len()
+            ));
+        }
+
+        // Convert to MagicBytes (safe since we validated length is 4)
+        name_bytes.try_into().unwrap()
```
bewakes: Use of `unwrap`. This function should return a result as it is already returning one. At the very least, need to use `expect`.

bewakes: Instead of unwrap, maybe there's something like this that can be done:

```rust
// Check for length
// ...
let mut buf: [u8;4] = [0; 4];
buf.copy_from_slice(name_bytes);
buf.into()
```

crates/l1tx/src/filter/mod.rs
```diff
@@ -234,60 +228,4 @@ mod test {
         let deposits: Vec<_> = try_parse_tx_deposit(&tx, &filter_conf).collect();
         assert!(deposits.is_empty(), "Should find no deposit request");
     }
-
-    #[test]
-    fn test_deposit_tx_with_test_vector() {
```
bewakes, line 239: Why is the test removed?

prajwolrg: The test is removed because it was using fixed test transactions which had "strata" as magic bytes, but we now only support 4-bytes magic. The tests with custom magic bytes are added [here](https://github.com/alpenlabs/alpen/blob/main/crates/asm/subprotocols/bridge-v1/src/txs/deposit/validation.rs#L173) inside asm bridge subprotocol.

PR #990, opened 2025-08-16
feat(sequencer): implement get latest unproven checkpoint idx API

irnb:
## Description

This PR adds get_latest_unproven_checkpoint_idx, preventing the prover client from re-proving already-proven checkpoints after restarts and ensuring the proven-checkpoint sequence remains contiguous (no gaps).

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Two implementation options:

1. **Use the existing API and queries.**
   This is suboptimal: in the worst case we issue multiple DB reads and chain several async calls to find the latest unproven checkpoint.

2. **Add a dedicated DB query.**
   Better, but it still requires scanning backward, which isn’t ideal.

The long-term solution is to add a **secondary index on unproven checkpoints**. For now, I implemented the stopgap in **RocksDB**, since there’s an open PR #979  for the **Sled** migration. After that PR merges, I’ll update the query (and introduces a secondary index for O(1) lookups.) to avoid doing the work twice.

I’d really appreciate feedback on the non-RocksDB parts, those are the most helpful to review at this stage.

## Checklist

## Related Issues

- STR-1567
- STR-1569

### DIFF
--- a/functional-tests/tests/dbtool/revert_chainstate_delete_blocks.py
+++ b/functional-tests/tests/dbtool/revert_chainstate_delete_blocks.py
@@ -3,7 +3,7 @@ import flexitest
 from envs import net_settings, testenv
 from mixins.dbtool_mixin import SequencerDbtoolMixin
 from utils.dbtool import send_tx
-from utils.utils import ProverClientSettings, wait_until_with_value
+from utils.utils import ProverClientSettings
 
 
 @flexitest.register
@@ -148,30 +148,6 @@ class RevertChainstateDeleteBlocksTest(SequencerDbtoolMixin):
         self.reth.start()
         self.seq.start()
         self.seq_signer.start()
-
-        def fetch_latest_checkpoint():
-            try:
-                return self.seqrpc.strata_getLatestCheckpointIndex()
-            except Exception as exc:
-                self.warning(f"failed to read latest checkpoint index: {exc}")
-                return None
-
-        current_checkpoint = wait_until_with_value(
-            fetch_latest_checkpoint,
-            lambda value: value is not None,
-            timeout=60,
-            step=1.0,
-            error_with="Timeout fetching latest checkpoint index after restart",
-        )
-        self.info(f"Checkpoint index before wait: {current_checkpoint}")
-
-        seq_waiter.wait_until_latest_checkpoint_at(
-            checkpt_idx_before_revert + 1,
-            timeout=20,
-        )
-
-        updated_checkpoint = fetch_latest_checkpoint()
-        self.info(f"Checkpoint index after wait: {updated_checkpoint}")
 
         # Wait for block production to resume
         seq_waiter.wait_until_chain_tip_exceeds(old_ol_block_number + 1, timeout=30)
--- a/functional-tests/tests/dbtool/revert_chainstate_fn.py
+++ b/functional-tests/tests/dbtool/revert_chainstate_fn.py
@@ -3,7 +3,6 @@ import flexitest
 from envs import net_settings, testenv
 from mixins.dbtool_mixin import FullnodeDbtoolMixin
 from utils.dbtool import send_tx
-from utils.utils import wait_until_with_value
 
 
 @flexitest.register
@@ -169,30 +168,6 @@ class RevertChainstateFnTest(FullnodeDbtoolMixin):
         self.seq_signer.start()
         self.follower_1_reth.start()
         self.follower_1_node.start()
-
-        def fetch_latest_checkpoint():
-            try:
-                return self.seqrpc.strata_getLatestCheckpointIndex()
-            except Exception as exc:
-                self.warning(f"failed to read latest checkpoint index: {exc}")
-                return None
-
-        current_checkpoint = wait_until_with_value(
-            fetch_latest_checkpoint,
-            lambda value: value is not None,
-            timeout=60,
-            step=1.0,
-            error_with="Timeout fetching latest checkpoint index after restart",
-        )
-        self.info(f"Checkpoint index before wait: {current_checkpoint}")
-
-        seq_waiter.wait_until_latest_checkpoint_at(
-            checkpt_idx_before_revert + 1,
-            timeout=120,
-        )
-
-        updated_checkpoint = fetch_latest_checkpoint()
-        self.info(f"Checkpoint index after wait: {updated_checkpoint}")
 
         # Wait for block production to resume to sync
         seq_waiter.wait_until_chain_tip_exceeds(old_seq_ol_block_number + 1, timeout=30)
--- a/functional-tests/tests/dbtool/revert_chainstate_seq.py
+++ b/functional-tests/tests/dbtool/revert_chainstate_seq.py
@@ -3,7 +3,7 @@ import flexitest
 from envs import net_settings, testenv
 from mixins.dbtool_mixin import SequencerDbtoolMixin
 from utils.dbtool import send_tx
-from utils.utils import ProverClientSettings, wait_until_with_value
+from utils.utils import ProverClientSettings
 
 
 @flexitest.register
@@ -148,30 +148,6 @@ class RevertChainstateSeqTest(SequencerDbtoolMixin):
         self.reth.start()
         self.seq.start()
         self.seq_signer.start()
-
-        def fetch_latest_checkpoint():
-            try:
-                return self.seqrpc.strata_getLatestCheckpointIndex()
-            except Exception as exc:
-                self.warning(f"failed to read latest checkpoint index: {exc}")
-                return None
-
-        current_checkpoint = wait_until_with_value(
-            fetch_latest_checkpoint,
-            lambda value: value is not None,
-            timeout=60,
-            step=1.0,
-            error_with="Timeout fetching latest checkpoint index after restart",
-        )
-        self.info(f"Checkpoint index before wait: {current_checkpoint}")
-
-        seq_waiter.wait_until_latest_checkpoint_at(
-            checkpt_idx_before_revert + 1,
-            timeout=30,
-        )
-
-        updated_checkpoint = fetch_latest_checkpoint()
-        self.info(f"Checkpoint index after wait: {updated_checkpoint}")
 
         # Wait for block production to resume
         seq_waiter.wait_until_chain_tip_exceeds(old_ol_block_number + 1, timeout=30)
--- a/functional-tests/tests/dbtool/revert_checkpointed_block_fn.py
+++ b/functional-tests/tests/dbtool/revert_checkpointed_block_fn.py
@@ -3,7 +3,6 @@ import flexitest
 from envs import net_settings, testenv
 from mixins.dbtool_mixin import FullnodeDbtoolMixin
 from utils.dbtool import send_tx
-from utils.utils import wait_until_with_value
 
 
 @flexitest.register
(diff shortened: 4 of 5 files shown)


bin/prover-client/src/checkpoint_runner/fetch.rs
```diff
@@ -4,10 +4,13 @@ use strata_rpc_api::StrataApiClient;
 use super::errors::CheckpointResult;
 use crate::checkpoint_runner::errors::CheckpointError;
 
-/// Fetches the latest checkpoint index from the sequencer client.
-pub(crate) async fn fetch_latest_checkpoint_index(cl_client: &HttpClient) -> CheckpointResult<u64> {
+/// Fetches the latest unproven checkpoint index from the sequencer client.
+/// This ensures sequential processing and prevents gaps in proof generation.
```
barakshani: actually this shouldn't been "latest" but "earliest", shouldn't it?

crates/db-store-rocksdb/src/checkpoint/db.rs
```diff
@@ -101,6 +101,48 @@ impl CheckpointDatabase for RBCheckpointDB {
     fn get_last_checkpoint_idx(&self) -> DbResult<Option<u64>> {
         Ok(rockbound::utils::get_last::<CheckpointSchema>(&*self.db)?.map(|(x, _)| x))
     }
+
+    fn get_latest_unproven_checkpoint_idx(&self) -> DbResult<Option<u64>> {
```
barakshani: This, I believe, should be oldest/earliest. The logic seems correct.

crates/db-store-rocksdb/src/checkpoint/db.rs
```diff
@@ -101,6 +101,48 @@ impl CheckpointDatabase for RBCheckpointDB {
...
+                            break;
+                        }
+                    }
+
+                    // Use saturating_add to prevent overflow
+                    let next_to_prove = latest_proven_idx.map_or(0, |idx| idx.saturating_add(1));
+
+                    Ok::<_, anyhow::Error>(match txn.get::<CheckpointSchema>(&next_to_prove)? {
+                        Some(entry)
+                            if entry.proving_status == CheckpointProvingStatus::PendingProof =>
```
barakshani, line 243: why do we need this extra check?

shouldn't we have another status for _currently being proven_ proofs?

bin/prover-client/src/checkpoint_runner/fetch.rs
evgenyzdanovich: to be clear, it should be "most recent" unproven checkpoint that is next to be proven.

barakshani: If checkpoint 100 is proven, but checkpoint 101, 102 and 103 aren't, we first need to prove 101. most recent is 103.
Right now there is always only one unproven checkpoint (we won't create 102 until 101 is proven), but in the future we may change this design.

crates/db-store-rocksdb/src/checkpoint/db.rs
delbonis: That would add more steps to the procedure and I'm not sure how much value it'd provide.  We track the proofs that are in-progress in the prover client, I don't think it needs to be persisted.

irnb:
We now derive `pending_proof_tree` from `put_checkpoint`, but that means existing PendingProof rows written before this build will never enter the index. 
On upgrade `get_latest_unproven_checkpoint_idx` will yield None until a brand-new checkpoint arrives. 
Do we want a migration/backfill (e.g. rebuild when the tree is empty) or is it acceptable because networks that adopt this branch will start from genesis?

@storopoli

crates/db-store-rocksdb/src/checkpoint/db.rs
irnb: wdyt about  `get_next_unproven_checkpoint_index`?

bin/prover-client/src/checkpoint_runner/fetch.rs
barakshani: @evgenyzdanovich let's get to an agreement over this.

crates/db-store-rocksdb/src/checkpoint/db.rs
barakshani: I'd prefer something else, "next" doesn't mean much to me in this case.

crates/db-store-rocksdb/src/checkpoint/db.rs
irnb: I was considering whether a prover/sequencer race could cause us to enqueue a proof for an index that just flipped to `ProofReady` in another db transaction. 
However, this doesn’t seem like a valid concern, since the prover client should handle such a scenario. That’s why I didn’t add this extra check in sled query.

storopoli:
> We now derive `pending_proof_tree` from `put_checkpoint`, but that means existing PendingProof rows written before this build will never enter the index. On upgrade `get_latest_unproven_checkpoint_idx` will yield None until a brand-new checkpoint arrives. Do we want a migration/backfill (e.g. rebuild when the tree is empty) or is it acceptable because networks that adopt this branch will start from genesis?
> 
> @storopoli

I don't see a reason to backport this.
Hence, no need to do a migration.

@sapinb @barakshani?

crates/db-store-rocksdb/src/checkpoint/db.rs
irnb: I think it depends on perspective 
from the prover’s side it’s the **next unproven** checkpoint,
while from the sequencer’s side it’s the **earliest unproven** checkpoint.

That said, I’m good with `get_earliest_unproven_checkpoint_idx` too.

irnb:
On this branch `ProverClientRpc::prove_latest_checkpoint` now short-circuits when the sequencer reports no “next unproven” checkpoint (`bin/prover-client/src/rpc_server.rs:133-163`), unlike the previous implementation that always tried the latest index.


fixing.

crates/db/src/traits.rs
```diff
@@ -205,6 +205,10 @@ pub trait CheckpointDatabase: Send + Sync + 'static {
     /// This method deletes all checkpoints with epoch index >= start_epoch.
     /// Returns a vector of deleted epoch indices.
     fn del_checkpoints_from_epoch(&self, start_epoch: u64) -> DbResult<Vec<u64>>;
+
+    /// Get the next checkpoint index that has PendingProof status.
+    /// Returns the highest index checkpoint that still needs proof generation.
```
barakshani: lowest?

crates/db-store-sled/src/checkpoint/db.rs
```diff
@@ -61,7 +66,24 @@ impl CheckpointDatabase for CheckpointDBSled {
     }
 
     fn put_checkpoint(&self, epoch: u64, entry: CheckpointEntry) -> DbResult<()> {
```
barakshani, line 68: This is _put in pending proof tree_, right? That is, we only "store" this checkpoint if it is pending a proof.

crates/db-store-sled/src/checkpoint/db.rs
```diff
@@ -61,7 +66,24 @@ impl CheckpointDatabase for CheckpointDBSled {
...
+
+        self.config.with_retry(
+            (&self.checkpoint_tree, &self.pending_proof_tree),
+            |(ct, pt)| {
+                ct.insert(&epoch, &entry)?;
+
+                if is_pending {
+                    pt.insert(&epoch, &())?;
+                } else {
+                    pt.remove(&epoch)?;
```
barakshani, line 79: I guess this function is really _put or remove_.

crates/db-store-sled/src/checkpoint/db.rs
```diff
@@ -131,13 +153,18 @@ impl CheckpointDatabase for CheckpointDBSled {
...
-                .compare_and_swap(epoch, old_item, None)?;
-        }
-        Ok(exists)
+        self.config.with_retry(
+            (&self.checkpoint_tree, &self.pending_proof_tree),
+            |(ct, pt)| {
+                let existing = ct.get(&epoch)?;
+                if existing.is_some() {
+                    ct.remove(&epoch)?;
+                    pt.remove(&epoch)?;
```
barakshani, line 162: We assume here that a checkpoint cannot be in the pending proof tree if it isn't in the checkpoint tree? Is the only way to insert checkpoints to the tree of pending proofs is via `put_checkpoint`? Similarly, is the only way to remove checkpoints from the checkpoint tree is either here or in `del_checkpoints_from_epoch`?

crates/db-tests/src/checkpoint_tests.rs
```diff
@@ -362,6 +365,134 @@ pub fn test_del_checkpoints_empty_database(db: &impl CheckpointDatabase) {
...
+        };
+        db.put_checkpoint(i, checkpoint).unwrap();
+    }
+
+    // Should return 3 (first unproven after last proven)
+    let result = db.get_next_unproven_checkpoint_idx().unwrap();
+    assert_eq!(result, Some(3));
+}
+
+pub fn test_get_next_unproven_checkpoint_idx_no_proven_yet(db: &impl CheckpointDatabase) {
```
barakshani: How is this different from `test_get_next_unproven_checkpoint_idx_all_pending`?

crates/db-tests/src/checkpoint_tests.rs
```diff
@@ -362,6 +365,134 @@ pub fn test_del_checkpoints_empty_database(db: &impl CheckpointDatabase) {
...
+        checkpoint.proving_status = CheckpointProvingStatus::PendingProof;
+        db.put_checkpoint(i, checkpoint).unwrap();
+    }
+
+    // Should return 0 (start from beginning when no proven checkpoints)
+    let result = db.get_next_unproven_checkpoint_idx().unwrap();
+    assert_eq!(result, Some(0));
+}
+
+pub fn test_get_next_unproven_checkpoint_idx_latest_proven_no_next(db: &impl CheckpointDatabase) {
```
barakshani: How is this different from `test_get_next_unproven_checkpoint_idx_all_ready`?

crates/db-tests/src/checkpoint_tests.rs
```diff
@@ -362,6 +365,134 @@ pub fn test_del_checkpoints_empty_database(db: &impl CheckpointDatabase) {
...
+        checkpoint.proving_status = CheckpointProvingStatus::ProofReady;
+        db.put_checkpoint(i, checkpoint).unwrap();
+    }
+
+    // Should return None (no next checkpoint to prove)
+    let result = db.get_next_unproven_checkpoint_idx().unwrap();
+    assert_eq!(result, None);
+}
+
+pub fn test_get_next_unproven_checkpoint_idx_next_after_proven_is_ready(
```
barakshani: How is this different from `test_get_next_unproven_checkpoint_idx_mixed_sequential`?

crates/storage/src/managers/checkpoint.rs
```diff
@@ -117,4 +117,12 @@ impl CheckpointDbManager {
     pub fn get_last_checkpoint_blocking(&self) -> DbResult<Option<u64>> {
         self.ops.get_last_checkpoint_idx_blocking()
     }
+
+    pub async fn get_next_unproven_checkpoint_idx(&self) -> DbResult<Option<u64>> {
+        self.ops.get_next_unproven_checkpoint_idx_async().await
+    }
+
+    pub fn get_next_unproven_checkpoint_idx_blocking(&self) -> DbResult<Option<u64>> {
+        self.ops.get_next_unproven_checkpoint_idx_blocking()
```
barakshani, line 126: Where is the implementation of this? I can't see it [here](https://github.com/alpenlabs/alpen/blob/STR-1567-latest-unproven-checkpoint-idx/crates/storage/src/ops/checkpoint.rs).

delbonis (approved):
Generally concur with Barak's concerns but otherwise looks good.

crates/db-tests/src/checkpoint_tests.rs
irnb: Removed redundant tests that were accidentally left in before the push.
thanks for mentioning it.

crates/db-store-sled/src/checkpoint/db.rs
irnb: Yep, we always persist the full `CheckpointEntry` in `checkpoint_tree`, and then treat `pending_proof_tree` as the lightweight index of “still needs proving” epochs. 
When the status is `PendingProof` we insert the epoch (`pt.insert(&epoch, &())?),` otherwise we scrub it (`pt.remove(&epoch)?`).

crates/db-store-sled/src/checkpoint/db.rs
irnb: https://github.com/alpenlabs/alpen/pull/990#discussion_r2414141307

crates/db-store-sled/src/checkpoint/db.rs
irnb: Yes. 
The invariant is that `pending_proof_tree` is updated only inside the three sled paths we have here: `put_checkpoint`, `del_checkpoint`, and `del_checkpoints_from_epoch`. All of them run under the same `with_retry((ct, pt), …)` guard, so the checkpoint payload and its pending flag move together. There’s no other insertion path.

crates/storage/src/managers/checkpoint.rs
irnb: It’s actually there: `crates/storage/src/ops/checkpoint.rs: Line 20 `comes from the `inst_ops_simple` macro expansion, which generates both the async and blocking flavors of each op. 

The macro wires `get_next_unproven_checkpoint_idx_blocking` straight through to the underlying `CheckpointDatabase`, so the manager just forwards to that generated method.

PR #991, opened 2025-08-16
Refactor HeaderVS with Genesis L1 View and Accumulated PoW Tracking

prajwolrg:
## Description

This PR introduces improvements and cleanup across the header verification system and related components:

* **Genesis L1 View Support**:

  * Added `GenesisL1View` definition in `rollup-params`.
  * Implemented `genesis_l1_view` in rollup parameters and datatool to fetch it directly.

* **Header Verification State (HeaderVS) Enhancements**:

  * Removed optional `headerVS` from `L1BlockManifest`.
  * Now tracks **total accumulated Proof of Work (PoW)**.
  * Simplified verification by enforcing **full verification only**.
  * Limited stored timestamps to the **latest 11**, removing reorg support.
  * Included `BtcParams` in state for better context.

* **Refactors and Cleanups**:

  * Removed unused `SyncAction::L2Genesis`.
  * Made `SyncState` non-optional.
  * Dropped `horizon height` from `RollupParams`.
  * Added BTC client support in datatool.

### Type of Change

- [X] Documentation update
- [X] Refactor
- [X] New or updated tests
- [X] Dependency Update

## Checklist

- [X] I have updated the documentation if needed.
- [X] My changes do not introduce new warnings.


crates/primitives/src/l1/header_verification.rs
```diff
@@ -156,6 +188,9 @@ impl HeaderVerificationState {
         // Set the target for the next block
         self.next_block_target = self.next_target(header);
 
+        // Update total accumulated PoW
+        self.total_accumulated_pow += header.difficulty(&self.params);
```
uncomputable: I think the header difficulty is the lower bound on the actual header work. Maybe [Header.work()](https://docs.rs/bitcoin/latest/bitcoin/block/struct.Header.html#method.difficulty) is the correct method.

prajwolrg: Looked more into [Header.work()](https://docs.rs/bitcoin/latest/bitcoin/block/struct.Header.html#method.work). Both are derived with the [bits](https://docs.rs/bitcoin/latest/bitcoin/block/struct.Header.html#structfield.bits) field, so both define the lower bound on the actual header work. 

`difficulty` is the relative measure whereas `work` seems to be the absolute measure.

I think it is fine using the lower bound on the actual header work, because the canonical chain is defined by blocks being built on top of it.

uncomputable: I see. It seems fine to sum the lower bound, since the block that is mined is the first one above that bound.

Is it fine to sum relative measures, though?

uncomputable: > Difficulty represents how difficult the current target makes it to find a block, relative to how difficult it would be at the highest possible target (highest target == lowest difficulty).

If the highest possible target is constant, than all difficulty measures are relative to the same value. So we could sum them.

prajwolrg: > Is it fine to sum relative measures, though?

Yes, I think so. But summing work seems would be a better choice.

bin/datatool/src/util.rs
```diff
@@ -279,10 +283,18 @@ fn exec_genparams(cmd: SubcParams, ctx: &mut CmdContext) -> anyhow::Result<()> {
...
-        .map_err(|e| anyhow::anyhow!("Invalid L1 block hash: {}", e))?;
-    let genesis_l1_blkid = L1BlockId::from(block_hash);
+    // Create Bitcoin RPC client and fetch genesis L1 view
+    let bitcoin_client = Client::new(
+        cmd.bitcoin_rpc_url,
+        cmd.bitcoin_rpc_user,
+        cmd.bitcoin_rpc_password,
+        None,
+        None,
+    )
```
delbonis: Have the user pass the block and height in, don't make them have a running client and ship the whole RPC protocol stack for that.

delbonis (changes requested):
Impl looks good, but we shouldn't be constraining the user's behavior around network setup.

bin/datatool/src/util.rs
prajwolrg: We require not only the block and height, but also the timestamp of epoch and last 11 timestamps for header consensus check. Do we pass all that as input?

prajwolrg: A lot of users won't generate the rollup parameters themselves and instead use the one that has been supplied. Also, we require the user has set up the ZkVm correctly to generate the VerifyingKey, so I think it's fine to require user to have a running bitcoin client.

functional-tests/utils/utils.py
```diff
@@ -367,10 +367,16 @@ def generate_params(
         "--epoch-slots", str(settings.epoch_slots),
         "--horizon-height", str(settings.horizon_height),
         "--genesis-l1-height", str(settings.genesis_trigger),
-        "--genesis-l1-hash", genesis_l1_blockhash,
```
delbonis, line 370: ~~Wait, I didn't realize it already *used* to be like this.  Why was this changed?~~

I see now, I didn't see the reply earlier.

bin/datatool/src/util.rs
delbonis: I would argue that it'd be better to have a script (or separate binary?) that extracts this pow state and exports it as a hex blob.  Or better, prints the arguments that strata-datatool would accept so they can be copy-pasted or used in a shell expansion in a shell script.

That's a little weird, but shipping the whole RPC client stack with the tool that's just supposed to be for working with protocol data structures is really annoying, especially when (as you say) most users won't actually run that command.

prajwolrg: I agree that it's better if `strata-datatool` does not require shiping the whole RPC client stack. At the same time, I think splitting this into a separate binary for the sole purpose of extracting consensus inputs would add more complexity than it saves.

delbonis: How does this interact with testing?  Do we have to wait for the bitcoin node to start up before we can set up the strata network?

prajwolrg: Yes, that's correct. In all the functional tests, we would need the bitcoin node to start first. Even now we run bitcoin node in all the envs for the functional tests.

crates/primitives/src/l1/header_verification.rs
prajwolrg: Updated in [2307c38](https://github.com/alpenlabs/alpen/pull/991/commits/2307c38ce0511edd15bd80cfa1b284f55692e8d1)

bin/datatool/src/args.rs
```diff
@@ -203,6 +200,34 @@ pub(crate) struct SubcParams {
...
+        option,
+        description = "bitcoin RPC password (required when btc-client feature is enabled)"
+    )]
+    pub(crate) bitcoin_rpc_password: String,
+
+    #[cfg(not(feature = "btc-client"))]
+    #[argh(
+        option,
+        description = "path to JSON-serialized genesis L1 view (required when btc-client feature is disabled)"
+    )]
```
delbonis: This should always be allowed even if the feature is enabled.

bin/datatool/src/util.rs
```diff
@@ -564,3 +561,140 @@ fn get_genesis_block_info(genesis_json: &str) -> anyhow::Result<BlockInfo> {
...
+/// * `cmd` - Command parameters containing Bitcoin RPC connection details and file paths
+///
+/// # Returns
+/// * `Ok(GenesisL1View)` - The successfully retrieved genesis L1 view
+/// * `Err(anyhow::Error)` - If RPC connection fails, file reading fails, or JSON parsing fails
+fn get_genesis_l1_view(cmd: &SubcParams) -> anyhow::Result<GenesisL1View> {
+    #[cfg(feature = "btc-client")]
+    {
+        // When the btc-client feature is enabled, we can fetch the genesis L1 view from a Bitcoin
+        // node.
```
delbonis: It should only look to these options if the view file is not set.  If the view file is set and any of these options are, then it should use that and print a warning.

bin/datatool/src/util.rs
delbonis, line 1: A lot of the stuff in here related to fetching the stuff from the client should be in a separate file.  Then we can just feature flag out the whole module, and have the function that does the high-level work of deciding what to do in main or something.

delbonis (changes requested):
As a side note, how do we construct the view file?

prajwolrg:
> As a side note, how do we construct the view file?

added a new command to generate it

delbonis (approved):

PR #995, opened 2025-08-18
Convert chain worker crate to use service framework

delbonis:
## Description

This PR is an initial attempt at converting a real service over to using the service framework.  I chose this crate because it's pretty isolated and is closest in design to the framework design.  The next steps would be to convert over other services.

This was done in large part with Claude Code.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


crates/chain-worker/src/builder.rs
```diff
@@ -95,28 +96,33 @@ impl<W> ChainWorkerBuilder<W> {
...
+        // Create the service builder and get command handle.
+        let mut service_builder =
+            ServiceBuilder::<ChainWorkerService<W>, _>::new().with_state(service_state);
+
+        // Create the command handle before launching.
+        let command_handle = service_builder.create_command_handle(64);
+
+        // Launch the service using the sync worker.
+        let _service_monitor = service_builder
+            .launch_sync("chain_worker", executor)
```
bewakes, line 121: Wouldn't it be better if the `launch_*` methods returned `(handle, monitor)`? That way user wouldn't have to remember calling `create_command_handle` beforehand. Also, it feels more natural. This would also possibly get rid of `expect` used for input inside `launch_async`.

crates/chain-worker/src/service.rs
```diff
@@ -0,0 +1,296 @@
...
+    fn prepare_block_context<'w>(
+        &'w self,
+        _l2bc: &L2BlockCommitment,
+    ) -> WorkerResult<WorkerExecCtxImpl<'w, W>> {
+        Ok(WorkerExecCtxImpl {
+            worker_context: &self.context,
+        })
+    }
+
+    fn try_exec_block(&mut self, block: &L2BlockCommitment) -> WorkerResult<()> {
```
bewakes, line 170: Don't you think having all these "action" methods in state feels a bit OOP-styled?

crates/service/src/status.rs
```diff
@@ -19,20 +25,84 @@ impl<S: Service> ServiceMonitor<S> {
...
 /// This is intended to be object-safe so that we can have a collection of
 /// monitors for heterogeneous service types.
 pub trait StatusMonitor {
+    /// Fetches the latest status as a boxed `dyn Any`.
+    fn fetch_status_any(&mut self) -> anyhow::Result<AnyStatus>;
+
     /// Fetches the latest status as a JSON value.
-    fn fetch_status(&mut self) -> anyhow::Result<serde_json::Value>;
+    fn fetch_status_json(&mut self) -> anyhow::Result<serde_json::Value>;
 }
```
bewakes: Why not have a associated type in this trait, possibly with `serde:Serializer/Deserializer` and let the callee choose to serialize it's way? That way this doesn't have to depend on serde_json and possibly remove the need of `Any` and all the casting that might need to be done.

Something like:
```rust
pub trait StatusMonitor {
    type Status: Serialize + Deserialize;
    fn fetch_status(&mut self) -> anyhow::Result<Self::Status>;
}
```
Or even associate it with service:
```rust
pub trait StatusMonitor {
    type Service: Service; // inside service, we can restrict status to be de/serializable.
    fn fetch_status(&mut self) -> anyhow::Result<Self::Service::Status>;
}
```

I am not sure how acting on a particular monitor status would work in this case though.

But wait, do we even need different status types for different services? Can we make things a lott simpler by having a single type for service status? Afterall a service can be in either of some predefined states like "spawning,initializing,ready,running,etc". Having these would help us get rid of all the `dyn`s and properly handle the monitor statuses as well.

crates/chain-worker/src/service.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    context: W,
+    chain_exec: ChainExecutor,
+    exec_ctl_handle: ExecCtlHandle,
+    cur_tip: L2BlockCommitment,
+    last_finalized_epoch: Option<EpochCommitment>,
+    status_channel: StatusChannel,
+    runtime_handle: Handle,
+    initialized: bool,
+}
```
bewakes: For me, a state contains the things that can change during the running of a service. All the other static things should go into the Context. I think you share the same idea as well.

bewakes (commented):
I was expecting that using the framework would actually significantly reduce the LoCs. Also I feel like some things like status and monitor can be simpler.

crates/service/src/status.rs
delbonis: > Why not have a associated type in this trait

Having associated types or generic functions would make it not dyn-safe.  The purpose of the trait is to use it as dyn and not have to export the service types everywhere we might want to monitor them, since it might have complex generic parameters (as it does in the chain worker's case).

> But wait, do we even need different status types for different services? Can we make things a lott simpler by having a single type for service status?

Then how would we inspect status attributes particular to the service that we're exporting, like the current tip or lasts finalized epoch? :slightly_smiling_face: 

The status type isn't meant to just indicate the running-state of the service, it's supposed to expose useful high-level properties that other services might want to depend on.

crates/chain-worker/src/service.rs
delbonis: It is, I don't love it, but it simplifies some things like having to keep repeating the `<W: WorkerContext + blah...>` type params and makes it easier to reason about the flow.

For what it is, it's fairly self-contained.

crates/chain-worker/src/builder.rs
delbonis: I had done that in an initial iteration but it gets annoying quickly, since I would have to add another pair of launch functions.  Services that don't use the command handle concept are perfectly sensible, so it's valid for a consumer to not *want* to call `create_command_handle`.  It's only even defined for services that use the tokio mpsc channel input.

It's also not entirely nonsensical to have a handle to a service before it's launched, so it doesn't seem like it's worth it to couple it together and force you to launch it before you can obtain the handle.  I can imagine some contrived scenario where you'd have some interdependency between services that couldn't be untangled where you'd want to take advantage of this.

crates/chain-worker/src/service.rs
delbonis: I'm generally with you, things that are truly immutable probably make sense to encapsulate into a separate params-like structure that's passed separately.  However this isn't a state machine, so we have more freedom to be pragmatic about passing data around.  I'm also trying to avoid dramatically changing the structure of the service in this PR, it's just trying to be a direct translation.

But to actually judge each of this.

* `shared`: keeping this because it might be useful later for metrics collection, but maybe we'll lean more into the new status system that comes with the service framework?
* `params`: will be refactored soon anyways
* `chain_exec`: this doesn't contain any meaningful internal state, so we could actually be constructing it on the fly, but that could change so I didn't want to remove it yet
* `exec_ctl_handle`: yeah this will be removed soon, as we're flipping the relationship so that the exec ctl tasks listen to the FCM instead of this task calling out to them
* `cur_tip`, `last_finalized_epoch`: mutable state, naturally makes sense
* `status_channel`, `runtime_handle`: kinda annoying that we need these, but we need them to be able to wait until the chain has started before we want to start processing commands
* `initialized`: this was actually something claude invented, it's not actually necessary, but it's useful as a safety check just to make sure that we don't crash trying to do chain stuff before the chain has started, and if somehow we do there's a good warning

crates/chain-worker/src/builder.rs
```diff
@@ -2,23 +2,24 @@ use std::sync::Arc;
...
 
-/// Builder for creating and launching a chain worker task.
+/// Builder for constructing and launching a chain worker service.
 ///
-/// This encapsulates all the initialization logic and dependencies needed
-/// to spawn a chain worker, preventing implementation details from leaking
-/// into the caller. The builder launches the task and returns a handle to it.
+/// This encapsulates all the initialization logic and dependencies needed to
+/// launch a chain worker using the service framework, preventing impl details
+/// from leaking into the caller.  The builder launches the service and returns
```
storopoli, line 21:
```suggestion
/// from leaking into the caller. The builder launches the service and returns
```
"trey was here".
You should add to your `CLAUDE.md` that it should remove double spacing on docstrings.

crates/chain-worker/src/builder.rs
```diff
@@ -70,14 +71,14 @@ impl<W> ChainWorkerBuilder<W> {
         self
     }
 
-    /// Launch the chain worker task and return a handle to it.
+    /// Launch the chain worker service and return a handle to it.
     ///
-    /// This method validates all required dependencies, creates the necessary
-    /// channels, spawns the worker task using the provided executor, and returns
+    /// This method validates all required dependencies, creates the service state,
+    /// uses ServiceBuilder to set up the service infrastructure, and returns
```
storopoli:
```suggestion
    /// uses [`ServiceBuilder`] to set up the service infrastructure, and returns
```

crates/chain-worker/src/handle.rs
```diff
@@ -1,116 +1,101 @@
...
+use strata_service::{CommandHandle, ServiceError};
+use tokio::sync::Mutex;
 
 use crate::{WorkerError, WorkerResult, message::ChainWorkerMessage};
 
+/// Handle for interacting with the chain worker service.
 #[derive(Debug)]
-#[allow(unused)]
 pub struct ChainWorkerHandle {
+    #[allow(unused)]
```
storopoli: Can we add a reason here?

```suggestion
    #[allow(unused, reason = "foo")]
```

storopoli: Or change to `#[expect(unused)]`

crates/chain-worker/src/service.rs
```diff
@@ -0,0 +1,296 @@
...
+
+/// Service state for the chain worker.
+#[derive(Debug)]
+pub struct ChainWorkerServiceState<W> {
+    #[allow(unused)]
+    shared: Arc<Mutex<WorkerShared>>,
+
+    #[allow(unused)]
+    params: Arc<Params>,
```
storopoli: Add `reason` or change to `expect`?

crates/chain-worker/src/service.rs
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /// Takes the block and post-state and inserts database entries to reflect
+    /// the epoch being finished on-chain.
+    ///
+    /// There's some bookkeeping here that's slightly weird since in the way it
+    /// works now, the last block of an epoch brings the post-state to the new
+    /// epoch.  So the epoch's final state actually has cur_epoch be the *next*
+    /// epoch.  And the index we assign to the summary here actually uses the
+    /// "prev epoch", since that's what the epoch in question is here.
```
storopoli, line 221:
```suggestion
    /// There's some bookkeeping here that's slightly weird since in the way it
    /// works now, the last block of an epoch brings the post-state to the new
    /// epoch. So the epoch's final state actually has cur_epoch be the *next*
    /// epoch. And the index we assign to the summary here actually uses the
    /// "prev epoch", since that's what the epoch in question is here.
```

crates/chain-worker/src/service.rs
```diff
@@ -0,0 +1,296 @@
...
+            *epoch_final_state,
+        );
+
+        debug!(?summary, "completed chain epoch");
+        context.store_summary(summary)?;
+
+        Ok(())
+    }
+
+    /// Updates the current tip as managed by the worker.  This does not persist
```
storopoli, line 265:
```suggestion
    /// Updates the current tip as managed by the worker. This does not persist
```

crates/service/src/builder.rs
```diff
@@ -106,7 +106,7 @@ impl<T: ServiceMsg, S: Service<Msg = T>> ServiceBuilder<S, TokioMpscInput<T>> {
     /// # Panics
     ///
     /// If an input has already been set.
-    pub fn create_command_handle(&mut self, capacity: usize) -> CommandHandle<S> {
+    pub fn create_command_handle(&mut self, capacity: usize) -> CommandHandle<T> {
```
storopoli, line 109: This is a very innocent but profound change.
Why do we need this diff?

crates/service/src/command.rs
```diff
@@ -1,18 +1,26 @@
...
+///
+/// This is essentially just a wrapper over a MPSC sender, but with some
+/// convenience functions for common patterns.  It's expected that an instance
+/// of this type will be used inside of a handle type for the particular
+/// service.
 #[derive(Debug)]
-pub struct CommandHandle<S: Service> {
-    tx: mpsc::Sender<S::Msg>,
+pub struct CommandHandle<M> {
+    tx: mpsc::Sender<M>,
```
storopoli, line 18: Why are we getting rid of the bounds?

crates/service/src/status.rs
```diff
@@ -19,20 +25,84 @@ impl<S: Service> ServiceMonitor<S> {
...
 ///
 /// This is intended to be object-safe so that we can have a collection of
 /// monitors for heterogeneous service types.
 pub trait StatusMonitor {
+    /// Fetches the latest status as a boxed `dyn Any`.
+    fn fetch_status_any(&mut self) -> anyhow::Result<AnyStatus>;
+
     /// Fetches the latest status as a JSON value.
-    fn fetch_status(&mut self) -> anyhow::Result<serde_json::Value>;
+    fn fetch_status_json(&mut self) -> anyhow::Result<serde_json::Value>;
```
storopoli: Intuitively these shouldn't need a `&mut self` since we are reading values.
What am I missing? Are we going to change the returned `status`?

crates/service/src/status.rs
bewakes: I see, the status means the summary of what's happening in the service. In that case we might need something for the health check as well.

But in any case, we should be using the actual types even if we might need to deal with a bunch of types as you mentioned in the comments for `GenericStatusMonitor`. Doesn't dealing with json, and tyring to extract keys out of it(especially in a method below `query_status_field()`) feel like working in javascript? :D

crates/chain-worker/src/service.rs
bewakes: >  I'm also trying to avoid dramatically changing the structure of the service in this PR, it's just trying to be a direct translation.

I would say if the change is worth it and makes things more streamlined, we should go for it since we don't have any other services using this framework. The more services we add more difficult it will be to make changes.

>  However this isn't a state machine, so we have more freedom to be pragmatic about passing data around.

It's not necessarily required to be working with state machine to have these sort of separation. You are probably right about the pragmatic-ness, but I still think this is something that can/should be cleanly facilitated by the framework.

crates/chain-worker/src/builder.rs
bewakes: Having to imagine some contrived scenario probably indicates that we don't need it. I think it's fine if there's need for another launch functions as long as users will be calling one function. Or maybe we could tweak some things in the framework itself?

Basically, there's two things I am concerned about:
1. The ergonomics
2. Safety: There's nothing stopping a user to create a handle, pass it around and try to use it before the service is even started. This is about "making invalid state impossible to represent".

crates/chain-worker/src/service.rs
bewakes: I guess it's fine.

crates/chain-worker/src/builder.rs
delbonis: oh but I want the double spaced docstrings :)

crates/service/src/builder.rs
delbonis: Otherwise we'd have to do something annoying like `CommandHandle<ChainWorkerService<StorageCtxImpl>>` and that makes any services using it dependent on how the chain worker service is instantiated by its direct consumer crate, or we'd have to leak the generic types all over the place.

I was trying to be clever and make it like how the `elliptic_curve` crate defines a type like `AffinePoint<C: Curve>`, which worked for the simple scenarios I was doing in my head, but it ended up not not working out since services aren't *entirely* self-contained.

crates/service/src/command.rs
delbonis: It's just the message type now, so no need for everything else.

crates/service/src/status.rs
delbonis: Oh good point.

crates/chain-worker/src/builder.rs
bewakes: I think we should have a consensus on whether to have single or double spaces and adhere to that as a convention. Not sure if that can be enforced via linter.

crates/service/src/builder.rs
storopoli: Ok, resolving

crates/chain-worker/src/builder.rs
storopoli: @bewakes Yes, but @delbonis you are aware that you're the only one that does this, right?

storopoli: We can do a crazy regex `'\.  +'` to detect these in CI if needed.

crates/chain-worker/src/builder.rs
delbonis: Copying from DM: Creating a command handle like this is an ergonomic improvement and it's completely valid to use the command handle to submit commands before the service is launched.

PR #999, opened 2025-08-22
STR-1089: Refactor csm to explicitly follow L1 blocks.

evgenyzdanovich:
## Description

- Dry up `ClientState` struct and leave only `L1Checkpoint`
- Adjust persistence layer to store `L1Block => ClientState`
- CSM refactor to explicitly follow `L1Block` and digest L1 reorgs.
- Some logic to avoid querying `ClientState` on startup of some tasks.
- Removal of the dead code here and there.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


bin/strata-dbtool/src/cmd/l1.rs
```diff
@@ -125,8 +125,8 @@ pub(crate) fn get_l1_summary(
 
     let (client_state_update, _) = get_latest_client_state_update(db, None)?;
     let (client_state, _) = client_state_update.into_parts();
-    let horizon_l1_height = client_state.horizon_l1_height();
-    let genesis_l1_height = client_state.genesis_l1_height();
+    let horizon_l1_height = 0u64; //TODO(QQ): client_state.horizon_l1_height();
```
delbonis: This "horizon" concept is being removed in the PR @prajwolrg is working on.

crates/consensus-logic/src/csm/ctl.rs
```diff
@@ -1,62 +1,45 @@
...
-            .sync_ev_man
-            .write_sync_event_async(sync_event.clone())
-            .await?;
-        let msg = CsmMessage::EventInput(ev_idx);
-        if self.csm_tx.send(msg).await.is_err() {
-            warn!(%ev_idx, "sync event receiver closed when submitting");
+    async fn submit_event_async(&self, sync_event: L1BlockCommitment) -> anyhow::Result<()> {
+        trace!(%sync_event, "submitting sync event");
+        if self.csm_tx.send(sync_event).await.is_err() {
+            warn!(%sync_event, "sync event receiver closed when submitting");
```
delbonis: Might want to remove "sync event" terminology here, since it'd lead to confusion since they're actually L1 block commitments.

crates/consensus-logic/src/csm/worker.rs
```diff
@@ -77,98 +55,64 @@ impl WorkerState {
...
         Ok(outp)
     }
 
-    fn update_bookeeping(&mut self, ev_idx: u64, state: Arc<ClientState>) {
-        debug!(%ev_idx, ?state, "computed new consensus state");
+    fn update_bookeeping(&mut self, last_height: u64, state: Arc<ClientState>) {
+        debug!(%last_height, ?state, "computed new consensus state");
         self.cur_state = state;
-        self.cur_state_idx = ev_idx;
+        self.last_height = last_height;
```
delbonis: Should this be the a block commitment instead of just the height?

crates/consensus-logic/src/csm/worker.rs
```diff
@@ -186,160 +130,127 @@ pub fn client_worker_task<E: ExecEngineCtl>(
...
-fn apply_action(
-    action: SyncAction,
-    state: &WorkerState,
-    _engine: &impl ExecEngineCtl,
-    _status_channel: &StatusChannel,
-) -> anyhow::Result<()> {
+fn apply_action(action: SyncAction, state: &WorkerState) -> anyhow::Result<()> {
     let ckpt_db = state.storage.checkpoint();
     match action {
         SyncAction::FinalizeEpoch(epoch_comm) => {
```
delbonis, line 280: A lot of this "sync action" business we can rework too.  Stuff like the checkpoints being finalized, we only care about the last checkpoint, since all of its parents are implicitly finalized too.  But all of that can be computed from the states in memory, so we don't actually need to store an explicit authoritative record that it's finalized (we might still write it down to accelerate queries, but that's not authoritative).

crates/primitives/src/l1/block.rs
```diff
@@ -67,6 +69,13 @@ pub struct L1BlockCommitment {
     blkid: L1BlockId,
 }
 
+impl fmt::Display for L1BlockCommitment {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // Delegate to Debug formatting
+        write!(f, "{self:?}")
```
delbonis: I do this even sneakier, `<Self as fmt::Debug>::fmt(self, f)`, so it skips having to do the macro expansion and extra formatting calls that would hopefully get inlined.

crates/rpc/api/src/lib.rs
```diff
@@ -135,14 +134,6 @@ pub trait StrataApi {
...
-    #[method(name = "getSyncEvent")]
-    async fn get_sync_event(&self, idx: u64) -> RpcResult<Option<SyncEvent>>;
-
-    /// Gets the index of the last written sync event.
-    #[method(name = "getLastSyncEventIdx")]
-    async fn get_last_sync_event_idx(&self) -> RpcResult<u64>;
-
     /// Gets the client update output produced as a result of the sync event idx given.
     #[method(name = "getClientUpdateOutput")]
     async fn get_client_update_output(&self, idx: u64) -> RpcResult<Option<ClientUpdateOutput>>;
```
delbonis: Can probably also rework this to be like a `get_block_client_state`, maybe?

crates/storage/src/lib.rs
```diff
@@ -73,17 +67,16 @@ where
...
     let checkpoint_db = db.checkpoint_db();
 
     let l1_block_manager = Arc::new(L1BlockManager::new(pool.clone(), l1_db));
     let l2_block_manager = Arc::new(L2BlockManager::new(pool.clone(), l2_db));
     let chainstate_manager = Arc::new(ChainstateManager::new(pool.clone(), chainstate_db));
 
-    let sync_event_manager = Arc::new(SyncEventManager::new(pool.clone(), sync_event_db));
     let client_state_manager = Arc::new(
-        ClientStateManager::new(pool.clone(), client_state_db).context("open client state")?,
+        ClientStateManager::new(pool.clone(), client_state_db, l1_block_manager.clone())
```
delbonis: Ah I'm realizing this naming is really unfortunate now.  Two things called "client state manager".

I kinda wanted to rename all the `FooManager` types to `FooDbManager` to indicate that they're db-related, but it feels like this isn't the place to do that.

crates/storage/src/managers/client_state.rs
```diff
@@ -4,16 +4,25 @@
...
-use crate::{cache, ops};
+use crate::{
+    cache,
+    ops::{self, client_state::ClientStateOps},
+    L1BlockManager,
+};
 
 #[expect(missing_debug_implementations)]
 pub struct ClientStateManager {
     ops: ops::client_state::ClientStateOps,
```
delbonis: Should this still be the ops type or the manager type?

bin/strata-dbtool/src/cmd/l1.rs
evgenyzdanovich: ACK, thanks. Will eventually rebase the PR and resolve if anything.

crates/consensus-logic/src/csm/ctl.rs
evgenyzdanovich: done

crates/primitives/src/l1/block.rs
evgenyzdanovich: done

crates/rpc/api/src/lib.rs
evgenyzdanovich: ACK. yeah, I'll rework RPC related stuff altogether soon.

crates/storage/src/managers/client_state.rs
evgenyzdanovich: ops is fine, I think - most of the persistence layer follows "ops wrapped in a manager" convention.

crates/storage/src/lib.rs
evgenyzdanovich: not sure if I understand your comment, could you please clarify?

crates/consensus-logic/src/csm/worker.rs
evgenyzdanovich: it's better to store commitment, fixed.

crates/consensus-logic/src/csm/worker.rs
evgenyzdanovich: I agree, currently it writes some additional records primarily to optimize queries and make logic simpler.

Also, `Vec<SyncAction>` is persistent as part of `ClientUpdateOutput`, do we still need actions to be persisted?

evgenyzdanovich (commented):
PTAL @delbonis

bin/strata-dbtool/src/cmd/checkpoint.rs
```diff
@@ -42,6 +41,10 @@ pub(crate) struct GetCheckpointArgs {
 #[argh(subcommand, name = "get-checkpoints-summary")]
 /// Get checkpoints summary
 pub(crate) struct GetCheckpointsSummaryArgs {
+    /// epoch index
```
krsnapaudel: Is this still accurate? Probably not. Also should this be from_epoch?

bin/strata-dbtool/src/cmd/l1.rs
```diff
@@ -31,6 +29,10 @@ pub(crate) struct GetL1ManifestArgs {
 #[argh(subcommand, name = "get-l1-summary")]
 /// Get L1 summary
 pub(crate) struct GetL1SummaryArgs {
+    /// block id
```
krsnapaudel: Same here. It's not a block id.

crates/consensus-logic/src/csm/client_transition.rs
```diff
@@ -43,186 +41,154 @@ impl EventContext for StorageEventContext<'_> {
...
+    // Actualize the previous state to handle the reorg.
+    match next_block_height.cmp(&(cur_height + 1)) {
+        Ordering::Less => {
+            // Canonical chain reorg case: switch to the canonical block right before the chain
+            // fork.
+            // Unconditionally take the new client state, even though it comes
+            // from the fork and the height is less than expected.
+            // The reason for that is btcio specific - we receive blocks with less height only
+            // if btcio sees a longer fork of bitcoin, thus eventually we receive a longer chain.
+            let pre_fork_block =
```
delbonis: Why is this being handled here?  Shouldn't this be in the CSM worker control logic?

crates/consensus-logic/src/csm/client_transition.rs
```diff
@@ -43,186 +41,154 @@ impl EventContext for StorageEventContext<'_> {
...
 
+// TODO(QQ): decouple checkpoint extraction, finalization and actions.
 fn handle_block(
-    state: &mut ClientStateMut,
-    block: &L1BlockCommitment,
-    block_mf: &L1BlockManifest,
-    _context: &impl EventContext,
+    mut cur_state: ClientState,
+    cur_height: u64,
+    next_block_mf: &L1BlockManifest,
```
delbonis: Does this manifest correspond to cur_height?

crates/consensus-logic/src/csm/worker.rs
delbonis: If we do still need it then it should probably persist it as part of the output.  But if it's not being used by anything then we can remove it entirely.

crates/consensus-logic/src/csm/worker.rs
```diff
@@ -184,163 +125,129 @@ pub fn client_worker_task<E: ExecEngineCtl>(
...
+
+    debug!(%incoming_block, "processed OK");
+
+    Ok(())
 }
 
-/// Repeatedly calls `handle_sync_event`, retrying on failure, up to a limit
-/// after which we return with the most recent error.
-fn handle_sync_event_with_retry(
+fn process_block(
```
delbonis, line 213: Weird nit but it seems like the names of `process_block` and `handle_block` are kinda backwards.  I would think of it as being the other way around, like "handle (new tip) block (notification)" and "process block (state transition)".

crates/consensus-logic/src/csm/worker.rs
```diff
@@ -36,140 +26,91 @@ pub struct WorkerState {
...
-        let outp = state_mut.into_update();
-
-        Ok(outp)
+        // Compute the state transition.
+        Ok(client_transition::process_block(
+            self.cur_state.as_ref().clone(),
+            self.cur_block,
+            next_block,
+            &context,
+            &self.params,
```
delbonis: I might be missing something but do we check that cur_state always corresponds to the block's parent?  This should be computing it like a tree, and not trying to compute a block's CSM STF without its parent already being there.  The state transitions should always be computed following the actual chain.

This also ties back to the comment in `client_transition::process_block` about the reorg checks that seem like they should be here.

There btcio reader sends a bunch of messages about reorgs that were previously more important, but if we can just simplify that to saying "here's the new chain tip" with the understanding that the manifests in the database are present and just compute whatever states we need to, then that seems like it'd avoid some of this reorg handling logic (since it's offloaded to the reader) and we can just publish that state as the CSM status.

delbonis (changes requested):
The main changes being asked for seem to look good.  I have some questions about how we track the bookkeeping of the current state in the CSM worker and how reorgs are handled/recognized.

bin/strata-dbtool/src/cmd/l1.rs
evgenyzdanovich: done

bin/strata-dbtool/src/cmd/checkpoint.rs
evgenyzdanovich: done.

I think it's okay to have ti with `height_from` (analogous to l1-summary)

bin/strata-dbtool/src/cmd/client_state.rs
```diff
@@ -14,7 +17,7 @@ use crate::{
 pub(crate) struct GetClientStateUpdateArgs {
     /// client state update index; defaults to the latest
```
krsnapaudel: help text needs an update.

bin/strata-dbtool/src/cmd/l1.rs
```diff
@@ -31,6 +29,10 @@ pub(crate) struct GetL1ManifestArgs {
 #[argh(subcommand, name = "get-l1-summary")]
 /// Get L1 summary
 pub(crate) struct GetL1SummaryArgs {
+    /// start height describing the summary queried.
```
krsnapaudel: All other help texts don't have . at the end.

bin/strata-dbtool/README.md
```diff
@@ -68,9 +68,12 @@ strata-dbtool get-client-state-update [<update_index>] [OPTIONS]
 Shows a summary of all L1 manifests in the database.
 
 ```bash
-strata-dbtool get-l1-summary [OPTIONS]
+strata-dbtool get-l1-summary [<height_from>] [OPTIONS]
 ```
 
+**Arguments:**
+- `height_from` - L1 height to look up the summary about.
```
krsnapaudel: For some reason, I can't comment on `update_index` on L62 above. It needs an update. For this argument help text and another below, good to keep it consistent with what is printed by help.

bin/strata-dbtool/README.md
```diff
@@ -190,12 +193,20 @@ strata-dbtool get-l2-block 858c390aaaabd7c457cb24c955d06fb9de0f6666d0b692e3b1a01
 Shows a summary of all checkpoints in the database.
 
 ```bash
-strata-dbtool get-checkpoints-summary [OPTIONS]
+strata-dbtool get-checkpoints-summary <height_from>  [OPTIONS]
 ```
 
+**Arguments:**
+- `height_from` - Start height to query checkpoints.
```
krsnapaudel: See previous comment about help text consistency.

krsnapaudel:
@evgenyzdanovich Left comments about a few nits.

crates/consensus-logic/src/csm/worker.rs
evgenyzdanovich: Made it more apparent and made sure the parent is always present (or computed beforehand) and (cur_state, cur_block) actually a parent of `next_block`.

also, cleaned in `client_transition` along the way

crates/consensus-logic/src/csm/client_transition.rs
evgenyzdanovich: removed, not needed anymore.

bin/strata-dbtool/src/cmd/client_state.rs
evgenyzdanovich: done

bin/strata-dbtool/src/cmd/l1.rs
evgenyzdanovich: done

bin/strata-dbtool/README.md
evgenyzdanovich: done

bin/strata-dbtool/README.md
evgenyzdanovich: done

crates/consensus-logic/src/csm/client_transition.rs
evgenyzdanovich: outdated, I guess.

crates/consensus-logic/src/csm/worker.rs
evgenyzdanovich: actions themselves aren't "used" anywhere, but they are exposed through some RPC methods (not sure if it's needed)

bin/strata-dbtool/src/cmd/checkpoint.rs
```diff
@@ -42,6 +41,10 @@ pub(crate) struct GetCheckpointArgs {
 #[argh(subcommand, name = "get-checkpoints-summary")]
 /// Get checkpoints summary
 pub(crate) struct GetCheckpointsSummaryArgs {
+    /// start l1 height to query checkpoints from.
```
krsnapaudel: Nit: trailing dot should be removed.

bin/strata-dbtool/README.md
```diff
@@ -56,10 +56,10 @@ strata-dbtool get-syncinfo
 Shows the latest client state update information.
 
 ```bash
-strata-dbtool get-client-state-update [<update_index>] [OPTIONS]
+strata-dbtool get-client-state-update [<block_id>] [OPTIONS]
 ```
 **Arguments:**
-- `update_index` - Client state update index (number), defaults to the latest
+- `block_id` -  Client state block_id.
```
krsnapaudel: Nit: Remove trailing dot in all arguments. Just for consistency.

bin/strata-dbtool/README.md
krsnapaudel: Same as above. Also help text says: "start height describing the summary queried"

bin/strata-dbtool/README.md
```diff
@@ -190,12 +193,20 @@ strata-dbtool get-l2-block 858c390aaaabd7c457cb24c955d06fb9de0f6666d0b692e3b1a01
 Shows a summary of all checkpoints in the database.
 
 ```bash
-strata-dbtool get-checkpoints-summary [OPTIONS]
+strata-dbtool get-checkpoints-summary <height_from>  [OPTIONS]
 ```
 
+**Arguments:**
+- `height_from` - Start l1 height to query checkpoints from.
```
krsnapaudel: Nit: Traliing dot.

crates/consensus-logic/src/csm/client_transition.rs
```diff
@@ -43,183 +39,104 @@ impl EventContext for StorageEventContext<'_> {
...
+    if let Some(buried_h) = buried_height {
+        let block = context.get_l1_block_manifest_at_height(buried_h).ok();
+        if let Some(b) = block {
+            if let Ok(cs) = context.get_client_state(&b.into()) {
+                return cs.get_last_checkpoint();
+            }
+        }
+    }
+    None
 }
```
delbonis, line 130: Ok I understand what this is doing better now with these cleanups.  I *believe* we can just record the "last finalized checkpoint" in the CSM state instead of having to go and fetch it like this.  The way it's written makes the state transition *sorta* like a pure function, but not over the CSM state itself but over the history of states.  Sorting this out should improve testability.

*However*, in an earlier version of this comment I suggested that we could remove the `EventContext` trait (which would need to be renamed anyways) in order to make this work.  But actually we don't want to do that!  For the ASM stuff coming down the pipeline, we need to be able to make provable queries about historical ASM states, so we do still need something similar, just with only the manifest querying traits.

crates/consensus-logic/src/csm/worker.rs
```diff
@@ -36,140 +26,91 @@ pub struct WorkerState {
...
-        let ev = self.get_sync_event_ok(ev_idx)?;
-
-        debug!(%ev_idx, ?ev, "processing sync event");
+    /// Given the next l1 block, does the state transition, returning next [`ClientState`].
+    pub fn advance_consensus_state(
+        &self,
+        next_block: &L1BlockManifest,
+    ) -> anyhow::Result<(ClientState, Vec<SyncAction>)> {
+        let id = next_block.blkid();
+        debug!(%id, "processing l1 block");
```
delbonis, line 71: This feels like it could benefit from having sanity checks about blocks following from their parents, but this might be refactored in a subsequent PR and seems to work now, so this can probably be left as-is for now.

delbonis (approved):
Okay so I went back and forth on a thing here in my head, so maybe this is all actually fine and we can merge, with some refactors coming shortly after.  Approving.

delbonis:
Force merging due to trivial changes since last approval.

PR #1000, opened 2025-08-22
Part 1: Mock Bridge foundations

voidash:
## Description

Introduces a mock bridge implementation system for the Alpen rollup, enabling complete end-to-end testing and development of bridge operations without requiring bridge client. 

Since the original [PR](https://github.com/alpenlabs/alpen/pull/994) got too big. This is attempt to break it down into small pieces such that it's easily reviewable. 

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
There will be two other PRs accompanying this foundational PR. While making suggestions please bear in mind that there the other two parts accompany
- bridge transactions
- tests

## Checklist


crates/rpc/types/src/types.rs
```diff
@@ -363,6 +364,21 @@ impl From<CheckpointEntry> for RpcCheckpointInfo {
...
+    pub deposit_idx: u32,
+    /// Corresponding deposit txid
+    pub deposit_txid: Txid,
+    /// Quantity of L1 asset, for Bitcoin this is sats.
+    pub amt: BitcoinAmount,
+    /// Destination [`Descriptor`] for the withdrawal
+    pub destination: Descriptor,
+    /// operator index
+    pub operator_idx: OperatorIdx,
+}
```
delbonis: This should be `RpcWithdrawalAssignment`.  Intents are basically just messages with the destination and amount, when they're assigned to a deposit and operator then that produces an assignment.

crates/rpc/api/src/lib.rs
```diff
@@ -82,6 +82,9 @@ pub trait StrataApi {
     #[method(name = "getCurrentDepositById")]
     async fn get_current_deposit_by_id(&self, deposit_id: u32) -> RpcResult<RpcDepositEntry>;
 
+    #[method(name = "getWithdrawalIntent")]
+    async fn get_current_withdrawal_intent(&self) -> RpcResult<Vec<RpcWithdrawalIntent>>;
```
delbonis: Reflecting the other comment, this should be `getCurWithdrawalAssignments` and `get_cur_withdrawal_assignments`.

crates/util/python-utils/src/bridge/drt.rs
```diff
@@ -1,31 +1,41 @@
...
     error::Error,
     parse::{parse_address, parse_el_address, parse_xonly_pk},
-    taproot::{bridge_wallet, new_bitcoind_client, sync_wallet, taproot_wallet, ExtractP2trPubkey},
+    taproot::{new_bitcoind_client, sync_wallet, taproot_wallet, ExtractP2trPubkey},
+    utils::parse_operator_keys,
 };
 
+/// Static vector to store DepositRequestData instances
+pub(crate) static DEPOSIT_REQUEST_DATA_STORAGE: LazyLock<Mutex<Vec<DepositRequestData>>> =
+    LazyLock::new(|| Mutex::new(Vec::new()));
```
delbonis: No globals!

crates/util/python-utils/src/bridge/drt.rs
```diff
@@ -115,13 +130,39 @@ fn deposit_request_transaction_inner(
...
         builder.finish().expect("drt: invalid psbt")
     };
     wallet.sign(&mut psbt, Default::default()).unwrap();
 
     let tx = psbt.extract_tx().expect("drt: invalid tx");
+
+    // Create DepositRequestData and add it to the static vector
+    {
+        let mut storage = DEPOSIT_REQUEST_DATA_STORAGE.lock().unwrap();
+        let stake_index = storage.len() as u32; // Use the position in vector as stake index
```
delbonis: This should be implemented by creating a type (exporting it to Python) which has this state, and then these functions could be functions on that type to manipulate it.

Or really, why does this even need to track this data?  I don't see it being used anywhere else.  We could just pass in the index as an argument.

Using a global like this will break, since it persists across test invocations, so when we spin down a test env and spin up the next env, tests running in that env will break because they'll be using the old stake chain index.

delbonis (changes requested):

crates/util/python-utils/src/bridge/drt.rs
```diff
@@ -36,45 +46,56 @@ use crate::{
 #[pyfunction]
 pub(crate) fn deposit_request_transaction(
     el_address: String,
-    musig_bridge_pk: String,
+    operator_keys: Vec<String>,
```
delbonis: From other PR, didn't realize this in initial review: Keys should be passed around, if not as some "key" type values, at least as byte arrays.

crates/rpc/api/src/lib.rs
```diff
@@ -82,6 +82,9 @@ pub trait StrataApi {
     #[method(name = "getCurrentDepositById")]
     async fn get_current_deposit_by_id(&self, deposit_id: u32) -> RpcResult<RpcDepositEntry>;
 
+    #[method(name = "getCurWithdrawalAssignments")]
+    async fn get_current_withdrawal_assignments(&self) -> RpcResult<Vec<RpcWithdrawalAssignment>>;
+
```
storopoli: Why the ones above are `getCurrent` (already in production) and this one is `getCur`? This is bad API consistency.

crates/rpc/types/src/types.rs
```diff
@@ -363,6 +364,21 @@ impl From<CheckpointEntry> for RpcCheckpointInfo {
...
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+/// Deposit entry for RPC corresponding to [`WithdrawalIntent`]
+pub struct RpcWithdrawalAssignment {
+    /// Corresponding deposit id
+    pub deposit_idx: u32,
+    /// Corresponding deposit txid
+    pub deposit_txid: Txid,
+    /// Quantity of L1 asset, for Bitcoin this is sats.
```
storopoli: Why this is the only one that deserves a punctuation `.`?
No love for the others :)

crates/util/python-utils/src/bridge/drt.rs
```diff
@@ -2,79 +2,94 @@ use std::str::FromStr;
...
+    taproot::{new_bitcoind_client, sync_wallet, taproot_wallet, ExtractP2trPubkey},
 };
 
 /// Generates a deposit request transaction (DRT).
 ///
 /// # Arguments
 ///
 /// - `el_address`: Execution layer address of the account that will receive the funds.
-/// - `musig_bridge_pk`: MuSig bridge X-only public key.
+/// - `operator_keys`: Private keys of operator
```
storopoli: This diff is wrong.

crates/util/python-utils/src/bridge/drt.rs
```diff
@@ -2,79 +2,94 @@ use std::str::FromStr;
...
         bitcoind_password.as_str(),
     )?;
     let signed_tx = serialize(&signed_tx);
-    Ok(signed_tx)
+
+    Ok((signed_tx, deposit_request_data))
+}
+
+fn build_timelock_miniscript(recovery_xonly_pubkey: XOnlyPublicKey) -> ScriptBuf {
+    let script = format!("and_v(v:pk({}),older({}))", recovery_xonly_pubkey, 1008);
```
storopoli:
```suggestion
    let script = format!("and_v(v:pk({recovery_xonly_pubkey}),older({1_008}))");
```

crates/util/python-utils/src/bridge/drt.rs
```diff
@@ -115,14 +124,32 @@ fn deposit_request_transaction_inner(
...
         builder.fee_rate(fee_rate);
         builder.finish().expect("drt: invalid psbt")
     };
     wallet.sign(&mut psbt, Default::default()).unwrap();
 
     let tx = psbt.extract_tx().expect("drt: invalid tx");
-    Ok(tx)
+
+    // Create DepositRequestData - stake_index will be managed by Python side
+    let deposit_request_outpoint = bdk_wallet::bitcoin::OutPoint {
```
storopoli: import the `OutPoint` above

crates/util/python-utils/src/bridge/drt.rs
```diff
@@ -115,14 +124,32 @@ fn deposit_request_transaction_inner(
...
+        txid: tx.compute_txid(),
+        vout: 0,
+    };
+
+    let deposit_request_data = DepositRequestData {
+        deposit_request_outpoint,
+        stake_index: 0, // This will be set by Python side when storing in the list
```
storopoli: This is bad. Why not make it an `Option<u64>` instead?
Having your API depending on the good will of someone over the Python FFI bridge is bad design.

crates/util/python-utils/src/bridge/drt.rs
storopoli: Also why not use the `RECOVER_DELAY`?

crates/util/python-utils/src/bridge/drt.rs
```diff
@@ -2,79 +2,94 @@ use std::str::FromStr;
...
         bitcoind_user.as_str(),
         bitcoind_password.as_str(),
     )?;
     let signed_tx = serialize(&signed_tx);
-    Ok(signed_tx)
+
+    Ok((signed_tx, deposit_request_data))
+}
+
+fn build_timelock_miniscript(recovery_xonly_pubkey: XOnlyPublicKey) -> ScriptBuf {
```
storopoli: Isn't this the same function as the one that it is way better documented and type safe that you've deleted above, `bridge_in_descriptor`?

You could get the first element from the tuple that it returns, `desc`, and compile to a `ScriptBuf`.

crates/util/python-utils/src/bridge/drt.rs
```diff
@@ -458,13 +439,26 @@ pub(crate) fn get_balance_recovery_inner(
...
+    rollup_str: &[u8; 4],
+    evm_address: &RethAddress,
+    take_back_key: &XOnlyPublicKey,
+) -> Vec<u8> {
+    let mut data = rollup_str.to_vec();
+    data.extend(take_back_key.serialize());
+    data.extend(evm_address.as_slice());
+
+    data
+}
```
storopoli: You don't need a `Vec<u8>` for this. All inputs have known at compile time sizes. You can use [`make_buf`](https://github.com/alpenlabs/make_buf).
If in doubt, check the documentation or this example: https://github.com/alpenlabs/alpen/blob/0293f84c23ce579dfd3d522b13d8cc2a0f185933/bin/alpen-cli/src/recovery.rs#L277-L283

Cc @Zk2u (and thanks for `make_buf`).

crates/util/python-utils/src/bridge/drt.rs
```diff
@@ -487,6 +481,18 @@ mod tests {
         });
     }
 
+    /// create test operator keys
+    fn create_test_operator_keys() -> XOnlyPublicKey {
+        use rand::thread_rng;
+
+        // Generate cryptographically secure random keys using secp256k1
+        let _secp = Secp256k1::new();
```
storopoli: delete this.

crates/util/python-utils/src/bridge/drt.rs
```diff
@@ -487,6 +481,18 @@ mod tests {
...
+
+        let pair = Keypair::new(SECP256K1, &mut thread_rng());
```
storopoli: You should use `EvenSecretKey`: https://github.com/alpenlabs/alpen/blob/0293f84c23ce579dfd3d522b13d8cc2a0f185933/crates/primitives/src/crypto.rs#L55-L57

And not a `Keypair`.

crates/util/python-utils/src/bridge/mod.rs
```diff
@@ -0,0 +1,16 @@
+//! Bridge transaction utilities for Alpen rollup
+//! This module provides functionality for creating and signing bridge transactions:
```
storopoli, line 2:
```suggestion
//! This module provides functionality for creating and signing bridge
transactions:
//!
```

crates/util/python-utils/src/bridge/musig_signer.rs
```diff
@@ -0,0 +1,195 @@
...
+use secp256k1::Parity;
+
+// Removed complex type dependencies - working with basic bitcoin types directly
+use crate::error::Error;
+
+/// MuSig2 signer for bridge transactions
+pub(crate) struct MusigSigner {
+    #[allow(dead_code)]
+    secp: Secp256k1<All>,
+}
```
storopoli: Just use the `SECP256K1` global const?

crates/util/python-utils/src/bridge/musig_signer.rs
```diff
@@ -0,0 +1,195 @@
...
+
+impl MusigSigner {
+    /// Create a new MuSig signer instance
+    pub(crate) fn new() -> Self {
+        Self {
+            secp: Secp256k1::new(),
```
storopoli: Just use the `SECP256K1` global const?

crates/util/python-utils/src/bridge/musig_signer.rs
```diff
@@ -0,0 +1,195 @@
...
+    /// * `keypairs` - Vector of operator keypairs for multi-signature
+    /// * `input_index` - Index of the input to sign (usually 0 for deposit transactions)
+    ///
+    /// # Returns
+    /// * `Result<Signature, Error>` - The aggregated taproot signature
+    pub(crate) fn sign_deposit_psbt(
+        &self,
+        psbt: &Psbt,
+        prevouts: &[TxOut],
+        tweak: Option<bdk_wallet::bitcoin::TapNodeHash>,
```
storopoli: full import at the top

crates/util/python-utils/src/bridge/musig_signer.rs
```diff
@@ -0,0 +1,195 @@
...
+    /// Create a new MuSig signer instance
+    pub(crate) fn new() -> Self {
+        Self {
+            secp: Secp256k1::new(),
+        }
+    }
+
+    /// Creates an aggregated signature for the deposit transaction PSBT using MuSig2
+    ///
+    /// # Arguments
```
storopoli, line 28:
```suggestion
    /// # Arguments
    ///
```

crates/util/python-utils/src/bridge/musig_signer.rs
```diff
@@ -0,0 +1,195 @@
...
+    /// * `input_index` - Index of the input to sign (usually 0 for deposit transactions)
+    ///
+    /// # Returns
+    /// * `Result<Signature, Error>` - The aggregated taproot signature
+    pub(crate) fn sign_deposit_psbt(
+        &self,
+        psbt: &Psbt,
+        prevouts: &[TxOut],
+        tweak: Option<bdk_wallet::bitcoin::TapNodeHash>,
+        keypairs: Vec<Keypair>,
```
storopoli: You should use `EvenSecretKey`: https://github.com/alpenlabs/alpen/blob/0293f84c23ce579dfd3d522b13d8cc2a0f185933/crates/primitives/src/crypto.rs#L55-L57

crates/util/python-utils/src/bridge/musig_signer.rs
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+        let mut full_pubkeys = Vec::new();
+        for kp in &keypairs {
+            // Convert to x-only and then to full public key with even parity
+            let (xonly, _parity) = XOnlyPublicKey::from_keypair(kp);
+            // Always use even parity for the full public key to ensure consistency
+            let even_full = xonly.public_key(Parity::Even);
+            full_pubkeys.push(even_full);
+        }
```
storopoli: `EvenSecretKey`

crates/util/python-utils/src/bridge/musig_signer.rs
```diff
@@ -0,0 +1,195 @@
...
+                .with_seckey(keypair.secret_key())
+                .with_message(sighash.as_byte_array());
+
+            // Generate a proper nonce seed
+            let mut nonce_seed = [0u8; 32];
+            thread_rng().fill_bytes(&mut nonce_seed);
+
+            let first_round: FirstRound =
+                FirstRound::new(ctx.clone(), nonce_seed, signer_index, spices)
+                    .map_err(|e| Error::Musig(format!("First round creation failed: {}", e)))?;
```
storopoli:
```suggestion
                    .map_err(|e| Error::Musig(format!("First round creation failed: {e}")))?;
```

Do this for all the other ones in this file?

crates/util/python-utils/src/bridge/musig_signer.rs
```diff
@@ -0,0 +1,195 @@
...
+
+        // Create first rounds with proper signer indices
+        for (signer_index, keypair) in keypairs.iter().enumerate() {
+            let spices = SecNonceSpices::new()
+                .with_seckey(keypair.secret_key())
+                .with_message(sighash.as_byte_array());
+
+            // Generate a proper nonce seed
+            let mut nonce_seed = [0u8; 32];
+            thread_rng().fill_bytes(&mut nonce_seed);
```
storopoli: This is bad, just use the [`SecNonceBuilder`](https://docs.rs/musig2/0.2.4/musig2/struct.SecNonceBuilder.html#method.new):

```rs
// Sample the seed automatically
let secnonce = musig2::SecNonceBuilder::new(&mut rand::rngs::OsRng)
    .build();
```

crates/util/python-utils/src/bridge/musig_signer.rs
```diff
@@ -0,0 +1,195 @@
...
+        let sighash = self.create_sighash(&psbt.unsigned_tx, prevouts, input_index)?;
+
+        // First round: Generate nonces
+        let mut first_rounds: Vec<FirstRound> = Vec::new();
+        let mut pub_nonces: Vec<PubNonce> = Vec::new();
+
+        // Create first rounds with proper signer indices
+        for (signer_index, keypair) in keypairs.iter().enumerate() {
+            let spices = SecNonceSpices::new()
+                .with_seckey(keypair.secret_key())
```
storopoli: `EvenSecretKey`

crates/util/python-utils/src/bridge/musig_signer.rs
```diff
@@ -0,0 +1,195 @@
...
+            sighash_type: TapSighashType::Default,
+        };
+
+        Ok(taproot_sig)
+    }
+
+    /// Creates the sighash for the transaction input
+    fn create_sighash(
+        &self,
+        tx: &bdk_wallet::bitcoin::Transaction,
```
storopoli: full import at the top

crates/util/python-utils/src/bridge/musig_signer.rs
```diff
@@ -0,0 +1,195 @@
...
+        prevouts: &[TxOut],
+        input_index: usize,
+    ) -> Result<bdk_wallet::bitcoin::TapSighash, Error> {
```
storopoli: full import at the top

crates/util/python-utils/src/bridge/types.rs
```diff
@@ -0,0 +1,166 @@
...
+    ) -> Self {
+        Self {
+            tag,
+            operator_idx,
+            deposit_idx,
+            deposit_txid,
+        }
+    }
+
+    pub(crate) fn op_return_data(&self) -> Vec<u8> {
```
storopoli: `make_buf`

crates/util/python-utils/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "maturin"
 
 [project]
 name = "strata_utils"
-version = "0.2.0-pre.1"
+version = "0.2.0-pre.11"
```
storopoli: Wow are we doing JavaScript addition here?
"1" + 1 = 11?

storopoli (changes requested):
I hate how Claude Code does not import stuff that it uses. Can you add some instruction in your `CLAUDE.md` so that it does that. Reviewing these sloppy inlined external definitions is annoying.

Most important I think we're going the wrong way here.
I think we should deprecate `strata-python-utils` since there is a LOT of code duplication with `alpen-cli` and we need to maintain a Python <> Rust FFI. Which is not trivial.

Do you think is possible to do this PR by using only `alpen-cli` (you can add `debug` commands) and calling it as a binary in `flexitest` like we do with other binaries in this repo?

crates/rpc/api/src/lib.rs
delbonis: I'm in favor of "cur", fwiw, it's a pretty universal term.

crates/rpc/types/src/types.rs
```diff
@@ -363,6 +364,21 @@ impl From<CheckpointEntry> for RpcCheckpointInfo {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+/// Deposit entry for RPC corresponding to [`WithdrawalIntent`]
```
delbonis: Attribute ordering.

crates/util/python-utils/src/bridge/types.rs
```diff
@@ -0,0 +1,166 @@
...
+    keys: Vec<Keypair>,
+}
+
+#[pymethods]
+impl OperatorKeys {
+    #[new]
+    pub fn new(keys: Vec<String>) -> PyResult<Self> {
+        Ok(Self {
+            keys: parse_operator_keys(&keys)?,
+        })
```
delbonis: This should take a Python `bytes` or even an already parsed `Keypair` type.  Keys should be parsed from strings *as early as possible*.  The Ethereum JS ecosystem is really bad about this.

crates/util/python-utils/src/bridge/types.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    #[getter]
+    fn total_amount(&self) -> u64 {
+        self.total_amount.to_sat()
+    }
+
+    #[getter]
+    fn x_only_public_key(&self) -> String {
+        self.x_only_public_key.to_string()
+    }
```
delbonis: Are these getters necessary if the fields are pub?  If yes, then the fields should not be pub.

delbonis (changes requested):
Also generally concur with Jose's comments.

crates/rpc/api/src/lib.rs
storopoli: Then we should normalize everything to either be `getCur` or `getCurrent`. We should pick one.

crates/util/python-utils/src/bridge/musig_signer.rs
voidash: https://docs.rs/musig2/0.1.0/src/musig2/rounds.rs.html#109
FirstRound creates SecNonce itself. Can you elaborate further on what you mean

storopoli: Why you need another nonce?

voidash: [SecNonce](https://docs.rs/musig2/0.1.0/musig2/struct.SecNonce.html#method.build) requires [NonceSeed](https://docs.rs/musig2/0.1.0/musig2/struct.NonceSeed.html#method.from-1). So it's just a nonce seed and not a nonce. We just have to provide random bytes [u8;32]

crates/util/python-utils/src/constants.rs
```diff
@@ -32,6 +27,17 @@ pub(crate) static DESCRIPTOR: LazyLock<&'static str> =
...
+pub(crate) const EE_ADDRESS_LEN: u8 = 20;
+
+/// For getting private key for the corresponding xpriv.
+#[allow(dead_code)]
+pub(crate) const GENERAL_WALLET_KEY_PATH: [ChildNumber; 3] = [
+    ChildNumber::Hardened { index: 20_000 },
+    ChildNumber::Hardened { index: 20 },
+    ChildNumber::Hardened { index: 101 },
+];
+
```
storopoli: This will not grow old nicely. We're duplicating code.
Can you please import the `strata-primitives`: https://github.com/alpenlabs/alpen/blob/ababb1f7d79993bdcb154223dc4240daef953bac/crates/primitives/src/constants.rs#L26-L82

crates/util/python-utils/src/constants.rs
```diff
@@ -32,6 +27,17 @@ pub(crate) static DESCRIPTOR: LazyLock<&'static str> =
 pub(crate) static CHANGE_DESCRIPTOR: LazyLock<&'static str> =
     LazyLock::new(|| Box::leak(format!("tr({XPRIV}/86'/1'/0'/1/*)").into_boxed_str()));
 
+/// Corresponds to address length of Execution environment.
+pub(crate) const EE_ADDRESS_LEN: u8 = 20;
```
storopoli: We should also define this in the `strata-primitives` `constants` module similar to this: https://github.com/alpenlabs/alpen/blob/ababb1f7d79993bdcb154223dc4240daef953bac/crates/primitives/src/constants.rs#L20-L22

crates/util/python-utils/src/taproot.rs
```diff
@@ -64,7 +64,8 @@ pub(crate) fn taproot_wallet() -> Result<Wallet, Error> {
         .map_err(|_| Error::Wallet))?
 }
 
-/// The bridge wallet used to get the recovery path of the deposit request transaction (DRT).
+/// The bridge wallet used for Withdrawal Fulfillment transaction.
+#[allow(dead_code)]
```
storopoli: Add a `reason = "foo"` or use `#[expect(dead_code)]`.

crates/util/python-utils/src/utils.rs
```diff
@@ -1,9 +1,61 @@
...
+) -> Result<(DescriptorTemplateOut, TapNodeHash), Error> {
+    let recovery_xonly_pubkey = recovery_address.extract_p2tr_pubkey()?;
+
+    let desc = bdk_wallet::descriptor!(
+        tr(UNSPENDABLE_PUBLIC_KEY, {
+            pk(bridge_pubkey),
+            and_v(v:pk(recovery_xonly_pubkey),older(RECOVER_DELAY))
+        })
+    )
+    .expect("valid descriptor");
```
storopoli: Do we still need this if we're moving to use `alpen-cli` for DRTs?

storopoli (changes requested):

functional-tests/factory/service.py
```diff
@@ -0,0 +1,107 @@
...
+from typing import Optional
+
+import flexitest
+
+
+class DisposableService(flexitest.Service):
+    """
+    Runs a single command and exits. Not long-lived; no start()/stop() lifecycle.
+    Stores only the result of the last run.
+    """
```
delbonis: I would say that this is *not* a service at all.  And actually calling it "disposable" with the docstring makes it appear more general than it actually is, which is designed more specifically for invoking subcommands.  I am not sure what a more appropriate name for this would be.

The factory concept referenced elsewhere still makes some sense despite it not being a service factory.

functional-tests/factory/service.py
```diff
@@ -0,0 +1,107 @@
...
+        if self._last_result is None:
+            return None
+        return self._last_result.stdout
+
+    # start and stop methods aren't to be used as this is not
+    def start(self):
+        pass
+
+    def stop(self):
+        pass
```
delbonis: The fact these are stubbed off should be a indication / red flag that the abstraction is being misused.

crates/util/python-utils/src/parse.rs
```diff
@@ -1,14 +1,37 @@
...
+    error::Error,
+};
 
-/// Parses a Execution Layer address.
-pub(crate) fn parse_el_address(el_address: &str) -> Result<RethAddress, Error> {
-    let el_address = el_address
-        .parse::<RethAddress>()
-        .map_err(|_| Error::ElAddress)?;
-    Ok(el_address)
+#[allow(unused)]
```
storopoli: Either remove it, or add a `reason = "foo"`, or change to `#[expect(unused)]`.

crates/util/python-utils/src/utils.rs
```diff
@@ -1,5 +1,6 @@
-use bdk_wallet::bitcoin::{Address, XOnlyPublicKey};
+use bdk_wallet::bitcoin::Address;
 use pyo3::prelude::*;
+use secp256k1::XOnlyPublicKey;
```
storopoli: Why this diff?

storopoli (dismissed):
ACK c047f38e32f63a401dbcc03cca08fb42b5aa8e0f

storopoli (dismissed):
ACK 4c80d9263bd20e8444bceb29dba54f5823bd2473

functional-tests/envs/testenv.py
```diff
@@ -58,9 +56,29 @@ class StrataRunContext(flexitest.RunContext):
...
+        bitcoind_config = BitcoindConfig(
+            rpc_url=f"http://localhost:{btc_svc.get_prop('rpc_port')}",
+            rpc_user=btc_svc.get_prop("rpc_user"),
+            rpc_password=btc_svc.get_prop("rpc_password"),
+        )
+        # get reth endpoint
+        reth_svc = self.get_service("reth")
+        reth_rpc_http_port = reth_svc.get_prop("eth_rpc_http_port")
+        rollup_cfg = env.rollup_cfg()
+
```
delbonis: Will have to rethink how this works when we support running the client without reth integrated, but this is fine for now.

delbonis (dismissed):
Okay I think this is fairly reasonable for now.

delbonis:
Fntest CI checks are failing, triggered it to re-run the job.

storopoli (approved):
ACK e69c4f5550bb9c07adfafc62eb4e1f0a43f254e9

functional-tests/factory/alpen_client.py
```diff
@@ -6,26 +6,52 @@
 from subprocess import CalledProcessError
 from typing import Optional
 
-from factory.config import BitcoindConfig
+from envs.env_control_builder import EnvControlBuilder, ServiceNotAvailable
 
 
-class AlpenCli:
+class AlpenCli(EnvControlBuilder):
```
delbonis: `AlpenCliBuilder`?

Edit: ah I see what's happening here

functional-tests/envs/env_control_builder.py
```diff
@@ -0,0 +1,57 @@
...
+
+        self.service_requirements[service_name] = transform_lambda
+        return self
+
+    def build(self):
+        """
+        Resolve all service requirements and return transformed configs.
+
+        Returns:
+            Dictionary mapping service names to their transformed values
```
delbonis, line 38: Should specify return type.

functional-tests/factory/alpen_client.py
```diff
@@ -37,15 +63,16 @@ def __init__(
...
+bridge_pubkey = "{self.pubkey}"
+magic_bytes = "{self.magic_bytes}"
 network = "regtest"
 seed = "838d8ba290a3066abb35b663858fa839"
 """
         with open(self.config_file, "w") as f:
             f.write(config_content)
 
         assert self.check_config(), "config file path should match"
+        return self
```
delbonis: This is merging the concept of the builder with the "thing that gets built".  All of the functions that "do stuff" exist on the builder, and it's impossible to know (at the call site) if the builder was properly initialized.

Conversely, there's a `build` method on the "thing that gets built" (naturally, since it's the same object as the builder).  Is it valid to call that after it's been built?  And the `with_` methods?

The purpose of the builder pattern is to encapsulate the variety of init procedures and then package the necessary components into the actual object that gets built by the builder, so that it's only normally possible to construct an instance of the object if it was inited properly (by the builder).  In Python that's a little hard because nobody can prevent you from directly constructing objects, but we can try to make it harder.

In simpler cases, a lot of this logic could be shifted to just exist in the `AlpenCli`'s constructor.  But that might not necessarily make sense with the separation between env "declaration time", env setup time, and test run time that we have.

I realize that yes, this is Python, so it's harder to do type-driven design like we have in Rust.  But it's *because* we're in Python that it's important to be careful about how we design these concepts and not let entropy pile up.  In Rust if we conflate concepts wrongly then it won't compile, but in Python we don't have the guardrails that prevent these kinds of design mixups and prevent things from "accidentally working" that shouldn't.

Also, I think this `build` impl contradicts the docstring for `build` on the parent class.

functional-tests/factory/alpen_client.py
```diff
@@ -6,26 +6,52 @@
...
+    def build(self):
+        """Build AlpenCli instance with resolved service configs"""
+        if not self.pubkey or not self.magic_bytes or not self.datadir:
+            raise ValueError("pubkey, magic_bytes, and datadir must be set before building")
+
+        # Get resolved configs from parent
+        try:
+            resolved_configs = super().build()
+        except ServiceNotAvailable:
+            return None
```
delbonis: This isn't how OO is supposed to work.  It should call a different method instead of hacking the class hierarchy.

When a class inherits from a class and overrides a method, it's supposed to implement some more specific *flavor* of that same behavior.  It's usually possible to delegate to the `super` impl, but when you do that it's because you want to delegate back to the parent's implementation.  Perhaps you added some conditionals around that behavior but ultimately want to maintain the guts of the parent's impl.

But in this case, the parent's impl is just being used as if it's another method but one we have to call weirdly.  The overload does something completely different than what the behavior is defined to be in the parent (according to docstring).

functional-tests/envs/env_control_builder.py
```diff
@@ -0,0 +1,57 @@
...
+
+class EnvControlBuilder:
+    """
+    Provides standardized interface for service lookup, validation, and retrieval
+    so that we can know if something is injectable to the Env or not and then build it
+    if it is injectable
+    """
+
+    def __init__(self, ctx: flexitest.LiveEnv):
+        self.ctx = ctx
```
delbonis: `ctx` or `env`?

And wouldn't it make more sense to have the env be passed to `build` instead?  That way we could construct a control builder independently of a particular env and *then* use it to build instances of the objects we're constructing with an env when it's ready.

delbonis (changes requested):

delbonis:
I think the general structure is better now, but this new builder concept to be reworked.

delbonis (approved):
Ok, I think this is in an acceptable state now.

storopoli (approved):
ACK ab51052561187aa4c1fcb763f617b703b2ddde88

PR #1003, opened 2025-08-26
db(stub-l2): make del_block_data clean heights and statuses

VolodymyrBg:
Align stub del_block_data with trait/rocksdb semantics:
- remove block from blocks, status from statuses
- remove ID from heights[slot]
- return false when block not present


krsnapaudel:
This is related to STR-1632.

delbonis (approved):

PR #1004, opened 2025-08-27
Str 1686: Add test mode in alpen-cli

voidash:
## Description

Adds a test-mode cargo feature flag that disables passwords and adds a new setting that directly takes the seed value

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

https://alpenlabs.atlassian.net/browse/STR-1686

### DIFF
--- a/bin/alpen-cli/src/cmd/change_pwd.rs
+++ b/bin/alpen-cli/src/cmd/change_pwd.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use argh::FromArgs;
 use rand_core::OsRng;
 use strata_cli_common::errors::{DisplayableError, DisplayedError};
@@ -12,7 +14,7 @@ pub struct ChangePwdArgs {}
 pub async fn change_pwd(
     _args: ChangePwdArgs,
     seed: Seed,
-    persister: impl EncryptedSeedPersister,
+    persister: Arc<dyn EncryptedSeedPersister>,
 ) -> Result<(), DisplayedError> {
     let mut new_pw =
         Password::read(true).internal_error("Failed to read the password entered by user")?;
--- a/bin/alpen-cli/src/cmd/mod.rs
+++ b/bin/alpen-cli/src/cmd/mod.rs
@@ -1,6 +1,7 @@
 use argh::FromArgs;
 use backup::BackupArgs;
 use balance::BalanceArgs;
+#[cfg(not(feature = "test-mode"))]
 use change_pwd::ChangePwdArgs;
 use config::ConfigArgs;
 use deposit::DepositArgs;
@@ -8,6 +9,7 @@ use drain::DrainArgs;
 use faucet::FaucetArgs;
 use receive::ReceiveArgs;
 use recover::RecoverArgs;
+#[cfg(not(feature = "test-mode"))]
 use reset::ResetArgs;
 use scan::ScanArgs;
 use send::SendArgs;
@@ -49,7 +51,9 @@ pub enum Commands {
     Faucet(FaucetArgs),
     Send(SendArgs),
     Receive(ReceiveArgs),
+    #[cfg(not(feature = "test-mode"))]
     ChangePwd(ChangePwdArgs),
+    #[cfg(not(feature = "test-mode"))]
     Reset(ResetArgs),
     Scan(ScanArgs),
     Config(ConfigArgs),
--- a/bin/alpen-cli/src/cmd/mod.rs
+++ b/bin/alpen-cli/src/cmd/mod.rs
@@ -1,21 +1,17 @@
 use argh::FromArgs;
 use backup::BackupArgs;
 use balance::BalanceArgs;
-#[cfg(not(feature = "test-mode"))]
-use change_pwd::ChangePwdArgs;
 use config::ConfigArgs;
 use deposit::DepositArgs;
 use drain::DrainArgs;
 use faucet::FaucetArgs;
 use receive::ReceiveArgs;
 use recover::RecoverArgs;
-#[cfg(not(feature = "test-mode"))]
-use reset::ResetArgs;
 use scan::ScanArgs;
 use send::SendArgs;
 use withdraw::WithdrawArgs;
 
-use crate::cmd::debug::DebugArgs;
+use crate::cmd::{change_pwd::ChangePwdArgs, debug::DebugArgs, reset::ResetArgs};
 
 pub mod backup;
 pub mod balance;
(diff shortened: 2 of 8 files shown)


bin/alpen-cli/src/cmd/mod.rs
```diff
@@ -1,19 +1,17 @@
...
 use faucet::FaucetArgs;
 use receive::ReceiveArgs;
 use recover::RecoverArgs;
-use reset::ResetArgs;
 use scan::ScanArgs;
 use send::SendArgs;
 use withdraw::WithdrawArgs;
 
-use crate::cmd::debug::DebugArgs;
+use crate::cmd::{change_pwd::ChangePwdArgs, debug::DebugArgs, reset::ResetArgs};
```
storopoli: This change is a no-op

bin/alpen-cli/src/cmd/change_pwd.rs
```diff
@@ -12,7 +14,7 @@ pub struct ChangePwdArgs {}
 pub async fn change_pwd(
     _args: ChangePwdArgs,
     seed: Seed,
-    persister: impl EncryptedSeedPersister,
+    persister: Arc<dyn EncryptedSeedPersister>,
```
storopoli: I don't think we need to `Arc` this.
You are changing compile-time monomorphization (`impl Trait`) for a run-time vtable (`dyn Trait`).

bin/alpen-cli/src/cmd/reset.rs
```diff
@@ -17,8 +19,8 @@ pub struct ResetArgs {
 
 pub async fn reset(
     args: ResetArgs,
-    persister: impl EncryptedSeedPersister,
-    settings: Settings,
+    persister: Arc<dyn EncryptedSeedPersister>,
+    settings: &Settings,
```
storopoli: same as above.
Although I like to move `Settings` to a read-only ref `&Settings`.

bin/alpen-cli/src/seed/seed_provider.rs
```diff
@@ -0,0 +1,116 @@
...
+#[cfg(feature = "test-mode")]
+use zeroize::Zeroizing;
+
+use super::{LoadOrCreateErr, Seed};
+use crate::{
+    cmd::{change_pwd::ChangePwdArgs, reset::ResetArgs},
+    settings::Settings,
+};
+
+#[async_trait(?Send)]
```
storopoli: This is acceptable, we don't plan to spawn different threads in the alpen-cli in `test` mode that will have it's own copy of `SecretStore`, since most of the calls will be single-threaded by a CLI binary call.

bin/alpen-cli/src/seed/seed_provider.rs
```diff
@@ -0,0 +1,116 @@
...
+use strata_cli_common::errors::DisplayedError;
+use terrors::OneOf;
+
+use super::{LoadOrCreateErr, Seed};
+use crate::{
+    cmd::{change_pwd::ChangePwdArgs, reset::ResetArgs},
+    settings::Settings,
+};
+
+#[async_trait(?Send)]
```
Zk2u: Async traits were stabilized multiple releases ago, this is an unnecessary dep

Zk2u: if it's moaning about mixing sync and async then make the async ones be sync and produce `impl Future<Target = ...>` instead

bin/alpen-cli/src/seed/seed_provider.rs
```diff
@@ -0,0 +1,116 @@
...
+            Ok(self.seed.clone())
+        }
+
+        async fn reset(
+            &self,
+            _args: ResetArgs,
+            _settings: &Settings,
+        ) -> Result<(), DisplayedError> {
+            eprintln!("Reset is disabled for test mode");
+            std::process::exit(1);
```
Zk2u: these should be just a single panic instead

bin/alpen-cli/src/seed/seed_provider.rs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        async fn change_pwd(
+            &self,
+            _args: ChangePwdArgs,
+            _seed: Seed,
+        ) -> Result<(), DisplayedError> {
+            eprintln!("change password is disabled for test mode");
+            std::process::exit(1);
```
Zk2u: panic

bin/alpen-cli/src/seed.rs
```diff
@@ -158,7 +159,7 @@ impl EncryptedSeed {
 }
 
 pub fn load_or_create(
-    persister: &impl EncryptedSeedPersister,
+    persister: Arc<dyn EncryptedSeedPersister>,
```
Zk2u: Why was this changed to an arc?

bin/alpen-cli/src/seed.rs
```diff
@@ -257,7 +258,7 @@ type LoadOrCreateErr = (
     IncorrectPassword,
 );
 
-pub trait EncryptedSeedPersister {
+pub trait EncryptedSeedPersister: Debug {
     fn save(&self, seed: &EncryptedSeed) -> Result<(), PersisterErr>;
```
Zk2u: Why was debug supertrait added?

bin/alpen-cli/src/seed/seed_provider.rs
voidash: That won't compile. However we can `Pin<Box<dyn Future<Ouput = ... >` . but that is what async trait does

bin/alpen-cli/src/seed.rs
voidash: Because Settings struct is Debug and it has secret_store: `Arc<dyn SecretStore>` . The trait SecretStore has `UserSeedProvider` struct which stores `EncryptedSeedProvider` so i had to add debug supertrait.

storopoli (approved):
ACK 3469fe5cacbf6cf111990fa8b18ba2cd0e1dd5bf
Waiting on Azz's approval before merging.

bin/alpen-cli/src/seed.rs
Zk2u: Please make it `Arc<dyn SecretStore + Debug>` instead to depend on `Debug` where we use it instead of adding it as a dep of `EncryptedSeedPersister`

bin/alpen-cli/src/seed/seed_provider.rs
Zk2u: Is it an issue of needing `Send`?
https://github.com/alpenlabs/strata-bridge/blob/6173345084f61f094cab1f7222cb05dc1c1200d7/crates/secret-service-proto/src/v2/traits.rs#L72-L91

Idk what the issue was, but I think this that won't solve it by itself because we want to dyn it right? Otherwise you can just do the `Pin<Box<>>` thing cos seems pointless adding another dep to do that for us

PR #1013, opened 2025-09-01
ASM Administration Subprotocol

prajwolrg:
## Description

Implements the ASM Administration Subprotocol.

Right now the implementation deliberately avoids the transaction parsing logic, because there we haven't finalized if we want to go with Schnorr Musig2 signature or the BLS signatures.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

I used Claude Code to add docstring and generate tests. 

## Checklist

      in the body of this PR.

## Related Issues

Supersedes #864


crates/primitives/src/roles.rs
```diff
@@ -0,0 +1,57 @@
...
+    PartialEq,
+    PartialOrd,
+    Ord,
+    Hash,
+    BorshSerialize,
+    BorshDeserialize,
+    Arbitrary,
+)]
+#[borsh(use_discriminant = false)]
+pub enum Role {
```
delbonis: What is going on with this derive order?  especially `Eq, PartialEq, PartialOrd, Ord` being mixed weird?

The guideline is to put the more the "essential" properties first, going to more specific ones, then the non-std ones in alphabetical order.  So *something* like: `Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Arbitrary, BorshDeserialize, BorshSerialize`.

delbonis (approved):
This is a good start.  We'll probably want to do another pass later after all the initial versions of all the main subprotocols are written to make them follow the same module structure, but if it works then that's fine.

crates/primitives/src/roles.rs
prajwolrg: Done in [01e22f1](https://github.com/alpenlabs/alpen/pull/1013/commits/01e22f1fe2f59147c1ecfefdd83ea6893b1cd2c6).

crates/crypto/src/multisig/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+pub mod config;
+pub mod errors;
+pub mod msg;
+pub mod vote;
+
+use strata_primitives::buf::{Buf32, Buf64};
+
+use crate::multisig::{errors::VoteValidationError, msg::MultisigPayload};
+
+pub type PubKey = Buf32;
```
storopoli, line 11: Can we have more documentation on this?
Are these X-only public keys? (Probably yes, otherwise they would be `Buf33` 😂).
What type of keys are these? I'd imagine BIP-340 Schnorr keys right?
Should the implementer be aware that it is not guaranteed that these are valid X-Only PKs (they can be odd PKs) or should the implementer guarantee that he/she will always be responsible for guaranteeing that they are valid only X-only PKs (always even, negating the sk if needed).

I don't see that constrained in the type, `Buf32`.
If you want you could use the `XOnlyPk` that is guaranteed to be a valid X-only PK and is `Borsh`-friendly: https://github.com/alpenlabs/alpen/blob/da69785a3c57baa1295b835f146085541dfb5bce/crates/primitives/src/l1/btc.rs#L717-L732

crates/crypto/src/multisig/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+pub type Signature = Buf64;
+
+// FIXME: handle
+pub fn aggregate_pubkeys(_keys: &[PubKey]) -> Result<PubKey, VoteValidationError> {
+    Ok(PubKey::default())
+}
```
storopoli, line 17: Use `musig2`, don't reinvent the wheel :)

storopoli (dismissed):
Some concerns over the public keys

storopoli:
Let's fix the conflicts and I can also approve this since a lot of my concerns will be tackled in #1027.

crates/asm/subprotocols/administration/src/authority.rs
```diff
@@ -0,0 +1,96 @@
...
+/// Manages multisignature operations for a given role and key set, with replay protection via a
+/// nonce.
+#[derive(Clone, Debug, Eq, PartialEq, Arbitrary, BorshDeserialize, BorshSerialize)]
+pub struct MultisigAuthority {
+    /// The role of this multisignature authority.
+    role: Role,
+    /// The public keys of all grant-holders authorized to sign.
+    config: MultisigConfig,
+    /// Sequence number for the multisig configuration. It increases on each valid action.
+    /// This is used to prevent replay attacks
```
barakshani:
```suggestion
    /// This is used to prevent replay attacks.

File: crates/asm/subprotocols/administration/src/authority.rs [hunk 5]
```diff
@@ -0,0 +1,96 @@
...
+                    .get(i as usize)
+                    .cloned()
+                    .ok_or(VoteValidationError::AggregationError)
+            })
+            .collect::<Result<_, _>>()?;
+
+        // 2. Aggregate those public keys into one.
+        let aggregated_key = aggregate_pubkeys(&signer_keys)?;
+
+        // 3. Compute the msg from the UpdateAction
```
barakshani:
```suggestion
        // 3. Compute the msg from the UpdateAction.

File: crates/asm/subprotocols/administration/src/authority.rs [hunk 6]
```diff
@@ -0,0 +1,96 @@
...
+        let msg_hash = compute_borsh_hash(action);
+        let payload = MultisigPayload::new(msg_hash, self.seqno);
+
+        // 4. Verify the aggregated signature against the aggregated pubkey
```
barakshani:
```suggestion
        // 4. Verify the aggregated signature against the aggregated pubkey.

File: crates/asm/subprotocols/administration/src/authority.rs [hunk 7]
```diff
@@ -0,0 +1,96 @@
...
+
+        Ok(())
+    }
+
+    /// Increments the nonce.
+    pub fn increment_seqno(&mut self) {
+        self.seqno += 1;
+    }
+
+    /// Get the current sequence number (for testing)
```
barakshani:
```suggestion
    /// Get the current sequence number (for testing).

File: crates/asm/subprotocols/administration/src/authority.rs [hunk 8]
```diff
@@ -0,0 +1,96 @@
...
+    /// Steps:
+    /// 1. Map voter indices to pubkeys (error if out of bounds).
+    /// 2. Aggregate pubkeys and compute payload hash.
+    /// 3. Verify aggregated signature.
+    pub fn validate_action(
+        &self,
+        action: &MultisigAction,
+        vote: &AggregatedVote,
+    ) -> Result<(), VoteValidationError> {
+        // 1. Collect each public key by index; error if out of bounds.
```
barakshani, line 58: how do you define out of bound here?
do we check that the map is onto? that is, that we use all pubkeys in the `config.keys()`?

crates/asm/subprotocols/administration/src/authority.rs
```diff
@@ -0,0 +1,96 @@
...
+                self.config
+                    .keys()
+                    .get(i as usize)
+                    .cloned()
+                    .ok_or(VoteValidationError::AggregationError)
+            })
+            .collect::<Result<_, _>>()?;
+
+        // 2. Aggregate those public keys into one.
+        let aggregated_key = aggregate_pubkeys(&signer_keys)?;
```
barakshani, line 72: is this not part of `vote.signature()`?

crates/asm/subprotocols/administration/src/config.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// MultisigConfig for [StrataAdministrator](Role::StrataAdministrator).
+    pub strata_administrator: MultisigConfig,
+
+    /// MultisigConfig for [StrataSequencerManager](Role::StrataSequencerManager).
+    pub strata_sequencer_manager: MultisigConfig,
+
+    /// The confirmation depth (CD) setting: after an update transaction receives this many
+    /// confirmations, the update is enacted automatically. During this confirmation period,
+    /// the update can still be cancelled by submitting a cancel transaction.
+    pub confirmation_depth: u32,
```
barakshani, line 24: so you decided to have a global confirmation depth?

crates/asm/subprotocols/administration/src/state.rs
```diff
@@ -0,0 +1,280 @@
...
+    /// List of configurations for multisignature authorities.
+    /// Each entry specifies who the signers are and how many signatures
+    /// are required to approve an action.
+    authorities: Vec<MultisigAuthority>,
+
+    /// Actions that can still be cancelled by CancelTx while waiting
+    /// for their block countdown to complete.
+    queued: Vec<QueuedUpdate>,
+
+    /// UpdateId for the next update
```
barakshani:
```suggestion
    /// UpdateId for the next update.

File: crates/asm/subprotocols/administration/src/state.rs [hunk 12]
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    /// Remove a queued update by swapping it out.
+    pub fn remove_queued(&mut self, id: &UpdateId) {
+        if let Some(i) = self.queued.iter().position(|u| u.id() == id) {
+            self.queued.swap_remove(i);
+        }
+    }
+
+    /// Get the next global update id
```
barakshani:
```suggestion
    /// Get the next global update id.

File: crates/asm/subprotocols/administration/src/state.rs [hunk 13]
```diff
@@ -0,0 +1,280 @@
...
+    pub fn next_update_id(&self) -> UpdateId {
+        self.next_update_id
+    }
+
+    /// Confirmation depth
```
barakshani:
```suggestion
    /// Confirmation depth.

File: crates/asm/subprotocols/administration/src/state.rs [hunk 14]
```diff
@@ -0,0 +1,280 @@
...
+    pub fn confirmation_depth(&self) -> u32 {
+        self.confirmation_depth
+    }
+
+    /// Increment the next global update id
```
barakshani:
```suggestion
    /// Increment the next global update id.

File: crates/asm/subprotocols/administration/src/state.rs [hunk 15]
```diff
@@ -0,0 +1,280 @@
...
+    /// from `queued` into `committed`.
+    pub fn process_queued(&mut self, current_height: u64) -> Vec<QueuedUpdate> {
+        let (ready, rest): (Vec<_>, Vec<_>) = std::mem::take(&mut self.queued)
+            .into_iter()
+            .partition(|u| u.activation_height() <= current_height);
+        self.queued = rest;
+        ready
+    }
+
+    /// Get a reference to the queued updates (for testing)
```
barakshani:
```suggestion
    /// Get a reference to the queued updates (for testing).

File: crates/asm/subprotocols/administration/src/state.rs [hunk 16]
```diff
@@ -0,0 +1,280 @@
...
+        self.confirmation_depth
+    }
+
+    /// Increment the next global update id
+    pub fn increment_next_update_id(&mut self) {
+        self.next_update_id += 1;
+    }
+
+    /// Process all queued updates and move any whose `activation_height` equals `current_height`
+    /// from `queued` into `committed`.
```
barakshani: what does this mean in practice? you seem to just remove those items from the queue

crates/asm/subprotocols/administration/src/handler.rs
```diff
@@ -0,0 +1,396 @@
...
+    let actions_to_enact = state.process_queued(current_height);
+
+    for action in actions_to_enact {
+        match action.action() {
+            UpdateAction::Multisig(update) => {
+                state.apply_multisig_update(update.role(), update.config());
+            }
+            UpdateAction::VerifyingKey(update) => match update.kind() {
+                ProofType::Asm => {
+                    // Emit Log
```
barakshani: shall we add explicit TODOs here?

crates/asm/subprotocols/administration/src/state.rs
barakshani: also just double checking, right now the function that call this does not have an error path, however if we were to revert that function, will this "mark as committed" be reverted as well? basically, is this a future?

crates/asm/subprotocols/administration/src/handler.rs
barakshani: even open a tech-debt ticket

crates/asm/subprotocols/administration/src/handler.rs
```diff
@@ -0,0 +1,396 @@
...
+            assert!(state.find_queued(&initial_next_id).is_none());
+        }
+    }
+
+    /// Test that cancel actions properly remove queued updates:
+    /// - First queue 5 update actions
+    /// - Then cancel each one individually
+    /// - Verify sequence numbers increment, queue shrinks, and updates are removed
+    #[test]
+    fn test_strata_administrator_cancel_action() {
```
barakshani, line 313: how about a test that cancel updates not in a sequential order? do you see value in adding this?

crates/asm/subprotocols/administration/src/handler.rs
```diff
@@ -0,0 +1,396 @@
...
+            cancel_action,
+            vote.clone(),
+            current_height,
+            &mut relayer,
+        );
+        assert!(matches!(res, Err(AdministrationError::UnknownAction(_))));
+    }
+
+    /// Test that attempting to cancel a non-existent action returns an error:
+    /// - Generate a random cancel action for an ID that doesn't exist
```
barakshani: this description should be modified

crates/asm/subprotocols/administration/src/state.rs
```diff
@@ -0,0 +1,280 @@
...
+/// multisignature authorities and any actions still pending execution.
+#[derive(Clone, Debug, Eq, PartialEq, BorshDeserialize, BorshSerialize)]
+pub struct AdministrationSubprotoState {
+    /// List of configurations for multisignature authorities.
+    /// Each entry specifies who the signers are and how many signatures
+    /// are required to approve an action.
+    authorities: Vec<MultisigAuthority>,
+
+    /// Actions that can still be cancelled by CancelTx while waiting
+    /// for their block countdown to complete.
```
barakshani: while technically true, isn't that a very narrowed view on the queue? the queue is basically a list of updates that is _waiting_ to be executed (and can be cancelled until then).

crates/asm/subprotocols/administration/src/state.rs
```diff
@@ -0,0 +1,280 @@
...
+    pub fn process_queued(&mut self, current_height: u64) -> Vec<QueuedUpdate> {
+        let (ready, rest): (Vec<_>, Vec<_>) = std::mem::take(&mut self.queued)
+            .into_iter()
+            .partition(|u| u.activation_height() <= current_height);
+        self.queued = rest;
+        ready
+    }
+
+    /// Get a reference to the queued updates (for testing)
+    #[cfg(test)]
```
barakshani: ?

crates/asm/subprotocols/administration/src/subprotocol.rs
```diff
@@ -0,0 +1,59 @@
...
+        handle_pending_updates(state, relayer, current_height);
+
+        for tx in txs {
+            if let Ok((action, vote)) = parse_tx_multisig_action_and_vote(tx) {
+                let _ = handle_action(state, action, vote, current_height, relayer);
+            }
+        }
+    }
+
+    fn process_msgs(
```
barakshani, line 85: for incoming messages to the admin subprotocol?

crates/asm/txs/administration/src/actions/cancel.rs
```diff
@@ -0,0 +1,33 @@
+use arbitrary::Arbitrary;
+use borsh::{BorshDeserialize, BorshSerialize};
+use strata_asm_common::TxInputRef;
+
+use crate::{actions::UpdateId, constants::CANCEL_TX_TYPE, error::AdministrationTxParseError};
+
+#[derive(Clone, Debug, Eq, PartialEq, Arbitrary, BorshDeserialize, BorshSerialize)]
+pub struct CancelAction {
+    /// ID of the update that needs to be cancelled
```
barakshani:
```suggestion
    /// ID of the update that needs to be cancelled.

File: crates/asm/txs/administration/src/parser.rs [hunk 24]
```diff
@@ -0,0 +1,49 @@
...
+
+        _ => Err(AdministrationTxParseError::UnknownTxType)?,
+    };
+    Ok((action, vote))
+}
+
+pub fn parse_aggregated_vote(
+    _tx: &TxInputRef<'_>,
+) -> Result<AggregatedVote, AdministrationTxParseError> {
+    Ok(AggregatedVote::new(vec![], Signature::default()))
```
barakshani, line 53: placeholder?

crates/asm/txs/administration/src/actions/cancel.rs
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    pub fn target_id(&self) -> &UpdateId {
+        &self.target_id
+    }
+}
+
+impl CancelAction {
+    /// Extracts a CancelAction from a transaction input.
+    /// This is a placeholder function and should be replaced with actual logic.
```
barakshani: what would be the actual logic?

crates/asm/txs/administration/src/actions/updates/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+use arbitrary::Arbitrary;
+use borsh::{BorshDeserialize, BorshSerialize};
+use strata_primitives::roles::Role;
+
+use crate::actions::updates::{
+    multisig::MultisigUpdate, operator::OperatorSetUpdate, seq::SequencerUpdate,
+    vk::VerifyingKeyUpdate,
+};
+
+/// An action that updates some part of the ASM
```
barakshani:
```suggestion
/// An action that updates some part of the ASM.

File: crates/asm/txs/administration/src/actions/mod.rs [hunk 27]
```diff
@@ -0,0 +1,19 @@
...
+
+pub use cancel::CancelAction;
+pub use updates::UpdateAction;
+
+pub type UpdateId = u32;
+
+/// A high‐level multisig operation that participants can propose.
+#[derive(Clone, Debug, Eq, PartialEq, Arbitrary, BorshDeserialize, BorshSerialize)]
+pub enum MultisigAction {
+    /// Cancel a pending action
```
barakshani:
```suggestion
    /// Cancel a pending action.

File: crates/asm/txs/administration/src/actions/mod.rs [hunk 28]
```diff
@@ -0,0 +1,19 @@
...
+    Cancel(CancelAction),
+    /// Propose an update
```
barakshani:
```suggestion
    /// Propose an update.

File: crates/asm/txs/administration/src/actions/cancel.rs (hunk 25 above)
Comment (barakshani) [reply crates/asm/txs/administration/src/actions/cancel.rs]: add TODO?

File: crates/asm/subprotocols/administration/src/authority.rs (hunk 8 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/administration/src/authority.rs:58]: Yes we need to check the bound. But that has been implemented as part of https://github.com/alpenlabs/alpen/pull/1027.

File: crates/asm/subprotocols/administration/src/authority.rs (hunk 9 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/administration/src/authority.rs:72]: No, the aggregated keys is recomputed from the indices.

File: crates/asm/subprotocols/administration/src/config.rs (hunk 10 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/administration/src/config.rs:24]: Yes.

File: crates/asm/subprotocols/administration/src/state.rs (hunk 16 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/administration/src/state.rs]: Nice catch. Previously there was separate `committed_updates`, but now we just remove it from the queue.

File: crates/asm/subprotocols/administration/src/handler.rs (hunk 18 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/administration/src/handler.rs:313]: Sure. I can modify an existing test to support that.

File: crates/asm/subprotocols/administration/src/handler.rs (hunk 19 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/administration/src/handler.rs]: nice catch. thanks!

File: crates/asm/subprotocols/administration/src/subprotocol.rs (hunk 22 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/administration/src/subprotocol.rs:85]: Yes.

File: crates/asm/txs/administration/src/parser.rs (hunk 24 above)
Comment (prajwolrg) [reply crates/asm/txs/administration/src/parser.rs:53]: yes. added a fixme.

File: crates/asm/txs/administration/src/actions/cancel.rs (hunk 25 above)
Comment (prajwolrg) [reply crates/asm/txs/administration/src/actions/cancel.rs]: The actual logic will be parsing the transaction payload.

File: crates/asm/subprotocols/administration/src/state.rs (hunk 20 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/administration/src/state.rs]: Updated the docstring.

File: crates/asm/txs/administration/src/actions/updates/operator.rs [hunk 29]
```diff
@@ -0,0 +1,48 @@
...
+
+use crate::error::AdministrationTxParseError;
+
+/// An update to the Bridge Operator Set:
+/// - removes the specified `old_members`
+/// - adds the specified `new_members`
+#[derive(Clone, Debug, Eq, PartialEq, Arbitrary, BorshDeserialize, BorshSerialize)]
+pub struct OperatorSetUpdate {
+    new_members: Vec<Buf32>,
+    old_members: Vec<Buf32>,
```
barakshani: "old" is a bit ambiguous here: maybe change to "add_members" and "remove_members"?

crates/asm/txs/administration/src/actions/updates/seq.rs
```diff
@@ -0,0 +1,37 @@
+use arbitrary::Arbitrary;
+use borsh::{BorshDeserialize, BorshSerialize};
+use strata_asm_common::TxInputRef;
+use strata_primitives::buf::Buf32;
+
+use crate::error::AdministrationTxParseError;
+
+/// An update to the public key of the sequencer
```
barakshani:
```suggestion
/// An update to the public key of the sequencer.

File: crates/asm/txs/administration/src/constants.rs [hunk 31]
```diff
@@ -0,0 +1,22 @@
...
+/// Transaction type that proposes an update to the on-chain multisignature configuration.
+pub const MULTISIG_CONFIG_UPDATE_TX_TYPE: TxType = 10;
+
+/// Transaction type that proposes an update to the set of authorized operators.
+pub const OPERATOR_UPDATE_TX_TYPE: TxType = 11;
+
+/// Transaction type that proposes an update to the sequencer configuration.
+pub const SEQUENCER_UPDATE_TX_TYPE: TxType = 12;
+
+/// Transaction type that proposes an update to the verifying key used by the protocol.
```
barakshani: what protocol? eventually this will allow change 3 keys, no? EE, OL, ASM

crates/crypto/src/multisig/config.rs
```diff
@@ -0,0 +1,259 @@
...
+}
+
+impl MultisigConfig {
+    /// Create a new config.
+    ///
+    /// # Errors
+    ///
+    /// Returns `MultisigConfigError` if:
+    /// - `EmptyKeys`: The keys list is empty
+    /// - `InvalidThreshold`: The threshold is not greater than half the number of keys or exceeds
```
barakshani: why setting this minimum bar?

crates/crypto/src/multisig/config.rs
```diff
@@ -0,0 +1,259 @@
...
+        &self,
+        update: &MultisigConfigUpdate,
+    ) -> Result<(), MultisigConfigError> {
+        // Ensure no duplicate new members
+        if let Some(duplicate) = update.new_members().iter().find(|m| self.keys.contains(*m)) {
+            // `duplicate` is a reference to the first member that already exists in `self.keys`.
+            return Err(MultisigConfigError::MemberAlreadyExists(*duplicate));
+        }
+
+        // Ensure old members exist
```
barakshani:
```suggestion
        // Ensure old members exist.

File: crates/crypto/src/multisig/config.rs [hunk 34]
```diff
@@ -0,0 +1,259 @@
...
+    ///
+    /// Returns `MultisigConfigError` if:
+    /// - `MemberAlreadyExists`: A new member already exists in the current configuration
+    /// - `MemberNotFound`: An old member to be removed doesn't exist in the current configuration
+    /// - `InvalidThreshold`: New threshold is less than half the updated member count
+    pub fn validate_update(
+        &self,
+        update: &MultisigConfigUpdate,
+    ) -> Result<(), MultisigConfigError> {
+        // Ensure no duplicate new members
```
barakshani:
```suggestion
        // Ensure no duplicate new members.

File: crates/crypto/src/multisig/config.rs [hunk 35]
```diff
@@ -0,0 +1,259 @@
...
+        if let Some(duplicate) = update.new_members().iter().find(|m| self.keys.contains(*m)) {
+            // `duplicate` is a reference to the first member that already exists in `self.keys`.
```
barakshani, line 117: what does _first_ mean here? is the list of keys/members ordered?

crates/crypto/src/multisig/config.rs
```diff
@@ -0,0 +1,259 @@
...
+            return Err(MultisigConfigError::MemberAlreadyExists(*duplicate));
+        }
+
+        // Ensure old members exist
+        if let Some(missing) = update
+            .old_members()
+            .iter()
+            .find(|m| !self.keys.contains(*m))
+        {
+            // `missing` is the first member that wasn’t found in `self.keys`
```
barakshani:
```suggestion
            // `missing` is the first member that wasn’t found in `self.keys`.

File: crates/crypto/src/multisig/config.rs [hunk 37]
```diff
@@ -0,0 +1,259 @@
...
+            return Err(MultisigConfigError::MemberNotFound(*missing));
+        }
+
+        // Ensure new threshold is strictly greater than half
```
barakshani:
```suggestion
        // Ensure new threshold is strictly greater than half.

File: crates/crypto/src/multisig/config.rs [hunk 38]
```diff
@@ -0,0 +1,259 @@
...
+        let updated_size =
+            self.keys.len() + update.new_members().len() - update.old_members().len();
+        let min_required = updated_size.div_ceil(2);
+
+        if (update.new_threshold() as usize) < min_required {
```
barakshani: @prajwolrg need to check against max (updated_size) as well.

crates/crypto/src/multisig/config.rs
```diff
@@ -0,0 +1,259 @@
...
+                min_required,
+                max_allowed: updated_size,
+            });
+        }
+
+        Ok(())
+    }
+
+    /// Applies an update to this configuration by removing old members, adding new members, and
+    /// updating the threshold. Must call `validate_update` first to ensure the update is valid.
```
barakshani: any reason not having a _validate_update_and_apply_ function to make sure the right flow happens?

crates/asm/txs/administration/src/constants.rs
prajwolrg: Right now, it only allows changes to OL and ASM verifying keys. Updated it.

crates/crypto/src/multisig/config.rs
prajwolrg: This allows validation before queuing, but the apply happens after N number of blocks.

prajwolrg: But I see there can be a case where two updates are lined up, where the second update can be invalid only after the first update has been applied. I don't expect that to happen, but will update it to handle that edge case as well.

Cargo.toml
```diff
@@ -106,6 +108,8 @@ rlp = "0.5.2"
 strata-asm-common = { path = "crates/asm/common" }
 strata-asm-logs = { path = "crates/asm/logs" }
 strata-asm-moho-program-impl = { path = "crates/asm/moho-program" }
+strata-asm-proto-administration = { path = "crates/asm/subprotocols/administration" }
+strata-asm-proto-administration-txs = { path = "crates/asm/txs/administration" }
```
delbonis, line 112: This crate name / file structure is a bit inconsistent.

Would call the latter one `strata-asm-tx-admin`.  And we can shorten `administration` to `admin` in all cases I think.

crates/crypto/src/multisig/config.rs
delbonis, line 1: Ordering of types and impls in this module is interleaved.

crates/crypto/src/multisig/config.rs
```diff
@@ -0,0 +1,259 @@
...
+    pub fn apply(&mut self, update: &MultisigConfigUpdate) {
+        // REVIEW: If we assert these lists are always sorted then we can do a more efficient
+        // merge-and-remove pass with both this and the new entries
+        // Remove specified old members
+        self.keys.retain(|key| !update.old_members().contains(key));
+        // Add new members
+        self.keys.extend_from_slice(update.new_members());
+        // Update threshold
+        self.threshold = update.new_threshold();
+    }
```
delbonis: `apply_update` to be more clear about what it does.

However, I am not sure I love this procedure because it does a lot of linear scans.  I wonder if it makes sense to make this operate on sorted lists to make merging updates easier?  This happens infrequently enough that it's probably not worth increasing the complexity to make it more efficient.

However, should we panic here if we try to remove keys that don't exist?  What if the threshold is greater than the new number of keys?  I understand there is a `validate_update` function, but we should still fail if we apply an update that is invalid with respect to the config.

Either this function should actually fallibly construct a new instance of the config (and then we can overwrite the old one with if it is successful), or do the additional checks while applying (even if they're redundant).

crates/crypto/src/multisig/mod.rs
delbonis, line 1: This module is designed for voting signatures, not just multisigs in general.  Should be named to reflect that, maybe even broken out into its own crate?

crates/crypto/src/multisig/mod.rs
```diff
@@ -0,0 +1,22 @@
+// TODO: This needs to be in a different crate. Maybe strata-crypto
+pub mod config;
+pub mod errors;
+pub mod msg;
+pub mod vote;
```
delbonis, line 5: Don't export all these unless it improves readability/ergonomics.  Should only export required types, and at the crate root level.

crates/crypto/src/multisig/config.rs
```diff
@@ -0,0 +1,259 @@
...
+    new_threshold: u8,
+}
+
+impl MultisigConfigUpdate {
+    pub fn new(new_members: Vec<PubKey>, old_members: Vec<PubKey>, new_threshold: u8) -> Self {
+        Self {
+            new_members,
+            old_members,
+            new_threshold,
+        }
```
delbonis, line 84: Should this sanity check that we're not adding and removing the same pubkeys?  Maybe that's actually fine.

crates/crypto/src/multisig/config.rs
delbonis: This is a policy-level constraint, not a technical one, so we should have some type that describes update constraints which we validate updates against.

In ref to other comment, that type would probably make sense to have that `config, update -> result<config>` function to do the validation-and-update checks.

crates/asm/subprotocols/administration/src/state.rs
```diff
@@ -0,0 +1,280 @@
...
+    /// cancelled, they are executed automatically once their activation height is reached.
+    queued: Vec<QueuedUpdate>,
+
+    /// UpdateId for the next update.
+    next_update_id: UpdateId,
+
+    /// The confirmation depth (CD) setting: after an update transaction receives this many
+    /// confirmations, the update is enacted automatically. During this confirmation period,
+    /// the update can still be cancelled by submitting a cancel transaction.
+    confirmation_depth: u32,
```
delbonis: Should this also be part of params instead of state?

prajwolrg: Oh yes, this makes sense to be part of params instead of state. Updating it at [b50ae16](https://github.com/alpenlabs/alpen/pull/1013/commits/b50ae16924b60459ce03650b97c0d49c52d6809d).

crates/crypto/src/multisig/config.rs
prajwolrg: The `old_members` will be replaced by the BitVec instead of the actual pub keys. This is done as part of https://github.com/alpenlabs/alpen/pull/1027.

crates/crypto/src/multisig/mod.rs
prajwolrg: This will be fixed in https://github.com/alpenlabs/alpen/pull/1027.

crates/crypto/src/multisig/mod.rs
prajwolrg: I'm not sure. What's the difference between voting signatures and multisigs in general?

crates/crypto/src/multisig/config.rs
prajwolrg: We **cannot** panic in the ASM STF. Panicking in the ASM STF means that the proof of the ASM STF proof cannot be generated, which means that there is no way to enact the update or continue Moho.


> This happens infrequently enough that it's probably not worth increasing the complexity to make it more efficient. 

I agree on this.

Will be doing the additional check. I'll do the validation again before applying the changes.

crates/crypto/src/multisig/config.rs
prajwolrg: This will be fixed in https://github.com/alpenlabs/alpen/pull/1027.

Cargo.toml
prajwolrg: I'll do that as part of https://github.com/alpenlabs/alpen/pull/1027 to avoid a lot of conflicts on rebase.

crates/crypto/src/multisig/config.rs
prajwolrg: Will be done as part of https://github.com/alpenlabs/alpen/pull/1013.

crates/crypto/src/multisig/config.rs
prajwolrg: Added revalidation [5aff27c](https://github.com/alpenlabs/alpen/pull/1013/commits/5aff27c48a5adf2f1d7ce9de22e86063ddd0bde2).

barakshani: yeah, that's tricky. We still need an answer to the question of whether we should line-up two updates without an explicit cancelation in between. Initially I thought that it would be easier to require the explicit cancelation, but I'm not sure anymore.

I assume that what you describe can only happen for the same update "type", and I also assume that it cannot happen if we have a cancellation in between. I don't know if these assumptions hold.

Anyway, I'm curious to see how you're going to address this : )

crates/crypto/src/multisig/config.rs
prajwolrg: Thanks. Good catch!

crates/crypto/src/multisig/config.rs
prajwolrg: I realize we don't need this minimum bar. The PRD doesn't require this.

crates/crypto/src/multisig/config.rs
```diff
@@ -0,0 +1,266 @@
...
+                min_required,
+                max_allowed: updated_size,
+            });
+        }
+
+        Ok(())
+    }
+
+    /// Applies an update to this configuration by removing old members, adding new members, and
+    /// updating the threshold. Must call `validate_update` first to ensure the update is valid.
```
barakshani:
```suggestion
    /// updating the threshold.
```
This part actually isn't needed anymore.

crates/crypto/src/multisig/config.rs
prajwolrg: In case there are multiple members, it means the first member.

crates/crypto/src/multisig/config.rs
prajwolrg: Yup. thanks for catching it.

crates/crypto/src/multisig/config.rs
```diff
@@ -0,0 +1,259 @@
...
+            .old_members()
+            .iter()
+            .find(|m| !self.keys.contains(*m))
+        {
+            // `missing` is the first member that wasn’t found in `self.keys`.
+            return Err(MultisigConfigError::MemberNotFound(*missing));
+        }
+
+        // Ensure new threshold doesn't exceed total number of keys.
+        let updated_size =
```
barakshani, line 132: @prajwolrg we need to check that this isn't 0 now, similar to checking above (in `try_new`) that the original list is not empty.

crates/crypto/src/multisig/config.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// - `EmptyKeys`: The keys list is empty
+    /// - `InvalidThreshold`: The threshold exceeds the total number of keys
+    pub fn try_new(keys: Vec<PubKey>, threshold: u8) -> Result<Self, MultisigConfigError> {
+        if keys.is_empty() {
+            return Err(MultisigConfigError::EmptyKeys);
+        }
+
+        let total_keys = keys.len();
+
+        if threshold as usize > total_keys {
```
barakshani, line 31: We should also check that `threshold` is positive. Any idea what would happen if we set it to 0?

crates/crypto/src/multisig/config.rs
```diff
@@ -0,0 +1,259 @@
...
+        {
+            // `missing` is the first member that wasn’t found in `self.keys`.
+            return Err(MultisigConfigError::MemberNotFound(*missing));
+        }
+
+        // Ensure new threshold doesn't exceed total number of keys.
+        let updated_size =
+            self.keys.len() + update.new_members().len() - update.old_members().len();
+
+        if update.new_threshold() as usize > updated_size {
```
barakshani, line 135: We should separately check that the threshold isn't 0, as in the previous comment.

crates/crypto/src/multisig/config.rs
```diff
@@ -0,0 +1,259 @@
...
+        // Add new members
+        self.keys.extend_from_slice(update.new_members());
+        // Update threshold
+        self.threshold = update.new_threshold();
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
```
barakshani, line 166: let's add tests for update that makes the list of keys empty and for a multisig with threshold=0.

crates/crypto/src/multisig/config.rs
prajwolrg: That's correct. Let me handle that.

functional-tests/tests/cl_reorg_resume_blockproduction.py
barakshani, line 1: Let's revert the entire changes here, as it not really part of this PR (nor we get it to solve the issue).

crates/crypto/src/multisig/config.rs
prajwolrg: I tried to work on this, but since we will be using BitVec in https://github.com/alpenlabs/alpen/pull/1027, it makes more sense to handle this properly there.

crates/crypto/src/multisig/config.rs
prajwolrg: I tried to work on this, but since we will be using BitVec in https://github.com/alpenlabs/alpen/pull/1027, it makes more sense to handle this properly there.

crates/crypto/src/multisig/config.rs
prajwolrg: I tried to work on this, but since we will be using BitVec in https://github.com/alpenlabs/alpen/pull/1027, it makes more sense to handle this properly there.

crates/crypto/src/multisig/config.rs
prajwolrg: I tried to work on this, but since we will be using BitVec in https://github.com/alpenlabs/alpen/pull/1027, it makes more sense to handle this properly there.

crates/crypto/src/multisig/config.rs
barakshani: wait, `threshold` remains an integer regardless. we should just add additional condition to check if it's 0.

functional-tests/tests/cl_reorg_resume_blockproduction.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, ctx: flexitest.InitContext):
         )
 
     def main(self, ctx: flexitest.RunContext):
+        return True
```
barakshani, line 37: probably worth removing this as well, no?

crates/crypto/src/multisig/mod.rs
delbonis: Well nothing in particular, but all of these interfaces and names are based around voting concepts.

crates/crypto/src/multisig/config.rs
delbonis: > We cannot panic in the ASM STF. Panicking in the ASM STF means that the proof of the ASM STF proof cannot be generated, which means that there is no way to enact the update or continue Moho.

This is true, but if we realize while applying the update that it produces a result that's inconsistent with our policy constraints, then that means the update was not validated before being applied.  This means that there's a bug in the code, because that should not be allowed, so we *don't want* to be able to construct a proof for that update.  We would want to catch it before deploying it by writing unit tests for that case. 

The rule of thumb is that a panic makes sense in cases where types aren't practically useful to enforce a constraint but where a "sufficiently advanced" compiler would be able to "figure out" that the panic could never be triggered based on conditions tested earlier in the function (after inlining and whatever) and would remove the code path.

functional-tests/tests/cl_reorg_resume_blockproduction.py
prajwolrg: This is here just so we don't have to delete the whole file.

PR #1019, opened 2025-09-03
STR-1703: Benchmarking Databases

storopoli:
## Description

Basic benchmarking of read and write operations for RocksDB implementations of `L1Database` and `L2Database` traits in `strata-db`.

This is the first benchmark in the `alpen` repo.

Done with the help of `claude`.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Also added a `just` recipe. Note that we don't need to change CI since it already covers `--benches`: https://github.com/alpenlabs/alpen/blob/7de932a8b45ad4015aea3acd6450b211bcc80d68/.github/workflows/lint.yml#L49-L49

I am planning to do a PR on top of #979 or once it get's merged redo this PR for `sled`. Posting `RocksDB` numbers here:

## Checklist

      in the body of this PR.

## Related Issues

STR-1703


benches/src/db/l1_data_gen.rs
```diff
@@ -0,0 +1,80 @@
...
+};
+
+/// Generate L1 block manifests for benchmarking.
+#[derive(Debug)]
+pub struct L1DataGenerator {
+    current_height: u64,
+    block_cache: HashMap<L1BlockId, L1BlockManifest>,
+}
+
+impl L1DataGenerator {
```
bewakes: Can we use `ArbitraryGenerator` here? I believe it should be feasible and it would definitely be cleaner. We already have `Arbitrary` derived in most of the structs.

benches/src/db/l2_data_gen.rs
```diff
@@ -0,0 +1,156 @@
...
+    header::{L2BlockHeader, SignedL2BlockHeader},
+};
+
+/// Generate L2 block bundles for benchmarking.
+#[derive(Debug)]
+pub struct L2DataGenerator {
+    current_height: u64,
+}
+
+impl L2DataGenerator {
```
bewakes: Same here, I wonder if we can use `ArbitraryGenerator`.

benches/benches/l1_database.rs
```diff
@@ -0,0 +1,275 @@
...
+use strata_primitives as _;
+#[allow(unused_imports)]
+use strata_state as _;
+use tempfile::TempDir;
+
+/// Transaction counts to test across benchmarks.
+const TX_COUNTS: &[usize] = &[1, 2, 3, 5, 10, 20, 50];
+
+/// Block counts to test for chain operations.
+const BLOCK_COUNTS: &[usize] = &[1, 2, 3, 5, 10, 20, 50];
```
bewakes: Do you think we should have counts at least in the order of 1000 to actually test db under load?

storopoli: My initial implementation had 100, 250 and 1_000 but I thought it was overkill. I'm reconsidering now. Will extend this range to thousands.

benches/src/db/l1_data_gen.rs
delbonis: I'm not sure how much we can rely on it, since we can't really control the data that gets generated.

storopoli: Done in 344e6a460aab76fce753d3e31213b135488a378c

benches/src/db/l2_data_gen.rs
storopoli: Done in 344e6a460aab76fce753d3e31213b135488a378c

benches/benches/l1_database.rs
storopoli: Done in 344e6a460aab76fce753d3e31213b135488a378c

PR #1027, opened 2025-09-09
Administration Subprotocol Transactions with correct multisig verification

prajwolrg:
## Description

I used Claude Code for refactoring and docstring.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

- [X] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/.github/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

## Todo from https://github.com/alpenlabs/alpen/pull/1013

- [ ] Rename `administrator` -> `admin` for brevity. Make the crate name consistent. Eg: `strata-asm-tx-admin`.
- [ ] Define policy-level constraint for multisig. [Ref](https://github.com/alpenlabs/alpen/pull/1013#discussion_r2342325794)
- [x] Add tests for update that makes the list of keys empty and for a multisig with threshold=0 [Ref](https://github.com/alpenlabs/alpen/pull/1013#discussion_r2344328038)
- [ ] Fix ordering of types and impls in the multisig module. [Ref1](https://github.com/alpenlabs/alpen/pull/1013#discussion_r2342302125), [Ref2](https://github.com/alpenlabs/alpen/pull/1013#discussion_r2342310201)

## Related Issues

https://github.com/alpenlabs/alpen/pull/1013


crates/crypto/src/multisig/traits.rs
```diff
@@ -0,0 +1,66 @@
...
+    /// The type used to represent a public key in this scheme.
+    type PubKey: Clone
+        + Debug
+        + PartialEq
+        + Eq
+        + Send
+        + Sync
+        + BorshSerialize
+        + BorshDeserialize
+        + for<'a> Arbitrary<'a>;
```
storopoli, line 23: I think it needs also to be `Ord`/`PartialOrd` due to the fact that we need to lexicographically sort the PubKeys in MuSig2.

storopoli: And probably `Into<Point>` from `musig2` crate: https://docs.rs/musig2/latest/musig2/struct.KeyAggContext.html#method.new

crates/crypto/src/multisig/schemes/schnorr/create.rs
```diff
@@ -0,0 +1,194 @@
...
+///
+/// # Arguments
+/// - `operators_privkeys`: Private keys of all operators participating in signing
+/// - `message`: The message to be signed (typically a sighash)
+/// - `tweak`: Optional tweak for taproot spending (merkle root)
+///
+/// # Returns
+/// The aggregated MuSig2 signature
+pub fn create_musig2_signature(
+    signer_secretkeys: &[SecretKey],
```
storopoli: Use `EvenSecretKey` from the `primitives`.

crates/crypto/src/multisig/traits.rs
prajwolrg: The ordering inside Musig2 is done after converting the `XOnlyPubKey` type to the Full PubKey. Since in the state we store only the `XOnlyPubKey`, I'm not sure if adding Ord is valuable.

crates/asm/subprotocols/administration/src/subprotocol.rs
```diff
@@ -0,0 +1,59 @@
...
+pub struct AdministrationSubprotocol;
+
+impl Subprotocol for AdministrationSubprotocol {
+    const ID: SubprotocolId = ADMINISTRATION_SUBPROTOCOL_ID;
+
+    type Params = AdministrationSubprotoParams;
+
+    type State = AdministrationSubprotoState;
+
+    type Msg = NullMsg<0>;
```
irnb: `pub struct NullMsg<const ID: SubprotocolId>;`

```suggestion
    type Msg = NullMsg< ADMINISTRATION_SUBPROTOCOL_ID >;
```

It’s not important, but it helps avoid future unknown problems.

prajwolrg: Could you submit the review in https://github.com/alpenlabs/alpen/pull/1013 instead?

crates/crypto/src/multisig/traits.rs
storopoli: Oh we do the conversion implicitly? Ok then

crates/crypto/src/multisig/schemes/schnorr/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+pub struct SchnorrScheme;
+
+impl CryptoScheme for SchnorrScheme {
+    type PubKey = Buf32;
+    type Signature = Buf64;
+    type AggregatedKey = Buf32;
+
+    /// Aggregates public keys using MuSig2 key aggregation.
+    ///
+    /// This is moved from the original `generate_agg_pubkey` function in `aggregation.rs`.
```
barakshani: Why is this important to note?

crates/crypto/src/multisig/mod.rs
```diff
@@ -1,22 +1,61 @@
...
+/// then performs the full verification process using the provided cryptographic scheme.
+///
+/// # Arguments
+/// * `config` - The multisig configuration containing keys and threshold
+/// * `signature` - The multisig signature containing signer indices and aggregated signature
+/// * `message_hash` - The message hash that was signed
+///
+/// # Returns
+/// Returns `Ok(())` if verification succeeds, or an error if:
+/// - Insufficient keys are selected (fewer than threshold)
```
barakshani:
```suggestion
/// - Threshold is not achieved

File: crates/crypto/src/multisig/schemes/schnorr/mod.rs [hunk 6]
```diff
@@ -0,0 +1,87 @@
...
+
+use bitcoin::{key::Parity, secp256k1::PublicKey, XOnlyPublicKey};
+use musig2::KeyAggContext;
+use strata_primitives::buf::{Buf32, Buf64};
+
+use crate::multisig::{errors::MultisigError, traits::CryptoScheme};
+
+/// Schnorr signature scheme using MuSig2 key aggregation.
+///
+/// This is the current implementation that was previously hardcoded in the multisig module.
```
barakshani: Why is this important to note?

crates/crypto/src/multisig/schemes/schnorr/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+        keys: impl Iterator<Item = &'k Self::PubKey>,
+    ) -> Result<Self::AggregatedKey, MultisigError>
+    where
+        Self::PubKey: 'k,
+    {
+        let a = aggregate_schnorr_keys(keys)?;
+        Ok(Buf32::from(a.serialize())) // FIXME:
+    }
+
+    /// Verifies a Schnorr signature against a message hash using an aggregated public key.
```
barakshani:
```suggestion
    /// Verifies a Schnorr signature against a message hash using a(n aggregated) public key.

File: crates/crypto/src/multisig/schemes/schnorr/mod.rs [hunk 8]
```diff
@@ -0,0 +1,87 @@
...
+    ///
+    /// This uses the existing `verify_sig` function logic, but properly implemented.
```
barakshani: What do you mean by "properly"? It makes it sounds like something else, perhaps `verify_sig`, isn't.

crates/asm/subprotocols/administration/src/handler.rs
```diff
@@ -181,6 +192,37 @@ mod tests {
...
+            .iter()
+            .map(|sk| sk.x_only_public_key(SECP256K1).0.into())
+            .collect();
+        let strata_sequencer_manager =
+            SchnorrMultisigConfig::try_new(strata_seq_manager_pks, 2).unwrap();
+
+        let config = AdministrationSubprotoParams {
+            strata_administrator,
+            strata_sequencer_manager,
+            confirmation_depth: 2016,
```
barakshani, line 223: This could be lower, no?

crates/asm/subprotocols/administration/src/authority.rs
```diff
@@ -35,52 +34,28 @@ impl MultisigAuthority {
...
 
-    /// Validate that `vote` approves `action` under the current config and nonce.
+    /// Validate that `signature` approves `action` under the current config and nonce.
     ///
-    /// Steps:
-    /// 1. Map voter indices to pubkeys (error if out of bounds).
-    /// 2. Aggregate pubkeys and compute payload hash.
-    /// 3. Verify aggregated signature.
+    /// Uses the generic multisig verification function to orchestrate the workflow.
     pub fn validate_action(
```
barakshani: nit but is "validate" the right term? we basically verify correctness of the threshold signature. Note that [here](https://github.com/alpenlabs/alpen/blob/8e5c3792106addcb9f1fd0d634abae5502814728/crates/asm/subprotocols/administration/src/handler.rs#L67) we also seem to _validate_ the action.

You write there:
> /// Processes a multisig action by validating the aggregated vote.

but actually here is where we validate the aggregation. See my other comment there.

crates/asm/subprotocols/administration/src/handler.rs
```diff
@@ -79,7 +79,7 @@ pub(crate) fn handle_pending_updates(
 pub(crate) fn handle_action(
```
barakshani, line 82: For a few lines above:
> /// Processes a multisig action by validating the aggregated vote.

note that we actually do more than validatig the aggregated vote. Also, you seem to not use the term "vote" anymore, but "sig".

crates/asm/subprotocols/administration/src/authority.rs
```diff
@@ -14,14 +13,14 @@ pub struct MultisigAuthority {
     /// The role of this multisignature authority.
     role: Role,
     /// The public keys of all grant-holders authorized to sign.
-    config: MultisigConfig,
+    config: SchnorrMultisigConfig,
     /// Sequence number for the multisig configuration. It increases on each valid action.
     /// This is used to prevent replay attacks.
     seqno: u64,
```
barakshani, line 19: Just to clarify, this is public knowledge, and when the admin sends a "admin update" it must include the next sequence number? If so, where do we verify that the correct number is used?

crates/crypto/src/multisig/mod.rs
```diff
@@ -1,22 +1,71 @@
...
+/// - Signer indices exceed the available keys count
+/// - Insufficient keys are selected (fewer than threshold)
+/// - Key aggregation fails
+/// - Signature verification fails
+pub fn verify_multisig<S: CryptoScheme>(
+    config: &config::MultisigConfig<S>,
+    signature: &signature::MultisigSignature<S>,
+    message_hash: &[u8; 32],
+) -> Result<(), MultisigError> {
+    // Validate that signer indices don't exceed available keys
```
barakshani: If this is the case, won't the `aggregate` function below fail, perhaps even `selected_keys` could not map correctly?

crates/asm/subprotocols/administration/src/handler.rs
```diff
@@ -101,7 +101,7 @@ pub(crate) fn handle_action(
     let authority = state
```
barakshani, line 104: Above we have `let target_action_id = cancel.target_id();` -- does this means that the admin in the "cancel update" message should include this value?

crates/crypto/src/multisig/schemes/schnorr/mod.rs
prajwolrg: Yes right. This is not needed. I had asked Claude for the refactor. I didn't realize this. Thanks for catching it.

crates/crypto/src/multisig/schemes/schnorr/mod.rs
prajwolrg: It's not important. Thanks for the catch.

crates/asm/subprotocols/administration/src/handler.rs
prajwolrg: Yes, it could be anything.

crates/crypto/src/multisig/schemes/schnorr/mod.rs
prajwolrg: I previously mocked the impl and later used Claude Code in the refactor. So, that might have caused it.

crates/asm/subprotocols/administration/src/authority.rs
prajwolrg: I've changed the function name to `verify_action_signature`.

crates/asm/subprotocols/administration/src/handler.rs
prajwolrg: Yes, I'll update the docstring.

crates/asm/subprotocols/administration/src/authority.rs
prajwolrg: Yes, this is a public knowledge, but it doesn't need to include the next sequencer number. The message to sign must include the sequence number. It is done [here](https://github.com/alpenlabs/alpen/blob/e99f37783617128a6a0fbc77eafd71edc318cec3/crates/asm/subprotocols/administration/src/authority.rs#L56) in `verify_action_signature`.

crates/crypto/src/multisig/mod.rs
prajwolrg: Yes, it could have failed. I've handled it differently by padding/shrinking the bitvec. Please see if anything is missing now.

crates/asm/subprotocols/administration/src/handler.rs
prajwolrg: Yes. The admin in the cancel update, must include the target id that they want to cancel.

crates/asm/subprotocols/administration/src/handler.rs
barakshani: oh, I see, you're actually not using it in test.

crates/asm/subprotocols/administration/src/authority.rs
barakshani: You mean it's not the _next_ number they have to sign but the _current_? How does the admin get to know what the value is to start with?

crates/asm/subprotocols/administration/src/handler.rs
barakshani: It's this value [here](https://github.com/alpenlabs/alpen/blob/main/crates/asm/subprotocols/administration/src/state.rs#L26), right? How would the admin get that value?

crates/crypto/src/multisig/mod.rs
barakshani: I only see the removal of code [here](https://github.com/alpenlabs/alpen/commit/8e2d8c6ba866d0ca50eefdf0ac0838f1f578658b). I didn't ask to change the impl, just wanted to make sure. Doing extra verification probably not going to hurt us here.

barakshani: Or did you mean to say that it was old code, and since you have handled it in a different way?

crates/asm/subprotocols/administration/src/handler.rs
prajwolrg: Each update gets an `UpdateId.` In the cancel, they'd have to sign that particular `UpdateId`.

They'd somehow have to maintain that on their own. It will be part of the off-chain infrastructure that will be responsible for gathering the individual nonces and signatures.

crates/crypto/src/multisig/mod.rs
prajwolrg: Yes, it was an old code.

With the new impl, I don't think we need the additional verification.

The changes are part of `apply_update` function [here](https://github.com/alpenlabs/alpen/blob/STR-1459-upgrade-txs/crates/crypto/src/multisig/config.rs#L232-L242).

crates/asm/subprotocols/administration/src/authority.rs
prajwolrg: No, I meant the `seqno` is not part of the payload that is posted onto L1. When signing the msg, they have to sign the message that already includes the correct `seqno`. The combination of the `seqno` with the action is done as part of `compute_sighash` [here](https://github.com/alpenlabs/alpen/blob/STR-1459-upgrade-txs/crates/asm/txs/administration/src/actions/mod.rs#L41-L47). If the correct expected `seqno` is not used, the action is invalid.

The admin would have to maintain what the next `seqno` is on their own. It will be part of the off-chain infrastructure that will be responsible for gathering the individual nonces and signatures.

They can always get what the next `seqno` is by inspecting the subprotocol state, but it'll be abstracted by the off-chain infrastructure.

barakshani: I feel like we use slightly different wording. For me, it's not really important if `seqno` is implicit or explicit, as long it's verifiable. When you say "`seqno` is not part of the payload that is posted onto L1", do you just mean that the bytes aren't explicit in the message (instead they will just be part of the hash digest)?

We'll need to put some thought into how the admins get `seqno`, and whether it's verifiable. Right now, do I understand correctly that we don't expose it?

crates/asm/subprotocols/administration/src/handler.rs
barakshani: We'll need to put some thought into how the admins get `UpdateId`, and whether it's verifiable. I guess one way would be to scan history? Right now, do I understand correctly that we don't expose it?

prajwolrg: `UpdateId`, are similar to seqno in the sense they are monotonically increasing, but are global across all roles. So everytime we have a new update, we assign it a new ID based on the latest `UpdateId`. 

I had previously used `UpdateId = sha2(Update)` to make the UpdateId, but Trey pointed out that is unncessary and it just increases the DA cost. 

I'm not sure what you mean by exposing it.
If you mean that there are no public functions to get the seqno, you're correct.

crates/asm/subprotocols/administration/src/authority.rs
prajwolrg: Yes, I mean't the bytes aren't explicit in the message and are part of hash digest.

Yes, right now there is no public function to get the `seqno`.

crates/asm/subprotocols/administration/src/handler.rs
barakshani: As above, I am concerned with he question of how the admin can obtain this value.

barakshani: What the purpose of `UpdateId`? It seems to me we only use it for cancellation.

crates/crypto/src/multisig/mod.rs
```diff
@@ -1,22 +1,62 @@
...
+pub use errors::MultisigError;
 
-// FIXME: handle
-pub fn aggregate_pubkeys(_keys: &[PubKey]) -> Result<PubKey, VoteValidationError> {
-    Ok(PubKey::default())
-}
+use crate::multisig::traits::CryptoScheme;
+
+/// Generic multisig verification function that takes a multisig configuration, multisig signature,
+/// and message hash, then performs the full verification process using the provided cryptographic
```
barakshani:
```suggestion
/// Generic multisig verification function that takes a multisig configuration, message hash
/// and the multi-signature, then performs the full verification process using the provided cryptographic

File: crates/crypto/src/multisig/mod.rs [hunk 16]
```diff
@@ -1,22 +1,62 @@
...
+/// scheme.
+///
+/// # Arguments
+/// * `config` - The multisig configuration containing keys and threshold
+/// * `signature` - The multisig signature containing signer indices and aggregated signature
```
barakshani:
```suggestion
/// * `signature` - The multi-signature containing signer indices and aggregated signature

File: crates/crypto/src/multisig/mod.rs [hunk 17]
```diff
@@ -1,22 +1,62 @@
...
+/// * `message_hash` - The message hash that was signed
+///
+/// # Returns
+/// Returns `Ok(())` if verification succeeds, or an error if:
+/// - Signer indices exceed the available keys count
```
barakshani: do we actually check for that in the current version?

crates/crypto/src/multisig/mod.rs
barakshani, line 1: @sampkaALP please check the changes in this crates.

crates/asm/subprotocols/administration/src/handler.rs
```diff
@@ -64,22 +64,25 @@ pub(crate) fn handle_pending_updates(
     }
 }
 
-/// Processes a multisig action by validating the aggregated vote.
+/// Processes a multisig action by validating the aggregated signature and executing the requested
+/// operation.
```
barakshani:
```suggestion
/// Processes a multisig action (an admin "change" message) by validating the aggregated 
/// signature and executing the requested operation.

File: crates/asm/subprotocols/administration/src/handler.rs [hunk 19]
```diff
@@ -253,6 +300,79 @@ mod tests {
         }
     }
 
+    /// Test that multisig actions reject invalid sequence numbers.
+    ///
+    /// Verifies sequence number validation prevents replay attacks by rejecting
```
barakshani:
```suggestion
    /// Verifies that sequence number validation prevents replay attacks by rejecting

File: crates/asm/subprotocols/administration/src/state.rs [hunk 20]
```diff
@@ -229,23 +229,29 @@ mod tests {
...
+        for i in 0..initial_members.len() {
+            if rng.gen_bool(0.3) {
+                // 30% chance to remove each member
+                members_to_remove_bitvec.set(i, true);
+                removal_count += 1;
+            }
+        }
+
+        let new_size = initial_members.len() + new_members.len() - removal_count;
+        let new_threshold = std::cmp::max(1, new_size); // Ensure valid threshold (at least majority)
```
barakshani: this isn't majority, right? It's the entire set (unless the set is empty)

crates/crypto/src/multisig/config.rs
```diff
@@ -1,44 +1,59 @@
...
 
 /// Configuration for a multisignature authority:
 /// who can sign (`keys`) and how many of them must sign (`threshold`).
-#[derive(Clone, Debug, Eq, PartialEq, BorshDeserialize, BorshSerialize)]
-pub struct MultisigConfig {
+#[derive(Debug, Clone, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct MultisigConfig<S: CryptoScheme> {
     /// The public keys of all grant-holders authorized to sign.
-    pub keys: Vec<PubKey>,
+    pub keys: Vec<S::PubKey>,
```
barakshani, line 13: it is possible we have duplicates here?

crates/crypto/src/multisig/config.rs
barakshani, line 1: @sampkaALP please check the changes.

crates/asm/subprotocols/administration/src/state.rs
```diff
@@ -256,13 +262,17 @@ mod tests {
...
         assert_eq!(updated_auth.threshold(), new_threshold as u8);
 
-        // Verify that old members were removed
-        for old_member in &members_to_remove {
-            assert!(
-                !updated_auth.keys().contains(old_member),
-                "Old member {:?} was not removed",
-                old_member
-            );
+        // Verify that old members indicated by bitvec were removed
```
barakshani: this works only if there are no duplicate pk in the list. there should not be deuplicates! I left a comment on crates/crypto/src/multisig/config.rs.

crates/asm/subprotocols/bridge-v1/src/txs/test_utils/deposit.rs
```diff
@@ -1,130 +1,23 @@
...
-        return second_round
-            .finalize()
-            .expect("Failed to finalize MuSig2 signature");
-    }
-
-    panic!("No signers available to finalize signature");
-}
-
-/// Creates a test deposit transaction with proper MuSig2 signatures following the SPS-50
+// Creates a test deposit transaction with proper MuSig2 signatures following the SPS-50
```
barakshani:
```suggestion
/// Creates a test deposit transaction with proper MuSig2 signatures following the SPS-50

File: crates/asm/subprotocols/administration/src/handler.rs (hunk 14 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/administration/src/handler.rs:104]: It's a unique identifier for the update. 

Yes, it's only used for update cancellation.

But in the some off-chain dashboard, we can use it as identifier for all the different proposals that have been acted upon.

File: crates/crypto/src/multisig/config.rs (hunk 21 above)
Comment (prajwolrg) [reply crates/crypto/src/multisig/config.rs:13]: No, it's not possible to have duplicates here. We check for duplicates and return an error in case of duplicate.

File: crates/asm/subprotocols/administration/src/state.rs (hunk 20 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/administration/src/state.rs]: Nice catch! Previously there was an additional requirement in the threshold, such that it is always greater than half of the total number of keys. I've removed that now. It's upto the signer set to choose a good threshold. The only checks on the threshold is that it is not zero, and it less than or equal to the total number of signers.

File: crates/asm/subprotocols/administration/src/state.rs (hunk 22 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/administration/src/state.rs]: Yes, there should be no duplicates. It should return an error in case of duplicate.

File: crates/crypto/src/multisig/mod.rs (hunk 17 above)
Comment (prajwolrg) [reply crates/crypto/src/multisig/mod.rs]: Nice catch. We don't need it in the new implementation.

File: crates/asm/subprotocols/administration/src/handler.rs (hunk 14 above)
Comment (barakshani) [reply crates/asm/subprotocols/administration/src/handler.rs:104]: I mean, what's the purpose of this identifier?
In addition, we already have (role, sequence) as one-to-one map, isn't it?

File: crates/crypto/src/multisig/mod.rs (hunk 15 above)
Comment (prajwolrg) [reply crates/crypto/src/multisig/mod.rs]: I'm reconsidering the name for this. `MultisigSignature` sounds a bit weird. I'm thinking of `AggregatedSignature` (similar to Ethereum). ChatGPT suggested `QuorumSignature`.  Any suggestions? CC: @storopoli @delbonis @sampkaALP

File: crates/asm/subprotocols/administration/src/handler.rs (hunk 14 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/administration/src/handler.rs:104]: `seqno` increases on each action by the multisig, whether that is a new update or the cancellation and is role-specific.

`UpdateId` is global to all roles.

The purpose right now is only to help in cancellation. If we don't support cancellation, it's not needed.

File: crates/crypto/src/multisig/mod.rs (hunk 15 above)
Comment (barakshani) [reply crates/crypto/src/multisig/mod.rs]: `AggregatedSignature` sounds good.

Comment (delbonis) [reply crates/crypto/src/multisig/mod.rs]: `AggregatedSignature` is fine, "quorum" invokes consensus algo concepts which this isn't *really*.

File: crates/asm/subprotocols/administration/src/handler.rs (hunk 14 above)
Comment (delbonis) [reply crates/asm/subprotocols/administration/src/handler.rs:104]: I think it makes sense to have since it's useful to be able to refer to actions across different contexts.

File: crates/asm/txs/administration/src/actions/mod.rs [hunk 24]
```diff
@@ -17,3 +25,37 @@ pub enum MultisigAction {
...
+    }
+
+    pub fn tx_type(&self) -> TxType {
+        match self {
+            MultisigAction::Cancel(_) => CANCEL_TX_TYPE,
+            MultisigAction::Update(update) => match update {
+                UpdateAction::Multisig(_) => MULTISIG_CONFIG_UPDATE_TX_TYPE,
+                UpdateAction::OperatorSet(_) => OPERATOR_UPDATE_TX_TYPE,
+                UpdateAction::Sequencer(_) => SEQUENCER_UPDATE_TX_TYPE,
+                UpdateAction::VerifyingKey(_) => ASM_STF_VK_UPDATE_TX_TYPE, // FIXME:
```
barakshani: Can you add what should be fixed?

delbonis (dismissed):
did not review in deep detail but existing comments cover the kinds of things I would probably have commented on

crates/crypto/src/multisig/mod.rs
barakshani: Taking about naming, I also feel that `MultisigAction` is not natural: would prefer to call is `AdminAction` or even `AdminChange` (or presumably someone can come up with a better name).
I'm writing this here, because if we change to `AdminAction`, we can also change the above to `AdminSignature`.

crates/asm/txs/administration/src/parser.rs
```diff
@@ -1,54 +1,113 @@
...
-
-pub fn parse_tx_multisig_action_and_vote(
+use strata_crypto::multisig::SchnorrMultisigSignature;
+use strata_l1tx::envelope::parser::{enter_envelope, extract_until_op_endif};
+
+use crate::{actions::MultisigAction, errors::AdministrationTxParseError};
+
+/// Parses a transaction to extract both the multisig action and the aggregated signature.
+///
+/// This function extracts the administrative action from the taproot leaf script embedded
```
barakshani, line 11: why taproot? it is clear that op_return isn't feasible?

crates/crypto/src/multisig/config.rs
barakshani: We check it only for update, but not when we creating it the firs time. Is that right?

crates/crypto/src/multisig/config.rs
```diff
@@ -47,93 +62,156 @@ impl MultisigConfig {
...
-    /// Ensures new members don't already exist, old members exist, and new threshold doesn't
-    /// exceed the updated member count.
+    /// Ensures new members don't already exist, remove indices are valid, and new threshold
+    /// is within valid bounds for the updated member count.
     ///
     /// # Errors
     ///
-    /// Returns `MultisigConfigError` if:
+    /// Returns `MultisigError` if:
     /// - `MemberAlreadyExists`: A new member already exists in the current configuration
```
barakshani, line 183: we already err if the update itself contains the same member multiple times (the pk is duplicated)

crates/crypto/src/multisig/config.rs
```diff
@@ -47,93 +62,156 @@ impl MultisigConfig {
...
 
-        // Ensure old members exist.
-        if let Some(missing) = update
-            .old_members()
-            .iter()
-            .find(|m| !self.keys.contains(*m))
-        {
-            // `missing` is the first member that wasn’t found in `self.keys`.
-            return Err(MultisigConfigError::MemberNotFound(*missing));
+        // Ensure no duplicate members to add
```
barakshani: this should actually be above -- here we check something slightly different.

crates/crypto/src/multisig/config.rs
```diff
@@ -47,93 +62,156 @@ impl MultisigConfig {
...
-    /// - `InvalidThreshold`: New threshold exceeds the updated member count
-    pub fn validate_update(
-        &self,
-        update: &MultisigConfigUpdate,
-    ) -> Result<(), MultisigConfigError> {
-        // Ensure no duplicate new members.
-        if let Some(duplicate) = update.new_members().iter().find(|m| self.keys.contains(*m)) {
-            // `duplicate` is a reference to the first member that already exists in `self.keys`.
-            return Err(MultisigConfigError::MemberAlreadyExists(*duplicate));
+    /// - `RemovalBitVecTooLong`: The removal bitvec is longer than the current member count
```
barakshani: @prajwolrg do we actually check this?

crates/crypto/src/multisig/mod.rs
barakshani: @prajwolrg I'm a little bit uncomfortable with bitvec. We need to make sure that we don't have a previous (pending) update that adds/remove members such that, after applying it, the bitvec will map to the incorrect members. Given that we don't expect changes frequently, I'd say it's better to specify the members pk's explicitly.

crates/crypto/src/multisig/config.rs
barakshani: See my other note [here](https://github.com/alpenlabs/alpen/pull/1027#discussion_r2347247828) on removal.

crates/crypto/src/multisig/config.rs
```diff
@@ -144,18 +222,28 @@ impl MultisigConfig {
...
+    /// during the retain operation instead of searching through the removal list.
+    pub fn apply_update(&mut self, update: &MultisigConfigUpdate<S>) -> Result<(), MultisigError> {
         self.validate_update(update)?;
-        // REVIEW: If we assert these lists are always sorted then we can do a more efficient
-        // merge-and-remove pass with both this and the new entries
-        // Remove specified old members
-        self.keys.retain(|key| !update.old_members().contains(key));
+
+        // Remove members by index using retain
+        // Pad/shrink the bitvec to exact size of keys
```
barakshani: what do you mean by "shrink"? it sounds like the original list was too large, which the previous function (validate_update) should err on.

crates/crypto/src/multisig/schemes/schnorr/mod.rs
barakshani: please remove then

barakshani (changes requested):
Nice work!
Please see a few comments with @prajwolrg mention I'd like you to further clarify.

crates/crypto/src/multisig/config.rs
prajwolrg: No we also check it when creating for the first time. I updated the logic so that the checks are the same.

crates/crypto/src/multisig/config.rs
prajwolrg: The bitvec is represented as u8. So, there are only 5 members, the bitvec will still be initialized with length 8, so had to shrink it. 

But as per your suggestion, I've updated the update to explicitly use PublicKey instead of bitvec in [41ff435](https://github.com/alpenlabs/alpen/pull/1027/commits/41ff43513c4cc597294ab60a38a5ff97d071e52b).

crates/asm/txs/administration/src/parser.rs
prajwolrg: Using OP_RETURN is feasible. 

This is done primarily to save the cost. When the payload is > 126 bytes, using commit-reveal is less costly. Analysis [here](https://www.notion.so/Commit-Reveal-vs-OP_RETURN-Payload-Analysis-266901ba000f80eca4c4e509fb83c810#266901ba000f80eca4c4e509fb83c810).

crates/crypto/src/multisig/mod.rs
prajwolrg: We have two roles: Strata Administrator and Strata Sequencer Manager. `AdminAction` and `AdminSignature` seems more like the action and signature from the Strata Administrator.

But since the whole subprotocol is named as Administration Subprotocol, it makes sense as well.

crates/asm/txs/administration/src/parser.rs
barakshani: yeah, I understand, but what's the payload size?

prajwolrg: Payload size is the size of the information that needs to be provided. For example when the action is a new VerifyingKey, the size of the new VerifyingKey is 386 bytes. When the action is Cancel, we only require UpdateId (4 bytes). The payload for different transaction size is mentioned [here](https://www.notion.so/SPS-65-Strata-Administration-Subprotocol-265901ba000f80e583d7ff093da6b369?source=copy_link#26b901ba000f80389c3ddc55580da4d0)

barakshani (approved):

delbonis (dismissed):

crates/crypto/src/multisig/mod.rs
storopoli: Yeah `AggregatedSignature` is nice.

crates/crypto/src/multisig/schemes/schnorr/create.rs
```diff
@@ -0,0 +1,194 @@
...
+    let secp = Secp256k1::new();
+
+    // Adjust both public keys and private keys for even parity
+    let adjusted_keys: Vec<(PublicKey, EvenSecretKey)> = signer_secretkeys
+        .iter()
+        .map(|sk| {
+            let pk = PublicKey::from_secret_key(&secp, sk);
+            (pk, *sk)
+        })
+        .collect();
```
storopoli: If they are `EvenSecretKey`s they will always generate even PKs.
Hence you can instantiate a `KeyAggContext` with `XOnlyPublicKey`s and not `PublicKey`s since it takes an iterator for anything that `impl Into<Point>`: https://docs.rs/musig2/latest/musig2/struct.KeyAggContext.html#method.new

Which is annoying but you can use [`From<(XOnlyPublicKey, Parity)>`](https://docs.rs/secp/0.6.0/secp/struct.Point.html#impl-From%3C(XOnlyPublicKey,+Parity)%3E-for-Point)

crates/crypto/src/multisig/schemes/schnorr/create.rs
```diff
@@ -0,0 +1,194 @@
...
+/// - `operators_privkeys`: Private keys of all operators participating in signing
+/// - `message`: The message to be signed (typically a sighash)
+/// - `tweak`: Optional tweak for taproot spending (merkle root)
+///
+/// # Returns
+/// The aggregated MuSig2 signature
+pub fn create_musig2_signature(
+    signer_secretkeys: &[EvenSecretKey],
+    message: &[u8; 32],
+    tweak: Option<[u8; 32]>,
```
storopoli: I don't understand how this function interfaces with the whole idea of the multi-signature authorities.
Since, in order to run this function, you need to have all the secret keys.

The function should be broken down into rounds of MuSig2, where at no point we have the need to have knowledge or possession of all the secret keys.

It should take in all rounds the aggregated public key; then in round 1 should take the `PubNonce`s and output an `AggNonce`; then in round 2 should take all `PartialSignatures` and spit out a `Signature` by aggregating all the partials.

storopoli: How to communicate with other signers I don't think is in scope of this PR.

crates/crypto/src/multisig/schemes/schnorr/create.rs
storopoli: This is not important, resolving

crates/crypto/src/multisig/schemes/schnorr/create.rs
barakshani: This seems to be copied from crates/asm/subprotocols/bridge-v1/src/txs/test_utils/deposit.rs (I believe by Claude or a relative). Do we only use it for testing, as in crates/asm/txs/administration/src/test_utils/mod.rs?
Maybe we should move it there, to stress it's a testing tool.

prajwolrg: This is solely for testing purposes. It is not meant to be use elsewhere. Let me restructure it better reflect that. Fixed in [54dc670](https://github.com/alpenlabs/alpen/pull/1027/commits/54dc670f8940b0e90aa8eaa47e9610d2bd4c5d18).

storopoli (approved):
ACK 54dc670f8940b0e90aa8eaa47e9610d2bd4c5d18

storopoli:
Ok I'm putting this in the queue. Nice work @prajwolrg!

PR #1029, opened 2025-09-11
feat(asm): implement debug subprotocol v1

irnb:
## Description
  - Create new debug-v1 subprotocol with mock data injection capabilities
    * Support fake ASM log injection via FAKE_ASM_LOG_TX_TYPE
    * Support fake withdrawal intent creation via FAKE_WITHDRAW_INTENT_TX_TYPE

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

      in the body of this PR.

## Related Issues
STR-1444

### DIFF
--- a/crates/asm/logs/src/lib.rs
+++ b/crates/asm/logs/src/lib.rs
@@ -3,8 +3,6 @@
 //! This crate provides structured log types for the Anchor State Machine (ASM) in the Strata
 //! protocol. It defines various log entry types that capture important events within the system.
 
-use borsh::{BorshDeserialize, BorshSerialize};
-
 pub mod asm_stf;
 pub mod checkpoint;
 pub mod constants;
@@ -17,13 +15,3 @@ pub use checkpoint::CheckpointUpdate;
 pub use deposit::DepositLog;
 pub use export::NewExportEntry;
 pub use forced_inclusion::ForcedInclusionData;
-
-/// Enum wrapping all supported ASM log types.
-#[derive(Debug, Clone, BorshSerialize, BorshDeserialize)]
-pub enum AsmLogType {
-    AsmStfUpdate(AsmStfUpdate),
-    CheckpointUpdate(CheckpointUpdate),
-    DepositLog(DepositLog),
-    NewExportEntry(NewExportEntry),
-    ForcedInclusionData(ForcedInclusionData),
-}
--- a/crates/asm/logs/src/lib.rs
+++ b/crates/asm/logs/src/lib.rs
@@ -21,14 +21,9 @@ pub use forced_inclusion::ForcedInclusionData;
 /// Enum wrapping all supported ASM log types.
 #[derive(Debug, Clone, BorshSerialize, BorshDeserialize)]
 pub enum AsmLogType {
-    /// ASM state transition function update.
     AsmStfUpdate(AsmStfUpdate),
-    /// Checkpoint update.
     CheckpointUpdate(CheckpointUpdate),
-    /// Deposit log.
     DepositLog(DepositLog),
-    /// New export entry.
     NewExportEntry(NewExportEntry),
-    /// Forced inclusion data.
     ForcedInclusionData(ForcedInclusionData),
 }
--- a/crates/asm/logs/src/lib.rs
+++ b/crates/asm/logs/src/lib.rs
@@ -3,6 +3,8 @@
 //! This crate provides structured log types for the Anchor State Machine (ASM) in the Strata
 //! protocol. It defines various log entry types that capture important events within the system.
 
+use borsh::{BorshDeserialize, BorshSerialize};
+
 pub mod asm_stf;
 pub mod checkpoint;
 pub mod constants;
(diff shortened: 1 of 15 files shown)


crates/asm/subprotocols/debug-v1/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+//!
+//! # Transaction Types
+//!
+//! The debug subprotocol supports the following transaction types:
+//!
+//! - **`FAKE_ASM_LOG_TX_TYPE` (1)**: Injects arbitrary log messages into the ASM log output,
+//!   simulating logs that would normally originate from the bridge subprotocol. Example: Deposit
+//!   events (locking funds in n/n multisig)
+//!
+//! - **`FAKE_WITHDRAW_INTENT_TX_TYPE` (2)**: Creates withdrawal intents that are sent to the bridge
```
barakshani: maybe use "mock" instead of "fake"?

crates/asm/subprotocols/debug-v1/src/constants.rs
```diff
@@ -0,0 +1,32 @@
+use strata_l1_txfmt::{SubprotocolId, TxType};
+
+/// Debug subprotocol ID (set to u8::MAX to avoid production conflicts).
+pub(crate) const DEBUG_SUBPROTOCOL_ID: SubprotocolId = u8::MAX;
+
+/// Transaction type for fake ASM log injection.
+pub(crate) const FAKE_ASM_LOG_TX_TYPE: TxType = 1;
+
+/// Transaction type for fake withdrawal intent creation.
+pub(crate) const FAKE_WITHDRAW_INTENT_TX_TYPE: TxType = 2;
```
barakshani: maybe use "mock" instead of "fake"?

crates/asm/logs/src/lib.rs
```diff
@@ -15,3 +17,13 @@ pub use checkpoint::CheckpointUpdate;
...
+
+/// Enum wrapping all supported ASM log types.
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize)]
+pub enum AsmLogType {
+    AsmStfUpdate(AsmStfUpdate),
+    CheckpointUpdate(CheckpointUpdate),
+    DepositLog(DepositLog),
+    NewExportEntry(NewExportEntry),
+    ForcedInclusionData(ForcedInclusionData),
+}
```
delbonis: Don't like this, logs are defined to be "just bytes", like EVM logs.  Defining it this way means that logs not from this list means they're *invalid*, which isn't actually true.   We extract a type ID according to the msgfmt spec and *then* we can pick the concrete type we parse the body using.  And borsh does the wrong thing here, the prefixing is much less efficient then as defined.

We *should* have parsing for each of these types, but it shouldn't be defined together since it's not a closed set.

crates/asm/spec-debug/src/lib.rs
delbonis, line 1: This is very clever.

crates/asm/subprotocols/debug-v1/src/txs.rs
```diff
@@ -0,0 +1,270 @@
...
+/// Type alias for fake ASM log injection info.
+pub(crate) type FakeAsmLogInfo = AsmLogType;
+
+/// Type alias for fake withdrawal info.
+pub(crate) type FakeWithdrawInfo = WithdrawOutput;
+
+/// Parsed debug transaction types.
+pub(crate) enum ParsedDebugTx {
+    /// ASM log injection transaction.
+    FakeAsmLog(FakeAsmLogInfo),
```
delbonis: On the above comment, don't actually have to parse the encoded log.  We can just construct the `AsmLogEntry` directly.

I see that `AsmLogEntry` doesn't *have* a `from_raw` constructor fn, but that should be added.  I *like* the traits to aid in safely constructing them, but I was *very precise* in the spec docs about how these types should work and be composed to ensure safe forwards compatibility without surprises.  The `AsmLogEntry` type should be reworked to be wrapper around a `Vec<u8>` (instead of `OwnedMsg`) and have some kind of `try_as_msg` fn that returns a `Option<&MsgRef>` instead.

crates/asm/subprotocols/debug-v1/src/constants.rs
```diff
@@ -0,0 +1,26 @@
...
+/// Offset of amount field in auxiliary data.
+pub(crate) const AMOUNT_OFFSET: usize = 0;
+
+/// Offset of descriptor field in auxiliary data.
+pub(crate) const DESCRIPTOR_OFFSET: usize = AMOUNT_OFFSET + AMOUNT_SIZE;
+
+/// Minimum auxiliary data length for mock withdrawal intent.
+///
+/// Format: `[amount: 8 bytes][descriptor: variable]`
+pub(crate) const MIN_MOCK_WITHDRAW_INTENT_AUX_DATA_LEN: usize = AMOUNT_SIZE + 20; // +20 for minimum descriptor size
```
prajwolrg: nit: maybe create a separate constant for minimum descriptor size?

crates/asm/subprotocols/debug-v1/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[dependencies]
+borsh = { workspace = true }
+serde = { workspace = true }
+thiserror = { workspace = true }
+
+strata-asm-common = { workspace = true }
+strata-asm-logs = { workspace = true }
+strata-l1-txfmt = { workspace = true }
+strata-primitives = { workspace = true }
```
prajwolrg: nit: use `.workspace = true`. This makes @storopoli happy 😅.

prajwolrg (commented):
Looks good to me except minor nits.

crates/asm/subprotocols/debug-v1/Cargo.toml
prajwolrg: also workspace lints.

irnb: Thanks for the review; fixed the nits 😄. 

I’m not quite sure what you mean by “workspace lints”  could you elaborate a bit?

prajwolrg: I meant adding 

```toml
[lints]
workspace = true
```

but it seems to already been done. i somehow missed it.

prajwolrg (approved):

delbonis:
Failing checks ahhh merging

PR #1035, opened 2025-09-15
add common types and traits for alpen ee implementation

sapinb:
## Description
Add some common types and traits to start implementation of alpen-ee.
Should be considered subject to change as development proceeds.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
All changes are to be done additively, without breaking existing evm implementation until development is complete and OL/EE implementation can be switched.

## Checklist

      in the body of this PR.


crates/reth/ee/primitives/src/lib.rs
delbonis, line 1: There's a bunch of stuff here being redefined which should probably just be re-exports.  Also should not be using alloy here anywhere.

crates/reth/ee/sequencer/src/batch.rs
```diff
@@ -0,0 +1,132 @@
...
+        &self.proof
+    }
+}
+
+/// EE state update with proof sent to OL by EE sequencer
+/// Will correspond 1-1 to a batch
+#[derive(Debug, Clone)]
+pub struct EEUpdate {
+    /// update sequence no. Must equal expected sequence number in account
+    sequencer_no: u64,
```
delbonis: extra r

crates/reth/ee/sequencer/src/block.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// Maps 1-1 to an eth block and holds additional block production metadata
+#[derive(Debug, Clone)]
+pub struct BlockPackage {
+    /// blocknumber
+    number: u64,
+    /// Blockhash of ee block
+    blockhash: EEBlockHash,
+    /// Blockhash of parent ee block
+    parent_blockhash: EEBlockHash,
```
delbonis: missing

* chunk
* parent package ID
* exec block ID
* summary hash

also we don't need parent block hash

crates/reth/ee/sequencer/src/block.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// All input messages must be in correct order without any gaps
+    input_msgs: IndexedVec<InboundMsgEnvelope>,
```
delbonis, line 16: Why does this need to be indexed vec?

crates/reth/ee/sequencer/src/lib.rs
delbonis, line 1: Avoid exporting internal module structure, explicitly re-export the items that make sense to be exported at this level.

crates/reth/ee/sequencer/src/message.rs
delbonis, line 1: Message structure (which is mostly commented) here is a little weird conflates different types of outputs and is missing some fields.

crates/reth/ee/node/src/account_state.rs
delbonis, line 1: I'm not sure why these types are in this crate?

delbonis (changes requested):
I am maybe a little bit confused about what the scope here is.  There's a lot of concepts that are clearly borrowed from SPS-exec-chunk-protocol but changed in ways that I don't understand the rationale of.  I understand that there was some related work going on which was not based on that spec, but then why does it sorta follow it?  I would appreciate some discussion on this.  Some of it seems like it's trying to be more closely coupled to EVM concepts, which I think is not a good choice and was trying to avoid doing in the chunk spec doc.

Also, much of this EE SDK stuff should be in strata-common or maybe a new strata-sdk repo, shouldn't it?

crates/reth/ee/node/src/account_state.rs
storopoli: Maybe move them to `crates/reth/primitives`?

crates/reth/ee/node/src/consensus.rs
```diff
@@ -0,0 +1,53 @@
...
+
+pub trait OLConsensusTracker {
+    fn subscribe(&self) -> impl Stream<Item = OLConsensusUpdate>;
+}
+
+/// Consensus update for blocks not yet checkpointed/proven on OL/L1.
+/// Represents new blocks produced and signed by sequencer.
+/// TODO: rename
+#[derive(Debug, Clone)]
+pub struct PreConsensusUpdate {
```
storopoli, line 37: `PendingConsensusUpdate`?

crates/reth/ee/primitives/src/indexed_vec.rs
```diff
@@ -0,0 +1,42 @@
+/// Represents a segment of a list with an incrementing index
+#[derive(Debug, Clone)]
+pub struct IndexedVec<T> {
+    base_idx: u64,
+    inner: Vec<T>,
+}
```
storopoli, line 6: Can we use this? https://github.com/alpenlabs/alpen/blob/7048ed3468b8a587fae580c2f2b108838c7e4703/crates/primitives/src/indexed.rs#L3-L11

Cargo.toml
```diff
@@ -59,7 +63,6 @@ members = [
   "crates/test-utils/test-utils",
   "crates/test-utils/tx-indexer",
   "crates/util/mmr",
-  "crates/db-tests",
```
storopoli, line 62: Nice you removed the duplicated entry :)

crates/reth/ee/node/src/account_state.rs
sapinb: My intention is to keep this ee implementation separate from the existing reth interfacing code, so that existing tests, etc are not broken until we are ready to swap the implementations once development work is done. Mentioned above in the PR.

crates/reth/ee/sequencer/src/block.rs
sapinb: Because the messages must be read in order. Using indexedvec to indicate it while retaining access to index.

crates/reth/ee/sequencer/src/block.rs
sapinb: I'll change the name for this struct. 
This isnt meant to be an exact implementation of SPS-chunk-protocol.
In this simplified implementation, as the single ee sequencer (and ee prover) is the only entity that needs the additional metadata present in the package chain described in SPS-chunk-protocol. So this is initially being implemented as additional evm block metadata stored by sequencer in its own db (and keyed by evm blockhash). Most of this data is also only required until the prover creates a proof for this chunk (referred as batch here).

crates/reth/ee/primitives/src/lib.rs
sapinb: The reth related parts, including this will eventually be split into a separate repo, so I wanted to keep the datatypes used inside closer to the types used in reth, leading to some redefinitions of common types.

crates/reth/ee/sequencer/src/message.rs
sapinb: This message and state structures are meant to be internal to the ee implementation with subset of fields relevant to it. These are not the network types shared with OL.

crates/reth/ee/primitives/src/indexed_vec.rs
sapinb: Didnt want to bring in borsh when not required. Likely doesn't matter a lot.

crates/reth/ee/sequencer/src/block.rs
delbonis: But the outputs aren't part of a global indexing (like state queue entries were, for example).

sapinb:
closing in favor of #1052

PR #1036, opened 2025-09-16
Ol stf draft

bewakes:
## Description
This is a draft and PoC implementation of OL STF. It is a PoC but  not meant to discard as it follows the spec document to most extent. Some specific details like snark verification and mmr, and possibly some validation are omitted for simplicity.

The PR consists of two aspects:
1. **Add `services/stf-runner` crate**: This is the main change for this PR. All the new OL related types from various docs are present right here. Not the best way to do but I did not want to interfere with other crates. Later we can either swap or replace the types and traits. Also, used the name `stf-runner` to be different from `chain-executor`, although similar in functionalities, to reduce confusion.
2. ~**Add `bin/stf-runner-demo` binary(using CLAUDE)**: The sole purpose is to see in action how the state and roots change. This has generator for deposit logs, and ol transactions. See corresponding readme for info on how to run this. This is meant to be _removed_ as this PR evolves.~

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

      in the body of this PR.


crates/services/stf-runner/src/block.rs
```diff
@@ -0,0 +1,361 @@
...
+pub struct AsmManifest {
+    blockid: Buf32,
+    logs: Vec<AsmLogEntry>,
+}
+/// A log entry for an account operation
+#[derive(Debug, Clone)]
+pub struct OLLog {
+    account_serial: u32,
+    account_id: AccountId,
+    payload: Vec<u8>, // TODO: make this typed, serialization can be done at the edges
```
delbonis, line 109: Having both the serial and ID is redundant, we should only need the serial.

Also no, the payload should *not* be typed, *because* serialization can be done at the edges.

crates/chainexec/src/output.rs
```diff
@@ -50,23 +50,24 @@ impl EpochExecutionOutput {
...
     ///
     /// This is NOT a state diff, that requires more precise tracking.
-    write_batch: WriteBatch,
+    write_batch: WB,
 }
 
-impl BlockExecutionOutput {
-    pub fn new(computed_state_root: Buf32, logs: Vec<LogMessage>, write_batch: WriteBatch) -> Self {
+impl<WB, L> BlockExecutionOutput<WB, L> {
+    pub fn new(computed_state_root: Buf32, logs: Vec<L>, write_batch: WB) -> Self {
```
delbonis, line 70: Why were these made generic?

crates/chaintsn/src/context.rs
```diff
@@ -80,16 +81,16 @@ impl BlockHeaderContext for L2HeaderAndParent {
 /// Accessor for fetch and manipulate the state we're building on top of.
 ///
 /// This is supersceding the `StateCache` type.
-pub trait StateAccessor {
+pub trait StateAccessor<S = Chainstate> {
```
delbonis: Why was this made generic?

crates/chaintsn/src/context.rs
```diff
@@ -80,16 +81,16 @@ impl BlockHeaderContext for L2HeaderAndParent {
...
     ///
     /// This is a transitional accessor that we will deprecate and remove soon.
-    fn state_untracked(&self) -> &Chainstate;
+    fn state_untracked(&self) -> &S;
 
     /// Gets a mut ref to the state.
     ///
     /// This is a transitional accessor that we will deprecate and remove soon.
-    fn state_mut_untracked(&mut self) -> &mut Chainstate;
+    fn state_mut_untracked(&mut self) -> &mut S;
```
delbonis, line 93: Actually should be removing this function.

crates/chaintsn/src/context.rs
```diff
@@ -12,6 +12,7 @@ use thiserror::Error;
 ///
 /// Does NOT provide access to chainstate information.  This is primarily
 /// involving block headers.  It will probably also provide L1 manifests.
+// TODO/NOTE: There's already a similar trait called `L2Header`, might need to merge those.
 pub trait BlockHeaderContext {
```
delbonis, line 16: Trying to expose only a minimal amount of information about the block header to avoid an issue with the cyclical state root we had a while ago.

crates/services/stf-runner/src/lib.rs
delbonis, line 1: This conflates a whole bunch of STF related concepts with service related concepts.  Because of that we can't use this in the OL proofs, it needs to be refactored out.  This OL STF implementation currently exists in the chainstf crate and it would have made sense to be done there, since the chain executor crate already is designed around that.

crates/services/stf-runner/src/block.rs
delbonis, line 1: Why are all these block types being redefined here instead of being used from more appropriate crates?  Also, the struct defs and impl blocks for them are not well-organized here, they should be close together.

I think it would make sense to split apart the tx and manifest related concepts into separate modules from the high-level block structure in order to aid code navigation.

crates/services/stf-runner/src/block.rs
```diff
@@ -0,0 +1,361 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub enum TransactionPayload {
+    GenericAccountMessage {
+        target: AccountId,
+        payload: Vec<u8>,
+    },
+    SnarkAccountUpdate {
+        target: AccountId, // is the transaction supposed to update the state of this target? looks
```
delbonis: Yes, that is what the spec says.

crates/services/stf-runner/src/block.rs
```diff
@@ -0,0 +1,361 @@
...
+        // Hash all header fields in a deterministic order
+        hasher.update(self.timestamp.to_be_bytes());
+        hasher.update(self.slot.to_be_bytes());
+        hasher.update(self.epoch.to_be_bytes());
+        hasher.update(self.parent_blockid.as_ref());
+        hasher.update(self.body_root.as_ref());
+        hasher.update(self.state_root.as_ref());
+
+        Buf32::new(hasher.finalize().into())
+    }
```
delbonis, line 289: This is redundant because we're going to be using SSZ.

crates/services/stf-runner/src/block.rs
```diff
@@ -0,0 +1,361 @@
...
+    /// The account id this transaction belongs to. Maybe we should also store sequencer pubkey
+    /// along with vk? and then we can have transactions to update the pubkey if sequencer needs to
+    /// rotate. Just a thought.
+    pub fn account_id(&self) -> AccountId {
+        match self.payload() {
+            TransactionPayload::SnarkAccountUpdate { target, .. } => *target,
+            // FIXME: this is probably not correct for Generic Account Message
+            TransactionPayload::GenericAccountMessage { target, .. } => *target,
+        }
+    }
```
delbonis: Transactions do not *belong to* accounts, they just exist as part of a block.

crates/services/stf-runner/src/ledger.rs
```diff
@@ -0,0 +1,194 @@
...
+
+#[derive(Debug, Error)]
+pub enum LedgerError {}
+
+pub type LedgerResult<T> = Result<T, LedgerError>;
+
+/// Interface for accessing and modifying accounts ledger
+pub trait LedgerProvider {
+    /// Root of the current accounts ledger. For example root of the accounts trie.
+    fn root(&self) -> LedgerResult<Buf32>;
```
delbonis: Don't want to expose this because we don't want to guarantee that it's always available.

crates/services/stf-runner/src/ledger.rs
```diff
@@ -0,0 +1,194 @@
...
+    fn account_state(&self, acct_id: &AccountId) -> LedgerResult<Option<AccountState>>;
+
+    /// Convenient method for accessing state via serial.
+    fn account_serial_state(&self, serial: AccountSerial) -> LedgerResult<Option<AccountState>> {
+        if let Some(acct_id) = self.account_id(serial)? {
+            self.account_state(&acct_id)
+        } else {
+            Ok(None)
+        }
+    }
```
delbonis, line 37: Should have `get_` prefix.

crates/services/stf-runner/src/ledger.rs
```diff
@@ -0,0 +1,194 @@
...
+        message: SnarkAccountMessageEntry,
+    ) -> LedgerResult<()>;
+
+    /// Consume input messages. Most likely updates some input index in state
+    fn consume_messages(
+        &mut self,
+        acct_id: AccountId,
+        from_idx: u64,
+        to_idx: u64,
+    ) -> LedgerResult<()>;
```
delbonis: What does this mean?  Accounts consume messages immediately, they don't need to explicitly re-consume them at this level.

crates/services/stf-runner/src/ledger.rs
```diff
@@ -0,0 +1,194 @@
...
+}
+
+/// Simplest in-memory ledger. All it has is an in-memory map of acct id to list of messages.
+#[derive(Debug, Clone)]
+pub struct InMemoryVectorLedger {
+    pub acct_msgs: HashMap<AccountId, Vec<SnarkAccountMessageEntry>>,
```
delbonis: What does this field represent?  Can you explain why it's exposed here?

crates/services/stf-runner/src/stf.rs
```diff
@@ -0,0 +1,202 @@
...
+// FIXME: There's a lot of func arguments :o
+pub fn process_block(
+    _prev_state: &OLState, // Possibly redudant with state_accessor
+    prev_header: &OLBlockHeader,
+    block: &OLBlock,
+    params: &RollupParams,
+    // state accessor is expected to allow CRUD operations on state
+    state_accessor: &mut impl StateAccessor<OLState>,
+    // ledger_provider is expected to allow CRUD operations on accounts ledger
+    ledger_provider: &mut impl LedgerProvider,
```
delbonis, line 104: Why is OLState being passed in explicitly if we have a `StateAccessor<OLState>` also?  Why does `LedgerProvider` exist separately when `StateAccessor` was created to be there to support accessing accounts in a structured way?

crates/services/stf-runner/src/service.rs
```diff
@@ -0,0 +1,91 @@
...
+    type Msg = StfRunnerMessage;
+
+    type Status = StfRunnerStatus;
+
+    fn get_status(_s: &Self::State) -> Self::Status {
+        StfRunnerStatus {}
+    }
+}
+
+impl AsyncService for StfRunnerService {
```
delbonis, line 23: It would be a sync service not an async one, but this already exists in the chain executor crate.

crates/services/stf-runner/src/block.rs
bewakes: > Having both the serial and ID is redundant, we should only need the serial.

Oh yes, having both is redundant.

Regarding payload,  we should have some struct that gets serialized into `payload: Vec<u8>`, right? There's got to be something to serialize.

crates/chainexec/src/output.rs
bewakes: These are transitional. Since the current chainstate differs significantly from the OLState in the spec, I did not want to make intrusive changes on the `Chainstate` and hence a new `OLState` which should be merged/replaced in the future. The generic is just to support both, with default being the Chainstate. I guess Log generic can be removed since the `LogMessage` and `OLLog` are pretty similar and thus only one should exist.

crates/chaintsn/src/context.rs
bewakes: The same reason as above, this is a transition and will be removed.

crates/chaintsn/src/context.rs
bewakes: I see. We can probably tackle this with supertraits.

crates/services/stf-runner/src/lib.rs
bewakes: I am not sure acutally. As far as I understand, the prover ultimately needs to call `process_block` and for that we should have prover specific implementation of `StateAccessor` or `LedgerProvider`(I know these are separate right now, but we will  probably need to merge them). Did I miss anything?

crates/services/stf-runner/src/block.rs
bewakes: Like I said in the description, the current types differ quite a bit and I did not want to alter things. I believe we can simply replace the types once we have gaps figured out.

crates/services/stf-runner/src/block.rs
bewakes: Okay, so the name is confusing. This kind of reminds me of "upbound"/"downbound" conversation. :D

crates/services/stf-runner/src/block.rs
bewakes: Yes, until SSZ. :) I hope the ssz integration won't be disruptive.

crates/services/stf-runner/src/block.rs
bewakes: Yes, but they are on behalf of some account right? From the comment above, the tx is on behalf of the `target` account.

crates/services/stf-runner/src/ledger.rs
bewakes: But a ledger always has a root right, even if it is empty.

crates/services/stf-runner/src/ledger.rs
bewakes: I am going to have to think more on this.

crates/services/stf-runner/src/stf.rs
bewakes: > Why is OLState being passed in explicitly if we have a StateAccessor<OLState> also?

See the comment on the first argument, I do intend to remove it.


> Why does LedgerProvider exist separately when StateAccessor was created to be there to support accessing accounts in a structured way?

We should probably merge them, I had the thought as well. But `StateAccessor` is already huge so I wanted to separate out the logic.

crates/services/stf-runner/src/block.rs
```diff
@@ -0,0 +1,361 @@
...
+        let current_header = self.signed_header.header();
+
+        if current_header.slot() > 0 && current_header.slot() != prev_header.slot() + 1 {
+            return Err(format!("Invalid block slot {}", current_header.slot()));
+        }
+        if current_header.slot() > 0
+            && *current_header.parent_blockid() != prev_header.compute_header_root()
+        {
+            return Err("Invalid parent block ID".to_string());
+        }
```
barakshani: Just curious, any reason not to have
```
if current_header.slot() {
      if current_header.slot() != prev_header.slot() + 1 {...}
      if *current_header.parent_blockid() != prev_header.compute_header_root() {...}
}
```

crates/services/stf-runner/src/block.rs
```diff
@@ -0,0 +1,361 @@
...
+
+        // Check epoch progression - epoch should not decrease
+        if current_header.epoch() != 0 && current_header.epoch() < prev_header.epoch() {
```
barakshani: What if the epoch number increases by more than 1?

crates/services/stf-runner/src/ledger.rs
barakshani: From the OL perspective once a message is in the queue that's all there is. The ledger/account should run their own logic of consuming messages from the queue, but that's independent of the OL.

crates/services/stf-runner/src/ledger.rs
barakshani, line 1: This supposed to be an account on the OL? There are some functionalities here that I'd defer to a later stage, like messaging. I think that right now the only kind of message we have is deposits, and I'm not sure if this is treated generically.

crates/services/stf-runner/src/state.rs
```diff
@@ -0,0 +1,103 @@
...
+/// Represents the state of the OL Chain
+#[derive(Debug, Clone, Default)]
+pub struct OLState {
+    accounts_root: Buf32,
+    l1_view: L1View,
+    cur_slot: u64,
+    cur_epoch: u64,
+}
+
+/// Represents the view of Layer 1 blockchain from the perspective of the OL
```
barakshani:
```suggestion
/// Represents the view of the layer 1 blockchain from the perspective of the OL

File: crates/services/stf-runner/src/state.rs [hunk 18]
```diff
@@ -0,0 +1,103 @@
+use strata_primitives::buf::Buf32;
+
+/// Represents the state of the OL Chain
+#[derive(Debug, Clone, Default)]
+pub struct OLState {
+    accounts_root: Buf32,
```
barakshani, line 7: How about `state_root`, if we're not going to have any greater state tree?

crates/services/stf-runner/src/state.rs
```diff
@@ -0,0 +1,103 @@
...
+    l1_view: L1View,
+    cur_slot: u64,
+    cur_epoch: u64,
+}
```
barakshani, line 18: We should have another field for total BTC.

crates/services/stf-runner/src/state.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    pub fn set_cur_slot(&mut self, cur_slot: u64) {
+        self.cur_slot = cur_slot;
+    }
+
+    pub fn set_cur_epoch(&mut self, cur_epoch: u64) {
+        self.cur_epoch = cur_epoch;
+    }
+
+    pub fn compute_root(&self) -> Buf32 {
```
barakshani, line 87: Why root? We just hash the OL state, no?

crates/services/stf-runner/src/ledger.rs
```diff
@@ -0,0 +1,194 @@
...
+        self.account_states.insert(id, state);
+        self.acct_msgs.insert(id, Vec::new());
+        self.invalidate_root_cache();
+    }
+
+    fn invalidate_root_cache(&mut self) {
+        self.root_cache = None;
+    }
+
+    pub fn compute_root(&mut self) -> LedgerResult<Buf32> {
```
barakshani, line 74: What exactly do we compute here?

crates/services/stf-runner/src/ledger.rs
```diff
@@ -0,0 +1,194 @@
...
+        use sha2::{Digest, Sha256};
+        let mut hasher = Sha256::new();
+
+        // Sort account IDs for deterministic ordering
+        let mut sorted_accounts: Vec<_> = self.account_states.keys().collect();
+        sorted_accounts.sort();
+
+        for account_id in sorted_accounts {
+            hasher.update(account_id.as_ref());
+            if let Some(state) = self.account_states.get(account_id) {
```
barakshani, line 87: How come we may not have `Some()` here?

crates/services/stf-runner/src/ledger.rs
```diff
@@ -0,0 +1,194 @@
...
+use crate::account::{AccountId, AccountSerial, AccountState, SnarkAccountMessageEntry};
+
+#[derive(Debug, Error)]
+pub enum LedgerError {}
+
+pub type LedgerResult<T> = Result<T, LedgerError>;
+
+/// Interface for accessing and modifying accounts ledger
+pub trait LedgerProvider {
+    /// Root of the current accounts ledger. For example root of the accounts trie.
```
barakshani, line 18: What is "accounts ledger"? Do you mean "ledger accounts"?

crates/services/stf-runner/src/ledger.rs
```diff
@@ -0,0 +1,194 @@
...
+    fn root(&self) -> LedgerResult<Buf32>;
+
+    /// Get account id from serial
+    fn account_id(&self, serial: AccountSerial) -> LedgerResult<Option<AccountId>>;
+
+    /// Get an account state
+    fn account_state(&self, acct_id: &AccountId) -> LedgerResult<Option<AccountState>>;
+
+    /// Convenient method for accessing state via serial.
+    fn account_serial_state(&self, serial: AccountSerial) -> LedgerResult<Option<AccountState>> {
```
barakshani: This name is confusing. Why not `get_account_state_by_serial()`?

crates/services/stf-runner/src/stf.rs
```diff
@@ -0,0 +1,202 @@
...
+    state_accessor.set_slot(block.signed_header().header().slot());
+
+    // Set accounts root
+    state_accessor.set_accounts_root(ledger_provider.root()?);
+
+    // Check state root
+    let new_state = state_accessor.get_toplevel_state().clone();
+    let state_root = new_state.compute_root();
+
+    // NOTE: Need to check state root matches or not. That should happen outside this.
```
barakshani: Why outside?

crates/chaintsn/src/context.rs
bewakes: This is being used at a bunch of places. Will remove it gradually.

crates/services/stf-runner/src/ledger.rs
bewakes: I guess I was overthinking on this. We don't need this.

crates/services/stf-runner/src/block.rs
bewakes: There's no reason :), I just didn't notice this can be simplified. Will update, thanks.

crates/services/stf-runner/src/block.rs
bewakes: Yep, good catch. Updated.

crates/services/stf-runner/src/ledger.rs
bewakes: Yes, the `consume_messages` is not required actually. The only message that the impl deals with is the deposit messages. All other inter-EE messages are ignored.

crates/services/stf-runner/src/state.rs
bewakes: I think it is still best to abstract out the accounts even if we won't have tree. The accounts root can simply be hash of the account serialization.
Also, i don't think we should keep state_root in the state itself. Maybe a prev_state_root, but not sure - can add if needed.

crates/services/stf-runner/src/state.rs
bewakes: Won't that be redundant? We can compute it from accounts, right? We can have some provider interface that will cache the total if needed, but having that in state feels redundant.

crates/services/stf-runner/src/state.rs
bewakes: yes, once ssz is there the root will be simpler. This is until then.

crates/services/stf-runner/src/ledger.rs
bewakes: This is an abstraction. We compute the accounts root here, or if we have a trie then we could just return the trie root from here.

crates/services/stf-runner/src/ledger.rs
bewakes: That's mostly because of the data structure. This uses a HashMap which doesn't guarantee the existence of a key.

crates/services/stf-runner/src/ledger.rs
bewakes: Totally agree. This is what happens when you try to shorten names :) Updated.

crates/services/stf-runner/src/stf.rs
bewakes: The idea is that block producer also needs to call this. Until the block is processed, it cannot know the state root and hence can't be compared here.

crates/services/stf-runner/src/ledger.rs
bewakes: No, I mean "accounts ledger". The way I am thinking and naming here is that we have a ledger(collection/book) of accounts. And this interface is to interact(perform CRUD) with that ledger.

crates/services/stf-runner/src/ledger.rs
bewakes: Done.

crates/services/stf-runner/src/ledger.rs
bewakes: Actually this is not necessary. I thought it would be convenient to hold messages separately. But yeah, this doesn't need to exist. Good catch.

bewakes:
> The demo runner crate should be replaced a bunch of unit tests in the stf crate(s)

Totally agree @delbonis. I'll be adding tests along the way as more clarity is gained -  at this stage of the PR these feel maintenance burden. The binary is something claude is very good at creating, I can just quickly see what's happening and can be safely removed after a point.

crates/services/stf-runner/src/service.rs
bewakes: I see, will keep that in mind. Can I ask what's the rationale for this being Sync and not Async?

crates/services/stf-runner/src/ledger.rs
barakshani: OK, is there any ledger which is not accounts ledger?

crates/services/stf-runner/src/ledger.rs
barakshani: Doesn't `sorted_accounts` above take care of that?

crates/services/stf-runner/src/state.rs
barakshani: Fair point, I see value in having this field where we can prove later on that it's the sum of BTC in all accounts (and separately each account will provide a proof that the sum of BTC in their (sub)accounts equal this number). But I agree that it isn't a hard necessity.

crates/services/stf-runner/src/state.rs
barakshani: 1. I meant changing the name.
2. I sound like the root isn't really a root, just hash, as I ask [here](https://github.com/alpenlabs/alpen/pull/1036#discussion_r2353407937). What do I miss?

crates/services/stf-runner/src/stf.rs
barakshani: ok, but then the block header isn't set yet, which you access (validate) above.

crates/services/stf-runner/src/ledger.rs
bewakes: I see what you meant. Yeah, sorting the keys should do it. My bad :) Fixed.

crates/services/stf-runner/src/stf.rs
bewakes: true, I've now separated out the block header validation and block execution. The flow is:
~`validate_continuity_sig_and_body_root` -> `execute_block` -> `validate_state_and_logs_root`~
`block.pre_exec_validate` -> `execute_block` -> `block.post_exec_valiadate`.

crates/services/stf-runner/src/ledger.rs
barakshani: Just to be clear, I didn't mean to ask to change this. I was just curious to understand why it's needed. I don't think having this extra check is bad.

crates/services/stf-runner/src/ledger.rs
bewakes: A ledger can be of anything right? The thing is, I am treating ledger as the collection of accounts. And it makes sense to define root for the whole collection and hence the term "accounts ledger".

barakshani: It's not that I disagree, but I think the main theme is let's try to keep things simple. This work is for OL and STF, not snark accounts. So I recommend we scope it according to that. WDYT?

crates/services/stf-runner/src/stf.rs
barakshani: I will need to understand this better. Eventually, I don't see a real logical problem here, only with naming (and potentially some struct). But I think that it will be great to clarify it on the spec first.

crates/services/stf-runner/src/ledger.rs
bewakes: I agree on the simplification part. What's happening is that the stf requires this interface and I need to give it a sensible name. :) 
There's also `StateAccessor` which serves similar purpose. At some point I'll merge them. Maybe simply `StateAccessor`/`StateProvider` will make more sense.
Till then I am keepin this as is. Will ask for review when that's done.

crates/services/stf-runner/src/state.rs
bewakes: Thanks for the clarification.
1. Even though we will have single account and a trivial implementation of account ledger, defining things based on the end-goal concepts and terminologies will help us in an easy transition. Unless it is very confusing.
2. There are two aspects to an OL state: account related(all accounts related stuffs) and non account related(like curr slot, epoch, etc). The accounts summary/commitment is the `account root` because keeping all accounts in state would be bulky. Now, the state root is some commitment of `account_root` + serialized(non-account data). So, `account_root` and `state_root` are different. Even though we say root, they would probably be some form of hash.

crates/services/stf-runner/src/state.rs
bewakes: yeah, we can add it when we feel its necessity.

crates/services/stf-runner/src/service.rs
delbonis: all of the DB calls are blocking and IO bound, so there's literally nothing gained by being async

PR #1038, opened 2025-09-17
STR-1297: Make ASM state progress within CSM machinery

evgenyzdanovich:
## Description

initial **ASM** integration. 

**Includes**:

- implement ASM as a service.
- run ASM service in `strata-client` alongside existing CSM.
- define RocksDB and SLED tables for ASM persistence.
- define minimal interface to notify subprotocols (via subprotocol service handle).

**Doesn't include**:

- versioning
- storage of aux inputs

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

      in the body of this PR.

### DIFF
--- a/functional-tests/tests/dbtool/revert_checkpointed_block_fn.py
+++ b/functional-tests/tests/dbtool/revert_checkpointed_block_fn.py
@@ -173,7 +173,7 @@ class RevertCheckpointedBlockFnTest(FullnodeDbtoolMixin):
         self.follower_1_node.start()
 
         # Wait for block production to resume to sync
-        seq_waiter.wait_until_chain_tip_exceeds(old_seq_ol_block_number + 1, timeout=30)
+        seq_waiter.wait_until_chain_tip_exceeds(old_seq_ol_block_number + 1, timeout=120)
 
         # Wait for full node to catch up to sequencer
         fn_waiter = self.create_strata_waiter(self.follower_1_rpc)


bin/strata-client/src/main.rs
```diff
@@ -332,12 +333,18 @@ fn start_core_tasks(
...
         &storage,
+        bitcoin_client.clone(),
         engine.clone(),
         params.clone(),
         status_channel.clone(),
     )?
     .into();
 
+    // Creating a combined block consumer both for CSM and for ASM, so they are in-sync.
+    // TODO: Ideally, ASM service would use it's own reader task with its own reader config.
```
barakshani:
```suggestion
    // TODO: Ideally, ASM service would use its own reader task with its own reader config.

File: crates/asm/worker/src/builder.rs [hunk 2]
```diff
@@ -0,0 +1,113 @@
...
+    service::AsmWorkerService,
+    state::AsmWorkerServiceState,
+    traits::WorkerContext,
+};
+
+/// Builder for constructing and launching an ASM worker service.
+///
+/// This encapsulates all the initialization logic and dependencies needed to
+/// launch an ASM worker using the service framework, preventing impl details
+/// from leaking into the caller.  The builder launches the service and returns
```
barakshani:
```suggestion
/// from leaking into the caller. The builder launches the service and returns

File: crates/asm/worker/src/service.rs [hunk 3]
```diff
@@ -0,0 +1,110 @@
...
+            skipped_blocks.push((block, pivot_block));
+
+            // Update the loop state.
+            pivot_anchor = ctx.get_anchor_state(&parent_block_id);
+            pivot_block = parent_block_id;
+        }
+
+        // We reached the height before genesis (while traversing), but didn't find genesis state.
+        if pivot_block.height() < genesis_height {
+            warn!("ASM haven't found pivot anchor state at genesis.");
```
barakshani:
```suggestion
            warn!("ASM hasn't found pivot anchor state at genesis.");

File: crates/consensus-logic/src/sync_manager.rs [hunk 4]
```diff
@@ -171,3 +195,30 @@ fn spawn_chain_worker(
...
+    // blocks while following canonical chain (and "canonicity" of l1 chain is imposed by the l1
+    // block manager).
+    let context = AsmWorkerCtx::new(
+        handle.clone(),
+        bitcoin_client,
+        storage.l1().clone(),
+        storage.asm().clone(),
+    );
+
+    // Use the new builder API to launch the worker and get a handle
```
barakshani:
```suggestion
    // Use the new builder API to launch the worker and get a handle.

File: crates/db-store-rocksdb/src/asm/schemas.rs [hunk 5]
```diff
@@ -0,0 +1,10 @@
+use strata_primitives::l1::L1BlockCommitment;
+use strata_state::asm_state::AsmState;
+
+use crate::{define_table_with_seek_key_codec, define_table_without_codec, impl_borsh_value_codec};
+
+// ASM state per block schema and corresponding codecs implementation
```
barakshani:
```suggestion
// ASM state per block schema and corresponding codecs implementation.

File: crates/db-store-sled/src/asm/schemas.rs [hunk 6]
```diff
@@ -0,0 +1,10 @@
+use strata_primitives::l1::L1BlockCommitment;
+use strata_state::asm_state::AsmState;
+
+use crate::{define_table_with_seek_key_codec, define_table_without_codec, impl_borsh_value_codec};
+
+// ASM state per block schema and corresponding codecs implementation
```
barakshani:
```suggestion
// ASM state per block schema and corresponding codecs implementation.

File: crates/asm/common/src/state.rs [hunk 7]
```diff
@@ -12,7 +12,7 @@ use crate::{AsmError, Mismatched, Subprotocol, SubprotocolId};
 /// receiving protocol transactions at L1 and updating its storage. A zk-SNARK proof
 /// attests that the transition from the previous ASM state to the new state
 /// was performed correctly on the given L1 block.
-#[derive(Clone, Debug, BorshSerialize, BorshDeserialize)]
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize, PartialEq)]
```
prajwolrg: nit (derive order): use built-in derives first before those that come from external crates.

crates/asm/worker/src/state.rs
```diff
@@ -0,0 +1,170 @@
...
+            .iter()
+            .map(|(k, v)| {
+                (
+                    *k,
+                    AuxPayload {
+                        data: v.data().to_vec(),
+                    },
+                )
+            })
+            .collect();
```
prajwolrg, line 117: We'd have to handle the aux request properly, but I think it's fine to skip it for now and tackle it separately later.

crates/state/src/asm_state.rs
```diff
@@ -0,0 +1,31 @@
+//! State bookkeeping necessary for ASM to run.
+
+use borsh::{BorshDeserialize, BorshSerialize};
+use strata_asm_common::{AnchorState, AsmLogEntry};
+use strata_asm_stf::AsmStfOutput;
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize, PartialEq)]
```
prajwolrg: nit: break lines.

crates/state/src/asm_state.rs
```diff
@@ -0,0 +1,31 @@
...
+pub struct AsmState {
+    state: AnchorState,
+    logs: Vec<AsmLogEntry>,
+}
```
prajwolrg, line 13: I don't think we require logs for the ASM to run.

crates/storage/src/ops/asm.rs
```diff
@@ -0,0 +1,17 @@
...
+use strata_db::traits::*;
+use strata_primitives::l1::L1BlockCommitment;
+use strata_state::asm_state::AsmState;
+
+use crate::exec::*;
+
+inst_ops_simple! {
+    (<D: AsmDatabase> => AsmDataOps) {
+        put_asm_state(block: L1BlockCommitment, state: AsmState) => ();
+        get_asm_state(block: L1BlockCommitment) => Option<AsmState>;
```
prajwolrg, line 14: I think it might be useful to put the AsmLog and AnchorState separately.

crates/db-store-sled/src/asm/schemas.rs
```diff
@@ -0,0 +1,10 @@
...
+define_table_with_seek_key_codec!(
+    /// A table to store ASM state per l1 block.
+    (AsmBlockSchema) L1BlockCommitment => AsmState
+);
```
prajwolrg, line 16: Maybe separate out the schema for logs and the state.

delbonis (changes requested):
This is really unfortunate, there's been some serious miscommunication here, I'm not sure where that happened.

At this level, subprotocols are completely encapsulated within the ASM infrastructure.  CSM doesn't need to be aware of them, they are not separate workers, so the `CommandHandle` doesn't exist, there is not a separate ASM worker.  There doesn't need to be a separate `AsmManager` or ASM DB types because it should be a field inside the CSM state.

The way it should have worked it that the CSM state struct would be extended like this:

```rs
pub struct ClientState {
    // Last *finalized* checkpoint.
    pub(crate) last_finalized_checkpoint: Option<L1Checkpoint>,

    // Last *seen* checkpoint.
    pub(crate) last_seen_checkpoint: Option<L1Checkpoint>,

    /// ASM state computed as of this block.
    pub(crate) asm_state: AnchorState, // sorry the name is weird, we could change that
}
```

And then most of the job of the CSM STF is:

* looking at the old state, figuring out what aux input there needs to be ("preprocessing", probably empty for now)
* invoking the ASM with that (`compute_asm_transition`)
* responding to emitted logs to update checkpoint fields and maybe other things in the future

Then the CSM STF output which gets persisted would include the new CSM state (containing the new ASM state) and the ASM aux input (which we'll later use for generating ASM proofs).

So this PR has a lot more moving parts than was necessary.

evgenyzdanovich:
@delbonis I was under impression that it's better to create a separate ASM worker because it seems that part of CSM (specifically, `ClientState` and its stf) will eventually be replaced by Checkpoint Subprotocol (and thus, we will need some other suitable "driver"). 
Also, the notion of "following l1 chain" is kinda duplicated as well (in both the CSM and ASM worker).

Once we are ready, we just create a (thin) subprotocol service that will do "responding to emitted logs to update checkpoint fields/ DB/signal on the status channel/communication to FCM/etc". This subprotocol service will be registered and data to it is passed by this ASM worker service.

P.S. Current impl already supports registering subprotocol handles(`crates/asm/worker/src/builder.rs`) where logs/state are passed to subprotocol service/task.


-------------------
A random philosophical note. 
The approach I had in mind allows us building from the ground up and switch parts along the way (but there can be some boilerplate/duplication). 
If I understand you correctly now, what you are proposing is to start growing ASM more tighly coupled to CSM and do more like refactors along the way (which, IMO, is a trickier thing to do).

Not a strong proponent of any approach, as long as we agree on it within the team (and ideally before impl - although that's not always possible:)

crates/asm/worker/src/service.rs
```diff
@@ -0,0 +1,110 @@
...
+        Ok(Response::Continue)
+    }
+}
+
+/// Status information for the ASM worker service.
+#[derive(Clone, Debug, Serialize)]
+pub struct AsmWorkerStatus {
+    pub is_initialized: bool,
+    pub cur_block: Option<L1BlockCommitment>,
+}
```
bewakes, line 111: This module is missing unit tests, particularly around `process_input`.

crates/db-store-rocksdb/src/asm/db.rs
```diff
@@ -0,0 +1,54 @@
...
+        iterator.seek_to_last();
+        let opt = iterator.rev().next().map(|res| res.unwrap().into_tuple());
+
+        Ok(opt)
+    }
+}
+
+#[cfg(feature = "test_utils")]
+#[cfg(test)]
+mod tests {}
```
bewakes: I would suggest to add tests here unless there's a good reason to defer it to some other PR.

crates/db-store-sled/src/asm/db.rs
```diff
@@ -0,0 +1,29 @@
...
+        Ok(self.asm_state_tree.get(&block)?)
+    }
+
+    fn get_latest_asm_state(&self) -> DbResult<Option<(L1BlockCommitment, AsmState)>> {
+        // Relying on the lexicographical order of L1BlockCommitment.
+        let mut iter = self.asm_state_tree.iter().rev();
+        let res = iter.next();
+        Ok(res.transpose()?)
+    }
+}
```
bewakes, line 56: I would suggest to add tests here unless there's a good reason to defer it to some other PR.

bewakes (changes requested):
I was also under the assumption that there would be csm re-structured as a service and that it would internally run ASM stf.

My main suggestions are regarding units tests that some of the important modules are missing.

crates/asm/worker/src/state.rs
evgenyzdanovich: okay, cool

crates/state/src/asm_state.rs
evgenyzdanovich: Well, that's just an "umbrella" state handy for bookkeeping / passing data around

crates/storage/src/ops/asm.rs
evgenyzdanovich: I agree. However, putting it separately means there potentially can be a case when state exists in the DB while logs are absent (or vice verca). This is quite annoying to handle. 
With `AsmState` in the ops interface (instead of bare `AnchorState` and `Vec<AsmLogEntry>`) this is achievable.

crates/asm/worker/src/service.rs
evgenyzdanovich: I'll add tests for `process_input` later (once we fully integrate checkpoint v0 subprotocol).

crates/db-store-rocksdb/src/asm/db.rs
evgenyzdanovich: done

crates/db-store-sled/src/asm/db.rs
evgenyzdanovich: done

crates/db-store-sled/src/asm/schemas.rs
evgenyzdanovich: separating schemas is fine, done.

crates/state/src/asm_state.rs
evgenyzdanovich: done

evgenyzdanovich:
@delbonis I re-read your "New service primitive" doc and modified the ASM service's status (also removed explicit subprotocol handlers), so that async code can listen to ASM log/state updates (as specified in the doc).

crates/asm/worker/src/handle.rs
```diff
@@ -1,24 +1,39 @@
...
+use crate::{AsmWorkerService, WorkerContext};
+
+type Monitor<W> = ServiceMonitor<AsmWorkerService<W>>;
+
 /// Handle for interacting with the ASM worker service.
 #[derive(Debug)]
-pub struct AsmWorkerHandle {
+pub struct AsmWorkerHandle<W: WorkerContext + Send + Sync + 'static> {
     command_handle: CommandHandle<L1BlockCommitment>,
+    service_monitor: Monitor<W>,
```
delbonis, line 15: This is super ugly, I should have converted this like I did the command handle type to just use the status type, so that the monitor type doesn't have to itself be generic over the context (and make all the consumers be aware of the concrete context type).  It's fun and quirky to have it be generic over the service but it's actually cumbersome when the service itself is generic.  Maybe I should have reworked how we'd use a context trait instead, but that sounds like it's getting close to needing HKTs.

crates/state/src/asm_state.rs
```diff
@@ -31,3 +33,33 @@ impl AsmState {
         &self.state
     }
 }
+
+/// Serde serialization of logs as hex strings.
+mod asm_log_entry_vec_hex {
```
delbonis: Is it not enough to add some attrs to `AsmLogEntry`?

delbonis:
@evgenyzdanovich 

I'm not against rewriting "the stuff that will go around the ASM STF invocation itself" entirely.  I can see the value in having a more clean foundation to build on top of.

> Logs are stored in DB and interpreted by the subprotocol service logic (additional, separate async task alongside subprotocol state which is internal to ASM). 

I agree we shouldn't be interpreting things like bridge subproto logs in this service.  But at the very least we'd need to consume the vk change upgrade log and record that outside of the `AnchorState` type itself to correctly control execution going forwards.  So that's why I was pointing towards reusing/extending this here.  I believe there would be similar bookkeeping needs for how we'd eventually, although I am not sure where that left off after the initial implementation.

So my thinking was if we're going to have that anyways, we might as well use it for other essential components like taking the checkpoint subproto logs and remember those as part of the same lifecycle, so we can make it available and not need to run another service just to track that.

The bridge related messages it would make more sense to run as a separate service because it's more "on the side", so I concur on that point.

> But logic paths are different oftentimes: for example btcio internals are different (CSM currently consumes L1BlockManifest which is in an old format, ASM consumes pure BTC blocks and does parsing of sps-50 under the hood).

Yeah this does have some messiness around the filters that are irrelevant now.

However, I had suggested that we should maintain this L1BlockManifest type but just incrementally convert it over to having SPS-50 tagged txs.  I had originally drafted all the spec docs for the ASM around taking full blocks but I think it makes more sense in practice to separate it out into two phases, one where we scrape the txs (is really fast and doesn't need the reset of the CSM state) and then the second where we really invoke the main ASM execution.  This would necessitate the proof being two steps.

> Sorry, I don't see any coupling here, could you please elaborate? ASM-as-a-service runs its STF (consuming pure btc blocks) and storing/making an output opaque asm logs.

Maybe this was changed but I was seeing what looked like "subprotocol command handles".  I interpreted this as having a service for each subprotocol that responds to the logs it produces.

> in this discussion I frankly hold only one strong opinion - ideally I'd like to see CSM being a service eventually in the end.

Completely agree here.

delbonis (changes requested):
^

crates/asm/worker/src/handle.rs
evgenyzdanovich: I vaguely plan on working around "integrating CSM with ASM logs" later, together I'll adjust MonitorService (the way it's done in CommandHandle), this will improve the ergonomics a bit.

crates/state/src/asm_state.rs
evgenyzdanovich: yeah, added.

evgenyzdanovich:
@delbonis thanks for the feedback.

>But at the very least we'd need to consume the vk change upgrade log and record that outside of the AnchorState type itself to correctly control execution going forwards.
Agree. This is pretty much the first impl, we'll iterate on that soon.

>So my thinking was if we're going to have that anyways, we might as well use it for other essential components like taking the checkpoint subproto logs and remember those as part of the same lifecycle, so we can make it available and not need to run another service just to track that.
well, this is (now) makes sense to me. I think in the (near) future we can end up having what you described (in some shape).

> Maybe this was changed but I was seeing what looked like "subprotocol command handles". I interpreted this as having a service for each subprotocol that responds to the logs it produces.

Yeah, I removed it (well, after re-reading your initial doc) and made sure the data is available through service monitor.

crates/asm/worker/src/state.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub(crate) anchor: Option<AsmState>,
+
+    /// Current anchor block.
+    pub(crate) blkid: Option<L1BlockCommitment>,
+
+    /// ASM spec for ASM STF.
+    asm_spec: StrataAsmSpec,
+
+    /// A runtime handle.
+    _runtime_handle: Handle,
```
delbonis: Can we remove this now?

crates/state/src/lib.rs
```diff
@@ -23,18 +24,53 @@ pub mod operation;
...
+
+        Ok(())
+    }
+}
+
+impl CombinedBlockSubmitter {
+    pub fn new(submitters: Vec<std::sync::Arc<dyn BlockSubmitter>>) -> Self {
+        Self { submitters }
+    }
 }
```
delbonis, line 74: *I think* would be better to reverse the relationship between the ASM worker and the L1 reader.  The L1 reader updates its status to include the new block tip and the ASM worker's `Input` monitors for those state updates and follows the chain being published there, processing blocks as needed.

I may be misunderstanding something, but what use does combining multiple block submitters have here?

crates/storage/src/lib.rs
```diff
@@ -20,6 +20,7 @@ use strata_db::traits::DatabaseBackend;
 #[derive(Clone)]
 #[expect(missing_debug_implementations)]
 pub struct NodeStorage {
+    asm_manager: Arc<AsmManager>,
```
delbonis: Should this be `AsmStateManager`/`asm_state_manager`?

crates/state/src/lib.rs
evgenyzdanovich: > I may be misunderstanding something, but what use does combining multiple block submitters have here?

to not spawn a separate btcio task and reuse the existing data channel (for CSM). The reason is `btcio` reading still needs some additional changes (unrelated to this PR), so I decided not touch that part of the code yet and just reuse the same piece. *Soon*, it'll definitely go away (it's currently only used in sync_manager alongside csm).

> I think would be better to reverse the relationship between the ASM worker and the L1 reader

Strongly agree. I have some plans to address it in the future.

crates/asm/worker/src/state.rs
evgenyzdanovich: done

crates/storage/src/lib.rs
evgenyzdanovich: done

evgenyzdanovich:
> needing wrapping container to track STF version, remembering computed aux inputs, etc.

Thanks, I left a TODO to address that. It's much easier to work on separately later. Right now, the main objective is to fully integrate checkpoint v0 with csm (which doesn't require neither of those things) and the current impl roughly suffice.

I'd much rather prefer to shift much of our discussions to next PR(s) and have polishing there, but merge this one soon(er) so we can start working on the actual integration (CSM consuming ASM logs).

prajwolrg (approved):

crates/state/src/lib.rs
delbonis: ok great, will approve now

delbonis (approved):

PR #1039, opened 2025-09-18
feat(btcio): sps-50 compatibility in btcio

irnb:
## Description

Crates:
- ```crates/asm/subprotocols/checkpointing-v0```
- ```crates/asm/txs/checkpointing```
- ```crates/btcio/```

This PR introduces an SPS-50–compatible `btcio` writer, removes the old parsing logic from the `btcio` reader, and implements `checkpointing-v0`.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

      in the body of this PR.

## Related Issues

STR-1724


crates/test-utils/tx-indexer/src/lib.rs
```diff
@@ -1,5 +1,7 @@
 //! Transaction indexing tests and helpers.
 
+// TODO: Update this crate so the module works with the ASM + SPS-50 tag flow, allowing it to drive
+// ASM-focused tests.
```
prajwolrg, line 4: For the TODO, let's create a ticket and add the link here.

crates/asm/txs/checkpointing/src/tag.rs
```diff
@@ -0,0 +1,21 @@
...
+        CHECKPOINTING_V0_SUBPROTOCOL_ID,
+        OL_STF_CHECKPOINT_TX_TYPE,
+        &[],
+    )
+    .expect("checkpoint tag encoding");
+
+    config
+        .encode_tag_buf(&tag)
+        .expect("checkpoint tag encoding")
+}
```
prajwolrg: This should be outside of here. Maybe this can be moved into `strata-l1tx` instead. It can take that as the argument, or it might also makes sense in the `strata-common` repo.

crates/btcio/src/reader/event.rs
```diff
@@ -25,16 +24,16 @@ pub(crate) struct BlockData {
     // TODO remove?
     block: Block,
 
-    /// Transactions in the block that contain protocol operations
-    relevant_txs: Vec<RelevantTxEntry>,
+    /// Transaction indexes in the block that contain SPS-50 tags
+    tagged_tx_indices: Vec<u32>,
```
prajwolrg, line 27: This feels weird. We need to handle this in a better way. I imagine eventually the BlockData will be SPS-50 parsed transactions. 

Since this is meant to be stepping stone, I think it's fine, but we should add a TODO with a new ticket.

crates/btcio/src/writer/builder.rs
```diff
@@ -669,13 +706,15 @@ mod tests {
         .unwrap(); // should be 33 bytes
 
         let inp_txn = get_txn_from_utxo(utxo, &ctx.sequencer_address);
+        let header = vec![0xAAu8, 0xBB, 0xCC];
```
prajwolrg, line 719: `header` is confusing. I think this should be envelope tag.

crates/asm/subprotocols/checkpointing-v0/src/types.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// Sequencer public key for signature verification
+    pub sequencer_pubkey: Buf32,
+    /// Whether to skip proof verification for testing (current system compatibility)
+    pub skip_proof_verification: bool,
+    /// Genesis L1 block commitment
+    pub genesis_l1_block: L1BlockCommitment,
+    /// Rollup verifying key for proof verification
+    /// Optional to support testing environments without proof verification
+    pub rollup_verifying_key: Option<strata_primitives::proof::RollupVerifyingKey>,
+}
```
prajwolrg, line 53: Few points on this:
1. It makes sense to put the `sequencer_pubkey` and  `rollup_verifying_key` in the state as we want to support updating that even in v1. (is fine for v0).
2. I think we can also make the `rollup_verifying_key` non-optional. 
3. Also related with (2). We can also remove `skip_proof_verification`. We already have Native in the `RollupVerifyingKey` enum where the proof verification is already skipped. We will most likely be replacing the verifying key with the predicate, so it should be fine to skip that as well.

crates/asm/subprotocols/checkpointing-v0/src/verification.rs
```diff
@@ -0,0 +1,162 @@
...
+fn verify_checkpoint_signature(
+    _signed_checkpoint: &SignedCheckpoint,
+    _expected_pubkey: &strata_primitives::buf::Buf32,
+) -> Result<bool, CheckpointV0Error> {
+    // TODO: Implement actual signature verification
+    // For now, accept all signatures for feature parity testing
+    // In real implementation, this would:
+    // 1. Extract signature and pubkey from signed checkpoint
+    // 2. Verify signature against checkpoint data
+    // 3. Check pubkey matches expected sequencer key
```
prajwolrg: Even for feature parity, I think we are already checking the signature verification.

crates/asm/txs/checkpointing/src/tag.rs
evgenyzdanovich: agree here.

crates/btcio/src/reader/event.rs
evgenyzdanovich: I don't think it's necessary as a field (BlockData is not persisted, unlike `L1TxManifest`).
`tagged_tx_indices` can be computed in-place (in `generate_l1txs`), `generate_l1txs` can be extented with `ctx: &ReaderContext<R>` (to fetch the magic bytes and construct the parser in place).

crates/asm/subprotocols/checkpointing-v0/src/subprotocol.rs
```diff
@@ -0,0 +1,153 @@
...
+
+/// Process a single checkpoint transaction (v0 implementation)
+///
+/// This function implements the core checkpoint transaction processing:
+/// 1. Extract signed checkpoint from SPS-50 envelope (following admin pattern)
+/// 2. Create verification context from transaction and anchor state
+/// 3. Delegate to checkpoint verification logic
+/// 4. Extract and forward withdrawal messages if checkpoint is accepted
+///
+/// NOTE: This bridges current checkpoint format with SPS-50 envelope parsing
```
barakshani: What exactly do you mean by that? Does the current checkpoint structure follow SPS-50?

crates/asm/subprotocols/checkpointing-v0/src/verification.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    if expected_output != actual_output {
+        logging::warn!(
+            epoch = checkpoint.batch_info().epoch(),
+            "Checkpoint proof public values mismatch"
+        );
+        return Err(CheckpointV0Error::InvalidProof);
+    }
+
+    let is_empty_proof = proof_receipt.proof().is_empty();
```
barakshani, line 98: I know this code will be changed, but still, we should make sure this code will not end up in production, so let's add a big TODO to remove this.

barakshani, line 89: I know this code will be changed, but still, we should make sure this code will not end up in production, so let's add a big TODO to remove this.

crates/asm/subprotocols/checkpointing-v0/src/verification.rs
```diff
@@ -0,0 +1,156 @@
...
+        verify_rollup_groth16_proof_receipt(&proof_receipt, &state.rollup_verifying_key)
+    {
+        logging::warn!("Groth16 verification failed: {err:?}");
+        return Err(CheckpointV0Error::InvalidProof);
+    }
+
+    Ok(())
+}
+
+fn verify_state_transition(
```
barakshani: There is actually no state transition here, right? We just check that the previous epoch (as declared in the checkpoint) ends where this epoch starts. Maybe something like _verify_continuous_transition_? or _states_? or _consistent_transition_?

crates/asm/subprotocols/checkpointing-v0/src/subprotocol.rs
```diff
@@ -0,0 +1,153 @@
...
+    let signed_checkpoint = extract_signed_checkpoint_from_envelope(tx)?;
+
+    // 2. Process checkpoint using v0 verification logic
+    process_checkpoint_v0(state, &signed_checkpoint, current_l1_height)?;
+
+    // 3. Extract withdrawal messages for bridge forwarding
+    let withdrawal_intents = extract_withdrawal_messages(signed_checkpoint.checkpoint())
+        .map_err(|e| CheckpointV0Error::StateTransitionError(e.to_string()))?;
+
+    // Forward each withdrawal message to the bridge subprotocol
```
barakshani, line 167: This is interesting! The OL already "knows" about this withdrawal intent, right? Is this needed then _only_ for the bridge node itself (that eventually will also run ASM)?

crates/asm/subprotocols/checkpointing-v0/src/types.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    /// Update state with a newly verified checkpoint
+    pub fn update_with_checkpoint(&mut self, checkpoint: PrimitivesCheckpoint, l1_height: u64) {
+        let epoch = checkpoint.batch_info().epoch();
+        self.last_checkpoint = Some(checkpoint);
+        self.last_checkpoint_l1_height = l1_height;
+        self.current_verified_epoch = epoch;
+    }
+
+    /// Get the current epoch we've verified
```
barakshani:
```suggestion
    /// Get the latest verified epoch

File: crates/asm/subprotocols/checkpointing-v0/src/types.rs [hunk 12]
```diff
@@ -0,0 +1,118 @@
...
+    pub fn current_epoch(&self) -> u64 {
+        self.current_verified_epoch
+    }
+
+    /// Get the epoch value we expect from the next checkpoint.
```
barakshani:
```suggestion
    /// Get the epoch value we expect for the next checkpoint.

File: crates/asm/subprotocols/checkpointing-v0/src/types.rs [hunk 13]
```diff
@@ -0,0 +1,118 @@
...
+    /// # Returns
+    /// `true` if the epoch can be accepted, `false` otherwise
+    pub fn can_accept_epoch(&self, epoch: u64) -> bool {
+        if self.last_checkpoint.is_none() {
+            epoch == 0
+        } else {
+            epoch == self.current_verified_epoch + 1
+        }
+    }
+}
```
barakshani, line 112: Can't this be replaced with checking if `epoch == expected_next_epoch()` you've just defined?

crates/btcio/src/reader/event.rs
```diff
@@ -24,16 +23,16 @@ pub(crate) struct BlockData {
     // TODO remove?
```
storopoli, line 23: Maybe remove this TODO comment? Since we're not storing the `RelevantTxEntry` anymore?

crates/btcio/src/writer/builder.rs
```diff
@@ -93,6 +103,12 @@ pub(crate) async fn build_envelope_txs<R: Reader + Signer + Wallet>(
     payloads: &[L1Payload],
     ctx: &WriterContext<R>,
 ) -> anyhow::Result<(Transaction, Transaction)> {
+    // SPS-50 envelopes currently carry a single protocol transaction, so btcio enforces
+    // exactly one payload until multi-payload envelopes are supported.
+    if payloads.len() != 1 {
+        return Err(EnvelopeError::InvalidPayloadCount(payloads.len()).into());
+    }
+
```
storopoli, line 111: Why not take a stack allocated array then?
`s/payloads: &[L1Payload]/payload: [L1Payload; 1]/`
Or even a single `&L1Payload`?

crates/btcio/src/writer/context.rs
```diff
@@ -3,11 +3,14 @@ use std::sync::Arc;
 use bitcoin::Address;
 use bitcoind_async_client::traits::{Reader, Signer, Wallet};
 use strata_config::btcio::WriterConfig;
-use strata_primitives::params::Params;
+use strata_primitives::{l1::payload::L1Payload, params::Params};
 use strata_status::StatusChannel;
 
+/// Callable that produces SPS-50 tag bytes for the reveal transaction OP_RETURN.
```
storopoli, line 9:
```suggestion
/// Callable that produces SPS-50 tag bytes for the reveal transaction `OP_RETURN`.
```

crates/btcio/src/writer/builder.rs
```diff
@@ -154,6 +154,10 @@ pub fn create_envelope_transactions(
         .finalize(SECP256K1, public_key)
         .map_err(|_| anyhow!("Could not build taproot spend info"))?;
 
+    let op_return_push = PushBytesBuf::try_from(envelope_tag.to_vec())
+        .map_err(|_| EnvelopeError::InvalidEnvelopeTag)?;
```
storopoli, line 158: I don't think you need to heap-allocate this into `Vec`.
`PushBytes` in `rust-bitcoin` has a [`TryFrom<&'a [u8]>`](https://docs.rs/bitcoin/0.32.7/bitcoin/blockdata/script/struct.PushBytes.html#impl-TryFrom%3C%26%5Bu8%5D%3E-for-%26PushBytes)

crates/btcio/src/writer/builder.rs
```diff
@@ -154,6 +154,10 @@ pub fn create_envelope_transactions(
...
+    let op_return_script = ScriptBuf::new_op_return(op_return_push.clone());
```
storopoli, line 159: Also here I don't think you need to clone.
The `new_op_return` function signature is:

```rs
pub fn new_op_return<T: AsRef<PushBytes>>(data: T) -> Self
```

bin/strata-client/src/main.rs
```diff
@@ -401,6 +406,26 @@ fn start_sequencer_tasks(
...
+
+    let tag_encoder: Arc<EnvelopeTagEncoder> =
+        Arc::new(move |payload| -> anyhow::Result<Vec<u8>> {
+            match payload.payload_type() {
+                L1PayloadType::Checkpoint => {
+                    let mut tag = Vec::with_capacity(6);
+                    tag.extend_from_slice(&magic_bytes);
+                    tag.push(CHECKPOINT_V0_SUBPROTOCOL_ID);
+                    tag.push(OL_STF_CHECKPOINT_TX_TYPE);
+                    Ok(tag)
```
storopoli, line 421: Don't allocate.
We have `make_buf!` that is a macro with compile-time stack allocations instead. See as an example: https://github.com/alpenlabs/alpen/blob/8e396c3e8dddac7eb99b53be66b40df1bcaa02d5/crates/util/python-utils/src/bridge/types.rs#L41-L50

PR #1052, opened 2025-09-23
Create common account and EE types

delbonis:
## Description

This just adds all of the basic account types that we will end up using everywhere and various EE-related types that are around the interface boundaries of various components.  There's also some minor refactors to the MMR code.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

- [X] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/.github/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR. (N/A)

### DIFF
--- a/crates/snark-acct-types/src/lib.rs
+++ b/crates/snark-acct-types/src/lib.rs
@@ -1,5 +1,5 @@
 //! Types relating to snark accounts and the snark account proof interface.
-#![expect(unused)] // in-development
+#![expect(unused, reason = "in development")] // in-development
 
 mod accumulators;
 mod messages;


crates/acct-types/src/id.rs
```diff
@@ -0,0 +1,41 @@
...
+/// Raw primitive version of an account ID.  Defined here for convenience.
+pub type RawAcctTypeId = u16;
+
+/// Distinguishes between account types.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
+#[repr(u16)]
+pub enum AcctTypeId {
+    Empty = 0,
+    Snark = 1,
+}
```
prajwolrg, line 57: Should probably add more context about what the empty accounts and snark accounts are and their usage.

My assumption is that the ledger accounts are `AcctTypeId::Snark` and system accounts are `AcctTypeId::Empty`.

sapinb (changes requested):
Just a nit, but can we use Account* instead of Acct* ?

crates/snark-acct-types/src/outputs.rs
```diff
@@ -0,0 +1,66 @@
+//! Account output types that get applied to the ledger.
+
+use strata_acct_types::{AcctId, MsgPayload};
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct UpdateOutputs {
+    transfers: Vec<OutputTransfer>,
+    messages: Vec<OutputMessage>,
+}
```
prajwolrg, line 10: Are these both converted to `MessageEntry` by the OL?

crates/acct-types/src/id.rs
```diff
@@ -0,0 +1,41 @@
...
+
+impl From<RawAcctTypeId> for AcctTypeId {
+    fn from(value: RawAcctTypeId) -> Self {
+        // SAFETY: this is safe, right?
```
bewakes: This is not because isn't `AcctTypeId::from(1000)` undefined? There's nothing that stops us from trying to create AcctTypeId with values > 2.

crates/ee-chain-types/src/block.rs
```diff
@@ -0,0 +1,89 @@
...
+/// Container for an execution block that signals additional data with it.
+// TODO better name, using an intentionally bad one for now
+// TODO SSZ
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ExecBlockNotpackage {
+    /// Execution blkid, which commits to things more cleanly.
+    exec_blkid: Hash,
+
+    /// Encoded block hash to more cleanly link things.
+    block_raw_hash: Hash,
```
sapinb: whats the difference between exec_blkid and block_raw_hash ? Which represents the exec blockhash ?
And shouldnt the blockhash commit to the entire block data anyway ?

crates/snark-acct-types/src/update.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    /// MMR proofs for each of the inbox messages we processed.
+    ///
+    /// These may be updated by the sequencer.
+    inbox_proofs: Vec<MessageEntryProof>,
+
+    /// MMR proofs for each piece of ledger data we referenced.
+    ///
+    /// These may be updated by the sequencer.
+    ledger_ref_proofs: LedgerRefProofs,
```
prajwolrg, line 164: Is the merkle proofs provided by the OL or the EE?

I think EE can just provide the `AccumulatorClaims` and the OL Sequencer is responsible for fetching all the Merkle Proofs. Then we don't need the proof to be updated by the sequencer.

crates/acct-types/src/id.rs
```diff
@@ -0,0 +1,41 @@
...
+/// Incrementally assigned account serial number.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct AcctSerial(u32);
+
+/// Identifier for a "subject" within the scope of an execution environment.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct SubjectId([u8; 32]);
+
+/// Raw primitive version of an account ID.  Defined here for convenience.
+pub type RawAcctTypeId = u16;
```
bewakes: There doesn't seem to be ways to create these types. These should either have `::new()`, or the wrapped item be public, or a `From` impl.

crates/acct-types/src/mmr.rs
```diff
@@ -0,0 +1,24 @@
+//! Concrete orchestration layer MMR types.
+
+use strata_mmr::Sha256Hasher;
+
+/// The basic hasher we use for all the MMR stuff.
+///
+/// This is SHA-256 with the full 32 byte hash.
+// TODO should this be blake3 and be only 20 bytes or something?
```
prajwolrg, line 8: Right now, most ZkVm have internal custom circuit for sha2 and sha3, but not for blake3.

crates/acct-types/src/state.rs
```diff
@@ -0,0 +1,127 @@
...
+/// Intrinsic account fields.
+// TODO SSZ
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct IntrinsicAcctState {
+    // immutable fields, these MUST NOT change
+    /// Account type, which determines how we interact with it.
+    ty: u16,
+
+    /// Account serial number.
+    serial: AcctId,
```
bewakes: You mean `AcctSerial`?

crates/acct-types/src/state.rs
```diff
@@ -0,0 +1,127 @@
...
+    errors::{AcctError, AcctResult},
+    id::{AcctId, AcctTypeId},
+    mmr::Hash,
+};
+
+type Root = Hash;
+
+/// Account state.
+// TODO SSZ
+// TODO builder
```
bewakes, line 12: Please add builder, or at least constructor methods.

bewakes (changes requested):
Looks good except some of the structs can't be created because they lack constructors/builders.

crates/snark-acct-types/src/outputs.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    pub fn messages(&self) -> &[OutputMessage] {
+        &self.messages
+    }
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct OutputTransfer {
+    dest: AcctId,
+    value: u64,
```
sapinb, line 37: OutputTransfer should also have a message payload or subject, to indicate subject or other address inside the destination.

crates/ee-chain-types/src/block.rs
delbonis: `exec_blkid` is the application-specific ID of the block, used to verify the header chain.  `raw_block_hash` is the hash of the *encoded* block, not application-specifically, used to verify that some blob of bytes if the block in question.

Strictly speaking we probably only need *one* for the proof protocol to be secure, but it makes it harder to write more generic code.

crates/acct-types/src/id.rs
delbonis: `Empty` is just a dummy value.  All of these refer to "ledger" accounts in the taxonomy.

crates/snark-acct-types/src/outputs.rs
delbonis:
```
class OutputTransfer(Container):
	"""
	Describes a simple value transfer from some account to another.
	"""
	destination: AcctId
	transferred_value: uint64
```

Only messages result in messages entries.

crates/snark-acct-types/src/update.rs
delbonis: Right now you are correct that it doesn't matter.

When we have p2p gossiped txs, these MUST be provided by the party sending the tx so that regular nodes are not expected to index all of the historical data to be able to verify transactions.  Apply all of the logic from the "stateless Ethereum" efforts to this with regard to verifying txs state proofs.

crates/acct-types/src/id.rs
delbonis: ah balls I didn't think about this, C leaves a lasting legacy on you

crates/snark-acct-types/src/outputs.rs
delbonis: Transfers a OL ledger level concept and are not messages.  They don't have data payloads and are not intended to be received by a subject.

crates/acct-types/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+//! Account system common type definitions.
+#![allow(unused)] // in-development
```
storopoli: Can you move this to
```suggestion
#![expect(unused)] // in-development
```

crates/acct-types/src/id.rs
```diff
@@ -0,0 +1,41 @@
+use std::{fmt, mem};
+
+/// Universal account identifier.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct AcctId([u8; 32]);
+
+/// Incrementally assigned account serial number.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct AcctSerial(u32);
```
storopoli: 3 things:

1. I don't see anywhere in this PR the `AccountSerial` being used or incremented.
2. Why not calling `AccountSequenceNumber`?
3. You should implement a `increment` function that `+=1` since this is an opaque type with a non-pub inner value.

crates/acct-types/src/id.rs
storopoli: In order to make it safe you need to do switch to a `TryFrom`... Yeah that's annoying I know. Or you can do very backwards bending type gymnastics.

crates/acct-types/src/messages.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    pub fn payload(&self) -> &MsgPayload {
+        &self.payload
+    }
+}
+
+/// Contents of a message, ie the data and sent value payload components.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct MsgPayload {
+    value: u64, // TODO convert to BitcoinAmount
```
storopoli: Putting a blocker in this TODO.

crates/acct-types/src/state.rs
```diff
@@ -0,0 +1,127 @@
...
+    intrinsics: IntrinsicAcctState,
+    encoded_state: Vec<u8>,
+}
+
+impl AcctState {
+    pub fn ty(&self) -> u16 {
+        self.intrinsics.ty()
+    }
+
+    pub fn serial(&self) -> AcctId {
```
storopoli: As @bewakes already pointed out this should be a `AcctSerial`

crates/acct-types/src/state.rs
```diff
@@ -0,0 +1,127 @@
...
+    // immutable fields, these MUST NOT change
+    /// Account type, which determines how we interact with it.
+    ty: u16,
+
+    /// Account serial number.
+    serial: AcctId,
+
+    // mutable fields, which MAY change
+    /// Native asset (satoshi) balance.
+    balance: u64,
```
storopoli: Should be a `BitcoinAmount`

crates/ee-chain-types/src/block.rs
storopoli: I'm lost...

storopoli: This should give rise to better docstrings... Because even with your explanation I'm utterly lost...

crates/snark-acct-types/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+//! Types relating to snark accounts and the snark account proof interface.
+#![allow(unused)] // in-development
```
storopoli:
```suggestion
#![expect(unused)] // in-development
```

crates/snark-acct-types/src/messages.rs
```diff
@@ -0,0 +1,72 @@
+//! Snark account types.
+
+use strata_acct_types::{AcctId, MsgPayload, RawMerkleProof};
+
+/// Message entry in an account inbox.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct MessageEntry {
+    /// The source account ID of the message.
+    source: AcctId,
```
storopoli: Do we need a `destination`?

crates/snark-acct-types/src/outputs.rs
```diff
@@ -0,0 +1,66 @@
...
+}
```
storopoli, line 38: Shouldn't this be a `BitcoinAmount`?

Cargo.toml
```diff
@@ -334,6 +339,7 @@ reqwest = { version = "0.12.7", default-features = false, features = [
   "zstd",
   "json",
 ] }
+rlp = "0.5.2"
```
storopoli, line 350: Can you please move this standalone entry to use the new `workspace` version?
https://github.com/alpenlabs/alpen/blob/e2a4138a3f18e375a7f26b6a7f24572f12cb93ae/crates/mpt/Cargo.toml#L17

storopoli (changes requested):
Some comments. Looks good so far!

crates/snark-acct-types/src/messages.rs
delbonis: we don't!

crates/snark-acct-types/src/outputs.rs
delbonis: Trying to avoid including huge deps like the primitives crate (which would eventually become a dep on this crate), so I've redefined a placeholder.

storopoli: Ok you can leave as a TODO or open a ticket? We can discuss refactoring some things out of primitives.

crates/acct-types/src/id.rs
delbonis: I had a couple places where I accidentally used AcctId instead of AcctSerial, so it was actually intended to be used.

I think it might actually make sense to add a seqno type, will add.

crates/acct-types/src/id.rs
delbonis: added a macro to help with all this

crates/acct-types/src/mmr.rs
delbonis: we'll figure this out later

crates/snark-acct-types/src/messages.rs
delbonis: (it's contextual)

crates/acct-types/src/amount.rs
```diff
@@ -0,0 +1,11 @@
+use crate::impl_thin_wrapper;
+
+/// Describes an amount of bitcoin.
+///
+/// This will eventually be replaced with the more general one, which I am not
+/// using here to avoid creating a dependency mess.
```
storopoli, line 8: This should not be a docstring

crates/acct-types/src/macros.rs
```diff
@@ -0,0 +1,29 @@
...
+            }
+        }
+
+        impl Into<$inner> for $target {
+            fn into(self) -> $inner {
+                self.0
+            }
+        }
+    };
+}
```
storopoli, line 33: Should also do `DeRef`/`DeRefMut`.

crates/ee-chain-types/src/lib.rs
```diff
@@ -3,7 +3,7 @@
 //! This is primarily at the boundary between the internal EE account state and
 //! the execution env chain.  These are not generally involved in the
 //! orchestration layer protocol.
-#![allow(unused)] // in-development
+//#![expect(unused)] // in-development
```
storopoli: remove this :)

crates/acct-types/src/macros.rs
storopoli: Ooof on second thought. You need `DeRef`/`DeRefMut` for `BitcoinAmmount` but not for the others that are opaque.
What about two macros: `impl_thin_opaque_wrapper` which is the current impl and `impl_thin_transparent_wrapper` which impls also `DeRef`/`DeRefMut` (you can come up with alternative names, I don't mind)

crates/acct-types/src/id.rs
```diff
@@ -55,6 +55,7 @@ pub enum AcctTypeId {
 }
 
 impl AcctTypeId {
+    // This MUST always be the last entry in `AcctTypeId`.
     const MAX_RAW: RawAcctTypeId = 1;
```
storopoli: Is there a way we can unit test this?
This will not age well

storopoli (changes requested):

crates/ee-chain-types/src/block.rs
sapinb: I didnt understand either

crates/acct-types/src/id.rs
prajwolrg: Is SystemAccount not differentiated by the `AcctTypeId`? I was basing my thoughts based on [this](https://www.notion.so/SPS-acct-sys-Strata-OL-account-system-1e4901ba000f803aa54cf6834883f466?source=copy_link#1e4901ba000f80feaaa2fc5adc32b27c).

storopoli:
> > That would start to make the type names get realllllyyyy loooonggg.
> 
> @delbonis We dont have an 8 byte limit of variable names :)

I prefer long and explicit variable names.

> Account* instead of Acct*

`Acct` can be confused with accumulator, especially when MMRs are close by.
The compiler will compile that away :)

crates/acct-types/src/id.rs
delbonis: Not cleanly.  Although what I actually should be using is the int_enum crate.

delbonis:
@storopoli I don't really see how acct could ever be confused for accumulator and while compiler will compile type names away but I only fit so many characters on my screen.  But I'll change it anyways since clearly it's objectionable.

crates/acct-types/src/id.rs
delbonis: System accounts are not ledger accounts so they don't have a state.

crates/acct-types/src/amount.rs
delbonis: I think it should be a docstring to clarify to users what to expect with it.

crates/acct-types/src/macros.rs
delbonis: added with new macros

PR #1062, opened 2025-09-25
backport (dbtool) : STR 1728 Fix strata-dbtool functional tests

krsnapaudel:
## Description

Backport of #1042 to `releases/0.2.0` branch. Some issues were identified in backport testing. The fixes are being merged to `main` as well.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

      in the body of this PR.


functional-tests/factory/factory.py
```diff
@@ -271,6 +271,11 @@ def create_fullnode(
         svc = flexitest.service.ProcService(props, cmd, stdout=logfile)
         svc.start()
         _inject_service_create_rpc(svc, rpc_url, name)
+
+        def _datadir_path() -> str:
+            return datadir
+
+        svc.datadir_path = _datadir_path
```
delbonis, line 278: We could just do `svc.datadir_path = datadir` instead of making it a thunk.

functional-tests/tests/dbtool/revert_chainstate_fn.py
```diff
@@ -0,0 +1,229 @@
...
+            timeout=30,
+        )
+
+        # Wait for full node to catch up to sequencer
+        wait_until_l2_synced_to_height(
+            self.follower_1_rpc,
+            old_seq_ol_block_number + 1,
+            error_with="expected blocks not produced after revert chainstate",
+            timeout=30,
+        )
```
delbonis: Can we use any of the waiter classes?

delbonis (changes requested):
It feels like there's a lot of duplication with the waits that could be consolidated.  A lot of these tests generally have a lot of shared behavior so I'm wondering if this could be consolidated into a base class that manages the lifecycle and then each test would inherit from that instead.

functional-tests/tests/dbtool/revert_chainstate_fn.py
krsnapaudel: This is backport to `releases/0.2.0`. The waiter classes are not there.

krsnapaudel:
> It feels like there's a lot of duplication with the waits that could be consolidated. A lot of these tests generally have a lot of shared behavior so I'm wondering if this could be consolidated into a base class that manages the lifecycle and then each test would inherit from that instead.

Some parts yes. But they are slightly different. Example: full node vs sequencer. Checkpointed block or not. Should pass or fail.

functional-tests/factory/factory.py
krsnapaudel: I realized that the value needs to be a function and not a string path.

PR #1065, opened 2025-09-25
log: Bitcoin `Buf32` types reverse hex

storopoli:
## Description

Some 32-byte bitcoin things are represented as `Buf32`.
That creates an issue when we log them because block explorers reverse the bytes as little-endian for things like TXID, WTXID and block hashes.

This PR fixes that by adding custom `Debug`/`Display` impls that reverse these bytes for these bitcoin tuple struct wrappers around `Buf32`.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The tickets STR-1478, STR-1494, STR-1498, and STR-1478 suggest creating a new `Buf32R` type.
But I disagree. The type Tetris is pretty solid in the codebase and I don't want to add more types to represent the same thing, zkVM-friendly 32-bytes.

Hence, I went with the custom `Debug`/`Display` impl with a quick dirty fix in `btcio` in 653bb2f4e3ffddf1f958df6c43d817f6c7c4b107

## Checklist

      in the body of this PR.

## Related Issues

STR-1478
STR-1494
STR-1478
STR-1498

### DIFF
--- a/crates/primitives/src/l1/block.rs
+++ b/crates/primitives/src/l1/block.rs
@@ -4,6 +4,7 @@ use arbitrary::Arbitrary;
 use bitcoin::{hashes::Hash, BlockHash};
 use borsh::{BorshDeserialize, BorshSerialize};
 use const_hex as hex;
+use hex::encode_to_slice;
 use serde::{Deserialize, Serialize};
 
 use crate::{buf::Buf32, hash::sha256d};
@@ -138,8 +139,11 @@ impl fmt::Debug for L1BlockId {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let mut bytes = self.0 .0;
         bytes.reverse();
-        let hex_str = hex::encode(bytes);
-        f.write_str(&hex_str)
+        let mut buf = [0u8; 64]; // 32 bytes * 2 for hex
+        encode_to_slice(bytes, &mut buf).expect("buf: enc hex");
+        // SAFETY: hex encoding always produces valid UTF-8
+        let hex_str = unsafe { str::from_utf8_unchecked(&buf) };
+        f.write_str(hex_str)
     }
 }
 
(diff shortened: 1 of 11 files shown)


barakshani:
STR-1478 --> STR-1243

crates/primitives/src/l1/status.rs
```diff
@@ -29,3 +32,48 @@ pub struct L1Status {
     /// Number of published reveal transactions.
     pub published_reveal_txs_count: u64,
 }
+
+// Custom debug implementation to print the txid in little endian
```
delbonis, line 36: Can't we just replace `Buf32` with a `BitcoinTxid`?

crates/primitives/src/l1/block.rs
```diff
@@ -116,3 +133,23 @@ impl L1BlockCommitment {
...
+    }
+}
+
+// Custom display implementation to print the block hash in little endian
+impl fmt::Display for L1BlockId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut bytes = self.0 .0;
+        bytes.reverse();
+        let hex_str = hex::encode(bytes);
+        f.write_str(&hex_str)
```
delbonis: This doesn't *really* matter but can we do the no alloc encoding with shrex?

delbonis (changes requested):

crates/primitives/src/l1/status.rs
storopoli: No because of zkVM restrictions

crates/primitives/src/l1/block.rs
storopoli: Yes, but then you have some `unsafe` stuff. I did this in a separate commit in 62be4777b7e85760524b7c30bc5d8bff55e0c878. Take a look.

crates/primitives/src/l1/status.rs
prajwolrg: What specific zkVM restrictions? I don't think there should be any restrictions by zkVM to allow for this.

storopoli: Thanks @prajwolrg! This is out of the scope of this PR, but I'll create a new ticket and assign to a new cleanup/tech-debt sprint.

[STR-1759](https://alpenlabs.atlassian.net/browse/STR-1759).

delbonis:
Looks good so merging.

PR #1066, opened 2025-09-25
EE account runtime

delbonis:
## Description

This is the skeleton of the EE account runtime framework and associated supporting libraries.  This is based on [this design](https://www.notion.so/Exec-env-runtime-framework-25a901ba000f808890f3f26288d25b49) but with some improvements discovered in the process of implementation.  There is a divergence from the design doc to not have the separate commit-prove OL confirmation process for blocks and instead directly post the new exec blkid in the update extradata field.  I am also planning on having a more general DA system that could be both diff based and summary based.

This PR also contains a new strata-simple-ee crate, which is used for testing the EE framework to ensure that it does validation in the ways we expect.  But it's standalone, so aside from the fact there's no authorization, we could actually deploy it to a real system as a demo.  We can also build a bunch of tooling around it to make a standalone client for it, if we wanted to.  I would love to keep using it to build out that without needing to pull around all the Reth baggage. (@sapinb may be interested).

Claude was used for:

* the simple EE impl
* iterating on the segment/update builder types
* writing most of the tests (with proper inspection to ensure they make sense)
* automatically fixing lints, moving modules around to hande the cleanup

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

      in the body of this PR.


crates/ee-acct-runtime/src/exec_processing.rs
```diff
@@ -0,0 +1,106 @@
...
+
+/// Verifies a commit against the account's current state using a coinput which
+/// reveals all the EE blocks and notpackage data, accumulating effects in the
+/// verification state.
+pub fn verify_commit<E: ExecutionEnvironment>(
+    vstate: &mut UpdateVerificationState,
+    astate: &EeAccountState,
+    commit: &CommitMsgData,
+    coinp: &CommitCoinput,
+    ee: &E,
```
storopoli: Shouldn't we be using something similar to the `strata-predicate` here?
I'm fine if this implies as leaving this as a TODO.

crates/ee-acct-runtime/src/exec_processing.rs
```diff
@@ -0,0 +1,106 @@
...
+    if ps_sr != prev_header_sr {
+        return Err(EnvError::MismatchedCoinput);
+    }
+
+    // 4. Go through each block and process them.
+    for block_data in coinp.blocks() {
+        process_block(vstate, &mut partial_state, block_data, ee)?;
+    }
+
+    // We don't update the
```
storopoli: Ooooo cliffhanger! Update what?!

crates/ee-acct-runtime/src/process_update.rs
```diff
@@ -0,0 +1,186 @@
+//! High-level procedures for processing updates.
+//!
+//! There are two main functions here.  The first is
+//! [`verify_and_apply_update_operation`], which is what we use in the proof to
+//! carefully check and update the account state.
+//!
+//! The second is [`apply_update_operation_unconditionall`], which is used
+//! outside the proof, after verifying the proof, to update our view of the
+//! state, presumably with information extracted from DA.
```
storopoli: Perfect! Thanks for these docs :)

crates/ee-acct-runtime/src/update_processing.rs
```diff
@@ -0,0 +1,186 @@
...
+            // Other type or invalid message, skip.
+            continue;
+        };
+
+        // Verify the message.  This relies on the private coinput.
+        verify_message(&mut vstate, state, &eem, &meta, coinp, operation, ee)?;
+
+        // Then apply the message.  This doesn't rely on the private coinput.
+        apply_message(state, &eem, &meta, operation)?;
+    }
```
storopoli, line 138: This is not an atomic operation.
What happens if we process multiple `(inp, coinp)`s and one or more of these are invalid. Then the `EeAccountState` would be changed in the loop even if the `verify_accumulated_state` below fails.

Is this expected? If not, we should accumulate all the changes in the `EeAccountState`, and then apply them after the `verify_accumulated_state` below to make this `verify_and_apply_update_operation` atomic and without side effects if it fails.

storopoli (changes requested):

crates/ee-acct-runtime/src/exec_processing.rs
delbonis: Uh, where?

crates/ee-acct-runtime/src/update_processing.rs
delbonis: This state is small and in memory, so we're not modifying an "authoritative" instance.  We'd discard the modified state if the function returns an err.  From the context of the proof, the err paths are never taken.

Maybe `verify_message` is a misleading name.  It's not verifying the *message* per se, it's challenging the prover to potentially "do something with" the message (via the coinput).  If they can't then they can't construct the proof.

crates/ee-acct-runtime/src/exec_processing.rs
storopoli: In the bunch of checks. Check 1, Check 2, Check 3 etc in the body of the function.

crates/ee-acct-runtime/src/update_processing.rs
storopoli: Solving, thanks for the explanation

crates/ee-acct-runtime/src/exec_processing.rs
delbonis: fixed this oops

crates/ee-acct-runtime/src/exec_processing.rs
delbonis: oh I've refactored this anyways

crates/ee-acct-runtime/src/update_processing.rs
```diff
@@ -0,0 +1,303 @@
...
+            .map_err(make_inp_err_indexer(i))?;
+
+        // Then apply the message.  This doesn't rely on the private coinput.
+        apply_message(astate, &msg, &extra).map_err(make_inp_err_indexer(i))?;
+    }
+
+    // Make sure there are no more leftover coinputs we haven't recognized.
+    if coinp_iter.next().is_some() || cnt != operation.processed_messages().len() {
+        return Err(EnvError::MismatchedCoinputCnt);
+    }
```
storopoli, line 143: `cnt` is set to the loop index `i`, so after processing `n` messages it still equals `n-1`. Any non-empty operation therefore triggers `MismatchedCoinputCnt`, meaning no update can be accepted.

crates/ee-acct-runtime/src/verification_state.rs
```diff
@@ -0,0 +1,174 @@
...
+        let Some(exp_next) = self.expected_next() else {
+            return Err(EnvError::MalformedCoinput);
+        };
+
+        if input != exp_next {
+            return Err(EnvError::MalformedCoinput);
+        }
+
+        Ok(())
+    }
```
storopoli, line 184: The tracker never increments `self.consumed`. After the first match, the next input is compared against the same expected entry (and `consumed()` stays zero), so multi-input segments fail. Even if the entries matched, the consumed count fed into later consistency checks stays at `0`, causing `ConflictingPublicState`.

delbonis: oh good catch, making this generic so will fix

delbonis:
This PR is essentially ready now.  There is one missing issue, that some types should be converted over to use SSZ so that we can use the `hash_tree_root` function to compute their commitments, in particular for `EeAccountState`.  This is kinda essential for the system to be secure, since right now we're actually not doing the final account state check necessary to verify an account update, although it's not really a blocker for ongoing development.

The other issue is this stupid thing I don't know what to do about, which is breaking the doc and lint CI jobs, it's some lockfile issue:

```
error[E0277]: `DefaultHashBuilder` doesn't implement `Debug`
  --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/revm-inspectors-0.27.2/src/edge_cov.rs:24:5
   |
20 | #[derive(Clone, Debug)]
   |                 ----- in this derive macro expansion
...
24 |     hash_builder: DefaultHashBuilder,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Debug` is not implemented for `DefaultHashBuilder`
```

PR #1054, opened 2025-09-24
STR 1598 Move checkpoint types

krsnapaudel:
## Description

Move checkpoint types to `checkpoint-types` crate from `primitives`.

This PR was created with help from Cursor AI.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Not sure whether `EpochSummary` should be in `ol-chain-types`.

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [x] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [x] New and existing tests pass with my changes.
- [x] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/.github/CONTRIBUTING.md#ai-assistance-notice)
      in the body of this PR.

## Related Issues

[STR-1598](https://alpenlabs.atlassian.net/browse/STR-1598)

[STR-1598]: https://alpenlabs.atlassian.net/browse/STR-1598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1067, opened 2025-09-25
New state accessor skeleton

delbonis:
## Description

For additional details, see the lib.rs in the new crate.

There's various different contexts that we'll need to run the OL STF logic, which have different demands on how state is manipulated:

* classically, to update state
* classically, to generate DA
* within a snark proof
* maybe some tracing/hooking reasons

Instead of having one concrete type hierarchy that attempts to handle all of these use cases, we can instead flip it around and have the STF functions be generic over the object hierarchy *in its entirety* instead of simply doing it for the base-level interface to the spec.  This is a design that I borrow from something I saw in a different context within various projects, but in particular *I think* in Reth, it just took me a while to realize that it also applies here.  This does not supersede the `LedgerInterface` defined in the snark acct spec, it only supersedes the existing `StateAccessor` trait in the chaintsn crate.

### Type of Change

- [ ] Documentation update
- [x] Refactor **(maybe?)**
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

      in the body of this PR.

### DIFF
--- a/crates/ledger-types/src/account.rs
+++ b/crates/ledger-types/src/account.rs
@@ -1,0 +1,65 @@
+use strata_acct_types::{
+    AccountSerial, AccountTypeId, AcctResult, BitcoinAmount, RawAccountTypeId,
+};
+use strata_snark_acct_types::MessageEntry;
+
+/// Abstract account state.
+pub trait IAccountState: Sized {
+    /// Type representing snark account state.
+    type SnarkAccountState: ISnarkAccountState;
+
+    /// Gets the account serial.
+    fn serial(&self) -> AccountSerial;
+
+    /// Gets the account's balance.
+    fn balance(&self) -> BitcoinAmount;
+
+    /// Sets the account's balance.
+    fn set_balance(&self, amt: BitcoinAmount);
+
+    /// Gets the account raw type ID.
+    fn raw_ty(&self) -> AcctResult<RawAccountTypeId>;
+
+    /// Gets the parsed account type ID, if valid.
+    fn ty(&self) -> AcctResult<AccountTypeId>;
+
+    /// Gets the account type state, if valid.
+    fn get_type_state(&self) -> AcctResult<AccountTypeState<Self>>;
+
+    /// Sets the account type state.
+    fn set_type_state(&self, state: AccountTypeState<Self>) -> AcctResult<()>;
+}
+
+/// Account type state enum.
+pub enum AccountTypeState<T: IAccountState> {
+    /// Empty accounts with no state.
+    Empty,
+
+    /// Snark account with snark account state.
+    Snark(T::SnarkAccountState),
+}
+
+/// Abstract snark account state.
+pub trait ISnarkAccountState: Sized {
+    // TODO accumulator accessors
+
+    /// Gets the next inbox msg index.
+    fn get_next_inbox_msg_idx(&self) -> u64;
+
+    /// Gets the update seqno.
+    // TODO convert to Seqno
+    fn seqno(&self) -> u64;
+
+    /// Gets the inner state root hash.
+    fn inner_state_root(&self) -> [u8; 32];
+
+    /// Inserts a message into the inbox.  Performs no other operations.
+    fn insert_inbox_message(&mut self, entry: MessageEntry) -> AcctResult<()>;
+
+    /// Increments the sequence number by some amount.
+    // TODO convert to Seqno
+    fn increment_seqno(&mut self, amt: u64) -> AcctResult<u64>;
+
+    /// Sets the inner state root.
+    fn set_inner_state_root(&mut self, state: [u8; 32]);
+}
--- a/crates/ledger-types/src/lib.rs
+++ b/crates/ledger-types/src/lib.rs
@@ -1,0 +1,24 @@
+//! Ledger data types.
+//!
+//! This crate is NOT about the basic data structures themselves.  This crate
+//! focuses on how we access the ledger data structures in different contexts.
+//!
+//! We present a trait that represents the various types of structures we
+//! interact with in the ledger's state, and expose accessor functions on it.
+//! The different impls of these traits are tailored for different contexts.  In
+//! some contexts we care about tracing DA generation, in others we may be doing
+//! blocking fetches from disk we want to trace for later proof generation.
+//!
+//! We use the `I` prefix convention which is normally uncommon in Rust to refer
+//! to these abstract data structures.  This is because the "ordinary" struct
+//! versions of these data structure we use on the wire are the "real" versions
+//! we want to think of them as being, but these traits are standins for those.
+//! Making up new names for these items would crate too much confusion.
+
+mod account;
+mod state_accessor;
+mod toplevel;
+
+pub use account::{AccountTypeState, IAccountState, ISnarkAccountState};
+pub use state_accessor::StateAccessor;
+pub use toplevel::IToplevelState;
--- a/crates/ledger-types/src/state_accessor.rs
+++ b/crates/ledger-types/src/state_accessor.rs
@@ -1,0 +1,38 @@
+use strata_acct_types::{AccountId, AccountSerial, AcctResult};
+
+use crate::{
+    account::{AccountTypeState, IAccountState},
+    toplevel::IToplevelState,
+};
+
+/// Opaque interface for manipulating the ledger state.
+///
+/// This exists because we want to make this generic across the various
+/// different contexts we'll be manipulating state.
+pub trait StateAccessor {
+    /// Type representing the toplevel chainstate.
+    type ToplevelState: IToplevelState;
+
+    /// Type representing ledger account state.
+    type AccountState: IAccountState;
+
+    /// Gets the toplevel state.
+    fn toplevel(&self) -> &Self::ToplevelState;
+
+    /// Checks if an account exists.
+    fn check_account_exists(&self, id: AccountId) -> AcctResult<bool>;
+
+    /// Fetches an account by ID, if it exists.
+    fn fetch_account_state(&self, id: AccountId) -> AcctResult<Option<Self::AccountState>>;
+
+    /// Stores an existing account entry's state, if it exists.
+    fn store_account_state(&mut self, id: AccountId, state: Self::AccountState) -> AcctResult<()>;
+
+    /// Creates a new account as some ID with some type state, if that ID
+    /// doesn't exist, returning the serial.
+    fn create_new_account(
+        &mut self,
+        id: AccountId,
+        state: AccountTypeState<Self::AccountState>,
+    ) -> AcctResult<AccountSerial>;
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14789,6 +14789,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "strata-ledger-types"
+version = "0.1.0"
+dependencies = [
+ "strata-acct-types",
+ "strata-snark-acct-types",
+]
+
+[[package]]
 name = "strata-mmr"
 version = "0.3.0-alpha.1"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ members = [
   "crates/evmexec",
   "crates/key-derivation",
   "crates/l1tx",
+  "crates/ledger-types",
   "crates/mpt",
   "crates/primitives",
   "crates/proof-impl/checkpoint",
@@ -144,6 +145,7 @@ strata-eectl = { path = "crates/eectl" }
 strata-evmexec = { path = "crates/evmexec" }
 strata-key-derivation = { path = "crates/key-derivation" }
 strata-l1tx = { path = "crates/l1tx" }
+strata-ledger-types = { path = "crates/ledger-types" }
 strata-mmr = { path = "crates/util/mmr" }
 strata-mpt = { path = "crates/mpt" }
 strata-ol-chain-types = { path = "crates/ol-chain-types" }
--- a/crates/ledger-types/Cargo.toml
+++ b/crates/ledger-types/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "strata-ledger-types"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+strata-acct-types.workspace = true
+strata-snark-acct-types.workspace = true
+
+[lints]
+workspace = true
--- a/crates/ledger-types/src/toplevel.rs
+++ b/crates/ledger-types/src/toplevel.rs
@@ -1,0 +1,4 @@
+/// Abstract toplevel chainstate.
+pub trait IToplevelState {
+    // TODO accessors
+}


crates/ledger-types/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+//! to these abstract data structures.  This is because the "ordinary" struct
+//! versions of these data structure we use on the wire are the "real" versions
+//! we want to think of them as being, but these traits are standins for those.
+//! Making up new names for these items would crate too much confusion.
+//!
+//! As for structure, this design is based around a "toplevel" state that is not
+//! ever actually directly accessed.  Below it, there are two parts:
+//!
+//! * A "global" state that is treated using the DA framework directly.
+//! * An "accounts" table, which are selectively loaded.
```
bewakes, line 24: My understanding of "global" state was that it corresponds to the toplevel state, and "account" state corresponds to the more detailed state of an account. Is global state only for DA?

crates/da-framework/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "strata-da-framework"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+strata-codec = { git = "https://github.com/alpenlabs/strata-common" }
```
storopoli: Shouldn't we add this to the Workspace `Cargo.toml`?

crates/ledger-types/src/coin.rs
```diff
@@ -0,0 +1,63 @@
...
+
+/// A linear coin that must be explicitly created and destroyed.  This allows us
+/// to more safely reason about flow of funds between components and reduce the
+/// complexity of bookkeeping by making it more likely that accounting bugs are
+/// turned into panics.
+///
+/// Triggers a panic when dropped without [`safely_consume_unchecked`] being
+/// called.
+#[derive(Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub struct Coin(BitcoinAmount);
```
storopoli, line 15: Why not also do a `From<BitcoinAmount>`? Do you want all the new `Coin`s to be constructed with the `new_unchecked` method?

crates/ledger-types/src/coin.rs
```diff
@@ -0,0 +1,63 @@
...
+        // Since this is just flat bytes, we can do this to safely destroy
+        // ourselves without calling `Drop::drop`.
+        mem::forget(self);
+    }
+}
+
+impl Drop for Coin {
+    fn drop(&mut self) {
+        let amt: u64 = self.amt().into();
+        panic!("coin: accidentally destroyed value ({amt} sats)");
```
storopoli, line 43: This is really clever! We cannot never destroy `Coin`s.
What happens in a withdrawal? Are they sent to the "blackhole" bridge pre-compile withdrawal address?

crates/ledger-types/src/state_accessor.rs
```diff
@@ -0,0 +1,50 @@
...
+    type GlobalState: IGlobalState;
+
+    /// Type representing a ledger account's state.
+    type AccountState: IAccountState;
+
+    /// Gets a ref to the global state.
+    fn global(&self) -> &Self::GlobalState;
+
+    /// Gets a mut ref to the global state.
+    fn globlal_mut(&mut self) -> &mut Self::GlobalState;
```
storopoli: typo

```suggestion
    fn global_mut(&mut self) -> &mut Self::GlobalState;
```

crates/da-framework/src/codec.rs
```diff
@@ -0,0 +1,7 @@
+//! Codec re-exports from strata-codec.
+
+// Re-export everything from strata-codec
+pub use strata_codec::{Codec, CodecError, Decoder, Encoder, decode_buf_exact, encode_to_vec};
```
storopoli, line 4: Will this re-exporting bite us in the future?

crates/da-framework/src/errors.rs
```diff
@@ -0,0 +1,20 @@
...
+pub enum DaError {
+    #[error("context missing required data")]
+    InsufficientContext,
+}
+
+/// Errors generated in building a diff.
+#[derive(Debug, Error)]
+pub enum BuilderError {
+    #[error("tried to set a value out of allowed bounds")]
+    OutOfBoundsValue,
```
storopoli, line 16: Can we add more context?

```suggestion
    OutOfBoundsValue(usize, usize),
```

crates/da-framework/src/register.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// This only really makes sense for registers since they're the only type
+    /// we can consistently do this with.
+    pub fn compare(orig: &T, new: &T) -> Self {
+        if new == orig {
+            Self::new_unset()
+        } else {
+            Self::new_set(new.clone())
+        }
+    }
+}
```
storopoli, line 56: Shouldn't this be a manual `Eq` for `DaRegister`?

crates/da-framework/src/register.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+        Ok(())
+    }
+}
+
+impl<T> Default for DaRegister<T> {
+    fn default() -> Self {
+        Self { new_value: None }
+    }
+}
```
storopoli, line 75: I think we can do this with a `#[derive(Default)]` in `DaRegister` no?

crates/da-framework/src/counter.rs
```diff
@@ -0,0 +1,289 @@
...
+        })
+    }
+}
+
+/// Counter schemes.
+pub mod counter_schemes {
+    inst_direct_ctr_schemes! {
+        CtrU64ByU8(u64, u8);
+        CtrU64ByU16(u64, u16);
+        CtrU32ByU8(u64, u8);
```
storopoli:
```suggestion
        CtrU32ByU8(u32, u8);
```

crates/ledger-types/src/state_accessor.rs
```diff
@@ -0,0 +1,50 @@
...
+    account::{AccountTypeState, IAccountState},
+    global_state::IGlobalState,
+};
+
+/// Opaque interface for manipulating the chainstate, for all of the parts
+/// directly under the toplevel state.
+///
+/// This exists because we want to make this generic across the various
+/// different contexts we'll be manipulating state.
+pub trait StateAccessor {
```
storopoli, line 14: https://github.com/alpenlabs/alpen/blob/9ba835b12cd7f4e48746da429c89d8645fa588cc/crates/chaintsn/src/context.rs#L78-L81

storopoli (changes requested):

crates/da-framework/Cargo.toml
delbonis: yeah I think this was something Claude messed up, I've already fixed it

crates/ledger-types/src/coin.rs
delbonis: We'd call `safely_consume_unchecked` and then it's just a matter of ensuring that we call that in exactly the right place.

crates/da-framework/src/codec.rs
delbonis: I hope not.  I wasn't sure how far to go with stripping this out, but I can probably go all the way with it.

crates/da-framework/src/errors.rs
delbonis: I can add more variants to cover this.  I didn't want to make it overly specific because, for example with a counter, we could have a range that's like [a, a+255) and so trying to set it to a-1 is actually out of bounds but it isn't like a "size limit" like that variant would allow us to express.

crates/da-framework/src/register.rs
delbonis: Don't think so?  It's returning a `DaRegister<T>` not a `bool`.

crates/ledger-types/src/state_accessor.rs
delbonis: Yeah the new one will superscede the old one.  I love telescoping supersession :upside_down_face: .

crates/da-framework/src/register.rs
delbonis: Actually no, using the derive impl breaks since it relies on `T: Default`, but we don't actually want to make that constraint.  We *could*, but there's no reason to.

crates/ledger-types/src/coin.rs
delbonis: Yeah I want us to be very explicit about how we're creating/destroying these.  With `From` that allows for some more-implicit conversions which feels like it'd encourage mistakes.

crates/ledger-types/src/coin.rs
```diff
@@ -0,0 +1,63 @@
+//! Coin abstraction to ensure we don't accidentally create or destroy funds.
+
+use std::{mem, ops::Drop};
+
+use strata_acct_types::BitcoinAmount;
+
+/// A linear coin that must be explicitly created and destroyed.  This allows us
+/// to more safely reason about flow of funds between components and reduce the
+/// complexity of bookkeeping by making it more likely that accounting bugs are
+/// turned into panics.
```
bewakes, line 10: This is interesting. But looks risky too because of the usage of mem::forget, which may leak memory(though possibly not in this particular case).
Is this the only way we can achieve accounting safety?

bewakes: Also, it would be great to know with examples how exactly it allows us to safely reason about flow of funds.

crates/da-framework/src/compound.rs
```diff
@@ -0,0 +1,336 @@
...
+        }
+    };
+}
+
+impl_shift_bitmap!(u8);
+impl_shift_bitmap!(u16);
+impl_shift_bitmap!(u32);
+
+/// Safer interface around a [`ShiftBitmap`] that ensures we don't overflow.
+pub struct BitReader<T: ShiftBitmap> {
```
bewakes: It might be a good idea to prepend/append the name with something like `Seq`, since this reader reads in sequence. Same for the writer below. The way it is indicates that random access is possible.

crates/da-framework/src/counter.rs
```diff
@@ -0,0 +1,289 @@
+//! Simple counter type.
```
bewakes, line 1: This is a neat abstraction via `CounterScheme`. It definitely allows for better extensibility. However, given this is just a counter and most likely will be used for numbers only in practice, I feel like it doesn't need to be this complicated in terms of boilerplate and macros.

What do you think of [the one that I implemented](https://github.com/alpenlabs/alpen/blob/fe296556a5cfc5f1cefabd1c6608b01824249a16/crates/da-lib/src/num.rs) a while ago, particular for type like counter? It's called `NumDiff` but serves the same purpose and is relatively simpler with comparable wire size. Also, this has more robust compile time guarantees. It doesn't have the builder and codec considerations but those can be added.

crates/ledger-types/src/account.rs
```diff
@@ -0,0 +1,96 @@
...
+use strata_snark_acct_types::MessageEntry;
+
+use crate::coin::Coin;
+
+type Seqno = u64;
+
+/// Abstract account state.
+pub trait IAccountState: Sized {
+    /// Type representing snark account state.
+    type SnarkAccountState: ISnarkAccountState;
```
bewakes, line 13: I believe there are other account types that might be added in future. Wouldn't having this associated `SnarkAccountState` restrict that? Maybe this can be renamed to something generic like `InnerAccountState`.

crates/ledger-types/src/account.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// Gets the account type state, if valid.
+    fn get_type_state(&self) -> AcctResult<AccountTypeState<Self>>;
+
+    /// Sets the account type state.
+    fn set_type_state(&self, state: AccountTypeState<Self>) -> AcctResult<()>;
+}
+
+/// Account type state enum.
+#[derive(Clone, Debug)]
+pub enum AccountTypeState<T: IAccountState> {
```
bewakes, line 42: What do you think of `AccountInnerState`? The current name implies the state of an "account type", which is a bit confusing.

crates/ledger-types/src/l1vs.rs
```diff
@@ -0,0 +1,65 @@
...
+#[derive(Clone, Debug)]
+pub struct AsmManifest {
+    // TODO
+}
+
+/// Transitional type until we update crate hierarchy.
+#[derive(Clone, Debug)]
+pub struct EpochCommitment {
+    // TODO
+}
```
storopoli, line 24: Will these be fixed during this PR?

crates/da-framework/src/errors.rs
storopoli: I see, that's fine as it is...

crates/ledger-types/src/state_accessor.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// Type representing the global chainstate.
+    type GlobalState: IGlobalState;
```
bewakes, line 16: This is the toplevel state right?

Also, Is it really necessary to have trait bounds for `GlobalState` and `AccountState`?Because neither's internals are being used as argument or return types in the trait methods.

Basically, I am not sure if trait bounds for Global State and Account State are needed because it seems to me that accessing/modifying state is being done via this accessor itself, it should be sufficient to define those getters/setters/modifiers in the `StateAccessor` trait. I would even say that with mutable getters for both states, we might even not need to have fine grained getters/setters. I might be wrong or missing something though.

crates/ledger-types/src/state_accessor.rs
storopoli: Ok, will you tackle the superseding in this PR? If not, can you open a ticket 
for that?

crates/da-framework/src/queue.rs
```diff
@@ -0,0 +1,231 @@
...
+        // Remove things from this that are redundant.
+        self.new_entries.drain(..self.consumed_new_entries());
+
+        let tail = self.new_entries;
+        let incr_front = self.new_front_pos - self.original_front_pos;
+        Ok(DaQueue { tail, incr_front })
+    }
+}
+
+// TODO tests for the wacky logic up ^there
```
bewakes: Could you add these? If it is not feasible right now, please create ticket(s). Ideally we want every module to have tests which would also greatly help on reviewing and understanding the code.

bewakes (changes requested):
Great work overall!

Requesting change for missing tests and for some clarifying questions. There are some suggestions on the names which are not critical blockers but would make things much cleaner.

crates/ledger-types/src/account.rs
delbonis: We would add new associated types for each account type.

crates/ledger-types/src/lib.rs
delbonis: I'm trying to be cleaner with how the types are composed.  I'm using the term "global" to refer to the concrete state that's always available and then "toplevel" to refer to the container for that and the other types of state.

Previously we had to specify "these fields are managed by diffs directly and these fields are updated indirectly" about the toplevel state, but that is cumbersome.  So structuring it this way lets us say that the toplevel state is where we split out the different handling pathways and consolidate them together, then each component just has to worry about a single model of how diffs are handled.

crates/ledger-types/src/coin.rs
delbonis: > which may leak memory(though possibly not in this particular case).

Calling `mem::forget` here is to discard the value without invoking `Drop`.  There's no allocs so there's no memory leaks possible.

>  Is this the only way we can achieve accounting safety?

It's leaning into Rust's type system and [faking true linear types](https://en.wikipedia.org/wiki/Substructural_type_system) by panicking if we don't consume the type correctly.

The goal is to make it easier to compose across different components without having to mentally do global reasoning to think about the correctness of the system.

> Also, it would be great to know with examples how exactly it allows us to safely reason about flow of funds.

```rs
let coin = thing.get_coin();
spend_coin(coin); // consumes
spend_coin(coin); // can't call again, value already consumed
```

If `coin` was just `BitcoinAmount` instead then we'd have just duplicated funds.

crates/ledger-types/src/account.rs
delbonis: Using "inner" might be more strictly correct, but when I started writing it out in the public interface like this it started to feel like it was exposing internals, which is not the intention I was going for.

crates/ledger-types/src/state_accessor.rs
delbonis: > This is the toplevel state right?

Yes.

> Also, Is it really necessary to have trait bounds for GlobalState and AccountState?Because neither's internals are being used as argument or return types in the trait methods.

Yes they are?  This depends on the internals, by way of `AccountTypeState`:

```rs
    fn create_new_account(
        &mut self,
        id: AccountId,
        state: AccountTypeState<Self::AccountState>,
    ) -> AcctResult<AccountSerial>;
```

But really, we can't just do

```rs
fn global(&self) -> &impl IGlobalState;
fn global_mut(&mut self) -> &mut impl IGlobalState;
```

because we need to enforce that it's the same type.  Further, we can't do

```rs
fn update_account_state(&mut self, id: AccountId, state: impl IAccountState) -> AcctResult<()>;
```

because then the `StateAccessor`'s impl of `update_account_state` would need to be generic over *that* impl, and that creates a whole can of worms.

In a language like Java you'd be forced to use the interface type there, but then normally you'd declare that you can't pass an instance to it that you didn't obtain from it so that it's safe to just make a cast to the concrete class and throw an exception otherwise.  But we're in Rust so we can just propagate the types we want throughout the whole consumer.

```java
public void updateAccountState(AccountId id, IAccountState state) {
    if (!(state instanceof MyAccountState)) {
        throw IllegalArgumentException("used foreign object type");
    }
    MyAccountState realState = (MyAccountState) state;
    // ...
}
```

crates/da-framework/src/counter.rs
delbonis: I had gone with a similar approach initially but if I remember right there was some issues I ran into when it came to making the codec/builder work cleanly, it ended up being a mess of type constraints that felt unergonomic.

These DA components are from a while ago that I never actually created a PR for.

crates/da-framework/src/queue.rs
delbonis: Added tests.

crates/ledger-types/src/l1vs.rs
delbonis: Waiting on #1054 for this.

crates/ledger-types/src/coin.rs
bewakes: I see, this sounds cool. I would approach it differently though, to try to achieve compile time safety instead of runtime panicing. What do you think of this?

We annotate each coin with a phantom marker(Spent or Unspent), make it non-Copy and non-Clone. There would be a single function that would create coins, possibly some deposit related function. Now, we can make the OL stf or some sub routine that the stf calls accept Unspent coins and return Spent coins. The coin spending would be done by a single consume function, so that needs to  be called ensuring consumption. High level signatures:

```rust
pub struct Spent;
pub struct Unspent;

// no copy, no clone
struct Coin<SpentOrUnspent>  {
     value: u64,
     state: PhantomData<SpentOrUnspent>
}

// Called by deposit function
fn mint_coin(acct: AccountId, value: u64) -> Coin<Unspent>;

// Called within stf
fn consume(coin: Coin<Unspent>) -> Coin<Spent>;  // Coin<Spent> would be a part of STF somehow, so that this needs to be called.
```

crates/ledger-types/src/account.rs
bewakes: That doesn't feel right because we would have to change the trait impls everytime we have a new type which is not scalable. Basically this is like trying to enforce enum variants on the type level. An account state has a single internal type state. So I believe there should be a single associated type.

On a side note, I honestly feel like traits are somewhat being overused. From what I understand about the system, we will have single instances for toplevel account and internal account type variants. The benefits of "tratifying" everything is not very clear at this point. Sure, it makes things more generic — but do we really need that?

crates/da-framework/src/counter.rs
bewakes: I haven't tried to incorporate codec and builder in my implementation so I can't really say how messy it gets. But if you have already tried it and it did not work out then that's probably fine - though I still incline towards that simpler approach.

crates/ledger-types/src/coin.rs
delbonis: There's not really a purpose in keeping a Coin around after it's been spent, and allowing a "spent" kind of coin would encourage misuse by reintroducing the possibility for semantic confusion.

crates/ledger-types/src/state_accessor.rs
delbonis: I expect that that whole crate will be massively changed when we implement the OL STF for real, so I don't see a reason to explicitly say to do it.

crates/ledger-types/src/account.rs
delbonis: > we would have to change the trait impls everytime we have a new type which is not scalable.

Yes, this is expected, when we introduce a new concept we sometimes have to extend the components that interface with it.  We'd already have to go update every (non-wildcarded) `match` on account when we add a new account type.  I don't agree with the argument that this design is not scalable but the status quo is.

> An account state has a single internal type state. So I believe there should be a single associated type.

There is one field, but its interpretation varies depending on the account type.  There will be multiple account types.  How do we express the different account types as a single Rust type?

Yes, we can "just use an enum", but how do we extend the abstraction through that to each account type('s decoded state)?  We can't express all possible account state operations without disambiguating somewhere.  We'd also want to avoid having to repeatedly decode/encode/redecode account state as we manipulate it.

> The benefits of "tratifying" everything is not very clear at this point. [...] but do we really need that?

Yes we do, I discussed it in the PR description.  For one, this is the simplest way to to support OL DA without baking that bookkeeping directly into every data structure.  The functions on these traits correspond to the kinds of operations that we express in DA.  It also makes it easy to keep a clean separation between how state is expressed when it's loaded from the database and how it's expressed when we provide it as a private input to the proof.

One of the examples from Reth that I was referring to: https://reth.rs/docs/reth_revm/trait.ExecuteEvm.html

crates/ledger-types/src/l1vs.rs
storopoli: Merged. Please rebase.

crates/ledger-types/src/state_accessor.rs
storopoli: Resolving this and Cc'ing @bewakes

crates/ledger-types/src/coin.rs
bewakes: I agree there's not much benefit in keeping it. But this gives us compile time safety which is a much bigger payoff. Also we could just drop them afterwards.

PR #1072, opened 2025-09-29
feat(asm): checkpoint v0 subprotocol implementation

irnb:
## Checkpointing v0 (ASM only)

**Crates:**

* `crates/asm/subprotocols/checkpoint-v0`
* `crates/asm/txs/checkpoint`

### Overview

This PR introduces **checkpointing v0**, which includes integration with both the bridge and administration subprotocols.

### Key Responsibility

* Extraction of withdrawal intents from checkpoint
* Message routing to the bridge
* Processing of update messages received from the administration subprotocol
* Simple proof verification aligned with the TN1 checkpoint proof verification flow

⚠️ Note: Some soundness checks are missing in checkpointing v0 proof verification. We’re aware of these gaps and plan to address them in the next iteration.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

      in the body of this PR.

## Related Issues

STR-1712
STR-1710


crates/asm/subprotocols/checkpoint-v0/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[lints]
+workspace = true
+
+[dependencies]
+# ASM dependencies
+borsh.workspace = true
+strata-asm-common.workspace = true
+strata-asm-logs.workspace = true
+strata-asm-proto-bridge-v1.workspace = true
```
prajwolrg, line 14: This implies that the `strata-asm-proto-bridge-v1` cannot have `strata-asm-proto-checkpointing-v0` as a dependency. Although this is fine for now, we must design the subprotocols such that each subprotocol can send the messages to any other subprotocol. 

One way can be to create a separate crate for the msgs for each subprotocol. So for each subprotocol we might need to have three crates:
1. Subprotocol Implementation
2. Transaction parsing
3. Interprotocol Msg and Logs

It might also be acceptable for now that we use the `strata-asm-logs` for all the InterprotoMsg. 

WDYT? CC: @delbonis

irnb: Oh, good catch 
thanks for mentioning it. I’ll create a ticket so we don’t forget.

Cargo.toml
delbonis, line 1: Don't need to have the `ing` on these names, `checkpoint` is clear enough and keeps the crate names slightly shorter.

crates/asm/subprotocols/admin/src/handler.rs
```diff
@@ -140,6 +172,20 @@ pub(crate) fn handle_action(
...
+    relayer.relay_msg(&msg);
+}
+
+fn relay_checkpointing_rollup_vk(relayer: &mut impl MsgRelayer, rollup_vk: RollupVerifyingKey) {
+    let msg = CheckpointingIncomingMsg::UpdateRollupVerifyingKey(rollup_vk);
+    relayer.relay_msg(&msg);
+}
+
+fn deserialize_rollup_vk(vk: &VerifyingKey) -> Result<RollupVerifyingKey, serde_json::Error> {
+    serde_json::from_slice(vk.as_bytes())
```
delbonis, line 186: I'm not sure if this is how it was already written previously but does this have to be JSON here for some reason?

crates/asm/subprotocols/checkpoint-v0/Cargo.toml
delbonis: This is sorta an example of a "separate interface and implementation" situation.  When in doubt, it would make sense to separate the message types into separate crates so that other subprotos can depend on them without having to depend on the impl.  Especially when they send messages to each other (as that would cause cyclical dependencies).

crates/asm/subprotocols/checkpointing-v0/src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+    #[error("Invalid epoch: expected {expected}, got {actual}")]
+    InvalidEpoch { expected: u64, actual: u64 },
+
+    /// Serialization error
+    #[error("Serialization error")]
+    SerializationError,
+
+    /// Invalid transaction type
+    #[error("Unsupported transaction type: {0}")]
+    UnsupportedTxType(String),
```
delbonis: Can just make this be a u8.

crates/asm/subprotocols/checkpoint-v0/src/error.rs
```diff
@@ -0,0 +1,52 @@
+//! Error types for checkpointing v0 subprotocol
+
+use strata_asm_proto_checkpointing_txs::CheckpointTxError;
+use thiserror::Error;
+
+/// Errors that can occur during checkpoint verification and processing
+#[derive(Debug, Error)]
+pub enum CheckpointV0Error {
```
delbonis, line 8: Having `Error` in these variant names is redundant.

crates/asm/subprotocols/checkpointing-v0/src/verification.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    Ok(())
+}
+
+fn ensure_batch_epochs_consistent(checkpoint: &Checkpoint) -> Result<(), CheckpointV0Error> {
+    let info_epoch = checkpoint.batch_info().epoch();
+    let transition_epoch = checkpoint.batch_transition().epoch;
+    if info_epoch != transition_epoch {
+        return Err(CheckpointV0Error::StateTransitionError(
+            "batch info and transition epochs differ".to_string(),
```
delbonis: Make these into different error variants.

crates/asm/txs/checkpointing/src/parser.rs
```diff
@@ -0,0 +1,67 @@
...
+    })?;
+
+    Ok(chain_state.pending_withdraws().entries().to_vec())
+}
+
+fn parse_envelope_payload(script: &ScriptBuf) -> CheckpointTxResult<Vec<u8>> {
+    let mut instructions = script.instructions();
+
+    enter_envelope(&mut instructions)
+        .map_err(|e| CheckpointTxError::EnvelopeParse(e.to_string()))?;
```
delbonis: This could do a bunch of copies, we should be able to get a `impl Iterator<Item = &[u8]>` out of this and then convert that into a `Read` so we can directly parse the structure.

crates/asm/txs/checkpointing/src/errors.rs
```diff
@@ -0,0 +1,28 @@
...
+    #[error("checkpoint transaction missing taproot leaf script in first input witness")]
+    MissingLeafScript,
+
+    /// Failed to parse the envelope script structure.
+    #[error("failed to parse checkpoint envelope script: {0}")]
+    EnvelopeParse(String),
+
+    /// Failed to deserialize data embedded inside the envelope.
+    #[error("failed to deserialize checkpoint payload: {0}")]
+    Deserialization(String),
```
delbonis: It's fine to expose the envelope error instead of forcing conversion to string.

crates/asm/subprotocols/checkpointing-v0/src/subprotocol.rs
```diff
@@ -0,0 +1,268 @@
...
+        anchor_pre: &AnchorState,
+        _aux_input: &Self::AuxInput,
+        relayer: &mut impl MsgRelayer,
+        _params: &Self::Params,
+    ) {
+        // Get current L1 height from anchor state
+        let current_l1_height = anchor_pre.chain_view.pow_state.last_verified_block.height();
+
+        for tx in txs {
+            let result = process_checkpoint_transaction_v0(state, tx, current_l1_height, relayer);
```
delbonis: This function does only get called with txs that have already been checked to match the subproto ID, but maybe we should add a check to make sure it's the checkpoint tx type before we try to parse it as a checkpoint tx and `continue` early if it's not.  There's nothing stopping people from sending those txs, and we wouldn't want to waste time *trying* to parse them even if they were valid for some reason.  Not sure how much that helps us here.

crates/asm/subprotocols/checkpointing-v0/src/subprotocol.rs
```diff
@@ -0,0 +1,268 @@
...
+                        let txid = tx.tx().compute_txid();
+                        logging::info!("Successfully processed checkpoint transaction: {txid:?}");
+                    } else {
+                        let txid = tx.tx().compute_txid();
+                        logging::warn!("Rejected checkpoint transaction: {txid:?}");
+                    }
+                }
+                Err(e) => {
+                    let txid = tx.tx().compute_txid();
+                    logging::warn!("Error processing checkpoint transaction {txid:?}: {e:?}");
```
delbonis: If we log the txid in all cases we could move this `compute_txid` call up to before the match, no?

crates/asm/subprotocols/admin/src/handler.rs
prajwolrg: I think we can use bincode here. Doesn't have to be JSON.

I was planning to work on using the predicate-fmt once this was merged, so this can be handwaved for now.

crates/asm/subprotocols/checkpointing-v0/src/subprotocol.rs
prajwolrg: I'd argue to put the compute_txid() inside the `logging::warn!`, because when we use `strata_asm_common::logging`, all the tracing macros becomes a noop inside ZkVm, avoiding the txid calculation.

crates/asm/subprotocols/checkpointing-v0/src/error.rs
storopoli: Yes please...

crates/asm/subprotocols/checkpoint-v0/src/error.rs
storopoli: Maybe not, if you do `pub use error::CheckpointV0Error` in the `lib.rs`/`mod.rs`.
Or you do a `prelude` module...

storopoli: I would say to keep the `*Error` suffix in the enum and remove from the variants like
`MyError::BadStuff` instead of `MyError::BadStuffError`.

crates/asm/subprotocols/checkpointing-v0/src/verification.rs
```diff
@@ -0,0 +1,148 @@
...
+fn verify_checkpoint_signature(
+    signed_checkpoint: &SignedCheckpoint,
+    cred_rule: &CredRule,
+) -> Result<(), CheckpointV0Error> {
+    if verify_signed_checkpoint_sig(signed_checkpoint, cred_rule) {
+        Ok(())
+    } else {
+        Err(CheckpointV0Error::InvalidSignature)
+    }
+}
```
storopoli: I don't get...
Why creating this wrapper?! Just call the original `verify_signed_checkpoint_sig` whatever you're calling it and do a `map_err(CheckpointV0Error::InvalidSignature)`...

crates/asm/subprotocols/checkpointing-v0/src/verification.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    if is_empty_proof {
+        // TODO: Remove empty-proof acceptance once the predicate framework lands so this
+        // code cannot reach production unchanged.
+        if allow_empty {
+            logging::warn!(
+                epoch = checkpoint.batch_info().epoch(),
+                "Accepting empty checkpoint proof"
+            );
+            return Ok(());
```
storopoli: You could do a combination of both:

1. feature-gating to a `test-foo` or something; and
2. enabling only on debug builds (non-release mode).

crates/asm/subprotocols/checkpoint-v0/src/types.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// The last verified checkpoint
+    pub last_checkpoint: Option<PrimitivesCheckpoint>,
+    /// Last L1 block where we got a valid checkpoint
+    pub last_checkpoint_l1_height: u64,
+    /// Current epoch we've verified up to
+    pub current_verified_epoch: u64,
+    /// Credential rule governing signature verification
+    pub cred_rule: CredRule,
+    /// Rollup verifying key used for proof verification
+    pub rollup_verifying_key: RollupVerifyingKey,
```
storopoli, line 35: better readability
```suggestion
    /// The last verified checkpoint
    pub last_checkpoint: Option<PrimitivesCheckpoint>,

    /// Last L1 block where we got a valid checkpoint
    pub last_checkpoint_l1_height: u64,

    /// Current epoch we've verified up to
    pub current_verified_epoch: u64,

    /// Credential rule governing signature verification
    pub cred_rule: CredRule,

    /// Rollup verifying key used for proof verification
    pub rollup_verifying_key: RollupVerifyingKey,
```

crates/asm/subprotocols/checkpoint-v0/src/types.rs
```diff
@@ -0,0 +1,106 @@
...
+/// some SPS-62 concepts for future compatibility.
+/// Configuration parameters don't need serialization - they're provided at init.
+#[derive(Clone, Debug)]
+pub struct CheckpointV0VerificationParams {
+    /// Genesis L1 block commitment
+    pub genesis_l1_block: L1BlockCommitment,
+    /// Credential rule governing signature verification
+    pub cred_rule: CredRule,
+    /// Rollup verifying key for proof verification
+    pub rollup_verifying_key: RollupVerifyingKey,
```
storopoli, line 52:
```suggestion
    /// Genesis L1 block commitment
    pub genesis_l1_block: L1BlockCommitment,

    /// Credential rule governing signature verification
    pub cred_rule: CredRule,

    /// Rollup verifying key for proof verification
    pub rollup_verifying_key: RollupVerifyingKey,
```

crates/asm/subprotocols/checkpointing-v0/src/subprotocol.rs
storopoli: Yes, great point @prajwolrg

crates/asm/subprotocols/checkpointing-v0/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+//!
+//! # SPS-62 Compatibility Notes
+//!
+//! This is checkpointing v0, which prioritizes feature parity with the current system.
+//! Future versions will be fully SPS-62 compliant. Current SPS-62 concepts incorporated:
+//!
+//! - Envelope transaction structure (SPS-50)
+//! - Basic verification flow concepts
+//! - Placeholder structures for future SPS-62 migration
+// Module declarations
```
storopoli: Do we need this comment?

crates/asm/subprotocols/checkpointing-v0/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "strata-asm-proto-checkpointing-v0"
```
storopoli:
```suggestion
name = "strata-asm-proto-checkpoint-v0"
```

storopoli: As Trey pointed out, rename the dir as well

crates/asm/txs/checkpointing/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "strata-asm-proto-checkpointing-txs"
```
storopoli:
```suggestion
name = "strata-asm-proto-checkpoint-txs"
```
As Trey already suggested, also rename the dir.

crates/asm/txs/checkpointing/src/parser.rs
```diff
@@ -0,0 +1,67 @@
...
+use crate::{
+    constants::OL_STF_CHECKPOINT_TX_TYPE,
+    errors::{CheckpointTxError, CheckpointTxResult},
+};
+
+/// Extract the signed checkpoint payload from an SPS-50-tagged transaction input.
+///
+/// Ensures the transaction carries the expected checkpoint tx-type tag, unwraps the taproot
+/// envelope script from the first input witness, parses the embedded payload, and deserializes it
+/// into a [`SignedCheckpoint`].
```
storopoli: Make this procedure an item list instead of prose?

storopoli (changes requested):
Some suggestions on code quality

crates/asm/subprotocols/checkpoint-v0/Cargo.toml
irnb: STR-1713

crates/asm/subprotocols/admin/src/handler.rs
```diff
@@ -21,44 +23,68 @@ use crate::{
...
+            }
             UpdateAction::OperatorSet(_update) => {
-                // TODO: STR-1721 Set an InterProtoMsg to the Bridge Subprotocol;
+                // TODO: STR-1721 Set an InterProtoMsg to the Bridge Subprotocol
             }
-            UpdateAction::Sequencer(_update) => {
-                // TODO: STF-1721 Send a InterprotoMsg to the Checkpoint subprotocol
+            UpdateAction::Sequencer(update) => {
+                let new_key = update.into_inner();
+                relay_checkpointing_sequencer_update(relayer, new_key);
```
barakshani: Why this is actually part of the checkpoint subprotocol? I don't have a better alternative, but still I wonder.

prajwolrg: This is part of the admin subprotocol. We are passing `InterProtoMsg` to checkpoint subprotocol. Right now checkpoint subprotocol is responsible for verifying if the Checkpoint message is signed by the expected sequencer.

crates/asm/subprotocols/checkpointing-v0/src/subprotocol.rs
irnb: This check is already present, it’s actually the very first condition evaluated in the function’s control flow. 

```
pub fn extract_signed_checkpoint_from_envelope(
      tx: &TxInputRef<'_>,
) -> CheckpointTxResult<SignedCheckpoint> {
      let tx_type = tx.tag().tx_type();
      if tx_type != OL_STF_CHECKPOINT_TX_TYPE {
            return Err(CheckpointTxError::UnexpectedTxType {
                  expected: OL_STF_CHECKPOINT_TX_TYPE,
                  actual: tx_type,
            });
       }
 ....
 }
```

That said, I agree it makes sense to move it up to the higher layer for clarity and separation of concerns.

delbonis (approved):

storopoli (approved):
ACK 449380fa19971488863ecf5ea3be308c95874dfa

prajwolrg (approved):

PR #1073, opened 2025-09-29
perf: optimize operator tracking with bitmap implementation

prajwolrg:
## Description

Implements a memory-efficient `OperatorBitmap` structure to replace `Vec<OperatorIdx>` across the bridge subprotocol to reduce memory usage for operator set tracking while improving performance of set operations.
  
 I used Claude Code for generating tests and updating docstrings.

### Type of Change

- [ ] Documentation update
- [X] Refactor
- [X] New or updated tests
- [X] Dependency Update

## Checklist

      in the body of this PR.

### DIFF
--- a/crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
+++ b/crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
@@ -443,10 +443,7 @@ impl AssignmentTable {
 
 #[cfg(test)]
 mod tests {
-    use strata_primitives::{
-        l1::{BitcoinBlockHeight, L1BlockId},
-        operator::OperatorIdx,
-    };
+    use strata_primitives::l1::{BitcoinBlockHeight, L1BlockId};
     use strata_test_utils::ArbitraryGenerator;
 
     use super::*;
--- a/crates/asm/subprotocols/bridge-v1/src/state/deposit.rs
+++ b/crates/asm/subprotocols/bridge-v1/src/state/deposit.rs
@@ -9,7 +9,6 @@ use arbitrary::Arbitrary;
 use borsh::{BorshDeserialize, BorshSerialize};
 use strata_primitives::{
     l1::{BitcoinAmount, OutputRef},
-    operator::OperatorIdx,
     sorted_vec::SortedVec,
 };
 
--- a/crates/asm/subprotocols/bridge-v1/src/state/bitmap.rs
+++ b/crates/asm/subprotocols/bridge-v1/src/state/bitmap.rs
@@ -6,7 +6,7 @@
 use arbitrary::Arbitrary;
 use bitvec::prelude::*;
 use borsh::{BorshDeserialize, BorshSerialize};
-use strata_primitives::bridge::OperatorIdx;
+use strata_primitives::operator::OperatorIdx;
 
 use crate::BitmapError;
 


crates/asm/subprotocols/bridge-v1/src/errors.rs
```diff
@@ -40,24 +58,6 @@ pub enum DepositValidationError {
...
-    #[error("failed to parse withdrawal fulfillment tx")]
-    WithdrawalTxParse(#[from] WithdrawalParseError),
-
-    #[error("failed to parse withdrawal fulfillment tx")]
-    WithdrawalTxProcess(#[from] WithdrawalValidationError),
-
-    #[error("unsupported tx type {0}")]
-    UnsupportedTxType(TxType),
-}
-
```
storopoli, line 60: Is this diff needed?

crates/asm/subprotocols/bridge-v1/src/errors.rs
```diff
@@ -105,3 +105,23 @@ pub enum WithdrawalCommandError {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            BitmapError::IndexOutOfBounds(idx) => {
+                write!(f, "Index {} is out of bounds for sequential bitmap", idx)
+            }
+        }
+    }
+}
+
+impl std::error::Error for BitmapError {}
```
storopoli: Why not just add `thiserror` to the `Cargo.toml` and add the `Error` to the derive?
Less code to maintain.

storopoli: Actually it's already there!
So, no excuses 😂

crates/asm/subprotocols/bridge-v1/src/errors.rs
prajwolrg: Yes, I think this helps in readability.

crates/asm/subprotocols/bridge-v1/src/state/deposit.rs
```diff
@@ -123,10 +122,23 @@ impl DepositEntry {
     }
 
     /// Returns the historical set of operators that formed the N/N multisig.
-    pub fn notary_operators(&self) -> &[OperatorIdx] {
+    ///
+    /// # Returns
+    ///
+    /// A reference to the operator bitmap that contains the operators active for this deposit.
```
Rajil1213: Nit: This `Returns` section is redundant with the top-level heading (I do like the body of this section more than the top-level heading though).

crates/asm/subprotocols/bridge-v1/src/state/deposit.rs
```diff
@@ -123,10 +122,23 @@ impl DepositEntry {
...
+    pub fn notary_operators(&self) -> &OperatorBitmap {
         &self.notary_operators
     }
 
+    /// Returns the indices of operators that formed the N/N multisig for this deposit.
+    ///
+    /// # Returns
+    ///
+    /// Vector containing [`OperatorIdx`] for operators that were active for this deposit.
```
Rajil1213: Nit: As in the previous getter, this `Returns` section is redundant.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -91,17 +93,188 @@ impl OperatorEntry {
...
+/// - Create a bitmap with `new_with_size(5, true)` to get operators 0, 1, 2, 3, 4 active
+/// - Use `is_active(idx)` to check if an operator is active
+/// - Use `try_set(idx, false)` to deactivate an operator
+/// - Use `to_indices()` to get a Vec of active operator indices
+/// - Sequential index constraint: can only add operators one by one without gaps
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct OperatorBitmap {
+    /// Bitmap where bit `i` is set if operator index `i` is active.
+    /// Uses `BitVec<u8>` for dynamic sizing and memory efficiency.
+    bits: BitVec<u8>,
```
Rajil1213: Does this mean that it supports upto `u8::MAX` operators?

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -91,17 +93,188 @@ impl OperatorEntry {
...
+    fn deref(&self) -> &Self::Target {
+        &self.bits
+    }
+}
+
+impl DerefMut for OperatorBitmap {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.bits
+    }
+}
```
Rajil1213: This makes it convenient to the implementor of this crate but it also exposes the internal `BitVec<u8>` APIs to the consumers. This makes `OperatorBitmap` "behave" in the same way as a `BitVec<u8>`. This may or may not be desirable in all contexts. So I'd encourage thinking through this a bit more to see if this is really what we want.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -231,20 +414,40 @@ impl OperatorTable {
...
     }
 
+    /// Returns whether this operator is part of the current N/N multisig set.
+    ///
+    /// Operators in the current multisig are eligible for new task assignments, while operators
+    /// not in the current multisig are preserved in the table but not assigned new tasks.
+    ///
+    /// # Returns
+    ///
+    /// `true` if the operator is in the current multisig, `false` otherwise.
```
Rajil1213: Based on the implementation of `is_active`, this also returns `false` if the index is out-of-bounds. So that should be mentioned.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -231,20 +414,40 @@ impl OperatorTable {
...
+    }
+
+    /// Returns a reference to the bitmap of currently active multisig operators.
+    ///
+    /// The bitmap tracks which operators are part of the current N/N multisig configuration.
+    /// This is used for assignment creation and deposit processing.
+    ///
+    /// # Returns
+    ///
+    /// Reference to the [`OperatorBitmap`] representing current multisig membership.
```
Rajil1213: Nit: This section is redundant.

crates/asm/subprotocols/bridge-v1/src/errors.rs
```diff
@@ -104,4 +104,17 @@ pub enum WithdrawalCommandError {
...
+    BitmapError(#[from] BitmapError),
+}
+
+/// Error type for OperatorBitmap operations.
+#[derive(Debug, Error, PartialEq)]
+pub enum BitmapError {
+    /// Attempted to set a bit at an index that would create a gap in the bitmap.
+    /// Only sequential indices are allowed.
+    #[error("Index {0} is out of bounds for sequential bitmap")]
+    IndexOutOfBounds(OperatorIdx),
```
Rajil1213: It might be useful to also mention the bounds here.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -204,23 +222,26 @@ impl AssignmentEntry {
         &mut self,
         new_operator_fee: BitcoinAmount,
         seed: L1BlockId,
-        current_active_operators: &[OperatorIdx],
+        current_active_operators: &OperatorBitmap,
     ) -> Result<(), WithdrawalCommandError> {
-        self.previous_assignees.push(self.current_assignee);
+        self.previous_assignees
+            .try_set(self.current_assignee, true)?;
```
Rajil1213: Since this change introduces a new error path, it should be documented.

Rajil1213 (changes requested):
There is something to be said about whether this kind of optimization is even needed given that there are not going to be that many operators. Barring that, I've left a few nits (both on the APIs and the docs) but there are no major blockers.

storopoli (dismissed):
ACK a26e6a13085283f7f31494ba7e618ee6ba73f66f

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
prajwolrg: No. This means that we are using `Vec<u8>`, when storing the bitmap. By default BitVec uses `Vec<usize>`.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
prajwolrg: Thought about it and it made sense to remove the `Deref`. Thanks for the encouragement. Appreciate it.

prajwolrg:
> There is something to be said about whether this kind of optimization is even needed given that there are not going to be that many operators. Barring that, I've left a few nits (both on the APIs and the docs) but there are no major blockers.

The optimization is not that significant. I worked on it because it was brought up in some old PR.

crates/asm/subprotocols/bridge-v1/src/state/deposit.rs
```diff
@@ -42,7 +40,7 @@ use crate::errors::DepositValidationError;
 /// formed the N/N multisig when this deposit was locked. Any one honest operator
 /// from this set can properly process user withdrawals. We store this historical
 /// set because the active operator set may change over time.
-#[derive(Clone, Debug, Eq, PartialEq, BorshDeserialize, BorshSerialize, Serialize, Deserialize)]
+#[derive(Clone, Debug, Eq, PartialEq, BorshDeserialize, BorshSerialize)]
```
Rajil1213, line 43: What is the reason for this change? Note that this is a breaking change (and it actually breaks the mock checkpoint tooling in the bridge because it reads the deposit table from a JSON file).

crates/asm/subprotocols/bridge-v1/src/errors.rs
```diff
@@ -95,13 +95,61 @@ pub enum WithdrawalCommandError {
...
+    )]
+    MismatchedBitmapLengths {
+        notary_len: usize,
+        previous_len: usize,
+    },
+
+    /// Current active operators bitmap is shorter than notary operators bitmap.
+    /// This indicates a system inconsistency since operator indices are only appended.
+    #[error(
+        "Current active operators bitmap length ({active_len}) is shorter than notary operators length ({notary_len}). This should never happen as operator bitmaps only grow."
```
Rajil1213, line 131: This is a crucial detail in the sense that the operator indices continue to grow regardless of exits. So it is theoretically possible for an operator index to exceed `u32::MAX` (max for `OperatorIdx`) if there are a lot of unique operators joining the set over the lifetime of the bridge. I feel like this needs to be documented somewhere (possibly as part of the documentation on `OperatorBitmap` itself).

crates/asm/subprotocols/bridge-v1/src/state/deposit.rs
prajwolrg: This is the internal representation of the in the subprotocol state and I don't think we should be directly exposing this type outside of the crate. It shouldn't be a breaking change (yet) because the subprotocol hasn't been used, and there exists a separate `DepositEntry` type in the `strata-state`.

crates/asm/subprotocols/bridge-v1/src/errors.rs
prajwolrg: I've tried to add it in the relevant places in [82e2570](https://github.com/alpenlabs/alpen/pull/1073/commits/82e25707bbb9bd31e66d862a742e16adffd4dc38).

crates/asm/subprotocols/bridge-v1/src/state/deposit.rs
Rajil1213: Oh okay. Should it still be `pub` if it's an internal representation only?

prajwolrg: The types are pub within pub(crate) modules, which makes them functionally crate-private due to the module boundary.

I explored changing these to explicit `pub(crate)` and it triggered a lot of lint issues `unreachable_pub` and `dead_code`. 

I don't want to make a lot of changes until we have a clearer picture of what external consumers need. 

I am also not a big fan of having separate types just for RPCs. So it might make sense to add `Serialize`, `Deserialize` back to this struct.

What do you think is the right direction here?

Rajil1213: If there is a separate DeposiyEntry type in the state crate which does have serialize/deserialize and this type is going to be internal anyways, nothing is going to break when adding or removing these traits. So we can add these back if they become necessary. I think it's fine as is.

PR #1078, opened 2025-10-01
STR-1739: ASM Logs in CSM

evgenyzdanovich:
## Description

**Currently DRAFT WIP**.

Done (can be reviewed):

- additional adapters in the service framework for listeners service.
- integration of `ASM` Logs in the `CSM`.
- "servization" of `CSM` worker as a listener.

In progress:
- failing tests (btcio writer?)

**Note:**
parts of the work were done with the use of Claude.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

      in the body of this PR.


crates/consensus-logic/src/sync_manager.rs
```diff
@@ -133,23 +136,40 @@ pub fn start_sync_tasks<E: ExecEngineCtl + Sync + Send + 'static>(
...
         status_channel,
     })
 }
 
+fn spawn_csm_listener<W: strata_asm_worker::WorkerContext + Send + Sync + 'static>(
+    executor: &TaskExecutor,
+    params: Arc<Params>,
+    storage: Arc<NodeStorage>,
+    status_channel: StatusChannel,
+    asm_monitor: &strata_service::ServiceMonitor<strata_asm_worker::AsmWorkerService<W>>,
```
delbonis: This is the kind of situation I wanted to avoid that I was gesturing when I mentioned that we should slim down the service monitor's def to just have the monitor type and not be generic over the whole service.  Because now the CSM crate is required to know abou the ASM worker context trait.

crates/service/src/status.rs
```diff
@@ -30,6 +30,44 @@ impl<S: Service> ServiceMonitor<S> {
...
+    ///
+    /// This allows another service to listen to status updates from this service.
+    /// Returns an MPSC receiver that will receive status updates, and spawns a task
+    /// to forward watch channel updates to it.
+    ///
+    /// The listener task will automatically exit when the monitored service exits.
+    pub fn create_listener_input(
+        &self,
+        executor: &strata_tasks::TaskExecutor,
+    ) -> crate::TokioMpscInput<S::Status> {
```
delbonis: Should import this at the top instead of using `crate::` here.

crates/service/src/status.rs
```diff
@@ -30,6 +30,44 @@ impl<S: Service> ServiceMonitor<S> {
...
+            while watch_rx.changed().await.is_ok() {
+                // Get the new status and send it
+                let status = watch_rx.borrow_and_update().clone();
+                if tx.send(status).await.is_err() {
+                    // Receiver dropped, exit
+                    break;
+                }
+            }
+            // Watch channel closed - monitored service exited
+        });
```
delbonis, line 65: ~~I think it should be possible to make a new input type that behaves like `TokioMpscInput` but for watch channels so we don't have to launch a service just to feed inputs from one place to another.~~

Ah you did make that already.

crates/service/src/adapters.rs
```diff
@@ -284,12 +284,121 @@ impl<T> FromIterator<T> for VecInput<T> {
...
+///
+/// When the watch channel closes (monitored service exits), this adapter returns
+/// `None` to signal the listener should also exit.
+pub struct StatusMonitorInput<S: Service> {
+    rx: watch::Receiver<S::Status>,
+    closed: bool,
+    /// Optional Tokio runtime handle for sync services.
+    /// Required when using this input with `SyncService`.
+    handle: Option<tokio::runtime::Handle>,
+}
```
delbonis: This only being optional makes is so there's nothing stopping you from misusing it.  We can just not implement the sync trait on this and rely on a consumer to use the adapter if they want to use it from a sync context.

Also in line with the other points, we could loosen this up a bit and call it `TokioWatchInput` and make it generic over just the message type instead of the whole service.

crates/csm-worker/src/service.rs
```diff
@@ -0,0 +1,67 @@
...
+
+        // Process each checkpoint update log
+        for log in logs {
+            if let Err(e) = process_checkpoint_log(state, log, &asm_block) {
+                error!(%asm_block, err = %e, "Failed to process checkpoint log");
+                // Continue processing other logs instead of failing completely
+            }
+        }
+
+        info!(%asm_block, "CSM successfully processed ASM logs.");
```
delbonis: Since `%asm_block` is in all of these, we could wrap this in a span that has that parameter so the individual logs don't need it.  However, the harness kinda already does this so maybe we don't need it?

```rs
        let input_span = debug_span!("handlemsg", %service, ?input);
```

But this *is* pretty verbose, which is why it's `debug!` and not `info!`.  We could still have a span with *just* the block ID.

crates/csm-worker/src/service.rs
```diff
@@ -0,0 +1,67 @@
...
+        info!(%asm_block, "CSM is processing ASM logs.");
+
+        // Track which block we're processing
+        state.last_asm_block = Some(asm_block);
+
+        // Extract checkpoint logs from ASM status
+        let logs = asm_status.logs();
+
+        if logs.is_empty() {
+            trace!(%asm_block, "No logs in ASM status update");
```
delbonis: If we're printing when there's no logs then maybe we should print the quantity is when there *are* logs?

crates/csm-worker/src/service.rs
```diff
@@ -0,0 +1,67 @@
...
+            return Ok(Response::Continue);
+        }
+
+        // Process each checkpoint update log
+        for log in logs {
+            if let Err(e) = process_checkpoint_log(state, log, &asm_block) {
+                error!(%asm_block, err = %e, "Failed to process checkpoint log");
+                // Continue processing other logs instead of failing completely
+            }
```
delbonis: It would be better to match on the log type first and then call the handler fn instead of assuming it's a checkpoint and returning early if it's not a checkpoint.

delbonis (changes requested):
Really happy with how this is coming along far.  Main comments are on service input handling and logging.

Do you think you should make the change to how `ServiceMonitor` is defined in this PR or should we do it in a different one?

crates/service/src/status.rs
evgenyzdanovich: done

crates/service/src/adapters.rs
evgenyzdanovich: nice catch (on banning weird sync interface and rely on adapter).

done.

crates/consensus-logic/src/sync_manager.rs
evgenyzdanovich: done

crates/service/src/lib.rs
```diff
@@ -97,12 +97,11 @@ mod status;
 mod sync_worker;
 mod types;
 
+// Re-export commonly used types for listener pattern
+pub use adapters::TokioWatchInput;
 pub use adapters::*;
```
delbonis: This seems redundant?

crates/consensus-logic/src/sync_manager.rs
```diff
@@ -133,23 +136,40 @@ pub fn start_sync_tasks<E: ExecEngineCtl + Sync + Send + 'static>(
...
+    storage: Arc<NodeStorage>,
+    status_channel: StatusChannel,
+    asm_monitor: &strata_service::ServiceMonitor<AsmWorkerStatus>,
+) -> anyhow::Result<()> {
+    // Create CSM worker state
+    let csm_state = CsmWorkerState::new(params, storage, status_channel)?;
+
+    // Create input that listens to ASM status updates via MPSC
+    // This spawns a forwarder task that converts watch channel updates to MPSC messages
+    let async_input = asm_monitor.create_listener_input(executor);
```
delbonis: I think these comments are out of date now.

crates/asm/worker/src/builder.rs
```diff
@@ -84,7 +84,7 @@ impl<W> AsmWorkerBuilder<W> {
         let command_handle = service_builder.create_command_handle(64);
 
         // Launch the service using the sync worker.
-        let service_monitor: strata_service::ServiceMonitor<AsmWorkerService<W>> = service_builder
+        let service_monitor = service_builder
             .launch_sync("asm_worker", executor)
             .map_err(|e| WorkerError::Unexpected(format!("failed to launch service: {}", e)))?;
```
delbonis: Would it make more ergonomic sense to change the return type and have this just propagate up the anyhow error instead of wrapping it?  Since it's not really an error *with the worker*, it's an error with launching the worker's *service*, and the service framework uses anyhow errors for better tracking there.

And then since I see `_runtime` is unused in this function can we get rid of that too?  Or is it being used somewhere I am not realizing?

crates/asm/worker/src/handle.rs
```diff
@@ -27,13 +28,13 @@ impl<W: WorkerContext + Send + Sync + 'static> AsmWorkerHandle<W> {
     /// Allows other services to listen to status updates.
     ///
     /// Can be useful for logic that want to listen to logs/updates of ASM state.
-    pub fn get_monitor(&self) -> &Monitor<W> {
+    pub fn get_monitor(&self) -> &ServiceMonitor<AsmWorkerStatus> {
```
delbonis: Could maybe be better named just `monitor` (and probably rename the field).

delbonis (changes requested):
Changes to services look good, this is mostly nits, plus the other review comments.

crates/csm-worker/src/processor.rs
```diff
@@ -0,0 +1,163 @@
...
+    let next_state = ClientState::new(last_finalized, recent.clone());
+
+    // Check if we need to finalize an epoch
+    let old_final_epoch = cur_state.get_declared_final_epoch();
+    let new_final_epoch = next_state.get_declared_final_epoch();
+
+    let should_finalize = match (old_final_epoch, new_final_epoch) {
+        (None, Some(new)) => Some(new),
+        (Some(old), Some(new)) if new.epoch() > old.epoch() => Some(new),
+        _ => None,
```
storopoli, line 133: `update_client_state_with_checkpoint` replaces the client state with `ClientState::new(last_finalized, recent.clone())`, where `recent` is set to `None` whenever we receive a checkpoint log whose epoch is ≤ the currently tracked one.

For a repeated/out-of-order `CheckpointUpdate`, the `else` branch runs, `recent` becomes `None`, and the newly stored `ClientState` ends up with `last_seen_checkpoint == None`.

https://github.com/alpenlabs/alpen/blob/6de96dff5879a976eabe1677be5ea46fc9909927/crates/csm-worker/src/processor.rs#L94-L99

In practice that means a harmless duplicate log erases the currently tracked checkpoint and also reassigns `last_finalized_checkpoint` to the most recently seen (even though nothing new was finalized).

Downstream components that rely on CSM will immediately observe an empty checkpoint state.

This is an issue

crates/csm-worker/src/sync_actions.rs
```diff
@@ -0,0 +1,73 @@
...
+                }
+            };
+
+            ckpt_entry.confirmation_status = CheckpointConfStatus::Confirmed(l1_reference);
+
+            ckpt_db.put_checkpoint_blocking(epoch, ckpt_entry)?;
+        }
+    }
+
+    Ok(())
```
storopoli, line 70: When we `create_checkpoint_from_update` with an empty proof:

https://github.com/alpenlabs/alpen/blob/6de96dff5879a976eabe1677be5ea46fc9909927/crates/csm-worker/src/processor.rs#L144-L162

For epochs where no entry exists yet, the DB now stores a `CheckpointProvingStatus::ProofRead` checkpoint whose proof/sidecar are completely blank. Any later consumer expecting the verified proof will fail, and we effectively discard the only authenticated witness for that checkpoint.

Is that a problem?

crates/csm-worker/src/processor.rs
```diff
@@ -0,0 +1,163 @@
...
+        .try_into_log()
+        .map_err(|e| anyhow::anyhow!("Failed to deserialize CheckpointUpdate: {}", e))?;
+
+    let epoch = checkpoint_update.batch_info.epoch();
+
+    info!(
+        %epoch,
+        %asm_block,
+        checkpoint_txid = ?checkpoint_update.checkpoint_txid,
+        "Processing checkpoint update from ASM log"
```
barakshani: Are there other components that process ASM logs from checkpoints? If so, I'd clarify that it's the _CSM_ that is processing.

crates/csm-worker/src/processor.rs
```diff
@@ -0,0 +1,163 @@
...
+    );
+
+    // Create L1 checkpoint reference from the log data
+    let l1_reference = CheckpointL1Ref::new(
+        *asm_block,
+        checkpoint_update.checkpoint_txid.inner_raw(),
+        checkpoint_update.checkpoint_txid.inner_raw(), // TODO: get wtxid if available
+    );
+
+    // Create L1Checkpoint for client state
```
barakshani, line 61: Can you explain what is happening here? Do we always have a _single_ `CHECKPOINT_UPDATE_LOG_TYPE` per checkpoint, so this is where we inform the client state about the existence of the checkpoint emitting this log?

crates/csm-worker/src/processor.rs
```diff
@@ -0,0 +1,163 @@
...
+/// Update client state with a new checkpoint.
+fn update_client_state_with_checkpoint(
+    state: &mut CsmWorkerState,
+    new_checkpoint: L1Checkpoint,
+    epoch: u64,
+) -> anyhow::Result<()> {
+    // Get the current client state
+    let cur_state = state.cur_state.as_ref();
+
+    // Determine if this checkpoint should be the last finalized or just recent
```
barakshani: Can you clarify what these terms mean? It looks like _recent_ is always the tip while _last finalized_ is the one just before the tip.

crates/csm-worker/src/processor.rs
```diff
@@ -0,0 +1,163 @@
...
+    let (last_finalized, recent) = match cur_state.get_last_checkpoint() {
+        Some(existing) => {
+            // If the new checkpoint is for a later epoch, it becomes recent
+            if epoch > existing.batch_info.epoch() {
+                (Some(existing.clone()), Some(new_checkpoint))
+            } else {
+                // Otherwise keep existing
+                (Some(existing.clone()), None)
```
barakshani, line 114: Actually here it seems that my interpretation in the previous comment breaks. If we don't use these terms elsewhere, perhaps just call it _newer_, or something like that.

crates/csm-worker/src/processor.rs
```diff
@@ -0,0 +1,163 @@
...
+            }
+        }
+        None => {
+            // First checkpoint becomes recent
```
barakshani:
```suggestion
            // New checkpoint is the first checkpoint, and it is marked recent

File: crates/csm-worker/src/processor.rs (hunk 17 above)
Comment (barakshani) [inline crates/csm-worker/src/processor.rs:114]: Why do we move the existing checkpoint to be final? Shouldn't this depend on the number of block confirmations?

File: crates/csm-worker/src/processor.rs [hunk 19]
```diff
@@ -0,0 +1,163 @@
...
+
+    // Create new client state
+    let next_state = ClientState::new(last_finalized, recent.clone());
+
+    // Check if we need to finalize an epoch
+    let old_final_epoch = cur_state.get_declared_final_epoch();
+    let new_final_epoch = next_state.get_declared_final_epoch();
+
+    let should_finalize = match (old_final_epoch, new_final_epoch) {
+        (None, Some(new)) => Some(new),
```
barakshani, line 131: For someone with not enough context this seems weird: we just pulled this value as a (new) final epoch, so why do we need to finalise it?? Is this finalisation for storage purposes?

crates/csm-worker/src/sync_actions.rs
```diff
@@ -0,0 +1,73 @@
...
+use strata_state::operation::SyncAction;
+use strata_storage::NodeStorage;
+use tracing::*;
+
+/// Apply a sync action to storage.
+pub(crate) fn apply_action(action: SyncAction, storage: &Arc<NodeStorage>) -> anyhow::Result<()> {
+    let ckpt_db = storage.checkpoint();
+    match action {
+        SyncAction::FinalizeEpoch(epoch_comm) => {
+            // For the fork choice manager this gets picked up later.  We don't have
```
barakshani:
```suggestion
            // For the fork choice manager this gets picked up later. We don't have

File: crates/csm-worker/src/sync_actions.rs [hunk 21]
```diff
@@ -0,0 +1,73 @@
...
+            // to do anything here *necessarily*.
```
barakshani: I don't understand this comment.

crates/csm-worker/src/processor.rs
```diff
@@ -0,0 +1,163 @@
...
+    // Create empty sidecar - checkpoint was already verified by ASM
+    let sidecar = CheckpointSidecar::new(vec![]);
+
+    Checkpoint::new(
+        update.batch_info.clone(),
+        BatchTransition {
+            epoch,
+            chainstate_transition: update.chainstate_transition,
+        },
+        Default::default(), // Empty proof - actual proof was already verified by ASM
```
barakshani, line 180: This makes me wonder: why do we need to store historical checkpoints at all?

storopoli: Related: https://github.com/alpenlabs/alpen/pull/1078#discussion_r2399551133

crates/csm-worker/src/service.rs
prajwolrg, line 65: All the ASM Logs are meant to be consumed. Right now there are two entities that consume the ASM Logs:
1. **Moho** consumes `ASM_STF_UPDATE_LOG` and `NEW_EXPORT_ENTRY_LOG`.
2. **CSM** should consume all the remaining logs - `DEPOSIT_LOG`, `FORCED_INCLUSION_LOG`, `CHECKPOINT_UPDATE_LOG` and `OL_STF_UPDATE_LOG`.

`DEPOSIT_LOG` and `FORCED_INCLUSION_LOG` is later needed by the OL Block Builder, so it should be saved somewhere in the DB. 

----

On a different note, we might have to rename CSM to better reflect this. Right now CSM means Client State Manager or Checkpoint State Manager. CC: @delbonis @storopoli @barakshani.

crates/csm-worker/src/processor.rs
prajwolrg: Related: https://github.com/alpenlabs/alpen/pull/1078/files#r2409192201

crates/csm-worker/src/processor.rs
prajwolrg: We always have a single `CHECKPOINT_UPDATE_LOG` per checkpoint transaction. In case there are two valid checkpoint transactions in a single block, there will be two checkpoint update logs, and two client state updates. But since each log is handled separately in `CsmWorkerService::process_input`, it should be fine.

crates/csm-worker/src/processor.rs
prajwolrg: This seems to come from the older design. In the old design, the checkpoint was considered final if the new checkpoint builds on top of it, and since the checkpoint was recursive and proved the validity of the inclusion of checkpoint it is built upon, it was okay to consider it finalized as we were only looking at blocks that were "mature".

crates/csm-worker/src/processor.rs
prajwolrg: https://github.com/alpenlabs/alpen/pull/1078/files#r2409211335

crates/csm-worker/src/processor.rs
prajwolrg: The `CheckpointUpdate` log cannot be repeated/out-of-order unless there is a reorg. In case of reorg the ClientState reconstruction should be handled separately. But I do see the issue with the implementation.

I think it's better if we save the valid `L1Checkpoint` in the DB separately, and then on the DbManager have logic for what the finalized and latest checkpoint is.

crates/csm-worker/src/processor.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    Ok(())
+}
+
+/// Create a [`Checkpoint`] from a [`CheckpointUpdate`] log.
+///
+/// Note: The log doesn't contain the full signed checkpoint, so we reconstruct
+/// what we can. The signature verification was already done by ASM.
+fn create_checkpoint_from_update(update: &CheckpointUpdate) -> Checkpoint {
```
prajwolrg, line 168: This function seems to be created for compatibility reason to avoid making larger changes, which is fine for me. This will be largely changed as we move to the new OL STF, where the new checkpoint structure will be different than the existing ones. We should add a TODO to clean it up if that's the case.

prajwolrg (changes requested):
Apart from the log handling, looks good to me.

---

It seems you’ve intentionally avoided making major changes to the existing data structures and legacy code in this PR, which is fine. Maybe we should also keep a track of the changes that needs to be done in the upcoming PRs.

crates/csm-worker/src/service.rs
evgenyzdanovich: done

crates/csm-worker/src/service.rs
evgenyzdanovich: done

crates/csm-worker/src/processor.rs
evgenyzdanovich: correct

crates/csm-worker/src/processor.rs
evgenyzdanovich: done

crates/csm-worker/src/processor.rs
evgenyzdanovich: done, thx

crates/csm-worker/src/processor.rs
evgenyzdanovich: agree, thx

crates/csm-worker/src/service.rs
evgenyzdanovich: made it better with a separate little dispatching based on a type ID

crates/csm-worker/src/service.rs
evgenyzdanovich: agree, this is too verbose. 
I removed the `asm_block` as it already contained in the handlemsg.

I also don't see any point in creating a span because it will be an inner span for `handlemsg` and it's even more verbose.

crates/csm-worker/src/processor.rs
```diff
@@ -0,0 +1,178 @@
...
+use strata_primitives::prelude::*;
+use strata_state::{
+    client_state::{CheckpointL1Ref, ClientState, L1Checkpoint},
+    operation::{ClientUpdateOutput, SyncAction},
+};
+use tracing::*;
+
+use crate::{state::CsmWorkerState, sync_actions::apply_action};
+
+pub(crate) fn process_logs(
```
bewakes: Could you rename this to `process_log` since it takes in a single log?

crates/csm-worker/src/processor.rs
```diff
@@ -0,0 +1,178 @@
...
+    log: &AsmLogEntry,
+    asm_block: &L1BlockCommitment,
+) -> anyhow::Result<()> {
+    // Assert that the dispatch is fine.
+    assert_eq!(log.ty(), Some(CHECKPOINT_UPDATE_LOG_TYPE));
+
+    // Deserialize the checkpoint update using the AsmLog trait
+    let checkpoint_update: CheckpointUpdate = log
+        .try_into_log()
+        .map_err(|e| anyhow::anyhow!("Failed to deserialize CheckpointUpdate: {}", e))?;
```
bewakes: Suggestion: It might be a better idea to `try_into_log()` at the calling site and have this function just work with `CheckpointUpdate`. This way we get better compile time safety and we don't need `assert_eq!` calls.

crates/csm-worker/src/processor.rs
```diff
@@ -0,0 +1,178 @@
...
+    Checkpoint::new(
+        update.batch_info.clone(),
+        BatchTransition {
+            epoch,
+            chainstate_transition: update.chainstate_transition,
+        },
+        Default::default(), // Empty proof - actual proof was already verified by ASM
+        sidecar,
+    )
+}
```
bewakes, line 183: I would strongly suggest to add unit tests, particularly for `process_checkpoint_log`.

barakshani: Let's please make sure this happens.

crates/csm-worker/src/sync_actions.rs
evgenyzdanovich: The checkpoint entry is being created somewhere by the sequencer (with a valid proof).
 
What you observed is a fallback: we create a checkpoint entry and only set put in case the expected entry doesn't exist: `csm-worker/sync_actions.rs`. If it (expectedly) exist, the proof is intact.

Discrepancy can happen with example for the cases where we delete checkpoints via `dbtool`.

crates/csm-worker/src/processor.rs
evgenyzdanovich: Correct. Currently, in this logic there are still traces of the older design.
in the future it'll be simplified and parts of it taken down.

Left a comment, resolving for now.

crates/csm-worker/src/processor.rs
evgenyzdanovich: > why do we need to store historical checkpoints at all?
it general, I don't think we *must* store all historical checkpoints (we only need just last *few*, at the worst case). 
In practice, it's for easier bookkeeping and to easier guarantee our storage is persistent.

evgenyzdanovich: on a related comment, I mentioned that having checkpoint with an empty proof is a fallback. We will have checkpoint entry with a proper non-empty proof (unless the checkpoint was deleted).

crates/csm-worker/src/processor.rs
evgenyzdanovich: yeah, working on that.

evgenyzdanovich: done

crates/csm-worker/src/processor.rs
evgenyzdanovich: agree. I left it as a TODO.

crates/csm-worker/src/processor.rs
```diff
@@ -0,0 +1,499 @@
...
+        Some(CHECKPOINT_UPDATE_LOG_TYPE) => {
+            let ckpt_upd = log
+                .try_into_log()
+                .map_err(|e| anyhow::anyhow!("Failed to deserialize CheckpointUpdate: {}", e))?;
+
+            return process_checkpoint_log(state, &ckpt_upd, asm_block);
+        }
+        Some(log_type) => {
+            warn!(log_type, "not yet supported");
+        }
```
delbonis, line 31: Should just silently discard unsupported log types here, since they're not wrong we just don't care about them.

prajwolrg: added a `debug!` log for now in [66e4877](https://github.com/alpenlabs/alpen/pull/1078/commits/66e487701f8183682b8496da454be75f269309e0) to help with debugging.

PR #1081, opened 2025-10-03
feat!: replacing all `u64`s with proper bitcoin `Amount` and `absolute::Height`

storopoli:
## Description

Replaces all the `u64`s for proper `rust-bitcoin` `Amount`s and `absolute::Height`.

Planned with the help of Cursor's ChatGPT-5 and executed with the help of Claude Code Sonnet-5.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Serde vs Bincode is annoying... See 7df38dcfe08c14144d0002c0216257cbe75cc0d9
I also did a `TODO: rename to max_address_length` that I saw somewhere and fixed the tests to use that new renamed field.

## Checklist

      in the body of this PR.

## Related Issues

STR-1446

### DIFF
--- a/functional-tests/envs/rollup_params_cfg.py
+++ b/functional-tests/envs/rollup_params_cfg.py
@@ -80,7 +80,7 @@ class RollupConfig(BaseModel):
     evm_genesis_block_state_root: StrBuf32
     l1_reorg_safe_depth: int
     target_l2_batch_size: int
-    address_length: int
+    max_address_length: int
     deposit_amount: int
     rollup_vk: RollupVk
     dispatch_assignment_dur: int


crates/primitives/src/l1/block.rs
```diff
@@ -115,20 +240,41 @@ impl fmt::Debug for L1BlockCommitment {
...
+    pub fn new(height: absolute::Height, blkid: L1BlockId) -> Self {
         Self { height, blkid }
     }
 
-    pub fn height(&self) -> u64 {
+    /// Create a new L1 block commitment from a u64 height.
+    ///
+    /// Returns `None` if the height is invalid (greater than u32::MAX).
+    pub fn from_height_u64(height: u64, blkid: L1BlockId) -> Option<Self> {
+        let height = absolute::Height::from_consensus(height as u32).ok()?;
```
delbonis, line 263: Honestly it might make sense to "just panic" in this case.  If all calls to this function are all just panics (and there's no sensible error handling path) then there's no point.

Maybe I'm wrong, not sure what the right path to take here is.  I was wrong for trying to be future-proof with using u64 here.

storopoli: We still need 1 more hard-fork for the u32 timestamp bug in bitcoin 😂

storopoli: So, we'll never know...

PR #1083, opened 2025-10-07
Re-enable disabled tests requiring bridge

voidash:
## Description

for https://alpenlabs.atlassian.net/browse/STR-1727

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

      in the body of this PR.

### DIFF
--- a/functional-tests/tests/el_bridge_precompile.py
+++ b/functional-tests/tests/el_bridge_precompile.py
@@ -6,7 +6,7 @@ from envs import net_settings, testenv
 from mixins.bridge_out_precompile_contract_mixin import BridgePrecompileMixin
 from utils import *
 from utils.constants import PRECOMPILE_BRIDGEOUT_ADDRESS
-from utils.wait.strata import StrataWaiter
+from utils.wait.reth import RethWaiter
 
 withdrawal_intent_event_abi = {
     "anonymous": False,
@@ -65,7 +65,7 @@ class ElBridgePrecompileTest(BridgePrecompileMixin):
             }
         )
 
-        receipt_waiter = StrataWaiter(web3.eth, self.logger, 60, 0.5)
+        receipt_waiter = RethWaiter(web3.eth, self.logger, 60, 0.5)
         receipt = receipt_waiter.wait_until_tx_included_in_block(txid.hex())
 
         assert receipt.status == 1, "precompile transaction failed"
--- a/functional-tests/utils/wait/strata.py
+++ b/functional-tests/utils/wait/strata.py
@@ -321,19 +321,3 @@ class StrataWaiter(RpcWaiter):
             timeout=timeout,
             error_with=f"Timeout: Checkpoint index did not increment to expected value({idx})",
         )
-
-    def wait_until_tx_included_in_block(self, txid: str):
-        def _query():
-            try:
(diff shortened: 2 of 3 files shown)


functional-tests/utils/utils.py
```diff
@@ -652,3 +653,28 @@ def check_initial_eth_balance(rethrpc, address, debug_fn=print):
...
+    else:
+        path = os.path.join(ctx.datadir_root, env, "_init")
+
+    print(path)
+    priv_keys = []
+    opkeys = sorted(
+        filter(lambda file: file.startswith("opkey"), os.listdir(path)),
+        key=lambda x: int("".join(filter(str.isdigit, x))),
+    )
+    print(opkeys)
```
delbonis: Are these prints needed?  If so then use `logging` so they get picked up properly.

Edit: oh I see these were from before this PR, it'd be good to convert over the prints in these updated tests

functional-tests/tests/el_bridge_precompile.py
```diff
@@ -46,24 +48,36 @@ def main(self, ctx: flexitest.RunContext):
...
+            except Exception as e:
+                return e
+
+        receipt = wait_until_with_value(
+            check_transaction_and_blocks,
+            lambda result: not isinstance(result, Exception),
+            error_with="Transaction receipt for txid not available",
+            timeout=60,
+            step=2,
+        )
```
delbonis: This is basically doing a "wait until EVM tx included in block" operation.  If there isn't already waiter for this then can we add it to one so it's available everywhere?

functional-tests/tests/el_bridge_precompile.py
```diff
@@ -46,24 +49,24 @@ def main(self, ctx: flexitest.RunContext):
...
             }
         )
-        self.debug(txid.to_0x_hex())
-
-        # build block
-        time.sleep(2)
 
-        receipt = web3.eth.get_transaction_receipt(txid)
+        receipt_waiter = StrataWaiter(web3.eth, self.logger, 60, 0.5)
+        receipt = receipt_waiter.wait_until_tx_included_in_block(txid.hex())
```
delbonis: This is constructing a `StrataWaiter` but you're passing a Web3 RPC handle into it.  I wish we had added slightly better type hints so that mypy could catch this.

functional-tests/utils/wait/strata.py
```diff
@@ -321,3 +321,19 @@ def wait_until_latest_checkpoint_at(self, idx: int, timeout: int | None = None):
...
+            except Exception as e:
+                return e
+
+        result = self._wait_until_with_value(
+            _query,
+            lambda result: not isinstance(result, Exception),
+            error_with="Transaction receipt for txid not available",
+            timeout=self.timeout,
+        )
+        return result
```
delbonis: This is on `StrataWaiter` which must be constructed with a Strata client's RPC handle.  `get_transaction_receipt` doesn't exist for Strata clients, only in the web3 eth_ namespace, so this function *shouldn't* work.  Move it to `RethWaiter`.

voidash: it works because StrataWaiter takes rpc instance as the argument. But it should be present in RethWaiter.

PR #1085, opened 2025-10-08
(ol): Add basic types

bewakes:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

      in the body of this PR.

### DIFF
--- a/crates/ol/chain-types/src/lib.rs
+++ b/crates/ol/chain-types/src/lib.rs
@@ -1,8 +1,10 @@
 //! All the types related to OL chain.
 mod block;
 mod common;
+mod log;
 mod transaction;
 
 pub use block::*;
 pub use common::*;
+pub use log::*;
 pub use transaction::*;
--- a/crates/ol/chain-types/src/log.rs
+++ b/crates/ol/chain-types/src/log.rs
@@ -1,0 +1,30 @@
+use strata_acct_types::AccountSerial;
+
+/// Log emitted during OL block execution.
+#[derive(Clone, Debug)]
+pub struct OLLog {
+    /// Account this log is related to.
+    // TODO: Determine if this should be here or inside payload
+    account_serial: AccountSerial,
+
+    /// Opaque log payload.
+    payload: Vec<u8>,
+    // TODO: add more concrete fields.
+}
+
+impl OLLog {
+    pub fn new(account_serial: AccountSerial, payload: Vec<u8>) -> Self {
+        Self {
+            account_serial,
+            payload,
+        }
+    }
+
+    pub fn account_serial(&self) -> AccountSerial {
+        self.account_serial
+    }
+
+    pub fn payload(&self) -> &[u8] {
+        &self.payload
+    }
+}


crates/ol/state-types/src/lib.rs
delbonis, line 1: Don't export the internal module structure unless it improves usability/readability.

crates/ol/state-types/src/state.rs
```diff
@@ -0,0 +1,121 @@
+use strata_primitives::buf::Buf32;
+
+/// The Orchestration Layer(OL) State.
+#[derive(Debug, Clone)]
+pub struct OLState {
+    accounts_root: Buf32,
+    l1_view: L1View,
+    cur_slot: u64,
+    cur_epoch: u64,
+    l1_recorded_epoch: EpochCommitment,
```
delbonis: Should be part of L1ViewState.

crates/ol/state-types/src/state.rs
```diff
@@ -0,0 +1,121 @@
...
+    epoch: u64,
+    /// State root at the end of the epoch.
+    state_root: Buf32,
+    /// Terminal slot.
+    terminal_slot: u64,
+    /// Terminal block id.
+    terminal_blockid: Buf32,
+}
+
+impl EpochCommitment {
```
delbonis: Not a state type, really a chain type like OLBlockId would be.

crates/ol/state-types/src/state.rs
```diff
@@ -0,0 +1,121 @@
...
+
+/// The Orchestration Layer(OL) State.
+#[derive(Debug, Clone)]
+pub struct OLState {
+    accounts_root: Buf32,
+    l1_view: L1View,
+    cur_slot: u64,
+    cur_epoch: u64,
+    l1_recorded_epoch: EpochCommitment,
+    total_btc_bridged: u64, // sats
```
storopoli: Can we use `BitcoinAmount`?
I don't want to redo https://alpenlabs.atlassian.net/browse/STR-1446 again!

crates/ol/state-types/src/state.rs
```diff
@@ -0,0 +1,121 @@
...
+        self.cur_epoch
+    }
+
+    pub fn l1_recorded_epoch(&self) -> &EpochCommitment {
+        &self.l1_recorded_epoch
+    }
+
+    pub fn total_btc_bridged(&self) -> u64 {
+        self.total_btc_bridged
+    }
```
storopoli: `BitcoinAmount`

crates/ol/chain-types/src/transaction.rs
```diff
@@ -0,0 +1,84 @@
...
+    },
+}
+
+impl TransactionPayload {
+    pub fn target(&self) -> AccountId {
+        match self {
+            TransactionPayload::SnarkAccountUpdate { target, .. } => *target,
+            TransactionPayload::GenericAccountMessage { target, .. } => *target,
+        }
+    }
```
delbonis: Should return an `Option<&AccountId>` since we can't assume that txs will always have targets.

crates/ol/chain-types/src/transaction.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub fn target(&self) -> AccountId {
+        self.payload().target()
+    }
+
+    pub fn type_id(&self) -> TxTypeId {
+        match self.payload {
+            TransactionPayload::GenericAccountMessage { .. } => 1,
+            TransactionPayload::SnarkAccountUpdate { .. } => 2,
+        }
+    }
```
delbonis, line 38: This should be on `TransactionPayload`.  Also, should have a `TxType` enum instead of just using u16s (use int_enum) to help with this.

crates/ol/chain-types/src/block.rs
```diff
@@ -0,0 +1,213 @@
...
+#[derive(Clone, Debug)]
+pub struct OLBlockHeader {
+    /// The timestamp the block was created at.
+    timestamp: u64,
+    /// Slot the block was created for.
+    slot: Slot,
+    /// Epoch the block was created in.
+    epoch: Epoch,
+    /// Parent block id.
+    parent_blk_id: OLBlockId,
```
delbonis: `blkid` when it's an abbreviation

crates/ol/chain-types/src/block.rs
```diff
@@ -0,0 +1,213 @@
...
+}
+
+impl AsmManifest {
+    pub fn new(blockid: Buf32, logs: Vec<AsmLogEntry>) -> Self {
+        Self { blockid, logs }
+    }
+
+    pub fn blockid(&self) -> &Buf32 {
+        &self.blockid
+    }
```
delbonis, line 239: For consistency this should also be `blkid`, or to reduce ambiguity, `l1blkid`.

crates/ol/chain-types/src/block.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub fn state_root(&self) -> Buf32 {
+        self.state_root
+    }
+}
+
+/// OL block body containing transactions and l1 updates
+#[derive(Clone, Debug)]
+pub struct OLBlockBody {
+    /// The transactions contained in an OL block.
+    txs: Vec<OLTransaction>,
```
delbonis: Wrap this in a "segment" structure to improve ergonomics.

crates/ol/chain-types/src/transaction.rs
```diff
@@ -82,3 +85,37 @@ impl TransactionExtra {
...
+
+    #[inline]
+    fn try_from(v: u16) -> Result<Self, Self::Error> {
+        match v {
+            1 => Ok(TxTypeId::GenericAccountMessage),
+            2 => Ok(TxTypeId::SnarkAccountUpdate),
+            _ => Err(TxTypeError::InvalidTxType(v)),
+        }
+    }
+}
```
delbonis: Use `int-enum` which just does this automatically.

bewakes: It does, but I am not sure if it's best to add a dependency for something trivial like this. Plus we get to use custom error variant like `TxTypeError::InvalidTxType`.

bewakes: I see it's already being used. Cool.

crates/ol/chain-types/src/block.rs
prajwolrg, line 1: Could we do something like

```rust
#[derive(Clone, Debug)]
pub struct Signed<T> {
    inner: T,
    signature: Buf64,
}

```

and then wrap `Block` and `Header` into `Signed<Block>` and `Signed<Header>`?

delbonis: There's a little more to that than it seems because a "signed block" and a "signed header" should actually have identical signatures, since what we want to be signing is the header.  We'd need some scheme to expose "what is the message we're signing/verifying".

crates/ol/chain-types/src/common.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    pub fn slot(&self) -> Slot {
+        self.slot
+    }
+
+    pub fn blkid(&self) -> &OLBlockId {
+        &self.blkid
+    }
+}
```
delbonis, line 28: I'll approve this PR with this, but this should be removed soon now that a version exists in `identifiers`.

Same for the rest of the types here actually.

delbonis (approved):

PR #1098, opened 2025-10-15
SSZ 1: SPS-62 checkpoint-types

krsnapaudel:
## Description

SSZ support for checkpoint types per SPS-62.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

      in the body of this PR.

## Related Issues

[STR-1910](https://alpenlabs.atlassian.net/browse/STR-1910)

[STR-1910]: https://alpenlabs.atlassian.net/browse/STR-1910?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -630,8 +630,8 @@ dependencies = [
  "serde_json",
  "serde_with",
  "thiserror 2.0.16",
- "tree_hash",
- "tree_hash_derive",
+ "tree_hash 0.10.0",
+ "tree_hash_derive 0.10.0",
 ]
 
 [[package]]
@@ -3778,7 +3778,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8d162beedaa69905488a8da94f5ac3edb4dd4788b732fadb7bd120b2625c1976"
 dependencies = [
  "data-encoding",
- "syn 1.0.109",
+ "syn 2.0.106",
 ]
 
 [[package]]
@@ -12093,7 +12093,7 @@ version = "0.29.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9465315bc9d4566e1724f0fffcbcc446268cb522e60f9a27bcded6b19c108113"
 dependencies = [
- "bitcoin_hashes 0.13.0",
+ "bitcoin_hashes 0.14.0",
  "rand 0.8.5",
  "secp256k1-sys 0.10.1",
  "serde",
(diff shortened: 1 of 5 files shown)


crates/checkpoint-types-ssz/specs/checkpoint_types.ssz
delbonis, line 1: Should we be converting over to using types that [look like this](https://www.notion.so/SPS-62-OL-checkpoint-structure-and-verification-225901ba000f80cfb3cff69d0c916054) and *then* convert to SSZ or vice versa?  Converting over the existing types I think is kinda a wasted effort since we know we'll be changing them.

crates/identifiers-ssz/src/conversions.rs
```diff
@@ -0,0 +1,175 @@
...
+impl From<borsh::L1BlockCommitment> for crate::L1BlockCommitment {
+    fn from(value: borsh::L1BlockCommitment) -> Self {
+        Self {
+            height: value.height_u64(),
+            blkid: FixedVector::from(value.blkid().as_ref().to_vec()),
+        }
+    }
+}
+
+impl From<crate::L1BlockCommitment> for borsh::L1BlockCommitment {
```
delbonis: Only use `crate::` imports in `use` decls.

crates/identifiers-ssz/Cargo.toml
delbonis, line 1: What is the rationale for introducing new crates just for the SSZ serialization?

My thought was that instead of making new crates for the SSZ definitions, we should convert the *existing* crates over to using the SSZ definitions, removing the borshy ones (which we want to eliminate anyways) and manually adding the extra conversion impls where needed.  This way we won't need an additional conversion step before serialization / after deserialization that does what is essentially an extra copy.

delbonis (changes requested):

crates/identifiers-ssz/Cargo.toml
krsnapaudel: I did have an exchange with Claude Code about this. I felt the conversion was not necessary. The original rationale was to avoid introducing breaking changes and allow gradual migration. I agree with your suggestion. So will update the PR soon.

crates/checkpoint-types-ssz/specs/checkpoint_types.ssz
krsnapaudel: I updated the PR to add SSZ to SPS-62 checkpoint types. According to @prajwolrg, the work on SPS-62 types is not done yet. I will create other PRs for types that already exist.

PR #1102, opened 2025-10-16
SSZ 3: snark-acct-types

krsnapaudel:
## Description

SSZ support for `snark-acct-types`. Also requires SSZ support added to `strata-merkle` from `strata-common`. See [PR](https://github.com/alpenlabs/strata-common/pull/36).

The PR was created with help from Claude Code.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

      in the body of this PR.

## Related Issues

[STR-1841](https://alpenlabs.atlassian.net/browse/STR-1841)

[STR-1841]: https://alpenlabs.atlassian.net/browse/STR-1841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1415,7 +1415,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1426,7 +1426,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2564,7 +2564,7 @@ dependencies = [
  "bitflags 2.10.0",
  "cexpr",
  "clang-sys",
- "itertools 0.12.1",
+ "itertools 0.13.0",
  "log",
  "prettyplease",
  "proc-macro2",
@@ -2584,7 +2584,7 @@ dependencies = [
  "bitflags 2.10.0",
  "cexpr",
  "clang-sys",
- "itertools 0.12.1",
+ "itertools 0.13.0",
  "proc-macro2",
  "quote",
  "regex",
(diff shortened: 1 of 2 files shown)


crates/snark-acct-types/src/outputs.rs
```diff
@@ -1,90 +1,102 @@
...
+    pub fn extend_messages<I>(&mut self, iter: I)
+    where
+        I: Iterator<Item = OutputMessage>,
+    {
+        for item in iter {
+            self.messages
+                .push(item)
+                .expect("messages list capacity exceeded");
+        }
+    }
```
delbonis: The arg types can be written as `impl IntoIterator<Item = _>` and then *I think* you can just use `.extend` on the lists, if that's present.

However, I wonder if it's bad API design to just panic here instead of maybe returning a bool and resetting the length?

crates/snark-acct-types/ssz/update.ssz
delbonis, line 1: Please add blank lines before doc comment blocks to improve readability, this is especially relevant in contexts where we don't have syntax highlighting.

delbonis (changes requested):
Generally good, some minor suggestions.

crates/snark-acct-types/src/outputs.rs
krsnapaudel: `.extend` is not there in `VariableList`. I am going to add `try_extend_*`. This does require collect entries from the iterator. The `iter().map()` on caller side and the `collect()` on callee side is not ideal. But good to check capacity I think.

crates/ee-acct-runtime/src/verification_state.rs
```diff
@@ -82,19 +84,43 @@ impl UpdateVerificationState {
...
+            // rather than continue with inconsistent state.
+            eprintln!(
+                "ERROR: Failed to extend messages: {}. Current: {}, Adding: {}, Max: {}",
+                e,
+                self.accumulated_outputs.messages().len(),
+                outputs.output_messages().len(),
+                MAX_MESSAGES
+            );
+            panic!("output messages capacity exceeded");
+        }
```
delbonis, line 123: We should handle this in a more graceful way, like propagating the error up.  But we can do that later.

crates/snark-acct-types/src/outputs.rs
```diff
@@ -1,90 +1,122 @@
...
+        if needed > MAX_MESSAGES as usize {
+            return Err("messages capacity would be exceeded");
+        }
+
+        for item in items {
+            self.messages.push(item).expect("capacity already checked");
+        }
+
+        Ok(())
+    }
```
delbonis, line 88: Well this isn't that useful, if we were going to collect it into a vec anyways we'd just accept a vec, or really a slice.  But actually we'd want to use an `ExactSizeIterator`.  But this is something we can worry about later.

delbonis (approved):

PR #1104, opened 2025-10-17
STR 1772 typed auxiliary input framework

irnb:
## Description

Typed auxiliary input framework for ASM

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

You can refer to the [design document ](https://www.notion.so/ASM-STF-Logs-Auxiliary-Inputs-and-Design-Considerations-293901ba000f80d18938f55942e0e6fc#293901ba000f80bba27ddd3f7fa808bf)for the overall approach to log commitments in `history_mmr` and the `MMR tree` construction. 

## Checklist

      in the body of this PR.

## Related Issues

STR-1772


crates/asm/common/src/tx.rs
```diff
@@ -12,12 +17,18 @@ use strata_l1_txfmt::TagDataRef;
...
     /// Create a new `TxInput` referencing the given `Transaction`.
-    pub fn new(tx: &'t Transaction, tag: TagDataRef<'t>) -> Self {
-        TxInputRef { tx, tag }
+    pub fn new(tx: &'t Transaction, tag: TagDataRef<'t>, index: RequesterL1Index) -> Self {
+        TxInputRef { tx, tag, index }
+    }
+
+    /// create a new `TxInput` referencing the given `Transaction` with index 0 for tests
+    pub fn new_for_test(tx: &'t Transaction, tag: TagDataRef<'t>) -> Self {
+        TxInputRef { tx, tag, index: 0 }
```
prajwolrg: This should not be here.

crates/asm/common/src/aux.rs
```diff
@@ -1,56 +1,66 @@
...
+        self.l1_tx_index
+    }
+
+    /// Tries to merkle-proof validate the aux payload for the given subprotocol.
+    pub fn validate<S: Subprotocol>(&self) -> Result<(), AsmError> {
+        match &self.data {
+            AuxResponseSpec::Single(_resp) => todo!(),
+            AuxResponseSpec::Range(_resps) => todo!(),
+        }
     }
```
prajwolrg: `validate` logic must take some additional argument for validation. The additional argument can depend upon the type of AuxPayload.

crates/asm/aux/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+            cohashes: proof.cohashes().to_vec(),
+        }
+    }
+}
+
+impl SerializableMerkleProof {
+    pub fn into_merkle_proof(self) -> MerkleProof {
+        MerkleProof::from_cohashes(self.cohashes, self.index)
+    }
+}
```
prajwolrg: I think we can avoid creating a new type and add proper serialization directly to the type.

crates/asm/aux/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+pub struct HistoricalAuxResponse {
+    pub log: AuxLog,
+    pub proof: SerializableMerkleProof,
+}
+
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize)]
+pub struct HistoricalAuxRangeResponse {
+    pub logs: Vec<AuxLog>,
+    pub proofs: Vec<SerializableMerkleProof>,
+}
```
prajwolrg: This looks wrong. This seems to only provide the InclusionProof of the logs. But we also need to validate the completeness of the logs.

For the checkpoint subprotocol, we need to validate that _**all**_ the logs were applied in the right order. This approach doesn't seem to encapsulate that.

crates/asm/common/src/aux.rs
```diff
@@ -1,56 +1,66 @@
...
         &self.data
     }
+
+    pub fn requester_index(&self) -> RequesterL1Index {
+        self.l1_tx_index
+    }
+
+    pub fn encode(&self) -> Vec<u8> {
+        borsh::to_vec(&self.data).expect("asm: serialize aux request")
+    }
```
prajwolrg: If we need `encode`, we must also need `decode`.

crates/asm/stf/src/manager.rs
```diff
@@ -65,23 +66,22 @@ impl<S: Subprotocol, R: MsgRelayer, C: AuxInputCollector> SubprotoHandler for Ha
...
 
-        // TODO better error handling
-        let parsed_aux = <S::AuxInput as BorshDeserialize>::try_from_slice(aux_input_data)
-            .map_err(|e| AsmError::Deserialization(S::ID, e))
-            .unwrap();
-
+        for aux in aux_input_data {
+            aux.validate::<S>()
+                .expect("asm: invalid aux payload for subprotocol");
+        }
```
prajwolrg: This feels weird. The manager should not do any additional validation except for the type validations.

crates/asm/common/Cargo.toml
```diff
@@ -10,6 +10,9 @@ workspace = true
 bitcoin.workspace = true
 borsh.workspace = true
 serde.workspace = true
+strata-acct-types.workspace = true
+strata-asm-aux.workspace = true
+strata-asm-logs.workspace = true
```
prajwolrg: The initial design was for `strata-asm-common` to be unaware of the specific logs or the auxiliary inputs.

prajwolrg (changes requested):

crates/asm/common/src/subprotocol.rs
```diff
@@ -122,20 +115,21 @@ pub trait Subprotocol: 'static {
...
-    /// * `aux_inputs` - Auxiliary data previously requested and validated
+    /// * `aux_resolver` - Interface for retrieving auxiliary data previously requested
     /// * `relayer` - Interface for sending messages to other subprotocols and emitting logs
     /// * `params` - Subprotocol's current params
     fn process_txs(
         state: &mut Self::State,
         txs: &[TxInputRef<'_>],
         anchor_pre: &AnchorState,
-        aux_input: &Self::AuxInput,
+        aux_resolver: &dyn AuxResolver,
```
delbonis: We shouldn't be using `dyn` unless we absolutely have to.

If we're giving up on making aux inputs generic and saying "they have to be one of these types" (defined in the enum), then we might as well just pass a vec of them here instead.

crates/asm/common/src/subprotocol.rs
```diff
@@ -122,20 +115,21 @@ pub trait Subprotocol: 'static {
     ///
```
delbonis, line 116: ^Above here, we should make `pre_process_txs` blocked behind a feature (`preprocess`?) so that it doesn't make it into the proof binaries.

crates/asm/common/src/aux.rs
```diff
@@ -1,56 +1,136 @@
...
+
+/// Provides access to pre-fetched auxiliary responses during transaction execution.
+pub trait AuxResolver {
+    /// Returns all verified historical ASM logs associated with a given L1 transaction.
+    fn historical_logs(&self, tx_index: L1TxIndex) -> AuxResolveResult<Vec<AsmLogEntry>>;
+
+    /// Returns the deposit request transaction referenced by a bridge deposit, if present.
+    // TODO: consider changing return type based on bridge subprotocol needs
+    fn deposit_request_tx(&self, tx_index: L1TxIndex) -> AuxResolveResult<Option<Vec<u8>>>;
+}
```
delbonis: Similarly, if we're saying that "the aux has to be one of these types" then this adds indirection and we can just have an enum for the entries that would be.

delbonis (changes requested):
If the more generic approach that I was sketching out previously turned out to be more trouble than it's worth then we should just go all the way with it and define upfront the types of queries that can be made so that we have a more structured interface.

We can develop this more and upstream it into the main ASM spec.

I mentioned using an enum for this in the comments, but alternatively we could consider a little more restricted approach that's easier to deal with.

```rs
struct AuxRequest {
  asm_log_queries: Vec<AuxLogQuery>,
  header_queries: Vec<HeaderQuery>
}

struct AuxData {
  asm_log_oracles: Vec<AuxLogOracleData>,
  header_oracles: Vec<HeaderOracleData>
}
```

Making everything be an enum provided as a vec makes it a little more loose (ie. "does the interleaving order have meaning?", although it shouldn't).

This doesn't have as demanding encoding stability requirements since it'd be a private input anyways.

crates/asm/common/src/subprotocol.rs
prajwolrg: We can also feature flag it behind the `!target_os=zkvm`, so it not included for ZkVms. But having preprocess feature might be a better approach.

crates/asm/common/src/aux.rs
```diff
@@ -1,56 +1,136 @@
...
+    )]
+    InvalidLogProof {
+        /// Subprotocol identifier.
+        subprotocol: SubprotocolId,
+        /// L1 transaction index within the block.
+        tx_index: L1TxIndex,
+        /// Hash of the L1 block whose logs were being proven.
+        block_hash: BlockHash,
+    },
+}
```
prajwolrg: let's also include the tx_type for more helpful information.

crates/asm/common/src/log.rs
```diff
@@ -106,3 +108,13 @@ impl AsmLogEntry {
...
+
+/// Computes the log-leaf hash committed into the header MMR for a given block.
+pub fn compute_log_leaf(block_hash: &BlockHash, logs: &[AsmLogEntry]) -> [u8; 32] {
+    let logs_hash = compute_borsh_hash(&logs);
+    let mut payload = Vec::with_capacity(64);
+    payload.extend_from_slice(&block_hash.to_byte_array());
+    payload.extend_from_slice(logs_hash.as_ref());
+    let leaf: [u8; 32] = sha256d(&payload).into();
+    leaf
+}
```
prajwolrg: We don't require using `sha256d` here. It just increases the proving cost.

crates/asm/aux/src/response.rs
```diff
@@ -0,0 +1,65 @@
...
+        let index = u64::deserialize_reader(reader)?;
+        let cohashes = Vec::<[u8; 32]>::deserialize_reader(reader)?;
+        let proof = LogMmrProof::from_cohashes(cohashes, index);
+        Ok(Self {
+            block_hash,
+            logs,
+            proof,
+        })
+    }
+}
```
prajwolrg: Should we derive this manually? I think it should be possible to use derive directly. We define `L1BlockId` in btc-types. We should be able to use that.

crates/asm/common/src/aux.rs
```diff
@@ -1,56 +1,87 @@
...
+        let cohashes = Vec::<[u8; 32]>::deserialize_reader(reader)?;
+        let index = u64::deserialize_reader(reader)?;
+        let proof = LogMmrProof::from_cohashes(cohashes, index);
+        Ok(Self {
+            block_hash: BlockHash::from_byte_array(block_hash_bytes),
+            logs,
+            proof,
+        })
     }
+}
```
prajwolrg: I think we should be able to derive BorshSerialize/Deserialize here. We can also use `L1BlockId` that is defined in the `strata-identifiers` that implements the borsh serialization deserialization.

crates/asm/aux/src/request.rs
```diff
@@ -0,0 +1,31 @@
...
+
+use bitcoin::Txid;
+use strata_asm_common::{L1TxIndex, SubprotocolId};
+
+/// Table mapping subprotocol IDs to their corresponding auxiliary requests.
+pub type AuxRequestTable = BTreeMap<SubprotocolId, AuxRequestEnvelope>;
+#[derive(Debug)]
+pub struct AuxRequestEnvelope {
+    pub asm_log_queries: Vec<AuxLogQuery>,
+    pub drt_tx_queries: Vec<DrtTxQuery>,
```
prajwolrg: For the spec, it makes sense to make it more generic. Instead of requesting DRT Transaction, the spec should support requesting a transaction in general.

crates/asm/common/src/aux.rs
prajwolrg: In case of custom implementation, we should add tests to verify that it works.

crates/asm/common/src/tx.rs
```diff
@@ -12,12 +14,18 @@ use strata_l1_txfmt::TagDataRef;
 pub struct TxInputRef<'t> {
     tx: &'t Transaction,
     tag: TagDataRef<'t>,
+    /// Index of the transaction within the containing L1 block.
+    l1_tx_index: L1TxIndex,
```
prajwolrg: nit: all the fields here are related with l1_tx. It doesn't make much sense to add the prefix.

crates/asm/aux/src/utils.rs
```diff
@@ -0,0 +1,33 @@
...
+        let leaf = compute_log_leaf(&block_logs.block_hash, &block_logs.logs);
+        if !mmr.verify(&block_logs.proof, &leaf) {
+            return Err(format!(
+                "Invalid log proof for subprotocol {:?}, tx_index {}, block_hash {}",
+                subprotocol, tx_index, block_logs.block_hash
+            ));
+        }
+    }
+    Ok(())
+}
```
prajwolrg: The verification logic will be different for different type of AuxInput. This assumes a single verification for all AuxInput which is incorrect. We should define the verification logic where that is defined and add tests to verify that the verification works.

crates/util/mmr/src/lib.rs
```diff
@@ -29,6 +30,35 @@ pub struct CompactMmr<H: MerkleHash> {
...
+    {
+        let (entries, cap_log2, roots) = <(u64, u8, Vec<H>)>::deserialize(deserializer)?;
+        Ok(Self {
+            entries,
+            cap_log2,
+            roots,
+        })
+    }
+}
+
```
prajwolrg: In case of custom serialization/deserialization logic, we need to add tests that it works.

prajwolrg (changes requested):
This is starting to take a good shape. We still need to add more tests to make sure that the aux data is correct.

crates/asm/aux/src/request.rs
delbonis: ^ this is very important, at this level of abstraction we shouldn't be designing for particular protocols, talking about "DRTs" makes this primitive not useful for anything but the particular kind of bridge protocol design that uses that format of DRT

crates/util/mmr/src/lib.rs
delbonis: Might be better to just go and add the serialize/deserialize impls to `CompactMmrN` anyways.

crates/asm/common/src/state.rs
```diff
@@ -36,7 +41,26 @@ pub struct ChainViewState {
...
+}
+
+impl ChainViewState {
+    /// Log2 of the number of peaks retained in the header log MMR.
+    pub const HEADER_MMR_CAP_LOG2: usize = 64;
+
+    /// Creates an empty compact MMR ready to accept the first block leaf.
+    pub fn empty_history_mmr() -> HistoryMmrCompact {
+        HistoryMmr::new(Self::HEADER_MMR_CAP_LOG2).to_compact()
+    }
```
delbonis: This function shouldn't be on `ChainViewState`.

crates/asm/common/src/state.rs
```diff
@@ -1,6 +1,11 @@
 use borsh::{BorshDeserialize, BorshSerialize};
 use serde::{Deserialize, Serialize};
 use strata_asm_types::HeaderVerificationState;
+use strata_mmr::{CompactMmr, MerkleMr64, Sha256Hasher};
+
+pub type HistoryMmrHash = [u8; 32];
+pub type HistoryMmr = MerkleMr64<Sha256Hasher>;
+pub type HistoryMmrCompact = CompactMmr<HistoryMmrHash>;
```
delbonis: A very similar definition exists in `acct-types`.  We could move these out to a new crate somewhere so we only have to define it once.

crates/asm/common/src/aux.rs
```diff
@@ -1,56 +1,87 @@
...
+#[derive(Debug, Clone, BorshDeserialize, BorshSerialize)]
+pub struct AuxData {
+    pub asm_log_oracles: Vec<BlockLogsOracleData>,
+    pub drt_tx_oracles: Vec<DrtTxOracleData>,
+}
+
+#[derive(Debug, Default, Clone, BorshDeserialize, BorshSerialize)]
+pub struct AuxInput {
+    pub data: BTreeMap<L1TxIndex, AuxData>,
 }
```
delbonis, line 154: How would we pass aux data if there wasn't one particular tx that the request pertained to?  For example, if it was aggregated from multiple txs or the requests were from on from data in the subprotos state.

Really we would just want to pass the `AuxData` as the whole struct and have the subproto impl figure out what it needs from that instead of enforcing a structure binding the data entries to a tx.

crates/asm/common/src/aux.rs
```diff
@@ -1,56 +1,87 @@
...
-    /// Processed auxiliary input data as raw bytes.
-    ///
-    /// This `Vec<u8>` must deserialize into one
-    /// `<P as Subprotocol>::AuxInput` for the corresponding subprotocol P.
-    pub data: Vec<u8>,
+/// Per-block set of historical logs together with an MMR inclusion proof.
+#[derive(Debug, Clone)]
+pub struct BlockLogsOracleData {
+    pub block_hash: BlockHash,
+    pub logs: Vec<AsmLogEntry>,
```
delbonis: These fields should just be an `AsmManifest` (or whatever it is that gets put into the history mmr) instead of flattening the fields out into this container.

crates/asm/common/src/aux.rs
```diff
@@ -1,56 +1,87 @@
...
+/// Responsible for recording auxiliary requests emitted during preprocessing.
+pub trait AuxRequestCollector: Any {
+    /// Records that the transaction at `tx_index` requires auxiliary data described by `request`.
+    fn request_aux_input(&mut self, tx_index: L1TxIndex, request: AuxRequestSpec);
 
-    pub fn data(&self) -> &[u8] {
-        &self.data
-    }
+    /// Exposes the collector as a `&mut dyn Any` for downcasting.
+    fn as_mut_any(&mut self) -> &mut dyn Any;
```
delbonis: Definitely just replace this with a container for individual requests.  It should be like the converse to the `AuxData` (just not bound to txs, the only thing we need binding to would be the subprotocol ID).  It's just adding indirection like this, we don't get anything out of it being generic.

crates/asm/common/src/subprotocol.rs
```diff
@@ -197,7 +193,7 @@ pub trait SubprotoHandler {
         txs: &[TxInputRef<'_>],
         relayer: &mut dyn MsgRelayer,
         anchor_state: &AnchorState,
-        aux_input_data: &[u8],
+        aux_input: &AuxInput,
```
delbonis: Maybe pass by value to reduce clones?

crates/asm/common/src/aux.rs
irnb: We actually need this feature to properly handle multiple Aux requests that can appear within a single block.

> For example:
> The subprotocol defines two transaction types: Type1 and Type2.
> Type1 transactions require AuxInput, while Type2 transactions do not.
> Suppose an L1Block contains the following sequence: Type1Tx, Type2Tx, Type1Tx, Type1Tx.

If we don’t know which transaction each piece of Aux data corresponds to, we can’t reliably access the correct data. That’s why having the requesting transaction index can be helpful, it allows the subprotocol developer to easily retrieve the correct response needed for processing a specific transaction type.

If we were to instead embed this information(requester tx index) inside the Aux data itself, the subprotocol would have to search through multiple entries to find the expected response, which is inefficient.
By maintaining this mapping, we effectively provide an indirect link between requests and their responses, making data access simpler and more predictable for subprotocol developer.

That said, I’m happy to hear other ideas or alternative approaches if you think there’s a cleaner way to handle this. 🙏

crates/asm/common/src/aux.rs
irnb: In `crates/asm/common/src/log.rs`, the MMR leaf is defined as:

```rust
/// AsmLogsHash = Hash(Vec<AsmLog>)
/// MMRLeafHash = Hash(L1BlockHash, AsmLogsHash)
pub fn compute_logs_leaf(block_hash: &L1BlockId, logs: &[AsmLogEntry]) -> [u8; 32] {
    let logs_hash = compute_borsh_hash(&logs);
    let mut payload = Vec::with_capacity(64);
    payload.extend_from_slice(block_hash.as_ref());
    payload.extend_from_slice(logs_hash.as_ref());
    let leaf: [u8; 32] = sha256(&payload).into();
    leaf
}
```

This is exactly what we commit to the MMR: the **L1 block hash** and the **hash of all ASM-emitted logs** for that block’s processing.

By contrast, `AsmManifest` includes additional fields and carries a header as raw bytes rather than the block hash:

```rust
// crates/asm/chain-types/src/manifest.rs
pub struct AsmManifest {
    /// Encoded ASM logs as raw bytes
    pub encoded_logs: Vec<u8>,

    /// Encoded header as raw bytes (not decoded bitcoin::Header)
    pub encoded_header: Vec<u8>,

    /// Additional encoded data from ASM execution
    pub encoded_data: Vec<u8>,
    // TODO: Add other fields as needed based on actual ASM execution output
}
```

If we were to replace the AsmLogsOracleData fields with `AsmManifest`, verification would incur extra work: we’d need to deserialize the header to reconstruct the **block hash** and parse the logs to recompute the **logs hash**. In other words, the manifest introduces an additional verification cycle compared to committing directly to `(block_hash, logs_hash)` in the AsmLogsOracleData.

We can handle this type conversion within `asm/worker`.
Please let me know if I’m misunderstanding anything here.

crates/asm/common/src/subprotocol.rs
irnb: Subprotocol  just read from `AuxInput`
If we pass by value, we'd need to clone at the call site in `stage.rs:106` to move the value out of the `BTreeMap`, which would increase clones rather than reduce them

crates/asm/common/src/aux.rs
delbonis: Can you expand on how the different types would break it?

Perhaps it would be better to have a procedure that takes a list of oracle data, verifies the proof, and then generates something like this:

```
struct AuxAsdfIdk {
  asm_manifest_queries: BTreeMap<Height, AsmManifest>
  // whatever else
}
```

That we could query directly.

I'm opposed to the txid coupling because, as I mentioned above, it enforces by definition that aux entries correspond exactly 1:1 to specific txs and doesn't allow the request to be generated through other sources.  Think like in an EVM context, trying to enforce a structure so that each tx gets just the storage slots it needs and no others.  It complicates the bookkeeping because there's shared state across txs and the same storage slots are often used across multiple txs.

But I guess this is an implementation detail and doesn't appear in the proof interface, so it's not a horribly breaking change if we wanted to change it.  So we *could* just ship it without the ability to make statements like the above.

crates/asm/common/src/aux.rs
delbonis: > In crates/asm/common/src/log.rs, the MMR leaf is defined as:

I am not sure why it was defined this way. It *should* be a SSZ hash tree root, although we don't have that right now.  As a stopgap, we should just do a flat hash of the borsh-encoded manifest.

> By contrast, AsmManifest includes additional fields and carries a header as raw bytes rather than the block hash:

Ah I'm sorry, I wasn't aware it was defined this way there.  This is what the def *should* be, which is closer to how I've defined it elsewhere:

```rs
struct AsmManifest {
  encoded_header: Vec<u8>, // or [u8; 80] I guess?
  wtxids_root: [u8; 32],
  logs: Vec<AsmLog>,
}

struct AsmLog(Vec<u8>);
```

I'm not sure what `encoded_data` represents in the above def.  The ASM state maybe?  Not sure.

> we’d need to deserialize the header to reconstruct the block hash

We *can* just hash it directly, we know it's valid because that's how the accumulator is constructed, so it should be fine.

crates/asm/common/src/log.rs
```diff
@@ -31,7 +35,7 @@ pub trait AsmLog: BorshSerialize + BorshDeserialize {
 /// [`AsmLogEntry::from_msg`], and retrieve typed data using [`AsmLogEntry::try_into_log`]
 /// or check if it's a valid SPS-52 message using [`AsmLogEntry::try_as_msg`].
 /// TODO(QQ): FIX borsh here.
```
prajwolrg, line 33: @evgenyzdanovich is this message still relevant here?

PR #1107, opened 2025-10-27
Str 1413 sync and fcm refactor

purusang:
## Description

Refactored ForkChoiceManager (FCM) to be shareable across threads and removed duplicate block tracker from sync crate. I still need some critical feedback on where fcm should be initialized, design choices or anything felt by reviewers. Thanks!

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Changes

1. FCM now returns `Arc<RwLock<ForkChoiceManager>>` instead of plain `ForkChoiceManager`
2. Moved FCM initialization to `init_fcm_after_genesis()` in fork_choice_manager.rs
3. Added wrapper methods: `get_chain_tracker_finalized_epoch()`, `is_chain_tracker_block_seen()`
4. L2SyncState now only holds FCM reference and queries through it
5. Sync worker uses immutable state reference since no local mutations needed
  
## Checklist

      in the body of this PR.

## Related Issues

This work is based on discussions with @sapinb and @irnb to take fcm out of `tracker_task` which started from ticket 1413.

### DIFF
--- a/crates/sync/src/worker.rs
+++ b/crates/sync/src/worker.rs
@@ -239,6 +239,7 @@ async fn handle_new_block<T: SyncClient>(
 
     // send ForkChoiceMessage::NewBlock for all pending blocks in correct order
     // The FCM will handle updating the chain_tracker
+    let mut last_block_id = None;
     while let Some(block) = fetched_blocks.pop() {
         context
             .storage
@@ -246,13 +247,38 @@ async fn handle_new_block<T: SyncClient>(
             .put_block_data_async(block.clone())
             .await?;
         let block_idx = block.header().slot();
+        let block_id = block.header().get_blockid();
+        last_block_id = Some(block_id);
         debug!(%block_idx, "l2 sync: sending chain tip msg");
         context
             .sync_manager
-            .submit_chain_tip_msg_async(ForkChoiceMessage::NewBlock(block.header().get_blockid()))
+            .submit_chain_tip_msg_async(ForkChoiceMessage::NewBlock(block_id))
             .await;
         debug!(%block_idx, "l2 sync: sending chain tip sent");
     }
+
+    // Wait for FCM to process the messages by checking if the chain tracker has the block
+    // This prevents a race condition where we query state before FCM processes our messages
+    if let Some(expected_block_id) = last_block_id {
+        let mut attempts = 0;
+        const MAX_ATTEMPTS: u32 = 100; // 10 seconds max wait
+        while attempts < MAX_ATTEMPTS {
+            if state.has_block(&expected_block_id) {
+                debug!(?expected_block_id, "fcm has processed blocks");
+                break;
+            }
+            // Short delay to give FCM time to process
+            tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
+            attempts += 1;
+        }
+        if attempts >= MAX_ATTEMPTS {
+            warn!(
+                ?expected_block_id,
+                "fcm did not process block after max attempts"
+            );
+        }
+    }
+
     Ok(())
 }
 


crates/consensus-logic/src/fork_choice_manager.rs
```diff
@@ -261,7 +274,7 @@ pub fn init_forkchoice_manager(
     csm_finalized_epoch: Option<EpochCommitment>,
     genesis_blkid: L2BlockId,
     chain_worker: Arc<ChainWorkerHandle>,
-) -> anyhow::Result<ForkChoiceManager> {
+) -> anyhow::Result<Arc<RwLock<ForkChoiceManager>>> {
```
storopoli, line 277: Just a sanity check. Why are you using `RwLock`? Are you expecting much more reads than writes? Or in other words, are you expecting having multiple shared references most of the time and very few exclusive references?

crates/consensus-logic/src/fork_choice_manager.rs
```diff
@@ -335,6 +348,63 @@ pub fn init_forkchoice_manager(
...
+        storage
+            .l2()
+            .get_blocks_at_height_blocking(0)
+            .unwrap()
+            .first()
+            .cloned()
+            .expect("genesis should be in")
+    });
+
+    let (_, init_state) = storage.client_state().fetch_most_recent_state()?.unwrap();
```
krsnapaudel: Similar comment here for `unwrap()`.

crates/consensus-logic/src/fork_choice_manager.rs
```diff
@@ -335,6 +348,63 @@ pub fn init_forkchoice_manager(
...
+    params: &Arc<Params>,
+    chain_worker: Arc<ChainWorkerHandle>,
+    status_channel: &StatusChannel,
+) -> anyhow::Result<Arc<RwLock<ForkChoiceManager>>> {
+    // Wait for genesis block
+    info!("waiting for genesis before starting forkchoice logic");
+    let genesis_block_id = handle.block_on(async {
+        while storage
+            .l2()
+            .get_blocks_at_height_blocking(0)
```
krsnapaudel: This could cause a panic, right? Claude Code suggests something like this:
```
let genesis_block_id = handle.block_on(async {
    loop {
        match storage.l2().get_blocks_at_height_blocking(0) {
            Ok(blocks) if !blocks.is_empty() => {
                return blocks[0];  // Safe indexing
            }
            Ok(_) => {
                debug!("genesis not yet available, waiting...");
                tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
            }
            Err(e) => {
                warn!(err = %e, "error checking for genesis, retrying...");
                tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
            }
        }
    }
})?;
```
Same for other unwrap()s and expect()s. Probably better to handle the errors than panic.

crates/sync/src/state.rs
```diff
@@ -1,96 +1,44 @@
...
-use tracing::debug;
+use std::sync::{Arc, RwLock};
 
-use crate::L2SyncError;
+use strata_consensus_logic::fork_choice_manager::ForkChoiceManager;
+use strata_ol_chain_types::L2BlockId;
+use strata_primitives::epoch::EpochCommitment;
 
-#[derive(Debug)]
+#[derive(Clone)]
```
krsnapaudel: Lost `Debug`. Is this intentional?

krsnapaudel (changes requested):

crates/consensus-logic/src/fork_choice_manager.rs
purusang: Up to this point there will be only 2 threads sharing this and yes more reads than writes. In future it can be shared any other worker thread that might need fcm.

crates/sync/src/state.rs
purusang: It was intentional as fcm does not implement Debug. We do not need Clone as well so let me remove it.

crates/consensus-logic/src/fork_choice_manager.rs
storopoli: Good point. One quick nitpick I have this in my memories/rules (whatever your LLM thing uses):

> When writing Rust code, don't inline your imports like `tokio::time::sleep(std::time::Duration::from_secs(1))` but import them at the top of the file using `use` statements.

I do have these rules/memories for all my quirks in my development environment: `gh` CLI, conventional commit messages, Nix, Fish shell, and Jujutsu version control.

PR #1111, opened 2025-10-29
STR 1801 BlockAssemblyContext interface

krsnapaudel:
## Description

Defines the `BlockAssemblyContext` trait API and mempool types for OL block assembly.

### What Block Assembly Needs
Get parent state → determine current slot/epoch and L1 view
Prepare L1 update → fetch ASM logs, scan for checkpoint, create manifests
Prepare transactions → get from mempool, validate, build tx segment
Compute slot/epoch → increment based on parent state
Assemble body → combine tx segment + L1 update
Execute block → state transition
Create header → commitment roots + new state root

**The Context API and purpose**
* `get_latest_l1_block()`: L1 bounds for fetching ASM logs
* `get_asm_logs_range(from, to)`: ASM logs for manifests
* `get_mempool_transactions(limit)`: Pending transactions
*  `clear_mempool_transactions(range)`: Remove included transactions

### Design Decisions
Range-based ASM logs fetching
`fn get_asm_logs_range(from_block: L1BlockCommitment, to_block: L1BlockCommitment)`
Uses range [from, to] instead of (from, count). Also not fetching ASM states. Block assembly only needs logs to scan for `CheckpointUpdate` and create `AsmManifest`.

Three error types
`Database` for errors from storage
`MempoolError` for mempool operations with domain-specific variants
`InvalidRangeError`: `from_block` height  > `to_block` height

For `OLState`, we could use `StateAccessor` abstraction. So not included in context.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Not sure if we need additional methods or types.

## Checklist

      in the body of this PR.

## Related Issues

[STR-1801](https://alpenlabs.atlassian.net/browse/STR-1801)

[STR-1801]: https://alpenlabs.atlassian.net/browse/STR-1801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/ol/block-assembly/Cargo.toml
+++ b/crates/ol/block-assembly/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "strata-ol-block-assembly"
+version = "0.1.0"
+edition = "2024"
+
+[lints]
+workspace = true
+
+[dependencies]
+strata-asm-common.workspace = true
+strata-db.workspace = true
+strata-identifiers.workspace = true
+strata-ol-chain-types-new.workspace = true
+strata-ol-mempool.workspace = true
+strata-primitives.workspace = true
+
+thiserror.workspace = true
--- a/crates/ol/block-assembly/src/context.rs
+++ b/crates/ol/block-assembly/src/context.rs
@@ -1,0 +1,76 @@
+//! Block assembly context trait.
+
+use strata_asm_common::AsmLogEntry;
+use strata_identifiers::OLTxId;
+use strata_ol_chain_types_new::OLTransaction;
+use strata_primitives::l1::L1BlockCommitment;
+
+use crate::error::BlockAssemblyError;
+
+/// Context interface for block assembly.
+///
+/// Provides access to external resources needed for OL block assembly:
+/// - ASM logs for L1 updates (checkpoints, deposits)
+/// - Mempool for pending transactions
+///
+/// # Design Notes
+///
+/// This trait provides access to external data sources needed during block assembly:
+/// - **ASM logs**: Read-only access to validated L1 data (checkpoints, deposits)
+/// - **Mempool**: Read and write access to pending transactions
+///
+/// The context provides access to external resources
+/// without owning or mutating them directly.
+pub trait BlockAssemblyContext {
+    /// Gets ASM's latest processed L1 block commitment (tip).
+    ///
+    /// Returns `Ok(None)` if no L1 blocks processed yet, `Err` with [`BlockAssemblyError`] if error
+    /// occurs.
+    fn get_latest_l1_block(&self) -> Result<Option<L1BlockCommitment>, BlockAssemblyError>;
+
+    /// Fetches ASM logs in the [`L1BlockCommitment`] range `[from_block, to_block]` (inclusive).
+    ///
+    /// Returns entries in ascending order by L1 block height.
+    /// Logs are used for scanning checkpoints and creating ASM manifest entries.
+    /// Returns empty [`Vec`] if no logs available in range.
+    ///
+    /// # Errors
+    ///
+    /// - [`BlockAssemblyError::InvalidRange`] - if `from_block` height > `to_block` height
+    /// - [`BlockAssemblyError::Database`] - if database error occurs
+    fn get_asm_logs_range(
+        &self,
+        from_block: L1BlockCommitment,
+        to_block: L1BlockCommitment,
+    ) -> Result<Vec<(L1BlockCommitment, Vec<AsmLogEntry>)>, BlockAssemblyError>;
+
+    /// Gets pending [`OLTransaction`] entries from mempool (up to `limit`).
+    ///
+    /// Returns transactions with their [`OLTxId`] values.
+    /// Returns empty [`Vec`] if no transactions available (not an error).
+    ///
+    /// # Errors
+    ///
+    /// - [`BlockAssemblyError::Mempool`] - if mempool error occurs
+    fn get_mempool_transactions(
+        &self,
+        limit: u64,
+    ) -> Result<Vec<(OLTxId, OLTransaction)>, BlockAssemblyError>;
+
+    /// Removes included transactions from mempool.
+    ///
+    /// Attempts to clear each transaction with the given [`OLTxId`].
+    /// Returns the list of transaction IDs that were successfully removed.
+    /// Returns empty [`Vec`] if no transactions were removed.
+    ///
+    /// This operation is idempotent - already-removed transactions will simply
+    /// not appear in the returned list.
+    ///
+    /// # Errors
+    ///
+    /// - [`BlockAssemblyError::Mempool`] - if mempool error occurs (connectivity, internal error)
+    fn clear_mempool_transactions(
+        &self,
+        txids: &[OLTxId],
+    ) -> Result<Vec<OLTxId>, BlockAssemblyError>;
+}
--- a/crates/ol/block-assembly/src/error.rs
+++ b/crates/ol/block-assembly/src/error.rs
@@ -1,0 +1,20 @@
+//! Error types for block assembly operations.
+
+use strata_db::errors::DbError;
+use strata_ol_mempool::MempoolError;
+
+/// Errors that can occur during block assembly operations.
+#[derive(Debug, thiserror::Error)]
+pub enum BlockAssemblyError {
+    /// Database operation failed.
+    #[error("db: {0}")]
+    Database(#[from] DbError),
+
+    /// Mempool operation failed.
+    #[error("mempool: {0}")]
+    Mempool(#[from] MempoolError),
+
+    /// Invalid L1 block range where `from_block` height > `to_block` height.
+    #[error("invalid L1 block height range (from {from_height} to {to_height})")]
+    InvalidRange { from_height: u64, to_height: u64 },
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14081,6 +14081,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "strata-ol-block-assembly"
+version = "0.1.0"
+dependencies = [
+ "strata-asm-common",
+ "strata-db",
+ "strata-identifiers",
+ "strata-ol-chain-types-new",
+ "strata-ol-mempool",
+ "strata-primitives",
+ "thiserror 2.0.16",
+]
+
+[[package]]
 name = "strata-ol-chain-types"
 version = "0.3.0-alpha.1"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ members = [
   "crates/l1tx",
   "crates/ledger-types",
   "crates/mpt",
+  "crates/ol/block-assembly",
   "crates/ol/chain-types",
   "crates/ol/mempool",
   "crates/ol/state-types",
@@ -183,6 +184,7 @@ strata-l1tx = { path = "crates/l1tx" }
 strata-ledger-types = { path = "crates/ledger-types" }
 strata-mmr = { path = "crates/util/mmr" }
 strata-mpt = { path = "crates/mpt" }
+strata-ol-block-assembly = { path = "crates/ol/block-assembly" }
 strata-ol-chain-types = { path = "crates/ol-chain-types" }
 strata-ol-chain-types-new = { path = "crates/ol/chain-types" }
 strata-ol-chainstate-types = { path = "crates/ol-chainstate-types" }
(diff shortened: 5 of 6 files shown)


prajwolrg:
This might overlap with some work done by @evgenyzdanovich as part of https://github.com/alpenlabs/alpen/pull/1078. ASM Logs are made available by the [`AsmStateManager`](https://github.com/alpenlabs/alpen/blob/main/crates/storage/src/managers/asm.rs).

krsnapaudel:
> This might overlap with some work done by @evgenyzdanovich as part of #1078. ASM Logs are made available by the [`AsmStateManager`](https://github.com/alpenlabs/alpen/blob/main/crates/storage/src/managers/asm.rs).

I guess the implementation of this would use [`AsmStateManager`](https://github.com/alpenlabs/alpen/blob/main/crates/storage/src/managers/asm.rs) then.

crates/csm-types/src/provider.rs
```diff
@@ -0,0 +1,66 @@
...
+    InvalidRange(u64, u64),
+}
+
+/// Provider interface for accessing CSM state, ASM logs, and L1 information.
+pub trait CsmProvider {
+    /// ASM log entry type. Implementations should use `strata_asm_common::AsmLogEntry`.
+    ///
+    /// This is defined as an associated type to avoid requiring `strata-asm-common`
+    /// as a direct dependency in this crate (keeps `csm-types` lightweight).
+    type AsmLogEntry;
```
delbonis: It makes sense for this to depend on `strata-asm-common`, this shouldn't be necessary.

crates/ol/block-assembly/src/context.rs
```diff
@@ -0,0 +1,54 @@
...
+    ) -> Result<Vec<(L1BlockCommitment, Vec<AsmLogEntry>)>, DbError>;
+
+    /// Gets pending [`OLTransaction`] entries from mempool (up to `limit`).
+    ///
+    /// Returns transactions with their [`MempoolIndex`] values.
+    /// Returns empty [`Vec`] if no transactions available (not an error).
+    fn get_mempool_transactions(
+        &self,
+        limit: u64,
+    ) -> Result<Vec<(MempoolIndex, OLTransaction)>, MempoolError>;
```
bewakes: Do you think it will be safer to use `TxId` instead of `MempoolIndex`? Especially while clearing the mempool txs. Just in case somehow the txs get re-ordered(although we won't probably have fees and ordering right now).

crates/ol/mempool/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+    DuplicateTransaction,
+
+    /// Transaction failed validation (for add_transaction).
+    #[error("Transaction validation failed: {0}")]
+    InvalidTransaction(String),
+}
+
+/// Mempool interface for block assembly.
+///
+/// Block assembly uses this trait to fetch pending transactions and clear
```
bewakes: Block assembly doesn't need to use this because the relevant methods are already defined in `BlockAssemblyContext`.
Also, this ticket might be useful: https://alpenlabs.atlassian.net/browse/STR-1803

crates/ol/block-assembly/src/context.rs
```diff
@@ -0,0 +1,66 @@
...
+    ///
+    /// Range is `(start_index, end_index)` where both bounds are [`MempoolIndex`] values,
+    /// inclusive. Either clears all transactions in range or clears none (atomic operation).
+    ///
+    /// # Errors
+    ///
+    /// - [`BlockAssemblyError::Mempool`] - if mempool error occurs (e.g., transaction not found)
+    fn clear_mempool_transactions(
+        &self,
+        range: (MempoolIndex, MempoolIndex),
```
bewakes: Do you think using `TxId` instead of `MempoolIndex` will be safer because same index might refer to a different transaction at a later time? Although we won't possibly have txs ordering right now.

crates/ol/block-assembly/src/context.rs
```diff
@@ -0,0 +1,66 @@
...
+use strata_ol_chain_types_new::{OLBlockId, OLTransaction};
+use strata_ol_mempool::MempoolIndex;
+use strata_ol_state_types::OLState;
+use strata_primitives::l1::L1BlockCommitment;
+
+use crate::error::BlockAssemblyError;
+
+/// Context interface for block assembly.
+pub trait BlockAssemblyContext {
+    /// Gets parent block's [`OLState`].
```
bewakes:
```suggestion
    /// Gets block's [`OLState`] corresponding to a block id.
```
Because the function doesn't say anything about the parent.

bewakes (commented):
Looks good so far. Some comments.

crates/ol/block-assembly/src/context.rs
krsnapaudel: Yes if it's removing using an array, not a range.

crates/ol/mempool/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+use strata_ol_chain_types_new::OLTransaction;
+use strata_primitives::buf::Buf32;
+
+/// Transaction identifier.
+///
+/// Uniquely identifies a transaction in the mempool. Used by block assembly
+/// to track which transactions to clear after inclusion.
+///
+/// TODO: Import once defined.
+pub type TxId = Buf32;
```
delbonis: This should be in the `identifiers` crate and be called `OLTxid` or similar.

Actually, there's a macro to define wrappers for ID types like this instead of it being a plain alias.

bewakes:
Actually @krsnapaudel although it doesn't hurt to have the mempool interface to be in this PR, this PR is supposed to provide interfaces for block assembly. The original plan was to have two providers for block assembly: asm and mempool providers, but it makes more sense to have them into a single `BlockAssemblyContext` trait which you have done in this PR. The mempool interface is for the implementation of actual mempool and would be better off in a separate PR(I think it is even fine to have it together with the implementation, str-1803).

krsnapaudel:
> I'm still kinda confused by the purpose of this PR. Is there an implementation of the mempool/blockasm that exists in another branch that this is building towards? What purpose does this `Mempool` trait serve in relation to that? It's not apparent clear to me how these pieces are fitting together. For example, the `BlockAssemblyContext` trait exposes mempool txs, implying that they're external to it, but that's something I would expect the blockasm thread would be keeping in its internal state because it's actively manipulating them to produce block templates.

@delbonis: My understanding is that blockasm thread would query `NodeStorage` and `Mempool` impl. Implementation of `BlockAssemblyContext` would look something like this:

```
struct OLBlockAssemblyContext {
  storage: Arc<NodeStorage>,
  mempool: Arc<dyn Mempool>,
}

impl BlockAssemblyContext for OLBlockAssemblyContext {
...
}
```

crates/ol/block-assembly/src/context.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    /// Removes included transactions from mempool atomically.
+    ///
+    /// Clears transactions with the given [`OLTxId`] values.
+    /// Either clears all transactions or clears none (atomic operation).
+    ///
+    /// # Errors
+    ///
+    /// - [`BlockAssemblyError::Mempool`] - if mempool error occurs (e.g., transaction not found)
+    fn clear_mempool_transactions(&self, txids: &[OLTxId]) -> Result<(), BlockAssemblyError>;
```
bewakes: It might be safer to return lists containing `removed` and `missing`. If we error out when some tx is missing then there might be a situation where only a subset of the range are removed. And the next time we call this method again with same params, there'll be another error.

Basically having this return the lists allows to have idempotency resulting in safer access. What do you think?

crates/ol/block-assembly/src/context.rs
```diff
@@ -0,0 +1,63 @@
...
+
+use crate::error::BlockAssemblyError;
+
+/// Context interface for block assembly.
+pub trait BlockAssemblyContext {
+    /// Gets a block's [`OLState`].
+    ///
+    /// Returns `Ok(None)` if block doesn't exist, `Err` with [`BlockAssemblyError`] if error
+    /// occurs.
+    fn get_ol_state(&self, block_id: OLBlockId) -> Result<Option<&OLState>, BlockAssemblyError>;
```
bewakes: It should be fine to return owned `OLState` here since it will be most likely be fetched from db and won't exist inside the trait implementor.

crates/ol/block-assembly/src/context.rs
krsnapaudel: Sure. Sounds good.

crates/ol/block-assembly/src/context.rs
krsnapaudel: Based on discussion with Claude Code. `get_ol_state` is not needed in context. It will be accessed and updated with `StateAccessor`. I have updated the trait.

crates/ol/block-assembly/src/context.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// Returns transactions with their [`OLTxId`] values.
+    /// Returns empty [`Vec`] if no transactions available (not an error).
+    ///
+    /// # Errors
+    ///
+    /// - [`BlockAssemblyError::Mempool`] - if mempool error occurs
+    fn get_mempool_transactions(
+        &self,
+        limit: u64,
+    ) -> Result<Vec<(OLTxId, OLTransaction)>, BlockAssemblyError>;
```
storopoli, line 58: How are we ordering mempool transactions, by arrival ordering, FIFO?

crates/ol/block-assembly/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+workspace = true
+
+[dependencies]
+strata-asm-common.workspace = true
+strata-db.workspace = true
+strata-identifiers.workspace = true
+strata-ol-chain-types-new.workspace = true
+strata-ol-mempool.workspace = true
+strata-primitives.workspace = true
+thiserror.workspace = true
```
storopoli, line 17: We generally leave a gap between internal/external deps. See https://www.notion.so/Engineering-Rust-Best-Practices-14c901ba000f8025a974ec8e2a70b4af?source=copy_link#14d901ba000f80ec87f1fb93dae2b52f

```suggestion
strata-primitives.workspace = true

thiserror.workspace = true
```

crates/ol/mempool/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "strata-ol-mempool"
+version = "0.1.0"
+edition = "2024"
+
+[lints]
+workspace = true
+
+[dependencies]
+strata-identifiers.workspace = true
+thiserror.workspace = true
```
storopoli, line 12: Same as above

```suggestion
strata-identifiers.workspace = true

thiserror.workspace = true
```

crates/ol/block-assembly/src/context.rs
storopoli: I mean this is non-blocking and shouldn't be in the trait itself. It is particular to the upcoming implementation.
Solving this.

bewakes (approved):

crates/ol/block-assembly/src/error.rs
```diff
@@ -0,0 +1,22 @@
...
+/// Errors that can occur during block assembly operations.
+#[derive(Debug, thiserror::Error)]
+pub enum BlockAssemblyError {
+    /// Database operation failed.
+    #[error("Database error: {0}")]
+    Database(#[from] DbError),
+
+    /// Mempool operation failed.
+    #[error("Mempool error: {0}")]
+    Mempool(#[from] MempoolError),
```
delbonis, line 15: * `db: {0}`
* `mempool: {0}`

It's always redundant to use "error" in error messages.

crates/ol/block-assembly/src/error.rs
```diff
@@ -0,0 +1,22 @@
...
+
+    /// Invalid L1 block range where `from_block` height > `to_block` height.
+    #[error(
+        "Invalid L1 block range: from_block height ({from_height}) > to_block height ({to_height})"
```
delbonis: This error string seems a little hard to read imo, would be cleaner as

`invalid L1 block height range (from {from_height} to {to_height})`

bewakes (approved):
Great, this is good enough for the interface.

In the implementation though, the block assembler(sequencer) needs to insert inbox messages proofs and l1 references proofs into the transactions. That can stay out of this interface.

PR #1124, opened 2025-11-03
STR-1680 ee fullnode sync with ol

sapinb:
## Description
Adds alpen-client binary for alpen evm execution environment and tasks for syncing evm chain using state updates from OL.
Most implementations are currently defined inside this binary crate. They will be split into other crates in subsequent PRs.

- [x] Add alpen-client binary for alpen evm execution environment.
- [x] Add OL Tracker
- [x] Add engine control
- [x] Storage implementation using sled database
- [x] Storage implementation using rocksdb
~~OL Client implementation using OL rpc~~

AI use disclosure: Mostly for test boilerplate

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

      in the body of this PR.

### DIFF
--- a/bin/alpen-client/src/ol_tracker/state.rs
+++ b/bin/alpen-client/src/ol_tracker/state.rs
@@ -460,11 +460,7 @@ mod tests {
             mock_storage
                 .expect_ee_account_state()
                 .times(1)
-                .returning(|_| {
-                    Err(crate::traits::error::StorageError::database(
-                        "disk error",
-                    ))
-                });
+                .returning(|_| Err(crate::traits::error::StorageError::database("disk error")));
 
             let result = build_tracker_state(best_state, &ol_status, &mock_storage).await;
 
--- a/bin/alpen-client/src/ol_tracker/error.rs
+++ b/bin/alpen-client/src/ol_tracker/error.rs
@@ -5,7 +5,8 @@ use crate::traits::error::{OlClientError, StorageError};
 /// Error type for OL tracker operations.
 ///
 /// Errors are categorized into:
-/// - **Recoverable**: Transient failures that can be retried (network issues, temporary DB failures)
+/// - **Recoverable**: Transient failures that can be retried (network issues, temporary DB
+///   failures)
 /// - **NonRecoverable**: Fatal errors requiring intervention (no fork point found, data corruption)
 #[derive(Debug, Error)]
 pub(crate) enum OlTrackerError {


crates/ee-acct-runtime/tests/common/mod.rs
```diff
@@ -42,7 +42,7 @@ pub fn assert_update_paths_match(
     let mut unconditional_state = initial_state.clone();
     strata_ee_acct_runtime::apply_update_operation_unconditionally(
         &mut unconditional_state,
-        operation,
+        &operation.clone().into(),
```
storopoli: I think you can just do an into with an owned type here no and avoid the clone, right?

crates/snark-acct-types/src/update.rs
```diff
@@ -75,6 +75,69 @@ impl UpdateOperationData {
     }
 }
 
+/// Subset of UpdateOperationData fields needed for unconditional update.
+/// This represents all the data of an update operation that is available to a DA-only synced OL
+/// fullnode.
+#[derive(Clone, Debug)]
+pub struct UpdateOperationUnconditionalData {
```
storopoli: So this is an `UpdateOperationData`.
without:

* `ledger_refs: LedgerRefs`
* `outputs: UpdateOutputs`

I haven't found any mention of `UpdateOperationUnconditionalData` in the specs, I think we need to update the specs.

bin/alpen-client/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+workspace = true
+
+[features]
+default = ["sled"]
+sled = ["dep:sled", "dep:typed-sled", "alpen-reth-db/sled"]
+rocksdb = [
+  "alpen-reth-db/rocksdb",
+  "dep:rockbound",
+  "dep:strata-db-store-rocksdb",
+]
```
storopoli, line 17: Argh... I wanna so bad remove RocksDB!!! But soon....

bin/alpen-client/src/config.rs
```diff
@@ -0,0 +1,33 @@
...
+pub(crate) struct AlpenEeParams {
+    /// Account id of current EE in OL
+    pub account_id: AccountId,
+
+    /// Genesis blockhash of execution chain
+    pub genesis_blockhash: B256,
+
+    /// Genesis stateroot of execution chain
+    pub genesis_stateroot: B256,
+}
```
storopoli, line 65: If you want these to remain constant, then make the fields non-`pub` or `pub(crate,super)` (or even the new fancy [`pub(in foo::bar)`](https://doc.rust-lang.org/reference/visibility-and-privacy.html#pubin-path-pubcrate-pubsuper-and-pubself)) and add only accessors with shared/non-exclusive/read-only references.

bin/alpen-client/src/genesis.rs
```diff
@@ -0,0 +1,18 @@
+use alloy_primitives::B256;
+use reth_chainspec::ChainSpec;
+
+pub(crate) struct BlockInfo {
+    pub blockhash: B256,
+    pub stateroot: B256,
+}
```
storopoli, line 16: Same as above, make this non-`pub`.

bin/alpen-client/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+    #[arg(
+        long,
+        value_name = "CHAIN_OR_PATH",
+        default_value = "testnet",
+        value_parser = chain_value_parser,
+        required = false,
+    )]
+    pub custom_chain: Arc<ChainSpec>,
+
+    /// Rpc of sequener's reth node to forward transactions to.
```
storopoli: typo on sequencer's

bin/alpen-client/src/main.rs
```diff
@@ -1,4 +1,4 @@
-#![allow(unused_crate_dependencies, reason = "temporary")]
+#![warn(unused_crate_dependencies, reason = "temporary")]
```
storopoli: You can remove this now. the `warn` is the same as not having this.

bin/alpen-client/src/main.rs
```diff
@@ -70,19 +71,37 @@ fn main() {
                 account_id: AccountId::new([0; 32]),
                 genesis_blockhash: genesis_info.blockhash,
                 genesis_stateroot: genesis_info.stateroot,
+                genesis_ol_blockid: OLBlockId::null(), // TODO
```
storopoli: (blocking comment)

bin/alpen-client/src/ol_tracker/state.rs
```diff
@@ -0,0 +1,63 @@
...
+use tokio::sync::watch;
+use tracing::{debug, error, warn};
+
+use crate::{
+    config::AlpenEeConfig,
+    traits::{ol_client::OlClient, storage::Storage},
+};
+
+/// Number of Ol blocks to process in one cycle
+const MAX_BLOCKS_FETCH: u64 = 10;
```
storopoli: Should this be configurable and exposed?

bin/alpen-client/src/ol_tracker/task.rs
```diff
@@ -0,0 +1,199 @@
...
+use tokio::sync::watch;
+use tracing::{debug, error, warn};
+
+use crate::{
+    ol_tracker::OlTrackerState,
+    traits::{ol_client::OlClient, storage::Storage},
+};
+
+/// Number of Ol blocks to process in one cycle
+const MAX_BLOCKS_FETCH: u64 = 10;
```
storopoli: You already defined the same constant in `bin/alpen-client/src/ol_tracker/state.rs`.
Also, should this be configurable and exposed to the user?

bin/alpen-client/src/ol_tracker/task.rs
```diff
@@ -0,0 +1,199 @@
...
+    let best_ol_block = &ol_status.latest;
+
+    debug!(
+        local_slot = state.ol_block.slot(),
+        ol_slot = best_ol_block.slot(),
+        "check best ol block"
+    );
+
+    if best_ol_block.slot() < state.ol_block.slot() {
+        // local view of chain is ahead of Ol, should not typically happen
```
storopoli: We should emit a `warn!` or even `error!` log here.

bin/alpen-client/src/ol_tracker/task.rs
```diff
@@ -0,0 +1,199 @@
...
+        .await?;
+
+    let res = new_blocks
+        .iter()
+        .cloned()
+        .zip(operations)
+        .map(|(block, operations)| OlBlockOperations { block, operations })
+        .collect();
+
+    Ok(TrackOlAction::Extend(res))
```
storopoli: I still think that all of this should be in a `if best_ol_block.slot() > state.ol_block.slot() {...}` just for safe keeping and then outside all these `if`s add an `error`/`panic` or `ERROR` log that something REALLY BAD HAS HAPPENED.

bin/alpen-client/src/ol_tracker/task.rs
```diff
@@ -0,0 +1,199 @@
...
+            error!(
+                slot = ol_block.slot(),
+                %error,
+                "failed to store ee account state"
+            );
+            return;
+        }
+        // 3. update local state
+        state.ee_state = ee_state;
+        // TODO: notify state update
```
storopoli: (blocking comment)

bin/alpen-client/src/ol_tracker/task.rs
```diff
@@ -0,0 +1,199 @@
...
+                slot = ol_block.slot(),
+                %error,
+                "failed to apply ol block operation"
+            );
+            return;
+        }
+        // 2. Persist corresponding ee state for every ol block
+        if let Err(error) = ctx
+            .storage
+            .store_ee_account_state(ol_block, &ee_state)
```
storopoli, line 218: Does this have retries?
I gave up in the wall of macros in `strata-storage-common` when I keep hitting "go to definition".
This will bite us in the future if we don't have a retry mechanism. We had an annoying issue in the prover a couple months ago.

bin/alpen-client/src/traits/ol_client.rs
```diff
@@ -0,0 +1,164 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `start_slot` - The starting slot number (inclusive)
+    /// * `end_slot` - The ending slot number (inclusive)
+    ///
+    /// # Returns
+    ///
+    /// A vector of block commitments for the specified slot range.
+    async fn block_commitments_in_range_checked(
```
storopoli: this should be rearranged to come right after the `block_commitments_in_range`. In other words, move one function upwards.

bin/alpen-client/src/traits/ol_client.rs
```diff
@@ -0,0 +1,164 @@
...
+                expected_result_len,
+                blocks.len()
+            ));
+        }
+        Ok(blocks)
+    }
+
+    /// Retrieves update operations for the specified blocks with validation.
+    ///
+    /// This is a checked version of `get_update_operations_for_blocks` that validates
```
storopoli: doclink `get_update_operations_for_blocks`

bin/alpen-client/src/traits/ol_client.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// A vector where each element contains the update operations for the
+    /// corresponding block in the input vector.
+    async fn get_update_operations_for_blocks(
+        &self,
+        blocks: Vec<OLBlockId>,
+    ) -> eyre::Result<Vec<Vec<UpdateOperationUnconditionalData>>>;
+
+    /// Retrieves block commitments for a range of slots with validation.
+    ///
+    /// This is a checked version of `block_commitments_in_range` that validates:
```
storopoli: doclink `get_update_operations_for_blocks`

bin/alpen-client/src/traits/storage.rs
```diff
@@ -0,0 +1,85 @@
...
+        let Some(base_idx) = items.first().map(|item| item.ol_block.slot()) else {
+            return Ok(());
+        };
+        let truncate_idx = to_slot.saturating_sub(base_idx);
+
+        items.truncate(truncate_idx as usize);
+
+        Ok(())
+    }
+}
```
storopoli: Shouldn't this be only in `#[cfg(tests)]`?

bin/alpen-client/src/ol_tracker/handle.rs
```diff
@@ -31,6 +32,7 @@ impl OlTrackerHandle {
             storage,
             ol_client,
             ee_state_tx,
+            max_blocks_fetch: max_block_fetch.unwrap_or(DEFAULT_MAX_BLOCKS_FETCH),
```
storopoli: Oh, this is already configurable!

bin/alpen-client/src/ol_tracker/task.rs
```diff
@@ -12,13 +12,14 @@ use crate::{
     traits::{ol_client::OlClient, storage::Storage},
 };
 
-/// Number of Ol blocks to process in one cycle
-const MAX_BLOCKS_FETCH: u64 = 10;
+/// Default number of Ol blocks to process in one cycle
+pub(crate) const DEFAULT_MAX_BLOCKS_FETCH: u64 = 10;
```
storopoli: Ooops, I was reviewing commit by commit.

bin/alpen-client/src/ol_tracker/task.rs
```diff
@@ -228,3 +232,319 @@ async fn handle_reorg<TStorage, TOlClient>(
...
+        bytes[0] = id;
+        OLBlockCommitment::new(slot, Buf32::new(bytes).into())
+    }
+
+    /// Helper to create a test tracker state
+    fn make_test_state(slot: u64, id: u8) -> OlTrackerState {
+        OlTrackerState {
+            ee_state: EeAccountState::new([0u8; 32], BitcoinAmount::zero(), vec![], vec![]),
+            ol_block: make_block_commitment(slot, id),
+        }
```
storopoli: We should do `Arbitrary` on these?
Probably a good tech-debt ticket, right?

bin/alpen-client/src/traits/error.rs
```diff
@@ -0,0 +1,33 @@
...
+    UnexpectedOperationCount { expected: usize, actual: usize },
+
+    #[error("network error: {0}")]
+    Network(String),
+
+    #[error("rpc error: {0}")]
+    Rpc(String),
+
+    #[error(transparent)]
+    Other(#[from] eyre::Error),
```
storopoli: Add docstrings for these variants and for the enum itself.
It will be really useful for developers when trying to use these errors while coding.

Cargo.toml
storopoli, line 1: This whole 5bbd633d947dd34efd396af9cac7d79f8fd94a61 could be a standalone PR, but let's keep it as it is.

storopoli (changes requested):
This is very promising, a couple of comments.
I am solving stuff that I saw that were handled later since I reviewed this commit by commit.

crates/ee-acct-runtime/tests/common/mod.rs
sapinb: operation is a &UpdateOperationData. So don't have an owned type to into here.

An option is to change UpdateOperationData to hold UpdateOperationUnconditionalData
```
pub struct UpdateOperationData {
  unconditional: UpdateOperationUnconditionalData,
  ledger_refs: LedgerRefs,
  outputs: UpdateOutputs,
}
```
So the ref can be taken internally, and is also consistent with `UpdateOperationUnconditionalData` being a subset of `UpdateOperationData`.

wdyt ?

sapinb: updating as described above

bin/alpen-client/src/ol_tracker/task.rs
sapinb: ~~no retries yet. will do.~~

There are retries for transaction conflicts, but not for other failures as part of db transactions.
sled: bin/alpen-client/src/db/sled/db.rs `transaction_with_retry`
rocksdb: bin/alpen-client/src/db/rocksdb/db.rs `with_optimistic_txn`

bin/alpen-client/src/traits/ol_client.rs
sapinb: `block_commitments_in_range` is inside `OlClient` trait. These _checked are helper functions that work on the OlClient trait. 
Putting these as default methods inside the trait caused issues while testing with mockall, so keeping these outside as separate helper functions.

bin/alpen-client/src/traits/storage.rs
sapinb: was using this in main before implementing the db backed storage. Will remove this now.

bin/alpen-client/src/genesis.rs
sapinb: done

bin/alpen-client/src/main.rs
sapinb: updated

bin/alpen-client/src/ol_tracker/task.rs
sapinb: done

bin/alpen-client/src/ol_tracker/task.rs
sapinb: updated as requested and added and unreachable!() at the end.

bin/alpen-client/src/traits/error.rs
sapinb: added docstrings

crates/ee-acct-runtime/tests/common/mod.rs
storopoli: Ok, either way was fine.

bin/alpen-client/src/ol_tracker/task.rs
storopoli: Thanks, that makes it way more iron-clad.

bin/alpen-client/src/ol_tracker/task.rs
storopoli: Ok I've seen the new ticket. Solving this for now.

bin/alpen-client/src/traits/ol_client.rs
storopoli: I mean rearrange the methods (reorder) inside the trait. glob the `block_commitments_in_range` with `block_commitments_in_range_checked` and the other `foo` with `foo_checked`.

crates/snark-acct-types/src/update.rs
```diff
@@ -10,6 +10,71 @@ use crate::{
...
+        outputs: UpdateOutputs,
+        extra_data: Vec<u8>,
+    ) -> Self {
+        Self {
+            unconditional: UpdateOperationUnconditionalData {
+                seq_no,
+                new_state,
+                processed_messages,
+                extra_data,
+            },
```
delbonis: This naming is pretty confusing now and the constructors are a bit wonky.  I would think that it should be possible to construct the unconditional form without constructing the full version.  The way the conversions are being defined here makes it kinda feel like it's a hack to make the types fit, but really we should be thinking about how the protocol data structures are actually composed together and what they represent.

We should also think through this a little more sophisticatedly.  Since there's the form the update takes in the tx, the data that's persisted in DA, and the data that's necessary offchain but is reconstructible from DA.

In the tx:

* proof state
* extra data
* messages
* ledger refs
* outputs

Persisted in DA directly:

* proof state
* extra data

Reconstructible from DA, corresponding to an update:

* proof state
* extra data
* messages

So I'm proposing something like this:

```py
class ProofState(Container): # already exists, we should just use that def
  ...

ExtraData = List[byte, ...]

# thing we store in DA
class UpdateStateData(Container):
  proof_state: ProofState
  extra_data: ExtraData

# thing we use to update the state
class UpdateInputData(Container):
  update: UpdateStateData
  messages: List[..., ...]

# thing we use to do checks against the ledger
class UpdateOperationData(Container):
  input: UpdateInputData
  ledger_refs: ...
  outputs: ...

# for the tx: thing that authenticates everything inside
class UpdateContainer(Container):
  operation: UpdateOperationData
  message_proofs: List[..., ...]
  ledger_ref_proofs: List[..., ...]
  update_proof: List[byte, ...]
```

crates/storage-common/src/lib.rs
delbonis, line 1: Avoid reexporting the internal module structure, export the items that this crate is actually "exporting".

I know this is basically how it was before but that's bad practice that we should have started discouraging earlier. :/

crates/storage-common/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+//! Common storage utils for the Alpen codebase.
+
+pub mod cache;
+pub mod exec;
+
+// these re-exports are required for exec::inst_ops* macros
+pub use paste;
+pub use threadpool;
+pub use tracing;
```
delbonis: These *can* be made `#[doc(hidden)]` if that works.

delbonis:
Oh also it seems like there's "a lot of stuff" happening in `alpen-client` like `ol_tracker` that would be better to exist as its own crate so that it's more reusable.

crates/storage-common/src/lib.rs
sapinb: I dont see how the inst_ops macros can be used in another crate without these. I could put these inside a "pub mod private" that I've seen somewhere. Otherwise, need to have these 3 in dependency of whatever crate uses the macro, which is more confusing.

crates/storage-common/src/lib.rs
sapinb: sure. will do.

crates/snark-acct-types/src/update.rs
sapinb: updated according to this

crates/storage-common/src/lib.rs
delbonis: oh shit then maybe make it `#[doc(hidden)]`?

bin/alpen-client/src/ol_tracker/reorg.rs
```diff
@@ -0,0 +1,75 @@
...
+use super::{
+    ctx::OlTrackerCtx,
+    state::{build_tracker_state, OlTrackerState},
+};
+use crate::traits::{
+    ol_client::{block_commitments_in_range_checked, chain_status_checked, OlClient},
+    storage::Storage,
+};
+
+pub(super) async fn handle_reorg<TStorage, TOlClient>(
```
storopoli: This is an important function.
Please add docstrings.

bin/alpen-client/src/ol_tracker/reorg.rs
```diff
@@ -0,0 +1,75 @@
...
+    let mut fork_point = None;
+
+    // walk back from latest block check
+    let mut max_slot = ol_status.latest().slot();
+    while max_slot >= genesis_slot {
+        let min_slot = max_slot
+            .saturating_sub(ctx.reorg_fetch_size)
+            .max(ctx.params.genesis_ol_slot());
+
+        warn!(min_slot, max_slot, "checking slot range for fork point");
```
storopoli:
```suggestion
        warn!(%min_slot, %max_slot, "checking slot range for fork point");
```

bin/alpen-client/src/ol_tracker/reorg.rs
```diff
@@ -0,0 +1,75 @@
...
+    let Some(last_common) = fork_point else {
+        error!(
+            %genesis_slot, "reorg: could not find ol fork block till ol genesis slot",
+        );
+        eyre::bail!("reorg: could not find ol fork block");
+    };
+
+    let slot = last_common.ol_slot();
+
+    warn!(slot, "reorg: found fork point; starting db rollback");
```
storopoli:
```suggestion
    warn!(%slot, "reorg: found fork point; starting db rollback");
```

bin/alpen-client/src/ol_tracker/reorg.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    // revert own db and state to fork point
+    ctx.storage.rollback_ee_account_state(slot).await?;
+
+    // update own state
+    let next_state = build_tracker_state(last_common, &ol_status, ctx.storage.as_ref()).await?;
```
storopoli: What happens if we revert own db with success and then fail on updating the `next_state`?
Should these two operations be combined into an atomic no-side-effects one?

bin/alpen-client/src/ol_tracker/reorg.rs
```diff
@@ -5,71 +5,635 @@ use super::{
...
+    },
+    storage::{EeAccountStateAtBlock, Storage},
 };
 
-pub(super) async fn handle_reorg<TStorage, TOlClient>(
-    state: &mut OlTrackerState,
-    ctx: &OlTrackerCtx<TStorage, TOlClient>,
-) -> eyre::Result<()>
+/// Finds the last common block state between local storage and remote chain.
+pub(super) async fn find_fork_point<TStorage, TOlClient>(
```
storopoli, line 16: Oh, ok the docstrings are here. In the "add reorg tests" commit 😂 🙄

bin/alpen-client/src/ol_tracker/reorg.rs
```diff
@@ -5,71 +5,635 @@ use super::{
...
+where
+    TStorage: Storage,
+{
+    let slot = fork_state.ol_slot();
+
+    info!(slot, "rolling back to fork point");
+
+    storage.rollback_ee_account_state(slot).await?;
+
+    let next_state = build_tracker_state(fork_state.clone(), ol_status, storage).await?;
```
storopoli: Same comment as above about what happens if db reversal works but `next_state` fails?

crates/storage-common/src/lib.rs
```diff
@@ -4,6 +4,10 @@ pub mod cache;
 pub mod exec;
 
 // these re-exports are required for exec::inst_ops* macros
-pub use paste;
-pub use threadpool;
-pub use tracing;
+
```
storopoli, line 7: Why the empty line?

storopoli (changes requested):
I still think we have a potential time bomb in `reorg.rs`.

What happens if db reversal works but `build_tracker_state` fails in `rollback_to_fork_point`?

If the rollback succeeds but `build_tracker_state` fails, the system is left inconsistent.

Since `build_tracker_state` queries storage for confirmed/finalized blocks

https://github.com/alpenlabs/alpen/blob/1f5d1734511377a3dc1e183096c12f8fb6080501/bin/alpen-client/src/ol_tracker/state.rs#L97-L116

If the confirmed/finalized blocks were after the fork point, they were deleted during rollback, so `effective_account_state` returns `None` and the function errors.

bin/alpen-client/src/ol_tracker/reorg.rs
storopoli: Solved in 04cd6a65f6c2d23761dd92657f201dbf7b1cfd8e

crates/storage-common/src/lib.rs
sapinb: added `#[doc(hidden)]` and also prefixed the reexports with a `_` to indicate as internal.

bin/alpen-client/src/ol_tracker/reorg.rs
sapinb: Thanks for spotting this issue. I think moving the build_tracker_state before the rollback should be sufficient to mitigate this. Ensuring there are no fallible operations between db rollback and state mutation. Will update

crates/storage-common/src/lib.rs
sapinb: does it make a difference ? Its just some more spacing for aesthetics. 😄

bin/alpen-client/src/ol_tracker/reorg.rs
sapinb: updated.

bin/alpen-client/src/ol_tracker/reorg.rs
sapinb: updated.

bin/alpen-client/src/ol_tracker/reorg.rs
sapinb: updated, as stated in a comment above

delbonis (approved):
Ok I think this is in a good position for now.

PR #1127, opened 2025-11-05
Str 1781 impl evm ee

purusang:
## Description

It is a work heavily based on `strata-proofimpl-evm-ee-stf/src/executor.rs::process_block()` and used tons of help from Claude understanding the traits, process_block() function, evm data types, .... and to generate code.
Thanks!

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [x] Dependency Update

## Checklist

      in the body of this PR.

## Related Issues
STR-1781

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14174,7 +14174,6 @@ dependencies = [
  "alpen-reth-evm",
  "alpen-reth-primitives",
  "itertools 0.14.0",
- "proptest",
  "reth-chainspec",
  "reth-consensus-common",
  "reth-errors",
--- a/crates/evm-ee/Cargo.toml
+++ b/crates/evm-ee/Cargo.toml
@@ -43,7 +43,6 @@ itertools.workspace = true
 
 [dev-dependencies]
 alloy-eips.workspace = true
-proptest.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 


crates/evm-ee/src/execution.rs
```diff
@@ -0,0 +1,348 @@
...
+
+        // Step 14: Create BlockOutputs
+        // TODO: Convert withdrawal_intents to OutputTransfer
+        // WithdrawalIntent has Descriptor destination, OutputTransfer needs AccountId
+        // This conversion requires business logic to map Bitcoin descriptors to AccountIds
+        let outputs = BlockOutputs::new_empty();
+        // for intent in withdrawal_intents {
+        //     let account_id = convert_descriptor_to_account_id(&intent.destination)?;
+        //     outputs.add_transfer(OutputTransfer::new(account_id, intent.amt));
+        // }
```
delbonis: Descriptors describe bitcoin-level concepts, they don't contain account IDs.  We don't currently support cross-EE transfers, so the withdrawal intents should just be expressed as withdrawal messages being sent to the bridge gateway account.  See the message types spec.

crates/evm-ee/src/execution.rs
```diff
@@ -0,0 +1,348 @@
...
+    type WriteBatch = EvmWriteBatch;
+    type Block = EvmBlock;
+
+    fn execute_block_body(
+        &self,
+        pre_state: &Self::PartialState,
+        exec_payload: &ExecPayload<'_, Self::Block>,
+        // TODO: get feedbacks from Trey if this field can be unused in Eth context
+        _inputs: &BlockInputs,
+    ) -> EnvResult<ExecBlockOutput<Self>> {
```
delbonis, line 87: This function is *huge*, we should try to figure out how to break it apart into components, if possible.

crates/evm-ee/src/execution.rs
```diff
@@ -0,0 +1,348 @@
...
+        // Step 1: Create EVM config with AlpenEvmFactory
+        let evm_config =
+            EthEvmConfig::new_with_evm_factory(self.chain_spec.clone(), AlpenEvmFactory::default());
+
+        // Step 2: Prepare data for witness DB
+        // Build block_hashes map from ancestor headers (for BLOCKHASH opcode)
+        // Build bytecode_by_hash map from bytecodes (for contract execution)
+        use reth_primitives_traits::SealedHeader;
+        use revm_primitives::{B256, map::HashMap};
```
delbonis: Move these out.

crates/evm-ee/src/execution.rs
```diff
@@ -0,0 +1,348 @@
...
+            .bytecodes()
+            .iter()
+            .map(|code| (code.hash_slow(), code))
+            .collect();
+
+        // Step 3: Initialize witness database from EthereumState
+        let db = {
+            let trie_db = TrieDB::new(pre_state.ethereum_state(), block_hashes, bytecode_by_hash);
+            WrapDatabaseRef(trie_db)
+        };
```
delbonis: This is doing a lot of work that seems like it might be able to be moved into the `PartialState` impl so that we don't have to re-do it for each block we execute.

crates/evm-ee/src/execution.rs
```diff
@@ -0,0 +1,348 @@
...
+    fn execute_block_body(
+        &self,
+        pre_state: &Self::PartialState,
+        exec_payload: &ExecPayload<'_, Self::Block>,
+        // TODO: get feedbacks from Trey if this field can be unused in Eth context
+        _inputs: &BlockInputs,
+    ) -> EnvResult<ExecBlockOutput<Self>> {
+        // Step 1: Create EVM config with AlpenEvmFactory
+        let evm_config =
+            EthEvmConfig::new_with_evm_factory(self.chain_spec.clone(), AlpenEvmFactory::default());
```
delbonis: Can this be moved into the constructor?

crates/evm-ee/src/execution.rs
```diff
@@ -0,0 +1,348 @@
...
+        EthPrimitives::validate_block_post_execution(
+            &block,
+            self.chain_spec.clone(),
+            &execution_output,
+        )
+        .map_err(|_| EnvError::InvalidBlock)?;
+
+        // Step 9: Accumulate logs bloom
+        use alloy_consensus::TxReceipt;
+        use revm_primitives::alloy_primitives::Bloom;
```
delbonis: Move all of these `use` statements out.

crates/evm-ee/src/execution.rs
```diff
@@ -0,0 +1,348 @@
...
+            .execute(&block)
+            .map_err(|_| EnvError::InvalidBlock)?;
+
+        // Step 8: Validate block post-execution
+        EthPrimitives::validate_block_post_execution(
+            &block,
+            self.chain_spec.clone(),
+            &execution_output,
+        )
+        .map_err(|_| EnvError::InvalidBlock)?;
```
delbonis: This should probably be moved to `verify_outputs_against_header` if possible.

Is `validate_header` checking the block's structural consistency?  If so then we probably want to move that early.  Cheaper checks should go before more expensive ones if they're independent.

crates/evm-ee/src/execution.rs
```diff
@@ -0,0 +1,348 @@
...
+
+        // Get the intrinsics from the payload
+        let intrinsics = exec_payload.header_intrinsics();
+
+        // Get computed commitments from the write batch
+        let state_root = output.write_batch().state_root();
+        let logs_bloom = output.write_batch().logs_bloom();
+
+        // Build the complete header with both intrinsics and computed commitments
+        use alloy_consensus::Header;
```
delbonis: Move these `use` out.

crates/evm-ee/src/execution.rs
```diff
@@ -0,0 +1,348 @@
...
+            execution_output.state,
+            vec![execution_output.result.receipts],
+            block_number,
+            vec![execution_output.result.requests],
+        );
+        let hashed_post_state = executor_outcome.hash_state_slow::<KeccakKeyHasher>();
+
+        // Step 12: Compute state root
+        // Clone the pre-state, merge the hashed post state, and compute the new state root
+        let mut updated_state = pre_state.ethereum_state().clone();
```
delbonis: I *believe* this `.clone()` is **really expensive**!  Is there no way to compute the state root without applying it to an instance of the state?  It's hugely preferable to be able to compute the state root from a view of the state with the write batch without actually having to apply it.

crates/evm-ee/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+//! EVM Execution Environment implementation for Alpen.
+//!
+//! This crate provides an implementation of the `ExecutionEnvironment` trait
+//! for Ethereum Virtual Machine (EVM) block execution, enabling EVM blocks
+//! to be executed and proven within the Alpen rollup system.
+
+pub mod execution;
+pub mod types;
```
delbonis: Avoid exporting internal module structure.

crates/evm-ee/src/types.rs
```diff
@@ -0,0 +1,348 @@
...
+///
+/// Also includes witness data needed for block execution:
+/// - Contract bytecodes for executing contract calls
+/// - Ancestor block headers for the BLOCKHASH opcode
+#[derive(Clone, Debug)]
+pub struct EvmPartialState {
+    /// The sparse Merkle Patricia Trie state from RSP
+    ethereum_state: EthereumState,
+    /// Contract bytecodes needed for execution
+    /// FIXME: ensure bytecodes and ancestor_header are right to be placed in this struct
```
delbonis: This is correct, they make sense to be included here so we can look them up.  However, this indexing forces us to reindex it each time.  Surely we can make this a `BTreeMap<B256, Bytecode>` so that we can do the lookup directly, and use a better container structure for the ancestor headers.

Also, conventionally, FIXMEs shouldn't be doc comments, so they won't appear in docs, just use `//`.

crates/evm-ee/src/types.rs
```diff
@@ -0,0 +1,348 @@
...
+        ))
+    }
+
+    fn decode(_dec: &mut impl strata_codec::Decoder) -> Result<Self, CodecError> {
+        // TODO: Implement proper decoding for EthereumState
+        // For now, we'll return an error as this needs RSP's deserialization
+        Err(CodecError::InvalidVariant(
+            "EthereumState decoding not implemented",
+        ))
+    }
```
delbonis: Making a note not to forget this impl.

Also, `InvalidVariant` isn't when there's an enum variant discriminator that's not recognized.  Not sure what a more appropriate error variant would be here.

crates/evm-ee/src/types.rs
```diff
@@ -0,0 +1,348 @@
...
+            "HashedPostState encoding not implemented",
+        ))
+    }
+
+    fn decode(_dec: &mut impl strata_codec::Decoder) -> Result<Self, CodecError> {
+        // TODO: Implement proper decoding for HashedPostState
+        Err(CodecError::InvalidVariant(
+            "HashedPostState decoding not implemented",
+        ))
+    }
```
delbonis: Also note on this.

crates/evm-ee/src/types.rs
```diff
@@ -0,0 +1,348 @@
...
+            "Header encoding not implemented",
+        ))
+    }
+
+    fn decode(_dec: &mut impl strata_codec::Decoder) -> Result<Self, CodecError> {
+        // TODO: Implement proper decoding for Header using Alloy's encoding
+        Err(CodecError::InvalidVariant(
+            "Header decoding not implemented",
+        ))
+    }
```
delbonis: Also note on this.

crates/evm-ee/src/types.rs
```diff
@@ -0,0 +1,348 @@
...
+            "BlockBody encoding not implemented",
+        ))
+    }
+
+    fn decode(_dec: &mut impl strata_codec::Decoder) -> Result<Self, CodecError> {
+        // TODO: Implement proper decoding for transactions
+        Err(CodecError::InvalidVariant(
+            "BlockBody decoding not implemented",
+        ))
+    }
```
delbonis: Also this.

crates/evm-ee/src/types.rs
```diff
@@ -0,0 +1,348 @@
...
+impl Codec for EvmBlock {
+    fn encode(&self, _enc: &mut impl strata_codec::Encoder) -> Result<(), CodecError> {
+        // TODO: Implement proper encoding for block
+        Err(CodecError::InvalidVariant("Block encoding not implemented"))
+    }
+
+    fn decode(_dec: &mut impl strata_codec::Decoder) -> Result<Self, CodecError> {
+        // TODO: Implement proper decoding for block
+        Err(CodecError::InvalidVariant("Block decoding not implemented"))
+    }
```
delbonis: And this.

crates/evm-ee/src/execution.rs
purusang: @delbonis  I had a discussion with @MdTeach regarding optimizing these steps and I came to know it will require refactors in guest code that consumes it. Since I am not fully aware of that side I want to propose these optimization tasks after I do [impl EE for prover](https://alpenlabs.atlassian.net/browse/STR-1782). I can create a ticket as a reminder if you agree to this.

crates/evm-ee/src/execution.rs
purusang: Computing a Merkle Patricia Trie root requires applying changes to the trie structure. The `ExecutionEnvironment::execute_block_body` trait takes `&Self::PartialState` (immutable), which forces the clone. The reference implementation (proof-impl/evm-ee-stf) mutates `input.parent_state` directly to avoid this. Should we revisit the trait design to allow mutable state?

CLONE_OPTIMIZATION_ANALYSIS.md
delbonis, line 1: don't forget to delete this

crates/evm-ee/src/execution.rs
```diff
@@ -0,0 +1,309 @@
...
+            // This should be a well-known account ID for the bridge gateway
+            // that handles withdrawal intents from the EVM to L1
+            let bridge_gateway_account = AccountId::from([0u8; 32]);
+
+            // Create message to bridge gateway with withdrawal amount and intent data
+            let payload = MsgPayload::new(BitcoinAmount::from_sat(intent.amt), msg_data);
+            let message = SentMessage::new(bridge_gateway_account, payload);
+            outputs.add_message(message);
+        }
+    }
```
delbonis: This function is neither a constructor or an accessor so it shouldn't be an associated function, move it out.

On the address, define a const somewhere conspicuous for the acccount ID (for now, as `[1; 32]` as 0 is reserved).

If possible, remove this withdrawal_txid because it's not part of [the message format](https://www.notion.so/SPS-bridge-acct-msgs-Bridge-account-messages-1df901ba000f80e9b90fc85008f7c68e?source=copy_link#1df901ba000f802ba9a7c2b5eeb67b5f).

crates/evm-ee/src/execution.rs
```diff
@@ -0,0 +1,309 @@
...
+
+        // Step 6: Validate block post-execution
+        // Note: This validates execution-dependent fields (receipts root, gas used, requests)
+        // and cannot be moved to verify_outputs_against_header as it requires the full block
+        // and execution_output which are not available in that context
+        EthPrimitives::validate_block_post_execution(
+            &block,
+            Arc::clone(self.evm_config.chain_spec()),
+            &execution_output,
+        )
```
delbonis: Don't call `Arc::clone` like this, just do `.clone()` unless there's a good reason to do otherwise.

Also, is it *really really* required to clone these in the first place?  It's cheap (due to `Arc`), sure, but this is just a validation function, so *surely* it can take it by-ref?

crates/evm-ee/src/utils.rs
```diff
@@ -0,0 +1,68 @@
...
+
+/// Builds an Alloy block from exec payload and recovers transaction senders.
+///
+/// This constructs an AlloyBlock from the header and body in the exec payload,
+/// then recovers the sender addresses from transaction signatures.
+pub(crate) fn build_and_recover_block(
+    exec_payload: &ExecPayload<'_, EvmBlock>,
+) -> EnvResult<RecoveredBlock<AlloyBlock<TransactionSigned>>> {
+    let header = exec_payload.header_intrinsics().clone();
+    let body = exec_payload.body().body().clone();
```
delbonis, line 26: The block body is large, this clone forced or can we maybe take ownership?

crates/evm-ee/src/execution.rs
```diff
@@ -0,0 +1,309 @@
...
+        let transactions = block.into_transactions();
+        let withdrawal_intents =
+            collect_withdrawal_intents_from_execution(transactions, &execution_output.receipts);
+
+        // Step 9: Convert execution outcome to HashedPostState
+        let block_number = exec_payload.header_intrinsics().number;
+        let hashed_post_state = compute_hashed_post_state(execution_output, block_number);
+
+        // Step 10: Compute state root
+        let state_root = pre_state.compute_state_root_with_changes(&hashed_post_state);
```
delbonis: Not a change request, can you link this function?  I can't find it anywhere.

crates/evm-ee/src/utils.rs
```diff
@@ -0,0 +1,68 @@
...
+    block_number: u64,
+) -> HashedPostState {
+    let executor_outcome = ExecutionOutcome::new(
+        execution_output.state,
+        vec![execution_output.result.receipts],
+        block_number,
+        vec![execution_output.result.requests],
+    );
+    executor_outcome.hash_state_slow::<KeccakKeyHasher>()
+}
```
delbonis, line 43: I might be misunderstanding something, but why do we need these receipts/requests when computing the post state root?  I would think we *just* need `execution_output.state` and little else.  Receipts and requests are outputs, but hashing them in the state root implies that they're part of the state and we'd need them in the *next* block.  But I might be misunderstanding what the EVM state trie contains, in which case ignore this.

crates/evm-ee/src/utils.rs
```diff
@@ -0,0 +1,68 @@
...
+
+/// Collects withdrawal intents from executed transactions and their receipts.
+pub(crate) fn collect_withdrawal_intents_from_execution(
+    transactions: Vec<TransactionSigned>,
+    receipts: &[EthereumReceipt],
+) -> Vec<alpen_reth_primitives::WithdrawalIntent> {
+    let executed_txns = transactions.iter();
+    let receipt_refs = receipts.iter();
+    let tx_receipt_pairs = executed_txns.zip(receipt_refs);
+    collect_withdrawal_intents(tx_receipt_pairs).collect()
```
delbonis: I realize this is defined in the other file, but `collect_withdrawal_intents` being prefixed with `collect_` is a bit weird when it returns an iterator we call `.collect()` on.

Also this consumes `transactions` which doesn't seem necessary.  Convert it to `&[]`?  Or use `impl IntoIterator<Item = &_>`?

Also I think this silently misbehaves when the two lists passed aren't the same length.

crates/evm-ee/src/utils.rs
```diff
@@ -0,0 +1,68 @@
...
+}
+
+/// Accumulates logs bloom from all receipts in the execution output.
+pub(crate) fn accumulate_logs_bloom(receipts: &[EthereumReceipt]) -> Bloom {
+    let mut logs_bloom = Bloom::default();
+    receipts.iter().for_each(|r: &EthereumReceipt| {
+        logs_bloom.accrue_bloom(&r.bloom());
+    });
+    logs_bloom
+}
```
delbonis: Shouldn't this be part of the alpen revm crate?  This is just a regular EVM impl function, not specific to the EE infra.

.gitignore
```diff
@@ -273,3 +273,6 @@ provers/sp1/proofs/
 
 # Docker env
 docker/prover-client/prover-client.env
+
+# AI Agent's context files local to developers
+CONTEXT_*.md
```
delbonis: Put this in your `.git/info/exclude`.

crates/evm-ee/src/execution.rs
storopoli: Just pass a `&self.evm_config.chain_spec()` and maybe it will accept.

.gitignore
storopoli: Yes, heavy user of `.git/info/exclude`. It's your local `.gitignore`

crates/evm-ee/src/execution.rs
purusang: [compute_state_root_with_changes()](https://github.com/alpenlabs/strata/blob/STR-1781-impl-evm-ee/crates/evm-ee/src/types/mod.rs#L198)

crates/evm-ee/src/utils.rs
purusang: Must clone it as it is being referenced from `&exec_payload`.

crates/evm-ee/src/utils.rs
purusang: Absolutely right in this context.

crates/evm-ee/src/execution.rs
krsnapaudel: The `withdrawal_txid` was added for monitoring to work in Testnet 1. We need to consider how monitoring will work when we remove it. We do get the txid itself using `eth_getLogs`, but need to figure how linking the txid with withdrawal fulfillments will work. Please check @sapinb's comment on [the PRD for monitoring](https://docs.google.com/document/d/14ELO-teA_yfik1nvXOPzCwf0rn9n8KOmsnBl_C0C1pA/edit?tab=t.0).

storopoli: Maybe we need to review the spec @delbonis?

delbonis: The *right* way to expose this is to

* on the EVM side: index the withdrawal initiation logs to pick up the txids
* on the OL side: index the snark account updates to match the EVM withdrawal indexes with messages sent to the bridge gateway account, and then the OL logs getting emitted from that
* on the L1 side: index checkpoints to match the OL withdrawal log indexes with withdrawal assignments

I agree this is way out of scope of this PR.  But we're no longer under the release pressure we were under when this feature was initially implemented, so I don't think we should incur technical debt to directly translate over the implementation to the new context, we should do it right.

crates/evm-ee/src/utils.rs
delbonis: I see that.  What I am suggesting is to have this function consume the `ExecPayload` instead of forcing the clone.  I realize that `execute_block_body` takes it by ref, but would it break everything to change the interface?

Alternatively, do we really *need* to convert to the alloy types (which take ownership)?  Or can we use different impls of the revm [`Block`](https://reth.rs/docs/reth_ethereum/evm/primitives/revm/context_interface/block/trait.Block.html) and [`Transaction`](https://reth.rs/docs/reth/revm/revm/context/transaction/trait.Transaction.html) traits which let us do more borrowing?

crates/evm-ee/src/types/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+
+/// Helper function to decode raw bytes with length prefix.
+///
+/// This reads a u32 length prefix, then reads that many bytes and returns them as a Vec<u8>.
+fn decode_bytes_with_length(dec: &mut impl strata_codec::Decoder) -> Result<Vec<u8>, CodecError> {
+    let len = u32::decode(dec)? as usize;
+    let mut bytes = vec![0u8; len];
+    dec.read_buf(&mut bytes)?;
+    Ok(bytes)
+}
```
delbonis: Move these to a separate module (`codec_shims`?) and re-export if needed.  Also, consider using varints instead of u32 if we expect these to usually be small.

crates/evm-ee/src/types/partial_state.rs
```diff
@@ -0,0 +1,215 @@
...
+    fn compute_state_root(&self) -> EnvResult<Hash> {
+        let state_root = self.ethereum_state.state_root();
+        Ok(state_root.into())
+    }
+}
+
+impl Codec for EvmPartialState {
+    fn encode(&self, enc: &mut impl strata_codec::Encoder) -> Result<(), CodecError> {
+        // Encode EthereumState using bincode (it has serde support and is a complex MPT structure)
+        let ethereum_state_bytes = bincode::serialize(&self.ethereum_state)
```
delbonis: Is there no way to encode this except by using bincode?  Bincode is horrible.  It's a hack on top of serde's data model and in this context it's not a stable format because it depends on very specific things about the Reth implementation of the types that makes it not as portable as typical serde encodings would be.

crates/evm-ee/src/types/partial_state.rs
```diff
@@ -0,0 +1,215 @@
...
+            block_hashes.insert(parent_header.number(), child_header.parent_hash());
+        }
+
+        // Build bytecode_by_hash from bytecodes for contract execution
+        // Since bytecodes are already indexed by hash in the BTreeMap, we can iterate directly
+        let bytecode_by_hash: HashMap<B256, &revm::state::Bytecode> = self
+            .bytecodes
+            .iter()
+            .map(|(hash, code)| (*hash, code))
+            .collect();
```
delbonis: This is still doing *quite* a lot of work, and it gets called over and over again for every block.  It really doesn't seem like it needs to be repeated like this.

We might have to provide our own implementation of `revm::database::Database` but that should be a very easy shim down to this type instead of using `TrieDB`.  I looked at [the impl of `TrieDb`](https://github.com/succinctlabs/rsp/blob/main/crates/executor/client/src/io.rs) and it doesn't seem like there's much to it, so you shouldn't feel bound to using it as-is.

crates/evm-ee/src/types/block.rs
```diff
@@ -0,0 +1,77 @@
...
+    type Body = EvmBlockBody;
+
+    fn from_parts(header: Self::Header, body: Self::Body) -> Self {
+        Self { header, body }
+    }
+
+    fn check_header_matches_body(header: &Self::Header, body: &Self::Body) -> bool {
+        // Validate that the transactions root in the header matches the body's transactions
+        // This uses Alloy's proofs::calculate_transaction_root to compute the root
+        use alloy_consensus::proofs::calculate_transaction_root;
```
delbonis: clean up

crates/evm-ee/src/types/block.rs
```diff
@@ -0,0 +1,77 @@
...
+        Ok(())
+    }
+
+    fn decode(dec: &mut impl strata_codec::Decoder) -> Result<Self, CodecError> {
+        // Decode header and body separately using their Codec implementations
+        let header = EvmHeader::decode(dec)?;
+        let body = EvmBlockBody::decode(dec)?;
+        Ok(Self { header, body })
+    }
+}
```
delbonis: You can use [the macro](https://github.com/alpenlabs/strata-common/blob/main/crates/codec/src/macros.rs#L22) when it's a simple case like this.

crates/evm-ee/src/types/write_batch.rs
```diff
@@ -0,0 +1,94 @@
...
+    /// Consumes self and returns the underlying HashedPostState.
+    pub fn into_hashed_post_state(self) -> HashedPostState {
+        self.hashed_post_state
+    }
+}
+
+impl Codec for EvmWriteBatch {
+    fn encode(&self, enc: &mut impl strata_codec::Encoder) -> Result<(), CodecError> {
+        // Encode HashedPostState using bincode (it has serde support)
+        let hashed_post_state_bytes = bincode::serialize(&self.hashed_post_state)
```
delbonis: Same comment about bincode.

crates/evm-ee/src/execution.rs
```diff
@@ -0,0 +1,309 @@
...
+    type PartialState = EvmPartialState;
+    type WriteBatch = EvmWriteBatch;
+    type Block = EvmBlock;
+
+    fn execute_block_body(
+        &self,
+        pre_state: &Self::PartialState,
+        exec_payload: &ExecPayload<'_, Self::Block>,
+        // TODO: get feedbacks from Trey if this field can be unused in Eth context
+        _inputs: &BlockInputs,
```
delbonis: This is how deposits get provided to the EE.

There is the "withdrawals" header for validator that we're hijacking, but that's not free.  We have to check that field against this arg which is the actual authenticated inputs from the OL.

crates/reth/evm/src/utils.rs
```diff
@@ -22,37 +23,52 @@ pub(crate) fn wei_to_sats(wei: U256) -> (U256, U256) {
...
-
-            let event = WithdrawalIntentEvent::decode_log(log).ok()?;
-            let destination = Descriptor::from_bytes(&event.destination).ok()?;
-
-            Some(WithdrawalIntent {
-                amt: event.amount,
-                destination,
-                withdrawal_txid: txid,
+/// A [`Descriptor`], if invalid does not create a [`WithdrawalIntent`].
+pub fn withdrawal_intents<'a>(
```
delbonis: Okay this is a worse name because it's just a noun not a verb.

`extract_withdrawal_intents`, `generate_withdrawal_intents`, etc are all maybe suitable names.

crates/reth/evm/src/utils.rs
```diff
@@ -22,37 +23,52 @@ pub(crate) fn wei_to_sats(wei: U256) -> (U256, U256) {
...
+    transactions: &'a [TransactionSigned],
+    receipts: &'a [Receipt],
+) -> impl Iterator<Item = WithdrawalIntent> + 'a {
+    assert_eq!(
+        transactions.len(),
+        receipts.len(),
+        "transactions and receipts must have the same length"
+    );
```
delbonis, line 44: Either make sure to note this panic condition in the doc comment (`# Panics`) or make the return type be a result.

crates/reth/evm/src/utils.rs
```diff
@@ -22,37 +23,52 @@ pub(crate) fn wei_to_sats(wei: U256) -> (U256, U256) {
...
+        .iter()
+        .zip(receipts.iter())
+        .flat_map(|(tx, receipt)| {
+            let txid = Buf32((*tx.hash()).into());
+            receipt.logs.iter().filter_map(move |log| {
+                if log.address != BRIDGEOUT_PRECOMPILE_ADDRESS {
+                    return None;
+                }
+
+                let event = WithdrawalIntentEvent::decode_log(log).ok()?;
```
delbonis, line 56: I realize this is defined in the other file, but `Event`?  That's a term that isn't used anywhere else afaik.

crates/evm-ee/src/types/partial_state.rs
purusang: 🧠

crates/reth/evm/src/utils.rs
purusang: Yeah, what do you think would be a better name for it?

crates/evm-ee/src/execution.rs
purusang: Does this also means we will have validator withdrawals too? How do we order them in that case. Please link me the code or doc.

crates/evm-ee/src/execution.rs
barakshani: > I agree this is way out of scope of this PR

Let's not make PRs bigger than they should be, especially since Puru is new to the codebase, and it's better to take it step by step. Let's create a ticket for the extra work, and if needed start working on it in the same sprint. But I suggest we keep the context of each PR smaller, rather than bigger.

crates/evm-ee/src/execution.rs
delbonis: No no, we're (ab)using the mechanism that's used to implement validator withdrawals in order to implement our deposits.  So we have to compare the block inputs to that list, since that's the list that's exposed to the EVM machinery, *unless* we just enforce that that list is always empty and we have our own mechanism for it.

crates/reth/evm/src/utils.rs
delbonis: WithdrawalIntentCallData maybe?

crates/evm-ee/src/execution.rs
purusang: Created a [ticket](https://alpenlabs.atlassian.net/browse/STR-1911) for it.

crates/evm-ee/src/execution.rs
```diff
@@ -0,0 +1,297 @@
...
+    for intent in withdrawal_intents {
+        // Encode withdrawal intent data: descriptor bytes + txid
+        let mut msg_data = Vec::new();
+        msg_data.extend_from_slice(&intent.destination.to_bytes());
+        msg_data.extend_from_slice(&intent.withdrawal_txid.0);
+
+        let bridge_gateway_account = AccountId::from(BRIDGE_GATEWAY_ACCOUNT);
+
+        // Create message to bridge gateway with withdrawal amount and intent data
+        let payload = MsgPayload::new(BitcoinAmount::from_sat(intent.amt), msg_data);
```
delbonis: Make sure this follows the message format properly.  There's no need to include the txid here anymore.

crates/reth/evm/src/utils.rs
purusang: I have noted this, will change this in another work following this pr.

crates/evm-ee/src/execution.rs
purusang: Noted.

crates/evm-ee/src/execution.rs
```diff
@@ -0,0 +1,309 @@
...
```
delbonis: Use [this message data](https://github.com/alpenlabs/alpen/blob/main/crates/ol/msg-types/src/withdrawal.rs#L36) to construct the message.

crates/evm-ee/src/types/partial_state.rs
```diff
@@ -0,0 +1,283 @@
...
+}
+
+impl ExecPartialState for EvmPartialState {
+    fn compute_state_root(&self) -> EnvResult<Hash> {
+        let state_root = self.ethereum_state.state_root();
+        Ok(state_root.into())
+    }
+}
+
+impl Codec for EvmPartialState {
```
delbonis, line 185: Probably should have fuzzing/proptesting on this.

delbonis (changes requested):

crates/evm-ee/src/types/partial_state.rs
purusang: @delbonis I have created a [ticket](https://alpenlabs.atlassian.net/browse/STR-1934) for it to unblock other tasks as proptesting it with `EthereumState` became a bit complicated.

PR #1130, opened 2025-11-06
fix(bridge-subprotocol): fulfillment deadline in reassignments

prajwolrg:
## Description

The current implementation doesn't update the `exec_deadline` during reassignment. This fixes that. It also avoids making changes to the operator fees during reassignments.

It also makes the following refactors:
- **Rename `exec_deadline` to `fulfillment_deadline`** for clarity .
- **Move `assignment_duration` from `BridgeV1State` to `AssignmentTable`** for better encapsulation.

### Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update
- [X] Refactor
- [X] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
+++ b/crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
@@ -228,7 +228,7 @@ impl AssignmentEntry {
     ///
     /// # Parameters
     ///
-    /// - `new_deadline` - The new absolute Bitcoin block height deadline for execution
+    /// - `new_deadline` - The new absolute Bitcoin block height deadline for fulfillment
     /// - `seed` - L1 block ID used as seed for deterministic random selection
     /// - `current_active_operators` - Bitmap of currently active operator indices
     ///
@@ -396,7 +396,7 @@ impl AssignmentTable {
 
     /// Reassigns all expired assignments to new randomly selected operators.
     ///
-    /// Iterates through all assignments and reassigns those whose execution deadlines
+    /// Iterates through all assignments and reassigns those whose fulfillment deadlines
     /// have passed (current height >= fulfillment_deadline). Each expired assignment is
     /// reassigned using the provided seed for deterministic random operator selection.
     ///
@@ -407,11 +407,9 @@ impl AssignmentTable {
     ///
     /// # Parameters
     ///
-    /// - `operator_fee` - The new operator fee for reassigned withdrawals
-    /// - `current_height` - The current Bitcoin block height for expiration comparison
-    /// - `new_deadline` - The new absolute Bitcoin block height deadline for execution
     /// - `current_active_operators` - Bitmap of currently active operator indices
-    /// - `seed` - L1 block ID used as seed for deterministic random selection
+    /// - `l1_block` - The L1 block commitment used to derive the current height, seed, and new
+    ///   fulfillment deadline
     ///
     /// # Returns
     ///


crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -227,8 +227,10 @@ impl AssignmentEntry {
     ///
     /// # Parameters
     ///
+    /// - `new_operator_fee` - The new operator fee for the withdrawal
```
barakshani: is this new operator or new fee? why do we need this?

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -227,8 +227,10 @@ impl AssignmentEntry {
...
+    /// - `new_deadline` - The new absolute Bitcoin block height deadline for execution
```
```suggestion
    /// - `new_deadline` - The new absolute Bitcoin block height deadline for fulfillment

File: crates/asm/subprotocols/bridge-v1/src/state/assignment.rs [hunk 3]
```diff
@@ -399,29 +403,40 @@ impl AssignmentTable {
     ///
     /// # Parameters
     ///
+    /// - `operator_fee` - The new operator fee for reassigned withdrawals
```
barakshani: let's be consistent: above this is called `new_operator_fee`.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -399,29 +403,40 @@ impl AssignmentTable {
...
     /// let seed = L1BlockId::from([0u8; 32]);
     ///
-    /// table.reassign_expired_assignments(current_height, &active_operators, seed)?;
+    /// let reassigned = table.reassign_expired_assignments(
+    ///     operator_fee,
+    ///     current_height,
+    ///     new_deadline,
+    ///     &active_operators,
+    ///     seed
+    /// )?;
```
barakshani: just wondering: what exactly do we show here? what does the example clarify?

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -580,14 +596,70 @@ mod tests {
...
+        assert_eq!(
+            assignment.exec_deadline(),
+            new_deadline,
+            "Exec deadline should be updated to the new deadline after reassignment"
+        );
+        assert_ne!(
+            assignment.exec_deadline(),
+            initial_deadline,
+            "Exec deadline should have changed from initial value"
+        );
```
barakshani: isn't this redundant?

storopoli: These should probably be an `assert!(assignment.exec_deadline() > new_dealine > initial_deadline)` no?

storopoli (changes requested):

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
delbonis: ^ yeah it's not useful to have examples for internal functions like this that aren't meant to be used outside of a particular algorithm

Also, I'm not sure this actually belongs as a member function, it might make more sense as a separate fn.

storopoli (approved):
ACK a6c40833557d5bc0f7641789e676f439acc12e2a

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
prajwolrg: I realize this is redundant for now. The idea was that since the `operator_fee` is stored in the bridge state, during each reassignment we use the latest operator_fee from the state. Since we don't support updating the operator_fee, this just seems to make things complicated. I've removed this altogether.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
barakshani: nagging because you gave the thumb up

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
barakshani: this is removed now, right?

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -399,50 +407,85 @@ impl AssignmentTable {
...
+    /// * `withdrawal_cmd` - The withdrawal command to be assigned
+    /// * `current_active_operators` - Bitmap of currently active operators eligible for assignment
+    /// * `l1_block` - The L1 block commitment used to anchor the assignment and calculate the
+    ///   deadline
+    ///
+    /// # Returns
+    ///
+    /// Returns `Ok(())` if the assignment was created and added successfully, or an error if
+    /// the assignment creation failed (e.g., no operators available).
+    pub fn add_new_assignment(
```
barakshani, line 460: is this new??

barakshani: If previously we only had `create_with_random_assignment`, then why do we need this?

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
prajwolrg: yes.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
prajwolrg: Previously this was done in the `state/bridge.rs::create_assignment_withdrawal`. We moved some of the logic from there to this function. It's a refactor for better encapsulation.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
prajwolrg: that's alright. i appreciate you pointing out the things i missed.

delbonis (approved):

PR #1140, opened 2025-11-11
feat: handle cooperative transactions

sistemd:
## Description

Parse and handle cooperative withdrawal transactions.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

- [X] My changes do not introduce new warnings.

### DIFF
--- a/crates/asm/txs/bridge-v1/src/constants.rs
+++ b/crates/asm/txs/bridge-v1/src/constants.rs
@@ -6,6 +6,11 @@ use strata_l1_txfmt::{SubprotocolId, TxType};
 /// and must match the `subprotocol_id` checked in `SectionState::subprotocol()`.
 pub const BRIDGE_V1_SUBPROTOCOL_ID: SubprotocolId = 2;
 
+/// Transaction type identifier for deposit transactions in the Bridge V1 subprotocol.
 pub const DEPOSIT_TX_TYPE: TxType = 1;
+
+/// Transaction type identifier for withdrawal transactions in the Bridge V1 subprotocol.
 pub const WITHDRAWAL_TX_TYPE: TxType = 2;
+
+/// Transaction type identifier for cooperative transactions in the Bridge V1 subprotocol.
 pub const COOPERATIVE_TX_TYPE: TxType = 3;
--- a/crates/asm/txs/bridge-v1/src/errors.rs
+++ b/crates/asm/txs/bridge-v1/src/errors.rs
@@ -112,9 +112,11 @@ pub enum CooperativeParseError {
     #[error("Invalid transaction type: expected type to be {COOPERATIVE_TX_TYPE}, got {0}")]
     InvalidTxType(TxType),
 
+    /// The transaction is missing bridge input.
     #[error("Transaction is missing bridge input")]
     MissingBridgeInput,
 
+    /// The transaction is missing cooperative output that fulfills the user withdrawal request.
     #[error("Transaction is missing output that fulfilled user withdrawal request")]
     MissingWithdrawalOutput,
 }


crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -331,6 +333,63 @@ impl BridgeV1State {
...
+            return Err(CooperativeValidationError::DepositUtxoMismatch(Mismatch {
+                expected: expected_utxo,
+                got: actual_utxo,
+            }));
+        }
+
+        // Validate withdrawal destination against assignment command
+        let expected_destination = assignment.withdrawal_command().destination().to_script();
+        let actual_destination = cooperative_info.withdrawal_destination.clone();
+        if expected_destination != actual_destination {
```
storopoli, line 383: I think you can avoid the `clone()` here by comparing like

```suggestion
        let expected_destination = assignment.withdrawal_command().destination();
        let actual_destination = cooperative_info.withdrawal_destination;
        if expected_destination != actual_destination {
```

I think both of them are [`bosd::Descriptor` that implements `Eq`](https://github.com/alpenlabs/bitcoin-bosd/blob/8dd2dce9e8284c746a2f5b836d56bc9a960742a6/src/descriptor.rs#L65-L73) so that should work with references as well instead of owned values.

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -331,6 +333,63 @@ impl BridgeV1State {
...
+    ///
+    /// - `Ok(())` - If the withdrawal is valid according to assignment information
+    /// - `Err(CooperativeValidationError)` - If validation fails for any reason
+    ///
+    /// # Errors
+    ///
+    /// Returns error if:
+    /// - No assignment exists for the referenced deposit
+    /// - The withdrawal specifications don't match the assignment
+    fn validate_cooperative(
```
storopoli, line 358: According to the ticket as noted by @barakshani we should also validate:

> amount: the assigned operator should get an agreed fee, and the rest of the funds (potentially minus mining fees?) should go to the user. However, if all operators colluded to give the user an incorrect amount, they can just steal from the bridge.

crates/asm/txs/bridge-v1/src/cooperative/parse.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// The index of the deposit that the operator wishes to receive payout from later.
+    /// This must be validated against the operator's assigned deposits in the state's assignments
+    /// table to ensure the operator is authorized to claim this specific deposit.
+    pub deposit_idx: u32,
+
+    /// The UTXO of the deposit being spent in this cooperative transaction.
+    pub deposit_utxo: BitcoinOutPoint,
+
+    /// The scriptPubKey where the withdrawal funds are being sent.
+    pub withdrawal_destination: ScriptBuf,
```
storopoli, line 28: Don't use `ScriptBuf`.
We created BOSD exactly for this.
`ScriptBuf`s can be non-relay-safe (non-standard) spks and an attack vector.
See https://github.com/alpenlabs/bitcoin-bosd/blob/main/SPECIFICATION.md
Use `bosd::Descriptor` here.

This is part of the "Don't parse, validate" principle in the Engineering Best Practices.

crates/asm/txs/bridge-v1/src/test_utils/withdrawal_fulfillment.rs
```diff
@@ -69,3 +70,55 @@ pub fn create_test_withdrawal_fulfillment_tx(
...
+    tagged_payload.push(BRIDGE_V1_SUBPROTOCOL_ID); // 1 byte subprotocol ID
+    tagged_payload.push(COOPERATIVE_TX_TYPE); // 1 byte transaction type
+
+    // Auxiliary data: [DEPOSIT_IDX]
+    tagged_payload.extend_from_slice(&cooperative_info.deposit_idx.to_be_bytes()); // 4 bytes
+
+    // Create OP_RETURN script with the tagged payload
+    let op_return_script = ScriptBuf::new_op_return(
+        PushBytesBuf::try_from(tagged_payload).expect("Tagged payload should fit in push bytes"),
+    );
```
storopoli, line 100: We don't like allocations here 😂.
That's why @Zk2u created `make_buf`: https://github.com/alpenlabs/make_buf
Use that instead of `Vec::new()` for `OP_RETURN`s.
See an example here
https://github.com/alpenlabs/alpen/blob/b1f49c73d7b58072aaf2d7587afaa03797bbd086/bin/strata-test-cli/src/bridge/dt.rs#L219-L235

crates/asm/txs/bridge-v1/src/constants.rs
```diff
@@ -8,3 +8,4 @@ pub const BRIDGE_V1_SUBPROTOCOL_ID: SubprotocolId = 2;
 
 pub const DEPOSIT_TX_TYPE: TxType = 1;
 pub const WITHDRAWAL_TX_TYPE: TxType = 2;
+pub const COOPERATIVE_TX_TYPE: TxType = 3;
```
storopoli, line 16: Can we document these constants?

crates/asm/txs/bridge-v1/src/errors.rs
```diff
@@ -95,3 +96,25 @@ pub enum WithdrawalParseError {
...
+    /// The transaction type byte in the tag does not match the expected withdrawal fulfillment
+    /// transaction type.
+    #[error("Invalid transaction type: expected type to be {COOPERATIVE_TX_TYPE}, got {0}")]
+    InvalidTxType(TxType),
+
+    #[error("Transaction is missing bridge input")]
+    MissingBridgeInput,
+
+    #[error("Transaction is missing output that fulfilled user withdrawal request")]
+    MissingWithdrawalOutput,
```
storopoli, line 121: Please add docstrings here. They help out when using these variants as a developer because we don't see the `#error` macro in the LSP but the docstrings we do.

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
sistemd: From what I can see, the ticket says:

> Two more things that we **may** want to check/validate (**tbd**):

Then explains:

> amount: the assigned operator **should** get an agreed fee. ... However, if all operators colluded to give the user an incorrect amount, they can just steal from the bridge.

My interpretation of this comment was the opposite from what you're saying: there's no real point in verifying the amount – if all of the operators are colluding, they can just steal money anyway. In other words, the "verification" of the amount is ultimately up to the operators at signing time.

crates/asm/txs/bridge-v1/src/test_utils/withdrawal_fulfillment.rs
sistemd: This is creating a transaction for tests. Are you sure this really matters? It's just test code anyway. `create_test_withdrawal_fulfillment_tx` works the same way.

crates/asm/txs/bridge-v1/src/cooperative/parse.rs
sistemd: For the record, the [withdrawal fulfillment side](https://github.com/alpenlabs/alpen/blob/d76d20c3aeb9104c49cd3a922ee9d941770ff0e3/crates/asm/txs/bridge-v1/src/withdrawal_fulfillment/parse.rs#L40) also uses a raw `ScriptBuf` here. I don't think that is by mistake, either. AFAIU (and please correct me if I'm wrong!) the `bosd::Descriptor` contains an additional tag which differentiates common types of locking scripts, and also contains the actual script data. The main problem with raw locking scritps is that you don't necessarily know exactly what it is without the unlocking script.

But we don't have a tag here. We're working with an onchain transaction. The locking script could be anything, there is no way to convert it to a deterministic `bosd::Descriptor`.

What am I missing here?

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
sistemd: In your code, `actual_destination` is a `ScriptBuf` and `expected_destination` is a `&Descriptor`, leading to a compiler error indicating that the two cannot be compared.

crates/asm/txs/bridge-v1/src/errors.rs
sistemd: https://github.com/alpenlabs/alpen/pull/1140/commits/cdac3e8566d6b4057555896394ca5cae39c889e5

crates/asm/txs/bridge-v1/src/constants.rs
sistemd: https://github.com/alpenlabs/alpen/pull/1140/commits/cdac3e8566d6b4057555896394ca5cae39c889e5

crates/asm/txs/bridge-v1/src/cooperative/parse.rs
storopoli: I think that both should've been `bosd::Descriptor`.
Cc @prajwolrg for the `git blame` here.
@delbonis might have some insights...

crates/asm/txs/bridge-v1/src/test_utils/withdrawal_fulfillment.rs
storopoli: Yeah, ok I get your point.
Yes, for testing is fine.
Solving this.

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
storopoli: Yes, solving this because we're already discussing `bosd::Descriptor`s below.

PR #1149, opened 2025-11-15
STR-1815: prover tasks exposed as a separate service (via service framework)

evgenyzdanovich:
## Description

Expose PaaS: Prover-as-a-Service - a minimalistic Service that allows to define, execute, track and store proving tasks(programs). Think of PaaS as breeding  **Strata Service Framework** with **ZKaleido**.

The Prover Service is a general-purpose framework for managing proof generation with robust task lifecycle management, worker pooling, retry logic, and graceful error handling while remaining completely agnostic to the underlying proof system (thanks to ZKaleido).

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6975,7 +6975,7 @@ source = "git+https://github.com/alpenlabs/moho#274f1b9a9b3961a650c19de2e1b45ae6
 dependencies = [
  "borsh",
  "hex",
- "sha2 0.10.9",
+ "sha2",
  "ssz",
  "ssz_codegen",
  "ssz_derive",
@@ -11288,7 +11288,7 @@ dependencies = [
  "opentelemetry_sdk 0.31.0",
  "tracing",
  "tracing-opentelemetry 0.32.0",
- "tracing-subscriber 0.3.20",
+ "tracing-subscriber 0.3.22",
  "url",
 ]
 
@@ -13473,7 +13473,7 @@ version = "0.1.0"
 dependencies = [
  "digest 0.10.7",
  "proptest",
- "sha2 0.10.9",
+ "sha2",
  "ssz",
  "ssz_codegen",
  "ssz_derive",
@@ -13782,7 +13782,7 @@ dependencies = [
  "secp256k1 0.29.1",
  "serde",
  "serde_json",
- "sha2 0.10.9",
+ "sha2",
  "ssz",
  "ssz_derive",
  "strata-checkpoint-types",
@@ -14010,8 +14010,8 @@ dependencies = [
  "serde",
  "tokio",
  "tracing",
- "tracing-opentelemetry",
- "tracing-subscriber 0.3.20",
+ "tracing-opentelemetry 0.27.0",
+ "tracing-subscriber 0.3.22",
 ]
 
 [[package]]
@@ -14390,7 +14390,7 @@ dependencies = [
  "rand 0.8.5",
  "serde",
  "serde_json",
- "sha2 0.10.9",
+ "sha2",
  "ssz",
  "ssz_derive",
  "strata-codec",
@@ -14471,7 +14471,7 @@ dependencies = [
  "borsh",
  "digest 0.10.7",
  "serde",
- "sha2 0.10.9",
+ "sha2",
  "ssz",
  "ssz_codegen",
  "ssz_derive",
@@ -16100,7 +16100,7 @@ dependencies = [
  "tracing",
  "tracing-core",
  "tracing-log",
- "tracing-subscriber 0.3.20",
+ "tracing-subscriber 0.3.22",
  "web-time",
 ]
 
@@ -16150,7 +16150,7 @@ version = "0.8.3"
 source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.8.3#f16bea1f0c7925296e8a4c82d358bd1bb9fe71d8"
 dependencies = [
  "digest 0.10.7",
- "sha2 0.10.9",
+ "sha2",
  "smallvec",
  "ssz",
  "ssz_primitives",


crates/paas/src/registry/core.rs
```diff
@@ -0,0 +1,291 @@
...
+}
+
+/// Trait for providing inputs for a specific zkVM program
+pub trait InputProvider<P, Prog>: Send + Sync + 'static
+where
+    P: ProgramType,
+    Prog: ZkVmProgram,
+{
+    /// Provide input for the given program
+    fn provide_input<'a>(
```
bewakes: Can it be named `get_input` instead? While calling this method, it seems like the caller is providing something to it instead of receiving from it.

crates/paas/src/worker.rs
```diff
@@ -0,0 +1,202 @@
...
+
+        if should_skip {
+            return;
+        }
+
+        // Spawn proof generation task
+        let state = self.state.clone();
+        let backend_for_guard = backend.clone();
+
+        spawn(async move {
```
bewakes: nit: This part can be in a separate function. The function is very long.

crates/paas/tests/unit_test.rs
```diff
@@ -0,0 +1,247 @@
...
+    assert_eq!(config.calculate_delay(10), 30);
+
+    // Test should_retry
+    assert!(config.should_retry(0));
+    assert!(config.should_retry(4));
+    assert!(!config.should_retry(5));
+}
+
+#[test]
+fn test_zkvm_backend_serialization() {
```
bewakes: If there is a deserialization, maybe we could do a round-trip test?

crates/paas/src/config.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// Number of workers per backend
+    pub worker_count: HashMap<B, usize>,
+
+    /// Polling interval for checking pending tasks (milliseconds)
+    pub polling_interval_ms: u64,
+}
+
+/// Retry policy configuration
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RetryConfig {
```
bewakes, line 49: There's some primitives for retrying in `crates/common/src/retry`. It might be a good idea to see if that fits this case and thus potentially reduce the duplication.

bewakes (commented):
Great work!

crates/paas/src/config.rs
evgenyzdanovich: Agree, thanks!

Moreover, ideally, I envision one unified `retry config` that sits at `strata-common` which is used everywhere in our repos.

Right now, it will be clunky in one way or another:
- if I use `RetryConfig` from `crates/common` => **then the `paas` crate would be tied to this  Alpen repo**. However, I can see `paas` being useful in bridge as well, so mid term I won't be surprised seeing paas in `strata-common`. So, TLDR: we are making a dependency that in the future we are gonna break anyway it seems.
- If I keep duplicated version, there's an obvious duplication that you pointed out.

I'll look into it a bit more and decide how we progress from nowon

crates/paas/src/registry/core.rs
delbonis: `fetch_input` might be more descriptive since just `get_` implies it's a cheap operation and it matches the underlying call in this impl

crates/paas/src/config.rs
delbonis: Arguably we could move some/all of that common crate to the strata-common repo as a more generic thing.

crates/paas/src/worker.rs
```diff
@@ -0,0 +1,202 @@
...
+}
+
+impl<P: Prover> WorkerPool<P> {
+    /// Create a new worker pool
+    pub fn new(state: Arc<ProverServiceState<P>>) -> Self {
+        Self { state }
+    }
+
+    /// Main worker pool loop
+    pub async fn run(self) {
```
delbonis: Don't like having `run` functions like this as it kinda conflates the "worker state" and supplementary things like input handles.  Not sure if that applies here.

Can this use the service framework?  Or could it use a more general threadpool?

crates/paas/src/worker.rs
```diff
@@ -0,0 +1,202 @@
...
+            for task_id in pending.into_iter().chain(retriable) {
+                self.try_spawn_task(task_id).await;
+            }
+
+            // Sleep before next poll
+            sleep(Duration::from_millis(
+                self.state.config().workers.polling_interval_ms,
+            ))
+            .await;
+        }
```
delbonis: Don't like polling, can this use a channel?

crates/paas/src/task.rs
```diff
@@ -0,0 +1,114 @@
...
+    + Eq
+    + std::hash::Hash
+    + std::fmt::Debug
+    + Send
+    + Sync
+    + Serialize
+    + for<'de> Deserialize<'de>
+    + 'static
+{
+}
```
delbonis: Do we really need to be generic to this degree or can we just use a `u64` or maybe a UUID or something like that?

bin/prover-client/src/service/task.rs
```diff
@@ -0,0 +1,61 @@
...
+/// Routing key for ProofContext - used for dynamic dispatch in Prover Service registry
+///
+/// This enum represents the different variants of ProofContext without carrying
+/// the actual data, allowing the registry to route requests to the correct handler.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum ProofContextVariant {
+    EvmEeStf,
+    ClStf,
+    Checkpoint,
+}
```
delbonis, line 19: The name `ProofContextVariant` feels pretty obtuse to me, but the doc calls this a "routing key".  I get the idea we're trying to capture here but could we pick something more precise?

crates/paas/src/worker.rs
evgenyzdanovich: agree, your reasoning applies here :) addressing it.

bin/prover-client/src/service/adapters.rs
```diff
@@ -0,0 +1,128 @@
...
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        self.0.source()
+    }
+}
+
+impl From<anyhow::Error> for ProofStorageError {
+    fn from(e: anyhow::Error) -> Self {
+        Self(e)
+    }
+}
```
storopoli, line 42: What do we get in benefits by having all this code instead of using either `dyn std::error::Error` or `anyhow:Error` itself?

What am I missing? Maybe we should just use these instead.

EDIT: This is a binary in `bin/`. Just use `anyhow!`.

bin/prover-client/src/service/task_store.rs
```diff
@@ -0,0 +1,189 @@
...
+        }
+    }
+
+    /// Convert from serializable form to paas TaskId
+    fn from_serializable_task_id(ser: &SerializableTaskId) -> TaskId<ProofTask> {
+        let backend = match ser.backend {
+            0 => strata_paas::ZkVmBackend::Native,
+            1 => strata_paas::ZkVmBackend::SP1,
+            2 => strata_paas::ZkVmBackend::Risc0,
+            _ => strata_paas::ZkVmBackend::Native, // Default fallback
```
storopoli, line 44: `Native` should have a `#[default]` in `strata_paas::ZkVmBackend`.
Then you can change this code to something like:

```suggestion
            _ => strata_paas::ZkVmBackend::Default()
```

If the default changes then you don't need to hunt down all these things that fly under the radar in the codebase, and you don't need the `// Default fallback` comment.

crates/paas/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "strata-paas"
+version = "0.1.0"
+edition = "2021"
+
```
storopoli, line 5: add workspace lints!!!

crates/paas/CURRENT_PROBLEMS.md
storopoli, line 1: artifact from claude code

crates/paas/PHASE_1_2_COMPLETE.md
storopoli, line 1: artifact from claude code

crates/paas/PHASE_1_COMPLETE.md
storopoli, line 1: artifact from claude code

crates/paas/REFACTOR_PLAN.md
storopoli, line 1: artifact from claude code

crates/paas/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+/// ZkVm backend identifier
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum ZkVmBackend {
+    /// Native execution (no proving)
+    Native,
+    /// SP1 prover
+    SP1,
+    /// RISC0 prover
+    Risc0,
+}
```
storopoli, line 54: As I said above

```suggestion
/// ZkVm backend identifier
#[derive(Debug, Clone, PartialEq, Eq, Default, Hash, Serialize, Deserialize)]
pub enum ZkVmBackend {
    #[default]
    /// Native execution (no proving)
    Native,

    /// SP1 prover
    SP1,

    /// RISC0 prover
    Risc0,
}
```

crates/paas/src/registry.rs
```diff
@@ -0,0 +1,339 @@
...
+{
+    /// Routing key type (usually an enum discriminant)
+    type RoutingKey: Eq + Hash + Clone + Send + Sync + std::fmt::Debug + 'static;
+
+    /// Extract routing key for handler lookup
+    fn routing_key(&self) -> Self::RoutingKey;
+}
+
+/// Type-erased input container
+pub struct BoxedInput(pub Box<dyn Any + Send + Sync>);
```
storopoli: I really hate this being a `pub` in this crate.
It belongs to `strata-common` or something like this.

docs/paas-communication-analysis.md
storopoli, line 1: artifact from claude code

docs/paas-runtime-handle-refactor.md
storopoli, line 1: artifact from claude code

docs/service-framework-execution-analysis.md
storopoli, line 1: artifact from claude code

docs/worker-async-execution-analysis.md
storopoli, line 1: artifact from claude code

docs/worker-proving-methods-analysis.md
storopoli, line 1: artifact from claude code

Cargo.toml
```diff
@@ -256,8 +258,8 @@ moho-runtime-interface = { git = "https://github.com/alpenlabs/moho", branch = "
 moho-types = { git = "https://github.com/alpenlabs/moho", branch = "STR-1797-container-mmr" }
 
 # zkaleido
-zkaleido = { git = "https://github.com/alpenlabs/zkaleido", tag = "v0.1.0-alpha-rc17" }
-zkaleido-native-adapter = { git = "https://github.com/alpenlabs/zkaleido", tag = "v0.1.0-alpha-rc17" }
+zkaleido = { git = "https://github.com/alpenlabs/zkaleido", tag = "v0.1.0-alpha-rc18" }
+zkaleido-native-adapter = { git = "https://github.com/alpenlabs/zkaleido", tag = "v0.1.0-alpha-rc18" }
```
storopoli, line 267: We should try to update `zkaleido` in a separate PR that is trivial to merge/review etc.

storopoli (changes requested):
I am not against using Claude Code for big tasks quite the opposite, I think we should. But please take a look at what you're committing. There are several Claude Code generated artifacts that do not belong in the codebase.

As a side note, we should have a maintainable repo-wide `CLAUDE.md` (or `AGENTS.md`).

docs/service-framework-execution-analysis.md
evgenyzdanovich: this was intentionally converted to a draft and currently not in a reviewable state, those files hold necessary context. I'll clean those once it's in review again

evgenyzdanovich:
@delbonis could you please take another look, esp on `crates/paas` ?

Cargo.toml
evgenyzdanovich: Well, I published and pushed a necessary (mostly ergonomics) tag in zkaleido required for this PR... we were missing some blanket impls in zkaleido that made me annoyed (because it forced me to add unnecessary boilerplate).

I understand we could have made it more reviewable with regards to potentially multiple PRs, but it is what it is, sorry for that :)

crates/paas/src/config.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+}
+
+/// Worker pool configuration
+#[derive(Debug, Clone)]
+pub struct WorkerConfig<B: Clone + Eq + std::hash::Hash> {
+    /// Number of concurrent tasks per backend (semaphore capacity)
+    pub worker_count: HashMap<B, usize>,
+}
```
delbonis, line 42: Unclear what `B` represents here, I *assume* it's some kind of task ID, but I'm not sure.  Should we define a trait for this that captures the concept more precisely?  Or like my prior comment, if it's just a task ID does it need to be generic?

crates/paas/src/error.rs
```diff
@@ -0,0 +1,94 @@
...
+    #[error("Worker pool error: {0}")]
+    WorkerPool(String),
+
+    /// Configuration error
+    #[error("Configuration error: {0}")]
+    Config(String),
+
+    /// Internal error
+    #[error("Internal error: {0}")]
+    Internal(#[from] anyhow::Error),
```
delbonis, line 39: Redundant to use `error` in error messages.

Also is there some slightly more clever `From`/`Into` impls we can do in the `anyhow::Error` case to keep the context propagation linked together nicer?

crates/paas/src/error.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    /// Permanent failure that should not be retried
+    #[error("Permanent failure: {0}")]
+    PermanentFailure(String),
+
+    /// Invalid state transition
+    #[error("Invalid state transition from {from:?} to {to:?}")]
+    InvalidTransition {
+        from: crate::task::TaskStatus,
+        to: crate::task::TaskStatus,
```
delbonis: Import this at the top.

crates/paas/src/handler/remote.rs
```diff
@@ -0,0 +1,228 @@
...
+    executor: TaskExecutor,
+    _phantom: std::marker::PhantomData<(P, Prog)>,
+}
+
+impl<P, F, S, R, Prog> std::fmt::Debug for RemoteProofHandler<P, F, S, R, Prog> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("RemoteProofHandler")
+            .field("fetcher", &std::any::type_name::<F>())
+            .field("storer", &std::any::type_name::<S>())
+            .field("resolver", &std::any::type_name::<R>())
```
delbonis, line 78: ok this is pretty clever

crates/paas/src/handler/remote.rs
```diff
@@ -0,0 +1,228 @@
...
+where
+    P: ProgramType + Send + Sync + 'static,
+    F: InputFetcher<P> + Clone + Send + Sync + 'static,
+    F::Input: Into<Prog::Input> + 'static,
+    F::Error: 'static,
+    S: ProofStorer<P> + Clone + Send + Sync + 'static,
+    S::Error: 'static,
+    R: HostResolver<P> + Clone + Send + Sync + 'static,
+    Prog: ZkVmProgram + ZkVmRemoteProgram + Send + Sync + 'static,
+    Prog::Input: Send + Sync + 'static,
```
delbonis, line 121: Might be able to consolidate the `Clone + Sync + Send + 'static` bounds onto the traits being referenced here so we don't have to repeat them in every context like this.

crates/paas/src/handler/traits.rs
```diff
@@ -0,0 +1,213 @@
...
+// ============================================================================
+
+/// Type-erased input container for dynamic dispatch
+///
+/// Since `ProofHandler::fetch_input()` needs to work with any input type,
+/// we use type erasure via `Box<dyn Any>`. The handler can then downcast
+/// to the concrete type when executing the proof.
+///
+/// This allows PaaS to remain generic while handlers work with specific types.
+pub struct BoxedInput(pub Box<dyn Any + Send + Sync>);
```
delbonis: Maybe can hide the inner field here especially since we expose the `.downcast` anyways?

crates/paas/src/persistence.rs
```diff
@@ -0,0 +1,54 @@
...
+#[derive(Debug, Clone)]
+pub struct TaskRecord<T>
+where
+    T: Clone + Eq + std::hash::Hash + std::fmt::Debug + Send + Sync + 'static,
+{
+    pub task_id: T,
+    pub uuid: String,
+    pub status: TaskStatus,
+    pub created_at: Instant,
+    pub updated_at: Instant,
```
delbonis: Can probably make these private?

delbonis: Can probably make these private?

Similar comment as previously, this is a *lot* of type bounds to carry around that we could probably consolidate, *or* just make task IDs not generic.

crates/paas/src/persistence.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+/// Trait for persistent task storage
+///
+/// Implementations should be database-backed for production use.
+/// This trait enables idempotent task submission and crash recovery.
+pub trait TaskStore<P: ProgramType>: Send + Sync + 'static {
```
delbonis, line 79: Just want to sanity check that this gets called only in blocking-safe contexts so we don't block the executor.

Edit: :grimacing: I think it might not be! (`ProverService`) It'd probably be really awkward to use the `inst_ops` interface on this because the types are generic, but also it'd be awkward to make that a sync service.  Not sure the best approach here.

crates/paas/src/service/builder.rs
```diff
@@ -0,0 +1,194 @@
...
+        executor.spawn_critical_async("paas_retry_scheduler", async move {
+            retry_scheduler.run().await;
+            Ok(())
+        });
+
+        // Launch service
+        let monitor = service_builder
+            .launch_async("prover", executor)
+            .await
+            .map_err(crate::error::ProverServiceError::Internal)?;
```
delbonis: ahhhh `crate::`

crates/paas/src/service/runtime.rs
```diff
@@ -0,0 +1,148 @@
...
+                            let state_clone = state.clone();
+                            let task_id_clone = task_id.clone();
+                            state.executor.handle().spawn(async move {
+                                state_clone.execute_and_track(task_id_clone).await;
+                            });
+                        } else {
+                            debug!(?task_id, %uuid, "Task already exists, returning existing UUID");
+                        }
+
+                        completion.send(uuid).await;
```
delbonis, line 70: Wait so if we do use UUIDs here then why do we have a separate task_id concept?

crates/paas/src/service/runtime.rs
```diff
@@ -0,0 +1,148 @@
...
+        Self {
+            config: self.config.clone(),
+            task_store: self.task_store.clone(),
+            handlers: self.handlers.clone(),
+            semaphores: self.semaphores.clone(),
+            executor: self.executor.clone(),
+            retry_scheduler: self.retry_scheduler.clone(),
+        }
+    }
+}
```
delbonis: Uhhhh should this be in the `state` module?

crates/paas/src/service/state.rs
```diff
@@ -0,0 +1,581 @@
...
+
+        // Transition to Queued
+        if let Err(e) = self.update_status(&task_id, TaskStatus::Queued) {
+            error!(?task_id, ?e, "Failed to update status to Queued");
+            return TaskResult::Failed {
+                uuid: uuid.clone(),
+                error: format!("Status update failed: {}", e),
+            };
+        }
+        info!(?task_id, %uuid, "Task queued");
```
delbonis: There's a *lot* of logs here (and throughout the whole service) that have repeated parameters.  We can use a tracing span to group all of these together and make the logging more structure and not need to repeat the `?task_id, %uuid` on every log.  (Just make sure to use `.instrument` with the span on a future so it gets tracked properly.)

Also, I think some of these probably don't need to be `info!` but could be `debug!` or even `trace!`.

crates/paas/src/service/state.rs
```diff
@@ -0,0 +1,581 @@
...
+
+        // Transition to Proving
+        if let Err(e) = self.update_status(&task_id, TaskStatus::Proving) {
+            error!(?task_id, ?e, "Failed to update status to Proving");
+            return TaskResult::Failed {
+                uuid: uuid.clone(),
+                error: format!("Status update failed: {}", e),
+            };
```
delbonis, line 323: Could add a constructor to `TaskResult` for this to make it a little more ergonomic.  Also does `Uuid` not implement `Copy`?

delbonis (changes requested):
This is really sophisticated and I'm really happy with what's being done here.  However there's also a lot of code so I have a lot of comments on a few design choices.

crates/paas/src/config.rs
storopoli: Of course it's Bitcoin, come on it's @evgenyzdanovich

crates/paas/src/service/runtime.rs
evgenyzdanovich: well, thanks for asking!!! Hopefully you can advise here, on the naming as well :) 

 1. `UUID` is for submitting and polling on the proving task via handler (via commands::SubmitTask + commands::GetStatusByUuid).
 2. `task_id` is basically a unique `input` for a certain zk program against certain host that identifies the prove. Think of `task_id` like Checkpoint(13), or EvmEeEst(start_slot,end_slot). This concept is needed for ergonomic work with potentially several registered programs, so things like `prover_service_handle.submit_task(EmvEeStf::new(1,2))` are possible. Also, it's needed for idempotency and some storage.
 
There is two way mapping between `uuid` and `task_id` and user of the crate can use it anyhow they want (either via uuid with polling or with task_id "directly"). I'm not sure what API is more useful to the user of the crate (my best guess is both) and would appreciate opinion.

My opinion is I would keep both, but certainly rename and/or leave explicit documentation. 

+CC @prajwolrg

prajwolrg: I like the two ids approach for reasons pointed out.

crates/paas/src/error.rs
evgenyzdanovich: done.

Hm, I'm also curious, but unaware of such

crates/paas/src/config.rs
evgenyzdanovich: it stands for "Backend". done.

crates/paas/src/handler/remote.rs
evgenyzdanovich: somewhat prettified it, done

crates/paas/src/handler/traits.rs
evgenyzdanovich: done

crates/paas/src/service/runtime.rs
evgenyzdanovich: done

crates/paas/src/service/state.rs
evgenyzdanovich: yep, done, thanks!

crates/paas/src/service/state.rs
evgenyzdanovich: done

crates/paas/src/persistence.rs
evgenyzdanovich: done

crates/paas/src/persistence.rs
evgenyzdanovich: I'm not sure I see the problem, can you please elaborate? What's exactly blocking?

for now (as a least minimal measure) I'm leaving it as a TODO to handle it soon later.

storopoli (dismissed):
ACK 7bac33e57ea473bfb7b854544e0c8d34bef696d7

prajwolrg (dismissed):
LGTM!

crates/paas/src/persistence.rs
delbonis: The database read (if not cached) suspend the thread when the kernel is waiting for the disk and then the async executor thread won't be able to run other tasks.

crates/paas/src/service/runtime.rs
delbonis: ah my head hurts

crates/paas/src/persistence.rs
evgenyzdanovich: Thanks. I'm more curious what's different between the usage of db in this PR versus how we **already** access db in other services in async context (for example ASM or CSM) and what should be our best practice here. Maybe the problem is more systematic...

I'll hop on your office hours and would be glad to discuss.

delbonis (dismissed):

crates/paas/src/persistence.rs
evgenyzdanovich: that's also relevant in the context of aux data resolver in asm (mmr db for instance) as there certainly will be db interactions.

evgenyzdanovich: +CC @prajwolrg

storopoli: Solving this to unblock the merge queue

PR #1150, opened 2025-11-17
Update withdrawal fulfillment tx and OperatorClaimUnlock

prajwolrg:
## Description

This PR makes two changes:

#### 1.  Reduce the auxiliary data from 40 bytes to 4 bytes for withdrawal fulfillment transaction.

We now include only the deposit index (u32, big-endian) in the OP_RETURN payload.

Removed:
* Operator index (4 bytes)
* Deposit transaction ID (32 bytes)

These fields were unnecessary: the operator index is implied because only the assigned operator makes the front payment, and the deposit TXID does not affect validation. The fulfillment transaction is now considered valid as long as the right amount is fulfilled to the right user within the deadline.

This change reduces payload size and simplifies validation logic.

#### Remove TxIds from OperatorClaimUnlock
Bridge proof and counterproof only requires `deposit_idx` and `operator_idx`. So remove unrequired TxIds.

### Type of Change

- [x] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/asm/txs/bridge-v1/src/withdrawal_fulfillment/parse.rs
+++ b/crates/asm/txs/bridge-v1/src/withdrawal_fulfillment/parse.rs
@@ -82,7 +82,7 @@ pub fn parse_withdrawal_fulfillment_tx<'t>(
     }
 
     let mut decoder = BufDecoder::new(tx.tag().aux_data());
-    let withdrawal_auxdata = WithdrawalFulfillmentTxTagData::decode(&mut decoder).unwrap();
+    let withdrawal_auxdata = WithdrawalFulfillmentTxTagData::decode(&mut decoder)?;
 
     let withdrawal_fulfillment_output = &tx
         .tx()
@@ -195,7 +195,7 @@ mod tests {
 
         let mut tx = create_test_withdrawal_fulfillment_tx(&info);
 
-        // Mutate the OP_RETURN output to have shorter aux len
+        // Mutate the OP_RETURN output to have shorter aux len - this should fail
         let aux_data = vec![0u8; WITHDRAWAL_FULFILLMENT_TX_AUX_DATA_LEN - 1];
         let tagged_payload = create_tagged_payload(
             BRIDGE_V1_SUBPROTOCOL_ID,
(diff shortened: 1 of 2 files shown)


crates/asm/txs/bridge-v1/src/withdrawal_fulfillment/parse.rs
```diff
@@ -1,41 +1,28 @@
...
 
 /// Minimum length of auxiliary data for withdrawal fulfillment transactions.
-pub const WITHDRAWAL_FULFILLMENT_TX_AUX_DATA_LEN: usize =
-    OPERATOR_IDX_SIZE + DEPOSIT_IDX_SIZE + DEPOSIT_TXID_SIZE;
+pub const WITHDRAWAL_FULFILLMENT_TX_AUX_DATA_LEN: usize = DEPOSIT_IDX_SIZE;
 
-/// Information extracted from a Bitcoin withdrawal transaction.
+/// Information extracted from a Bitcoin withdrawal fulfillment transaction.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct WithdrawalFulfillmentInfo {
```
delbonis, line 22: Maybe we should restructure this a little bit so that we have a cleaner delineation between the components.  It's making the parsing logic kinda complicated.

Have a `WithdrawalFulfillmentTxTagData` that we can `Codec::decode` from the aux data field on the header directly, and then have a `WithdrawalFullfillmentTxInfo` that contains that and whatever other data we extract from the tx, like the output(s).

prajwolrg: That's a good idea.

crates/asm/txs/bridge-v1/src/withdrawal_fulfillment/tag.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use proptest::prelude::*;
+    use strata_codec::BufDecoder;
+
+    use super::*;
+
+    proptest! {
```
Rajil1213, line 27: Nice!

PR #1153, opened 2025-11-18
feat(bridge-txs): add slash tx

prajwolrg:
## Description

This PR adds slash transaction parsing logic. It also defines the auxiliary data that is expected in the SPS-50 header of the slash transaction.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/asm/txs/bridge-v1/src/constants.rs
+++ b/crates/asm/txs/bridge-v1/src/constants.rs
@@ -9,5 +9,5 @@ pub const BRIDGE_V1_SUBPROTOCOL_ID: SubprotocolId = 2;
 pub const DEPOSIT_REQUEST_TX_TYPE: TxType = 0;
 pub const DEPOSIT_TX_TYPE: TxType = 1;
 pub const WITHDRAWAL_FULFILLMENT_TX_TYPE: TxType = 2;
-pub const SLASH_TX_TYPE: TxType = 4;
 pub const COMMIT_TX_TYPE: TxType = 3;
+pub const SLASH_TX_TYPE: TxType = 4;
--- a/crates/asm/txs/bridge-v1/src/slash/info.rs
+++ b/crates/asm/txs/bridge-v1/src/slash/info.rs
@@ -9,14 +9,14 @@ pub struct SlashInfo {
     /// SPS-50 auxiliary data from the transaction tag.
     header_aux: SlashTxHeaderAux,
     /// Previous outpoint referenced second input (stake connector).
-    second_input_outpoint: BitcoinOutPoint,
+    second_inpoint: BitcoinOutPoint,
 }
 
 impl SlashInfo {
-    pub fn new(header_aux: SlashTxHeaderAux, second_input_outpoint: BitcoinOutPoint) -> Self {
+    pub fn new(header_aux: SlashTxHeaderAux, second_inpoint: BitcoinOutPoint) -> Self {
         Self {
             header_aux,
-            second_input_outpoint,
+            second_inpoint,
         }
     }
 
@@ -25,6 +25,6 @@ impl SlashInfo {
     }
 
     pub fn second_input_outpoint(&self) -> &BitcoinOutPoint {
-        &self.second_input_outpoint
+        &self.second_inpoint
     }
 }
--- a/crates/asm/txs/bridge-v1/src/slash/mod.rs
+++ b/crates/asm/txs/bridge-v1/src/slash/mod.rs
@@ -11,7 +11,7 @@
 //!
 //! ### Inputs
 //! - 1. **Contest slash connector**: Locked to the N-of-N multisig with a relative timelock
-//! - 2. **Stake connector**: Locked to the N-of-N pure multisig, without any timelocks.
+//! - 2. **Stake connector**: Locked to the N-of-N multisig..
 //!
 //! Only the stake connector is validated. The contest connector carries a relative timelock that
 //! is enforced on-chain, but the bridge subprotocol does not store the timelock that is used,


crates/asm/txs/bridge-v1/src/slash/aux.rs
```diff
@@ -0,0 +1,13 @@
+use arbitrary::Arbitrary;
+use strata_bridge_types::OperatorIdx;
+use strata_codec::Codec;
+
+/// Auxiliary data in the SPS-50 header for bridge v1 deposit transactions.
```
Rajil1213:
```suggestion
/// Auxiliary data in the SPS-50 header for bridge v1 transactions.
```

crates/asm/txs/bridge-v1/src/slash/parse.rs
```diff
@@ -0,0 +1,109 @@
...
+///
+/// Parses a slash transaction following the SPS-50 specification and extracts the auxiliary
+/// metadata along with the stake connector outpoint (input index 1).
+///
+/// # Parameters
+/// - `tx` - Reference to the transaction input containing the slash transaction and tag data
+///
+/// # Returns
+/// - `Ok(SlashInfo)` on success
+/// - `Err(SlashTxParseError)` if auxiliary data cannot be decoded, the input count is wrong, or the
```
Rajil1213: What auxiliary data?

crates/asm/txs/bridge-v1/src/constants.rs
```diff
@@ -8,3 +8,4 @@ pub const BRIDGE_V1_SUBPROTOCOL_ID: SubprotocolId = 2;
 
 pub const DEPOSIT_TX_TYPE: TxType = 1;
 pub const WITHDRAWAL_FULFILLMENT_TX_TYPE: TxType = 2;
+pub const SLASH_TX_TYPE: TxType = 4;
```
Rajil1213: All of these should really be defined in a `common` repo or a separate ASM repo so that the bridge and the ASM parsing logic can stay consistent.

crates/asm/txs/bridge-v1/src/errors.rs
```diff
@@ -95,3 +95,17 @@ pub enum WithdrawalParseError {
     #[error("Transaction is missing output that fulfilled user withdrawal request")]
     MissingUserFulfillmentOutput,
 }
+
+/// Errors that can occur when parsing withdrawal fulfillment transactions.
```
Rajil1213:
```suggestion
/// Errors that can occur when parsing slash transaction.
```

Rajil1213 (changes requested):
I only have some minor comments. The only thing I want to flag is whether there are plans to move the asm stuff into a separate repo. There are transaction types defined here that are relevant to the bridge as well and we should keep them consistent. I'd feel better if they were moved to a different repo because the `alpen` repo will evolve faster than the ASM stuff and there is no way to refer to a specific version of the ASM stuff while referring to a different version of the rest of the `strata-*` crates (from the bridge repo).

crates/asm/txs/bridge-v1/src/slash/parse.rs
prajwolrg: `SlashTxHeaderAux`. I'll update the docstring to better reflect that.

Rajil1213 (dismissed):
I'm not sure why the nix builds are failing, but the code looks good to me.

crates/asm/txs/bridge-v1/src/errors.rs
```diff
@@ -95,3 +95,14 @@ pub enum WithdrawalParseError {
     #[error("Transaction is missing output that fulfilled user withdrawal request")]
     MissingUserFulfillmentOutput,
 }
+
+/// Errors that can occur when parsing slash transaction.
+#[derive(Debug, Error)]
+pub enum SlashTxParseError {
+    /// The auxiliary data in the withdrawal fulfillment transaction is invalid
```
barakshani: withdrawal fulfillment transaction?

crates/asm/txs/bridge-v1/src/slash/mod.rs
```diff
@@ -0,0 +1,39 @@
+//! Slash Transaction Parser
+//!
+//! This module provides functionality for parsing Bitcoin slash transactions
+//! that follow the SPS-50 specification for the Strata bridge protocol.
+//!
+//! ## Slash Transaction Structure
+//!
+//! A slash transaction is posted by a watchtower if the operator hasn't posted the
+//! Contested Payout Transaction within the allotted time frame. The transaction spends
+//! the operator stake and distributes it to watchtowers.
```
barakshani:
```suggestion
//! A slash transaction is posted by a watchtower if the claim was contested but the 
//! operator hasn't posted the Contested Payout Transaction within the allotted time 
//! frame. The transaction spends the operator stake and distributes it to watchtowers.
```
just to clarify the optimistic payout does not lead to slashing : )

crates/asm/txs/bridge-v1/src/slash/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+//!
+//! ### Inputs
+//! - 1. **Contest slash connector**: Locked to the N-of-N multisig with a relative timelock
+//! - 2. **Stake connector**: Locked to the pure N-of-N multisig.
```
barakshani: pure?

crates/asm/txs/bridge-v1/src/slash/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+//!
+//! Only the stake connector is validated. The contest connector carries a relative timelock that
+//! is enforced on-chain, but the bridge subprotocol does not store the timelock that is used,
+//! so ASM cannot verify it and skips validation of that input. This is sufficient because the
```
barakshani, line 18: This is sufficient because we only really care if the stake is available or not: practically, the implication of slash is that we no longer assign (and update the bridge address) -- for this we only need to know if the stake is available. it's not different from the unstaking tx.

In other words, if we had a way to query about a specific utxo (and check if it's still not spent), then we may have done that and not really care how it was spent (slash or unstake).

crates/asm/txs/bridge-v1/src/slash/aux.rs
```diff
@@ -0,0 +1,13 @@
...
+use strata_codec::Codec;
+
+/// Auxiliary data in the SPS-50 header for bridge v1 slash transaction.
+///
+/// This represents the type-specific auxiliary bytes that appear after the magic, subprotocol,
+/// and tx_type fields in the OP_RETURN output at position 0.
+#[derive(Debug, Clone, PartialEq, Eq, Arbitrary, Codec)]
+pub struct SlashTxHeaderAux {
+    /// The index of the operator being slashed.
+    pub operator_idx: OperatorIdx,
```
delbonis: Don't make `pub`.

crates/asm/txs/bridge-v1/src/lib.rs
```diff
@@ -9,6 +9,7 @@
 pub mod constants;
 pub mod deposit;
 pub mod errors;
+pub mod slash;
 pub mod withdrawal_fulfillment;
```
delbonis, line 15: should make a note to un-export these in the future

crates/asm/txs/bridge-v1/src/slash/parse.rs
```diff
@@ -0,0 +1,109 @@
...
+/// Index of the stake connector input.
+const STAKE_INPUT_INDEX: usize = 1;
+
+/// Information extracted from a Bitcoin slash transaction.
+#[derive(Debug, Clone, PartialEq, Eq, Arbitrary)]
+pub struct SlashInfo {
+    /// SPS-50 auxiliary data from the transaction tag.
+    pub header_aux: SlashTxHeaderAux,
+    /// Previous outpoint referenced second input (stake connector).
+    pub second_input_outpoint: BitcoinOutPoint,
```
delbonis: No `pub`

storopoli:
> I'm not sure why the nix builds are failing, but the code looks good to me.

It is fixed in 6f3cf4cfac9cba7425802d9cf292706ea3e0d0c6

crates/asm/txs/bridge-v1/src/slash/mod.rs
storopoli: You've mean no timelock?

prajwolrg: Yes. I've tried to clarify it.

crates/asm/txs/bridge-v1/src/errors.rs
prajwolrg: nice catch. it is slash tx.

crates/asm/txs/bridge-v1/src/slash/mod.rs
prajwolrg: I'm not sure if I follow fully here. 

When I said it is sufficient I meant in terms of validating whether the transaction is indeed a slash transaction. i.e. a malicious entity (maybe a malicious operator) cannot simply post a transaction on L1 that resembles slash transaction, which can cause a honest entity to be slashed.

crates/asm/txs/bridge-v1/src/slash/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+//!
+//! ## Slash Transaction Structure
+//!
+//! A slash transaction is posted by a watchtower if the claim was contested but the operator hasn't
+//! posted the Contested Payout Transaction within the allotted time frame. The transaction spends
+//! the operator stake and distributes it to watchtowers.
+//!
+//! ### Inputs
+//! - 1. **Contest slash connector**: Locked to the N-of-N multisig with a relative timelock
+//! - 2. **Stake connector**: Locked to the N-of-N pure multisig, without any timelocks.
```
barakshani: I don't get it, just write N-of-N multisig, why adding what it is not?

crates/asm/txs/bridge-v1/src/slash/mod.rs
barakshani: > transaction is identified by its SPS-50 type

But some other (non-slash) tx can also have an SPS-50 header.

>Only the stake connector is validated

I read that comment as explaining why only validating the stake connector is validated.

crates/asm/txs/bridge-v1/src/constants.rs
```diff
@@ -9,4 +9,5 @@ pub const BRIDGE_V1_SUBPROTOCOL_ID: SubprotocolId = 2;
 pub const DEPOSIT_REQUEST_TX_TYPE: TxType = 0;
 pub const DEPOSIT_TX_TYPE: TxType = 1;
 pub const WITHDRAWAL_FULFILLMENT_TX_TYPE: TxType = 2;
+pub const SLASH_TX_TYPE: TxType = 4;
```
barakshani: [spec](https://www.notion.so/Glock-based-Strata-Bridge-v1-2b4901ba000f80a388e2e9b1fb5ead0c?source=copy_link#2bc901ba000f8098a25de4b7bd846869) says `ff`

Rajil1213: The reason an SPS-50 header was used in the Slash transaction was to maintain uniformity with the rest of the transactions. My understanding then was that the ASM can reason about UTXOs being spent so it does not need to have a proper `tx_type` (using `ff` seemed reasonable because if we increase the `tx_type` numbers one at a time, it'd be some time before we reach `ff`). So, I'm okay with the `tx_type` now being set to `4` but the spec must be updated accordingly.

barakshani: it's stake, not slash. resolving this...

crates/asm/txs/bridge-v1/src/slash/info.rs
```diff
@@ -0,0 +1,30 @@
...
+
+use crate::slash::SlashTxHeaderAux;
+
+/// Information extracted from a Bitcoin slash transaction.
+#[derive(Debug, Clone, PartialEq, Eq, Arbitrary)]
+pub struct SlashInfo {
+    /// SPS-50 auxiliary data from the transaction tag.
+    header_aux: SlashTxHeaderAux,
+    /// Previous outpoint referenced second input (stake connector).
+    second_input_outpoint: BitcoinOutPoint,
```
Rajil1213: In the bridge, we use the term `inpoint` to refer to this (for lack of a better term).

crates/asm/txs/bridge-v1/src/constants.rs
```diff
@@ -9,4 +9,5 @@ pub const BRIDGE_V1_SUBPROTOCOL_ID: SubprotocolId = 2;
...
 pub const COMMIT_TX_TYPE: TxType = 3;
```
Rajil1213, line 12: NIT: move this line up so that these are ordered in ascending order by value.

Rajil1213 (approved):
There are a few nits but otherwise, this looks good.

PS: Actually, I do have a question (which can be taken up somewhere else and not on this PR). I see that the project structure for ASM is: `~/asm/txs/bridge-v1/...`. In the bridge, we're calling the first mainnet release version `v1` but I don't think the `v1` used here is the same as that one. Why is it called bridge-_v1_?

crates/asm/txs/bridge-v1/src/slash/info.rs
prajwolrg: I like it.

prajwolrg:
> There are a few nits but otherwise, this looks good.
> 
> PS: Actually, I do have a question (which can be taken up somewhere else and not on this PR). I see that the project structure for ASM is: `~/asm/txs/bridge-v1/...`. In the bridge, we're calling the first mainnet release version `v1` but I don't think the `v1` used here is the same as that one. Why is it called bridge-_v1_?

It's almost the same thing. IIRC we wanted to keep separate crates in case we update a subprotocol later in time. If there were any modification needed in the subprotocol, we would have an entirely new crate with the updated code and we would call it a different version. CC: @delbonis who suggested the naming scheme in the first place.

storopoli (approved):
ACK 0a3f676f76dca4676c060491dc9a4ef7867adb35

PR #1157, opened 2025-11-21
STR-1987: Thin OL STF

delbonis:
## Description

This PR implements a "thin" implementation of the OL STF spec, as drafted on Notion.  Some variations were made to improve bookkeeping with the context types, which will be used to update the upstream spec.  Some other contrivances were made pending SSZification, where we'd normally want to use the SSZ hash_tree_root.

It includes the high-level block verification and assembly routines, including doing the special casing around the state roots.  There are also supplementary helper types involved in this to aid in block construction.

I also have uncommitted changes that add proper tests for the main STF functionality, along with improving `Codec` impls in simple cases, using the new derive macro.  I'll have this shortly as this is a little messy due to all the different moving parts.

This does NOT have:

* full snark account functionality/verification
* history MMRs (only useful for snark accounts)
* support for transfer outputs from updates
* anything outside of the subset of the upstream spec needed to implement the planned EE functionality

Claude was used for a lot of the annoying mechanical "fill in the blanks" stuff, adding Codec impls, writing tests, etc.  All of the general structure of the primary STF code was implemented by myself by hand.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/identifiers/src/acct.rs
+++ b/crates/identifiers/src/acct.rs
@@ -60,7 +60,7 @@ impl AccountId {
 impl fmt::Display for AccountId {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let mut buf = [0; SUBJ_ID_LEN * 2];
-        hex::encode_to_slice(&self.0, &mut buf).expect("ident/acct: encode hex");
+        hex::encode_to_slice(self.0, &mut buf).expect("ident/acct: encode hex");
         // SAFETY: correct lengths
         f.write_str(unsafe { str::from_utf8_unchecked(&buf) })
     }
@@ -192,7 +192,7 @@ impl_opaque_thin_wrapper!(SubjectId => RawSubjectId);
 impl fmt::Display for SubjectId {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let mut buf = [0; SUBJ_ID_LEN * 2];
-        hex::encode_to_slice(&self.0, &mut buf).expect("ident/subj: encode hex");
+        hex::encode_to_slice(self.0, &mut buf).expect("ident/subj: encode hex");
         // SAFETY: correct lengths
         f.write_str(unsafe { str::from_utf8_unchecked(&buf) })
     }
--- a/crates/ol/stf/src/test_utils.rs
+++ b/crates/ol/stf/src/test_utils.rs
@@ -11,7 +11,10 @@ use strata_ol_state_types::OLState;
 
 use crate::{
     ExecResult,
-    assembly::{BlockComponents, CompletedBlock, ConstructBlockOutput, construct_block, execute_and_complete_block},
+    assembly::{
+        BlockComponents, CompletedBlock, ConstructBlockOutput, construct_block,
+        execute_and_complete_block,
+    },
     context::{BlockContext, BlockInfo},
     errors::ExecError,
     verification::verify_block,
@@ -28,8 +31,8 @@ pub fn execute_block(
     execute_and_complete_block(state, block_context, components)
 }
 
-/// Execute a block and return the construct output which includes both the completed block and execution outputs.
-/// This is useful for tests that need to inspect the logs.
+/// Execute a block and return the construct output which includes both the completed block and
+/// execution outputs. This is useful for tests that need to inspect the logs.
 pub fn execute_block_with_outputs(
     state: &mut OLState,
     block_info: &BlockInfo,
--- a/crates/ol/stf/src/tests_snark.rs
+++ b/crates/ol/stf/src/tests_snark.rs
@@ -186,35 +186,33 @@ fn test_snark_account_deposit_and_withdrawal() {
 
     for log in logs {
         // Check if it's a snark account update log (from the snark account)
-        if log.account_serial() == snark_serial {
-            if let Ok(update_log) =
-                strata_codec::decode_buf_exact::<SnarkAccountUpdateLogData>(&log.payload())
-            {
-                snark_update_found = true;
-                // The update log indicates the snark account was updated
-                assert_eq!(update_log.new_msg_idx(), 0, "Message index should be 0");
-            }
+        if log.account_serial() == snark_serial
+            && let Ok(update_log) =
+                strata_codec::decode_buf_exact::<SnarkAccountUpdateLogData>(log.payload())
+        {
+            snark_update_found = true;
+            // The update log indicates the snark account was updated
+            assert_eq!(update_log.new_msg_idx(), 0, "Message index should be 0");
         }
 
         // Check if it's a withdrawal intent log (from the bridge gateway)
-        if log.account_serial() == BRIDGE_GATEWAY_ACCT_SERIAL {
-            if let Ok(withdrawal_log) =
-                strata_codec::decode_buf_exact::<SimpleWithdrawalIntentLogData>(&log.payload())
-            {
-                withdrawal_found = true;
+        if log.account_serial() == BRIDGE_GATEWAY_ACCT_SERIAL
+            && let Ok(withdrawal_log) =
+                strata_codec::decode_buf_exact::<SimpleWithdrawalIntentLogData>(log.payload())
+        {
+            withdrawal_found = true;
 
-                // Verify the withdrawal details
-                assert_eq!(
-                    withdrawal_log.amt, withdrawal_amount,
-                    "Withdrawal amount should match"
-                );
+            // Verify the withdrawal details
+            assert_eq!(
+                withdrawal_log.amt, withdrawal_amount,
+                "Withdrawal amount should match"
+            );
 
-                assert_eq!(
-                    withdrawal_log.dest.as_slice(),
-                    withdrawal_dest_desc.as_slice(),
-                    "Withdrawal destination should match"
-                );
-            }
+            assert_eq!(
+                withdrawal_log.dest.as_slice(),
+                withdrawal_dest_desc.as_slice(),
+                "Withdrawal destination should match"
+            );
         }
     }
 


delbonis:
This PR looks larger than it really is because there's a lot of types that had a `Codec` impl applied to them.  This was required to make hash commitments used in the STF verification checks work.

crates/ol/chain-types/src/log_payloads.rs
```diff
@@ -0,0 +1,206 @@
...
+        self.new_msg_idx
+    }
+
+    /// Get the extra data as bytes.
+    pub fn extra_data(&self) -> &[u8] {
+        self.extra_data.as_ref()
+    }
+}
+
+#[cfg(test)]
```
bewakes, line 72: Adding proptest instead of manually testing various conditions will cover more cases and will be succinct.

crates/ol/msg-types/src/withdrawal.rs
```diff
@@ -0,0 +1,335 @@
...
+        Ok(Self { fees, dest_desc })
+    }
+}
+
+/// Message data for withdrawal fee bump.
+///
+/// This message type is sent by accounts that want to bump the fee for
+/// a pending withdrawal in the withdrawal intents queue.
+#[derive(Debug, Clone, PartialEq, Eq, Codec)]
+pub struct WithdrawalFeeBumpMsgData {
```
bewakes: Do we need this for mainnet?

crates/ol/stf/src/account_processing.rs
```diff
@@ -0,0 +1,140 @@
...
+        BRIDGE_GATEWAY_ACCT_ID => {
+            handle_bridge_gateway_message(state, sender, msg, context)?;
+        }
+
+        // Any other address we assume is a ledger account, so we have to look it up.
+        _ => {
+            // TODO adapt the account state traits to make this all more
+            // amendable to avoiding copies/clones
+
+            // Make a copy of the account state.  I don't love this.
```
bewakes, line 43: We can possibly get away with this by flattening the accessors/modifiers in `StateAccessor`, which will also prevent the possible "self-send" issue. But it's fine for now and we can discuss that later.

crates/ol/chain-types/src/log_payloads.rs
```diff
@@ -0,0 +1,206 @@
...
+use strata_acct_types::VarVec;
+use strata_codec::{Codec, CodecError, Decoder, Encoder};
+use strata_codec_derive::Codec;
+
+/// Payload for a simple withdrawal intent log.
+///
+/// This log is emitted when a withdrawal intent is created through the bridge.
+/// It contains the withdrawal amount and destination information.
+#[derive(Debug, Clone, PartialEq, Eq, Codec)]
+pub struct SimpleWithdrawalIntentLogData {
```
bewakes, line 11: Do we really need to prefix this as "Simple"? It just seems to obscure why it is a special intent log data. Will there be other kinds of withdrawal intent log data?

delbonis:
Rebased onto main to get SSZ, removed most recent commits so I can take a more realistic approach using that instead.

crates/ol/chain-types/src/block.rs
```diff
@@ -1,9 +1,11 @@
...
-use strata_primitives::buf::{Buf32, Buf64};
+use strata_asm_common::AsmManifest;
+use strata_codec::{Codec, CodecError, Decoder, Encoder, encode_to_vec};
+use strata_codec_derive::Codec;
+use strata_identifiers::{Buf32, Buf64, L1BlockId, hash::raw};
 
 use crate::{Epoch, OLBlockId, OLTransaction, Slot};
 
-/// The Orchestration Layer(OL) block.
+/// Signed full orchestration layer block.
```
barakshani: Why "full"? what's non full block?

crates/ol/chain-types/src/block.rs
```diff
@@ -18,20 +20,21 @@ impl OLBlock {
...
 
-    pub fn body(&self) -> &OLBlockBody {
-        &self.body
-    }
-
     pub fn signed_header(&self) -> &SignedOLBlockHeader {
         &self.signed_header
     }
 
+    /// Returns the actual block header inside the signed header structure.
```
barakshani:
```suggestion
    /// Returns the block header inside the signed header structure.

File: crates/ol/chain-types/src/block.rs [hunk 7]
```diff
@@ -18,20 +20,21 @@ impl OLBlock {
...
         self.signed_header.header()
     }
+
+    pub fn body(&self) -> &OLBlockBody {
+        &self.body
+    }
 }
 
-/// OL Block header with signature.
+/// OL header with signature.
```
barakshani:
```suggestion
/// OL block header with signature.

File: crates/ol/chain-types/src/block.rs [hunk 8]
```diff
@@ -47,13 +50,16 @@ impl SignedOLBlockHeader {
...
-        self.signature
+    /// This MUST be a schnorr signature for now.
+    pub fn signature(&self) -> &Buf64 {
+        &self.signature
     }
 }
 
-/// OL Block header without signature.
-#[derive(Clone, Debug)]
+/// OL header.
```
barakshani:
```suggestion
/// OL block header.

File: crates/ol/chain-types/src/block.rs [hunk 9]
```diff
@@ -134,28 +147,76 @@ pub struct OLBlockBody {
...
+    pub fn new(tx_segment: OLTxSegment, l1_update: Option<OLL1Update>) -> Self {
         Self {
             tx_segment,
             l1_update,
         }
     }
 
-    pub fn txs(&self) -> &[OLTransaction] {
-        self.tx_segment.txs()
+    pub fn new_regular(tx_segment: OLTxSegment) -> Self {
```
barakshani: what's a _regular_?

crates/ol/chain-types/src/block.rs
```diff
@@ -134,28 +147,76 @@ pub struct OLBlockBody {
...
+
+    /// Computes the hash commitment of this block body.
+    pub fn compute_hash_commitment(&self) -> Buf32 {
+        // Encode the block body and hash it
+        let encoded = encode_to_vec(self).expect("block body encoding should succeed");
+        let hash = raw(&encoded);
+        hash
+    }
+
+    /// Checks if the body looks like an epoch terminal.  Ie. if the L1 update is present.
```
barakshani: why 'probably'? what else could it be?

crates/ol/msg-types/src/withdrawal.rs
```diff
@@ -0,0 +1,332 @@
...
+use strata_identifiers::SubjectId;
+
+/// Message type ID for withdrawal initiation.
+pub const WITHDRAWAL_MSG_TYPE_ID: u16 = 0x03;
+
+/// Message type ID for withdrawal fee bump.
+pub const WITHDRAWAL_FEE_BUMP_MSG_TYPE_ID: u16 = 0x04;
+
+/// Message type ID for withdrawal rejection.
+pub const WITHDRAWAL_REJECTION_MSG_TYPE_ID: u16 = 0x05;
```
barakshani, line 15: what are these?

crates/ol/msg-types/src/withdrawal.rs
```diff
@@ -0,0 +1,332 @@
...
+
+/// Message data for withdrawal initiation to the bridge gateway account.
+///
+/// This message type is sent by accounts that want to trigger a withdrawal.
+/// The value sent with the message should be equal to the predetermined
+/// static withdrawal size.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct WithdrawalMsgData {
+    /// Fees in satoshis to be paid to the operator.
+    pub fees: u32,
```
barakshani: do we need this here? we already have this in the bridge subprotocol (right now fees are fixed)

crates/ol/msg-types/src/withdrawal.rs
```diff
@@ -0,0 +1,332 @@
...
+    /// Bitcoin Output Script Descriptor describing the withdrawal output.
+    pub dest_desc: VarVec<u8>,
+}
+
+impl WithdrawalMsgData {
+    /// Create a new withdrawal message data instance.
+    pub fn new(fees: u32, dest_desc: Vec<u8>) -> Option<Self> {
+        // Ensure the destination descriptor isn't too long
```
barakshani: this is for the bitcoin address, right? Some bitcoin addresses, like bech32, have some error correction associated I think, so we may want to add this at some point.

crates/ol/msg-types/src/withdrawal.rs
```diff
@@ -0,0 +1,332 @@
...
+        Ok(Self { fees, dest_desc })
+    }
+}
+
+/// Message data for withdrawal fee bump.
+///
+/// This message type is sent by accounts that want to bump the fee for
+/// a pending withdrawal in the withdrawal intents queue.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct WithdrawalFeeBumpMsgData {
```
barakshani, line 94: we can keep this, but then let's add "Not Supported"

crates/ol/state-types/src/epochal.rs
barakshani, line 1: please clarify the name of this file

crates/ol/state-types/src/epochal.rs
```diff
@@ -0,0 +1,99 @@
...
+use strata_codec::{Codec, CodecError, Decoder, Encoder};
+use strata_identifiers::L1BlockCommitment;
+use strata_ledger_types::*;
+
+#[derive(Clone, Debug)]
+pub struct EpochalState {
+    total_ledger_funds: BitcoinAmount,
+    cur_epoch: u32,
+    last_l1_block: L1BlockCommitment,
+    checkpointed_epoch: EpochCommitment,
```
barakshani, line 15: can you add some doc string that explains for what time these fields are up to date?

I think _total funds_ is for end of previous epoch (epoch `j`), while _current epoch_ is the epoch that has already started (epoch `j+1`), last l1 block commitment corresponds to the last l1 block at the time of the previous epoch (when we sealed epoch `j`). Would be good to have this clarified.

crates/ol/state-types/src/ledger.rs
```diff
@@ -0,0 +1,140 @@
+//! Accounts system types.
+//!
+//! This uses the "transitional" types described in the OL STF spec.
```
barakshani, line 3: "transitional" as this is temporary?

crates/ol/state-types/src/snark_account.rs
```diff
@@ -0,0 +1,127 @@
...
+use strata_codec_derive::Codec;
+use strata_codec_utils::CodecSsz;
+use strata_ledger_types::ISnarkAccountState;
+use strata_merkle::CompactMmr64B32;
+use strata_snark_acct_types::{MessageEntry, Seqno};
+use tree_hash::TreeHash;
+
+#[derive(Clone, Debug)]
+pub struct NativeSnarkAccountState {
+    seqno: Seqno,
```
barakshani, line 12: I don't see this is the spec, what exactly this field represent?

crates/ol/state-types/src/snark_account.rs
```diff
@@ -0,0 +1,127 @@
...
+        let hash = <MessageEntry as TreeHash>::tree_hash_root(&entry);
+        self.inbox_mmr
+            .add_leaf(hash.into_inner())
+            .expect("ol/state: mmr add_leaf");
+        Ok(())
+    }
+}
+
+#[derive(Clone, Debug, Codec)]
+pub struct ProofState {
```
barakshani, line 76: What exactly is this proof state? why do we have here a thing called "next message read" which does not seem related to proofs.

The reason I'm asking is to note that we may have the account sending a few updates, while none of them is proven. I am not sure if this is taken into account -- like is this the last proven state or something?

crates/ol/stf/src/account_processing.rs
```diff
@@ -0,0 +1,140 @@
...
+    };
+
+    if msg.ty() != WITHDRAWAL_MSG_TYPE_ID {
+        // Some other message type, just ignore.
+        return Ok(());
+    }
+
+    let Some(withdrawal_data) = msg.try_as_withdrawal() else {
+        // Invalid withdrawal message, just ignore.
+        return Ok(());
```
barakshani: you should be able to combine these two, you do the same check twice

crates/ol/stf/src/account_processing.rs
```diff
@@ -0,0 +1,140 @@
...
+    };
+
+    // Check if the withdrawal amount is in allowed denominations
+    let withdrawal_amt = payload.value();
+
+    // TODO move to params struct
+    let withdrawal_denoms = &[100_000_000];
```
barakshani, line 102: let's do it now, I don't like this being hardcoded

barakshani: (or at least open a tech debt ticket, put in the next cleanup sprint)

crates/ol/stf/src/account_processing.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    // Make that the amount is an appropriate denomination.
+    if !withdrawal_denoms.contains(&withdrawal_amt.into()) {
+        return Ok(());
+    }
+
+    // If it is, then we can emit a OL log with the amount and destination.
+    let log_data = SimpleWithdrawalIntentLogData {
+        amt: withdrawal_amt.into(),
+        dest: withdrawal_data.dest_desc.clone(),
```
barakshani: shouldn't we also check correctness of destination address (I think that now we only check address size)?

crates/ol/stf/src/assembly.rs
```diff
@@ -0,0 +1,266 @@
...
+    pub fn logs(&self) -> &[OLLog] {
+        &self.logs
+    }
+
+    /// Computes the block's logs root from the log.
+    pub fn compute_block_logs_root(&self) -> Buf32 {
+        // This is just a simple binary merkle tree.
+        if self.logs.is_empty() {
+            // Empty tree has null root
+            return Buf32::zero();
```
barakshani, line 56: FWIW, ethereum, _I think_, does not return zero in this case. It may be hash(0) or perhaps some minimal size root of all-0 tree. We can leave it be, but if the code below is generic enough, then we may want to discard this special case.

crates/ol/stf/src/context.rs
```diff
@@ -0,0 +1,277 @@
...
+        OLBlockCommitment::new(ph.slot(), blkid)
+    }
+
+    /// Checks if we're probably at the first block of an epoch.
+    ///
+    /// This checks if the epoch is greater than the previous header's epoch, or if we have no
+    /// parent block (meaning we're probably the genesis block).
+    ///
+    /// This doesn't behave correctly if the headers are not actually
+    /// consistent.
```
barakshani: what do you mean by that? it sounds like this can happen in a correct flow.

crates/ol/stf/src/chain_processing.rs
```diff
@@ -0,0 +1,59 @@
...
+    let state_cur_epoch = estate.cur_epoch();
+    let block_cur_epoch = context.cur_epoch() as u32;
+
+    // Special case for genesis block: both state and block are at epoch 0
+    if state_cur_epoch == 0 && block_cur_epoch == 0 {
+        // Genesis block - no epoch increment needed
+    } else {
+        // Regular epoch transition: block should be state_epoch + 1
+        let state_next_epoch = state_cur_epoch + 1;
+        if block_cur_epoch != state_next_epoch {
```
barakshani: just curios, why don't you do these two lines in one?

crates/ol/stf/src/chain_processing.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    // TODO insert into MMR
+
+    Ok(())
+}
+
+/// Processing that happens at the start of every slot (block).
+///
+/// This updates the global state to track the current slot number.
+pub fn process_slot_start<S: StateAccessor>(
```
barakshani: do we actually only call it for every slot, or only if there's a block (via `execute_block_inputs`)?

crates/ol/stf/src/assembly.rs
```diff
@@ -0,0 +1,266 @@
...
+    )?;
+
+    // 4. Compute the state root and remember it.
+    let pre_manifest_state_root = state.compute_state_root()?;
+
+    // 5. If it's the last block of an epoch, then call process_block_manifests,
+    // and compute the final state root and remember it.
+    //
+    // Then we use this to figure out what our state commitments should be.
+    let post_state_roots = if let Some(manifest_container) = block_exec_input.manifest_container() {
```
barakshani, line 120: we have `is_probably_terminal()` which we don't seem to use, though it seems to be defined on block, rather than _context_.

crates/ol/stf/src/manifest_processing.rs
```diff
@@ -0,0 +1,146 @@
...
+
+fn process_asm_log<S: StateAccessor>(
+    state: &mut S,
+    log: &AsmLogEntry,
+    real_height: L1Height,
+    context: &BasicExecContext<'_>,
+) -> ExecResult<()> {
+    // Try to parse the log as an SPS-52 message.
+    let Some(msg) = log.try_as_msg() else {
+        // Not a valid message format, skip it.
```
barakshani, line 77: ASM has already did filtering, right? so this case should never happen? let's have a debug log for the case it does happen

crates/ol/stf/src/manifest_processing.rs
```diff
@@ -0,0 +1,146 @@
...
+        return Ok(());
+    };
+
+    // Match on the type ID to determine how to process the log.
+    match msg.ty() {
+        strata_asm_manifest_types::DEPOSIT_INTENT_ASM_LOG_TYPE_ID => {
+            // Parse the deposit intent data, skip if it fails to parse.
+            let Ok(data) = log.try_into_log::<DepositIntentLogData>() else {
+                return Ok(());
```
barakshani, line 86: as above

crates/ol/stf/src/manifest_processing.rs
```diff
@@ -0,0 +1,146 @@
...
+            };
+            process_deposit_intent_log(state, &data, context)?;
+        }
+
+        strata_asm_manifest_types::CHECKPOINT_ACK_ASM_LOG_TYPE_ID => {
+            // Parse the checkpoint acknowledgment data, skip if it fails to parse.
+            let Ok(data) = log.try_into_log::<CheckpointAckLogData>() else {
+                return Ok(());
```
barakshani, line 94: as above

crates/ol/stf/src/transaction_processing.rs
```diff
@@ -0,0 +1,202 @@
...
+    target: &AccountId,
+    mut astate: S::AccountState,
+    update: &SnarkAccountUpdateContainer,
+    context: &TxExecContext<'_>,
+) -> ExecResult<()> {
+    let AccountTypeState::Snark(mut sastate) = astate.get_type_state()? else {
+        return Err(ExecError::IncorrectTxTargetType);
+    };
+
+    // 2. Call the snark account machinery to process the update.
```
barakshani, line 113: where's 1?

crates/ol/stf/src/verification.rs
```diff
@@ -0,0 +1,510 @@
...
+        self.manifest_container
+    }
+
+    /// Checks if the body implied by this input is implied to be a terminal block.
+    pub fn is_body_terminal(&self) -> bool {
+        self.manifest_container().is_some()
+    }
+}
+
+/// Verifies a block classically by executing it the normal way.
```
barakshani: what would be the other, abnormal, way?

crates/ol/stf/src/verification.rs
```diff
@@ -0,0 +1,510 @@
...
+
+    // 3. Call process_block_tx_segment for every block as usual.
+    let output_buffer = ExecOutputBuffer::new_empty();
+    let basic_ctx = BasicExecContext::new(block_info, &output_buffer);
+    let tx_ctx = TxExecContext::new(&basic_ctx, parent_header.as_ref());
+    transaction_processing::process_block_tx_segment(state, body.tx_segment(), &tx_ctx)?;
+
+    // 4. Check the state root.
+    // - if it's a nonterminal, then check against the header state root
+    // - if it *is* a terminal, then check against the preseal state root
```
barakshani, line 175: nit: you emphasize _is_ here, but also one line above you have "is" (it's). You may want to emphasize _terminal_ instead.

crates/ol/stf/src/verification.rs
```diff
@@ -0,0 +1,510 @@
...
+
+/// Verifies a block classically by executing it the normal way.
+///
+/// This closely aligns with `execute_block_inputs`.
+pub fn verify_block_classically<S: StateAccessor>(
+    state: &mut S,
+    header: &OLBlockHeader,
+    parent_header: Option<OLBlockHeader>,
+    body: &OLBlockBody,
+    exp: &BlockExecExpectations,
```
barakshani: isn't this part of the header?

crates/ol/stf/src/verification.rs
```diff
@@ -0,0 +1,510 @@
...
+    let computed_logs_root = compute_logs_root(&output_buffer.into_logs());
+    if computed_logs_root != exp.logs_root {
+        return Err(ExecError::ChainIntegrity);
+    }
+
+    Ok(())
+}
+
+/// Checks that headers are properly continuous.
+pub fn verify_header_continuity(
```
barakshani, line 223: verify header chain?

crates/ol/stf/src/verification.rs
```diff
@@ -0,0 +1,510 @@
...
+    header: &OLBlockHeader,
+    parent: Option<&OLBlockHeader>,
+) -> ExecResult<()> {
+    // Check parent linkages.
+    if let Some(ph) = parent {
+        // Simply check that the parent block makes sense.
+        let pblkid = ph.compute_blkid();
+        if *header.parent_blkid() != pblkid {
+            return Err(ExecError::BlockParentMismatch);
+        }
```
barakshani, line 233: Wait, how do we even get here? We get to this function from `verify_block_classically` which is passed a block (header and body) and a parent header -- what is the parent header? Isn't it the header of _the block (id) referenced in the block's header_? We already got this header from somewhere, say our db, so we already checked that the header and the block id makes sense.

If the purpose is merely a sanity check, we can have it, but let's make sure to write it is for sanity.

crates/ol/stf/src/verification.rs
```diff
@@ -0,0 +1,510 @@
...
+        if !((header.epoch() == ph.epoch()) || (header.epoch() == ph.epoch() + 1)) {
+            return Err(ExecError::SkipEpochs(ph.epoch(), header.epoch()));
+        }
+
+        // Check slots go in order.
+        //
+        // We're writing this in a weird way to make it easier to handle
+        // nonmonotonic slots in future consensus algos.
+        let slot_diff = header.slot() as i64 - ph.slot() as i64;
+        if slot_diff != 1 {
```
barakshani, line 253: we may have empty slots, no?

crates/ol/stf/src/verification.rs
```diff
@@ -0,0 +1,510 @@
...
+        leaf_hashes.push([0u8; 32]);
+    }
+
+    let tree = BinaryMerkleTree::from_leaves::<Sha256Hasher>(leaf_hashes)
+        .expect("power of two leaves should always succeed");
+
+    Buf32(*tree.root())
+}
+
+/// Expectations we have about a epoch's processing.
```
barakshani, line 318:
```suggestion
/// Expectations we have about an epoch's processing.

File: crates/ol/stf/src/output.rs [hunk 38]
```diff
@@ -0,0 +1,45 @@
...
+use std::cell::RefCell;
+
+use strata_ol_chain_types_new::OLLog;
+
+/// Collector for outputs that we can pass around between different contexts.
+#[derive(Clone, Debug)]
+pub struct ExecOutputBuffer {
+    // maybe we'll have stuff other than logs in the future
+    // TODO don't use refcell, this sucks
+    logs: RefCell<Vec<OLLog>>,
```
bewakes, line 12: It doesn't need to have refcell does it? It can exist as a normal mutable structure. It is being used in the `BasicExecContext` so that should instead be refcell.

crates/ol/stf/src/context.rs
```diff
@@ -0,0 +1,277 @@
...
+    pub fn is_probably_epoch_initial(&self) -> bool {
+        self.parent_header()
+            .is_none_or(|ph| self.epoch() > ph.epoch())
```
bewakes: Probably should check explicit increment by 1.

crates/ol/stf/src/test_utils.rs
```diff
@@ -0,0 +1,261 @@
...
+        // Non-terminal block has only final state root
+        BlockPostStateCommitments::Common(block.header().state_root().clone())
+    };
+
+    BlockExecExpectations::new(post_state_roots, block.header().logs_root().clone())
+}
+
+/// Verify a block using verify_block_classically and assert it succeeds.
+pub fn verify_block<S: StateAccessor>(
+    state: &mut S,
```
bewakes: The verification function is taking the mutable state. Does this mean it is the responsibility of caller to revert the state if this fails?

crates/acct-types/src/mmr.rs
```diff
@@ -1,9 +1,7 @@
 //! Concrete orchestration layer MMR types.
+// TODO move this to its own crate, why is it in acct-types?
```
storopoli, line 2: Yeah agreed.

crates/identifiers/src/acct.rs
```diff
@@ -13,6 +13,12 @@ pub struct AccountId(RawAccountId);
 
 impl_opaque_thin_wrapper!(AccountId => RawAccountId);
 
+impl AccountId {
+    pub fn zero() -> Self {
```
storopoli: come on I know you don't like this, but this could be a `const fn` no?

crates/ol/state-types/src/account.rs
```diff
@@ -0,0 +1,109 @@
+use strata_acct_types::{
+    AccountSerial, AccountTypeId, AcctResult, BitcoinAmount, RawAccountTypeId,
+};
+use strata_ledger_types::{AccountTypeState, Coin, IAccountState};
+
+use crate::snark_account::NativeSnarkAccountState;
+
+#[derive(Clone, Debug)]
```
storopoli: Shouldn't we do `Eq` and `PartialEq`?

crates/ol/state-types/src/account.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    fn take_balance(&mut self, amt: BitcoinAmount) -> AcctResult<Coin> {
+        self.balance = self
+            .balance
+            .checked_sub(amt)
+            .expect("ledger: underflow balance");
+        Ok(Coin::new_unchecked(amt))
+    }
+
+    // TODO should we remove this?  what value is it even still providing?
```
storopoli, line 62: Do we need this for zkVMs? If no, then just drop it.

crates/ol/state-types/src/account.rs
```diff
@@ -0,0 +1,109 @@
...
+        self.state = NativeAccountTypeState::from_generic(state);
+        Ok(())
+    }
+}
+
+/// Internal impl of account state types.
+#[derive(Clone, Debug)]
+pub(crate) enum NativeAccountTypeState {
+    Empty,
+    Snark(NativeSnarkAccountState),
```
storopoli, line 90: Can we add docstrings to these variants? That would be really helpful for future reference.

crates/ol/state-types/src/epochal.rs
storopoli: Why not `epoch-state.rs` or just `epoch.rs`?

crates/ol/state-types/src/epochal.rs
```diff
@@ -0,0 +1,57 @@
...
+        self.cur_epoch = epoch;
+    }
+
+    fn last_l1_blkid(&self) -> &L1BlockId {
+        self.last_l1_block.blkid()
+    }
+
+    fn last_l1_height(&self) -> L1Height {
+        // FIXME this conversion is weird
+        self.last_l1_block.height_u64() as u32
```
storopoli, line 50: And a source of future bugs. We should probably change `last_l1_block` to have a `height()` method that returns a `u32`.

crates/ol/state-types/src/epochal.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    fn append_manifest(&mut self, height: L1Height, mf: AsmManifest) {
+        // TODO actually insert into mmr
+        // FIXME make this conversion less weird
+        self.last_l1_block = L1BlockCommitment::from_height_u64(height as u64, *mf.blkid())
+            .expect("state: weird conversion")
```
storopoli, line 57: Same as above `L1BlockCommitment` should have a `from_height` that takes a `u32`. Although this conversion is not as problematic as the one above since we are expanding a `u32` into a `u64` and the opposite (above) is no-no.

crates/ol/state-types/src/ledger.rs
storopoli: That is transitory, like impermanent. I am almost confident that this means that it is related to transition, as in the `T` in `STF`.

crates/ol/state-types/src/snark_account.rs
storopoli: It is there
CleanShot 2025-11-24 at 14 29 22@2x
See: https://www.notion.so/SPS-ol-stf-Orchestration-layer-STF-v1-268901ba000f80b68179dc2e4b83fba0?source=copy_link#268901ba000f803e9bd0c32e13d5b26e

This is similar to EVM's nonces. Trey does not like the term nonce and prefer sequence number. Am I right, @delbonis?

crates/snark-acct-types/src/state.rs
```diff
@@ -17,8 +17,12 @@ pub struct Seqno(RawSeqno);
 impl_opaque_thin_wrapper!(Seqno => RawSeqno);
 
 impl Seqno {
+    pub fn zero() -> Self {
+        Self(0)
+    }
+
```
storopoli: this can also be `const fn` :)

crates/identifiers/src/epoch.rs
```diff
@@ -103,15 +104,14 @@ impl fmt::Display for EpochCommitment {
             .expect("Failed to encode first 2 bytes to hex");
         hex::encode_to_slice(last_2, &mut last_hex).expect("Failed to encode last 2 bytes to hex");
 
+        // SAFETY: we made sure of it
```
storopoli: maybe

```suggestion
        // SAFETY: hex strings are always UTF-8 encoded and there's no way that
        //         `BlkId` has non-hex encoded strings.
```

crates/ol/chain-types/src/block.rs
```diff
@@ -47,12 +48,15 @@ impl SignedOLBlockHeader {
         &self.header
     }
 
+    /// This MUST be a schnorr signature for now.
```
storopoli:
```
    /// A signature that validates an [`OLBlockHeader`].
    ///
    /// For now, must be a 64-byte Schnorr BIP-340 signature.
```

crates/ol/chain-types/src/block.rs
```diff
@@ -134,28 +138,33 @@ pub struct OLBlockBody {
     tx_segment: OLTxSegment,
 
     /// Updates from L1.
-    l1_update: L1Update,
+    l1_update: Option<L1Update>,
```
storopoli: This means that not all OL blocks will have `L1Update`s?
Then, maybe update the docstring please?

crates/ol/chain-types/src/block.rs
```diff
@@ -134,28 +138,33 @@ pub struct OLBlockBody {
...
 
-    pub fn txs(&self) -> &[OLTransaction] {
-        self.tx_segment.txs()
+    pub fn new_regular(tx_segment: OLTxSegment) -> Self {
+        Self::new(tx_segment, None)
     }
 
-    pub fn l1_update(&self) -> &L1Update {
-        &self.l1_update
+    // TODO convert to builder?
```
storopoli, line 204: No, we don't need the builder pattern, unless we have a bunch of `set` methods that could be optional or required. I don't think we have acquired critical mass here.

crates/ol/chain-types/src/block.rs
```diff
@@ -210,32 +206,32 @@ impl L1Update {
...
-        self.new_l1_blkid
+    /// If there are new manifests (which there should be), returns the blkid of
+    /// the last one.  This is the blkid that we use as the new L1 chain tip.
+    pub fn new_l1_blkid(&self) -> Option<&Buf32> {
+        self.manifests.last().map(|mf| mf.l1blkid())
     }
 }
 
 /// A manifest containing ASM data corresponding to a L1 block.
+// TODO maybe convert to using exported types from ASM crates?
```
storopoli: Yes, that is preferred but not blocking.

crates/asm/manifest-types/src/payloads.rs
```diff
@@ -0,0 +1,98 @@
...
+
+use strata_codec::{Codec, CodecError, Decoder, Encoder};
+use strata_identifiers::{AccountSerial, EpochCommitment, SubjectId};
+use strata_msg_fmt::TypeId;
+
+use crate::AsmLog;
+
+// Define type IDs for these log types
+pub const DEPOSIT_INTENT_ASM_LOG_TYPE_ID: TypeId = 100; // TODO: Confirm the actual type ID
+pub const CHECKPOINT_ACK_ASM_LOG_TYPE_ID: TypeId = 101; // TODO: Confirm the actual type ID
```
storopoli, line 15: Cannot find these in https://www.notion.so/SPS-63-OL-checkpointing-ASM-subprotocol-v1-201901ba000f80fba6bce24f77dbac4e
We should update that spec, right?

crates/ol/da/src/traits.rs
```diff
@@ -0,0 +1,15 @@
...
+use crate::errors::DaResult;
+
+/// Defines a DA scheme for some state accessor.
+pub trait DaScheme<S: StateAccessor> {
+    /// The encoded diff structure.
+    type Diff: Codec;
+
+    /// Applies a diff to the state.
+    fn apply_to_state(diff: Self::Diff, acc: &mut S) -> DaResult<()>;
+}
```
storopoli, line 15: @irnb eyes here please since you are touching DA specs right now.

crates/ol/stf/src/assembly.rs
```diff
@@ -0,0 +1,148 @@
+//! Block assembly flows.
+// TODO should this be in another crate?
```
storopoli: I think it's fine to keep it here.

crates/ol/chain-types/src/block.rs
```diff
@@ -165,6 +165,13 @@ impl OLBlockBody {
     pub fn l1_update(&self) -> Option<&OLL1Update> {
         self.l1_update.as_ref()
     }
+
+    /// Computes the hash commitment of this block body.
+    /// TODO: This will use SSZ merkle hashing when ready.
```
storopoli: This should be a comment not a docstring
```suggestion
    // TODO: This will use SSZ merkle hashing when ready.
```

crates/ol/chain-types/src/log.rs
```diff
@@ -26,4 +27,14 @@ impl OLLog {
     pub fn payload(&self) -> &[u8] {
         &self.payload
     }
+
+    /// Computes the hash commitment of this log.
+    /// TODO: This will use SSZ merkle hashing when ready.
```
storopoli, line 32: This should be a comment not a docstring
```suggestion
    // TODO: This will use SSZ merkle hashing when ready.
```

crates/ol/stf/src/assembly.rs
storopoli: FYI: https://ethereum.stackexchange.com/questions/35103/keccak-hash-of-null-values-result-in-different-hashes-for-different-types

crates/ol/msg-types/src/withdrawal.rs
delbonis: These are defined in the bridge messages spec doc.  Claude generated it on its own since I gave it that doc as a starting point.  I didn't remove it since it's not *wrong*, just premature.  They're unused as of right now.

crates/ol/msg-types/src/withdrawal.rs
delbonis: No it's for a BOSD, which does not have error correction.  Addresses are decoded into BOSDs near to the UI.  There's no value in having error correction at this level.

crates/ol/state-types/src/epochal.rs
delbonis: They're always up to date, from the perspective of the OL, will add docs.

crates/ol/stf/src/manifest_processing.rs
delbonis: The ASM does not do filtering like this and *will* emit logs that the OL doesn't care about, which right now includes ASM vk change logs and bridge utxo unlocking logs, but may include more in the future.

crates/ol/stf/src/verification.rs
delbonis: Maybe "classically" is an imprecise term here, but this is to contrast with syncing the epoch via DA blobs.

crates/ol/stf/src/verification.rs
delbonis: This is part of the stuff I need to rework to more cleanly separate the different layers of abstraction.

crates/ol/stf/src/verification.rs
delbonis: It's only verifying one step, not a whole chain.

crates/ol/stf/src/verification.rs
delbonis: This check makes sense to use in the proof since we need to ensure that the blocks in the proof are provided in a real chain.  It is also a useful sanity check when used outside the proof.

crates/ol/stf/src/verification.rs
delbonis: I prefer the term "skipped" that Ethereum uses, but yes.

crates/ol/chain-types/src/log_payloads.rs
delbonis: Yes there will be other kinds in the future.

crates/ol/stf/src/context.rs
delbonis: This is the simplest way to handle it.  The "most correct" way to do it would be to panic if it's >1, but we already do that on construction.

crates/ol/stf/src/test_utils.rs
delbonis: There's no need to actively do anything to revert since we won't be directly writing to the database.

crates/ol/stf/src/output.rs
delbonis: Yeah probably, that field used to be integrated into the context types but I reworked it into this type and that was retained.

crates/ol/stf/src/verification.rs
barakshani: It's implied by the word "block", otherwise would be "batch verification" : )

crates/ol/stf/src/account_processing.rs
delbonis: That doesn't prevent the issue with self-send (actually: reentrant access to self state).  The way to handle it is to defer external interactions until after we've updated the account's state.  This is where the `Coin` abstraction was useful, since we can model it as taking the funds out of the account, putting the account back in the table, and *then* sending the funds away.

crates/ol/chain-types/src/block.rs
delbonis: Ah this should probably be "common", I just used a different term here than the one I guess we've settled on.

crates/ol/state-types/src/epochal.rs
barakshani: Do we update `total_ledger_funds` which each block?

crates/ol/stf/src/verification.rs
barakshani: so shouldn't the check just make sure the new block's slot is greater than the previous block's slot?

crates/asm/common/src/log.rs
```diff
@@ -1,108 +1,2 @@
...
-        &self.0
-    }
-
-    /// Consume the log entry and return the raw bytes.
-    pub fn into_bytes(self) -> Vec<u8> {
-        self.0
-    }
-}
+// Re-export types from the manifest-types crate to maintain backward compatibility
+pub use strata_asm_manifest_types::{AsmLog, AsmLogEntry};
```
storopoli, line 2: Can you open a tech debt ticket to avoid having these annoying re-exports just to "maintain backward compatibility"?

crates/asm/manifest-types/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+mod manifest;
+mod payloads;
+
+pub use errors::*;
+pub use log::*;
+pub use manifest::*;
+pub use payloads::*;
+
+/// Type alias for a 32-byte hash.
+// TODO use Buf32 from identifiers?
```
storopoli, line 17: Yes, why not? We already have it and it has `ZeroizeOnDrop`.

crates/codec-utils/src/borsh_shim.rs
```diff
@@ -0,0 +1,113 @@
...
+    use borsh::{BorshDeserialize, BorshSerialize};
+    use strata_codec::{decode_buf_exact, encode_to_vec};
+
+    use super::*;
+
+    #[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+    struct TestStruct {
+        a: u32,
+        b: u64,
+    }
```
storopoli, line 64: All decoding/encoding should be `proptest!`ed. Can you do that?

crates/codec-utils/src/ssz_shim.rs
```diff
@@ -0,0 +1,110 @@
...
+mod tests {
+    use strata_codec::{decode_buf_exact, encode_to_vec};
+
+    use super::*;
+
+    #[derive(Debug, Clone, PartialEq, Eq, Decode, Encode)]
+    struct TestStruct {
+        a: u32,
+        b: u64,
+    }
```
storopoli, line 61: Same here, can you `proptest!`?

crates/identifiers/src/l1.rs
```diff
@@ -146,6 +158,33 @@ impl BorshDeserialize for L1BlockCommitment {
...
+        self.blkid.encode(enc)?;
+        Ok(())
+    }
+
+    fn decode(dec: &mut impl Decoder) -> Result<Self, CodecError> {
+        let height_u64 = u64::decode(dec)?;
+
+        #[cfg(feature = "bitcoin")]
+        let height = absolute::Height::from_consensus(height_u64 as u32)
+            .map_err(|_| CodecError::MalformedField("L1BlockCommitment.height"))?;
```
storopoli, line 179: These `foo_u64 as u32` will bite us in the future.

storopoli: Can't we just use `u32` in height for L1 stuff?

storopoli: Feel free to kick this to a ticket and assign to me. Will gladly tackle in a tech debt sprint soon.

crates/ol/stf/src/context.rs
storopoli: Shouldn't we have a better way to check for check for genesis?
Like

```rs
const fn is_genesis(&self) -> bool {}
```

You could add a field in the `OLBlockHeader` that we can safely skip during `Codec` serialization with something like

```rs
pub struct OLBlockHeader {
  #[codec(skip_serializing)]
  /// If the block is the genesis block.
  genesis: bool,
}
```
similar to [serde](https://serde.rs/attr-skip-serializing.html)

crates/ol/chain-types/src/log_payloads.rs
storopoli: Yes, all roundtrip (de)serialization stuff should be `proptest!`s.

crates/ol/msg-types/src/deposit.rs
```diff
@@ -0,0 +1,90 @@
...
+            dest_subject: SubjectId::from([42u8; 32]),
+        };
+
+        // Encode
+        let encoded = encode_to_vec(&msg_data).unwrap();
+
+        // Decode
+        let decoded: DepositMsgData = decode_buf_exact(&encoded).unwrap();
+
+        // Verify round-trip
```
storopoli: more roundtrips that should be `proptest!`s.

crates/ol/state-types/src/ledger.rs
```diff
@@ -0,0 +1,140 @@
...
+            "test: invalid serial sequencing"
+        );
+
+        if self.get_acct_entry_idx(&id).is_some() {
+            return Err(AcctError::AccountIdExists(id));
+        }
+
+        let serial = self.next_avail_serial();
+        let entry = TsnlAccountEntry::new(id, acct_state);
+        self.accounts.push(entry);
```
storopoli: This is broken!
It appends to the end of the vector. However, lookups use `binary_search_by_key`, which requires the vector to be sorted (see https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search_by_key).

CleanShot 2025-11-24 at 16 54 04@2x


Pushing a new ID that isn't greater than the last one will break future lookups.

storopoli: Also broken in the sense that fails to push the new ID to `self.serials`. Since `next_avail_serial()` relies on `self.serials.len()`, every subsequent account will try to claim the same serial number, causing collisions or panics...

crates/ol/chain-types/src/transaction.rs
```diff
@@ -1,36 +1,36 @@
...
 pub struct OLTransaction {
+    /// Any extra data associated with the transaction.
+    extra: TransactionAttachment,
+
     /// The actual payload for the transaction.
     payload: TransactionPayload,
-
-    /// Any extra data associated with the transaction.
-    extra: TransactionExtra,
 }
```
storopoli, line 20: Shouldn't this have a signature?
`process_single_tx` performs no signature verification. This means anyone can submit transactions on behalf of anyone else.
https://github.com/alpenlabs/alpen/blob/92ff00de80a7c8abb39cd9a93661a16d9fac6f60/crates/ol/stf/src/transaction_processing.rs#L36-L77

crates/ol/stf/src/manifest_processing.rs
```diff
@@ -0,0 +1,146 @@
...
+        process_asm_manifest(state, real_height, mf, context)?;
+    }
+
+    if let Some((last_height, last_mf)) = last {
+        // TODO this is where we would update the header, if we want to keep
+        // that as defined in the spec
+    }
+
+    Ok(())
+}
```
storopoli, line 48: The spec defines `check_l1_header_continuity` to ensure L1 manifests form a valid chain. The code here iterates through manifests but skips this check entirely, allowing discontinuous or invalid manifest chains to be processed.

crates/ol/stf/src/transaction_processing.rs
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    if let Some(max_slot) = atch.max_slot() {
+        if context.slot() > max_slot {
+            return false;
+        }
+    }
+
+    true
+}
```
storopoli, line 207: function only checks `min_slot` and `max_slot`. The spec mentions "Tx intrinsic checks (signatures correct, conditions, etc.)" should happen here, but the signature/intrinsic checks are absent and unmarked.

crates/ol/stf/src/account_processing.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        // Any other address we assume is a ledger account, so we have to look it up.
+        _ => {
+            // TODO adapt the account state traits to make this all more
+            // amendable to avoiding copies/clones
+
+            // Make a copy of the account state.  I don't love this.
+            let Some(mut acct_state) = state.get_account_state(target)?.cloned() else {
+                // If we don't find it then we can just ignore it.
+                // TODO do something with the funds we're throwing away by doing this
```
storopoli, line 46: This `TODO` is dangerous enough (loss of funds) that I think deserves a blocking review comment.

crates/identifiers/src/acct.rs
delbonis: this is actually a good thing to make const

crates/ol/state-types/src/epochal.rs
delbonis: Yeah these conversions are all fucked up because it was inconsistent in upstream crates.  It'll get sorted out when we transition more fully to SSZ.

crates/ol/chain-types/src/block_flags.rs
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    /// Assigns the `IS_TERMINAL` flag to some value.
+    pub fn set_is_terminal(&mut self, b: bool) {
+        if b {
+            self.0 |= IS_TERMINAL;
+        } else {
+            self.0 &= !IS_TERMINAL;
+        }
+    }
```
barakshani, line 26: What is this magic trick?

crates/ol/chain-types/src/block.rs
barakshani: >This means that not all OL blocks will have L1Updates?

Correct, only _terminal_ (last block of epoch) has it.

crates/asm/manifest-types/src/payloads.rs
barakshani: so... we don't have any spec that treats deposits properly, do we??? I guess we've forgotten what we're building here!

crates/ol/chain-types/src/block.rs
delbonis: Correct.

crates/ol/msg-types/src/withdrawal.rs
delbonis: I'm going to leave it so that we don't have to change the message format (and break compat) when we support a withdrawal fee market.

I was thinking of making it so that if the fees are set to be nonzero, but actually that could be a problem since then we would risk misuse resulting in loss of funds, so I think I'm going to just leave it as-is.

crates/ol/state-types/src/ledger.rs
delbonis: "Transitional" as in this is what we're using before we transition to having an actual account system where we only load in part of the state at a time.

crates/ol/state-types/src/snark_account.rs
delbonis: > we may have the account sending a few updates, while none of them is proven.

What you are describing is implemented as a (execution) sequencer account (which serve the same role as an EOA with respect to a rollup) sending an account message to a snark account.  Proving it is the "proven update" mechanism.  The message index is relevant to the "proof state" because it is tracking the messages that the proof needs to attest to the processing of.

I can't really think of a better name than `ProofState`.

crates/ol/stf/src/account_processing.rs
delbonis: ah this was some Claude confabulation

crates/ol/stf/src/context.rs
delbonis: I'll expose a `is_genesis_slot` as `slot == 0`.

crates/ol/stf/src/chain_processing.rs
delbonis: Good observation!  Since we don't have any skipped slots there isn't actually a case where this matters, but it does do block-specific things, so I'm renaming it.

crates/ol/stf/src/assembly.rs
delbonis: I've made the terminal flagging more systematic/explicit and changed this to `is_body_terminal`.

crates/ol/stf/src/transaction_processing.rs
delbonis: Numbering got mangled when I was moving some stuff around.  Some things are a little less feasible due to borrowing than it seemed in the spec doc, so kinda two layers of account access had to reverse themselves, which is annoying.  Fixed.

crates/ol/stf/src/verification.rs
delbonis: Until we've decided a way to make sense of that I just want to keep the behavior constrained to something predictable, so this just ensures it's always incrementing one at a time.

crates/ol/state-types/src/account.rs
delbonis: We can fix it in a future pr where we strip out a bunch of unnecessary checks involving `AcctResult`

crates/ol/state-types/src/account.rs
delbonis: I don't think it adds much value but I added it anyways.

crates/acct-types/src/mmr.rs
delbonis: I'll amend the other mmr ticket I created to reference this.

crates/asm/manifest-types/src/payloads.rs
delbonis: ahhhhhhhhhhhhhhhh

crates/asm/common/src/log.rs
delbonis: 1909

crates/codec-utils/src/borsh_shim.rs
delbonis: not sure how you'd proptest this particular case, unless you mean proptesting some arbitrary concrete types like that :thinking:

crates/ol/msg-types/src/deposit.rs
delbonis: converted more of these to use the Codec derive macro

crates/ol/state-types/src/ledger.rs
delbonis: Oh I had a whole response here about how it's safe but I realized I was mixing up sorting by serials and IDs in my head.

crates/ol/chain-types/src/transaction.rs
delbonis: Yep, we assume txs are authentic since only the sequencer that's assembling the block can put them there.  Transactions have no senders, they simply exist, so there's no semantically-sensical pubkey to check a signature against.

crates/ol/stf/src/transaction_processing.rs
delbonis: Yes, we haven't specified how this should work yet, so there's nothing to do.  We'll implement it later.

crates/ol/stf/src/account_processing.rs
delbonis: This is comparable to sending funds to a 0 address, which isn't invalid it's just stupid, so I'm not preventing it.

delbonis: This is comparable to sending funds to a 0 address, which isn't invalid it's just stupid, so I'm not preventing it.

crates/ol/chain-types/src/block_flags.rs
delbonis: We define the consts as bitmasks so we can use bithacks to manipulate the flags field

```
aaaaaaaa | 00000001 = aaaaaaa1
aaaaaaaa & 11111110 = aaaaaaa0
```

Calling it a bithack is maybe misleading since this is literally just the most sensical way to do it that isn't absurdly verbose.  Idk why Rust uses `!` for bitwise inversion instead of `~` like most languages but that's just what the syntax is.

crates/ol/stf/src/manifest_processing.rs
delbonis: I've realized that's actually fine, since a malicious sequencer can always just lie about the contents of the manifests anyways, even if they make the headers match.  What we care about is the manifests inserted into the manifest accumulator match ones that the (future) ASM has in its own history, which we check externally to the checkpoint proof.

I do have to update the spec for this.

crates/asm/manifest-types/src/lib.rs
delbonis: I have no idea, I didn't want to meaningfully change the impl beyond what I strictly had to to make the STF stuff work, just reorganize things, that's what it had iirc

crates/asm/manifest-types/src/payloads.rs
delbonis: Oh that spec doc has been abandoned afaik.  This def does come from the OL STF spec doc.

Cargo.toml
```diff
@@ -224,10 +234,11 @@ strata-zkvm-hosts = { path = "crates/zkvm/hosts" }
 
 # strata-common
 strata-codec = { git = "https://github.com/alpenlabs/strata-common" }
+strata-codec-derive = { git = "https://github.com/alpenlabs/strata-common" }
 strata-codec-tests = { git = "https://github.com/alpenlabs/strata-common" }
 strata-l1-envelope-fmt = { git = "https://github.com/alpenlabs/strata-common" }
 strata-l1-txfmt = { git = "https://github.com/alpenlabs/strata-common" }
-strata-merkle = { git = "https://github.com/alpenlabs/strata-common", branch = "merkle-proof-ssz", features = [
+strata-merkle = { git = "https://github.com/alpenlabs/strata-common", features = [
```
krsnapaudel, line 241: I think you should remove `ssz` feature from here.

delbonis:
Merging because all major issues resolved, will follow up on remainder later.

crates/ol/state-types/src/epochal.rs
storopoli: Do we have a ticket for that?

crates/ol/stf/src/transaction_processing.rs
storopoli: Please open a ticket for that?

crates/ol/stf/src/verification.rs
barakshani: add todo?

crates/ol/msg-types/src/withdrawal.rs
barakshani: it's a bit pre-mature, don't you think? We don't really know how the fee market will work

crates/ol/state-types/src/snark_account.rs
barakshani: I don't follow. I understand that right now we plan to have a single (update+proof) tx, but I was referring to a _contract_-like account that first gets updates ("commit") and later a proof. In this case, we may have several updates in a row before the proofs come in, so we need to have some pointer for last proven update.

Anyway, I still don't understand what this struct is.

crates/ol/chain-types/src/transaction.rs
barakshani: wait, this kinda makes all snark accounts' sequencing decentralised, doesn't it?
snark account should be programable to some minimal extent, no? we should allow an EE to have a pk that is only authorized to interact with it. I think.
I actually din't really know how other ecosystem treat this.

crates/ol/state-types/src/snark_account.rs
delbonis: > so we need to have some pointer for last proven update.

Using the term "update" here is causing confusion, the term you mean in the snark account framework is "message".

What you are describing is the `next_read_idx` field.  It's valid for snark account updates to not result in changing this field in the proof state, but when we have message inputs that get processed/acked by updates then it will.

PR #1159, opened 2025-11-21
feat(ol/rpc): Add OL RPC types and interface

bewakes:
## Description

Types and Traits defintions for OL RPC.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This PR was created with the help of Claude, especially the type conversions between inner and RPC types and documentation.

## Checklist

### DIFF
--- a/crates/rpc-new/api/src/lib.rs
+++ b/crates/rpc-new/api/src/lib.rs
@@ -1,14 +1,14 @@
 //! OL RPC API definitions.
 
 use jsonrpsee::{core::RpcResult, proc_macros::rpc};
-use strata_identifiers::{AccountId, OLBlockCommitment, OLBlockId, OLTxId};
+use strata_identifiers::{AccountId, OLTxId};
 use strata_primitives::proof::Epoch;
 use strata_rpc_types_new::*;
 
 /// Common OL RPC methods that are served by all kinds of nodes(DA, block executing).
 #[cfg_attr(not(feature = "client"), rpc(server, namespace = "strata"))]
 #[cfg_attr(feature = "client", rpc(server, client, namespace = "strata"))]
-pub trait OLCommonApi {
+pub trait OLClientRpc {
     /// Get an account's epoch summary for a given epoch.
     #[method(name = "getAccountEpochSummary")]
     async fn get_acct_epoch_summary(
@@ -16,39 +16,19 @@ pub trait OLCommonApi {
         account_id: AccountId,
         epoch: Epoch,
     ) -> RpcResult<RpcAccountEpochSummary>;
-}
 
-/// Api methods served by all block executing nodes.
-#[cfg_attr(not(feature = "client"), rpc(server, namespace = "strata"))]
-#[cfg_attr(feature = "client", rpc(server, client, namespace = "strata"))]
-pub trait OLFullNodeApi {
     /// Get current chain status (latest, confirmed, finalized).
-    #[method(name = "chainStatus")]
+    #[method(name = "getChainStatus")]
     async fn chain_status(&self) -> RpcResult<RpcOLChainStatus>;
-
-    /// Get block commitments for slot range [start_slot, end_slot].
-    #[method(name = "getBlockCommitmentsInRange")]
-    async fn get_block_commitments_in_range(
-        &self,
-        start_slot: u64,
-        end_slot: u64,
-    ) -> RpcResult<Vec<OLBlockCommitment>>;
 
     /// Get summaries associated with an account for given blocks.
     #[method(name = "getBlocksSummaries")]
     async fn get_blocks_summaries(
         &self,
         account_id: AccountId,
-        blocks: Vec<OLBlockId>,
+        start_slot: u64,
+        end_slot: u64,
     ) -> RpcResult<Vec<RpcAccountBlockSummary>>;
-
-    /// Get new messages associated with an account for given blocks.
-    #[method(name = "getBlocksNewMessages")]
-    async fn get_blocks_new_messages(
-        &self,
-        account_id: AccountId,
-        blocks: Vec<OLBlockId>,
-    ) -> RpcResult<Vec<Vec<RpcMessageEntry>>>;
 
     /// Submit transaction to the node. Returns immediately with tx ID.
     #[method(name = "submitTransaction")]
--- a/crates/rpc-new/types/src/types.rs
+++ b/crates/rpc-new/types/src/types.rs
@@ -4,20 +4,22 @@ use serde::{Deserialize, Serialize};
 use strata_acct_types::{AccountId, BitcoinAmount, MsgPayload};
 use strata_identifiers::{Buf32, OLBlockCommitment};
 use strata_ol_chain_types_new::{Slot, TransactionAttachment};
-use strata_primitives::{HexBytes, HexBytes32};
+use strata_primitives::{EpochCommitment, HexBytes, HexBytes32};
 use strata_snark_acct_types::{MessageEntry, ProofState, UpdateInputData, UpdateStateData};
 
 /// OL chain status with latest, confirmed, and finalized blocks.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct RpcOLChainStatus {
     pub latest: OLBlockCommitment,
-    pub confirmed: OLBlockCommitment,
-    pub finalized: OLBlockCommitment,
+    pub confirmed: EpochCommitment,
+    pub finalized: EpochCommitment,
 }
 
 /// Epoch summary for an account.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct RpcAccountEpochSummary {
+    /// The epoch commitment.
+    epoch_commitment: EpochCommitment,
     /// Final balance at the end of the epoch(sats).
     final_balance: u64,
     /// Final sequence number at the end of the epoch.
@@ -29,7 +31,7 @@ pub struct RpcAccountEpochSummary {
     // final update. Maybe it might make sense to have this field as a single RpcUpdateSummary ?
     updates: Vec<RpcUpdateSummary>,
     /// All new messages received in the epoch.
-    new_messages: Vec<RpcMessageEntry>,
+    new_received_messages: Vec<RpcMessageEntry>,
 }
 
 /// Describes an update to a snark account. This is derivable from L1.
@@ -56,9 +58,11 @@ pub struct RpcMsgPayload {
 
 impl From<MsgPayload> for RpcMsgPayload {
     fn from(payload: MsgPayload) -> Self {
+        let MsgPayload { data, value } = payload;
+        let data: Vec<u8> = data.into();
         Self {
-            value: payload.value().to_sat(),
-            data: payload.data().into(),
+            value: value.to_sat(),
+            data: data.into(),
         }
     }
 }
@@ -144,6 +148,8 @@ pub struct RpcOLTransaction {
 /// Block data associated with an account.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct RpcAccountBlockSummary {
+    /// Block commitment.
+    block_commitment: OLBlockCommitment,
     /// Block's slot.
     slot: Slot,
     /// Account's balance after the block execution(sats).
@@ -153,9 +159,7 @@ pub struct RpcAccountBlockSummary {
     /// Account's updates processed in the block.
     updates: Vec<RpcUpdateSummary>,
     /// New messages for the account in the block.
-    // NOTE: This might not be necessary for block summary if we have a different method to serve
-    // new messages in a block.
-    new_messages: Vec<RpcMessageEntry>,
+    new_received_messages: Vec<RpcMessageEntry>,
 }
 
 // Type conversions
@@ -165,7 +169,7 @@ impl From<MessageEntry> for RpcMessageEntry {
         Self {
             source: (*entry.source().inner()).into(),
             incl_epoch: entry.incl_epoch(),
-            payload: entry.payload().clone().into(),
+            payload: entry.payload.into(),
         }
     }
 }


crates/ol/rpc/api/README.md
delbonis, line 1: I don't think we need this file do we?

crates/rpc-new/api/src/lib.rs
delbonis, line 1: The namespace should be `strata` or `strataol` or something to that effect.

crates/rpc-new/types/src/types.rs
delbonis, line 1: This should be in a separate types crate.

crates/ol/rpc/api/src/types.rs
```diff
@@ -0,0 +1,260 @@
...
+        #[serde(with = "hex::serde")]
+        payload: Vec<u8>,
+    },
+    SnarkAccountUpdate {
+        #[serde(with = "hex::serde")]
+        target: [u8; 32],
+        update: RpcUpdateInputData,
+        #[serde(with = "hex::serde")]
+        update_proof: Vec<u8>,
+    },
```
delbonis: Struct-like enum variants are annoying to use.  Split these out as their own structs.

crates/ol/rpc/api/src/types.rs
```diff
@@ -0,0 +1,260 @@
...
+/// Block update inputs: block ID and its update input data.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct BlockUpdateInputs {
+    pub block_id: OLBlockId,
+    pub inputs: Vec<RpcUpdateInputData>,
+}
+
+// Type conversions
+use strata_ol_chain_types_new::TransactionExtra;
+use strata_snark_acct_types::{MessageEntry, ProofState, UpdateInputData, UpdateStateData};
```
delbonis: Top of file.

crates/ol/rpc/api/src/types.rs
```diff
@@ -0,0 +1,260 @@
...
+
+impl From<MessageEntry> for RpcMessageEntry {
+    fn from(entry: MessageEntry) -> Self {
+        Self {
+            source: *entry.source().inner(),
+            incl_epoch: entry.incl_epoch(),
+            payload: entry.payload().clone().into(),
+        }
+    }
+}
```
delbonis: Do we really need both of these?  This pattern repeats with a lot of these.

crates/ol/rpc/api/src/types.rs
```diff
@@ -0,0 +1,260 @@
...
+        target: [u8; 32],
+        update: RpcUpdateInputData,
+        #[serde(with = "hex::serde")]
+        update_proof: Vec<u8>,
+    },
+}
+
+/// Transaction extra: slot constraints.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RpcTransactionExtra {
```
delbonis: I thought we agreed to go back to the "attachment" terminology?

crates/rpc-new/api/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+        block_ids: Vec<OLBlockId>,
+    ) -> RpcResult<Vec<BlockMessages>>;
+
+    /// Get update inputs for specified blocks and account.
+    #[method(name = "updateInputsForBlocks")]
+    async fn update_inputs_for_blocks(
+        &self,
+        account_id: RpcAccountId,
+        blocks: Vec<OLBlockId>,
+    ) -> RpcResult<Vec<BlockUpdateInputs>>;
```
delbonis: "Update" is a verb, so this makes it sound like what the RPC does is update a block's inputs.  This concept is nonsensical.  Similar issues with the others where they need to be using verby names.

crates/ol/rpc/api/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+    #[method(name = "messagesForBlocks")]
+    async fn messages_for_blocks(
+        &self,
+        account_id: RpcAccountId,
+        blocks: Vec<OLBlockId>,
+    ) -> RpcResult<Vec<BlockMessages>>;
+
+    /// Submit transaction to mempool. Returns immediately with tx ID.
+    #[method(name = "submitTransaction")]
+    async fn submit_transaction(&self, tx: RpcOlTransaction) -> RpcResult<OLTxId>;
```
delbonis: Any node can have a tx submitted to it, not just sequencer nodes.

crates/ol/rpc/api/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+        &self,
+        start_slot: u64,
+        end_slot: u64,
+    ) -> RpcResult<Vec<OLBlockCommitment>>;
+}
+
+/// OL sequencer-specific RPC methods.
+#[cfg_attr(not(feature = "client"), rpc(server, namespace = "ol"))]
+#[cfg_attr(feature = "client", rpc(server, client, namespace = "ol"))]
+pub trait OlSequencerApi {
```
delbonis: By "sequencer-specific" what do you mean?  None of these are RPCs that I would expect could only be serviced by a node that's being used for sequencing.

Also, naming inconsistency here, `_for_block_range` vs `_for_blocks`.  The `_for_block_range` one is also an issue because the blkids that are passed might not be in a range, when the other ones give no indication that they should be in a range.

crates/ol/rpc/api/src/types.rs
bewakes: Interesting that `cargo fmt ` didn't catch this. Done

crates/rpc-new/api/src/lib.rs
bewakes: True. Prefixed with `get_`. Also happy to hear suggestions.

crates/ol/rpc/api/src/types.rs
bewakes: Oh yeah, not really. Claude was just trying to impress. :D

crates/ol/rpc/api/src/lib.rs
bewakes: I suppose we don't have mempool syncing among OL nodes, do we? And also, non-sequencer node doesn't need to maintain mempool either if we are not propagating mempool txs.

crates/ol/rpc/api/src/types.rs
bewakes: Yes, we did. Updated.

crates/ol/rpc/api/src/lib.rs
bewakes: This interface is based on the assumption that we won't have mempool for OL fullnodes because we won't be propagating mempool transactions.
So under that assumption, all the endpoints related to creating transactions to ultimately submit to mempool are "sequencer-specific" (like getting block messages and update inputs for an account).
The other general endpoints would be `chainStatus`, endpoints for block syncing(not decided yet how the interface should be), etc.

bewakes (commented):
Thanks @delbonis for your feedback. I have addressed most of them.

In this PR, I intend to include the methods that are required for EL sequencer to be able to create EL packages, and then ultimately submit transactions to OL. I had a discussion with @sapinb prior to writing this interface on what things are necessary and we agreed on this, but if I have missed anything @sapinb please let me know.

Agree with `Ol` -> `OL`. Made changes accordingly.

> Also, this should be in the rpc crate subtree shouldn't it?

Yes, we will have it there ultimately. Right now, I did not want to change what's there already as those will be replaced  by the new ones. It is also possible to create a separate module for these new methods in the current `rpc` crate. I am fine with doing that as well.

crates/ol/rpc/api/src/lib.rs
delbonis: > on the assumption that we won't have mempool for OL fullnodes because we won't be propagating mempool transactions.

This is an incorrect assumption.  It's not a product goal for mainnet, but it is an engineering goal long-term.

Plus anyways, if your client doesn't have an RPC to receive txs then how would you be able to submit a tx to your client for it to submit the sequencer?

crates/ol/rpc/types/src/types.rs
```diff
@@ -0,0 +1,238 @@
...
+    fn from(id: AccountId) -> Self {
+        RpcAccountId(*id.inner())
+    }
+}
+
+impl From<RpcAccountId> for AccountId {
+    fn from(rpc: RpcAccountId) -> Self {
+        AccountId::new(rpc.0)
+    }
+}
```
delbonis: For a type like this we should probably just make it RPC-safe like the blkid/commitment types are.

delbonis (changes requested):
Also don't forget to move the directories, maybe call it `newrpc`.

delbonis:
One issue with these RPCs as defined is that the data wouldn't be available at that level of granularity with DA sync, we'd only get the updates/messages at the epoch level.  We *should* have RPCs *like* this for EE client block production, to be able to understand what it's doing in realer-time.

For future reference in the interest of modularity, see [the Mosaic RPC crate hierarchy](https://github.com/alpenlabs/mosaic/tree/main/crates/rpc).  I would like to find a structure that doesn't force that anyone that consumes the RPC types crate not also depend on all of the upstream types, the best way that doesn't involve awkward restructuring would probably involve adding feature flags to be able to disable compilation of messy impl stuff.

crates/ol/rpc/api/src/lib.rs
bewakes: >  It's not a product goal for mainnet, but it is an engineering goal long-term.

I understand it is an engineering goal long term.

There are two scenarios:
1. If full nodes implement these methods then they need to either have sequencer's endpoint or some sort of relay logic(`if sequencer, then add to mempool, else relay to sequencer or other full node`).
2. If full nodes don't implement these methods then the EL sequencer can't just connect to any full node, it needs to connect to the sequencer node or else it can't submit the txs.

For our mainnet goal, I think option 2 is more desirable because it's simpler and, both EL and OL sequencers will be on the same network and run by us. Plus when we support proper propagation, we can simply move the impls around without any engineering hassle.

With option 1, we need to be careful about exposing sequencer's endpoint and I don't want to worry about relay logic when it's not actually needed. I presume we will have fullnodes other than the ones that we run ourselves.

crates/ol/rpc/types/src/types.rs
bewakes: Yeah, Agree.

delbonis:
> Do you mean something like this?
> [...]

The most ergonomic way to do it would be to group the updates with the messages that were processed in the update, and then separately list messages that haven't been processed in an update yet.  Grouping messages with the updates is essential for exec clients to be able to reconstruct EE state properly  You *could* provide it as a flat list of a subrange of messages with the index and then list the message indexes in the update, but then clients would have more work to do, so perhaps Sapin should be consulted on that.

Also, echoing the blkid in the response is redundant, you already know what the blkid is because you passed it as an argument.

crates/ol/stf/src/tests.rs
delbonis, line 1: Looking at this is making me think it might be useful to have a `TransactionBuilder`?  Have different constructors for each payload tx variant that make it future-extensible, and then setting the various attachment fields can be the builder fns.

crates/rpc-new/api/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+/// RPC Methods served only by sequencer node.
+#[cfg_attr(not(feature = "client"), rpc(server, namespace = "strata"))]
+#[cfg_attr(feature = "client", rpc(server, client, namespace = "strata"))]
+pub trait OLSequencerApi {
+    /// Submit transaction to mempool. Returns immediately with tx ID.
+    // TODO: This method will be served by all the nodes in future.
+    #[method(name = "submitTransaction")]
+    async fn submit_transaction(&self, tx: RpcOLTransaction) -> RpcResult<OLTxId>;
```
delbonis: I said before, not just sequencers are able to have transactions submitted to them!  *Even if* they're not running their own mempools!

There's very few RPCs that are *only* things that full nodes being configured to serve sequencers are able to serve!  Calling this `OLSequencerApi` implies that it's on the *sequencer* client, not the *full node being used with a sequencer client*.

crates/rpc-new/api/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+    ) -> RpcResult<RpcAccountEpochSummary>;
+}
+
+/// Api methods served by all block executing nodes.
+#[cfg_attr(not(feature = "client"), rpc(server, namespace = "strata"))]
+#[cfg_attr(feature = "client", rpc(server, client, namespace = "strata"))]
+pub trait OLFullNodeApi {
+    /// Get current chain status (latest, confirmed, finalized).
+    #[method(name = "chainStatus")]
+    async fn chain_status(&self) -> RpcResult<RpcOLChainStatus>;
```
delbonis, line 22: This is something any node can serve.

crates/rpc-new/types/src/types.rs
```diff
@@ -0,0 +1,245 @@
...
+    /// Hex-encoded data.
+    #[serde(with = "hex::serde")]
+    pub data: Vec<u8>,
+}
+
+impl From<MsgPayload> for RpcMsgPayload {
+    fn from(payload: MsgPayload) -> Self {
+        Self {
+            value: payload.value().to_sat(),
+            data: payload.data().to_vec(),
```
delbonis: Unnecessary copy.

crates/rpc-new/types/src/types.rs
```diff
@@ -0,0 +1,245 @@
...
+}
+
+// Type conversions
+
+impl From<MessageEntry> for RpcMessageEntry {
+    fn from(entry: MessageEntry) -> Self {
+        Self {
+            source: *entry.source().inner(),
+            incl_epoch: entry.incl_epoch(),
+            payload: entry.payload().clone().into(),
```
delbonis: Unnecessary copy.

crates/rpc-new/types/src/types.rs
```diff
@@ -0,0 +1,245 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RpcOLChainStatus {
+    pub latest: OLBlockCommitment,
+    pub confirmed: OLBlockCommitment,
+    pub finalized: OLBlockCommitment,
+}
+
+/// Epoch summary for an account.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RpcAccountEpochSummary {
```
delbonis, line 20: This is specific to snark accounts tbf.  Might also want to expose the "final next message idx" for convenience.

crates/rpc-new/types/src/types.rs
```diff
@@ -0,0 +1,245 @@
...
+use strata_identifiers::{Buf32, OLBlockCommitment};
+use strata_ol_chain_types_new::{Slot, TransactionAttachment};
+use strata_snark_acct_types::{MessageEntry, ProofState, UpdateInputData, UpdateStateData};
+
+/// OL chain status with latest, confirmed, and finalized blocks.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RpcOLChainStatus {
+    pub latest: OLBlockCommitment,
+    pub confirmed: OLBlockCommitment,
+    pub finalized: OLBlockCommitment,
```
delbonis: Do you think this should actually use `EpochCommitment`s for (L1) confirmed/finalized?

crates/rpc-new/types/src/types.rs
delbonis, line 1: We're being kinda inconsistent here about if RPC types have pub fields or priv fields.  Also it'd be good to clean up the repetitive `#[serde(with = "hex::serde")] foo: [u8; 32]` thing.  We could either do this *maybe* by altering `Buf32` to naturally serde when the format is human-readable to hex, or create a "hex buf" type that's equivalent to use in this context and use that here.

crates/ol/rpc/api/src/lib.rs
delbonis: It sounds like you might be mixing up use cases of the p2p gossip messages and RPC interfaces.

delbonis: Like, it's valid for nodes to return an error from the tx submit RPC if they know that they're unable to ever do anything with it, or drop/discard txs they realize they can't relay later.  We can log this even and the user can figure out it's happening when it's happening with their own node, or if it's our sequencing infra then it's fine to just not report anything.  Tx submission is not supposed to be a robust/reliable, it's like UDP.  You have to keep retrying until you recognize it's gotten (pre)confirmed, like TCP works under the hood.  Users('s wallets) are expected to keep trying to submit txs (through whatever means available to them) until either it concludes the tx can never be valid or the user decides to give up (and try something else) and abort trying to get the tx included.

Reliable/slashable preconfirmations would have a more sophisticated interactive procedure, since that requires a response directly from the sequencer that's scheduled to assemble a block and there's a bunch of machinery that has to be built for that (to keep track of the txs they've declared they're going to include and making sure they're not overpromising block/DA capacity), and this submission path is *much* simpler and less interactive.

crates/ol/stf/src/tests.rs
bewakes: Yeah, it sounds like a good idea to have builders. We can have them later though.

crates/rpc-new/api/src/lib.rs
bewakes: Is it? The chain status has `confirmed`, `finalized`, and `latest` OL block commitments, something only a block syncing node would have.

crates/rpc-new/api/src/lib.rs
bewakes: > not just sequencers are able to have transactions submitted to them!  Even if they're not running their own mempools!

That's understood. I just did not want to expose methods that are not currently supported. But if it's okay to error out for unsupported method, we can just move that to FullNodeApi.

crates/rpc-new/types/src/types.rs
bewakes: Yes, I think we already have a hexbuf type. will use that.

crates/rpc-new/types/src/types.rs
bewakes: how?

crates/rpc-new/types/src/types.rs
bewakes: How again? `.payload()` returns a reference.

crates/rpc-new/types/src/types.rs
bewakes: This is what sapin needs in the EL so leaving as is for now.

crates/rpc-new/api/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
```
sapinb, line 22: Chain status should be part of the common api. Both types of nodes can serve this.

crates/rpc-new/types/src/types.rs
```diff
@@ -0,0 +1,244 @@
...
+pub struct RpcOLTransaction {
+    pub payload: RpcTransactionPayload,
+    pub attachments: RpcTransactionAttachment,
+}
+
+/// Block data associated with an account.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RpcAccountBlockSummary {
+    /// Block's slot.
+    slot: Slot,
```
sapinb, line 154: Please add OLBlockCommitment here, to keep things similar to the Epoch Summary

crates/rpc-new/api/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+        end_slot: u64,
+    ) -> RpcResult<Vec<OLBlockCommitment>>;
+
+    /// Get summaries associated with an account for given blocks.
+    #[method(name = "getBlocksSummaries")]
+    async fn get_blocks_summaries(
+        &self,
+        account_id: AccountId,
+        blocks: Vec<OLBlockId>,
+    ) -> RpcResult<Vec<RpcAccountBlockSummary>>;
```
sapinb, line 31: Lets merge this with the method above:

`get_block_summaries(start_slot: u64, end_slot: u64) -> Vec<RpcAccountBlockSummary>`

crates/rpc-new/api/src/lib.rs
delbonis: Also this should be `getChainStatus`, or maybe something more specific like `getSyncStatus`.

crates/rpc-new/api/src/lib.rs
delbonis: That's not correct, all nodes can give answers for those fields.  For a checkpoint-only sync node it just happens to be that `latest`, to the best of its knowledge, will always be the same as `confirmed`.

crates/rpc-new/types/src/types.rs
delbonis: `from` takes ownership of the `MsgPayload` so it's fine to just disassemble it, add a `into_data` if you need to.

crates/rpc-new/types/src/types.rs
delbonis: Add a function to take ownership of it.  As in the other case, the function takes ownership of the value, so it's valid to move out the payload vec into the value you're constructing.

crates/rpc-new/api/src/lib.rs
bewakes: I see. That makes sense.

delbonis (approved):
will probably have to adjust the epoch summary types slightly later but this is important to merge for velocity reasons

storopoli (approved):
ACK 314e342d5fd855d25573c6268b880eb07700147c

PR #1160, opened 2025-11-21
fix: ensure checkpoint is verified against correct proof receipt

sapinb:
## Description
Verify checkpoint against proof receipt passed in input.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/consensus-logic/src/checkpoint_verification.rs
+++ b/crates/consensus-logic/src/checkpoint_verification.rs
@@ -2,67 +2,9 @@
 
 use strata_chaintsn::transition::verify_checkpoint_proof;
 use strata_checkpoint_types::{BatchTransition, Checkpoint};
-use strata_csm_types::L1Checkpoint;
 use strata_primitives::params::*;
 use tracing::*;
 use zkaleido::{ProofReceipt, ZkVmError, ZkVmResult};
-
-use crate::errors::CheckpointError;
-
-/// Verifies if a checkpoint if valid, given the context of a previous checkpoint.
-///
-/// If this is the first checkpoint we verify, then there is no checkpoint to
-/// check against.
-///
-/// This does NOT check the signature.
-// TODO reduce this to actually just passing in the core information we really
-// need, not like the height
-pub fn verify_checkpoint(
-    checkpoint: &Checkpoint,
-    prev_checkpoint: Option<&L1Checkpoint>,
-    params: &RollupParams,
-) -> Result<(), CheckpointError> {
-    // First thing obviously is to verify the proof.  No sense in continuing if
-    // the proof is invalid.
-    let proof_receipt = construct_receipt(checkpoint);
-    verify_proof_receipt_against_checkpoint(checkpoint, &proof_receipt, params)?;
-
-    // And check that we're building upon the previous state correctly.
-    if let Some(prev) = prev_checkpoint {
-        verify_checkpoint_extends(checkpoint, prev, params)?;
-    } else {
-        // If it's the first checkpoint we want it to be the initial epoch.
-        if checkpoint.batch_info().epoch() != 0 {
-            return Err(CheckpointError::SkippedGenesis);
-        }
-    }
-
-    Ok(())
-}
-
-/// Verifies that the a checkpoint extends the state of a previous checkpoint.
-fn verify_checkpoint_extends(
-    checkpoint: &Checkpoint,
-    prev: &L1Checkpoint,
-    _params: &RollupParams,
-) -> Result<(), CheckpointError> {
-    let epoch = checkpoint.batch_info().epoch();
-    let prev_epoch = prev.batch_info.epoch();
-    let last_tsn = prev.batch_transition;
-    let tsn = checkpoint.batch_transition();
-
-    // Check that the epoch numbers line up.
-    if epoch != prev_epoch + 1 {
-        return Err(CheckpointError::Sequencing(epoch, prev_epoch));
-    }
-
-    if last_tsn.chainstate_transition.post_state_root != tsn.chainstate_transition.pre_state_root {
-        warn!("checkpoint mismatch on L2 state!");
-        return Err(CheckpointError::MismatchL2State);
-    }
-
-    Ok(())
-}
 
 /// Constructs a receipt from a checkpoint.
 ///


crates/consensus-logic/src/checkpoint_verification.rs
```diff
@@ -72,12 +72,12 @@ pub fn construct_receipt(checkpoint: &Checkpoint) -> ProofReceipt {
...
 }
 
-/// Verify that the provided checkpoint proof is valid for the verifier key.
+/// Verify that the provided checkpoint and proof is valid for the verifier key.
 ///
 /// # Caution
 ///
 /// If the checkpoint proof is empty, this function returns an `Ok(())`.
-pub fn verify_proof(
+pub fn verify_proof_receipt_against_checkpoint(
```
barakshani, line 22: why the name change?

crates/chaintsn/src/transition.rs
```diff
@@ -170,7 +170,8 @@ fn process_l1_checkpoint(
     if ckpt_epoch != ckpt.batch_info().epoch() {
         return Err(OpError::MalformedCheckpoint);
     }
-    verify_checkpoint_proof(ckpt, params).map_err(|_| OpError::InvalidProof)?;
+    let proof_receipt = ckpt.construct_receipt();
```
barakshani, line 172: can you explain to me what has changed? previously we called `verify_checkpoint_proof` with the checkpoint, and then computed the proof receipt inside that function _from_ that checkpoint; now we first compute the proof receipt from the checkpoint, then pass the checkpoint and the receipt to the function. Why would the result be different?

barakshani: I thought that the issue was that compute the receipt from the checkpoint instead from the proof, while the checkpoint does not hold of the proof at this stage.

krsnapaudel (commented):
Do we still need this
Line 160 in `transition.rs`
```
let _receipt = ckpt.construct_receipt();
```

crates/chaintsn/src/transition.rs
krsnapaudel: With some AI help, I figured out that `verify_proof` was ignoring the `proof_receipt`, which should have been passed to `verify_checkpoint_proof`.

crates/consensus-logic/src/checkpoint_verification.rs
sapinb: To make clear what its actually doing. `verify_proof` was just too generic. This function checks that the provided proof receipt is valid and matches the provided checkpoint.

crates/chaintsn/src/transition.rs
sapinb: As @krsnapaudel mentioned above, the provided proof receipt was not getting checked, but the one inside checkpoint (ie. empty) was being used instead. This fix passes the proof receipt and checks against the checkpoint data, as it is exptected.

barakshani: The flow was: call `verify_checkpoint_proof(ckpt)` where inside we do `proof_receipt = ckpt.construct_receipt()` and use `proof_receipt` for verification.

The flow now is: do `proof_receipt = ckpt.construct_receipt();` and call `verify_checkpoint_proof(ckpt, proof_receipt)`, then use `proof_receipt` for verification.

why do we expect a different result? in both cases we use the same checkpoint to derive, in the same manner, the proof receipt.

crates/consensus-logic/src/checkpoint_verification.rs
barakshani: Do I understand correctly that this function is _only_ called when we are reading the checkpoint from L1? Why do we use the term _proof receipt_ an not just "proof"?

barakshani:
While we are making changes to these files, can we remove this text from checkpoint_verification.rs?
`    // First thing obviously is to verify the proof.  No sense in continuing if
    // the proof is invalid.`

Since we need all checks to pass, it applies generally. Actually it may be that the proof verification is more expensive, and so we'd rather start with verifying less expensive things (I don't ask to change the verification order, just to remove the text).

crates/consensus-logic/src/checkpoint_verification.rs
```diff
@@ -25,7 +25,7 @@ pub fn verify_checkpoint(
     // First thing obviously is to verify the proof.  No sense in continuing if
```
barakshani: Do we even use this function (`verify_checkpoint`) anywhere?

crates/chaintsn/src/transition.rs
barakshani: My conclusion is that _this_ flow was fine. The problem we experienced was with the path from `rpc_server.rs`, where we call `submit_checkpoint_proof(proof_receipt)`, but inside we call `verify_checkpoint_proof` without the proof receipt. Now it's fixed.

delbonis: Really we should be removing the one inside the checkpoint, we should only be constructing the receipt from the data we explicitly put in the checkpoint.  The receipt duplicates information that's put in the checkpoint.  This is where the confusion stems from, since there's duplicated data that we then have to go add checks for.

The new checkpoint format doesn't have this issue, so we should be trying to transition the various systems to working more like how that would work.

crates/consensus-logic/src/checkpoint_verification.rs
prajwolrg: `proof_receipt` includes both the proof and the public parameters.

crates/consensus-logic/src/checkpoint_verification.rs
sapinb: No. Its no longer anywhere after the ASM/CSM changes. Removing this.

crates/chaintsn/src/transition.rs
sapinb: I believe the main source of confusion is having a single type for both pending and complete checkpoints. Further confusion is added as an empty "proof" can mean either it is pending proof (known from a status elsewhere), or that `allow_empty_proofs` is enabled.
We can make pending checkpoint as a separate type (but this might require db to also store the 2 differently), or put proof under an enum to differentiate between a valid empty proof and an pending proof.
```rust
enum ProofStatus {
  Pending,
  Proven(Proof)
}
```

> constructing the receipt from the data we explicitly put in the checkpoint

I do not think this is the problem, as this issue specifically comes with verifying a proof is valid before we assign it in the checkpoint's constraints (public params). All other cases are validating that a checkpoint is valid.

barakshani: My conclusion is that _this_ flow was fine. The problem we experienced was with the path from `rpc_server.rs`, where we call `submit_checkpoint_proof(proof_receipt)`, but inside (via `verify_proof/verify_proof_receipt_against_checkpoint`) we call `verify_checkpoint_proof` without the proof receipt. Now it's fixed.

PR #1163, opened 2025-11-25
Feat: admin subprotocol hardware wallet compatibility.

irnb:
## Description

the admin subprotocol now uses individual `ECDSA` signatures instead of `MuSig2 aggregated Schnorr` signatures for hardware wallet compatibility.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


barakshani:
Does the admin spec represent this change?

crates/crypto/src/test_utils/schnorr.rs
```diff
@@ -149,7 +145,11 @@ mod tests {
 
         // Generate test private keys for 3 operators
         let operator_privkeys: Vec<EvenSecretKey> = (0..3)
-            .map(|_| EvenSecretKey::from(SecretKey::new(&mut OsRng)))
+            .map(|_| {
+                let mut sk_bytes = [0u8; 32];
+                OsRng.fill_bytes(&mut sk_bytes);
+                EvenSecretKey::from(SecretKey::from_slice(&sk_bytes).unwrap())
+            })
             .collect();
```
prajwolrg, line 153: Why are we generating the same private key everytime?

crates/crypto/src/threshold_signing/indexed_signatures/errors.rs
```diff
@@ -0,0 +1,88 @@
+//! Error types for threshold signing operations.
+
+use std::fmt;
```
prajwolrg: We can use `thiserror` crate here.

crates/crypto/src/threshold_signature/indexed/signature.rs
```diff
@@ -0,0 +1,207 @@
...
+use super::ThresholdSigningError;
+
+/// An individual ECDSA signature with its signer index.
+///
+/// The signature is in recoverable format (65 bytes): `recovery_id || r || s`.
+/// This format is used for hardware wallet compatibility (Ledger/Trezor native output).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct IndexedSignature {
+    /// Index of the signer in the ThresholdConfig keys array (0-255).
+    pub index: u8,
```
prajwolrg: I'm not sure why we limited ourselves to `u8` earlier. Maybe it's reasonable. But the PRD specifies there should be no limit. So maybe we should set this higher. We can also explore `VarInt` to save up on DA space, that gives us the best of both worlds.

crates/crypto/src/threshold_signing/indexed_signatures/verification.rs
```diff
@@ -0,0 +1,242 @@
...
+///    - The ECDSA signature is valid for the corresponding public key
+///
+/// # Returns
+///
+/// * `Ok(())` if all signatures are valid and threshold is met
+/// * `Err(ThresholdSigningError)` otherwise
+pub fn verify_threshold_signatures(
+    config: &ThresholdConfig,
+    signatures: &SignatureSet,
+    message_hash: &[u8; 32],
```
prajwolrg: I think the argument to the function should be `signatures: &[IndexedSignature]` and we should construct the `SignatureSet` inside **this** function. As of the current implementation, the `SignatureSet` comes from borsh deserialization where we are ensuring that we are using `SignatureSet::new` that checks for duplications . If we change the encoding/decoding, or derive the `BorshSerialize`/`BorshDeserialize`, it's easy to miss this check.

prajwolrg (changes requested):

crates/crypto/src/threshold_signature/indexed/signature.rs
barakshani: this is fine, a limit must be put practically.

crates/asm/subprotocols/admin/src/authority.rs
```diff
@@ -1,62 +1,61 @@
...
-        signature: &SchnorrMultisigSignature,
-    ) -> Result<(), MultisigError> {
+        signatures: &SignatureSet,
+    ) -> Result<(), ThresholdSigningError> {
         // Compute the msg to sign by combining UpdateAction with sequence no
         let sig_hash = action.compute_sighash(self.seqno);
 
-        // Use the generic multisig verification function
-        verify_multisig(&self.config, signature, &sig_hash.into())
+        // Verify each ECDSA signature against the corresponding public key
```
barakshani: I think that this function (`verify_threshold_signatures`) should be blind to the "threshold sig scheme" we are using. See my other comment on this function.

crates/crypto/src/threshold_signing/indexed_signatures/verification.rs
```diff
@@ -0,0 +1,242 @@
...
+/// 1. Check that the number of signatures meets the threshold
+/// 2. For each signature, verify that:
+///    - The signer index is within bounds
+///    - The ECDSA signature is valid for the corresponding public key
+///
+/// # Returns
+///
+/// * `Ok(())` if all signatures are valid and threshold is met
+/// * `Err(ThresholdSigningError)` otherwise
+pub fn verify_threshold_signatures(
```
barakshani: I think it would be better to have this "generic" verify-threshold-named function as a wrapper around the actual impl. The impl is now to verify a collection of ECDSA but in the future it may be something else. Since we are calling this function in several places, I think it would be good to not worry about the places where we call it (about the internal impl), and keep it generic.

crates/asm/subprotocols/admin/src/error.rs
```diff
@@ -13,7 +13,7 @@ pub enum AdministrationError {
     #[error("no pending update found for action_id = {0:?}")]
     UnknownAction(UpdateId),
 
-    /// Indicates a multisig error (configuration, aggregation, or signature validation).
+    /// Indicates a threshold signing error (configuration or signature validation).
```
barakshani:
```suggestion
    /// Indicates a threshold signature error (configuration or signature validation).
```
IIUC the problem is not in (threshold) _signing_ (each singing is done individually anyway), but in config related stuff.

crates/crypto/src/threshold_signature/indexed/config.rs
```diff
@@ -0,0 +1,297 @@
...
+use arbitrary::Arbitrary;
+use borsh::{BorshDeserialize, BorshSerialize};
+
+use super::{CompressedPublicKey, ThresholdSigningError};
+
+/// Configuration for a threshold signature authority.
+///
+/// Defines who can sign (`keys`) and how many must sign (`threshold`).
+#[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub struct ThresholdConfig {
```
barakshani, line 22: Previous implementation had
```
/// Phantom data to carry the crypto scheme type.
    #[borsh(skip)]
    _phantom: PhantomData<S>,
```
We don't want to keep it?

crates/crypto/src/threshold_signature/indexed/config.rs
```diff
@@ -0,0 +1,297 @@
...
+
+impl std::hash::Hash for CompressedPublicKey {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.serialize().hash(state);
+    }
+}
+
+/// Represents a change to the threshold configuration.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ThresholdConfigUpdate {
```
barakshani, line 151: as above re phantom

crates/crypto/src/threshold_signature/indexed/signature.rs
```diff
@@ -0,0 +1,207 @@
...
+use borsh::{BorshDeserialize, BorshSerialize};
+
+use super::ThresholdSigningError;
+
+/// An individual ECDSA signature with its signer index.
+///
+/// The signature is in recoverable format (65 bytes): `recovery_id || r || s`.
+/// This format is used for hardware wallet compatibility (Ledger/Trezor native output).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct IndexedSignature {
```
barakshani, line 30: what is the expectation, who is going to match a (signature) pk to the index?

crates/crypto/src/threshold_signing/indexed_signatures/verification.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    // Create the message for verification
+    let message = Message::from_digest_slice(message_hash)
+        .map_err(|_| ThresholdSigningError::InvalidMessageHash)?;
+
+    // Verify each signature
+    for indexed_sig in signatures.signatures() {
+        // Check index is in bounds
+        let index = indexed_sig.index as usize;
+        if index >= config.keys().len() {
```
barakshani: `>=` or `>`?
I guess it's fine (index starts from 0), but please double check.

crates/crypto/src/threshold_signing/indexed_signatures/verification.rs
```diff
@@ -0,0 +1,242 @@
...
+            .map_err(|_| ThresholdSigningError::InvalidSignature {
+                index: indexed_sig.index,
+            })?;
+
+        // Verify the recovered key matches the expected key
+        if &recovered_pubkey != expected_pubkey {
+            return Err(ThresholdSigningError::InvalidSignature {
+                index: indexed_sig.index,
+            });
+        }
```
barakshani: What's the benefit, if any, of this verification (extract publey and match) on using the pubkey (fetched by index) an applying the sig verification algorithm?

barakshani: shouldn't we give more detailed errors?

crates/crypto/src/threshold_signature/indexed/pubkey.rs
```diff
@@ -0,0 +1,173 @@
...
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+use super::ThresholdSigningError;
+
+/// A compressed secp256k1 public key (33 bytes).
+///
+/// This is a thin wrapper around `secp256k1::PublicKey` that adds Borsh
+/// serialization support. Unlike `EvenPublicKey`, this type does not
+/// enforce even parity - it accepts any valid compressed public key.
```
storopoli, line 17: Please also add that the don't enforce even parity because this key is supposed to be used with ECDSA.

crates/crypto/src/threshold_signature/indexed/config.rs
storopoli: Because it was generic over any `S` that is anything that implements the `CryptoScheme` trait.

```rs
#[derive(Debug, Clone, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
pub struct MultisigConfig<S: CryptoScheme> {
    /// The public keys of all grant-holders authorized to sign.
    pub keys: Vec<S::PubKey>,
    /// The minimum number of keys that must sign to approve an action.
    pub threshold: u8,
    /// Phantom data to carry the crypto scheme type.
    #[borsh(skip)]
    _phantom: PhantomData<S>,
}
```

Since this proposed implementation does not use trait abstraction and goes straight to the `secp256k1`'s `PublicKey` wrapper, we don't need to have `PhantomData` for the compiler.

crates/crypto/src/test_utils/schnorr.rs
storopoli: As Prajwol said, why do we need this diff?

crates/crypto/src/threshold_signature/indexed/config.rs
```diff
@@ -0,0 +1,297 @@
...
+
+/// Configuration for a threshold signature authority.
+///
+/// Defines who can sign (`keys`) and how many must sign (`threshold`).
+#[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub struct ThresholdConfig {
+    /// Public keys of all authorized signers.
+    keys: Vec<CompressedPublicKey>,
+    /// Minimum number of signatures required.
+    threshold: u8,
```
storopoli: You can do `NonZero<u8>` using Rust's std library, see: https://doc.rust-lang.org/stable/std/num/struct.NonZero.html

I think we should do this.
We enforce impossible conditions (invariants) using the type system. We also advocate that in our [Engineering Best Practices](https://www.notion.so/Engineering-Rust-Best-Practices-14c901ba000f8025a974ec8e2a70b4af?source=copy_link#14d901ba000f80008606ff02b84a908a)

crates/crypto/src/threshold_signing/indexed_signatures/errors.rs
```diff
@@ -0,0 +1,88 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ThresholdSigningError {
+    /// Not enough signatures to meet the threshold.
+    InsufficientSignatures { provided: usize, required: usize },
+    /// Invalid public key data.
+    InvalidPublicKey { index: usize, reason: String },
+    /// Invalid threshold value.
+    InvalidThreshold { threshold: u8, total_keys: usize },
+    /// Threshold cannot be zero.
+    ZeroThreshold,
```
storopoli: You can do `NonZero<u8>` using Rust's std library, see: https://doc.rust-lang.org/stable/std/num/struct.NonZero.html

I think we should do this.
We enforce impossible conditions (invariants) using the type system. We also advocate that in our [Engineering Best Practices](https://www.notion.so/Engineering-Rust-Best-Practices-14c901ba000f8025a974ec8e2a70b4af?source=copy_link#14d901ba000f80008606ff02b84a908a)

crates/crypto/src/threshold_signature/indexed/config.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// - `DuplicateAddMember`: The keys list contains duplicate members
+    /// - `ZeroThreshold`: The threshold is zero
+    /// - `InvalidThreshold`: The threshold exceeds the total number of keys
+    pub fn try_new(
+        keys: Vec<CompressedPublicKey>,
+        threshold: u8,
+    ) -> Result<Self, ThresholdSigningError> {
+        let mut config = ThresholdConfig {
+            keys: vec![],
+            threshold: 0,
```
storopoli: This is bad, why your placeholder `config` is already invalid with `0` as `threshold`. We should use `NonZero<u8>`.

crates/crypto/src/threshold_signature/indexed/config.rs
```diff
@@ -0,0 +1,297 @@
...
+
+        Ok(())
+    }
+}
+
+impl std::hash::Hash for CompressedPublicKey {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.serialize().hash(state);
+    }
+}
```
storopoli: Please import `use std::hash::Hash` in top of the file? Otherwise imports fly under the radar.

crates/crypto/src/threshold_signature/indexed/config.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    /// Check if there are no authorized signers.
+    pub fn is_empty(&self) -> bool {
+        self.keys.is_empty()
+    }
+
+    /// Validates that an update can be applied to this configuration.
+    pub fn validate_update(
+        &self,
+        update: &ThresholdConfigUpdate,
```
storopoli, line 79: Someone can totally skip this function.

This function should take a mutable `&mut self` and then change a (new I am proposing now) field in `ThresholdConfigUpdate` that is called `valid: bool` (of course it is `non-pub` and once this functions runs it then flips `valid` to `true`

crates/asm/txs/admin/Cargo.toml
```diff
@@ -15,7 +15,6 @@ strata-primitives.workspace = true
 
 arbitrary.workspace = true
 bitcoin = { workspace = true, optional = true }
-bitvec.workspace = true
```
storopoli, line 18: Should we not use this anymore @prajwolrg?

crates/crypto/src/threshold_signature/indexed/pubkey.rs
```diff
@@ -0,0 +1,173 @@
...
+pub struct CompressedPublicKey(PublicKey);
+
+impl CompressedPublicKey {
+    /// Create a new `CompressedPublicKey` from a byte slice.
+    ///
+    /// The slice must be exactly 33 bytes in compressed format (0x02 or 0x03 prefix).
+    pub fn from_slice(data: &[u8]) -> Result<Self, ThresholdSigningError> {
+        let pk =
+            PublicKey::from_slice(data).map_err(|e| ThresholdSigningError::InvalidPublicKey {
+                index: 0,
```
storopoli: Why are you hardcoding the index to `0`. If you don't have the `index` for certain `ThresholdSigningError::InvalidPublicKey` runtime errors, then make `index: Optional<u8>`?

crates/crypto/src/threshold_signature/indexed/pubkey.rs
```diff
@@ -0,0 +1,173 @@
...
+            })?;
+        Ok(Self(pk))
+    }
+
+    /// Get the inner `secp256k1::PublicKey`.
+    pub fn as_inner(&self) -> &PublicKey {
+        &self.0
+    }
+
+    /// Serialize to 33-byte compressed format.
```
storopoli, line 47: Add more information. You can just copy paste `secp256k1`'s docstrings on `Publickey::Serialize`:

> Serializes the key as a byte-encoded pair of values. In compressed form the y-coordinate is represented by only a single bit, as x determines it up to one bit.

crates/crypto/src/threshold_signature/indexed/pubkey.rs
```diff
@@ -0,0 +1,173 @@
...
+
+use super::ThresholdSigningError;
+
+/// A compressed secp256k1 public key (33 bytes).
+///
+/// This is a thin wrapper around `secp256k1::PublicKey` that adds Borsh
+/// serialization support. Unlike `EvenPublicKey`, this type does not
+/// enforce even parity - it accepts any valid compressed public key.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct CompressedPublicKey(PublicKey);
```
storopoli, line 27: This leaves in the `strata-crypto` crate.
Hence, I am deferring this decision to @AaronFeickert: should we do `Zeroize`/`ZeroizeOnDrop` for this wrapper around [`secp25k1::PublicKey`](https://docs.rs/secp256k1/latest/secp256k1/struct.PublicKey.html)?

Note that `secp256k1` library does not implement any zeroization method.

crates/crypto/src/threshold_signature/indexed/signature.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// Index of the signer in the ThresholdConfig keys array (0-255).
+    pub index: u8,
+    /// 65-byte recoverable ECDSA signature (recovery_id || r || s).
+    /// Using recoverable format for hardware wallet compatibility (Ledger/Trezor native output).
+    pub signature: [u8; 65],
```
storopoli: 👌🏻 love the docstrings

crates/crypto/src/threshold_signature/indexed/signature.rs
storopoli: We can start with `u8` for now and then do a breaking change in the future.

crates/crypto/src/threshold_signature/indexed/signature.rs
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    /// Get the r component (bytes 1-32).
+    pub fn r(&self) -> &[u8; 32] {
+        self.signature[1..33].try_into().unwrap()
+    }
+
+    /// Get the s component (bytes 33-64).
+    pub fn s(&self) -> &[u8; 32] {
+        self.signature[33..65].try_into().unwrap()
```
storopoli: No `unwrap`. Fine to do an `expect("IndexedSignature will always have a 65-length slice")`

crates/crypto/src/threshold_signature/indexed/signature.rs
storopoli: These fields should not be pub. You are already setting getters below right? Also for security ergononomics as well.

crates/crypto/src/threshold_signing/indexed_signatures/verification.rs
storopoli: I think it's `>`, but please double-check. I hate "off-by-one" errors.

crates/crypto/src/threshold_signing/indexed_signatures/verification.rs
```diff
@@ -0,0 +1,242 @@
...
+fn sign_ecdsa_recoverable(message_hash: &[u8; 32], secret_key: &secp256k1::SecretKey) -> [u8; 65] {
+    let message = Message::from_digest_slice(message_hash).expect("32 bytes");
+    let sig = SECP256K1.sign_ecdsa_recoverable(&message, secret_key);
+    let (recovery_id, compact) = sig.serialize_compact();
+
+    let mut result = [0u8; 65];
+    result[0] = recovery_id.to_i32() as u8;
+    result[1..65].copy_from_slice(&compact);
+    result
+}
```
storopoli: Move this to somewhere inside the `mod tests`.
Then you don't need to say that `This is a helper function for testing and creating signatures.` and you can drop the `#[cfg(test)]`.

crates/asm/subprotocols/admin/src/authority.rs
storopoli: Yes, if we are moving to this direction then we might do the trait-based thing with `PhantomData`.

crates/crypto/src/threshold_signature/indexed/pubkey.rs
irnb: This is already addressed in the current code. Lines 19-25 of pubkey.rs contain:
```
/// **Why no parity enforcement?** This key is used for ECDSA signature
/// verification (not Schnorr/BIP340). ECDSA signatures work with both
/// even and odd parity keys, unlike Schnorr which requires even parity
/// for x-only public keys.
```
Let me know if you'd like additional clarification in the docs.

crates/crypto/src/threshold_signature/indexed/pubkey.rs
irnb: Since public keys are not confidential data, what is the rationale for applying Zeroize to them?

crates/crypto/src/threshold_signature/indexed/signature.rs
irnb: The signer is responsible for providing their correct index when creating an `IndexedSignature`.

Config setup: `ThresholdConfig` has an ordered list of public keys (indices 0, 1, 2, ...)
Signing: Signer knows their index and includes it: `IndexedSignature { index, signature }`

crates/asm/subprotocols/admin/src/authority.rs
irnb: The explicit ECDSA implementation is intentional for this PR's scope. A few considerations:

> **Scope:** This PR is specifically about hardware wallet compatibility by moving from MuSig2 to ECDSA. 
> **No current need:** There's no other signature scheme being used for admin multisig. Adding abstraction for a single implementation would be premature optimization.
> **Complexity:** Making ThresholdConfig generic would cascade changes through the entire admin subprotocol 
> 

I've added a doc comment clarifying this design choice. If we need to support multiple signature schemes in the future, we can introduce the trait abstraction then with a clear use case driving the design.

crates/crypto/src/threshold_signature/indexed/config.rs
irnb: The `validate_update()` method is intentionally public for dry-run validation use cases  e.g., checking if an update would succeed before applying it, or showing validation errors to users before they submit. 

The "skip" concern is already addressed: `apply_update() `always calls `validate_update()` internally before making any changes. There's no way to mutate the config without going through `apply_update()`, which enforces validation.

is that sounds ok to you?

crates/crypto/src/threshold_signature/indexed/config.rs
```diff
@@ -31,11 +33,14 @@ impl ThresholdConfig {
         keys: Vec<CompressedPublicKey>,
         threshold: u8,
     ) -> Result<Self, ThresholdSignatureError> {
+        // Validate threshold is non-zero
+        let threshold = NonZero::new(threshold).ok_or(ThresholdSignatureError::ZeroThreshold)?;
```
storopoli: If you make `try_new` to accept a `threshold: NonZero<u8>` then you don't need the `ZeroThreshold` error variant.

storopoli (changes requested):

crates/crypto/src/test_utils/schnorr.rs
irnb: The diff is required for two reasons:

**Import path change:** The `multisig` module was renamed to `threshold_signature::musig2` in this PR, so the import had to be updated.
**Key generation fix:** The main branch code using `SecretKey::new(&mut OsRng)` doesn't compile `SecretKey::new() `requires the rand feature on `secp256k1`, which isn't enabled in the workspace. The verbose `from_slice` form is the fix.

The keys are not the same each run  `OsRng.fill_bytes() `generates different secure random bytes every time. Each of the 3 keys is unique and random.

crates/l1tx/src/utils.rs
```diff
@@ -11,7 +11,7 @@ use strata_asm_txs_bridge_v1::{
     withdrawal_fulfillment::WithdrawalFulfillmentInfo as BridgeV1WithdrawInfo,
 };
 use strata_asm_types::{DepositInfo, WithdrawalFulfillmentInfo};
-use strata_crypto::multisig::aggregate_schnorr_keys;
+use strata_crypto::threshold_signature::aggregate_schnorr_keys;
```
prajwolrg: It feels like `aggregate_schnorr_keys` should be outside of the `threshold_signature` module.

crates/crypto/src/threshold_signature/indexed/signature.rs
```diff
@@ -0,0 +1,256 @@
...
+}
+
+impl BorshDeserialize for IndexedSignature {
+    fn deserialize_reader<R: std::io::Read>(reader: &mut R) -> std::io::Result<Self> {
+        let index = u8::deserialize_reader(reader)?;
+        let mut signature = [0u8; 65];
+        reader.read_exact(&mut signature)?;
+        Ok(Self { index, signature })
+    }
+}
```
prajwolrg: I don't think we need a custom implementation of `BorshSerialize/BorshDeserialize`.

crates/crypto/src/threshold_signature/indexed/signature.rs
```diff
@@ -0,0 +1,256 @@
...
+    fn deserialize_reader<R: std::io::Read>(reader: &mut R) -> std::io::Result<Self> {
+        let count = u8::deserialize_reader(reader)?;
+        let mut signatures = Vec::with_capacity(count as usize);
+        for _ in 0..count {
+            signatures.push(IndexedSignature::deserialize_reader(reader)?);
+        }
+        SignatureSet::new(signatures)
+            .map_err(|e| std::io::Error::new(std::io::ErrorKind::InvalidData, e.to_string()))
+    }
+}
```
prajwolrg: I don't think we directly derive BorshSerialize/BorshDeserialize.

crates/crypto/src/threshold_signature/indexed/signature.rs
```diff
@@ -0,0 +1,256 @@
...
+/// # Hardware Wallet Compatibility
+///
+/// The first byte (header) can be in two formats:
+///
+/// 1. **Raw recovery ID** (0-3): Used by some signing libraries
+/// 2. **BIP-137 format** (27-42): Used by Bitcoin message signing in hardware wallets
+///    - 27-30: Uncompressed P2PKH
+///    - 31-34: Compressed P2PKH (most common for Ledger/Trezor)
+///    - 35-38: SegWit P2SH-P2WPKH
+///    - 39-42: Native SegWit P2WPKH
```
prajwolrg, line 22: nice! I didn't know about it. How did you realize this?

crates/crypto/src/threshold_signature/indexed/config.rs
```diff
@@ -0,0 +1,349 @@
...
+            ));
+        }
+
+        let threshold_u8 = u8::deserialize_reader(reader)?;
+        let threshold = NonZero::new(threshold_u8).ok_or_else(|| {
+            std::io::Error::new(std::io::ErrorKind::InvalidData, "threshold cannot be zero")
+        })?;
+        Ok(Self { keys, threshold })
+    }
+}
```
prajwolrg: I'm generally averse to manual implementation of the serialization/deserialization logic unless we have a good reason. 

I don't think DoS is an attack here since we will be reading this from bitcoin transaction which are limited to at max 4MB. If we cannot serialize NonZero variant, that might be an issue though.

crates/asm/txs/admin/Cargo.toml
prajwolrg: This was more of an optimization than a feature. `IndexedSignature` has an index in the struct. We can avoid this by removing index from the `IndexedSignature` (name would not make sense then), and then in the `SignatureSet` add a bitmap of the indices of the signers. This would also make it such that we don't require the validation of unique index in the `IndexedSignature`.

I'll leave it upto @irnb to decide what works best here.

crates/crypto/src/threshold_signing/indexed_signatures/verification.rs
irnb: **Recovery vs direct verification:** We use `recover_ecdsa` because hardware wallets produce BIP-137 recoverable signatures (65 bytes with recovery ID). Per the [secp256k1 library docs](https://github.com/bitcoin-core/secp256k1/blob/master/include/secp256k1_recovery.h): "Successful public key recovery guarantees that the signature passes verification." 

**Error granularity:** The signer index is sufficient, it tells us which signature failed. More detailed internal errors (recovery failed vs key mismatch) don't provide actionable information for us.

crates/crypto/src/threshold_signature/indexed/signature.rs
irnb: Previous experience with hardware wallets :smile: 

here's the spec: [BIP-137: Signatures of Messages using Private Keys](https://github.com/bitcoin/bips/blob/master/bip-0137.mediawiki).

crates/asm/txs/admin/Cargo.toml
irnb: Keeping the current design, it's simpler and the per-signature index overhead is minimal. 

We can revisit if space becomes a concern.

crates/crypto/src/threshold_signature/indexed/signature.rs
barakshani: I am not so happy about it. How would the signer know their own index?

storopoli (approved):
ACK 5cf0e391be1c6d3c34a22650ab97adf471cd97b8

From my side OK. Note that we have other review comments. Especially the trait-based one from @barakshani.

crates/crypto/src/threshold_signature/indexed/signature.rs
storopoli: Yes, I was also impressed with that.
Makes total sense to have a BIP for that.
I knew that we had a standard for signing messages, but I didn't knew how widely adopted it was by hardware wallet manufacturers.

crates/crypto/src/threshold_signature/indexed/signature.rs
irnb: The signer **does** know their own index, it's not something they have to guess.

**Here's how it works:**

* **Deterministic index:**
  `ThresholdConfig::keys()` is an ordered vector established `during admin setup`. That order (0, 1, 2, …) is part of the configuration we distribute off-chain to signers, along with their own public key. **(This is also available as part of the admin subprotocol.)**

* **How the signer knows their index:**
  When producing an `IndexedSignature`, the signer looks at the config they already have and picks their corresponding slot. There’s no ambiguity the config explicitly tells them their position.

* **Mis tagging is safely rejected:**
  If a signer accidentally uses the wrong index, verification will recover the pubkey from the signature and compare it against the key at that slot. If they don’t match, the signature is rejected with `InvalidSignature` or `SignerIndexOutOfBounds`.

Let me know if this still doesn’t make sense or if you’d like to explore an alternative approach, I can do that if needed.

barakshani: It makes sense, but feels a bit too cumbersome to me. Signers will download some app (the Admin tool) that they will have to run to post updates. The full details are still being worked out, but if we don't have to force them to add some config file, I'd rather us not follow this path. Note that admin set may change (though even if it changes, the order of existing admins doesn't need to change).

I would propose that filtering of correct pk is done on the ASM side. @prajwolrg what was the procedure on the previous implementation?

irnb: https://alpenlabs.atlassian.net/browse/STR-1946

crates/crypto/src/threshold_signature/indexed/pubkey.rs
AaronFeickert: If the use case will never result in a secret value, no need to implement it. But if the result could be secret (like for deriving Diffie-Hellman shared secrets, which share the same underlying type), it would be useful to implement at least `Zeroize`.

crates/crypto/src/threshold_signature/indexed/signature.rs
AaronFeickert: You're just saying that since an `unwrap` recently broke the internet...

prajwolrg (approved):

crates/crypto/src/threshold_signature/indexed/verification/ecdsa.rs
```diff
@@ -0,0 +1,191 @@
...
+        39..=42 => header - 39, // Native SegWit
+        _ => return Err(ThresholdSignatureError::InvalidSignatureFormat),
+    };
+    Ok(recid as i32)
+}
+
+/// Verifies each ECDSA signature in the set against the corresponding public key.
+///
+/// This function performs the actual ECDSA signature recovery and verification.
+/// It assumes the SignatureSet has already been validated for duplicates.
```
AaronFeickert: This comment seems a bit off. Is it not the case that a `SignatureSet` is already guaranteed not to contain duplicates? If so, this function's caller does not need to concern itself with this.

crates/crypto/src/threshold_signature/indexed/signature.rs
```diff
@@ -0,0 +1,219 @@
...
+        signatures.sort_by_key(|s| s.index);
+
+        // Check for duplicate indices
+        for window in signatures.windows(2) {
+            if window[0].index == window[1].index {
+                return Err(ThresholdSignatureError::DuplicateSignerIndex(
+                    window[0].index,
+                ));
+            }
+        }
```
AaronFeickert: Recommend using an internal hash set to check for duplicates instead. It's a very clean approach.

crates/crypto/src/threshold_signature/indexed/signature.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub fn compact(&self) -> [u8; 64] {
+        let mut compact = [0u8; 64];
+        compact.copy_from_slice(&self.signature[1..65]);
+        compact
+    }
+}
+
+/// A set of indexed ECDSA signatures for threshold verification.
+///
+/// Signatures are sorted by index and must not contain duplicates.
```
AaronFeickert: This comment seems a bit wonky. Since the constructor enforces uniqueness, why not change to indicate that a `SignatureSet` _will not contain_ duplicates?

AaronFeickert (changes requested):

crates/crypto/src/threshold_signature/indexed/verification/ecdsa.rs
```diff
@@ -0,0 +1,191 @@
...
+        35..=38 => header - 35, // SegWit P2SH
+        39..=42 => header - 39, // Native SegWit
+        _ => return Err(ThresholdSignatureError::InvalidSignatureFormat),
+    };
+    Ok(recid as i32)
+}
+
+/// Verifies each ECDSA signature in the set against the corresponding public key.
+///
+/// This function performs the actual ECDSA signature recovery and verification.
```
barakshani:
```suggestion
/// This function performs pk extraction from the ECDSA signature, and matches with the config.

File: crates/crypto/src/threshold_signature/indexed/verification/ecdsa.rs [hunk 36]
```diff
@@ -0,0 +1,191 @@
...
+        // Normalize the recovery ID from BIP-137 header format to raw 0-3
+        let recid_raw = normalize_recovery_id(indexed_sig.recovery_id())?;
+        let recovery_id = RecoveryId::from_i32(recid_raw)
+            .map_err(|_| ThresholdSignatureError::InvalidSignatureFormat)?;
+
+        let recoverable_sig =
+            RecoverableSignature::from_compact(&indexed_sig.compact(), recovery_id)
+                .map_err(|_| ThresholdSignatureError::InvalidSignatureFormat)?;
+
+        // Hardware wallets emit recoverable signatures (with a header). Recover first so we
```
barakshani:
```suggestion
        // Hardware wallets support recoverable public key from signatures. Recover first so we

File: crates/asm/subprotocols/admin/src/authority.rs [hunk 37]
```diff
@@ -1,62 +1,61 @@
...
+use strata_crypto::threshold_signature::{
+    SignatureSet, ThresholdConfig, ThresholdSignatureError, verify_threshold_signatures,
 };
 use strata_primitives::roles::Role;
 
-/// Manages multisignature operations for a given role and key set, with replay protection via a
-/// seqno.
-#[derive(Clone, Debug, Eq, PartialEq, Arbitrary, BorshDeserialize, BorshSerialize)]
+/// Manages threshold signature operations for a given role and key set, with replay protection via
+/// a seqno.
```
barakshani:
```suggestion
/// a sequence number.

File: crates/asm/subprotocols/admin/src/authority.rs [hunk 38]
```diff
@@ -1,62 +1,61 @@
...
     }
 
-    /// Verify that `signature` is a valid threshold signature for `action` under the current config
-    /// and seqno.
+    /// Verifies a set of ECDSA signatures against a threshold configuration.
     ///
-    /// Uses the generic multisig verification function to orchestrate the workflow.
+    /// This function is intentionally ECDSA-specific as part of the hardware wallet
+    /// compatibility design (BIP-137 format support). A trait-based abstraction
+    /// could be added in the future if multiple signature schemes are needed.
```
barakshani: Convert to a comment rather than a docstring.

crates/asm/txs/admin/src/test_utils/mod.rs
```diff
@@ -8,101 +8,106 @@ use bitcoin::{
...
+use strata_primitives::buf::Buf32;
 
 pub(crate) const TEST_MAGIC_BYTES: &[u8; 4] = b"ALPN";
 
-use crate::{actions::MultisigAction, constants::ADMINISTRATION_SUBPROTOCOL_ID};
+use crate::{
+    actions::MultisigAction, constants::ADMINISTRATION_SUBPROTOCOL_ID, parser::SignedPayload,
+};
+
+/// Creates an ECDSA recoverable signature for a message hash.
```
barakshani:
```suggestion
/// Creates an ECDSA signature with recoverable pk for a message hash.

File: crates/crypto/src/threshold_signature/indexed/mod.rs [hunk 40]
```diff
@@ -0,0 +1,17 @@
+//! Individual ECDSA signature set for threshold signatures (M-of-N).
```
barakshani: What does "individual" mean here? I think we can omit this word.

crates/crypto/src/threshold_signature/indexed/mod.rs
```diff
@@ -0,0 +1,17 @@
...
+//!
+//! This module provides types and functions for verifying a set of individual
```
barakshani: ditto

crates/crypto/src/threshold_signature/indexed/signature.rs
```diff
@@ -0,0 +1,219 @@
+//! Signature types for threshold signing.
+
+use borsh::{BorshDeserialize, BorshSerialize};
+
+use super::ThresholdSignatureError;
+
+/// An individual ECDSA signature with its signer index.
```
barakshani:
```suggestion
/// An ECDSA signature with its signer index.

File: crates/crypto/src/threshold_signature/indexed/signature.rs [hunk 43]
```diff
@@ -92,17 +94,12 @@ pub struct SignatureSet {
...
-        signatures.sort_by_key(|s| s.index);
-
-        // Check for duplicate indices
-        for window in signatures.windows(2) {
-            if window[0].index == window[1].index {
-                return Err(ThresholdSignatureError::DuplicateSignerIndex(
-                    window[0].index,
-                ));
+    /// The signatures will be checked for duplicates.
+    pub fn new(signatures: Vec<IndexedSignature>) -> Result<Self, ThresholdSignatureError> {
```
AaronFeickert, line 98: This change means the signatures are no longer sorted by key in the resulting struct. Is such sorting assumed elsewhere?

irnb: Actually, we don’t need the keys to be sorted by index, and there’s no assumption anywhere else.

AaronFeickert (approved):

PR #1166, opened 2025-11-26
STR-1684-ee-p-2-p-broadcast-gossip-of-head-block-hash-v2

storopoli:
## Description

Custom RLPx subprotocol to gossip the head block hash in `alpen-reth` and `reth-node`.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Done with the help of Cursor.

Based on [`reth`'s custom RLPx subprotocol example](https://github.com/paradigmxyz/reth/tree/v1.8.2/examples/custom-rlpx-subprotocol)

I have no idea why Nix CI is failing. I think it's because of a bad cache...

## Checklist

## Related Issues

STR-1684

### DIFF
--- a/functional-tests/envs/testenv.py
+++ b/functional-tests/envs/testenv.py
@@ -4,6 +4,7 @@ import logging
 import time
 
 import flexitest
+import requests as req
 
 from envs.rollup_params_cfg import RollupConfig
 from factory.alpen_client import AlpenCliBuilder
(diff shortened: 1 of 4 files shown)


bin/alpen-reth/src/main.rs
```diff
@@ -83,11 +95,84 @@ fn main() {
...
+                                    connections.remove(&peer_id);
+                                }
+                                HeadGossipEvent::HeadHash { peer_id, header } => {
+                                    // TODO(@storopoli): do something when we receive a single `Header`.
+                                    info!(target: "head-gossip", "Received head hash from peer {}: {:?}", peer_id, header.hash_slow());
+                                }
+                                HeadGossipEvent::HeadHashes { peer_id, headers } => {
+                                    // TODO(@storopoli): do something when we receive multiple `Header`s.
+                                    info!(target: "head-gossip", "Received {} head hashes from peer {}", headers.len(), peer_id);
+                                }
```
storopoli, line 137: @sapinb, these two TODOs are what to do when receiving `HeadHash{es}`.
Probably tackled in another PR.

bin/alpen-reth/src/main.rs
```diff
@@ -83,11 +95,84 @@ fn main() {
...
+            // This task will handle gossip events
+            let mut state_events = handle.node.provider.subscribe_to_canonical_state();
+            tokio::spawn(async move {
+                let mut connections: HashMap<PeerId, mpsc::UnboundedSender<HeadGossipCommand>> =
+                    HashMap::new();
+
+                loop {
+                    tokio::select! {
+                        Some(event) = gossip_rx.recv() => {
+                            match event {
```
delbonis, line 119: I don't think this should be in `main`, can we not split it out into a separate (service?) crate so that it could be reusable or at least modularized?

crates/reth/node/src/head_gossip/connection.rs
```diff
@@ -0,0 +1,159 @@
...
+    events: mpsc::UnboundedSender<HeadGossipEvent>,
+}
+
+impl Drop for HeadGossipConnection {
+    fn drop(&mut self) {
+        self.events
+            .send(HeadGossipEvent::Closed {
+                peer_id: self.peer_id,
+            })
+            .ok();
```
delbonis, line 105: Is there a reason we can't just interpret the channel being closed as a graceful shutdown?

crates/reth/node/src/head_gossip/handler.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// Channel for sending events to the node.
+    pub events: mpsc::UnboundedSender<HeadGossipEvent>,
+}
+
+/// The protocol handler for head gossip.
+#[derive(Debug)]
+pub struct HeadGossipProtocolHandler {
+    /// State of the head gossip protocol.
+    pub state: HeadGossipState,
+}
```
delbonis, line 64: avoid `pub`!

crates/reth/node/src/head_gossip/connection.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+}
+
+impl Stream for HeadGossipConnection {
+    type Item = BytesMut;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
```
delbonis, line 112: *might* want to have some tests on this just to be careful since it's a manual `Stream` impl

crates/reth/node/src/head_gossip/handler.rs
```diff
@@ -0,0 +1,84 @@
...
+    },
+
+    /// Multiple head hashes were received from a peer.
+    HeadHashes {
+        /// Peer that we received the new head hashes.
+        peer_id: PeerId,
+
+        /// Received [`Header`]s.
+        headers: Vec<Header>,
+    },
```
delbonis, line 49: Can these struct variants be split out as standalone structs?

crates/reth/node/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 //! Reth node implementation for the Alpen EE.
 mod engine;
 mod evm;
+pub mod head_gossip;
```
delbonis, line 4: is this necessary?

delbonis (changes requested):
Overall looks pretty good, main comment is on that big loop in `main.rs`.  I don't know enough of the Reth internals to be sure this makes the most sense but I don't see any major red flags.

bin/alpen-reth/src/main.rs
storopoli: NOTE: This should not go here, but here
https://github.com/alpenlabs/alpen/blob/b26ec1267986639b737a5ba91cc20bb3d826f9f6/bin/alpen-client/src/main.rs#L109-L117

crates/reth/node/src/head_gossip/handler.rs
```diff
@@ -0,0 +1,84 @@
...
+}
```
storopoli, line 50: We should create a wrapper struct that holds the `HeadHash` as one of the fields in case we need to add more stuff to be gossiped, and we should also add a `nonce`/`seq_no` (`u64`) to resolve conflicts, which in doubt choose the highest one. Also we should have some sort of authentication from the sequencer, probably a signature.

Discussed with Sapin in a meeting today, Cc @delbonis @bewakes

storopoli:
I had a meeting with Sapin to discuss the task and I am closing this in favor of #1170. I considered all review comments here (from @delbonis) and incorporated them in #1170. Especially the new "war on `pub`" stuff.

PR #1168, opened 2025-11-27
STR-1783: EE Startup check Reth DB consistency

storopoli:
## Description

Implements EE (reth) consistency recovery to handle block loss during sequencer restarts. When reth loses blocks after forced exits, the sequencer detects missing blocks on startup and resubmits them using stored execution payloads.
Changes:

- Add `SequencerDatabase` trait and Sled implementation for exec payload storage
- Persist exec payloads in `TemplateManagerHandle` when completing block templates
- Add startup consistency check that compares stored payloads with Reth and recovers missing blocks
- Add functional test simulating Reth data loss via snapshot/restore

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Done with the aide of Cursor using Claude Opus 4.5

- Storage overhead only affects sequencer nodes
- Recovery runs synchronously before other sequencer tasks start
- Uses exponential backoff retries for engine calls

Please review it commit by commit I guarantee it will be a delightful experience :)

## Checklist

## Related Issues

STR-1783

### DIFF
--- a/functional-tests/tests/ee_consistency_recovery.py
+++ b/functional-tests/tests/ee_consistency_recovery.py
@@ -1,0 +1,155 @@
+import flexitest
+
+from envs import testenv
+
+
+@flexitest.register
+class EeConsistencyRecoveryTest(testenv.StrataTestBase):
+    """
+    Tests that the sequencer can recover missing blocks in reth on startup.
+
+    The scenario simulates reth losing recent blocks due to consistency issues
+    during forced exits. The sequencer should detect this on startup and replay
+    the missing blocks using stored exec payloads.
+    """
+
+    def __init__(self, ctx: flexitest.InitContext):
+        ctx.set_env(testenv.BasicEnvConfig(110))
+
+    def main(self, ctx: flexitest.RunContext):
+        seq = ctx.get_service("sequencer")
+        seqrpc = seq.create_rpc()
+        reth = ctx.get_service("reth")
+        rethrpc = reth.create_rpc()
+
+        seq_waiter = self.create_strata_waiter(seqrpc)
+        reth_waiter = self.create_reth_waiter(rethrpc)
+
+        # Wait for genesis
+        seq_waiter.wait_until_genesis()
+        self.info("Genesis occurred, waiting for initial blocks...")
+
+        # Wait for some blocks to be produced
+        initial_blocks = 5
+        seq_waiter.wait_until_chain_tip_exceeds(initial_blocks)
+        self.info(f"Chain tip exceeded {initial_blocks} blocks")
+
+        # Get current state before snapshot
+        initial_eth_block = reth_waiter.get_current_block_number()
+        self.info(f"Initial ETH block number: {initial_eth_block}")
+
+        # Take a snapshot of reth's data directory
+        # This simulates the state before some blocks are lost
+        reth.stop()
+        self.info("Stopped reth, taking snapshot...")
+        reth.snapshot_datadir(1)
+
+        # Restart reth to continue producing blocks
+        reth.start()
+        rethrpc = reth.create_rpc()
+        reth_waiter = self.create_reth_waiter(rethrpc)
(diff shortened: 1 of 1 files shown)


bin/strata-client/src/main.rs
delbonis, line 1: Does this have to be in `main.rs`?

bin/strata-client/src/rpc_server.rs
delbonis, line 1: Use the ops/manager pattern, *don't* make the RPC server generic over the database interface so you can make the template manager generic over the database interface.  That's the low level interface that's supposed to be wrapped in the nicer mid-level interfaces.  *Also*, those are blocking database calls but the RPC server is all `async`, so you'd be blocking the executor.

crates/sequencer/src/block_template/handle.rs
```diff
@@ -105,6 +116,25 @@ impl TemplateManagerHandle {
...
             .await?;
 
+        // Persist exec payload to sequencer db for EE consistency recovery.
+        // This is done before saving to L2BlockManager to ensure the payload is persisted
+        // even if subsequent operations fail.
+        let slot = block_bundle.header().slot();
+        let exec_payload = block_bundle.accessory().exec_payload().to_vec();
+        if let Err(e) = self
+            .sequencer_db
+            .put_exec_payload(slot, template_id, exec_payload)
```
delbonis: Blocking the executor!

crates/db/types/src/traits.rs
```diff
@@ -351,3 +352,36 @@ pub trait L1BroadcastDatabase: Send + Sync + 'static {
...
     fn get_last_tx_entry(&self) -> DbResult<Option<L1TxEntry>>;
 }
+
+/// Sequencer-specific database for storing exec payloads.
+///
+/// This database stores execution payloads produced by the sequencer to enable
+/// recovery in case reth loses blocks on restart due to consistency checks.
+/// On sequencer startup, blocks stored here are treated as canonical and any
+/// missing blocks in reth will be resubmitted using the stored payloads.
+pub trait SequencerDatabase: Send + Sync + 'static {
```
delbonis, line 362: I thought we already had a trait for this?

storopoli: I could not find it. The only `SequencerDatabase` apart from what I am adding in this PR is by @sapinb 2 weeks ago here: https://github.com/alpenlabs/alpen/blob/6f3cf4cfac9cba7425802d9cf292706ea3e0d0c6/crates/storage-common/src/exec.rs#L48-L56

bin/strata-client/src/main.rs
storopoli: What do you mean with "this"? the `check_ee_consistency_and_recover`?

bin/strata-client/src/rpc_server.rs
storopoli: Removed the changes in `bin/strata-client/src/rpc_server.rs`

crates/sequencer/src/block_template/handle.rs
storopoli: The best I could do was a `tokio`'s `spawn_blocking` in dbd4e22ae989bc2488d2f4ec2c92a0de435685ed

bin/strata-client/src/rpc_server.rs
storopoli: See cff1fb858c4302560080006ecbcc3a36a0c88283

bin/strata-client/src/main.rs
sapinb, line 1: @storopoli This was meant to be something equivalent to `sync_chainstate_to_el` for the new evm ee client.

https://github.com/alpenlabs/alpen/blob/79a811052ea295c00d935d83f2c1a9cf92e9e4eb/bin/strata-client/src/el_sync.rs#L27

This should not require any changes on the `strata-client` side.

sapinb (changes requested):
The consistency checks should on the new ee client.

PR #1169, opened 2025-11-28
ee block assembly

sapinb:
## Description
Core logic for EE block assembly.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

There is a dependency to #1127 during building exec block package. This will be resolved in a followup PR after both are merged.

```
    // TODO: get using `EvmExecutionEnvironment`
    let raw_block_encoded_hash = [0u8; 32];
```

## Checklist

### DIFF
--- a/crates/alpen-ee/block-assembly/src/block.rs
+++ b/crates/alpen-ee/block-assembly/src/block.rs
@@ -10,6 +10,7 @@ use strata_snark_acct_types::MessageEntry;
 
 use crate::{package::build_block_package, payload::build_exec_payload};
 
+/// All inputs that control the next built block.
 #[derive(Debug)]
 pub struct BlockAssemblyInputs {
     /// EeAccountState of last block.
@@ -27,13 +28,18 @@ pub struct BlockAssemblyInputs {
     pub bridge_gateway_account_id: AccountId,
 }
 
+/// Outputs from block assembly
 #[derive(Debug)]
 pub struct BlockAssemblyOutputs {
+    /// Block package representing the OL inputs and outputs for this block.
     pub package: ExecBlockPackage,
+    /// Block payload including full exec block body.
     pub payload: AlpenBuiltPayload,
+    /// EeAccountState after applying the new block.
     pub account_state: EeAccountState,
 }
 
+/// Builds the next block using `inputs` and `payload_builder`.
 pub async fn build_next_exec_block<E: PayloadBuilderEngine<AlpenBuiltPayload>>(
     inputs: BlockAssemblyInputs,
     payload_builder: &E,
--- a/crates/alpen-ee/block-assembly/src/lib.rs
+++ b/crates/alpen-ee/block-assembly/src/lib.rs
@@ -1,4 +1,5 @@
-#![warn(unused_crate_dependencies, reason = "wip")]
+//! Handles assembly of EE blocks.
+
 mod block;
 mod package;
 mod payload;
--- a/crates/alpen-ee/block-assembly/src/package.rs
+++ b/crates/alpen-ee/block-assembly/src/package.rs
@@ -12,6 +12,7 @@ use strata_msg_fmt::{Msg as MsgTrait, OwnedMsg};
 use strata_ol_msg_types::{WithdrawalMsgData, WITHDRAWAL_MSG_TYPE_ID};
 use tracing::warn;
 
+/// Builds the block package based on execution inputs and results.
 pub(crate) fn build_block_package(
     bridge_gateway_account_id: AccountId,
     parsed_inputs: Vec<MsgData>,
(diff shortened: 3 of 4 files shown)


crates/alpen-ee/block-assembly/src/block.rs
```diff
@@ -0,0 +1,80 @@
...
+#[derive(Debug)]
+pub struct BlockAssemblyInputs {
+    /// EeAccountState of last block.
+    pub account_state: EeAccountState,
+    /// New inbox messages to be included in this block.
+    /// Can be empty.
+    pub inbox_messages: Vec<MessageEntry>,
+    /// Exec blkid of previous block.
+    pub parent_exec_blkid: Hash,
+    /// Timestamp of next block to be built in ms.
```
barakshani, line 23: you have here "last block", "this block" and "next block" (you actually even have "previous") -- is it _next_ or _this_?

crates/alpen-ee/block-assembly/src/block.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub timestamp_ms: u64,
+    /// Max number of deposits to process per block.
+    pub max_deposits_per_block: NonZero<u8>,
+    /// Account id for bridge gateway on ol.
+    pub bridge_gateway_account_id: AccountId,
```
barakshani, line 28: why aren't these global? i.e. outside of this struct

barakshani (commented):
How is reth block building integrating into this?

In particular, we limit deposits, I assume for "gas limitation" purposes:
1. do we always limit deposits to a fix amount independent of how much gas is spent?
2. should we, for internal computation, assume worth case full reth block (36M gas) + max deposits? Right now I don't know how we dealt with it.
3. shouldn't we have some priority for L1 txes over L2 txes (for example, when we have forced inclusion)? I would imagine we first consume all L1 txes -- see how much computation/gas it taxes -- before we start processing L2 txes.

Please clarify

crates/alpen-ee/block-assembly/src/block.rs
sapinb: Will clarify the comments and names. It is only last block and next block being built. current/this becomes confusing.

crates/alpen-ee/block-assembly/src/block.rs
sapinb: They can be made global and used to construct this input. This PR deals with only the libs that will be called to do the block assembly. The parts that call this will be in another PR. Those will use the globals, or configs where these are provided.

crates/alpen-ee/common/src/types/mod.rs
```diff
@@ -1,3 +1,4 @@
 pub mod consensus_heads;
 pub mod ee_account_state;
 pub mod ol_chain_status;
+pub mod payload_builder;
```
delbonis, line 4: pls make these non-pub if possible

crates/alpen-ee/block-assembly/src/block.rs
delbonis: Also they shouldn't be *globals*, they should be in a `Params` struct that's passed around in a more structured way.

crates/alpen-ee/block-assembly/src/block.rs
```diff
@@ -0,0 +1,74 @@
...
+pub struct BlockAssemblyInputs {
+    /// EeAccountState of last block.
+    pub account_state: EeAccountState,
+    /// New inbox messages to be included in this block.
+    /// Can be empty.
+    pub inbox_messages: Vec<MessageEntry>,
+    /// Exec blkid of previous block.
+    pub parent_exec_blkid: Hash,
+    /// Timestamp of next block to be built in ms.
+    pub timestamp_ms: u64,
```
storopoli, line 24: You already specified in the docstring that is in milliseconds.
```suggestion
    pub timestamp: u64,
```

storopoli (changes requested):
One minor nit. Agree with everyone's comments.
Also c2ac6356e4a4d47319b2bc0c86a2a0a9b08fc7cf can be squashed into  b548fab0daaca9e60deae9969a35154c17df9b6b for simplicity...

crates/alpen-ee/block-assembly/src/block.rs
barakshani: wait, you're blocking merge on this? I'm not sure that we have this privilege

sapinb: I prefer to have it more explicit, hence the _ms.

storopoli: > wait, you're blocking merge on this? I'm not sure that we have this privilege

This is nuanced. Since there were some changes requested by Trey, I've added this small minor nit. I am monitoring the PR. If somehow it is approved I'd approve without having this addressed.

Ok @sapinb solving this

PR #1170, opened 2025-11-28
feat: EE custom block header gossip

storopoli:
## Description

Custom RLPx subprotocol to gossip the head block header in `alpen-client` and `reth-node`.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Done with the help of Cursor.

Based on [`reth`'s custom RLPx subprotocol example](https://github.com/paradigmxyz/reth/tree/v1.8.2/examples/custom-rlpx-subprotocol)

## Checklist

## Related Issues

STR-1684

### DIFF
--- a/bin/alpen-client/src/gossip.rs
+++ b/bin/alpen-client/src/gossip.rs
@@ -22,9 +22,12 @@ pub(crate) struct GossipConfig {
     /// Sequencer's public key for signature validation.
     pub sequencer_pubkey: Buf32,
 
+    /// Whether the local node should produce and sign gossip messages.
+    pub sequencer_enabled: bool,
+
     /// Sequencer's private key for signing (only in sequencer mode).
     #[cfg(feature = "sequencer")]
-    pub sequencer_privkey: Buf32,
+    pub sequencer_privkey: Option<Buf32>,
 }
 
 /// Handles a gossip event (connection established/closed or package received).
@@ -196,6 +199,10 @@ fn broadcast_new_block(
     connections: &HashMap<PeerId, mpsc::UnboundedSender<AlpenGossipCommand>>,
     config: &GossipConfig,
 ) {
+    if !config.sequencer_enabled {
+        return;
+    }
+
     info!(
         target: "alpen-gossip",
         block_hash = ?tip.hash_slow(),
@@ -206,6 +213,14 @@ fn broadcast_new_block(
 
     #[cfg(feature = "sequencer")]
     {
+        let Some(sequencer_privkey) = config.sequencer_privkey else {
+            error!(
+                target: "alpen-gossip",
+                "Sequencer mode enabled but no private key configured; skipping broadcast"
+            );
+            return;
+        };
+
         let msg = AlpenGossipMessage::new(
             tip.clone(),
             // NOTE: we use the block number as the sequence number
@@ -213,7 +228,7 @@ fn broadcast_new_block(
             //       provides monotonic, unique sequence numbers for gossip messages.
             tip.number,
         );
-        let pkg = msg.into_package(config.sequencer_pubkey, config.sequencer_privkey);
+        let pkg = msg.into_package(config.sequencer_pubkey, sequencer_privkey);
 
         for (peer_id, sender) in connections {
             if sender
--- a/bin/alpen-client/src/main.rs
+++ b/bin/alpen-client/src/main.rs
@@ -29,7 +29,7 @@ use strata_acct_types::AccountId;
 use strata_identifiers::{CredRule, OLBlockId};
 use strata_primitives::buf::Buf32;
 use tokio::sync::{broadcast, mpsc};
-use tracing::info;
+use tracing::{error, info};
 
 use crate::{
     genesis::ee_genesis_block_info,
@@ -149,22 +149,40 @@ fn main() {
                     let state_events = node.provider.subscribe_to_canonical_state();
 
                     // Parse sequencer private key from environment variable (only in sequencer mode)
-                    #[cfg(feature = "sequencer")]
                     let gossip_config = {
-                        let privkey_str = env::var("SEQUENCER_PRIVATE_KEY")
-                            .map_err(|_| eyre::eyre!("SEQUENCER_PRIVATE_KEY environment variable is required when sequencer feature is enabled"))?;
-                        let sequencer_privkey = privkey_str
-                            .parse::<Buf32>()
-                            .map_err(|e| eyre::eyre!("Failed to parse SEQUENCER_PRIVATE_KEY as hex: {e}"))?;
-                        GossipConfig {
-                            sequencer_pubkey: ext.sequencer_pubkey,
-                            sequencer_privkey,
-                        }
-                    };
+                        #[cfg(feature = "sequencer")]
+                        {
+                            let sequencer_privkey = if ext.sequencer {
+                                let privkey_str = env::var("SEQUENCER_PRIVATE_KEY").map_err(|_| {
+                                    eyre::eyre!(
+                                        "SEQUENCER_PRIVATE_KEY environment variable is required when running with --sequencer"
+                                    )
+                                })?;
+                                Some(
+                                    privkey_str.parse::<Buf32>().map_err(|e| {
+                                        eyre::eyre!(
+                                            "Failed to parse SEQUENCER_PRIVATE_KEY as hex: {e}"
+                                        )
+                                    })?,
+                                )
+                            } else {
+                                None
+                            };
 
-                    #[cfg(not(feature = "sequencer"))]
-                    let gossip_config = GossipConfig {
-                        sequencer_pubkey: ext.sequencer_pubkey,
+                            GossipConfig {
+                                sequencer_pubkey: ext.sequencer_pubkey,
+                                sequencer_enabled: ext.sequencer,
+                                sequencer_privkey,
+                            }
+                        }
+
+                        #[cfg(not(feature = "sequencer"))]
+                        {
+                            GossipConfig {
+                                sequencer_pubkey: ext.sequencer_pubkey,
+                                sequencer_enabled: false,
+                            }
+                        }
                     };
 
                     let gossip_task = create_gossip_task(gossip_rx, state_events, preconf_tx, gossip_config);
@@ -224,6 +242,11 @@ pub struct AdditionalConfig {
     #[arg(long, required = false)]
     pub db_retry_count: Option<u16>,
 
+    /// Run the node as a sequencer. Requires the `sequencer` feature and a
+    /// `SEQUENCER_PRIVATE_KEY` environment variable.
+    #[arg(long, default_value_t = false)]
+    pub sequencer: bool,
+
     /// Sequencer's public key (hex-encoded, 32 bytes) for signature validation.
     #[arg(long, required = true, value_parser = parse_buf32)]
     pub sequencer_pubkey: Buf32,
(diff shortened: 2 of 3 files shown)


bin/alpen-client/src/main.rs
```diff
@@ -131,6 +161,125 @@ fn main() {
...
+                                        "Received gossip message"
+                                    );
+
+                                    // Forward the block hash to engine control task for fork choice update
+                                    let hash = Hash::from(block_hash.0);
+                                    if preconf_tx.send(hash).is_err() {
+                                        warn!(
+                                            target: "alpen-gossip",
+                                            "Failed to forward block hash to engine control (no receivers)"
+                                        );
```
storopoli: @sapinb is this how we send the block hash to the underlying reth node?

crates/reth/node/src/gossip/package.rs
```diff
@@ -0,0 +1,273 @@
...
+pub struct AlpenGossipMessage {
+    /// Block [`Header`].
+    header: Header,
+
+    /// Sequence number.
+    seq_no: u64,
+
+    /// Sequencer's signature.
+    signature: Buf64,
+}
```
storopoli, line 44: @sapinb Is this what you intended with the wrapper struct?
We can add more things as needed.
There's a `seq_no` for deciding on which block header to follow (still todo implement) and a `signature` to validate that it came from the sequencer.

bin/alpen-client/src/main.rs
```diff
@@ -131,6 +161,125 @@ fn main() {
...
+                                    debug!(
+                                        target: "alpen-gossip",
+                                        %peer_id,
+                                        "Gossip connection closed"
+                                    );
+                                    connections.remove(&peer_id);
+                                }
+                                AlpenGossipEvent::Message { peer_id, message } => {
+                                    // TODO(@storopoli): validate signature before processing.
+                                    //                   get the sequencer's public key from the config somehow.
```
storopoli: @sapinb Should we add the sequencer's public key to the `alpen-client` config?

crates/reth/node/src/gossip/message.rs
```diff
@@ -0,0 +1,273 @@
...
+
+    /// Gets the signature of the [`AlpenGossipMessage`].
+    pub fn signature(&self) -> &Buf64 {
+        &self.signature
+    }
+
+    /// Validates the signature of the [`AlpenGossipMessage`] given a public key.
+    pub fn validate_signature(&self, _public_key: &Buf32) -> bool {
+        unimplemented!("Signature validation is not implemented yet");
+    }
```
storopoli: Should we do this here? We might need to import `strata-crypto`.

bin/alpen-client/src/main.rs
```diff
@@ -131,6 +161,125 @@ fn main() {
...
+            // Subscribe to canonical state notifications for broadcasting new blocks
+            let mut state_events = handle.node.provider.subscribe_to_canonical_state();
+
+            // Spawn the gossip event handling task
+            let mut gossip_rx = gossip_rx;
+            spawn(async move {
+                let mut connections: HashMap<PeerId, mpsc::UnboundedSender<AlpenGossipCommand>> =
+                    HashMap::new();
+
+                loop {
```
delbonis: This is a *huge* block and feels like the meat of it should be moved into a separate module.

storopoli: Check 6b20823bc11da043f464a5076d96411481eeba08
(I've also rebased)

bin/alpen-client/src/main.rs
sapinb: yes. That will be required.

crates/reth/node/src/gossip/package.rs
sapinb: This is in line with my thoughts. Just one question: 
What data does the signature sign ? 
Everything in the AlpenGossipMessage with signature field blanked ?
Or do we need an inner container ?

storopoli: Oh fuck yeah, good point.
We should do an inner struct, that we hash and the sequencer signs.
Since this is EVM context, we should probably use Keccak-256.

storopoli: Ok I've added `AlpenGossipPackage` that has an `AlpenGossipMessage`, a public key and a signature in 22b3b11c9e52c5fff88f87818078bb9f597c6e25

bin/alpen-client/src/main.rs
storopoli: Ok done in a8f1b1d164a9df10fc70639eb396e723c3113c96

crates/reth/node/src/gossip/message.rs
storopoli: Done in 22b3b11c9e52c5fff88f87818078bb9f597c6e25

bin/alpen-client/src/gossip.rs
```diff
@@ -0,0 +1,218 @@
...
+                                        }
+                                    }
+                                }
+
+                                #[cfg(not(feature = "sequencer"))]
+                                {
+                                    warn!(
+                                        target: "alpen-gossip",
+                                        "Cannot broadcast: sequencer feature not enabled"
+                                    );
```
sapinb: The logs for any non-sequencer node are going to be full of this warning.
Can we just skip this ?

bin/alpen-client/src/gossip.rs
```diff
@@ -0,0 +1,218 @@
...
+                            if let Some(tip) = headers.last() {
+                                info!(
+                                    target: "alpen-gossip",
+                                    block_hash = ?tip.hash_slow(),
+                                    block_number = tip.number,
+                                    peer_count = connections.len(),
+                                    "Broadcasting new block to peers"
+                                );
+
+                                #[cfg(feature = "sequencer")]
```
sapinb: Is there a benefit to feature gating sequencer related code ?
Previously, this has just been a config.

bin/alpen-client/src/gossip.rs
```diff
@@ -0,0 +1,218 @@
...
+                                ?block_hash,
+                                "Package already seen, skipping"
+                            );
+                            continue;
+                        }
+                        seen_packages.insert(block_hash);
+
+                        // Cleanup seen set if it grows too large
+                        if seen_packages.len() > config.max_seen_packages {
+                            seen_packages.clear();
```
sapinb: We just clear the entire set ?
Can this cause network spikes around the max_seen_packages boundaries ?

crates/reth/node/src/gossip/package.rs
```diff
@@ -0,0 +1,359 @@
...
+const SEQ_NO_SIZE: usize = mem::size_of::<u64>();
+
+/// Gossip message types.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct AlpenGossipMessage {
+    /// Block [`Header`].
+    header: Header,
+
+    /// Sequence number.
+    seq_no: u64,
```
bewakes, line 30: Is this `seq_no` used to prevent receiving of duplicate message from a peer?

I don't see it being used though(except for encoding, decoding).

crates/reth/node/src/gossip/package.rs
```diff
@@ -0,0 +1,359 @@
...
+        Ok(Self {
+            message,
+            public_key,
+            signature,
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
bewakes, line 185: Does it make sense to use `proptest`s?

bin/alpen-client/src/gossip.rs
```diff
@@ -0,0 +1,218 @@
...
+}
+
+/// Creates the gossip event handling task.
+///
+/// This task manages:
+///
+/// - Connection tracking (establish/close)
+/// - Receiving gossip messages and forwarding block hashes to engine control
+/// - Broadcasting new canonical blocks to connected peers
+pub(crate) async fn create_gossip_task(
```
bewakes, line 255: nit: This is a very big function. It might be worthwhile to separate out the `select!` handlers.

bin/alpen-client/src/gossip.rs
```diff
@@ -0,0 +1,218 @@
...
+                        }
+                    }
+                }
+            },
+            res = state_events.recv() => {
+                match res {
+                    Ok(event) => {
+                        if let CanonStateNotification::Commit { new } = event {
+                            // Extract headers from the new chain segment
+                            let headers: Vec<Header> = new
```
bewakes: Since `headers` is only used to get the last element, its better to just do `new.headers.last().map(...)` to avoid unnecessary vector allocation.

bin/alpen-client/src/gossip.rs
delbonis: Yeah there's no reason to compile it out.

bin/alpen-client/src/gossip.rs
delbonis: Can we submit it to whatever peer that we're syncing from?  That's what we should be doing.

bin/alpen-client/src/gossip.rs
```diff
@@ -0,0 +1,218 @@
...
+                            lagged = n,
+                            "Canonical state subscription lagged"
+                        );
+                        continue;
+                    }
+                    Err(broadcast::error::RecvError::Closed) => {
+                        warn!(
+                            target: "alpen-gossip",
+                            "Canonical state subscription closed"
+                        );
```
delbonis: I don't think this needs a warning, we can just gracefully exit.

delbonis (approved):
generally agree with other comments but that's mostly nits at this point

bin/alpen-client/src/gossip.rs
storopoli: I prefer to keep feature-gated.
It is cleaner code to maintain and also avoids compiling sequencer code in non-sequencer binaries. Reduces the attack surface and bugs for non-sequencer nodes also making the binary smaller and leaner in resources.

storopoli:
> Can we submit it to whatever peer that we're syncing from? That's what we should be doing.

I don't get it. What do you mean?
Code has changed, Sapin asked to remove the `warn!`. You mean that non-sequencer nodes regossip the block header that they've seen? Isn't this what the code is doing already?

crates/reth/node/src/gossip/package.rs
storopoli: Ok I've replaced the LRU for a simple `seq_no` dedup in df983208033b331a42e6170a5949c4278bbbea0f

crates/reth/node/src/gossip/package.rs
storopoli: Yes, added `proptest!`s in 30303d52086ece5965cadc38fec7cb7ea8a839fa replacing the static ones.

bin/alpen-client/src/gossip.rs
storopoli: Yes see bc639675794ece87af9b1ca2aad7f5e81861f8da

bin/alpen-client/src/gossip.rs
storopoli: Great suggestion. Changed to:
```rs
        Ok(event) => {
            if let CanonStateNotification::Commit { new } = event {
                // Extract the last header from the new chain segment
                if let Some(tip) = new.headers().last().map(|h| h.header().clone()) {
                    broadcast_new_block(&tip, connections, config);
                }
            }
            true // Continue loop
```
see bc639675794ece87af9b1ca2aad7f5e81861f8da

bin/alpen-client/src/gossip.rs
storopoli: I first went with a LRU cache in bc639675794ece87af9b1ca2aad7f5e81861f8da then I just decided to use `seq_no` dedup in df983208033b331a42e6170a5949c4278bbbea0f

bewakes (approved):

PR #1171, opened 2025-11-29
Str 1919 strata new binary

bewakes:
## Description

A new strata binary. Cleans up main reusing the older config parsing, args and params.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
Used AI for mundane tasks, some docstrings/cleanups and feedback on the structures.

## Checklist

### DIFF
--- a/bin/strata/src/args.rs
+++ b/bin/strata/src/args.rs
@@ -37,7 +37,7 @@ pub(crate) struct Args {
     #[argh(
         option,
         short = 'd',
-        description = "datadir path that will contain databases"
+        description = "datadir path used mainly for databases"
     )]
     pub datadir: Option<PathBuf>,
 
@@ -64,24 +64,23 @@ pub(crate) struct Args {
 }
 
 impl Args {
-    /// Get strings of overrides gathered from args.
-    pub(crate) fn get_overrides(&self) -> Result<Vec<String>, InitError> {
+    /// Get strings of overrides gathered from user and internal attributes.
+    pub(crate) fn get_all_overrides(&self) -> Result<Vec<String>, InitError> {
         let mut overrides = self.overrides.clone();
-        overrides.extend_from_slice(&self.get_direct_overrides()?);
+        overrides.extend_from_slice(&self.get_internal_overrides()?);
         Ok(overrides)
     }
 
-    /// Overrides passed directly as args and not as overrides.
-    fn get_direct_overrides(&self) -> Result<Vec<String>, InitError> {
+    /// Overrides passed directly as args attributes.
+    fn get_internal_overrides(&self) -> Result<Vec<String>, InitError> {
         let mut overrides = Vec::new();
         if self.sequencer {
             overrides.push("client.is_sequencer=true".to_string());
         }
         if let Some(datadir) = &self.datadir {
-            let dd = datadir.to_str().ok_or(anyhow::anyhow!(
-                "Invalid datadir override path {:?}",
-                datadir
-            ))?;
+            let dd = datadir
+                .to_str()
+                .ok_or_else(|| InitError::InvalidDatadirPath(datadir.clone()))?;
             overrides.push(format!("client.datadir={dd}"));
         }
         if let Some(rpc_host) = &self.rpc_host {
--- a/bin/strata/src/config.rs
+++ b/bin/strata/src/config.rs
@@ -6,12 +6,14 @@ use crate::errors::ConfigError;
 
 type Override = (String, toml::Value);
 
-/// Parses an override string. Splits by '=' to get key and value, then splits
-/// the key by '.' which is the update path.
+/// Parses an override string. Splits by '=' to get key and raw str value, then parses the str
+/// value.
 pub(crate) fn parse_override(override_str: &str) -> Result<Override, ConfigError> {
     let (key, value_str) = override_str
         .split_once("=")
-        .ok_or(ConfigError::InvalidOverride(override_str.to_string()))?;
+        .ok_or(ConfigError::InvalidOverride {
+            override_str: override_str.to_string(),
+        })?;
     Ok((key.to_string(), parse_value(value_str)))
 }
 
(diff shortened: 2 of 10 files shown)


bin/strata/src/args.rs
```diff
@@ -0,0 +1,96 @@
...
+impl Args {
+    /// Get strings of overrides gathered from args.
+    pub(crate) fn get_overrides(&self) -> Result<Vec<String>, InitError> {
+        let mut overrides = self.overrides.clone();
+        overrides.extend_from_slice(&self.get_direct_overrides()?);
+        Ok(overrides)
+    }
+
+    /// Overrides passed directly as args and not as overrides.
+    fn get_direct_overrides(&self) -> Result<Vec<String>, InitError> {
```
delbonis: The way these two functions are related is ambiguous.  If anything, I would expect `get_direct_overrides` to do what `get_overrides` does.

Better names would be something like

* `get_user_overrides`
* `get_internal_overrides`

bin/strata/src/config.rs
```diff
@@ -0,0 +1,49 @@
...
+) -> Result<(), ConfigError> {
+    match path.split_once(".") {
+        None => {
+            table.insert(path.to_string(), value);
+            Ok(())
+        }
+        Some((key, rest)) => {
+            if let Some(t) = table.get_mut(key).and_then(|v| v.as_table_mut()) {
+                apply_override(rest, value, t)
+            } else if table.contains_key(key) {
```
delbonis: This handling is kinda weird, we're doing the table lookup twice.  I would expect a structure more like

```rs
match table.get_mut(key) {
  Some(toml::Value::Table(tbl)) => ...,
  Some(_) => ...,
  None => ...
}
```

Really, I would expect this `apply_overrides` function to accept a `target: &mut toml::Value` and, if the path is exhausted, we just `*target = value;`.

bin/strata/src/config.rs
```diff
@@ -0,0 +1,49 @@
+//! Configuration override parsing and application logic.
+
+use toml::value::Table;
+
+use crate::errors::ConfigError;
+
+type Override = (String, toml::Value);
+
+/// Parses an override string. Splits by '=' to get key and value, then splits
+/// the key by '.' which is the update path.
```
delbonis: This doc comment isn't accurate wrt the path.

bin/strata/src/errors.rs
```diff
@@ -0,0 +1,40 @@
...
+
+#[derive(Debug, Error)]
+pub(crate) enum InitError {
+    #[error("io: {0}")]
+    Io(#[from] io::Error),
+
+    #[error("unparsable params file: {0}")]
+    UnparsableParamsFile(#[from] SerdeError),
+
+    #[error("config: {0:?}")]
```
delbonis: `{0}` instead of `{0:?}`

bin/strata/src/errors.rs
```diff
@@ -0,0 +1,40 @@
...
+    MalformedConfig(#[from] ConfigError),
+
+    #[error("params: {0}")]
+    MalformedParams(#[from] ParamsError),
+
+    #[error("{0}")]
+    Anyhow(#[from] anyhow::Error),
```
delbonis: Can we structure the error handling better so we don't get this weird nesting?  thiserror errors should typically be leafs.

bin/strata/src/errors.rs
```diff
@@ -0,0 +1,40 @@
...
+    #[error("missing key: {0}")]
+    MissingKey(String),
+
+    /// Tried to traverse into a primitive.
+    #[error("can't traverse into non-table key: {0}")]
+    TraverseNonTableAt(String),
+
+    /// Invalid override string.
+    #[error("Invalid override: '{0}'")]
+    InvalidOverride(String),
```
delbonis: Inconsistent formatting, should be:

* `missing key '{0}'`
* `attempt to traverse into non-table key '{0}'`
* `invalid override string '{0}'`

bin/strata-new/src/args.rs
```diff
@@ -0,0 +1,139 @@
...
+    // Config non-overriding args
+    #[argh(option, short = 'c', description = "path to configuration")]
+    pub config: PathBuf,
+
+    // Config overriding args
+    /// Data directory path that will override the path in the config toml.
+    #[argh(
+        option,
+        short = 'd',
+        description = "datadir path that will contain databases"
```
storopoli:
```suggestion
        description = "datadir path used mainly for databases"
```

bin/strata-new/src/init_db.rs
```diff
@@ -0,0 +1,13 @@
+use std::{path::Path, sync::Arc};
+
+use strata_db_store_sled::{
+    SLED_NAME, SledBackend, SledDbConfig, init_core_dbs, open_sled_database,
+};
+
+/// Initialize database backend based on configured features
+pub(crate) fn init_database(datadir: &Path, retry_count: u16) -> anyhow::Result<Arc<SledBackend>> {
+    let sled_db = open_sled_database(datadir, SLED_NAME)?;
+    let retry_delay_ms = 200u64;
```
storopoli: This should be configurable with a sane fallback default `const`.

bin/strata-new/src/main.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    // wait for rpc to stop
+    rpc_handle.stopped().await;
+
+    Ok(())
+}
+
+fn do_startup_checks(_ctx: &NodeContext) {
+    // TODO: things like if bitcoin client is running or not, db consistency checks and any other
+    // checks prior to starting services, genesis checks etc.
```
storopoli: Can be moved to a ticket and I can take that since it will be very similar to [STR-1783](https://alpenlabs.atlassian.net/browse/STR-1783) that I'm already assigned.

bin/strata/src/args.rs
bewakes: The `get_overrides` here is collecting all the overrides: the internal(from the args attributes) and user provided overrides(via `args.overrides`. So, these would be better named as `get_all_overrides` and `get_internal_overrides`.

bin/strata/src/config.rs
bewakes: Yes, agreed. This should have a better pattern match which allows a single lookup.
Accepting a `target: &mut Value` would also work, I tried and it made the function body a bit bulkier. So leaving it as is for now.

bin/strata/src/errors.rs
bewakes: yes, done.

bin/strata-new/src/init_db.rs
bewakes: Done.

bin/strata-new/src/main.rs
bewakes: Created a ticket: https://alpenlabs.atlassian.net/browse/STR-1935

PR #1177, opened 2025-12-01
Str 1822 state accessor

bewakes:
## Description
This PR basically adds machinery to accumulate write batch and other auxiliary data like mmr entries during block execution.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/ol/state-types/src/writebatch.rs
+++ b/crates/ol/state-types/src/writebatch.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 
 use strata_acct_types::AccountId;
-use strata_ledger_types::AsmManifest;
+use strata_ledger_types::{AsmManifest, StateAccessor};
 use strata_snark_acct_types::MessageEntry;
 
 use crate::{AccountState, EpochalState, GlobalState};
@@ -11,23 +11,27 @@ use crate::{AccountState, EpochalState, GlobalState};
 /// This structure acts as a sparse, in-memory layer on top of a base state
 /// (which could be DB-backed). Reads check the overlay first, then fall through
 /// to the base. Writes populate the overlay.
-#[derive(Clone, Debug, Default)]
-pub struct WriteBatch {
+#[derive(Clone, Debug)]
+pub struct WriteBatch<S: StateAccessor> {
     /// Accounts modified during execution (CoW overlay).
     /// Presence in this map indicates the account was accessed mutably.
-    pub(crate) modified_accounts: HashMap<AccountId, AccountState>,
+    pub(crate) modified_accounts: HashMap<AccountId, S::AccountState>,
 
     /// Global state override. None means use base state.
-    pub(crate) global_state: Option<GlobalState>,
+    pub(crate) global_state: Option<S::GlobalState>,
 
     /// Epochal state override. None means use base state.
-    pub(crate) epochal_state: Option<EpochalState>,
+    pub(crate) epochal_state: Option<S::L1ViewState>,
 }
 
-impl WriteBatch {
+impl<S: StateAccessor> WriteBatch<S> {
     /// Create a new empty WriteBatch
     pub fn new() -> Self {
-        Self::default()
+        Self {
+            modified_accounts: Default::default(),
+            global_state: None,
+            epochal_state: None,
+        }
     }
 
     /// Check if an account exists in the overlay
@@ -36,37 +40,37 @@ impl WriteBatch {
     }
 
     /// Get a reference to a modified account, if it exists in the overlay
-    pub fn get_account(&self, id: &AccountId) -> Option<&AccountState> {
+    pub fn get_account(&self, id: &AccountId) -> Option<&S::AccountState> {
         self.modified_accounts.get(id)
     }
 
     /// Get a mutable reference to a modified account, if it exists in the overlay
-    pub fn get_account_mut(&mut self, id: &AccountId) -> Option<&mut AccountState> {
+    pub fn get_account_mut(&mut self, id: &AccountId) -> Option<&mut S::AccountState> {
         self.modified_accounts.get_mut(id)
     }
 
     /// Insert or update an account in the overlay
-    pub fn insert_account(&mut self, id: AccountId, state: AccountState) {
+    pub fn insert_account(&mut self, id: AccountId, state: S::AccountState) {
         self.modified_accounts.insert(id, state);
     }
 
     /// Get global state if overridden
-    pub fn global_state(&self) -> Option<&GlobalState> {
+    pub fn global_state(&self) -> Option<&S::GlobalState> {
         self.global_state.as_ref()
     }
 
     /// Get mutable global state, creating if needed
-    pub fn global_state_mut_or_insert(&mut self, base: &GlobalState) -> &mut GlobalState {
+    pub fn global_state_mut_or_insert(&mut self, base: &S::GlobalState) -> &mut S::GlobalState {
         self.global_state.get_or_insert_with(|| base.clone())
     }
 
     /// Get epochal state if overridden
-    pub fn epochal_state(&self) -> Option<&EpochalState> {
+    pub fn epochal_state(&self) -> Option<&S::L1ViewState> {
         self.epochal_state.as_ref()
     }
 
     /// Get mutable epochal state, creating if needed
-    pub fn epochal_state_mut_or_insert(&mut self, base: &EpochalState) -> &mut EpochalState {
+    pub fn epochal_state_mut_or_insert(&mut self, base: &S::L1ViewState) -> &mut S::L1ViewState {
         self.epochal_state.get_or_insert_with(|| base.clone())
     }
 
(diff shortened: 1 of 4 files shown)


crates/ol/state-types/src/tracking_state_accessor.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    /// Track which accounts were modified (excluding newly created)
+    modified_accounts: HashSet<AccountId>,
+
+    /// Track which accounts were newly created
+    created_accounts: HashSet<AccountId>,
+
+    /// Accumulate auxiliary data for database persistence
+    aux: ExecutionAuxiliaryData,
+}
```
delbonis: This doesn't make any sense, we can and should be writing directly into a write batch.  See discussion about write batch stacks in DMs.

crates/ol/state-types/src/writebatch.rs
```diff
@@ -0,0 +1,50 @@
...
+pub struct WriteBatch {
+    // Global state changes
+    pub new_slot: u64,
+
+    // L1 View / Epochal state changes
+    pub l1_view_writes: L1ViewWrites,
+
+    // Account changes - store full AccountState for now
+    // TODO: Consider storing diffs instead of full state for efficiency
+    pub new_accounts: Vec<(AccountId, AccountSerial, AccountState)>,
```
delbonis: * This should be structured better.
* Diffs are not relevant or needed here.
* Serials are redundant.

crates/ol/state-types/src/writebatch.rs
```diff
@@ -0,0 +1,50 @@
...
+use strata_ledger_types::AsmManifest;
+use strata_snark_acct_types::MessageEntry;
+
+use crate::AccountState;
+
+/// Batch of state changes from executing a block
+#[derive(Clone, Debug)]
+pub struct WriteBatch {
+    // Global state changes
+    pub new_slot: u64,
```
delbonis: Store the new global state so that we don't have to manually keep this in sync.

crates/ol/state-types/src/writebatch.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub struct L1ViewWrites {
+    pub cur_epoch: u32,
+    pub added_manifests: Vec<AsmManifest>,
+    pub asm_manifest_mmr: Mmr64,
+    pub asm_recorded_epoch: EpochCommitment,
+    pub total_ledger_balance: BitcoinAmount,
+}
```
delbonis: We can separately track the MMR appends so that we can just use the full new EpochalState instead of this form.

crates/ol/state-types/src/writebatch.rs
```diff
@@ -0,0 +1,50 @@
...
+pub struct ExecutionAuxiliaryData {
+    /// Messages added to each account's inbox during this block
+    /// Stored separately for DB indexing and queries
+    pub account_message_additions: HashMap<AccountId, Vec<MessageEntry>>,
+
+    /// ASM manifests processed during this block
+    /// Stored separately for DB indexing
+    // TODO: this might be redundant as asm might be saving this as well.
+    pub asm_manifests: Vec<AsmManifest>,
+}
```
delbonis: Track this separately.  It's not "for database persistence", it's for accumulator indexing.

crates/ol/state-types/src/toplevel.rs
```diff
@@ -85,11 +138,33 @@ impl StateAccessor for OLState {
...
-        Ok(self.ledger.get_account_state_mut(&id))
+        acct_id: AccountId,
+        state: Hash,
+        next_read_idx: u64,
+        seqno: Seqno,
+    ) -> AcctResult<()> {
+        let mut snark_state = self.get_snark_acct(&acct_id)?;
+        snark_state.set_proof_state_directly(state, next_read_idx, seqno);
+        self.set_snark_acct(&acct_id, snark_state)
     }
```
delbonis: Remove these.

crates/ol/state-types/src/toplevel.rs
```diff
@@ -125,3 +200,38 @@ impl StateAccessor for OLState {
         Ok(hash)
     }
 }
+
+impl OLState {
+    /// Apply a WriteBatch to this state
+    ///
+    /// This takes the changes accumulated during block execution and applies them
+    /// to the current state. Used when accepting blocks into the canonical chain.
+    pub fn apply_write_batch(&mut self, batch: crate::WriteBatch) -> AcctResult<()> {
```
delbonis: Import at top of file.

crates/ol/state-types/src/toplevel.rs
```diff
@@ -125,3 +200,38 @@ impl StateAccessor for OLState {
...
+        // Apply global state changes
+        self.global.set_cur_slot(batch.new_slot);
+
+        // Apply L1 view changes
+        self.epoch.set_cur_epoch(batch.l1_view_writes.cur_epoch);
+        self.epoch
+            .set_asm_recorded_epoch(batch.l1_view_writes.asm_recorded_epoch);
+        self.epoch
+            .set_total_ledger_balance(batch.l1_view_writes.total_ledger_balance);
```
delbonis: These should be updating the fields directly instead of going through the accessors.  We specifically want to bypass the trait interface in this context.

crates/ol/state-types/src/toplevel.rs
```diff
@@ -125,3 +200,38 @@ impl StateAccessor for OLState {
...
+
+        // Note: manifests were already applied during execution via append_manifest()
+        // They're included in WriteBatch for DB persistence, not for state application
+
+        // Apply new accounts
+        for (id, _serial, acct_state) in batch.new_accounts {
+            // Accounts should have been created during execution
+            // This just ensures they're in the final state
+            self.update_account_state(id, acct_state)?;
```
delbonis: `update_account_state` *was* defined to error if trying to create new accounts, since creating new accounts has consequences for the serial accumulator.

crates/ledger-types/src/state_accessor.rs
```diff
@@ -25,31 +28,68 @@ pub trait StateAccessor {
...
+    ///
+    /// # Note
+    /// Returns error for non-snark account
+    fn insert_inbox_message(&mut self, acct_id: AccountId, entry: MessageEntry) -> AcctResult<()>;
 
     /// Overwrites an existing account entry's state, if it exists.
     ///
     /// This refuses to create new accounts in order to avoid accidents like
     /// screwing up serials.
+    // TODO: might not be necessary with the above modifiers.
```
delbonis: Why would it not be necessary?

crates/ol/state-types/src/ledger.rs
```diff
@@ -84,6 +87,7 @@ impl TsnlLedgerAccountsTable {
         self.serials.push(id);
 
         // Sanity check.
+        // FIXME: remove runtime panics while it can be avoided with proper types(SortedVec)
         assert!(
             self.accounts.is_sorted_by_key(|e| e.id),
             "ol/state: accounts table not sorted by ID"
```
delbonis: We can't use SortedVec since it's not SSZ-able.

crates/ol/state-types/src/toplevel.rs
bewakes: Yeah, that makes sense. Need to use another method for creating account.

crates/ol/state-types/src/toplevel.rs
bewakes: Makes sense.

crates/ol/state-types/src/toplevel.rs
bewakes: These are a part of the flattened state accessor modifiers if you are indicating to the methods.

crates/ol/state-types/src/ledger.rs
bewakes: Can we make it SSZ-able?

crates/ledger-types/src/state_accessor.rs
bewakes: This would probably not be necessary because there are methods to update each attributes, unless we want to completely replace the account state.

crates/ol/state-types/src/tracking_state_accessor.rs
bewakes: Sure, writing directly into a write batch is also something that can be done. However, I do not see why this does not make sense.

crates/ol/state-types/src/writebatch.rs
bewakes: Oh yes. I overdid the writes structure. This structure actually is redundant.

crates/ol/state-types/src/writebatch.rs
bewakes: It _is_ also for db persistence. These need to be used later while updating accumulator proofs in block assembly.

> Track this separately

Do you mean separately from the writebatch or tracker?

PR #1180, opened 2025-12-01
Str 1918 guest code for eth ee

purusang:
## Description
Since there are some intersecting work between [EVM EE Guest code](https://alpenlabs.atlassian.net/browse/STR-1918) and [EE Prover](https://alpenlabs.atlassian.net/browse/STR-1782) I will be working on both in this pr. The replacement of guest code will force CL_STF to consume new types of outputs hence we cannot get passed all functional tests unless we keep the old guest code while new one is merged without integrating with prover.

The proof cycles for the added guest code is being appeared as `EVM EE STF (new) ` for the time being and should replace `EVM EE STF` before this pr gets merged. It is in early state so looking forward for suggestions on improving this work.
Till now new guest code is created that works with trait-based Eth block processing.

TODO: 
- guest code (STR-1918): handle `BlockOutputs` instead of `ExecSegment`. **I am looking for suggestion on this as [commented](https://github.com/alpenlabs/alpen/pull/1180#discussion_r2585101256) in the pr.** 
- EE Prover: (STR-1782): integrate new guest code with EE prover.

NOTE: I have used AI for this work.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues
STR-1918, STR-1782


crates/proof-impl/evm-ee-stf/src/lib.rs
```diff
@@ -1,80 +1,114 @@
...
+        // Need to decide approach:
+        // 1. Add transformation layer here ie. BlockOutputs to ExecSegment (backward compatible)
+        // 2. Update CL_STF to consume BlockOutputs directly (cleaner long-term)
+        // Decision to be made in PR review.
+        let _block_outputs = output.outputs();
     }
+
+    // TODO: Commit properly transformed outputs once decision is made
+    let empty_outputs: Vec<u8> = vec![];
+    zkvm.commit_serde(&empty_outputs);
```
purusang, line 113: @prajwolrg  @MdTeach  @delbonis   I need suggestion on this. Unless we handle block outputs, the old evm-ee-stf code cannot be replaced as there are multiple dependent section that will fail (prover-perf, prover-client & test-utils).

PR #1184, opened 2025-12-03
feat: handle slash tx

prajwolrg:
## Description

This PR implements the core logic for handling slash transactions in the Bridge V1 subprotocol. It adds validation to ensure that slash transactions correctly reference valid stake outputs locked to historical N/N multisig scripts and updates the state by removing the slashed operator.

This PR also makes refactors to the existing test structure by creating a separate `test_utils.rs` file.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
+++ b/crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
@@ -387,12 +387,9 @@ mod tests {
     use strata_test_utils::ArbitraryGenerator;
 
     use super::*;
-    use crate::{
-        state::withdrawal::WithdrawOutput,
-        test_utils::{
-            add_deposits, add_deposits_and_assignments, create_test_state,
-            create_withdrawal_info_from_assignment,
-        },
+    use crate::test_utils::{
+        add_deposits, add_deposits_and_assignments, create_test_state,
+        create_withdrawal_info_from_assignment,
     };
 
     /// Test successful deposit transaction processing.
--- a/crates/asm/subprotocols/bridge-v1/src/test_utils.rs
+++ b/crates/asm/subprotocols/bridge-v1/src/test_utils.rs
@@ -13,9 +13,7 @@ use strata_primitives::l1::{BitcoinAmount, L1BlockCommitment};
 use strata_test_utils::ArbitraryGenerator;
 
 use super::*;
-use crate::state::{
-    assignment::AssignmentEntry, config::BridgeV1Config, withdrawal::WithdrawOutput,
-};
+use crate::state::{assignment::AssignmentEntry, config::BridgeV1Config};
 
 /// A Mock MsgRelayer that does nothing.
 ///


crates/asm/subprotocols/bridge-v1/src/handler.rs
```diff
@@ -13,10 +14,13 @@ use crate::{
 /// - **Deposit**: Processes the deposit transaction without emitting logs
 /// - **WithdrawalFulfillment**: Processes the withdrawal and emits a withdrawal processed log via
 ///   the relayer to notify other components of the processed withdrawal
+/// - **Slash**: Validates the stake connector is locked to any valid N/N multisig (current or
```
barakshani: why not just saying _process_ as per the txes above?

crates/asm/subprotocols/bridge-v1/src/handler.rs
```diff
@@ -42,5 +47,63 @@ pub(crate) fn handle_parsed_tx<'t>(
...
+            // Validate that the stake connector (second input) is locked to a valid N/N
+            // multisig script from any historical configuration.
+            //
+            // This ensures this is a valid slash transaction that was presigned by an N/N
+            // multisig set and not some arbitrary transaction posted by someone else.
+            //
+            // We check against all historical scripts (which includes the current one)
+            // because when multiple operators are slashed in sequence, the second slash
+            // will reference a stake connector from the operator set configuration before
+            // the first slash occurred.
```
barakshani: I don't follow the need to discuss different slash and any timeline between them

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -124,6 +126,27 @@ pub struct OperatorTable {
     /// The key is automatically computed when the operator table is created or
     /// updated, ensuring it always reflects the current active multisig participants.
     agg_key: BitcoinXOnlyPublicKey,
+
+    /// Historical N/N multisig scripts from previous operator set configurations.
+    ///
+    /// This vector tracks all previous P2TR scripts that were used when the operator set
+    /// changed (due to slashing or unstaking). These historical scripts are necessary to validate
```
barakshani: >(due to slashing or unstaking)

could be entry/exit as well

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -124,6 +126,27 @@ pub struct OperatorTable {
...
+    /// slash transactions that reference stake connectors from previous operator configurations.
+    ///
+    /// # Use Case
+    ///
+    /// When multiple operators are slashed in sequence:
+    /// 1. First slash: Validated against current N/N multisig script, operator removed, new script
+    ///    computed
+    /// 2. Second slash: Must be validated against the previous N/N multisig script (before first
+    ///    slash) because the stake connector was created when that operator set was active
```
barakshani: why do we care about it? we validate against _all_ multisigs anyway. I think you're saying that we should not validate against the very last multisig, but:
1. is the vector deduplicated? That is, what if the very last multisig matches an old one
2. if someone is able to send an (illegitimate) slash tx with the very last (=current) multisig then we have other severe issues -- as the bridge multisig address is compromised.

barakshani: reading this again, you seems to focus on justifying why we validate against _all_ multisigs. If this is the intension, it's not clear to me why talking about "multiple operators are slashed in sequence" (what does in sequence mean? I read at as one immediately after the other)? there are other cases when this is required.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -240,35 +288,67 @@ impl OperatorTable {
...
+    /// > who (claim to) have identical individual public keys properly, and applications are not
+    /// > required to check for duplicate individual public keys. In fact, applications are
+    /// > recommended to omit checks for duplicate individual public keys in order to simplify
+    /// > error handling.
+    ///
+    /// [BIP-327]: https://github.com/bitcoin/bips/blob/master/bip-0327.mediawiki#public-key-aggregation
+    ///
+    /// In this implementation, only the administration subprotocol can add new members.
+    /// If the admin subprotocol chooses to add duplicate operators, we assume they have
+    /// a valid reason (e.g., weighted voting or specific trust arrangements) and allow it.
```
barakshani: I don't like this. The examples also doesn't really make sense to me.

@john-light is this something we want to allow?

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -285,9 +365,11 @@ impl OperatorTable {
 
             self.next_idx += 1;
         }
+    }
 
-        // Handle updates by modifying the bitmap directly
-        for &(idx, is_active) in updates {
+    /// Deactivates existing operators by their indices.
+    fn remove_operators(&mut self, indices: &[OperatorIdx]) {
```
barakshani, line 346: if we have duplicate operators, how many times do we need to remove them to completely remove the pk?

crates/asm/common/src/aux/provider.rs
```diff
@@ -134,6 +134,27 @@ impl VerifiedAuxData {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns `AuxError::BitcoinTxNotFound` if the transaction is not found.
+    pub fn get_bitcoin_txout(&self, outpoint: &OutPoint) -> AuxResult<&TxOut> {
+        let txid = outpoint.txid;
+        let tx = self
+            .txs
+            .get(&outpoint.txid)
+            .ok_or(AuxError::BitcoinTxNotFound { txid })?;
```
delbonis: (very slightly sloppy)

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -298,32 +380,44 @@ impl OperatorTable {
...
-                .into();
-        }
+    /// Builds a P2TR script for the provided aggregated operator key.
+    fn build_nn_script(agg_key: &BitcoinXOnlyPublicKey) -> BitcoinScriptBuf {
+        BitcoinScriptBuf::from(ScriptBuf::new_p2tr(
+            SECP256K1,
+            agg_key.to_xonly_public_key(),
+            None,
+        ))
     }
```
delbonis: Should be a standalone fn.

crates/asm/subprotocols/bridge-v1/src/subprotocol.rs
```diff
@@ -42,6 +43,47 @@ impl Subprotocol for BridgeV1Subproto {
...
+            // Parse transaction to extract structured data, then handle the preprocess transaction
+            // to get the auxiliary requests
+            match parse_tx(tx) {
+                Ok(parsed_tx) => {
+                    preprocess_parsed_tx(parsed_tx, state, collector);
+                    info!(tx_id = %tx.tx().compute_txid(), "Successfully pre-processed tx");
+                }
+                Err(e) => {
+                    error!(tx_id = %tx.tx().compute_txid(), error = %e, "Failed to process tx")
+                }
```
delbonis: Log params should be `txid`, I don't think we use `tx_id` *anywhere*.  Also error case log says "process" instead of "pre-process".

We definitely should not be logging as `info` in the happy case.  I'm not even sure we should be directly emitting logs in ASM stuff at all tbh.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
delbonis: This is a good question.  What would the implications of duplicate operators be?

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -240,35 +288,67 @@ impl OperatorTable {
...
     ) {
-        // Handle inserts first
-        for musig2_pk in inserts {
+        self.add_operators(add_members);
+        self.remove_operators(remove_members);
+
+        if !remove_members.is_empty() || !add_members.is_empty() {
+            self.recalculate_aggregated_key();
+            self.historical_nn_scripts
+                .push(Self::build_nn_script(&self.agg_key));
```
delbonis: This is a smart way to approach it.  Should we be doing more explicit tracking for the n-of-n master pubkey "versions"?

delbonis (changes requested):
Agree with Barak's comments in general.

crates/asm/subprotocols/bridge-v1/src/subprotocol.rs
prajwolrg: The log inside of ASM are turned to noop inside ZkVm (so no proving overhead). The logs are kept with the purpose of making debugging easier. I don't have a very strong opinion on this.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
prajwolrg: I hadn't thought we needed to track all the N/N scripts in the first place. So I wanted to highlight on that. 

It's not required for the deposit transaction as for deposit we only check against the current multisig. 

As you suggested, it's better to write it in case of `entry/exit`, rather than explicitly mentioning slashing.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
prajwolrg: > if we have duplicate operators, how many times do we need to remove them to completely remove the pk?

The operator entry are identified by `OperatorIdx`. If multiple `OperatorIdx` have the same pubkey, we need to remove all the OperatorIdx associated with that pubkey to completely remove that pk. Note that it's still not removed, we just mark it as inactive.

> What would the implications of duplicate operators be?

If there are two operator entries with the same public key, and both are active, it means the aggregation will take both of that. The new aggregated key with de-duplicated public keys are different than the original one where we had duplicated keys. So during the `MuSig2` signing, we need two separate signatures from the same signer, with the same message. The signer would also need to contribute randomness twice on the first round of `MuSig2`.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
prajwolrg: I'm not so sure. Since we are storing it in a `Vector`, in a sense, it already does it. We can be more explicit about this, but I don't know of a good reason.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
prajwolrg: I'd defer to @john-light on this as well. `MuSig2` allows for the same individual public key is allowed to occur more than once and as mentioned in the BIP, it is by design. 

We can reject the update in case there are duplicated public keys, or we can allow it and it's the responsibility of the admin to make sure there are no duplicates.

In practice, it might be an issue to the off-chain signature collection scheme of Musig2 as briefly touched upon [here](https://github.com/alpenlabs/alpen/pull/1184#discussion_r2588755762). CC: @Rajil1213

crates/asm/subprotocols/bridge-v1/src/handler.rs
```diff
@@ -48,3 +48,32 @@ pub(crate) fn handle_parsed_tx<'t>(
...
+
+/// Pre-processes a parsed transaction to collect auxiliary data requests.
+///
+/// This function inspects the transaction type and requests any additional data needed
+/// for full verification during the main processing phase. Currently handles:
+///
+/// - **Deposit transactions**: No auxiliary data required
+/// - **Withdrawal fulfillment transactions**: No auxiliary data required
+/// - **Slash transactions**: Requests the conflicting Bitcoin transaction referenced in
+///   the slash proof to enable verification of operator double-signing
```
Rajil1213: I think this needs rephrasing. What is the "slash proof"? What is potentially being double-spent? What is the "conflicting Bitcoin transaction"?

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -298,32 +326,35 @@ impl OperatorTable {
...
-
-            self.agg_key = aggregate_schnorr_keys(active_keys.iter())
-                .expect("Failed to generate aggregated key")
-                .into();
+    /// Recalculates the aggregated key based on currently active operators.
+    ///
+    /// # Panics
+    ///
+    /// Panics if there are no active operators.
+    fn recalculate_aggregated_key(&mut self) {
```
Rajil1213: Nit: Afaict, this API cannot guarantee that it is a *re*calculation for a given operator set. This could be called the first time a new operator set is created. So, might as well name this `calculate_aggregated_key`.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -240,35 +240,61 @@ impl OperatorTable {
...
-        inserts: &[EvenPublicKey],
+        add_members: &[EvenPublicKey],
+        remove_members: &[OperatorIdx],
     ) {
-        // Handle inserts first
-        for musig2_pk in inserts {
+        self.add_operators(add_members);
+        self.remove_operators(remove_members);
+
+        if !remove_members.is_empty() || !add_members.is_empty() {
```
Rajil1213: Nit:

```suggestion
        let did_change = !remove_members.is_empty() || !add_members.is_empty();
        if did_change {
```

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -406,7 +406,7 @@ mod tests {
     /// - `Vec<EvenPublicKey>` - MuSig2 public keys for bridge configuration
     fn create_test_operators() -> (Vec<EvenSecretKey>, Vec<EvenPublicKey>) {
         let secp = Secp256k1::new();
-        let mut rng = secp256k1::rand::thread_rng();
+        let mut rng = bitcoin::secp256k1::rand::thread_rng();
```
Rajil1213: Why the diff?

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
Rajil1213: Agree with Barak's comments but would like to focus on the fact that "in sequence" does not even matter. For all we know, two different operators might have been slashed and there is no way to know beforehand the multisig that those slash transactions belong to without iterating over all historical aggregated pubkeys.

crates/asm/common/src/aux/errors.rs
```diff
@@ -36,13 +36,23 @@ pub enum AuxError {
...
+    /// Bitcoin [`TxOut`](bitcoin::TxOut) not found.
+    #[error("Bitcoin transaction out not found for {txid:?}: {vout:?}")]
+    BitcoinTxOutNotFound {
+        /// The requested txid
+        txid: Txid,
+        /// The requested vout
+        vout: u32,
+    },
+
+    /// Bitcoin transaction not found by txid.
```
Rajil1213: What is this doc comment for?

crates/asm/subprotocols/bridge-v1/src/subprotocol.rs
delbonis: Are we sure the macros expand to `{}` in the zkvm builds?  If so then it's fine, just want to keep it relatively minimal so that if we re-exec the asm it doesn't spam the logs.

prajwolrg: Yes, that's the plan. See [here](https://github.com/alpenlabs/alpen/blob/main/crates/asm/common/src/logging.rs).

prajwolrg: Well it wasn't fully `{}`. Fixed it in [205776c](https://github.com/alpenlabs/alpen/pull/1184/commits/205776c736aa23f8ee0d2f768c53ee57847231cb).

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
prajwolrg: I removed `secp256k1` as dependency.

crates/asm/common/src/aux/errors.rs
prajwolrg: It seems to be leftover from older copies. Thanks for catching this.

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
barakshani: @john-light confirmed (private message) that duplicate operators isn't a product requirement. I feel that supporting this can only add complexity, so I advise to remove it.

crates/asm/subprotocols/bridge-v1/src/handler.rs
```diff
@@ -42,5 +47,113 @@ pub(crate) fn handle_parsed_tx<'t>(
...
+
+/// Pre-processes a parsed transaction to collect auxiliary data requests.
+///
+/// This function inspects the transaction type and requests any additional data needed
+/// for full verification during the main processing phase. Currently handles:
+///
+/// - **Deposit transactions**: No auxiliary data required
+/// - **Withdrawal fulfillment transactions**: No auxiliary data required
+/// - **Slash transactions**: Requests the conflicting Bitcoin transaction referenced in the slash
+///   proof to enable verification of operator double-signing
```
barakshani: I don't understand any of these

barakshani: we are not slashing because of double-signing: it looks like AI lost touch

crates/asm/subprotocols/bridge-v1/src/subprotocol.rs
```diff
@@ -42,6 +43,50 @@ impl Subprotocol for BridgeV1Subproto {
         Ok(BridgeV1State::new(params))
     }
 
+    /// Pre-processes transactions to collect auxiliary data requests.
+    ///
+    /// This function runs before the main transaction processing to identify and request
+    /// any auxiliary data needed for verification. For Bridge V1, this primarily handles:
+    ///
+    /// - **Slash transactions**: Requests the conflicting Bitcoin transaction referenced in the
+    ///   slash proof to enable verification of operator misbehavior
```
barakshani: fix

crates/asm/subprotocols/bridge-v1/src/handler.rs
```diff
@@ -42,5 +47,113 @@ pub(crate) fn handle_parsed_tx<'t>(
...
             Ok(())
         }
+        ParsedTx::Slash(info) => {
+            let stake_connector_script = &verified_aux_data
+                .get_bitcoin_txout(info.second_inpoint().outpoint())?
+                .script_pubkey;
+
+            // Validate that the stake connector (second input) is locked to a known N/N multisig
+            // script from any recorded configuration.
+            if !state
```
barakshani, line 57: it looks that at no place we verify that stake amount is correct. It makes me a bit uncomfortable, though I can't think of how this could be exploited (if we are here, the tx has a sps-50 tag with slash tx type, signed by N/N, and not previously spent). Still, if we are in the conclusion that it's not needed, I'd expect a comment somewhere explaining why we don't check for the stake amount.
I'm not sure if this is the right place, or perhaps here: https://github.com/alpenlabs/alpen/blob/main/crates/asm/txs/bridge-v1/src/slash/parse.rs

barakshani (approved):

crates/asm/subprotocols/bridge-v1/src/handler.rs
prajwolrg: This looks bad. I missed this. Thanks for catching.

crates/asm/subprotocols/bridge-v1/src/handler.rs
prajwolrg: This has been touched upon [here](https://github.com/alpenlabs/alpen/blob/main/crates/asm/txs/bridge-v1/src/slash/mod.rs#L31-L33).

barakshani: This is output of slash, I meant output of stake -- but I agree the rational here should be the same

crates/asm/subprotocols/bridge-v1/src/subprotocol.rs
barakshani: this isn't fixed yet

barakshani (approved):

PR #1188, opened 2025-12-04
ee block assembly pt 2

sapinb:
## Description
followup to #1169 

* [x] Adds chain tracker for exec chain (requesting a better name) and related persistence
* [x] Adds Payload builder implementation
* [x] Refactors and cleanup

AI use: for documentation and testing boilerplate

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

I'm going with full chain tracking (including unfinalized forks and orphans) even though it seems overkill in the current single sequencer scenario, as I have previously had to implement most "wont be needed for now" features during deployment anyway.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1180,7 +1180,6 @@ dependencies = [
  "strata-ee-acct-types",
  "strata-ee-chain-types",
  "strata-identifiers",
- "strata-snark-acct-types",
  "thiserror 2.0.17",
  "tokio",
  "tracing",
--- a/crates/alpen-ee/exec-chain/Cargo.toml
+++ b/crates/alpen-ee/exec-chain/Cargo.toml
@@ -9,9 +9,6 @@ workspace = true
 [dependencies]
 alpen-ee-common.workspace = true
 strata-acct-types.workspace = true
-strata-ee-acct-types.workspace = true
-strata-identifiers.workspace = true
-strata-snark-acct-types.workspace = true
 
 eyre.workspace = true
 thiserror.workspace = true
@@ -19,5 +16,7 @@ tokio.workspace = true
 tracing.workspace = true
 
 [dev-dependencies]
+strata-identifiers.workspace = true
+strata-ee-acct-types.workspace = true
 strata-ee-chain-types.workspace = true
 alpen-ee-common = { workspace = true, features = ["test-utils"] }


crates/alpen-ee/genesis/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+        Vec::new(),
+        Vec::new(),
+    )
+}
+
+pub fn build_genesis_exec_block_package(params: &AlpenEeParams) -> ExecBlockPackage {
+    // genesis_raw_block_encoded_hash: We dont really care about this for genesis block.
+    // Sufficient for it to be deterministic.
+    // Can be added to [`AlpenEeParams`] if correct value is required.
+    let genesis_raw_block_encoded_hash = [0; 32];
```
storopoli, line 26: This could be something interesting like how Satoshi encoded "Chancellor on Brink of Second Bailout for Banks" in the genesis block.

crates/alpen-ee/genesis/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+    let genesis_account_state = build_genesis_ee_account_state(params);
+    let genesis_ol_block =
+        OLBlockCommitment::new(params.genesis_ol_slot(), params.genesis_ol_blockid());
+    let block = ExecBlockRecord::new(
+        genesis_package,
+        genesis_account_state,
+        0,
+        genesis_ol_block,
+        0,
+        [0; 32],
```
storopoli: Also why are you repeating `[0; 32]` here?
Whatever we decide to be the genesis block hash, we should make it a `const` and pass it around instead. As this is right now can be error-prone.

crates/alpen-ee/exec-chain/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+#![expect(unused_crate_dependencies, reason = "wip")]
```
storopoli: Can we remove this before we merge? If no, please add a `// TODO(@sapinb): ...` next to it?

bin/alpen-client/src/payload_builder.rs
```diff
@@ -0,0 +1,105 @@
...
+#[derive(Debug)]
+pub(crate) struct AlpenRethPayloadEngine {
+    payload_builder_handle: PayloadBuilderHandle<AlpenEngineTypes>,
+    exec_engine: AlpenRethExecEngine,
+}
+
+impl AlpenRethPayloadEngine {
+    pub(crate) fn new(
+        payload_builder_handle: PayloadBuilderHandle<AlpenEngineTypes>,
+        beacon_engine_handle: ConsensusEngineHandle<AlpenEngineTypes>,
```
storopoli, line 26: Should we be using Ethereum consensus names if we are not a PoS chain?
I don't have a good opinion on this. Just raising awareness.

bin/alpen-client/src/payload_builder.rs
```diff
@@ -0,0 +1,105 @@
...
+            .context("failed build payload")?;
+
+        Ok(payload)
+    }
+}
+
+fn sats_to_gwei(sats: u64) -> Option<u64> {
+    // 1 BTC = 10^8 sats = 10^9 gwei
+    sats.checked_mul(10)
+}
```
storopoli: This probably should be inside `crates/` or maybe `strata-common`?

crates/alpen-ee/common/src/traits/storage/exec_block.rs
```diff
@@ -0,0 +1,888 @@
...
+    pub fn create_exec_block(
+        blocknum: u64,
+        parent_hash: Hash,
+        block_hash: Hash,
+        ol_slot: u64,
+    ) -> ExecBlockRecord {
+        let package = create_package(block_hash, block_hash);
+        let account_state = create_account_state(block_hash);
+        let ol_block = create_ol_block(ol_slot, blocknum as u8);
+        let timestamp_ms = 1000000 + blocknum * 1000;
```
storopoli: Sooooo many zeroooooooes

```suggestion
        let timestamp_ms = 1_000_000 + blocknum * 1_000;
```

crates/alpen-ee/database/src/sleddb/db.rs
```diff
@@ -160,21 +176,304 @@ impl EeNodeDb for EeNodeDBSled {
...
+
+                    if !block_exists {
+                        // save block data
+                        block_tree.insert(&hash, &db_block)?;
+                        // save payload
+                        payload_tree.insert(&hash, &payload)?;
+                    }
+
+                    // Always update blocks by height to handle the case where this hash
+                    // wasn't tracked at this height yet (though this should be rare)
```
storopoli: I don't understand the comment "Always update blocks by height..." since the code checks for `!block_exists`.

crates/alpen-ee/exec-chain/src/state.rs
```diff
@@ -0,0 +1,555 @@
...
+}
+
+/// Initializes chain state from storage using the last finalized block and all unfinalized blocks.
+pub async fn init_exec_chain_state_from_storage<TStorage: ExecBlockStorage>(
+    storage: Arc<TStorage>,
+) -> Result<ExecChainState, ExecChainStateError> {
+    let last_finalized_block = storage
+        .best_finalized_block()
+        .await?
+        .expect("cannot be empty");
```
storopoli: Should this be a recoverable error, or just panic as it is now?

crates/alpen-ee/genesis/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+        ExecBlockCommitment::new(
+            params.genesis_blockhash().into(),
+            genesis_raw_block_encoded_hash,
+        ),
+        BlockInputs::new_empty(),
+        BlockOutputs::new_empty(),
+    )
+}
+
+pub fn buid_genesis_exec_block(params: &AlpenEeParams) -> (ExecBlockRecord, Vec<u8>) {
```
storopoli: typo:

```suggestion
pub fn build_genesis_exec_block(params: &AlpenEeParams) -> (ExecBlockRecord, Vec<u8>) {
```

crates/alpen-ee/exec-chain/src/state.rs
sapinb: Not recoverable, but will switch to returning an Err (the caller will .expect()). Also added some comments to provide context.

Note: This was done to simplify the genesis block logic, which is moved to `alpen_ee_genesis::handle_finalized_exec_genesis` instead of creating the genesis block here.

crates/alpen-ee/database/src/sleddb/db.rs
sapinb: Yeah. This comment is quite obtuse. Updating it.

Its a technically possible, but unexpected inconsistent state of the db. (Block was absent in `exec_block_tree`, but its hash was tracked in `exec_blocks_by_height_tree`). Deduping makes sure this case isn't problematic.

bin/alpen-client/src/payload_builder.rs
sapinb: This is just what it is called inside reth, that we are passing here:
https://github.com/paradigmxyz/reth/blob/9c30bf7af5e0d45deaf5917375c9922c16654b28/crates/node/builder/src/rpc.rs#L331
I'm ok with using some other name else if it matters.

bin/alpen-client/src/payload_builder.rs
sapinb: Will place this in alpen-ee/common. This is only relevant inside the EE.

crates/alpen-ee/genesis/src/lib.rs
sapinb: genesis blockhash is part of the AlpenEeParams. The zeros are other deterministic things. I've added more context.

crates/alpen-ee/genesis/src/lib.rs
sapinb: AFAIK, this value will never need be used in the code, but just exists as its part of the struct. Its not part of anything on chain.

sapinb: But feel free to have fun with it later 😄

crates/alpen-ee/exec-chain/src/state.rs
storopoli: IT's fine to do expect in the `bin/` though...

PR #1190, opened 2025-12-07
feat(bridge-subprotocol): add parsing and handling of unstake tx

prajwolrg:
## Description

This PR implements the core logic for parsing and handling unstake transactions in the Bridge V1 subprotocol. It adds validation to ensure that unstake transactions correctly reference valid stake outputs locked to historical N/N multisig scripts and updates the state by removing the unstaking operator.

This PR also makes move `parser.rs` to from subprotocol crate to txs crate.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Builds on top of #1184.

## Checklist

### DIFF
--- a/crates/asm/subprotocols/bridge-v1/src/handler.rs
+++ b/crates/asm/subprotocols/bridge-v1/src/handler.rs
@@ -5,24 +5,16 @@ use strata_asm_txs_bridge_v1::parser::{ParsedDepositTx, ParsedTx};
 
 use crate::{SlashValidationError, errors::BridgeSubprotocolError, state::BridgeV1State};
 
-/// Handles parsed transactions and update the bridge state accordingly.
+/// Handles parsed bridge transactions.
 ///
-/// # Transaction Types and Log Behavior:
-/// - **Deposit**: Processes the deposit transaction without emitting logs
-/// - **WithdrawalFulfillment**: Processes the withdrawal and emits a withdrawal processed log via
-///   the relayer to notify other components of the processed withdrawal
-/// - **Slash**: Processes slash transactions after validating the stake connector is locked to any
-///   valid N/N multisig.
-///
-/// # Arguments
-/// * `state` - Mutable reference to the bridge state to be updated
-/// * `parsed_tx` - The parsed transaction to handle
-/// * `verified_aux_data` - Auxiliary data containing referenced transaction outputs
-/// * `relayer` - The message relayer used for emitting logs
+/// This function processes each transaction type according to its specific requirements:
+/// - Validating transaction-specific rules and constraints
+/// - Updating the bridge state
+/// - Emitting logs or relaying InterProtocolMsg if needed
 ///
 /// # Returns
 /// * `Ok(())` if the transaction was processed successfully
-/// * `Err(BridgeSubprotocolError)` if an error occurred during processing
+/// * `Err(BridgeSubprotocolError)` if validation fails or an error occurred during processing
 pub(crate) fn handle_parsed_tx<'t>(
     state: &mut BridgeV1State,
     parsed_tx: ParsedTx<'t>,
@@ -89,16 +81,7 @@ pub(crate) fn handle_parsed_tx<'t>(
 /// Pre-processes a parsed transaction to collect auxiliary data requests.
 ///
 /// This function inspects the transaction type and requests any additional data needed
-/// for full verification during the main processing phase. Currently handles:
-///
-/// - **Deposit transactions**: No auxiliary data required
-/// - **Withdrawal fulfillment transactions**: No auxiliary data required
-/// - **Slash transactions**: Requests the Bitcoin transaction spent by the stake connector (second
-///   input). We need this information to verify the stake connector is locked to a known N/N
-///   multisig.
-/// - **Unstake transactions**: Requests the Bitcoin transaction spent by the stake connector
-///   (second input). We need this information to verify the stake connector is locked to a known
-///   N/N multisig.
+/// for the main processing phase.
 pub(crate) fn preprocess_parsed_tx<'t>(
     parsed_tx: ParsedTx<'t>,
     _state: &BridgeV1State,
@@ -108,6 +91,8 @@ pub(crate) fn preprocess_parsed_tx<'t>(
         ParsedTx::Deposit(_) => {}
         ParsedTx::WithdrawalFulfillment(_) => {}
         ParsedTx::Slash(info) => {
+            // Requests the Bitcoin transaction spent by the stake connector (second input). We need
+            // this information to verify the stake connector is locked to a known N/N multisig.
             collector.request_bitcoin_tx(info.second_inpoint().0.txid);
         }
         ParsedTx::Unstake(_) => {}
--- a/crates/asm/txs/bridge-v1/src/unstake/mod.rs
+++ b/crates/asm/txs/bridge-v1/src/unstake/mod.rs
@@ -1,12 +1,20 @@
 //! Unstake Transaction Parser
 //!
-//! This module provides functionality for parsing Bitcoin unstake transactions
+//! This module provides functionality for parsing Bitcoin unstaking intent transactions
 //! that follow the SPS-50 specification for the Strata bridge protocol.
+//!
+//! ## Note on Terminology
+//!
+//! In the ASM context, "unstake transaction" refers to the **unstaking intent** transaction.
+//! While there is a separate unstaking transaction that actually transfers funds back to the
+//! operator, ASM only parses and validates the unstaking intent transaction. Once the unstaking
+//! intent transaction is observed on-chain, the operator is immediately removed from bridge
+//! duties, regardless of whether the final unstaking transaction is seen.
 //!
 //! ## Unstake Transaction Structure
 //!
-//! An unstake transaction is posted by an operator if it wants to exit from bridge duties and
-//! have its staked funds back.
+//! An unstaking intent transaction is posted by an operator when it wants to exit from bridge
+//! duties and have its staked funds returned.
 //!
 //! ### Inputs
 //! - 1. **Stake connector**: Locked to the N-of-N multisig and hashlock.
@@ -19,10 +27,6 @@
 //!     - Transaction type (1 byte): Unstake transaction type
 //!     - Auxiliary data (4 bytes):
 //!         - Operator index (4 bytes, encoded using [`strata_codec::Codec`] which uses big-endian)
-//!
-//! Additional output sends the stake to the operator, but ASM skips validating them because
-//! correctness is assumed to be enforced during presigning as they spend from the same N/N
-//! multisig.
 
 mod aux;
 mod info;
--- a/crates/asm/txs/bridge-v1/src/unstake/parse.rs
+++ b/crates/asm/txs/bridge-v1/src/unstake/parse.rs
@@ -23,6 +23,9 @@ pub const STAKE_INPUT_INDEX: usize = 0;
 /// Enforcing the length lets us index directly and fail fast on malformed witnesses.
 const STAKE_WITNESS_ITEMS: usize = 4;
 
+/// Index of the executed script in witness stack.
+const SCRIPT_INDEX: usize = 2;
+
 /// Parse an unstake transaction to extract [`UnstakeInfo`].
 ///
 /// Parses an unstake transaction following the SPS-50 specification and extracts the auxiliary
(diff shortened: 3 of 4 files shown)


crates/asm/subprotocols/bridge-v1/src/handler.rs
```diff
@@ -65,6 +61,26 @@ pub(crate) fn handle_parsed_tx<'t>(
             // Remove the slashed operator
             state.remove_operator(info.header_aux().operator_idx());
 
+            Ok(())
+        }
+        ParsedTx::Unstake(info) => {
```
barakshani, line 59: Do this?
```
ParsedTx::Slash(info) || ParsedTx::Unstake(info) => {
```
: )

crates/asm/subprotocols/bridge-v1/src/handler.rs
```diff
@@ -91,23 +110,24 @@ pub(crate) fn preprocess_parsed_tx<'t>(
         ParsedTx::Slash(info) => {
             collector.request_bitcoin_tx(info.second_inpoint().0.txid);
         }
+        ParsedTx::Unstake(info) => {
```
barakshani: as suggested abive

crates/asm/txs/bridge-v1/src/unstake/mod.rs
```diff
@@ -0,0 +1,40 @@
+//! Unstake Transaction Parser
+//!
+//! This module provides functionality for parsing Bitcoin unstake transactions
+//! that follow the SPS-50 specification for the Strata bridge protocol.
+//!
+//! ## Unstake Transaction Structure
+//!
+//! An unstake transaction is posted by an operator if they want to exit from the bridge and have
+//! their staked fund back.
```
barakshani:
```suggestion
//! An unstake transaction is posted by an operator if it wants to exit from bridge duties and 
//! have its staked funds back.

File: crates/asm/subprotocols/bridge-v1/src/handler.rs (hunk 1 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/bridge-v1/src/handler.rs:59]: The validation logic might be different.

File: crates/asm/subprotocols/bridge-v1/src/handler.rs (hunk 2 above)
Comment (prajwolrg) [reply crates/asm/subprotocols/bridge-v1/src/handler.rs]: The validation logic might be different.

File: crates/asm/subprotocols/bridge-v1/src/handler.rs (hunk 1 above)
Comment (barakshani) [reply crates/asm/subprotocols/bridge-v1/src/handler.rs:59]: if it would, then we'll need to change the code as well / anyway.

File: crates/asm/txs/bridge-v1/src/unstake/mod.rs [hunk 4]
```diff
@@ -0,0 +1,40 @@
...
+//! This module provides functionality for parsing Bitcoin unstake transactions
+//! that follow the SPS-50 specification for the Strata bridge protocol.
+//!
+//! ## Unstake Transaction Structure
+//!
+//! An unstake transaction is posted by an operator if it wants to exit from bridge duties and 
+//! have its staked funds back.
+//!
+//! ### Inputs
+//! - 1. **Unstaking Intent connector**: Locked to the N-of-N multisig with a relative timelock
```
barakshani: so this should be changed now

crates/asm/subprotocols/bridge-v1/src/handler.rs
prajwolrg: Yes, the validation happens here. so won't be able to do:

```rust
ParsedTx::Slash(info) || ParsedTx::Unstake(info) => {
```

barakshani: I'm saying that _in that case_ we will change the code and separate them (and the point in the previous message was that the code would need to be changed anyway)

crates/asm/txs/bridge-v1/src/errors.rs
```diff
@@ -1,10 +1,30 @@
...
+    DepositTxParse(#[from] DepositTxParseError),
+
+    #[error("failed to parse withdrawal fulfillment tx")]
+    WithdrawalTxParse(#[from] WithdrawalParseError),
+
+    #[error("failed to parse slash tx")]
+    SlashTxParse(#[from] SlashTxParseError),
+
+    #[error("failed to parse unstake tx")]
+    UnstakeTxParse(#[from] UnstakeTxParseError),
```
delbonis: Wouldn't we want to expose the underlying error messages in these?

crates/asm/txs/bridge-v1/src/errors.rs
```diff
@@ -142,3 +162,14 @@ pub enum SlashTxParseError {
...
+
+/// Errors that can occur when parsing unstake transaction.
+#[derive(Debug, Error)]
+pub enum UnstakeTxParseError {
+    /// The auxiliary data in the unstake transaction is invalid
+    #[error("Invalid auxiliary data")]
+    InvalidAuxiliaryData(#[from] CodecError),
+
+    #[error("Missing input at index {0}")]
+    MissingInput(usize),
```
delbonis, line 174: Hmm, maybe we could have some common tx parsing error type for variants like these?  It's very common/generic to be able to say the tx doesn't match the expected structure.  I'm a little worried about having an explosion of different error types that fan in to a big one (like above) but then we have to do annoying matching across the different variants if we want to catch a common condition like that.

crates/asm/txs/bridge-v1/src/lib.rs
```diff
@@ -11,7 +11,9 @@ pub mod constants;
 pub mod deposit;
 pub mod deposit_request;
 pub mod errors;
+pub mod parser;
 pub mod slash;
+pub mod unstake;
 pub mod withdrawal_fulfillment;
```
delbonis, line 17: can we unexport these?

crates/asm/txs/bridge-v1/src/errors.rs
prajwolrg: Will tackle this separately in a different PR. Created a [task](https://alpenlabs.atlassian.net/browse/STR-1970).

crates/asm/txs/bridge-v1/src/test_utils/unstake.rs
barakshani, line 1: Is this used for testing? If so, why isn't it part of test utils?

crates/asm/subprotocols/bridge-v1/src/handler.rs
```diff
@@ -80,6 +96,9 @@ pub(crate) fn handle_parsed_tx<'t>(
 /// - **Slash transactions**: Requests the Bitcoin transaction spent by the stake connector (second
 ///   input). We need this information to verify the stake connector is locked to a known N/N
 ///   multisig.
+/// - **Unstake transactions**: Requests the Bitcoin transaction spent by the stake connector
+///   (second input). We need this information to verify the stake connector is locked to a known
+///   N/N multisig.
```
Rajil1213: I think this doc comment is not relevant anymore.

crates/asm/txs/bridge-v1/src/unstake/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+//! 1. **OP_RETURN Output (Index 0)** (required): Contains SPS-50 tagged data with
+//!     - Magic number (4 bytes): Protocol instance identifier
+//!     - Subprotocol ID (1 byte): Bridge v1 subprotocol identifier
+//!     - Transaction type (1 byte): Unstake transaction type
+//!     - Auxiliary data (4 bytes):
+//!         - Operator index (4 bytes, encoded using [`strata_codec::Codec`] which uses big-endian)
+//!
+//! Additional output sends the stake to the operator, but ASM skips validating them because
+//! correctness is assumed to be enforced during presigning as they spend from the same N/N
+//! multisig.
```
Rajil1213: You should leave a note here to say that this is actually the `Unstaking Intent` transaction and not the `Unstaking` transaction. It might also be useful to clarify why we're parsing the former and not the latter.

crates/asm/txs/bridge-v1/src/unstake/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+
+mod aux;
+#[cfg(feature = "test-utils")]
+mod build;
```
Rajil1213: As suggested by Barak, why not just move it into the `test_utils` module instead of feature-gating?

crates/asm/txs/bridge-v1/src/test_utils/unstake.rs
prajwolrg: It's flagged behind `test-utils` feature. I think this is more cleaner as things are together. I will be updating the structure of other transactions to follow this pattern in the follow up PR.

crates/asm/txs/bridge-v1/src/unstake/parse.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    let witness_len = witness.len();
+    if witness_len != STAKE_WITNESS_ITEMS {
+        return Err(UnstakeTxParseError::InvalidStakeWitnessLen {
+            expected: STAKE_WITNESS_ITEMS,
+            actual: witness_len,
+        });
+    }
+    // With fixed layout, grab the script directly (index 2).
+    let script = ScriptBuf::from_bytes(witness[2].to_vec());
```
Rajil1213: Nit: I recommend using a `const SCRIPT_INDEX: usize = 2` here instead of hardcoding `2`.

crates/asm/txs/bridge-v1/src/unstake/script.rs
```diff
@@ -0,0 +1,299 @@
...
+        _ => return None,
+    };
+
+    // Skip instructions 1-5 (static opcodes - will be validated via reconstruction)
+    for _ in 0..5 {
+        instructions.next();
+    }
+
+    // Extract stake_hash from instruction 6
+    let stake_hash = match instructions.next() {
```
Rajil1213: If `instructions` is an `Iterator`, you should be able to directly do `.nth(STAKE_HASH_INDEX)`.

crates/asm/txs/bridge-v1/src/unstake/script.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    // STEP 3: Reconstruct what the script SHOULD be
+    let expected_script = stake_connector_script(stake_hash_bytes, pubkey);
+
+    // STEP 4: Byte-for-byte comparison - only return params if script matches exactly
+    if script == &expected_script {
+        Some((pubkey, stake_hash_bytes))
+    } else {
+        None
+    }
```
Rajil1213: You should be able to do this instead.

```suggestion
    (script == &expected_script).then(|| (pubkey, stake_hash_bytes))
```

crates/asm/txs/bridge-v1/src/unstake/script.rs
```diff
@@ -0,0 +1,299 @@
...
+            .push_opcode(OP_EQUALVERIFY)
+            .push_opcode(OP_SHA256)
+            .push_slice(stake_hash)
+            .push_opcode(OP_EQUAL)
+            .push_opcode(OP_DROP) // Extra instruction!
+            .into_script();
+
+        // Validation should fail
+        let result = validate_and_extract_script_params(&wrong_script);
+        assert!(result.is_none());
```
Rajil1213: It's good practice to add messages with these assertions directly instead of adding a comment.

```suggestion
        assert!(result.is_none(), "validation and extraction must fail");
```

crates/asm/subprotocols/bridge-v1/src/handler.rs
barakshani: why not?

Rajil1213: The implementation of the `pre_process` function for the Unstake transaction does not request any aux data.

crates/asm/txs/bridge-v1/src/test_utils/unstake.rs
prajwolrg: Discussed with Rajil again on what might be better and reached the conclusion that it is better inside `test-utils` in this specific context since the crate purpose is to parse transaction and not to build it. We only build it for testing purposes.

crates/asm/txs/bridge-v1/src/lib.rs
prajwolrg: Will do this in a follow up PR.

crates/asm/txs/bridge-v1/src/unstake/script.rs
```diff
@@ -0,0 +1,257 @@
...
+    // Extract pubkey from instruction at PUBKEY_INDEX
+    let pubkey = match script.instructions().nth(PUBKEY_INDEX) {
+        Some(Ok(Instruction::PushBytes(bytes))) if bytes.len() == 32 => {
+            bytes.as_bytes().try_into().ok()?
+        }
+        _ => return None,
+    };
+
+    // Extract stake_hash from instruction at STAKE_HASH_INDEX
+    let stake_hash = match script.instructions().nth(STAKE_HASH_INDEX) {
```
Rajil1213, line 53: This feels like something the compiler would optimize anyway but you shouldn't have to create the iterator twice. Either way this is too tiny an optimization to be a blocker for this PR.

```suggestion
    let mut instructions = script.instructions();

    let pubkey = match instructions.nth(PUBKEY_INDEX) {
    ...
    };

    let stake_hash = match instructions.nth(STAKE_HASH_INDEX) {
    ...
    };
```

PR #1191, opened 2025-12-08
STR-1808: implement the OL RPC

storopoli:
## Description

- Make the types in `crates/rpc-new/types/src/types.rs` non-`pub` with accessors.
- Adds the OL RPC Server to `bin/strata`
- Completes the OL RPC server implementation by integrating the mempool for transaction submission and implementing account state extraction for the remaining RPC methods.

### Type of Change

- [x] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- `get_acct_epoch_summary`: Fetches epoch commitment, retrieves OL state at terminal block, extracts account state including snark-specific data (`seqno`, `proof_state`)
- `get_blocks_summaries`: Iterates over slot range, retrieves OL state at each block, extracts account state with balance and sequence numbers
- `submit_transaction`: Converts RPC transaction to mempool format and submits via `MempoolHandle` with proper error code mapping
- Refactors: Renamed `includes_l2_block`/`includes_l1_block` to `l2_slot_at_or_before_end`/`l1_height_at_or_before_end` to clarify these only check upper bounds

The `extra_data` and `processed_msgs` fields in `RpcAccountEpochSummary`, and `updates`/`new_inbox_messages` in `RpcAccountBlockSummary` return empty vectors as the tracking infrastructure for these is not yet available.

Done with the assistance of Cursor.

Please review it commit by commit.

## Checklist

## Related Issues

STR-1808

### DIFF
--- a/bin/strata/src/rpc.rs
+++ b/bin/strata/src/rpc.rs
@@ -286,6 +286,11 @@ impl OLClientRpcServer for OLRpcServer {
             // Extract snark-specific data if applicable.
             // For non-snark accounts, these fields are zeroed since seqno and inbox
             // concepts don't apply to them.
+            //
+            // Note: Unlike `get_acct_epoch_summary` which returns a full `ProofState`,
+            // `RpcAccountBlockSummary` only has `next_inbox_msg_idx` as a separate field
+            // (no `inner_state`). This is by design - per-block summaries focus on
+            // tracking changes rather than full proof state.
             let (next_seq_no, next_inbox_msg_idx) = match account_state.as_snark_account() {
                 Ok(snark_state) => {
                     let seqno: u64 = *snark_state.seqno().inner();


bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,189 @@
...
+        // For now, use the first commitment if available
+        // TODO: This should be more sophisticated - we might need to determine which commitment
+        // corresponds to the canonical chain
+        let epoch_commitment = commitments.first().ok_or_else(|| {
+            ErrorObjectOwned::owned(
+                INVALID_PARAMS_CODE,
+                format!("No epoch commitment found for epoch {epoch}"),
+                None::<()>,
+            )
+        })?;
```
storopoli, line 77: @bewakes please take a look at this TODO and see if we should leave it or if there is a way around it.

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,189 @@
...
+                    format!("No chainstate found for terminal block {terminal_blkid}"),
+                    None::<()>,
+                )
+            })?;
+
+        // TODO: Access OL account state from execution environment state
+        // For now, return placeholder data as OL account state access is not yet implemented
+        // The account state is stored in the execution environment's state root (cur_state),
+        // but we need a way to access the actual account data from that root.
+        let _chainstate = chainstate.into_toplevel();
```
storopoli: @bewakes please take a look at this TODO and see if we should leave it or if there is a way around it.

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,189 @@
...
+        );
+
+        // TODO: Optionally persist to storage if there's a suitable table
+        // For now, just log and return unimplemented error
+
+        Err(ErrorObjectOwned::owned(
+            INTERNAL_ERROR_CODE,
+            "OL mempool not implemented; transaction was logged but not processed",
+            None::<()>,
+        ))
```
storopoli: @bewakes and @krsnapaudel please take a look at this TODO and see if we should leave it or if there is a way around it.

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,189 @@
...
+use tracing::{error, warn};
+
+/// OL RPC server implementation.
+pub(crate) struct OlRpcServer {
+    /// Storage backend.
+    storage: Arc<NodeStorage>,
+
+    /// Status channel.
+    status_channel: Arc<StatusChannel>,
+}
```
delbonis, line 37: I would like to encourage a structure [like this](https://github.com/alpenlabs/mosaic/tree/main/crates/rpc) to abstract out how the RPC server gets its stuff and enable better RPC testing (with mocks).

Also, for consistency this should be called `OLRpcServer`.

bin/strata/src/rpc.rs
delbonis: We don't have any of the indexing to support this RPC call yet, so it's just impossible to implement this.

bin/strata/src/rpc.rs
delbonis: >         // The account state is stored in the execution environment's state root (cur_state),

I'm not really sure what is meant by this?

bin/strata/src/rpc.rs
storopoli: Done in 814468b4b176f097e8f06a2dec841c8d14f362ce

bin/strata/src/rpc.rs
bewakes: This will be handled by the mempool

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,352 @@
...
+            .ok_or_else(|| {
+                ErrorObjectOwned::owned(
+                    INVALID_PARAMS_CODE,
+                    format!("No epoch summary found for epoch commitment {epoch_commitment}"),
+                    None::<()>,
+                )
+            })?;
+
+        // Get OL state at the terminal block
+        let terminal_commitment = OLBlockCommitment::new(
```
bewakes: The terminal slot and block id can be obtained from the epoch commitment itself. So I think we don't need to fetch epoch summary. Please check.

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,352 @@
...
+            epoch_summary.terminal().slot(),
+            *epoch_summary.terminal().blkid(),
+        );
+        let ol_state = self
+            .storage
+            .ol_state()
+            .get_toplevel_ol_state_async(terminal_commitment)
+            .await
+            .map_err(|e| {
```
bewakes, line 87: nit: Can we chain all/most of these with `and_then` ? All of this error handling is making the body very long.

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,352 @@
...
+            return Err(ErrorObjectOwned::owned(
+                INVALID_PARAMS_CODE,
+                "start_slot must be <= end_slot",
+                None::<()>,
+            ));
+        }
+
+        let mut summaries = Vec::new();
+
+        for slot in start_slot..=end_slot {
```
bewakes: If we fetch the blocks with slot then we also need to check if they are actually chained or not. Because right now we can't guarantee if they are chained or not, they might be from different trees.

Maybe first fetching the block from the end slot and then fetching the parent block can help?

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,352 @@
...
+                next_inbox_msg_idx,
+            ));
+        }
+
+        Ok(summaries)
+    }
+
+    async fn submit_transaction(&self, tx: RpcOLTransaction) -> RpcResult<OLTxId> {
+        // Convert RPC transaction to mempool transaction
+        let mempool_tx: OLMempoolTransaction = tx.try_into().map_err(|e| {
```
bewakes: This `map_err` is getting repititive. We really need a helper function that returns this map error function. we could then do something like 
```rust
result
    .map_err(error_fn(code, msg))
```

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,352 @@
...
+        Ok(txid)
+    }
+}
+
+/// Maps mempool errors to RPC errors with appropriate error codes.
+fn map_mempool_error_to_rpc(err: OLMempoolError) -> ErrorObjectOwned {
+    match &err {
+        // Capacity-related errors
+        OLMempoolError::MempoolFull { .. } | OLMempoolError::MempoolByteLimitExceeded { .. } => {
+            ErrorObjectOwned::owned(-32001, err.to_string(), None::<()>)
```
bewakes: maybe have a const for this `-32001` as well?

bin/strata/src/run_context.rs
```diff
@@ -22,4 +25,8 @@ pub(crate) struct RunContext {
     // Service handles
     pub asm_handle: AsmWorkerHandle,
     pub csm_monitor: ServiceMonitor<CsmWorkerStatus>,
+    // RPC dependencies
+    pub storage: Arc<NodeStorage>,
+    pub status_channel: Arc<StatusChannel>,
+    pub mempool_handle: MempoolHandle,
```
bewakes: There's a section for service handles above, should be moved there?

bin/strata/src/services.rs
```diff
@@ -36,24 +41,70 @@ pub(crate) fn start_services(nodectx: NodeContext) -> Result<RunContext> {
...
+        .status_channel
+        .get_chain_sync_status()
+        .map(|status| status.tip)
+        .ok_or_else(|| anyhow!("Chain sync status not available, cannot start mempool"))?;
+
+    let storage = nodectx.storage.clone();
+    let status_channel = (*nodectx.status_channel).clone();
+    let executor = nodectx.executor.clone();
+
+    nodectx.runtime.block_on(async {
```
bewakes, line 78: Not sure if we need to block_on here.

bin/strata/src/services.rs
```diff
@@ -36,24 +41,70 @@ pub(crate) fn start_services(nodectx: NodeContext) -> Result<RunContext> {
...
+    );
     Ok(())
 }
 
-async fn spawn_rpc(rpc_host: String, rpc_port: u16) -> Result<()> {
+/// Spawns the RPC server.
+async fn spawn_rpc(
+    rpc_host: String,
+    rpc_port: u16,
+    storage: Arc<strata_storage::NodeStorage>,
```
bewakes: do you think we can pass around `RunContext` here to reduce the params count? Also because all of these derive from the context itself.

storopoli: See 202cf5cb16d678a989b17d80e04727148289a21b

bin/strata/src/services.rs
storopoli: `block_on` is required because `start_services` is synchronous but we need to initialize the mempool which requires `async` operations. The mempool handle must be available before `RunContext` is constructed.

Added a comment in 202cf5cb16d678a989b17d80e04727148289a21b

bin/strata/src/rpc.rs
storopoli: See 4db5981a214c8ce78654fafb794f653ffb7f5301

bin/strata/src/rpc.rs
storopoli: See 4db5981a214c8ce78654fafb794f653ffb7f5301

bin/strata/src/rpc.rs
storopoli: See 6ca713775fbd3236acc29fa7772a97c8ac91cbd0

bin/strata/src/rpc.rs
storopoli: See 4db5981a214c8ce78654fafb794f653ffb7f5301

bin/strata/src/rpc.rs
storopoli: See 09aa632d5d620c2c66da5ca3fd1bc681e56dd00f

bin/strata/src/run_context.rs
storopoli: See https://github.com/alpenlabs/alpen/commit/202cf5cb16d678a989b17d80e04727148289a21b

bin/strata/src/rpc.rs
bewakes: Beautiful!

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,351 @@
...
+use strata_rpc_types_new::{
+    RpcAccountBlockSummary, RpcAccountEpochSummary, RpcOLChainStatus, RpcOLTransaction,
+};
+use strata_snark_acct_types::ProofState;
+use strata_status::StatusChannel;
+use strata_storage::NodeStorage;
+use tracing::error;
+
+/// Custom error code for mempool capacity-related errors.
+const MEMPOOL_CAPACITY_ERROR_CODE: i32 = -32001;
```
bewakes: nit: this should probably be with the other error code consts.

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,351 @@
...
+
+/// Creates an RPC error for database failures.
+fn db_error(e: impl Display) -> ErrorObjectOwned {
```
bewakes: nit: Just for readibility, this and the other helpers would be better off at the end of this module.

bewakes (dismissed):
Minor nits, but mergeable definitely.

Also, what do you think about having unit tests for rpc? We don't have unit tests for the old rpc either(maybe because they are used in the func tests). But I still think those might be valuable.
If you agree, we can create a ticket for unit testing of OL RPC.

crates/checkpoint-types/src/batch.rs
```diff
@@ -183,21 +183,22 @@ impl BatchInfo {
         &self.l1_range.1
     }
 
-    /// Check is whether the L2 slot is covered by the checkpoint
-    pub fn includes_l2_block(&self, slot: Slot) -> bool {
+    /// Check if the L2 slot is at or before the end of this checkpoint's L2 range.
+    ///
+    /// Note: This only checks the upper bound, not the lower bound. It returns `true`
+    /// if the slot is <= the checkpoint's final L2 slot. Use this to check if a slot
+    /// could be covered by this checkpoint (assuming it's after the previous checkpoint's end).
```
barakshani:
```suggestion
    /// is included in any batch up to this one (including).
```

> (assuming it's after the previous checkpoint's end)

why? you literally just write that the slot can be an any checkpoint.

bin/strata-client/src/rpc_server.rs
```diff
@@ -617,7 +617,10 @@ impl StrataApiServer for StrataRpcImpl {
         // that behavior here
```
barakshani: let's use this opportunity to fix the fixme

barakshani: btw, doesn't "Finalized check" and "Verified check" do exactly the same thing??

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,351 @@
...
+                internal_error(format!("Account error: {e}"))
+            })?
+            .ok_or_else(|| not_found_error(format!("Account {account_id} not found")))?;
+
+        // Extract snark-specific data if applicable
+        let (next_seq_no, proof_state) = match account_state.as_snark_account() {
+            Ok(snark_state) => {
+                let seqno: u64 = *snark_state.seqno().inner();
+                let inner_state = snark_state.inner_state_root();
+                let next_inbox_idx = snark_state.get_next_inbox_msg_idx();
```
barakshani, line 113: it is not possible to call this directly? https://github.com/alpenlabs/alpen/blob/main/crates/snark-acct-types/src/state.rs#L59

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,351 @@
...
+                .ok()
+                .and_then(|c| c.first().copied());
+            prev_commitments.unwrap_or_else(EpochCommitment::null)
+        } else {
+            EpochCommitment::null()
+        };
+
+        Ok(RpcAccountEpochSummary::new(
+            *epoch_commitment,
+            prev_epoch_commitment,
```
barakshani, line 135: I understand this struct is implemented elsewhere, but do you have any idea why we return prev epoch?

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,351 @@
...
+    ) -> RpcResult<Vec<RpcAccountBlockSummary>> {
+        if start_slot > end_slot {
+            return Err(invalid_params_error("start_slot must be <= end_slot"));
+        }
+
+        // Walk backwards from end_slot via parent references to ensure blocks are chained.
+        // This guarantees all returned blocks are from the same chain, not different forks.
+        let mut chain_blocks: Vec<(u64, OLBlockId)> = Vec::new();
+
+        // Get the starting block at end_slot
```
barakshani: starting block? too confusing.

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,351 @@
...
+
+        let Some(current_block_id) = end_block_ids.first().copied() else {
+            // No block at end_slot, return empty
+            return Ok(Vec::new());
+        };
+
+        // Walk backwards from end_slot to start_slot following parent references
+        let mut current_id = current_block_id;
+
+        loop {
```
barakshani, line 198: I'm not 100% sure I understand what is being done here, but it seems to me that because we may have a block tree -- for non final blocks -- we need to make sure that we return a chain on that tree. However, if at any point in time the current block is a final block, we are guaranteed to have a single chain, going backward, so it may be more efficient to just get all (previous) blocks from that point, instead of verifying parents.

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,351 @@
...
+            .ok_or_else(|| internal_error("Chain sync status not available"))?;
+
+        let latest = chain_sync_status.tip;
+        let confirmed = chain_sync_status.prev_epoch;
+        let finalized = chain_sync_status.finalized_epoch;
+
+        Ok(RpcOLChainStatus::new(latest, confirmed, finalized))
+    }
+
+    async fn get_blocks_summaries(
```
barakshani, line 157: maybe add some docstring description for this function?
we pass start/end blocks, which makes sense to me given the function name -- but then we also pass account id, which seems like we're requesting specific value per the account -- I am not sure what this will be used for, maybe how an EE views how to OL sees it, so will be good to clarify.

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,351 @@
...
+            };
+
+            // Extract snark-specific data if applicable
+            let (next_seq_no, next_inbox_msg_idx) = match account_state.as_snark_account() {
+                Ok(snark_state) => {
+                    let seqno: u64 = *snark_state.seqno().inner();
+                    let next_inbox_idx = snark_state.get_next_inbox_msg_idx();
+                    (seqno, next_inbox_idx)
+                }
+                Err(_) => (0, 0),
```
barakshani: why not Null? how can I know if `(next_seq_no, next_inbox_msg_idx) = (0,0)` or I hit the error path?

bin/strata/src/rpc.rs
```diff
@@ -0,0 +1,351 @@
...
+                error!(?e, %account_id, slot, "Failed to get account state");
+                internal_error(format!("Account error: {e}"))
+            })?;
+
+            let Some(account_state) = account_state else {
+                continue; // Account not found at this slot
+            };
+
+            // Extract snark-specific data if applicable
+            let (next_seq_no, next_inbox_msg_idx) = match account_state.as_snark_account() {
```
barakshani, line 294: above we return ProofState, why not here?

bin/strata/src/rpc.rs
storopoli: No. `as_snark_account()` returns `&NativeSnarkAccountState` from `strata_ol_state_types`; a different type that doesn't expose a `proof_state()` getter (the field is private).

barakshani: believe me, I don't know why we have this proof state.

bin/strata/src/rpc.rs
storopoli: The type was designed in #1159 by @bewakes.

bin/strata/src/rpc.rs
storopoli: If there's a desire to make these consistent, that would require modifying the `RpcAccountBlockSummary` type in `crates/rpc-new/types/src/account_summary.rs` to add a `proof_state` field (and include `inner_state`), which would be a separate change outside of this PR scope.

Clarified in 9bc370e539c36275f3b318f162aa0d6484a8c722

bin/strata/src/rpc.rs
storopoli: The issue is that for non-snark accounts, returning `(0, 0)` is indistinguishable from a snark account that legitimately has those values.

Options:
1. Make fields `Option<u64>` in `RpcAccountBlockSummary`. Requires changing types from PR #1159
2. Return error for non-snark accounts - but then the RPC only works for snark accounts.
3. Document the behavior

I went with 3 in fb413317670386f0c300dc0aa00891301aa66564

bin/strata/src/rpc.rs
storopoli: See 65f3193829367697e2e266d599d185a701787c06

bin/strata/src/rpc.rs
storopoli: See a5ede3bb549cea7c98bbdd0200ebd213125a4009

bin/strata/src/rpc.rs
storopoli: See e1176e72fb56f8193b7e821e435e16c6b4848e03

bin/strata-client/src/rpc_server.rs
storopoli: See d5e5f2b9b71189900515ee1da5805b06647bb17f

crates/checkpoint-types/src/batch.rs
storopoli: See 661a4fe0f2ea0eaadfe979444343e0a5a899a6b0

bin/strata/src/rpc.rs
storopoli: See 2a38a0f1785c36a26144e9a559c21555caf78460

bin/strata/src/rpc.rs
storopoli: See 2a38a0f1785c36a26144e9a559c21555caf78460

bin/strata/src/rpc.rs
barakshani: perfect thanks!

bin/strata-client/src/rpc_server.rs
```diff
@@ -612,28 +612,28 @@ impl StrataApiServer for StrataRpcImpl {
...
+                return Ok(L2BlockStatus::Verified(
+                    last_checkpoint.l1_reference.block_height(),
+                ));
             }
-        });
-        if let Some(l1_height) = verified_l1_height {
-            return Ok(L2BlockStatus::Verified(l1_height));
         }
 
         // Confirmed check
```
barakshani, line 639: this should be <=, I think.

barakshani (approved):

PR #1192, opened 2025-12-08
feat: add bitcoin transaction test helpers and tighten bridge slash validation

prajwolrg:
## Description

* Introduce `strata-test-utils-btcio` to centralize Bitcoin regtest helpers (address derivation, funding/mining, taproot/MuSig2 signing, and submission with single or aggregated keys) and reuse it across tests.
* Rework bridge slash test utilities to construct and broadcast real stake/slash transactions; handler now fetches the referenced stake prevout and rejects slashes unless the stake connector matches a recorded N/N multisig script.
* Misc cleanup: switch btcio to the shared test utils.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

I think some of these functions inside makes sense inside `strata-btcio` instead. I've deliberately avoided using `strata-btcio` because of cyclic dependencies it creates.

## Checklist

### DIFF
--- a/crates/asm/txs/admin/src/test_utils/mod.rs
+++ b/crates/asm/txs/admin/src/test_utils/mod.rs
@@ -13,10 +13,7 @@ use bitcoin::{
     transaction::Version,
 };
 use rand::{RngCore, rngs::OsRng};
-use strata_crypto::{
-    self,
-    threshold_signature::{IndexedSignature, SignatureSet},
-};
+use strata_crypto::threshold_signature::{IndexedSignature, SignatureSet};
 use strata_primitives::buf::Buf32;
 
 pub(crate) const TEST_MAGIC_BYTES: &[u8; 4] = b"ALPN";
--- a/crates/asm/worker/src/state.rs
+++ b/crates/asm/worker/src/state.rs
@@ -152,9 +152,7 @@ mod tests {
     struct TestEnv {
         pub _node: Node, // Keep node alive
         pub client: Arc<Client>,
-        pub context: MockWorkerContext,
         pub service_state: AsmWorkerServiceState<MockWorkerContext>,
-        pub params: Arc<Params>,
     }
 
     async fn setup_env() -> TestEnv {
@@ -182,10 +180,8 @@ mod tests {
 
         let params = Arc::new(params);
 
-        // 3. Setup Worker Context
+        // 3. Set worker context and initialize service state
         let context = MockWorkerContext::new();
-
-        // 4. Initialize Service State
         let mut service_state = AsmWorkerServiceState::new(context.clone(), params.clone());
 
         // Initialize: this should create genesis state based on our `genesis_l1_view`
@@ -201,9 +197,7 @@ mod tests {
         TestEnv {
             _node: node,
             client,
-            context,
             service_state,
-            params,
         }
     }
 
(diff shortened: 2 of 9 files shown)


bin/strata-test-cli/src/bridge/dt.rs
```diff
@@ -136,8 +139,8 @@ fn sign_deposit_transaction(
         .map_err(|e| Error::TxBuilder(format!("Sighash creation failed: {e}")))?;
 
     let msg = sighash.to_byte_array();
-    let tweak_bytes = Some(takeback_hash.to_byte_array());
-    let schnorr_sig = create_musig2_signature(signers, &msg, tweak_bytes);
+    let tweak = Musig2Tweak::TaprootScript(takeback_hash.to_byte_array());
```
irnb, line 142: For clarification: 
don’t DRTs only need a takeback path?

In this context, we’re assuming the DT will also require that path.
I’d appreciate any insight you can share on this.

crates/asm/txs/admin/src/test_utils/mod.rs
```diff
@@ -50,7 +50,7 @@ pub fn create_multisig_signature(
         .map(|index| privkeys[index])
         .collect();
 
-    let signature = create_musig2_signature(&selected_privkeys, &sighash.0, None);
+    let signature = create_musig2_signature(&selected_privkeys, &sighash.0, Musig2Tweak::None);
```
irnb: Shouldn’t we switch this to normal ECDSA?
I know it works in the current context, but after merging PR #1163, this might need to change.

crates/asm/worker/src/state.rs
```diff
@@ -131,3 +131,201 @@ impl<W: WorkerContext + Send + Sync + 'static> ServiceState for AsmWorkerService
...
+            Ok(())
+        }
+
+        fn get_network(&self) -> WorkerResult<Network> {
+            Ok(Network::Regtest)
+        }
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_asm_transition() {
```
irnb, line 288: I like this pattern. Maybe we could move these helper methods and the mock implementation to a separate place and make them public.

We could also generalize the testing pattern so that, for each scenario, in addition to the current logic, we send the ASM subprotocol related transactions as well. This should help reduce code duplication across tests.

irnb (commented):
Great work here!
I’ve left a few clarification questions and a suggestions.

crates/asm/txs/admin/src/test_utils/mod.rs
prajwolrg: This is mostly for bridge transactions. For multisig transactions, we need to create the signature set instead.

bin/strata-test-cli/src/bridge/dt.rs
prajwolrg: DRTs do need a takeback path. It is just encoding it in a different way.

crates/asm/worker/src/state.rs
prajwolrg: Yes, that's the plan.

evgenyzdanovich (approved):

PR #1193, opened 2025-12-09
STR-1924: asm aux input resolution

evgenyzdanovich:
## Description

ASM aux input resolution.

Covers STR-1924, STR-1925

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The impl of MMR db is not a thing I'm proud of and we certainly need to iterate on that (likely with inputs from Trey), but I intentionally made it quite short and directly pluggable into the existing architecture.

## Checklist

### DIFF
--- a/crates/consensus-logic/src/asm_worker_context.rs
+++ b/crates/consensus-logic/src/asm_worker_context.rs
@@ -115,7 +115,7 @@ impl WorkerContext for AsmWorkerCtx {
 
     fn store_manifest_hash(&self, index: u64, hash: [u8; 32]) -> WorkerResult<()> {
         self.asmman
-            .store_manifest_hash(index, hash)
+            .store_manifest_hash(index, Buf32(hash))
             .map_err(conv_db_err)
     }
 
@@ -127,7 +127,10 @@ impl WorkerContext for AsmWorkerCtx {
     }
 
     fn get_manifest_hash(&self, index: u64) -> WorkerResult<Option<[u8; 32]>> {
-        self.asmman.get_manifest_hash(index).map_err(conv_db_err)
+        self.asmman
+            .get_manifest_hash(index)
+            .map(|opt| opt.map(|buf| buf.0))
+            .map_err(conv_db_err)
     }
 }
 
--- a/crates/db/store-sled/src/asm/db.rs
+++ b/crates/db/store-sled/src/asm/db.rs
@@ -136,13 +136,13 @@ impl AsmDatabase for AsmDBSled {
         Ok(result)
     }
 
-    fn store_manifest_hash(&self, index: u64, hash: [u8; 32]) -> DbResult<()> {
-        self.manifest_hash_tree.insert(&index, &Buf32(hash))?;
+    fn store_manifest_hash(&self, index: u64, hash: Buf32) -> DbResult<()> {
+        self.manifest_hash_tree.insert(&index, &hash)?;
         Ok(())
     }
 
-    fn get_manifest_hash(&self, index: u64) -> DbResult<Option<[u8; 32]>> {
-        Ok(self.manifest_hash_tree.get(&index)?.map(|buf| buf.0))
+    fn get_manifest_hash(&self, index: u64) -> DbResult<Option<Buf32>> {
+        Ok(self.manifest_hash_tree.get(&index)?)
     }
 }
 
--- a/crates/db/types/src/traits.rs
+++ b/crates/db/types/src/traits.rs
@@ -62,12 +62,12 @@ pub trait AsmDatabase: Send + Sync + 'static {
     ///
     /// This is called after appending a manifest to the MMR to maintain
     /// a fast lookup index from leaf index to manifest hash.
-    fn store_manifest_hash(&self, index: u64, hash: [u8; 32]) -> DbResult<()>;
+    fn store_manifest_hash(&self, index: u64, hash: Buf32) -> DbResult<()>;
 
     /// Gets a manifest hash by MMR leaf index.
     ///
     /// Used by aux data resolver to retrieve manifest hashes for subprotocols.
-    fn get_manifest_hash(&self, index: u64) -> DbResult<Option<[u8; 32]>>;
+    fn get_manifest_hash(&self, index: u64) -> DbResult<Option<Buf32>>;
 }
 
 /// Database interface to control our view of L1 data.
--- a/crates/storage/src/managers/asm.rs
+++ b/crates/storage/src/managers/asm.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use strata_db_types::{traits::AsmDatabase, DbResult};
-use strata_primitives::l1::L1BlockCommitment;
+use strata_primitives::{buf::Buf32, l1::L1BlockCommitment};
 use strata_state::asm_state::AsmState;
 use threadpool::ThreadPool;
 
@@ -51,12 +51,12 @@ impl AsmStateManager {
     }
 
     /// Stores a manifest hash at the given MMR leaf index
-    pub fn store_manifest_hash(&self, index: u64, hash: [u8; 32]) -> DbResult<()> {
+    pub fn store_manifest_hash(&self, index: u64, hash: Buf32) -> DbResult<()> {
         self.ops.store_manifest_hash_blocking(index, hash)
     }
 
     /// Gets a manifest hash by MMR leaf index
-    pub fn get_manifest_hash(&self, index: u64) -> DbResult<Option<[u8; 32]>> {
+    pub fn get_manifest_hash(&self, index: u64) -> DbResult<Option<Buf32>> {
         self.ops.get_manifest_hash_blocking(index)
     }
 }
--- a/crates/storage/src/ops/asm.rs
+++ b/crates/storage/src/ops/asm.rs
@@ -1,7 +1,7 @@
 //! ASM data operation interface.
 
 use strata_db_types::traits::*;
-use strata_primitives::l1::L1BlockCommitment;
+use strata_primitives::{buf::Buf32, l1::L1BlockCommitment};
 use strata_state::asm_state::AsmState;
 
 use crate::exec::*;
@@ -12,7 +12,7 @@ inst_ops_simple! {
         get_asm_state(block: L1BlockCommitment) => Option<AsmState>;
         get_latest_asm_state() => Option<(L1BlockCommitment, AsmState)>;
         get_asm_states_from(from_block: L1BlockCommitment, max_count: usize) => Vec<(L1BlockCommitment, AsmState)>;
-        store_manifest_hash(index: u64, hash: [u8; 32]) => ();
-        get_manifest_hash(index: u64) => Option<[u8; 32]>;
+        store_manifest_hash(index: u64, hash: Buf32) => ();
+        get_manifest_hash(index: u64) => Option<Buf32>;
     }
 }


crates/asm/worker/src/aux_resolver.rs
```diff
@@ -0,0 +1,298 @@
...
+//! Fully implemented with on-demand MMR proof generation:
+//! - Maps L1 block heights to MMR indices using genesis offset
+//! - Fetches manifest hashes from fast lookup storage
+//! - Generates MMR proofs using `SledMmrDb`
+//! - Converts `MerkleProofB32` (SSZ type) to `MerkleProof<Hash32>` (ASM type)
+//!
+//! ### Bitcoin Transaction Resolution ⚠️
+//!
+//! Bitcoin transaction fetching requires transaction indexing infrastructure (see
+//! aux-resolver-blockers.md). Currently returns `WorkerError::BitcoinTxNotFound`.
```
prajwolrg: Can't we directly use `bitcoind-async-client::Client` here to fetch transactions?

crates/asm/worker/src/aux_resolver.rs
```diff
@@ -0,0 +1,298 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `txid` - Transaction ID to fetch
+    ///
+    /// # Errors
+    ///
+    /// Returns `WorkerError::BitcoinTxNotFound` if the transaction cannot be found.
+    fn fetch_bitcoin_tx(&self, _txid: &BitcoinTxid) -> WorkerResult<RawBitcoinTx> {
+        // TODO: Once bitcoind_async_client supports get_raw_transaction, use that directly.
```
prajwolrg: `bitcoind-async-client` already has `get_raw_transaction_verbosity_zero` and `get_raw_transaction_verbosity_one`. Both of them can can return `bitcoin::Transaction`.

crates/consensus-logic/src/asm_worker_context.rs
```diff
@@ -76,6 +96,55 @@ impl WorkerContext for AsmWorkerCtx {
...
+        // Convert BitcoinTxid to bitcoin::Txid
+        let bitcoin_txid = txid.inner();
+
+        // Fetch raw transaction hex from Bitcoin client
+        let raw_tx_hex = self
+            .handle
+            .block_on(
+                self.bitcoin_client
+                    .get_raw_transaction_verbosity_zero(&bitcoin_txid),
+            )
```
prajwolrg: `GetRawTransactionVerbosityZero` has public function `transaction` that does the hex decoding to convert into `bitcoin::Transaction`.

crates/storage/src/mmr_db/memory.rs
```diff
@@ -0,0 +1,564 @@
...
+/// - Append: O(log n) - compute path to new peak
+/// - Proof generation: O(log n) - traverse from leaf to peak
+/// - Storage: ~64 MB for 1M leaves (vs ~320 MB for stored proofs)
+///
+/// This is suitable for:
+/// - Testing
+/// - Short-lived processes
+/// - Production use with in-memory cache
+///
+/// For persistent storage, use `RocksDbMmrDb`.
```
prajwolrg:
```suggestion
/// For persistent storage, use `SledDbMmrDb`.
```

crates/storage/src/mmr_db_test.rs
```diff
@@ -0,0 +1,37 @@
...
+
+        // Try to generate a proof (this will fail if method doesn't exist)
+        // Checking available methods at compile time
+
+        // Convert to compact to see what's available
+        let compact: CompactMmr64 = mmr.to_compact();
+        println!("Compact MMR has {} peaks", compact.peaks_slice().len());
+
+        // Check if verification works
+        // compact.verify::<StrataHasher>(...) - need proof first
```
prajwolrg: We can generate the proof, and update the proof using `add_leaf_updating_proof_list` function in `strata-merkle`. See [here](https://github.com/alpenlabs/strata-common/blob/main/crates/merkle/src/mmr.rs#L924).

benches/src/db.rs
```diff
@@ -22,7 +22,7 @@ pub mod sled {
     /// Creates a temporary `Sled` database instance for benchmarking.
     pub fn create_temp_sled() -> (Arc<SledDb>, TempDir) {
         let temp_dir = TempDir::new().expect("Failed to create temp directory");
-        let db = open_sled_database(temp_dir.path(), "benchmark_db")
+        let (_, db) = open_sled_database(temp_dir.path(), "benchmark_db")
```
prajwolrg: Why does the API needs to be changed? I'm just curious.

crates/db/store-sled/src/asm/schemas.rs
```diff
@@ -14,3 +17,49 @@ define_table_with_seek_key_codec!(
...
+pub struct MmrMetadata {
+    pub num_leaves: u64,
+    pub mmr_size: u64,
+    pub peak_roots: Vec<[u8; 32]>,
+}
+
+define_table_without_codec!(
+    /// A table to store MMR metadata (singleton)
+    (AsmMmrMetaSchema) () => MmrMetadata
+);
```
prajwolrg: Interesting! I hadn't realized we could do singleton storage this way.

prajwolrg (changes requested):
This looks really nice. Thank you!

crates/storage/src/managers/asm.rs
```diff
@@ -49,4 +64,35 @@ impl AsmStateManager {
...
+        let mmr_node_tree = raw_db
+            .open_tree(b"AsmMmrNodeSchema")
+            .map_err(|e| strata_db_types::DbError::Other(e.to_string()))?;
+        let mmr_meta_tree = raw_db
+            .open_tree(b"AsmMmrMetaSchema")
+            .map_err(|e| strata_db_types::DbError::Other(e.to_string()))?;
+
+        SledMmrDb::new(mmr_node_tree, mmr_meta_tree)
+            .map_err(|e| strata_db_types::DbError::Other(e.to_string()))
+    }
```
prajwolrg: MMR database is also needed by OL. Maybe we can make the create_mmr_database take an argument instead and then be able to use different storages for different MMRs? This can be tackled separately as a different task as well.

crates/storage/src/mmr_db/types.rs
```diff
@@ -0,0 +1,110 @@
...
+/// efficient proof generation for arbitrary leaf positions.
+///
+/// ## Design Invariants
+///
+/// - Leaves are indexed from 0 sequentially
+/// - `append_leaf` is the only way to add data (append-only)
+/// - `num_leaves()` always returns the total number of leaves added
+/// - Proofs are valid against the current `root()`
+pub trait MmrDatabase: Send + Sync {
+    /// Append a new leaf to the MMR
```
prajwolrg: I think we should also have function to remove leaf. We should not be able remove leaf at arbitrary position though. Maybe a function like `pop` which removes the last leaf. 

This is needed to handle reorg properly. If we have already added something to `MmrDatabase` and there is a reorg on L1, we need to pop a few leaves and add back some new leaves.

crates/db/store-sled/src/asm/db.rs
```diff
@@ -2,13 +2,22 @@ use strata_db_types::{DbResult, traits::*};
...
 use crate::define_sled_database;
 
 define_sled_database!(
+    #[expect(
+        dead_code,
+        reason = "MMR trees accessed directly via raw sled DB in storage layer"
+    )]
     pub struct AsmDBSled {
         asm_state_tree: AsmStateSchema,
         asm_log_tree: AsmLogSchema,
```
prajwolrg, line 22: Instead of ASM Logs, maybe we should store `AsmManifest` instead.

crates/db/store-sled/src/lib.rs
```diff
@@ -44,15 +44,17 @@ pub fn open_sled_backend(
...
+    let (raw_db, sled_db) = open_sled_database(datadir, dbname)?;
+    SledBackend::new(raw_db, sled_db, ops_config)
         .map_err(|e| anyhow::anyhow!("Failed to initialize sled backend: {}", e))
         .map(Arc::new)
 }
 
 /// Complete Sled backend with all database types
 #[derive(Debug)]
 pub struct SledBackend {
+    /// Raw sled database handle for opening untyped trees (e.g., MMR database)
```
bewakes: I am curious why do we  want untyped trees. The `typed-sled` library's purpose was to have everything typed so that we can have safe db access.

crates/storage/src/mmr_db/sled.rs
```diff
@@ -0,0 +1,420 @@
...
+/// - **Metadata Tree**: metadata key -> MmrMetadata
+/// - **In-memory cache**: peak_roots, num_leaves, mmr_size
+///
+/// # Performance
+///
+/// - Append: O(log n) + Sled writes
+/// - Proof generation: O(log n) Sled reads
+/// - Storage: ~64 MB for 1M leaves
+#[derive(Debug, Clone)]
+pub struct SledMmrDb {
```
bewakes: This seems like something that should be alongside other db impls in `db/store-sled`. Any reason it is defined here in managers? because the current practice is to have typed dbs in `store-sled` and in `managers` we define `ops` which are wrapped by `*ManagerDb`.

crates/db/types/src/mmr_helpers.rs
bewakes, line 1: Most of the functions defined here seem universally useful to be added in the mmr library itself.

crates/db/types/src/mmr_helpers.rs
```diff
@@ -0,0 +1,348 @@
...
+//! ```text
+//!       7
+//!      /  \
+//!     3    6
+//!    / \  / \
+//!   1  2 4  5
+//!
+//! Position: [0, 1, 2, 3, 4, 5, 6, 7]
+//! Height:   [0, 0, 1, 0, 0, 1, 2, 0]
+//! Leaves:   [0, 1, x, 2, 3, x, x, 4]  (x = internal node)
```
bewakes, line 26: Not sure if the info here is correct. A couple of questions:
1. There are 7 elements, but the array has 8. Is zero index discarded?
2. For which element is the height zero, leaf or the root? *Update*: I think I got the answer: leaf is zero :)

bewakes: I think part of the confusion is because the element 1 is in the index 0 right? It might be helpful to name the elements by their index just like the referenced github repo [does](https://github.com/nervosnetwork/merkle-mountain-range?tab=readme-ov-file#construct).

crates/storage/src/mmr_db/sled.rs
```diff
@@ -0,0 +1,420 @@
...
+
+/// Metadata structure for serialization
+#[derive(Debug, Clone, borsh::BorshSerialize, borsh::BorshDeserialize)]
+struct MmrMetadata {
+    num_leaves: u64,
+    mmr_size: u64,
+    peak_roots: Vec<Hash>,
+}
+
+impl MmrDatabase for SledMmrDb {
```
bewakes: Since most of the methods here have almost identical logic in both the in-memory impl and persistent impl(and any other impl that might exist), I think the core mmr logic can be separated out and reused. For example, in `append_leaf` the only thing different are `.get_node()`, `.put_node()`, and possibly `.update_peaks()`.

Similar in `generate_proof()`, the only difference is `.get_node()`.

bewakes: One way might be to  define the append logic in the trait itself, and having the users to implement `get_node()`, `put_node() etc only.

crates/storage/src/lib.rs
```diff
@@ -104,3 +105,46 @@ where
         checkpoint_manager,
     })
 }
+
+/// Creates node storage from Sled backend with MMR database support
+///
+/// This function uses the raw sled database to create an AsmStateManager
```
bewakes: I am pretty sure we should not need the raw database. Please see if it's possible to use type-safe db interfaces from `typed-sled`.

crates/storage/src/managers/asm.rs
```diff
@@ -49,4 +64,35 @@ impl AsmStateManager {
     ) -> DbResult<Vec<(L1BlockCommitment, AsmState)>> {
         self.ops.get_asm_states_from_blocking(from_block, max_count)
     }
+
+    /// Creates a new MMR database instance for proof generation
+    ///
+    /// Opens the raw sled trees used for MMR storage. Requires that the manager
+    /// was created with `new_with_raw_db()`.
+    pub fn create_mmr_database(&self) -> DbResult<SledMmrDb> {
```
bewakes: Again, not sure why we need to create a mmr database here, I think it makes more sense to initialize databases like it's been done for other dbs. This again ties to whether we need raw db or not, whether the mmr db should be typed or not, which I believe should.

crates/consensus-logic/src/asm_worker_context.rs
```diff
@@ -19,6 +19,8 @@ pub struct AsmWorkerCtx {
     bitcoin_client: Arc<Client>,
     l1man: Arc<L1BlockManager>,
     asmman: Arc<AsmStateManager>,
+    /// Lazily initialized MMR database (created on first use)
+    mmr_db: std::sync::Mutex<Option<strata_storage::mmr_db::SledMmrDb>>,
```
bewakes: Is there a reason why manager is not used like we do for other database tables? This links with other comments as well. Perhaps there's some good reason which I can't see yet.

crates/asm/worker/src/aux_resolver.rs
evgenyzdanovich: yeah, there's some mess. I already have it implemented (in the worker context), just have to use it here.

crates/asm/worker/src/aux_resolver.rs
evgenyzdanovich: yeah, there's some mess. I already have it implemented (in the worker context), just have to use it here.

crates/storage/src/mmr_db/types.rs
evgenyzdanovich: great note, thanks.

crates/db/store-sled/src/lib.rs
evgenyzdanovich: yeah, thanks, iterating on it, it'll use typed tree.

crates/storage/src/mmr_db/sled.rs
evgenyzdanovich: agree, factoring out all the logic to appropriate places.

crates/storage/src/managers/asm.rs
evgenyzdanovich: agree, working on putting things where they belong.

crates/consensus-logic/src/asm_worker_context.rs
evgenyzdanovich: there's no apparent _good_ reason on that _anymore_, working on prettifying it and making sure we do the same abstractions like everywhere else.

crates/storage/src/mmr_db/sled.rs
delbonis: `MmrDbManager` would be the ultimate mid-level manager type name

crates/storage/src/mmr_db/memory.rs
```diff
@@ -0,0 +1,564 @@
...
+#[derive(Debug)]
+pub struct InMemoryMmrDb {
+    /// All MMR nodes indexed by position
+    nodes: HashMap<u64, Hash>,
```
delbonis: Can we just use a `Vec` for this?

crates/storage/src/mmr_db/types.rs
```diff
@@ -0,0 +1,110 @@
...
+/// Implementations of this trait maintain MMR data in a way that allows
+/// efficient proof generation for arbitrary leaf positions.
+///
+/// ## Design Invariants
+///
+/// - Leaves are indexed from 0 sequentially
+/// - `append_leaf` is the only way to add data (append-only)
+/// - `num_leaves()` always returns the total number of leaves added
+/// - Proofs are valid against the current `root()`
+pub trait MmrDatabase: Send + Sync {
```
delbonis: Maybe `MmrProvider`?  This name makes it seem like it's a low-level database type.

crates/db/types/src/traits.rs
```diff
@@ -57,6 +57,17 @@ pub trait AsmDatabase: Send + Sync + 'static {
...
+    /// Stores a manifest hash at the given MMR leaf index.
+    ///
+    /// This is called after appending a manifest to the MMR to maintain
+    /// a fast lookup index from leaf index to manifest hash.
+    fn store_manifest_hash(&self, index: u64, hash: [u8; 32]) -> DbResult<()>;
+
+    /// Gets a manifest hash by MMR leaf index.
+    ///
+    /// Used by aux data resolver to retrieve manifest hashes for subprotocols.
+    fn get_manifest_hash(&self, index: u64) -> DbResult<Option<[u8; 32]>>;
```
delbonis: I'm wondering if it makes sense to expose these at this level or if it would be better have the context traits in the ASM infrastructure be implemented by interacting with the MMR DB with a particular reserved MMR ID.  I was leaning towards the latter when I was writing the ticket but there might be some transactional domain / coherency reason to include this here anyways.

crates/storage/src/lib.rs
```diff
@@ -104,3 +105,46 @@ where
...
+/// with MMR database capabilities for proof generation.
+pub fn create_node_storage_with_sled(
```
delbonis: Kinda wanted this crate to avoid knowing about any particular database impl.

crates/db/store-sled/src/asm/schemas.rs
```diff
@@ -14,3 +17,49 @@ define_table_with_seek_key_codec!(
...
+    /// MMR node storage: position -> hash. Stores all MMR nodes for proof generation.
+    (AsmMmrNodeSchema) u64 => [u8; 32]
+);
+
+/// MMR metadata storage
+#[derive(Debug, Clone, borsh::BorshSerialize, borsh::BorshDeserialize)]
+pub struct MmrMetadata {
+    pub num_leaves: u64,
+    pub mmr_size: u64,
+    pub peak_roots: Vec<[u8; 32]>,
```
delbonis: Hmm, does this field make sense to store here?  Since we'd be storing regular MMR snapshots and reconstructing the head on the fly (with in-memory cache in the manager, ideally).

crates/db/store-sled/src/lib.rs
evgenyzdanovich: done, we use typed tree now.

crates/storage/src/mmr_db/sled.rs
evgenyzdanovich: removed in memory impl entirely for now (if needed we can bring it back quite easily).

crates/storage/src/lib.rs
evgenyzdanovich: yeah, done.

crates/consensus-logic/src/asm_worker_context.rs
evgenyzdanovich: done

crates/db/types/src/mmr_helpers.rs
evgenyzdanovich: agree, thanks for the hint!

however, IMO, I think the workflow is better if we first drive MMR persistence to a better state, prettify and make MMR useful in other contexts (unlike now, it's kind of part of ASM) and only then squeeze necessary common functionality to the lib. Otherwise, "premature" refactors can bring unnecessary mess.

I like Trey's vision on ID-enhanced MMRs, but the current impl needs additional work to make that vision into a reality.

crates/storage/src/mmr_db/types.rs
evgenyzdanovich: done!

crates/storage/src/managers/asm.rs
evgenyzdanovich: Agree, that's the plan, boss :) I only intend to deliver immediately necessary minimalistic logic that helps resolve auxes. That's also the reason why this current impl is "tied" to ASM in some way and is not _YET_ an independent piece of its own.

Once its submitted, THEN we cover ASM with heavy tests. After ASM is rock solid we can squeeze MMR and make it more usable for OL/account inboxes/etc (without worrying to break something on the ASM side by relying on, hopefully, rigorous tests).

benches/src/db.rs
evgenyzdanovich: the initial impl was working on the raw (untyped) sled trees that requires raw db interface. Now, it uses pretty much the same things as other existing persistence parts (`inst_ops` + trait + db + typed trees).

crates/asm/worker/src/aux_resolver.rs
evgenyzdanovich: done

crates/asm/worker/src/aux_resolver.rs
evgenyzdanovich: done!

crates/storage/src/mmr_db/memory.rs
evgenyzdanovich: I decided to remove the in-memory impl entirely since it's not clear as of now whether it's of any use. If needed, I can bring it back very quickly.

crates/storage/src/lib.rs
evgenyzdanovich: done, that was a leftover from the initial "prototyping" with claude.

crates/db/types/src/traits.rs
evgenyzdanovich: I like it, it makes quite a lot of sense actually, but IMO its better to handle it separately in another pr/task (and I'm ready to do it, just a bit later).

bin/strata-client/src/main.rs
```diff
@@ -103,7 +103,12 @@ fn main_inner(args: Args) -> anyhow::Result<()> {
 
     // Open and initialize database
     let database = init_db::init_database(&config.client.datadir, config.client.db_retry_count)?;
-    let storage = Arc::new(create_node_storage(database.clone(), pool.clone())?);
+    let asm_db_sled = database.asm_db_sled().clone();
+    let storage = Arc::new(strata_storage::create_node_storage_with_sled(
+        database.clone(),
+        asm_db_sled,
+        pool.clone(),
+    )?);
```
prajwolrg: Why is `asm_db` separated from node storage?

crates/storage/src/ops/mmr.rs
```diff
@@ -0,0 +1,15 @@
...
+use strata_db_types::traits::MmrDatabase;
+use strata_merkle::MerkleProofB32 as MerkleProof;
+
+use crate::exec::*;
+
+inst_ops_simple! {
+    (<D: MmrDatabase> => MmrDataOps) {
+        append_leaf(hash: [u8; 32]) => u64;
+        generate_proof(index: u64) => MerkleProof;
+        generate_proofs(start: u64, end: u64) => Vec<MerkleProof>;
```
delbonis: I think this should still probably be being done at a higher layer instead of making the database understand how to do it.  Since then it has to be aware of the hash function we're using for the mmr.

Also, which state index is the proof valid against?  What if we have to generate one for a slightly older state?

crates/db/store-sled/src/asm/db.rs
delbonis: Yes def, that's the thing we're actually operating on.

crates/storage/src/lib.rs
```diff
@@ -88,19 +120,65 @@ where
...
+        mmr_manager: None,
+        mmr_db: None,
+    })
+}
+
+/// Creates node storage from Sled backend with MMR database support
+///
+/// This function creates all storage managers including the MMR manager for proof generation.
+pub fn create_node_storage_with_sled(
+    db: Arc<impl DatabaseBackend + 'static>,
```
delbonis: Can probably just add `'static` to the trait bounds.

crates/storage/src/lib.rs
```diff
@@ -34,6 +34,26 @@ pub struct NodeStorage {
     // TODO maybe move this into a different one?
     // update: probably not, would require moving data around
     checkpoint_manager: Arc<CheckpointDbManager>,
+
+    mmr_manager: Option<Arc<MmrManager>>,
+
+    /// Direct access to MMR database for WorkerContext
+    mmr_db: Option<Arc<strata_db_store_sled::asm::SledMmrDb>>,
```
delbonis: Why are these `Option`?

delbonis (changes requested):

storopoli (dismissed):
ACK 6d6bedde471f6500aa66647f2a664f6e05a562ef

bin/strata-client/src/main.rs
evgenyzdanovich: done

crates/storage/src/ops/mmr.rs
evgenyzdanovich: done

storopoli (dismissed):
ACK 40108c3b39d1d21623254cfd9d70630d594671d4

crates/db/store-sled/src/asm/db.rs
evgenyzdanovich: completely agree. 

Left 2 todos for now: both to factor out mmr from here (once General MMR vision becomes reality) AND make sure that asm db stores and operates on manifests rather than logs directly.

storopoli (approved):
ACK 1cf55c1631914e9c91d297129c1c256f7cf0e70f

crates/db/store-sled/src/asm/mmr.rs
```diff
@@ -0,0 +1,507 @@
...
+    },
+    traits::MmrDatabase,
+};
+use strata_merkle::{CompactMmr64B32 as CompactMmr64, Sha256Hasher, hasher::MerkleHasher};
+use typed_sled::SledTree;
+
+use super::schemas::{AsmMmrMetaSchema, AsmMmrNodeSchema, MmrMetadata};
+
+// Hash type for 32-byte hashes
+type Hash = [u8; 32];
```
delbonis: This *might* be able to be switched to `Buf32` like we've been using elsewhere, not sure though.

crates/db/store-sled/src/asm/mmr.rs
```diff
@@ -0,0 +1,507 @@
...
+}
+
+impl Clone for SledMmrDb {
+    fn clone(&self) -> Self {
+        Self {
+            nodes_tree: self.nodes_tree.clone(),
+            metadata_tree: self.metadata_tree.clone(),
+        }
+    }
+}
```
delbonis: Why doesn't this like `#[derive(Clone)]`?

crates/db/store-sled/src/asm/mmr.rs
```diff
@@ -0,0 +1,507 @@
...
+        f.debug_struct("SledMmrDb")
+            .field("num_leaves", &metadata.as_ref().map(|m| m.num_leaves))
+            .field("mmr_size", &metadata.as_ref().map(|m| m.mmr_size))
+            .field("peak_roots", &metadata.as_ref().map(|m| &m.peak_roots))
+            .finish_non_exhaustive()
+    }
+}
+
+impl MmrDatabase for SledMmrDb {
+    fn append_leaf(&self, hash: Hash) -> DbResult<u64> {
```
delbonis: Ah wait I understand why this is as complicated as it is now, I was focused on the interfaces initially.  Is the idea with storing each node separately that we could directly reconstruct proofs for older nodes instead of having to update the proof the whole way to the state we're looking to have a proof against?  Because this *is* a faster approach actually, which I hadn't considered at all. :upside_down_face: I wonder if we should just do the addressing more directly and have an index for all nodes in all of the trees, with keys like `[ height: u32 ][ index: u64 ]`.

Or wait, *is* that something like what this is doing?  I'm not sure I understand the bookkeeping with the positions.

Also worried about race conditions here, should these all be done in the context of a transaction?  It'd probably be safe in practice since only one worker is ever expected to end up writing to the database at once but readers might see an inconsistent state in rare circumstances, like if a diagnostic/monitoring RPC call comes in at the wrong time.

evgenyzdanovich: >Also worried about race conditions here
Thanks, good note on races - I fixed it by using an-already-existing transactioning logic just like we do in other places.

> Or wait, is that something like what this is doing? I'm not sure I understand the bookkeeping with the positions.

Yes, the current impl stores all the nodes (peaks & intermediate & leafs) and it makes possible to generate any inclusion proofs against (even older) state in the future. Now it works only against the latest state, but it shouldn't be very complicated to generalize for inclusion proofs against (potentially) older state.

Yeah, the impl uses some neat tricks with bit-wise math that allows to "bookkeep" `[height][index]` as a plain index, so we never need double indexing in the internals. This'd allow us to represent the whole MMR tree in one Vec (+ some metadata), rather than map.
Having this approach allows us to store `mmr-id` -> `vec![nodes]` to supports multiple mmrs and expose nice access through `Handler` via single `mmr-id` lookup.

P.S. 
BUT, point noted, thank you! **I can see this is confusing not only to you, but Bibek also raised similar concerns before**, the indices are not very intuitive. 
I'll consider simplifying it and making it more direct and straightforward in the next iteration (as part of general MMR effort as described in the vision).

crates/db/store-sled/src/asm/mmr.rs
evgenyzdanovich: done, was a leftover

crates/db/store-sled/src/asm/mmr.rs
evgenyzdanovich: agree, done.

prajwolrg (approved):
ACK [ef37a55](https://github.com/alpenlabs/alpen/pull/1193/commits/ef37a55e1038bedf7411878a804113b3dcb7bed6).

crates/db/store-sled/src/asm/mmr.rs
storopoli: Should we accept as it is and move the simplification to a ticket and merge this (already huge) PR?

storopoli (approved):
ACK ef37a55e1038bedf7411878a804113b3dcb7bed6

crates/db/store-sled/src/asm/mmr.rs
delbonis: ok this is pretty cool, I hope we generalize this in a future PR to fulfill the needs of that jira ticket

storopoli: Before we forget. Can someone please dump all the context in a Jira ticket? @delbonis or @evgenyzdanovich?

evgenyzdanovich: that's already documented by Trey in this ticket about MMR, I don't think we need any extra

crates/db/types/src/mmr_helpers.rs
```diff
@@ -0,0 +1,331 @@
...
+/// 1 leaf  (0b1)   -> 1 peak  -> size = 2*1 - 1 = 1
+/// 2 leaves (0b10)  -> 1 peak  -> size = 2*2 - 1 = 3
+/// 3 leaves (0b11)  -> 2 peaks -> size = 2*3 - 2 = 4
+/// 7 leaves (0b111) -> 3 peaks -> size = 2*7 - 3 = 11
+/// ```
+pub fn leaf_index_to_mmr_size(index: u64) -> u64 {
+    let leaves_count = index + 1;
+    let peak_count = leaves_count.count_ones() as u64;
+    2 * leaves_count - peak_count
+}
```
barakshani, line 48: This implementation should work only for the last (most right) index. For other indices it depends on the total number of leaves on the tree/mmr.

Let's change the name to `leaves_to_mmr_size` and perhaps takes as input `leaves_count` instead?

crates/db/types/src/mmr_helpers.rs
```diff
@@ -0,0 +1,331 @@
...
+        }
+        peak_size >>= 1;
+    }
+
+    pos as u8
+}
+
+/// Calculate the offset to a node's parent
+///
+/// For a node at height h, parent is 2^(h+1) positions away
```
barakshani, line 100: this should depend on whether the node is left or right sibling. you seem to assume here left sibling, as you show on `parent_pos()` below.

barakshani (commented):
Very nice work!

PR #1194, opened 2025-12-09
Str 1920 new func tests setup

bewakes:
### Type of Change

This PR scaffolds a new functional tests framework aiming for less monkey-patching and more type safety and cleaner architecture in general.

This also adds the new tests in CI.

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## AI disclosure
This PR was created with a closely supervised help from Claude Code.

## Checklist

### DIFF
--- a/functional-tests-new/common/base_test.py
+++ b/functional-tests-new/common/base_test.py
@@ -2,7 +2,6 @@
 Base test class with common utilities.
 """
 
-import logging
 from typing import Literal, overload
 
 import flexitest
@@ -12,7 +11,6 @@ from common.services import (
     BitcoinService,
     StrataService,
 )
-from common.test_logging import get_test_logger
 
 
 class BaseTest(flexitest.Test):
@@ -34,31 +32,6 @@ class BaseTest(flexitest.Test):
         Things that need to be done before we run the test.
         """
         self.runctx = ctx
-
-    @property
-    def logger(self) -> logging.Logger:
-        """Get the current test's logger from thread-local context."""
-        return get_test_logger()
-
-    @property
-    def debug(self):
-        """Log at DEBUG level."""
-        return self.logger.debug
-
-    @property
-    def info(self):
-        """Log at INFO level."""
-        return self.logger.info
-
-    @property
-    def warning(self):
-        """Log at WARNING level."""
-        return self.logger.warning
-
-    @property
-    def error(self):
-        """Log at ERROR level."""
-        return self.logger.error
 
     # Overriding here to have `self.get_service` return a `ServiceWrapper[Rpc]` without boilerplate.
     def main(self, ctx) -> bool:  # type: ignore[override]
--- a/functional-tests-new/common/services/strata.py
+++ b/functional-tests-new/common/services/strata.py
@@ -2,7 +2,7 @@
 Strata service wrapper with Strata-specific health checks.
 """
 
-from typing import TypedDict, cast
+from typing import TypedDict
 
 from common.rpc import JsonRpcClient
 from common.services.base import RpcService
--- a/functional-tests-new/entry.py
+++ b/functional-tests-new/entry.py
@@ -23,6 +23,7 @@ from common.keepalive import KEEP_ALIVE_TEST_NAME, load_keepalive_test
 
 # Import environments
 from common.runtime import TestRuntimeWithLogging
+from common.test_logging import TestNameFilter
 from envconfigs.strata import StrataEnvConfig
 
 # Import factories
@@ -48,12 +49,14 @@ def disabled_tests() -> frozenset[str]:
 
 
 def setup_logging() -> None:
-    """Configure root logger."""
+    """Configure root logger with test name filter."""
     log_level = os.getenv("LOG_LEVEL", "INFO").upper()
     logging.basicConfig(
         level=getattr(logging, log_level, logging.INFO),
-        format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+        format="%(asctime)s - [%(test_name)s] - %(name)s - %(levelname)s - %(message)s",
     )
+    for handler in logging.root.handlers:
+        handler.addFilter(TestNameFilter())
 
 
 def parse_args(argv: list[str]) -> argparse.Namespace:
--- a/functional-tests-new/common/runtime.py
+++ b/functional-tests-new/common/runtime.py
@@ -1,21 +1,23 @@
 """
-Custom test runtime with logging context management.
+Custom test runtime with test name tracking for logging.
 """
 
 import flexitest
 
-from common.test_logging import test_logger_context
+from common.test_logging import set_current_test
 
 
 class TestRuntimeWithLogging(flexitest.TestRuntime):
     """
-    TestRuntime that sets up a thread-local logger context for each test.
+    TestRuntime that sets the current test name for automatic log tagging.
 
-    This allows library functions to call get_test_logger() from anywhere
-    without needing to pass logger instances around.
+    All logs will be tagged with the current test name via a logging filter.
     """
 
     def _exec_test(self, test_name: str, env):
-        """Wraps test execution with logger context."""
-        with test_logger_context(test_name):
+        """Wraps test execution with test name tracking."""
+        set_current_test(test_name)
+        try:
             return super()._exec_test(test_name, env)
+        finally:
+            set_current_test(None)
--- a/functional-tests-new/common/test_logging.py
+++ b/functional-tests-new/common/test_logging.py
@@ -1,45 +1,30 @@
 """
-Thread-local test logger context for use across the codebase.
+Test name injection for logging across the codebase.
 
-Provides a global logger that's automatically set per-test by the runtime,
-allowing library functions to log without tight coupling to test instances.
+Provides a logging filter that automatically tags all logs with the current test name.
 """
 
-import contextvars
 import logging
-from collections.abc import Generator
-from contextlib import contextmanager
 
-_current_test_logger: contextvars.ContextVar[logging.Logger | None] = contextvars.ContextVar(
-    "test_logger", default=None
-)
+_current_test_name: str | None = None
 
 
-def get_test_logger() -> logging.Logger:
+class TestNameFilter(logging.Filter):
     """
-    Get the current test's logger from anywhere in the codebase.
-
-    Raises RuntimeError if called outside a test context.
+    Logging filter that injects current test name into all log records.
     """
-    logger = _current_test_logger.get()
-    if logger is None:
-        raise RuntimeError("No test logger set. Are you calling this outside a test context?")
-    return logger
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        record.test_name = _current_test_name or "no-test"
+        return True
 
 
-@contextmanager
-def test_logger_context(test_name: str) -> Generator[logging.Logger, None, None]:
+def set_current_test(test_name: str | None) -> None:
     """
-    Context manager that sets up a test-specific logger for the duration of a test.
+    Set the current test name for logging.
 
-    Usage:
-        with test_logger_context("test_foo"):
-            # Now get_test_logger() will work anywhere
-            run_test()
+    This is called by the test runtime before each test execution.
+    All logs will be tagged with this test name until it's cleared.
     """
-    logger = logging.getLogger(f"test.{test_name}")
-    token = _current_test_logger.set(logger)
-    try:
-        yield logger
-    finally:
-        _current_test_logger.reset(token)
+    global _current_test_name
+    _current_test_name = test_name
--- a/functional-tests-new/common/wait.py
+++ b/functional-tests-new/common/wait.py
@@ -2,12 +2,13 @@
 Waiting utilities for test synchronization.
 """
 
+import logging
 import math
 import time
 from collections.abc import Callable
 from typing import Any, TypeVar
 
-from common.test_logging import get_test_logger
+logger = logging.getLogger(__name__)
 
 
 def wait_until(
@@ -26,7 +27,7 @@ def wait_until(
                 return
         except Exception as e:
             ety = type(e)
-            get_test_logger().warning(f"caught exception {ety}, will still wait for timeout: {e}")
+            logger.warning(f"caught exception {ety}, will still wait for timeout: {e}")
         time.sleep(step)
     raise AssertionError(error_with)
 
@@ -55,7 +56,7 @@ def wait_until_with_value(
                 return r
         except Exception as e:
             ety = type(e)
-            get_test_logger().warning(f"caught exception {ety}, will still wait for timeout: {e}")
+            logger.warning(f"caught exception {ety}, will still wait for timeout: {e}")
 
         time.sleep(step)
     raise AssertionError(error_with)
--- a/functional-tests-new/tests/test_node_version.py
+++ b/functional-tests-new/tests/test_node_version.py
@@ -1,9 +1,13 @@
 """Basic node functionality tests."""
 
+import logging
+
 import flexitest
 
 from common.base_test import StrataNodeTest
 from common.config import ServiceType
+
+logger = logging.getLogger(__name__)
 
 
 # NOTE: this is redundant and is just for setting up the func tests infra. Remove later.
@@ -21,14 +25,14 @@ class TestNodeVersion(StrataNodeTest):
         # Create RPC clients
         strata_rpc = strata.create_rpc()
 
-        self.info("Waiting for Strata RPC to be ready...")
+        logger.info("Waiting for Strata RPC to be ready...")
         strata.wait_for_rpc_ready(timeout=10)
 
         # Test protocol version
-        self.info("Checking protocol version...")
+        logger.info("Checking protocol version...")
         version = strata_rpc.strata_protocolVersion()
-        self.info(f"Protocol version: {version}")
+        logger.info(f"Protocol version: {version}")
         assert version == 1, f"Expected version 1, got {version}"
 
-        self.info("Test passed!")
+        logger.info("Test passed!")
         return True


functional-tests-new/common/base_test.py
```diff
@@ -0,0 +1,116 @@
...
+from common.config import ServiceType
+from common.rpc import JsonRpcClient
+from common.services import (
+    BitcoinServiceWrapper,
+    StrataServiceWrapper,
+)
+from common.wait import wait_until
+
+
+class BaseTest(flexitest.Test):
```
delbonis, line 16: `TestBase`?

functional-tests-new/common/base_test.py
```diff
@@ -0,0 +1,116 @@
...
+            self.wait_for(lambda: rpc.strata_protocolVersion() == 1, timeout=10)
+        """
+        wait_until(condition, error_with=error_msg, timeout=timeout, step=interval)
+
+    def wait_for_rpc_ready(
+        self,
+        rpc: JsonRpcClient,
+        method: str = "strata_protocolVersion",
+        timeout: int = 30,
+    ):
```
delbonis: These functions don't make sense to be here.

functional-tests-new/common/base_test.py
```diff
@@ -0,0 +1,116 @@
...
+            )
+        return svc
+
+    # Overriding here to have `self.get_service` return a `ServiceWrapper[Rpc]` without boilerplate.
+    def main(self, ctx) -> bool:  # type: ignore[override]
+        self.runctx = ctx
+        return self.run()
+
+    def run(self) -> bool:
+        raise NotImplementedError
```
delbonis: This is weird, it's re-doing what `premain` is designed for doing.

functional-tests-new/common/base_test.py
```diff
@@ -0,0 +1,116 @@
...
+    def premain(self, ctx: flexitest.RunContext):
+        """
+        Set up logging for the test.
+        This is the ONLY thing premain does - no service setup.
+        """
+        self.logger = logging.getLogger(f"test.{self.__class__.__name__}")
+        self.debug = self.logger.debug
+        self.info = self.logger.info
+        self.warning = self.logger.warning
+        self.error = self.logger.error
```
delbonis: I really want to move away from this pattern for logging.  It seemed convenient at the time but it breaks down when using library functions.  The test runtime class should set a global/thread-local logger just before running a test and unset it just after exiting (which I think it does now? with a context manager).

functional-tests-new/common/base_test.py
delbonis, line 1: It might be good to avoid making this assume that there's particular services always available.  It would be nice to reuse components of this for the mosaic client tests or if we support a no-L1 consensus mode then it'd be awkward to extricate it.  It makes sense to have a base test with universal stuff, but then maybe it'd make sense to have more test subclasses that specialize like this does.

functional-tests-new/common/config/params.py
```diff
@@ -0,0 +1,124 @@
...
+        ValueError: If n is not in valid byte range (0-255)
+    """
+    if not 0 <= n < 256:
+        raise ValueError(f"Byte value must be in range 0-255, got: {n}")
+    return bytes([n] * repeat).hex()
+
+
+@dataclass
+class L1BlockCommitment:
+    height: int = field(default=100)
```
delbonis, line 32: 100?

functional-tests-new/common/config/params.py
delbonis, line 1: This mod is going to be a *huge* pain to keep in sync with the params types starting in a couple of weeks when we restructure how params are defined across the different parts of the architecture.  Is it really worth being this verbosely type-safe when we can just have a fixed params file and override certain fields on an as-needed basis?

functional-tests-new/common/config/params.py
```diff
@@ -0,0 +1,124 @@
...
+class ProofPublishModeTimeout:
+    timeout: int = field(default=30)
+
+
+ProofPublishMode = Literal["strict"] | ProofPublishModeTimeout
+
+
+@dataclass
+class Checked:
+    schnorr_key: str  # TODO: more sophisticated, using pydantic?
```
delbonis, line 73: Calling this `Checked` does not at all explain what it's for.  I can infer it from context but others would probably have a hard time.

functional-tests-new/common/services/base.py
```diff
@@ -0,0 +1,120 @@
...
+from typing import Any, Generic, TypeVar
+
+import flexitest
+
+from common.wait import wait_until
+
+Rpc = TypeVar("Rpc")
+
+
+class ServiceWrapper(flexitest.service.ProcService, Generic[Rpc]):
```
delbonis: This feels like a *lot* of ceremony for something that's not actually providing a lot of value.  We can improve the `ProcService` type directly if it's needed.

What it's doing here with the status checks does provide value, but it might make more sense to be implemented as a mixin instead of coupling it to `ProcService` so that it can't be used with another type of service (like one launched as a docker container, which isn't implemented yet but was discussed).

However, the code to wait for it to become ready should be done as part of the env control logic, do we need to integrate this functionality into the service itself if we could rely on the env controller (`LiveEnv`) to manage its lifecycle?

functional-tests-new/common/services/bitcoin.py
```diff
@@ -0,0 +1,29 @@
...
+        if not self.check_status():
+            return False
+
+        try:
+            rpc = self.create_rpc()
+            # Try calling a simple RPC method
+            rpc.proxy.getblockchaininfo()
+            return True
+        except Exception:
+            return False
```
delbonis: I'm might be wrong but I'm pretty sure this is a perfect candidate to be a mixin.

This is a bit of an example of the kind of over-taxonomizing is a very common and valid critique of the object-oriented approach to software design that should be carefully watched for and avoided.  If there's something subtle that changes about the concepts modelled by the class hierarchy it can force you to completely rethink *everything*.

functional-tests-new/common/services/props.py
delbonis, line 1: More on this, this gets converted to a dict in the `ServiceWrapper` anyways.  What value is this really providing by being rigorously typed?  On the consumer side (in the test cases) there's enough indirection that mypy probably couldn't figure it out anyway, could it?

functional-tests-new/common/services/strata.py
delbonis, line 1: See same comment on the bitcoin service wrapper class.

functional-tests-new/entry.py
```diff
@@ -0,0 +1,297 @@
...
+
+    # Replace environment placeholder
+    test_code = test_code.replace("{ENV}", env_name)
+
+    # Load as module
+    module = types.ModuleType("__keepalive_dynamic_test__")
+    exec(test_code, module.__dict__)
+    sys.modules["__keepalive_dynamic_test__"] = module
+
+    return getattr(module, KEEP_ALIVE_TEST_NAME)
```
delbonis: Some of this keepalive stuff could be moved into another module to keep this one cleaner.

functional-tests-new/envconfigs/strata.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Basic environment: Bitcoin regtest + Strata sequencer.
+    """
+
+    def __init__(self, pre_generate_blocks: int = 0):
+        self.pre_generate_blocks = pre_generate_blocks
+
+    def init(self, ectx: flexitest.EnvContext) -> flexitest.LiveEnv:
+        btc_factory = cast(BitcoinFactory, ectx.get_factory(ServiceType.Bitcoin))
+        strata_factory = cast(StrataFactory, ectx.get_factory(ServiceType.Strata))
```
delbonis, line 22: This is rather verbose, can this be improved by adding a wrapper around `get_factory`?

functional-tests-new/common/config/params.py
bewakes: Renamed to `SchnorrVerify`.

functional-tests-new/common/config/params.py
bewakes: Actually I bit the bullet here understanding this might be a bit of a work until we finalize how the params are going to be.

functional-tests-new/common/services/base.py
bewakes: >  We can improve the ProcService type directly if it's needed.

I think that might be helpful. 

This "ceremony" is for a handful of  services that we define(reth, alpen-ee, prover, bitcoin). Rest of the code can benefit from this. The other values besides type safety this provides are:
1. We don't have to monkey patch service like `svc.create_rpc = _create_rpc` we did.
2. LSP Autocomplete. This _really_ is useful.
3. No more squiggly red lines in IDE. We do a couple of unavoidable `# ignore`s in the core library and use the interface downstream.

> do we need to integrate this functionality into the service itself if we could rely on the env controller (LiveEnv) to manage its lifecycle?

The `BasicEnvConfig` calls `wait_for_ready` during initialization. The service should define the lifecycle methods like `wait_for_ready` and those can be called from env or test cases as needed.

functional-tests-new/common/services/bitcoin.py
bewakes: Not sure if this is a candidate for mixin - note `rpc.proxy.getblockchaininfo()` and the mixin has to rely on existence of `create_rpc()` and `check_status()` methods. I would rather abstract out the parts except `rpc.proxy.getblockchaininfo()` in the base class.

functional-tests-new/common/config/params.py
bewakes: Trying to have a sensible default :)

functional-tests-new/common/base_test.py
bewakes: I don't know. Since this is not a test itself, I am more inclined towards this being `BaseTest` and the actual tests being `Test*`.

functional-tests-new/common/base_test.py
bewakes: Agree, I think it makes more sense to have them in the services themselves.

functional-tests-new/common/services/props.py
bewakes: You are right, this doesn't provide any significant value. I overdid this. Removed.

functional-tests-new/envconfigs/strata.py
bewakes: Not sure how we would do that and how that would make it less verbose. I think this might fine since we will only have about 5 factories.

functional-tests-new/common/base_test.py
bewakes: This is to be able to call `self.get_service(ServiceType) -> [Bitcoin|Strata]ServiceWrapper`. We could do that in main, but then the passed `ctx` would be unused.

functional-tests-new/common/base_test.py
bewakes: Yeah, that's a good point. I've made the base test thinner and added another class for strata.

functional-tests-new/common/base_test.py
delbonis: ah yeah that makes sense, I had the conventions the other way around in my head

functional-tests-new/envconfigs/strata.py
delbonis: That's kinda a lot of factories and this is a pretty repetitive pattern.  Is there some typevar thing here that could change it?

.github/workflows/functional-new.yml
```diff
@@ -0,0 +1,182 @@
...
+      - name: Cleanup Space
+        uses: ./.github/actions/cleanup # zizmor: ignore[unpinned-uses]
+
+      - name: Install bitcoind
+        env:
+          BITCOIND_VERSION: "29.0"
+          BITCOIND_ARCH: "x86_64-linux-gnu"
+        run: |
+          curl -fsSLO --proto "=https" --tlsv1.2 "https://bitcoincore.org/bin/bitcoin-core-$BITCOIND_VERSION/bitcoin-$BITCOIND_VERSION-$BITCOIND_ARCH.tar.gz"
+          curl -fsSLO --proto "=https" --tlsv1.2 "https://bitcoincore.org/bin/bitcoin-core-$BITCOIND_VERSION/SHA256SUMS"
```
storopoli: This will probably be done in #1200

```suggestion
          BITCOIND_VERSION: "30.0"
          BITCOIND_ARCH: "x86_64-linux-gnu"
        run: |
          curl -fsSLO --proto "=https" --tlsv1.3 "https://bitcoincore.org/bin/bitcoin-core-$BITCOIND_VERSION/bitcoin-$BITCOIND_VERSION-$BITCOIND_ARCH.tar.gz"
          curl -fsSLO --proto "=https" --tlsv1.3 "https://bitcoincore.org/bin/bitcoin-core-$BITCOIND_VERSION/SHA256SUMS"
```

functional-tests-new/common/base_test.py
delbonis: This is sorta a similar case of using the test itself to have convenience functions that are actually making calls into the context.  It's spooky action at a distance.  It would be better to structure it as `ctx.get_service(FooService)`.  Or make a specialized context class.

functional-tests-new/common/config/params.py
delbonis: But that's not something that exists for a type like this...?

functional-tests-new/common/config/params.py
delbonis: Well yeah that's what I'm saying, this is all going to be rewritten soon.

functional-tests-new/common/services/bitcoin.py
delbonis: That's what I'm suggesting, you make the common logic generic.  This would be naturally extended to apply to the `wait_for_ready`/etc logic that depends on this as well, actually.

But actually now that I'm thinking about it, what's the difference here between `check_health` and `check_status`?  It feels like we'd actually want to be overriding/extending the `check_status` fn?  I'm a bit confused on the impl choices here.

functional-tests-new/README.md
```diff
@@ -0,0 +1,286 @@
...
+Service connection info will be printed on startup.
+
+### Disabling Tests
+
+To temporarily disable flaky or WIP tests, add them to `disabled_tests()` in `entry.py`:
+
+```python
+def disabled_tests() -> frozenset[str]:
+    return frozenset(["test_flaky", "test_wip"])
+```
```
delbonis, line 250: Could this be done with an envvar to avoid editing the test harness and maybe accidentally committing those changes?

functional-tests-new/common/rpc.py
```diff
@@ -0,0 +1,117 @@
...
+        except json.JSONDecodeError as e:
+            self.logger.warning(f"Invalid JSON response: {resp.text}")
+            raise RpcError({"code": -1, "message": f"Invalid JSON: {e}"}) from e
+
+        if "error" in result:
+            error = result.get("error", {})
+            self.logger.warning(f"RPC error: {error}")
+            raise RpcError(error)
+
+        return result.get("result")
```
delbonis: This is being overly careful.  Might be better to structure this as:

```py
if "error" in response:
  error = response["error"]
  ...
elif "result" in response:
  result = response["result"]
  ...
else:
  raised RpcError("malformed response")
```

functional-tests-new/common/services/base.py
delbonis: Ok I'm seeing the structure of this a bit better now.

One more specific comment on this: This RPC factory arg seems like it'd be more ergonomic to expose by having child classes just override a class method.

```py
rpc_factory: Callable[[], Rpc] | None = None,
```

I get that this trying to make it more compositional and it's *probably* safe to assume that services always have RPCs we can call on them.  I believe the idea behind this was that we wouldn't have to necessarily make a subclass for the different types of services.  But if we're going to be making subclasses *anyways*, then does it really make sense to do it this way?  Not sure about this.

I believe if we leaned into that it might make it easier to use mixins to more cleanly add things to services like waiting for ready.

functional-tests-new/common/rpc.py
bewakes: Yeah, that makes sense.

functional-tests-new/README.md
bewakes: Yeah, that's a good point. Updated.

functional-tests-new/common/services/bitcoin.py
bewakes: Done.

functional-tests-new/envconfigs/strata.py
```diff
@@ -0,0 +1,56 @@
...
+import flexitest
+
+from common.config import BitcoindConfig, ServiceType
+from factories.bitcoin import BitcoinFactory
+from factories.strata import StrataFactory
+
+
+class StrataEnvConfig(flexitest.EnvConfig):
+    """
+    Straa environment: Initializes services required to run strata.
```
delbonis: Just noticed this: `Straa`

functional-tests-new/common/base_test.py
```diff
@@ -0,0 +1,91 @@
...
+        """Log at ERROR level."""
+        return self.logger.error
+
+    # Overriding here to have `self.get_service` return a `ServiceWrapper[Rpc]` without boilerplate.
+    def main(self, ctx) -> bool:  # type: ignore[override]
+        self.runctx = ctx
+        return self.run()
+
+    def run(self) -> bool:
+        raise NotImplementedError
```
delbonis, line 38: I think we can simplify this now can't we?  Since `premain` is assigning this field itself and this just repeats that and calls `run`.

functional-tests-new/common/base_test.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @property
+    def warning(self):
+        """Log at WARNING level."""
+        return self.logger.warning
+
+    @property
+    def error(self):
+        """Log at ERROR level."""
+        return self.logger.error
```
delbonis: Should still remove these.

delbonis (changes requested):
This is looking *pretty good* now, still a handful of comments.

@storopoli Do you have any feedback on those?

functional-tests-new/common/base_test.py
bewakes: Isn't it fine to have these helpers for base test now that the logger is the global logger within the context manager? See `get_test_logger()` in `logger` property.

delbonis: What purpose do they serve when they just call the managed global logger through a roundabout mechanism?

It's much clearer to understand what's happening when you read `logging.info()` than have to know that `self.info()` actually calls through multiple layers to get a contextvar to retrieve what is also set as the global logger anyways.  Calling `logging.info()` in some places and `self.info()` in others suggests that there's different behavior happening in the different cases, but there's not, and there shouldn't be.

delbonis: What I was saying in [this comment](https://github.com/alpenlabs/alpen/pull/1194#discussion_r2604358527) wasn't that I wanted to get rid of the `self.info = foo.info` monkeypatching and move to explicit declarations in the class, it's what I wanted all the tests to be able to just use `logging.info`.

bewakes: >  it's what I wanted all the tests to be able to just use logging.info

Okay, I am confused now. What's wrong with each modules doing `logger = logging.getLogger(__name__)` and then do `logger.info()` while the global setup is done at entry file?

Isn't that the best practice?

delbonis: When a test calls into `utils.wait_for_something` that we want the logs emitted there to be tracked as being associated with the test that is currently running.  It's not just about the modules that emitted the logs, that part is straightforward.

But this isn't even what I was referring to the initial comment on this subthread.  *I think* you already roughly did what I am describing, but you're still doing the setup for this redundant `self.info()`-style logging pattern here when it's made irrelevant with the logging context manager thing wraps the test run calls.

delbonis (approved):
Still has the issue with the weird casting on the service types in the env setup but that's relatively minor and can ignore that for now.

PR #1198, opened 2025-12-10
Str 1918 guest code ee

purusang:
## Description
This pr adds guest code for Alpen ee.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

##  AI usage declaration
I have used Clade extensively throughout the work and reviewed it myself before the generated code was included in the codebase.

## Checklist

## Related Issues
STR-1918

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14426,7 +14426,7 @@ dependencies = [
  "strata-simple-ee",
  "strata-snark-acct-types",
  "thiserror 2.0.17",
- "tree_hash 0.11.0",
+ "tree_hash 0.12.1",
 ]
 
 [[package]]
@@ -14926,7 +14926,7 @@ dependencies = [
  "sha2",
  "ssz",
  "ssz_derive",
- "strata-codec",
+ "strata-codec 0.1.0 (git+https://github.com/alpenlabs/strata-common?tag=v0.1.0-alpha-rc2)",
  "strata-ee-acct-runtime",
  "strata-ee-acct-types",
  "strata-ee-chain-types",
@@ -14934,7 +14934,7 @@ dependencies = [
  "strata-identifiers",
  "strata-snark-acct-types",
  "thiserror 2.0.17",
- "tree_hash 0.11.0",
+ "tree_hash 0.12.1",
  "zkaleido",
  "zkaleido-native-adapter",
 ]


crates/ee-acct-runtime/src/update_builder.rs
```diff
@@ -90,6 +90,7 @@ impl UpdateBuilder {
 
         // Accumulate outputs from all blocks in the segment
         for block in segment.blocks() {
+            // REVIEW: why `block_notpackage`?
```
storopoli: ?

crates/proof-impl/alpen-ee-acct/Readme.md
storopoli, line 1: don't name the file `Readme.md` please use `README.md` (all caps)

crates/proof-impl/alpen-ee-acct/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+borsh.workspace = true
+ssz.workspace = true
+strata-codec.workspace = true
+
+# Hashing (for guest block verification and state hashing)
+sha2.workspace = true
+tree_hash.workspace = true
+
+# Error handling
+thiserror.workspace = true
```
storopoli, line 24: I really prefer for simplicity and standardness amongst other crates:

```
FIRST BLOCK OF DEPS: `strata-*`, `alpen-*`, `ssz-*`, `tree_hash-*`. In other words, everything that is a git repo under `alpenlabs/` GitHub org.

SECOND BLOCK OF DEPS: Anything that is either in crates.io or a git repo outside of alpenlabs/ GitHub org.

File: crates/proof-impl/eth-ee-acct/src/lib.rs [hunk 3]
```diff
@@ -0,0 +1,170 @@
...
+use crate::guest_builder::build_commit_segments_from_blocks;
+
+// Borsh serialization implementation
+mod borsh_impl;
+
+// Guest-side block building
+mod guest_builder;
+
+// ZkVmProgram implementation
+pub mod program;
```
storopoli: Should this be `pub mod`? You are already doing a bunch of `pub use program::{Foo}` below.

crates/proof-impl/eth-ee-acct/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+#[derive(Clone, Debug)]
+pub struct EthEeAcctProofOutput {
+    /// Previous account state before the update
+    pub prev_state: ProofState,
+
+    /// New account state after the update
+    pub final_state: ProofState,
+
+    /// DA commitments (currently empty, will be populated by outer proof)
+    pub da_commitments: Vec<[u8; 32]>,
```
storopoli: Do we have a type alias for `[u8; 32]` or even `Vec<[u8; 32]>`. For example we adopted `Hash` in #1207. Do we have something planned for "DA types"?

Cc @irnb

crates/proof-impl/eth-ee-acct/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    /// Output messages generated during update
+    pub output_messages: Vec<OutputMessage>,
+
+    /// Input messages processed during update
+    pub input_messages: Vec<MessageEntry>,
+
+    /// Update metadata (new tip, processed counts)
+    pub extra_data: UpdateExtraData,
```
storopoli: Do we need all these fields to be `pub`? I'd really like to restrict the API in critical types such as prover/proving types.
Can we get accessors to all of them and mutable accessors for the ones that we need mutable access?

crates/proof-impl/eth-ee-acct/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+    // Build CommitChainSegment from blocks
+    let commit_segments = build_commit_segments_from_blocks(serialized_blocks)
+        .expect("Failed to build commit segments from blocks");
+
+    // Already raw bytes
+    let raw_prev_header = zkvm.read_buf();
+    let raw_partial_pre_state = zkvm.read_buf();
+
+    // Read genesis data for ChainSpec construction
+    let genesis: rsp_primitives::genesis::Genesis = zkvm.read_serde();
```
storopoli: Can you import `Genesis` in the top with `use` statements?

crates/proof-impl/eth-ee-acct/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    // Extract UpdateExtraData from the operation (already verified by runtime)
+    let extra_data = decode_buf_exact::<UpdateExtraData>(operation.extra_data())
+        .expect("Failed to decode UpdateExtraData");
+
+    // Build the complete public output
+    let proof_output = EthEeAcctProofOutput {
+        prev_state: prev_proof_state,
+        final_state: final_proof_state,
+        da_commitments: Vec::new(), // Empty for now, TODO: when do we actually fill it??
```
storopoli: @irnb or @prajwolrg can you guys comment on this TODO?

crates/proof-impl/eth-ee-acct/src/program.rs
```diff
@@ -0,0 +1,129 @@
...
+    {
+        let mut input_builder = B::new();
+
+        // Write SSZ-encoded data as raw buffers
+        input_builder.write_buf(&input.astate_ssz)?;
+        input_builder.write_buf(&input.operation_ssz)?;
+        input_builder.write_buf(&input.prev_proof_state_ssz)?;
+
+        // Write coinputs (Vec<Vec<u8>>) with borsh
+        input_builder.write_borsh(&input.coinputs)?;
```
storopoli: We are trying to migrate away from `Borsh` in zkVM environments.
I have to confess that I need to catch up and paint a general picture of how much Borsh (de)serialization we're using in zkVMs.
But why can't this be SSZ?
Cc @delbonis

crates/proof-impl/eth-ee-acct/src/program.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /// Execute the program with native host (for testing)
+    pub fn execute(
+        input: &<Self as ZkVmProgram>::Input,
+    ) -> ZkVmResult<<Self as ZkVmProgram>::Output> {
+        let host = Self::native_host();
+        <Self as ZkVmProgram>::execute(input, &host)
+    }
+}
```
storopoli: should this be feature-gated to `[cfg(tests)]` or maybe `[cfg(feature = "test-utils")]`?

crates/proof-impl/eth-ee-acct/src/lib.rs
storopoli: Why are we using `serde` inside a zkVM?
Cc @delbonis

crates/proof-impl/eth-ee-acct/src/program.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub serialized_blocks: Vec<Vec<u8>>,
+
+    /// Previous header (raw bytes)
+    pub raw_prev_header: Vec<u8>,
+
+    /// Partial pre-state (raw bytes)
+    pub raw_partial_pre_state: Vec<u8>,
+
+    /// Genesis data for constructing ChainSpec (serde-serialized)
+    pub genesis: rsp_primitives::genesis::Genesis,
```
storopoli: Why do all of these fields need to be `pub`? Same as the comment above.

storopoli: import `Genesis` with `use`.

crates/proof-impl/eth-ee-acct/src/guest_builder.rs
```diff
@@ -0,0 +1,98 @@
...
+//! - Building CommitBlockData with verified commitments
+//! - Assembling CommitChainSegment from multiple blocks
+
+use sha2::{Digest, Sha256};
+use ssz::{Decode, Encode};
+use strata_ee_acct_types::{CommitBlockData, CommitChainSegment};
+use strata_ee_chain_types::ExecBlockPackage;
+use strata_primitives::buf::Buf32;
+
+#[derive(Debug, thiserror::Error)]
```
storopoli: import `thiserror::Error` with `use`

crates/proof-impl/alpen-ee-acct/src/guest_builder.rs
```diff
@@ -0,0 +1,98 @@
...
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum GuestBuilderError {
+    #[error("SSZ decode error")]
+    SszDecode,
+    #[expect(dead_code, reason = "Reserved for future codec decoding errors")]
+    #[error("Codec decode error: {0}")]
+    CodecDecode(String),
+    #[error("Block hash mismatch: expected {expected:?}, computed {computed:?}")]
+    HashMismatch { expected: Buf32, computed: Buf32 },
+}
```
storopoli, line 15: Add docstrings. Also having `error` in `#[error]` macros are kinda redundant in error messages.

crates/proof-impl/eth-ee-acct/src/guest_builder.rs
```diff
@@ -0,0 +1,98 @@
...
+    #[error("SSZ decode error")]
+    SszDecode,
+    #[expect(dead_code, reason = "Reserved for future codec decoding errors")]
+    #[error("Codec decode error: {0}")]
+    CodecDecode(String),
+    #[error("Block hash mismatch: expected {expected:?}, computed {computed:?}")]
+    HashMismatch { expected: Buf32, computed: Buf32 },
+}
+
+pub(crate) type Result<T> = std::result::Result<T, GuestBuilderError>;
```
storopoli: import `std::result::Result`

crates/proof-impl/eth-ee-acct/src/guest_builder.rs
```diff
@@ -0,0 +1,98 @@
...
+
+/// Deserialize and build CommitBlockData from host-provided bytes (GUEST-SIDE)
+///
+/// The bytes format is: [exec_block_package (SSZ)][raw_block_body (strata_codec)]
```
storopoli:
```suggestion
/// The bytes format is: `[exec_block_package (SSZ)][raw_block_body (strata_codec)]`
```

crates/proof-impl/eth-ee-acct/src/guest_builder.rs
```diff
@@ -0,0 +1,98 @@
...
+/// 1. Deserializes ExecBlockPackage from SSZ
+/// 2. Calculates where the package ends
+/// 3. Extracts the raw block bytes
+/// 4. Verifies the hash of raw block bytes matches the commitment
+/// 5. Returns CommitBlockData
+///
+/// Note: We only verify the raw_block_encoded_hash (the cryptographic commitment).
+///
+/// # Arguments
+/// * `bytes` - Concatenated bytes from host: [exec_block_package][raw_block_body]
```
storopoli:
```suggestion
/// * `bytes` - Concatenated bytes from host: `[exec_block_package][raw_block_body]`
```

crates/proof-impl/eth-ee-acct/src/guest_builder.rs
```diff
@@ -0,0 +1,98 @@
...
+///
+/// # Returns
+/// CommitBlockData ready to be added to CommitChainSegment
```
storopoli:
```suggestion
/// [`CommitBlockData`] ready to be added to [`CommitChainSegment`].
```

crates/proof-impl/alpen-ee-acct/src/guest_builder.rs
```diff
@@ -0,0 +1,98 @@
...
+    // Decode ExecBlockPackage from SSZ
+    let package =
+        ExecBlockPackage::from_ssz_bytes(bytes).map_err(|_| GuestBuilderError::SszDecode)?;
+
+    // Calculate how many bytes the SSZ encoding used
+    let package_ssz = package.as_ssz_bytes();
+    let package_len = package_ssz.len();
+
+    // The rest is the raw block body
+    let raw_block = &bytes[package_len..];
```
storopoli: Add a check to make sure that `bytes` is at least `package_len` length

crates/proof-impl/alpen-ee-acct/src/guest_builder.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    // Verify the cryptographic commitment: hash of raw block bytes
+    let raw_block_hash = Buf32::from(<[u8; 32]>::from(Sha256::digest(raw_block)));
```
storopoli: You should try to use `Hash::from` from #1207 makes more sense here.

crates/proof-impl/eth-ee-acct/src/guest_builder.rs
```diff
@@ -0,0 +1,98 @@
...
+    Ok(CommitBlockData::new(package, raw_block.to_vec()))
+}
+
+/// Build CommitChainSegment(s) from host-provided block data (GUEST-SIDE)
+///
+/// This is the main entry point for processing blocks from the host.
+///
+/// # Arguments
+/// * `block_data_bytes` - Vec of serialized blocks, each in format
+///   [exec_block_package][raw_block_body]
```
storopoli:
```suggestion
///   `[exec_block_package][raw_block_body]`
```

crates/proof-impl/alpen-ee-acct/src/guest_builder.rs
```diff
@@ -0,0 +1,98 @@
...
+/// Build CommitChainSegment(s) from host-provided block data (GUEST-SIDE)
+///
+/// This is the main entry point for processing blocks from the host.
+///
+/// # Arguments
+/// * `block_data_bytes` - Vec of serialized blocks, each in format
+///   [exec_block_package][raw_block_body]
+///
+/// # Returns
+/// Vec<CommitChainSegment> - Currently returns one segment containing all blocks
```
storopoli:
```suggestion
/// [`Vec<CommitChainSegment>`] - Currently returns one segment containing all blocks
```

crates/proof-impl/eth-ee-acct/src/guest_builder.rs
```diff
@@ -0,0 +1,98 @@
...
+    for bytes in block_data_bytes {
+        let block_data = deserialize_and_build_block_data(&bytes)?;
+        block_data_list.push(block_data);
+    }
+
+    // Build one CommitChainSegment from all blocks
+    let segment = CommitChainSegment::new(block_data_list);
+
+    // Return as Vec for compatibility with existing code
+    Ok(vec![segment])
```
storopoli:
```suggestion
    // Deserialize and verify each block
    let block_data_list = block_data_bytes
        .into_iter()
        .map(|bytes| deserialize_and_build_block_data(&bytes))
        .collect::<Result<Vec<_>, _>>()?;
    
    // Build one CommitChainSegment from all blocks
    let segment = CommitChainSegment::new(block_data_list);

    // Return as Vec for compatibility with existing code
    Ok(vec![segment])
```

or if you make https://github.com/alpenlabs/alpen/blob/00a66b0daef2a2b511357fb4e434baba6b23513a/crates/ee-acct-types/src/commit.rs#L14-L17

take an `impl Iterator<Item = CommitBlockData>` you can do all in one big FP iterator call without having to partially allocate `Vec`s:


```suggestion
    // Deserialize and verify each block
    let block_data_list = block_data_bytes
        .into_iter()
        .map(|bytes| deserialize_and_build_block_data(&bytes)?);

    // Build one CommitChainSegment from all blocks
    let segment = CommitChainSegment::new(block_data_list);

    // Return as Vec for compatibility with existing code
    Ok(vec![segment])
```

crates/proof-impl/eth-ee-acct/src/borsh_impl.rs
```diff
@@ -0,0 +1,267 @@
+//! Borsh serialization implementation for EthEeAcctProofOutput
+//!
+//! This module provides Borsh serialization by wrapping SSZ-encoded bytes.
+//!
+//! Current approach: Each field is serialized to SSZ individually, then wrapped in Borsh.
+//!
+//! Once all inner types have proper SSZ support, this can be simplified to:
+//! - Derive SSZ Encode/Decode for the entire EthEeAcctProofOutput struct
+//! - Serialize the whole struct to SSZ bytes in one call
+//! - Wrap those single SSZ bytes in Borsh (much simpler than current per-field approach)
```
storopoli: We are trying to avoid Borsh in anything that touches zkVM and use SSZ.
Should we have this module?
Cc @delbonis

crates/ee-acct-types/src/errors.rs
```diff
@@ -77,6 +77,9 @@ pub enum EnvError {
     /// There was a tx that was invalid in a block, for some reason.
     #[error("invalid tx in a block")]
     InvalidBlockTx,
+
+    #[error("proof state mismatch")]
+    MismatchProofState,
```
storopoli, line 82: docstrings

crates/ee-acct-runtime/Cargo.toml
```diff
@@ -18,6 +18,7 @@ strata-snark-acct-types.workspace = true
 digest.workspace = true
 sha2.workspace = true
 thiserror = { workspace = true, optional = true }
+tree_hash.workspace = true
```
storopoli, line 23: this goes into the `strata-*` block above because it is a git repo under `alpenlabs` org.

crates/ee-acct-runtime/src/update_processing.rs
```diff
@@ -318,10 +319,17 @@ pub fn apply_final_update_changes(
...
 fn verify_acct_state_matches(
-    _astate: &EeAccountState,
-    _exp_new_state: &Hash,
+    astate: &EeAccountState,
+    exp_new_state: &Hash,
 ) -> Result<(), EnvError> {
-    // TODO use SSZ hash_tree_root
+    // Compute SSZ tree_hash_root
+    // REVIEW: should we use Sha256Hasher or Sha256MerkleHasher?
+    let computed_root = TreeHash::<Sha256Hasher>::tree_hash_root(astate);
```
storopoli, line 399: I think you can simplify it since `Sha256Hasher` is the default in `ssz-gen`'s `tree_hash` crate:

```suggestion
    let computed_root = astate.tree_hash_root()
```

The `Sha256MerkleHasher` is used internally by `TreeHash` implementations when they need to merkleize collections (lists, vectors, containers with multiple fields). You never pass it as a type parameter to `TreeHash`.

crates/ee-acct-runtime/src/update_builder.rs
```diff
@@ -219,3 +222,29 @@ impl UpdateBuilder {
...
+fn compute_new_proof_state(
+    state: &mut EeAccountState,
+    new_tip_blkid: Hash,
+    processed_messages_count: usize,
+) -> ProofState {
+    // Update last_exec_blkid before computing hash (matches apply_final_update_changes)
+    state.set_last_exec_blkid(new_tip_blkid);
+
+    // Compute the inner_state_root using tree_hash_root
+    let computed_hash = tree_hash::TreeHash::<tree_hash::Sha256Hasher>::tree_hash_root(state);
```
storopoli, line 241: Same as above:

```suggestion
    let computed_root = astate.tree_hash_root()
```

storopoli (changes requested):
Nice job on the documentation and docstrings!

crates/proof-impl/eth-ee-acct/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+mod guest_builder;
+
+// ZkVmProgram implementation
+pub mod program;
+
+pub use program::{EthEeAcctInput, EthEeAcctOutput, EthEeAcctProgram};
+
+/// Public output committed by the ETH-EE account update proof
+#[derive(Clone, Debug)]
+pub struct EthEeAcctProofOutput {
```
delbonis: This type MUST NOT exist.  We MUST be using the upstream [proof interface data structure from snark-acct-types](https://github.com/alpenlabs/alpen/blob/main/crates/snark-acct-types/src/proof_interface.rs#L8).  I realize this does not have any serialization defined, you can convert it to SSZ, which was the original design intention.

Making a borsh impl for this type as the public interface for this proof makes this non-compliant with the snark account spec.

crates/proof-impl/eth-ee-acct/src/program.rs
```diff
@@ -0,0 +1,129 @@
...
+    /// Previous ProofState (before the update) encoded as SSZ bytes
+    /// The guest will verify that tree_hash_root(astate) matches this state
+    pub prev_proof_state_ssz: Vec<u8>,
+
+    /// Coinput witness data for messages
+    pub coinputs: Vec<Vec<u8>>,
+
+    /// Serialized blocks for building CommitChainSegment in guest
+    /// Each `Vec<u8>` is: [exec_block_package (SSZ)][raw_block_body (strata_codec)]
+    pub serialized_blocks: Vec<Vec<u8>>,
```
delbonis: This is very bad ad-hoc serialization.  You should define a type that describes this structure.

This whole `EthEeAcctInput` should *probably* be `Codec`, but strictly speaking it's application-specific so it's not subject to the API stability concerns that other things are.

crates/proof-impl/eth-ee-acct/src/program.rs
```diff
@@ -0,0 +1,129 @@
...
+
+use zkaleido::{ProofType, PublicValues, ZkVmError, ZkVmInputResult, ZkVmProgram, ZkVmResult};
+use zkaleido_native_adapter::{NativeHost, NativeMachine};
+
+use crate::{EthEeAcctProofOutput, process_eth_ee_acct_update}; // From lib.rs
+
+/// Input for ETH-EE account proof generation
+/// This is the high-level input that the host provides
+#[derive(Debug)]
+pub struct EthEeAcctInput {
```
delbonis: This name implies that it's an input to the EE account, but that's confusing because that's also an accurate description of OL account messages.

*Really* what this type describes are the private inputs to the snark account proof for an EVM EE (runtime) account.

Also, there should be more structure to these fields.  Ie, separate fields that are initializing the EE account, fields we plug into the EE runtime directly to process the update, fields we init the EVM executor with, and whatever else.

delbonis (changes requested):
Also, using "eth" here is misleading as it suggests we're proving Ethereum.  We're *really* proving our specialized EVM chain's execution (really "Alpen").  I would suggest converting names that use "eth" everywhere here to "evm" or "alp"/"alpen".

crates/proof-impl/eth-ee-acct/src/program.rs
storopoli: Just curios: why `Codec`? Just want to understand the reason why?

I have in my mind the following rule of thumb: anything that touches DA needs to be be smallest size possible for Bitcoin DA costs, hence `Codec`; and anything that touches zkVMs needs to have the least amount of cycles needed for proving, hence SSZ which optimizes for proving costs and speed.

Is that the right mindset?

crates/ee-acct-runtime/src/update_builder.rs
purusang: I mean why the var name is like that?

crates/proof-impl/eth-ee-acct/src/program.rs
purusang: Had a discussion with @MdTeach , and summary was: having it feat-gated won't have much improvements as compared to a subtle lift to manage functional tests and prover-perf (proof cycle counter) while also making it work with other zkvmPrograms.

crates/proof-impl/eth-ee-acct/src/lib.rs
purusang: I am trying to understand why not and in that case what would you suggest?

crates/ee-acct-runtime/src/update_processing.rs
purusang: It does not compile, had tried it before.

crates/proof-impl/eth-ee-acct/src/program.rs
storopoli: Yes, you don't have practical improvements, but you definitely restrict users/developers using your API/library with the wrong assumptions.

When it is OK to use a native zkVM? I don' think in production right? Only in testing. Hence, the `"test-utils"` feature-gate.

It is not "How this will benefit us". It is more "Let's make sure that people (us in the future) don't shoot themselves in the foot".

image

crates/ee-acct-runtime/src/update_processing.rs
storopoli: Ok leave as it is, then.

purusang:
I need to pause the pr until we get to the same page regarding how we prove ee account updates. @sapinb proposes we should be doing that by using inner & outer proofs where inner proof proves chunks and outer proof proves chunks + da inclusion. I am documenting this and will be sharing by Monday. So reviewing after discussion on that document would make more sense. cc: @delbonis @sapinb @storopoli @barakshani

crates/proof-impl/alpen-ee-acct/Readme.md
delbonis, line 1: This file isn't really necessary tbh.  It'd be better on Notion along with other docs explaining the rest of the supporting infra for exec / exec proofs.

crates/proof-impl/alpen-ee-acct/src/guest_builder.rs
delbonis: Also for `HashMismatch` it should be `blkid mismatch (expected {0}, computed {1})`, and maybe called `BlkidMismatch` since that's more precise.

crates/proof-impl/eth-ee-acct/src/program.rs
delbonis: It doesn't *need* to be `Codec`, but just throwing it together like this is a problem.

crates/proof-impl/eth-ee-acct/src/guest_builder.rs
```diff
@@ -0,0 +1,98 @@
...
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum GuestBuilderError {
+    #[error("SSZ decode error")]
+    SszDecode,
+    #[expect(dead_code, reason = "Reserved for future codec decoding errors")]
+    #[error("Codec decode error: {0}")]
+    CodecDecode(String),
+    #[error("Block hash mismatch: expected {expected:?}, computed {computed:?}")]
+    HashMismatch { expected: Buf32, computed: Buf32 },
+}
```
delbonis, line 15: error string styling!

crates/proof-impl/eth-ee-acct/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+    let computed_hash = strata_identifiers::Hash::from(computed_hash.0);
+
+    // Compare with the expected inner_state from ProofState
+    if computed_hash != expected_proof_state.inner_state() {
+        panic!(
+            "Account state hash mismatch: computed {:?}, expected {:?}",
+            computed_hash,
+            expected_proof_state.inner_state()
+        );
+    }
```
delbonis: Should use `assert_eq!` here, and can probably avoid string interpolation.  This only ever uses `expected_proof_State.inner_state()` so we should just pass the inner state instead of the whole `ProofState`.

crates/ee-acct-runtime/src/update_builder.rs
sapinb: please rename not_package to package and forget it 😄

crates/proof-impl/eth-ee-acct/src/lib.rs
purusang: For this pr we won't handle DA so resolving it.

crates/proof-impl/eth-ee-acct/src/program.rs
purusang: [Planned](https://alpenlabs.atlassian.net/browse/STR-2059) it separately for tech debt sprint.

crates/proof-impl/alpen-ee-acct/src/guest_builder.rs
purusang: Please check recent commit, the approach is different now.

crates/proof-impl/alpen-ee-acct/src/outer.rs
```diff
@@ -0,0 +1,142 @@
...
+        outputs,
+        extra_data,
+    );
+
+    zkvm.commit_buf(&update_proof_params.as_ssz_bytes());
+}
+
+/// Verifies that chunk states are continuous: chunk[i].new_state == chunk[i+1].prev_state
+fn verify_state_continuity(chunk_outputs: &[ChunkProofOutput]) {
+    for i in 0..chunk_outputs.len() - 1 {
```
prajwolrg: nit: we can maybe use `chunk_outputs.windows(2)` to do this more idiomatically

crates/proof-impl/alpen-ee-acct/src/chunk_program.rs
```diff
@@ -0,0 +1,177 @@
...
+        // Write raw_blocks: count + items
+        let blocks_count = input.raw_blocks.len() as u32;
+        input_builder.write_buf(&blocks_count.to_le_bytes())?;
+        for block in &input.raw_blocks {
+            input_builder.write_buf(block)?;
+        }
+
+        // Write raw buffers
+        input_builder.write_buf(&input.raw_prev_header)?;
+        input_builder.write_buf(&input.raw_partial_pre_state)?;
```
prajwolrg, line 136: nit: We can serialize the input directly instead of writing each field.

purusang: I had tried to do that in the past but got reviews to avoid copies happening due to an extra layer of serialization. The struct has fields with mixed serializations (serde, SSZ, and raw buffers), so we cannot simply derive serialize on it. A manual implementation would still need to call as_ssz_bytes() on each SSZ field internally, copy those into an intermediate buffer, then write that buffer to input_builder—adding an extra copy compared to writing each field directly.

PR #1202, opened 2025-12-10
WriteBatch-generating and indexing state accessor layers

delbonis:
## Description

This PR is in support of STR-1822 to add state accessor layers fulfilling the original vision for the state accessor trait hierarchy in ledger-types.  This involved some moderate changes to how modifying accounts works in order to achieve.  The biggest change is that account updates have to be scoped within a transactional-ish closure.  This is necessary to ensure that modifications/indexing is actually remembered, by putting more explicit scoping.  This also simplifies some of the accessors on the account types that were more complex than they needed to be, which was a result of some older account design experimentation.

This increase in complexity was not apparent when the traits were initially defined, however this structure should be good for future needs now because any future need we might have would fall into one of these general patterns.

Claude was used pretty extensively.  It did a lot of lifetime bound troubleshooting when I was working through a more-efficient but apparently impossible idea involving GATs/HRTBs to improve ergonomics and avoid clones.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This PR is not 100% complete, it still needs:
* ~~proper unit tests to be added to ensure that these wrapper layers behave correctly~~
* snark account update indexing (right now it's just messages)
* ~~maybe: a read-only "write batch stack" state accessor impl that we can plug in underneath the write tracking layer we can use in-practice~~
* maybe: replicate the `Mut` pattern to have an `IStateAccessorMut` trait so:
    * the read-only "write batch stack" state accessor doesn't have to just fail on the functions that modify them
    * we can use with the on-disk state in the future without complicating things, since its interface would be purely write-batch based

## Checklist

### DIFF
--- a/crates/ol/state-support-types/src/indexer_layer.rs
+++ b/crates/ol/state-support-types/src/indexer_layer.rs
@@ -1061,7 +1061,7 @@ mod tests {
         match &writes.snark_state_updates()[0] {
             SnarkAcctStateUpdate::DirectSet(s) => {
                 assert_eq!(s.account_id(), account_id);
-                assert_eq!(s.state(), new_hash);
+                assert_eq!(s.state(), new_hash.0);
                 assert_eq!(s.next_read_idx(), next_read_idx);
                 assert_eq!(s.seqno(), seqno);
             }
@@ -1102,7 +1102,7 @@ mod tests {
         match &writes.snark_state_updates()[0] {
             SnarkAcctStateUpdate::Update(s) => {
                 assert_eq!(s.account_id(), account_id);
-                assert_eq!(s.inner_state(), inner_state);
+                assert_eq!(s.inner_state(), inner_state.0);
                 assert_eq!(s.next_read_idx(), next_read_idx);
                 assert_eq!(s.seqno(), seqno);
                 assert_eq!(s.extra_data(), extra_data);


storopoli:
> impossible idea involving GATs/HRTBs to improve ergonomics and avoid clones.

Would `RefCell`s help here?

storopoli (approved):
ACK a08969ce2cbdf670c34729bbfb07f4d801f8a13c

But let's wait for @bewakes' review as well.

delbonis:
Report from Claude on tests:

```
  Files Created/Modified:

  1. Cargo.toml - Added bitcoin dev-dependency for test helpers
  2. src/lib.rs - Added test module declarations:
    - #[cfg(test)] mod test_utils;
    - #[cfg(test)] mod tests;
  3. src/test_utils.rs (new file) - Shared test utilities:
    - test_account_id() - Create deterministic AccountIds
    - test_hash() - Create deterministic Hash values
    - test_snark_account_state() - Create fresh snark accounts
    - test_message_entry() - Create test inbox messages
    - setup_state_with_snark_account() - Setup OLState with a snark account
  4. src/indexer_layer.rs - Added 21 tests covering:
    - Pass-through tests (slot, epoch, get_account, check_exists, create, compute_state_root)
    - Write tracking tests (inbox messages, manifests, multiple writes, cross-account)
    - Modification flag tests (balance add/take, snark update, no modification on read)
    - State consistency tests (direct vs wrapped comparison)
    - Write data accuracy tests (pre-insertion index, correct account ID)
  5. src/write_tracking_layer.rs - Added 13 tests covering:
    - Copy-on-write tests (read fallback, write copies to batch, read prefers batch)
    - Account creation tests (create in batch, find by serial)
    - Global/epochal state tests (slot, epoch, ledger balance, manifest)
    - State root test (returns Unsupported)
    - Batch extraction tests (into_batch returns modifications)
    - Error handling tests (update nonexistent account)
  6. src/tests.rs (new file) - 8 combined layer integration tests:
    - test_indexer_over_write_tracking_basic - Basic layer composition
    - test_combined_inbox_message_tracking - Inbox writes through both layers
    - test_combined_manifest_tracking - Manifest writes through both layers
    - test_combined_balance_modification - Balance changes through both layers
    - test_combined_account_creation - Account creation through both layers
    - test_combined_global_state_modification - Slot/epoch changes
    - test_combined_multiple_operations - Complex multi-operation scenario
    - test_combined_layers_preserve_base_state - Verifies base state isolation
```

crates/ol/state-types/src/serial_map.rs
```diff
@@ -0,0 +1,406 @@
...
+/// This is useful when tracking newly-created accounts.
+#[derive(Clone, Debug)]
+pub struct SerialMap {
+    /// Tracks the serial of the first account ID.  This is invalid if the vec
+    /// is empty.
+    first: AccountSerial,
+
+    /// Tracks account IDs, ordered by serial.  If this is empty, then the first
+    /// serial is unset.
+    ids: Vec<AccountId>,
```
bewakes, line 14: > This is invalid if the vec is empty.

Can we make this impossible by construct? I think we can make the constructor take in the first serial and id and get rid of all the annoying if elses in the methods. Shouldn't this be sufficient for our use case where we can always assign `Serial(0)` to `AccountId::zero()` ?

bewakes: This is basically allowing invalid state representable due to which we need `assert!` here and there. Runtime panics are really not desirable.

bewakes: I suggest to use some sort of `NonEmptyVec` for the `ids` field now that we are here making everything type safe.

crates/ol/state-support-types/src/batch_diff_layer.rs
```diff
@@ -0,0 +1,544 @@
...
+/// Reads check each batch in reverse order (last = most recent), then fall back
+/// to the base state. All write operations return `AcctError::Unsupported` or
+/// silently no-op (for setters that return `()`).
+///
+/// The batch slice can be empty, making this a read-only wrapper for the base.
+/// This is useful for scenarios where you want to view state with pending
+/// changes applied without modifying anything.
+pub struct BatchDiffState<'batches, 'base, S: IStateAccessor> {
+    base: &'base S,
+    batches: &'batches [WriteBatch<S::AccountState>],
```
bewakes, line 24: Would it be more efficient if we just use a merged `WriteBatch` here? We would merge by consuming an earlier batch to avoid clones.
That way we don't have to linearly scan through batches to find accounts.

crates/ol/state-types/src/serial_map.rs
delbonis: > Can we make this impossible by construct?

It *is* impossible to construct an invalid `SerialMap`, in the sense that all pub functions (without `_unchecked`) will never panic.  This is ensured with the unit tests.

> I think we can make the constructor take in the first serial and id and get rid of all the annoying if elses in the methods.

Then we must have access to the state to initialize these counters when creating an new empty write batch.  This was intended to fix this issue.

crates/ol/state-support-types/src/batch_diff_layer.rs
delbonis: The idea is that we can keep track of each block's wb separately and we can revert the state on reorg just by doing `.pop()`.  If we want to avoid the linear scan we can just do the aggregation before construction and only pass a single write batch to the constructor, so this approach makes it strictly more useful.

I have half a mind to make this be actually be a `impl Fn() -> impl Iterator<Item = &'b WriteBatch<S::AccountState>>`.

storopoli (approved):
ACK be530c9720f1c2c95876f8d6fe5bb4ddac95fd4d

Check code formatting, is failing

delbonis:
First pass on that ci test run failed with `sync_fullnode_l2_lag`  timing out waiting for an epoch to be confirmed. :thinking: 

I didn't touch anything that could cause that.

crates/acct-types/src/state.rs
```diff
@@ -96,8 +81,8 @@ impl AccountIntrinsicState {
     }
 
     /// Attempts to parse the type into a valid [`AccountTypeId`].
-    pub fn ty(&self) -> AcctResult<AccountTypeId> {
-        AccountTypeId::try_from(self.raw_ty()).map_err(AcctError::InvalidAcctTypeId)
+    pub fn ty(&self) -> AccountTypeId {
+        AccountTypeId::try_from(self.raw_ty()).expect("acct: invalid id")
```
prajwolrg: The docstring doesn't seem updated. Should add something about panic.

bewakes (dismissed):
Looks good. This needs a rebase though.

storopoli (approved):
ACK 1a3ac46b0d738019340ed0a94f084ce7c0528949

You can do a fast-track merge after rebasing.

storopoli:
CleanShot 2025-12-17 at 11 04 09@2x

PR #1203, opened 2025-12-11
feat(key-derive): derive P2P operator keys using ed25519

storopoli:
## Description

`strata-p2p` v2 uses ed25519 for the operator message signing for reasons specified in the [`strata-p2p` v2 specs](https://www.notion.so/Technical-P2P-Scope-v2-1f2901ba000f80b1981ad4e2f1990415)

> LibP2P implementations ***must support*** Ed25519 ****keys and signatures. Ed25519 is the [[most well-implemented and supported authentication method in LibP2P](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types)](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types); and the only one that is mandatory to be supported in the protocol.
> 
> Ed25519 is generally preferred over ECDSA on secp256k1 (what the v1 of this spec used) for authentication for several important reasons:
> 
> - **Faster performance**: Ed25519 provides significantly faster signature verification (2-3x faster than ECDSA on secp256k1) and generation speed.
> - **Smaller signatures**: Ed25519 produces **64-byte signatures** compared to ECDSA on secp256k1's variable-length signatures (typically 70-72 bytes).
> - **Deterministic signatures**: Ed25519 is inherently deterministic, eliminating the need for a secure random number generator during signing. ECDSA on secp256k1 by default is not, only if paired with the [[RFC 6979](https://www.rfc-editor.org/rfc/rfc6979)](https://www.rfc-editor.org/rfc/rfc6979) for ECDSA deterministic nonces.

However, `strata-key-deriv` still uses secp256k1.

The `strata-bridge` uses `strata-key-deriv` to generate its operator keys, hence we need this PR to unlock https://github.com/alpenlabs/strata-bridge/pull/289.

### Type of Change

- [x] Documentation update
- [x] Refactor
- [x] New or updated tests
- [x] Dependency Update

## Notes to Reviewers

This PR has no new dependency since we already have `ed25519-dalek` in our `Cargo.lock` due to transitive dependencies.

Also this was partially done with the aid of Cursor.

## Checklist

## Related Issues

Part of STR-1650.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13856,6 +13856,7 @@ dependencies = [
  "arbitrary",
  "bitcoin",
  "borsh",
+ "ed25519-dalek",
  "musig2",
  "rand 0.8.5",
  "serde",
--- a/crates/bridge-types/Cargo.toml
+++ b/crates/bridge-types/Cargo.toml
@@ -12,14 +12,16 @@ strata-primitives.workspace = true
 arbitrary.workspace = true
 bitcoin.workspace = true
 borsh.workspace = true
+ed25519-dalek.workspace = true
 musig2.workspace = true
-rand.workspace = true
 serde.workspace = true
 sha2.workspace = true
 thiserror.workspace = true
 
 [dev-dependencies]
 strata-test-utils.workspace = true
+
+rand.workspace = true
 
 [features]
 default = ["std"]
--- a/crates/bridge-types/src/relay/util.rs
+++ b/crates/bridge-types/src/relay/util.rs
@@ -1,10 +1,7 @@
 use std::io;
 
-use bitcoin::secp256k1::{
-    schnorr::Signature, Keypair, Message, SecretKey, XOnlyPublicKey, SECP256K1,
-};
 use borsh::BorshSerialize;
-use rand::rngs::OsRng;
+use ed25519_dalek::{Signature, Signer, SigningKey, Verifier, VerifyingKey};
 use strata_primitives::buf::{Buf32, Buf64};
 use thiserror::Error;
 
@@ -53,7 +50,6 @@ impl MessageSigner {
         };
 
         let id: Buf32 = tmp_m.compute_id().into();
-        // WARN: I don't know if a global context is safe here, maybe.
         let sig = sign_msg_hash(&self.msg_signing_sk, &id);
         tmp_m.sig = sig;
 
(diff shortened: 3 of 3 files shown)


Cargo.toml
```diff
@@ -406,6 +406,7 @@ clap = { version = "4.5.39", default-features = false, features = ["derive"] }
 criterion = { version = "0.7.0", features = ["html_reports"] }
 deadpool = "0.12.1"
 digest = "0.10"
+ed25519-dalek = { version = "2.2.0", features = ["zeroize"] }
```
AaronFeickert, line 409: Very good dependency choice here.

crates/key-derivation/src/operator.rs
```diff
@@ -200,31 +201,9 @@ impl Zeroize for OperatorKeys {
...
-        unsafe {
-            *raw_ptr = if message.child_number.is_normal() {
-                ChildNumber::Normal { index: 0 }
-            } else {
-                ChildNumber::Hardened { index: 0 }
-            };
-        }
+        // NOTE: We don't need to zeroize the ed25519 signing key since it is already
+        //       zeroized on drop if we use `ed25519-dalek` with the `zeroize` feature.
+        //       Ref: <https://docs.rs/ed25519-dalek/latest/src/ed25519_dalek/signing.rs.html#665-666>.
```
AaronFeickert: This is not sufficient. It is the case that `SigningKey` is `ZeroizeOnDrop`, but the `Zeroize` trait is intended to capture _manual_ zeroization as well.

AaronFeickert: Just change to `message.zeroize()` and all will be well.

AaronFeickert (changes requested):

crates/key-derivation/src/operator.rs
storopoli: Ok this is very tricky. Check 88bbf6fd992df746626dcda1c07ce5f0bf45ec70

`ed25519-dalek` with the `zeroize` feature does NOT implement `Zeroize` only `ZeroizeOnDrop` for `SigningKey`. Hence, I cannot call `.zeroize()` on it.

See https://docs.rs/ed25519-dalek/latest/src/ed25519_dalek/signing.rs.html#658-666

The internal `SecretKey` has a `Zeroize` but that inner field is non-`pub` (which should be anyways).

AaronFeickert: Ooh, I forgot that there was a specific discussion about this [here](https://github.com/dalek-cryptography/curve25519-dalek/pull/782) and [here](https://github.com/dalek-cryptography/curve25519-dalek/pull/747) earlier this year. I disagree with the conclusion, but it is what it is. Good catch.

storopoli:
Ok just realized that we also have some leftovers in `bridge_types::relay` that were using secp2556k1. Done in https://github.com/alpenlabs/alpen/pull/1203/commits/01924deccf90f8cde81ae393ceed97c245770766

PR #1204, opened 2025-12-12
STR 1967 SSZ support for ee-acct-types

krsnapaudel:
## Description

Add SSZ support for `ee-acct-types` and fix dependent code.

This PR was created with help from Cursor AI.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Not sure about constant values. Please point to spec docs that have them.

## Checklist

### DIFF
--- a/crates/ee-acct-types/src/lib.rs
+++ b/crates/ee-acct-types/src/lib.rs
@@ -13,13 +13,25 @@ mod outputs;
 mod state;
 mod traits;
 
-pub use commit::{CommitBlockData, CommitChainSegment};
+// Include generated SSZ types from build.rs output
+#[allow(
+    clippy::all,
+    unreachable_pub,
+    clippy::allow_attributes,
+    reason = "generated code"
+)]
+mod ssz_generated {
+    include!(concat!(env!("OUT_DIR"), "/generated.rs"));
+}
+
+pub use commit::*;
 pub use errors::{EnvError, EnvResult, MessageDecodeError, MessageDecodeResult};
-pub use extra_data::UpdateExtraData;
+pub use extra_data::*;
 pub use inputs::ExecPayload;
 pub use messages::*;
 pub use outputs::ExecBlockOutput;
-pub use state::{EeAccountState, PendingFinclEntry, PendingInputEntry, PendingInputType};
+pub use ssz_generated::ssz::{self as ssz, state::*};
+pub use state::PendingInputType;
 pub use strata_identifiers::Hash;
 pub use traits::{
     BlockAssembler, ExecBlock, ExecBlockBody, ExecHeader, ExecPartialState, ExecutionEnvironment,
--- a/crates/ee-chain-types/src/lib.rs
+++ b/crates/ee-chain-types/src/lib.rs
@@ -16,9 +16,6 @@ mod ssz_generated {
 }
 
 // Publicly re-export only the SSZ items this crate's API intends to expose
-pub use ssz_generated::ssz::block::{
-    BlockInputs, BlockOutputs, ExecBlockCommitment, ExecBlockPackage, OutputTransfer,
-    SubjectDepositData,
-};
+pub use ssz_generated::ssz::block::*;
 
 mod block;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14367,12 +14367,21 @@ dependencies = [
 name = "strata-ee-acct-types"
 version = "0.1.0"
 dependencies = [
+ "proptest",
+ "ssz",
+ "ssz_codegen",
+ "ssz_derive",
+ "ssz_primitives",
+ "ssz_types",
  "strata-acct-types",
  "strata-codec",
  "strata-ee-chain-types",
  "strata-identifiers",
  "strata-msg-fmt",
+ "strata-test-utils-ssz",
  "thiserror 2.0.17",
+ "tree_hash 0.11.0",
+ "tree_hash_derive 0.11.0",
 ]
 
 [[package]]
--- a/crates/ee-acct-types/Cargo.toml
+++ b/crates/ee-acct-types/Cargo.toml
@@ -4,13 +4,26 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
+ssz.workspace = true
+ssz_derive.workspace = true
+ssz_primitives.workspace = true
+ssz_types.workspace = true
 strata-acct-types.workspace = true
 strata-codec.workspace = true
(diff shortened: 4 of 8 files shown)


crates/ee-acct-runtime/Cargo.toml
```diff
@@ -9,6 +9,7 @@ builders = ["dep:thiserror"]
 test-utils = []
 
 [dependencies]
+ssz.workspace = true
```
delbonis: ordering

crates/evm-ee/src/types/block.rs
```diff
@@ -42,3 +47,70 @@ impl ExecBlock for EvmBlock {
...
+        buf.extend_from_slice(&header_offset.to_le_bytes());
+        buf.extend_from_slice(&body_offset.to_le_bytes());
+
+        // Variable part: header and body data
+        self.header.ssz_append(buf);
+        self.body.ssz_append(buf);
+    }
+}
+
+impl Decode for EvmBlock {
```
delbonis: Don't really like this ad-hoc manual encode/decode.

SSZ isn't providing us much value here since the things we'd want to make SSZ proofs about (txs, receipts, etc.) are contained in an RLP structure, so we couldn't do that anyways.  I think this was fine being `Codec` as-is.

crates/evm-ee/src/types/partial_state.rs
delbonis, line 1: This really should not need to be SSZ, especially with all this added complexity!  Especially if it adds more bytes that we need to expose to the proof.  Arguably this might be better as rkyv so we can do better zero-copy deserialization.

crates/ee-acct-runtime/Cargo.toml
krsnapaudel: Based on what @storopoli said about ssz-* being Alpen internal, I tried to make this consistent in touched files.

storopoli: Yeah we're trying to have `alpen-*`, `strata-*` and `ssz-*` in the same dependency block in `Cargo.toml`s.
I think it's a useful criterion, since it denotes easily what is "under our control"/internal deps and what is not.

crates/ol/chain-types/ssz/transaction.ssz
```diff
@@ -30,12 +30,10 @@ class SnarkAccountUpdateTxPayload(Container):
...
 
-### Type aliases for union variant names
-GenericAccountMessage = GamTxPayload
-SnarkAccountUpdate = SnarkAccountUpdateTxPayload
-
 ### Transaction payload union.
-TransactionPayload = Union[GenericAccountMessage, SnarkAccountUpdate]
+class TransactionPayload(Union):
+    GenericAccountMessage: GamTxPayload
+    SnarkAccountUpdate: SnarkAccountUpdateTxPayload
```
delbonis, line 36: Wait since when are we allowed to do this?

crates/ee-chain-types/src/lib.rs
```diff
@@ -17,8 +17,9 @@ mod ssz_generated {
 
 // Publicly re-export only the SSZ items this crate's API intends to expose
 pub use ssz_generated::ssz::block::{
-    BlockInputs, BlockOutputs, ExecBlockCommitment, ExecBlockPackage, OutputTransfer,
-    SubjectDepositData,
+    BlockInputs, BlockInputsRef, BlockOutputs, BlockOutputsRef, ExecBlockCommitment,
+    ExecBlockCommitmentRef, ExecBlockPackage, ExecBlockPackageRef, OutputTransfer,
+    OutputTransferRef, SubjectDepositData, SubjectDepositDataRef,
 };
```
delbonis: At this point we should just do `::*` imports, unless this helps LLMs figure things out when trying to refer back to type defs?

crates/ee-acct-types/src/messages.rs
delbonis, line 1: Account messages are generally not expected to be SSZed.  They go on L1 and we would care about every byte, so we'd use Codec serialization where we have precise control.

crates/ee-acct-types/src/lib.rs
```diff
@@ -13,13 +13,35 @@ mod outputs;
...
+    extra_data::{UpdateExtraData, UpdateExtraDataRef},
+    messages::{
+        CommitMsgData, CommitMsgDataRef, DepositMsgData, DepositMsgDataRef, SubjTransferMsgData,
+        SubjTransferMsgDataRef,
+    },
+    state::{
+        EeAccountState, EeAccountStateRef, PendingFinclEntry, PendingFinclEntryRef,
+        PendingInputEntry, PendingInputEntryRef,
+    },
+    {self as ssz},
```
delbonis: Did this get messed up somehow?

But also with this generally, do we think this can safely be converted to wildcards? @storopoli

crates/ee-acct-types/ssz/extra_data.ssz
delbonis, line 1: This also goes on L1 so we shouldn't SSZ it.  This would have the same byte encoding in Codec for now but I'm not sure we can commit to this.

crates/ol/chain-types/ssz/transaction.ssz
krsnapaudel: Since [this PR](https://github.com/alpenlabs/ssz-gen/pull/51). We could not specify an enum with the following reqs:
1. Variant name is different from type.
2. The type came from external crate.
3. The type was a container.

crates/ee-acct-types/src/lib.rs
krsnapaudel: Using `::*`.

krsnapaudel:
> Sorry I didn't catch these the last time around. :/

I have addressed the comments. Hopefully good now.

crates/ee-chain-types/src/lib.rs
storopoli: I don't like doing `::*` I think it's bad practice and could introduce security and maintainability issues.

crates/ee-acct-types/src/lib.rs
storopoli: I don't like doing `::*` I think it's bad practice and could introduce security and maintainability issues.

storopoli: But fine, we can do it.

crates/ee-chain-types/src/lib.rs
storopoli: But fine, we can do it.

PR #1205, opened 2025-12-12
ci(functional): run tests in parallel

storopoli:
## Description

The idea behind this PR is to run all functional tests in parallel in CI. But this is really tricky to accomplish if we need to have coverage from the stuff that is being run in the functional tests.

So one trade-off is to give up coverage reports in PRs from the functional tests and we run them in parallel and we benefit from the fast pace of iterating over PRs. Instead, we only run coverage reports using the old functional test CI sequential setup when the commit hits main `branch`.

So, functional tests will run without coverage reports, but parallel and faster in PRs. And we get a full coverage report when the PR is merged to `main`. 

It does take 20min to run in a "cold" no-cache parallel run for functional tests now because of compiling the binaries.
However, when we have the cache "hot" it takes 8 min to run the functional tests.

This is a HUGE improvement over the average 33min that we have now.
Also this has a caveat. If your functional test fails (which is the case because of its flakiness) then you need to run the WHOLE thing again. Whereas, with the parallel approach you can trigger to run only the specific functional test that it failed (which can be in less than 5 minutes).

Below is an image showing what is run in a PR parallel setting versus `main` commits. As a bonus you can see that in a "hot cache" the functional tests is not the chokepoint anymore and it's actually the unit test 🎉

CleanShot 2025-12-12 at 15 45 13@2x

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

I had to do a trade-off. We cannot let the 60+ tests do a race condition to see who will write the cache slot. Hence I picked it up the `client_status` because it is one of the simplest, faster to run, and probably the one that is less flaky. Hence, that specific parallel runner will be the one that will write to the Rust cache, and all the other tests will use it. 

Used Cursor.

## Checklist

## Related Issues

STR-1989

### DIFF
--- a/functional-tests/envs/testenv.py
+++ b/functional-tests/envs/testenv.py
@@ -15,6 +15,36 @@ from utils.constants import *
 from utils.wait import ProverWaiter, RethWaiter, StrataWaiter
 
 
+def _retry_bitcoind_rpc(
+    logger: logging.Logger,
+    desc: str,
+    fn,
+    *,
+    timeout_s: float = 180.0,
+    interval_s: float = 1.0,
+):
+    """
+    Best-effort retry wrapper for bitcoind JSON-RPC calls.
+
+    We use it during env init to reduce flakiness when bitcoind is still warming up
+    (especially when CI runs each test in its own fresh runner).
+    """
+    deadline = time.monotonic() + timeout_s
+    last_exc: Exception | None = None
+
+    while time.monotonic() < deadline:
+        try:
+            return fn()
+        except Exception as e:  # noqa: BLE001 - broad retry is intentional for test harness
+            last_exc = e
+            logger.info(f"bitcoind RPC '{desc}' failed; retrying in {interval_s:.1f}s: {e!r}")
+            time.sleep(interval_s)
+
+    raise RuntimeError(
+        f"bitcoind RPC '{desc}' did not succeed within {timeout_s:.0f}s"
+    ) from last_exc
+
+
 class StrataTestBase(flexitest.Test):
     """
     Class to be used instead of flexitest.Test for accessing logger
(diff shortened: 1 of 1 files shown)


.github/workflows/functional.yml
```diff
@@ -51,10 +51,176 @@ jobs:
...
+          - prover/prover_el_precompiles
+          - prover/prover_el_selfdestruct
+          - prover/prover_el_selfdestruct_to_address
+          - prover/prover_schnorr_precompile
+          - sync/sync_bitcoin_reorg
+          - sync/sync_from_rpc
+          - sync/sync_fullnode_l2_lag
+          - sync/sync_fullnode_l2_lag_restart
+          - sync/sync_genesis
+          - sync/sync_l1_finalization
```
voidash: This means we have to append to `.github/workflows/functional.yml` everytime a new functional test is added otherwise it won't run in parallel. 

We might be able to read the folder contents and then pass it instead of us having to list the files ourselves

voidash: Something like

```
 - name: Discover tests
    id: discoverer
        run:
          tests=$(uv run python entry.py --list-tests-to-run-in-parallel-or-something-like-that)
          echo "tests=$tests" >> "$GITHUB_OUTPUT"
```

then 

```
 matrix:
        test: ${{ fromJson(discoverer.outputs.tests) }}
```

storopoli: Ooh yes, that's nice. See ad65b929803c46f4c244fea0ae4795388065bbf9

storopoli:
Damn this was really nice I had to refactor one of the commits and all the suite of tests ran in 14 min with the biggest bottleneck being unit tests and NOT functional tests because we took advantage of a "hot cache".

I imagine this would be the scenario that we'd encounter in a lot of our workflows and tasks around PRs.

voidash (approved):

storopoli (commented):
ACK 9dae1241d2256c3913e7f480f4ddfd0ec1bddcb6

PR #1207, opened 2025-12-15
refactor: use Hash as alias of Buf32

sapinb:
## Description
Converting Hash type to be an alias over Buf32 for better logging.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/identifiers/src/hash.rs
+++ b/crates/identifiers/src/hash.rs
@@ -6,6 +6,7 @@ use sha2::Sha256;
 
 use crate::buf::Buf32;
 
+/// Alias to Buf32 used as a universal type in EE.
 pub type Hash = Buf32;
 
 /// Direct untagged hash.


crates/identifiers/src/buf.rs
```diff
@@ -60,6 +60,7 @@ impl Zeroize for Buf20 {
 /// assert_eq!(buf, Buf32::from([0; 32]));
 /// ```
 #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Encode, Decode)]
+#[repr(transparent)]
```
storopoli, line 63: Why do we need this?

crates/identifiers/src/hash.rs
```diff
@@ -6,6 +6,8 @@ use sha2::Sha256;
 
 use crate::buf::Buf32;
 
+pub type Hash = Buf32;
```
storopoli, line 10: docstrings

crates/identifiers/src/buf.rs
sapinb: There is a conversion between FixedBytes<32> to Hash (previously to [u8; 32]) here

https://github.com/alpenlabs/alpen/pull/1207/changes#diff-eee9648bdfb82110c311eaeea0bcc66043d8282123f9acc7cbd033b74bef2afaL63-R64

Need to add repr(transparent) both here and in FixedBytes in ssz crate

storopoli: We should definitely add `cargo-miri` to the workflows now that we are starting to have bunch of `unsafe` code (will open a ticket). Can you please run `miri` locally just for sanity checks

storopoli: https://alpenlabs.atlassian.net/browse/STR-1992

PR #1209, opened 2025-12-15
SPS-62 checkpoint subprotocol

irnb:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/asm/txs/checkpoint/src/test_utils/mmr.rs
+++ b/crates/asm/txs/checkpoint/src/test_utils/mmr.rs
@@ -217,8 +217,7 @@ fn verified_aux_data_with_genesis(
 
 /// Creates verified auxiliary data for a range of L1 block heights.
 ///
-/// Convenience wrapper that uses genesis_height = 0.
-/// See [`verified_aux_data_with_genesis`] for full documentation.
+/// Convenience wrapper that uses `genesis_height = 0`.
 ///
 /// # Arguments
 ///
--- a/provers/sp1/guest-checkpoint/Cargo.lock
+++ b/provers/sp1/guest-checkpoint/Cargo.lock
@@ -203,12 +203,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "byteorder"
-version = "1.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
-
-[[package]]
 name = "cc"
 version = "1.2.41"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -262,12 +256,6 @@ checksum = "59ed5838eebb26a2bb2e58f6d5b5316989ae9d08bab10e0e6d103e656d1b0280"
 dependencies = [
  "libc",
 ]
-
-[[package]]
-name = "crunchy"
-version = "0.2.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "460fbee9c2c2f33933d720630a6a0bac33ba7053db5344fac858d4b8952d77d5"
 
 [[package]]
 name = "crypto-bigint"
@@ -627,9 +615,6 @@ name = "lazy_static"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bbd2bcb4c963f2ddae06a2efc7e9f3591312473c50c6685e1f298068316e66fe"
-dependencies = [
- "spin",
-]
 
 [[package]]
 name = "libc"
@@ -1046,12 +1031,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "48fd7bd8a6377e15ad9d42a8ec25371b94ddc67abe7c8b9127bec79bebaaae18"
 
 [[package]]
-name = "rustc-hex"
-version = "2.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e75f6a532d0fd9f7f13144f392b6ad56a32696bfcd9c78f797f16bbb6f072d6"
-
-[[package]]
 name = "rustversion"
 version = "1.0.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1183,8 +1162,8 @@ dependencies = [
 
 [[package]]
 name = "sizzle-parser"
-version = "0.11.1"
-source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.11.1#a22b12f83a569599744412990934683dd7466cc4"
+version = "0.12.0"
+source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.12.0#b1bbc429259c5575223ebfa1f979464a2f20dacf"
 dependencies = [
  "thiserror 2.0.17",
 ]
@@ -1247,12 +1226,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "spin"
-version = "0.9.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6980e8d7511241f8acf4aebddbb1ff938df5eebe98691418c4468d0b72a96a67"
-
-[[package]]
 name = "spki"
 version = "0.7.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1264,8 +1237,8 @@ dependencies = [
 
 [[package]]
 name = "ssz"
-version = "0.11.1"
-source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.11.1#a22b12f83a569599744412990934683dd7466cc4"
+version = "0.12.0"
+source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.12.0#b1bbc429259c5575223ebfa1f979464a2f20dacf"
 dependencies = [
  "hex",
  "itertools 0.13.0",
@@ -1277,8 +1250,8 @@ dependencies = [
 
 [[package]]
 name = "ssz_codegen"
-version = "0.11.1"
-source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.11.1#a22b12f83a569599744412990934683dd7466cc4"
+version = "0.12.0"
+source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.12.0#b1bbc429259c5575223ebfa1f979464a2f20dacf"
 dependencies = [
  "prettyplease",
  "proc-macro2",
@@ -1287,6 +1260,7 @@ dependencies = [
  "sizzle-parser",
  "ssz",
  "ssz_derive",
+ "ssz_primitives",
  "ssz_types",
  "syn",
  "toml",
@@ -1296,8 +1270,8 @@ dependencies = [
 
 [[package]]
 name = "ssz_derive"
-version = "0.11.1"
-source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.11.1#a22b12f83a569599744412990934683dd7466cc4"
+version = "0.12.0"
+source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.12.0#b1bbc429259c5575223ebfa1f979464a2f20dacf"
 dependencies = [
  "darling",
  "quote",
@@ -1306,8 +1280,8 @@ dependencies = [
 
 [[package]]
 name = "ssz_primitives"
-version = "0.11.1"
-source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.11.1#a22b12f83a569599744412990934683dd7466cc4"
+version = "0.12.0"
+source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.12.0#b1bbc429259c5575223ebfa1f979464a2f20dacf"
 dependencies = [
  "hex",
  "rand 0.8.5",
@@ -1316,8 +1290,8 @@ dependencies = [
 
 [[package]]
 name = "ssz_types"
-version = "0.11.1"
-source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.11.1#a22b12f83a569599744412990934683dd7466cc4"
+version = "0.12.0"
+source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.12.0#b1bbc429259c5575223ebfa1f979464a2f20dacf"
 dependencies = [
  "itertools 0.13.0",
  "serde",
@@ -1326,6 +1300,26 @@ dependencies = [
  "ssz_primitives",
  "thiserror 2.0.17",
  "tree_hash",
+]
+
+[[package]]
+name = "strata-acct-types"
+version = "0.1.0"
+dependencies = [
+ "digest",
+ "sha2",
+ "ssz",
+ "ssz_codegen",
+ "ssz_derive",
+ "ssz_primitives",
+ "ssz_types",
+ "strata-btc-types",
+ "strata-codec",
+ "strata-identifiers",
+ "strata-merkle",
+ "thiserror 2.0.17",
+ "tree_hash",
+ "tree_hash_derive",
 ]
 
 [[package]]
@@ -1351,9 +1345,10 @@ version = "0.1.0"
 dependencies = [
  "borsh",
  "strata-asm-common",
- "strata-checkpoint-types",
+ "strata-checkpoint-types-ssz",
  "strata-codec",
  "strata-codec-utils",
+ "strata-identifiers",
  "strata-msg-fmt",
  "strata-predicate",
  "strata-primitives",
@@ -1488,9 +1483,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "strata-checkpoint-types-ssz"
+version = "0.1.0"
+dependencies = [
+ "borsh",
+ "serde",
+ "ssz",
+ "ssz_codegen",
+ "ssz_derive",
+ "ssz_primitives",
+ "ssz_types",
+ "strata-identifiers",
+ "strata-ol-chain-types-new",
+ "thiserror 2.0.17",
+ "tree_hash",
+ "tree_hash_derive",
+]
+
+[[package]]
 name = "strata-codec"
 version = "0.1.0"
-source = "git+https://github.com/alpenlabs/strata-common#b504cf613eacb861629fbedb36b17a7e54d229f1"
+source = "git+https://github.com/alpenlabs/strata-common#4223a23387fd632876964bc8391ef2becee10090"
 dependencies = [
  "strata-codec-derive",
  "thiserror 2.0.17",
@@ -1499,7 +1512,7 @@ dependencies = [
 [[package]]
 name = "strata-codec-derive"
 version = "0.1.0"
-source = "git+https://github.com/alpenlabs/strata-common#b504cf613eacb861629fbedb36b17a7e54d229f1"
+source = "git+https://github.com/alpenlabs/strata-common#4223a23387fd632876964bc8391ef2becee10090"
 dependencies = [
  "quote",
  "syn",
@@ -1569,7 +1582,7 @@ dependencies = [
 [[package]]
 name = "strata-l1-txfmt"
 version = "0.1.0"
-source = "git+https://github.com/alpenlabs/strata-common#b504cf613eacb861629fbedb36b17a7e54d229f1"
+source = "git+https://github.com/alpenlabs/strata-common#4223a23387fd632876964bc8391ef2becee10090"
 dependencies = [
  "bitcoin",
  "thiserror 2.0.17",
@@ -1578,20 +1591,27 @@ dependencies = [
 [[package]]
 name = "strata-merkle"
 version = "0.2.0"
-source = "git+https://github.com/alpenlabs/strata-common#b504cf613eacb861629fbedb36b17a7e54d229f1"
+source = "git+https://github.com/alpenlabs/strata-common#4223a23387fd632876964bc8391ef2becee10090"
 dependencies = [
  "borsh",
  "digest",
  "serde",
  "sha2",
+ "ssz",
+ "ssz_codegen",
+ "ssz_derive",
+ "ssz_primitives",
+ "ssz_types",
  "strata-codec",
  "thiserror 2.0.17",
+ "tree_hash",
+ "tree_hash_derive",
 ]
 
 [[package]]
 name = "strata-msg-fmt"
 version = "0.1.0"
-source = "git+https://github.com/alpenlabs/strata-common#b504cf613eacb861629fbedb36b17a7e54d229f1"
+source = "git+https://github.com/alpenlabs/strata-common#4223a23387fd632876964bc8391ef2becee10090"
 dependencies = [
  "thiserror 2.0.17",
 ]
@@ -1610,6 +1630,26 @@ dependencies = [
  "strata-state",
  "thiserror 2.0.17",
  "tracing",
+]
+
+[[package]]
+name = "strata-ol-chain-types-new"
+version = "0.1.0"
+dependencies = [
+ "int-enum",
+ "ssz",
+ "ssz_codegen",
+ "ssz_derive",
+ "ssz_primitives",
+ "ssz_types",
+ "strata-acct-types",
+ "strata-asm-common",
+ "strata-codec",
+ "strata-identifiers",
+ "strata-snark-acct-types",
+ "thiserror 2.0.17",
+ "tree_hash",
+ "tree_hash_derive",
 ]
 
 [[package]]
@@ -1642,7 +1682,7 @@ dependencies = [
 [[package]]
 name = "strata-predicate"
 version = "0.1.0"
-source = "git+https://github.com/alpenlabs/strata-common#b504cf613eacb861629fbedb36b17a7e54d229f1"
+source = "git+https://github.com/alpenlabs/strata-common#4223a23387fd632876964bc8391ef2becee10090"
 dependencies = [
  "borsh",
  "hex",
@@ -1657,7 +1697,6 @@ dependencies = [
  "thiserror 2.0.17",
  "tree_hash",
  "tree_hash_derive",
- "zkaleido-sp1-groth16-verifier",
 ]
 
 [[package]]
@@ -1704,6 +1743,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "strata-snark-acct-types"
+version = "0.1.0"
+dependencies = [
+ "ssz",
+ "ssz_codegen",
+ "ssz_derive",
+ "ssz_primitives",
+ "ssz_types",
+ "strata-acct-types",
+ "thiserror 2.0.17",
+ "tree_hash",
+ "tree_hash_derive",
+]
+
+[[package]]
 name = "strata-state"
 version = "0.3.0-alpha.1"
 dependencies = [
@@ -1724,19 +1778,6 @@ name = "strsim"
 version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7da8b5736845d9f2fcb837ea5d9e2628564b3b043a70948a3f0b778838c5fb4f"
-
-[[package]]
-name = "substrate-bn"
-version = "0.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "72b5bbfa79abbae15dd642ea8176a21a635ff3c00059961d1ea27ad04e5b441c"
-dependencies = [
- "byteorder",
- "crunchy",
- "lazy_static",
- "rand 0.8.5",
- "rustc-hex",
-]
 
 [[package]]
 name = "subtle"
@@ -1899,8 +1940,8 @@ dependencies = [
 
 [[package]]
 name = "tree_hash"
-version = "0.11.1"
-source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.11.1#a22b12f83a569599744412990934683dd7466cc4"
+version = "0.12.0"
+source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.12.0#b1bbc429259c5575223ebfa1f979464a2f20dacf"
 dependencies = [
  "digest",
  "sha2",
@@ -1912,8 +1953,8 @@ dependencies = [
 
 [[package]]
 name = "tree_hash_derive"
-version = "0.11.1"
-source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.11.1#a22b12f83a569599744412990934683dd7466cc4"
+version = "0.12.0"
+source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.12.0#b1bbc429259c5575223ebfa1f979464a2f20dacf"
 dependencies = [
  "darling",
  "quote",
@@ -2052,21 +2093,6 @@ dependencies = [
  "borsh",
  "hex",
  "serde",
- "zkaleido",
-]
-
-[[package]]
-name = "zkaleido-sp1-groth16-verifier"
-version = "0.1.0"
-source = "git+https://github.com/alpenlabs/zkaleido?tag=v0.1.0-alpha-rc18#cd7839e5bbed46d8304cd95a20e956af56e22dfc"
-dependencies = [
- "blake3",
- "borsh",
- "hex",
- "serde",
- "sha2",
- "substrate-bn",
- "thiserror 1.0.69",
  "zkaleido",
 ]
 
--- a/provers/sp1/guest-cl-stf/Cargo.lock
+++ b/provers/sp1/guest-cl-stf/Cargo.lock
@@ -203,12 +203,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "byteorder"
-version = "1.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
-
-[[package]]
 name = "cc"
 version = "1.2.41"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -262,12 +256,6 @@ checksum = "59ed5838eebb26a2bb2e58f6d5b5316989ae9d08bab10e0e6d103e656d1b0280"
 dependencies = [
  "libc",
 ]
-
-[[package]]
-name = "crunchy"
-version = "0.2.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "460fbee9c2c2f33933d720630a6a0bac33ba7053db5344fac858d4b8952d77d5"
 
 [[package]]
 name = "crypto-bigint"
@@ -599,9 +587,6 @@ name = "lazy_static"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bbd2bcb4c963f2ddae06a2efc7e9f3591312473c50c6685e1f298068316e66fe"
-dependencies = [
- "spin",
-]
 
 [[package]]
 name = "libc"
@@ -1008,12 +993,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "48fd7bd8a6377e15ad9d42a8ec25371b94ddc67abe7c8b9127bec79bebaaae18"
 
 [[package]]
-name = "rustc-hex"
-version = "2.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e75f6a532d0fd9f7f13144f392b6ad56a32696bfcd9c78f797f16bbb6f072d6"
-
-[[package]]
 name = "rustversion"
 version = "1.0.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1145,8 +1124,8 @@ dependencies = [
 
 [[package]]
 name = "sizzle-parser"
-version = "0.11.1"
-source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.11.1#a22b12f83a569599744412990934683dd7466cc4"
+version = "0.12.0"
+source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.12.0#b1bbc429259c5575223ebfa1f979464a2f20dacf"
 dependencies = [
  "thiserror 2.0.17",
 ]
@@ -1209,12 +1188,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "spin"
-version = "0.9.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6980e8d7511241f8acf4aebddbb1ff938df5eebe98691418c4468d0b72a96a67"
-
-[[package]]
 name = "spki"
 version = "0.7.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1226,8 +1199,8 @@ dependencies = [
 
 [[package]]
 name = "ssz"
-version = "0.11.1"
-source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.11.1#a22b12f83a569599744412990934683dd7466cc4"
+version = "0.12.0"
+source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.12.0#b1bbc429259c5575223ebfa1f979464a2f20dacf"
 dependencies = [
  "hex",
  "itertools 0.13.0",
@@ -1239,8 +1212,8 @@ dependencies = [
 
 [[package]]
 name = "ssz_codegen"
-version = "0.11.1"
-source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.11.1#a22b12f83a569599744412990934683dd7466cc4"
+version = "0.12.0"
+source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.12.0#b1bbc429259c5575223ebfa1f979464a2f20dacf"
 dependencies = [
  "prettyplease",
  "proc-macro2",
@@ -1249,6 +1222,7 @@ dependencies = [
  "sizzle-parser",
  "ssz",
  "ssz_derive",
+ "ssz_primitives",
  "ssz_types",
  "syn",
  "toml",
@@ -1258,8 +1232,8 @@ dependencies = [
 
 [[package]]
 name = "ssz_derive"
-version = "0.11.1"
-source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.11.1#a22b12f83a569599744412990934683dd7466cc4"
+version = "0.12.0"
+source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.12.0#b1bbc429259c5575223ebfa1f979464a2f20dacf"
 dependencies = [
  "darling",
  "quote",
@@ -1268,8 +1242,8 @@ dependencies = [
 
 [[package]]
 name = "ssz_primitives"
-version = "0.11.1"
-source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.11.1#a22b12f83a569599744412990934683dd7466cc4"
+version = "0.12.0"
+source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.12.0#b1bbc429259c5575223ebfa1f979464a2f20dacf"
 dependencies = [
  "hex",
  "rand 0.8.5",
@@ -1278,8 +1252,8 @@ dependencies = [
 
 [[package]]
 name = "ssz_types"
-version = "0.11.1"
-source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.11.1#a22b12f83a569599744412990934683dd7466cc4"
+version = "0.12.0"
+source = "git+https://github.com/alpenlabs/ssz-gen?tag=v0.12.0#b1bbc429259c5575223ebfa1f979464a2f20dacf"
 dependencies = [
  "itertools 0.13.0",
  "serde",
@@ -1288,6 +1262,26 @@ dependencies = [
  "ssz_primitives",
  "thiserror 2.0.17",
  "tree_hash",
+]
+
+[[package]]
+name = "strata-acct-types"
+version = "0.1.0"
+dependencies = [
+ "digest",
+ "sha2",
+ "ssz",
+ "ssz_codegen",
+ "ssz_derive",
+ "ssz_primitives",
+ "ssz_types",
+ "strata-btc-types",
+ "strata-codec",
+ "strata-identifiers",
+ "strata-merkle",
+ "thiserror 2.0.17",
+ "tree_hash",
+ "tree_hash_derive",
 ]
 
 [[package]]
@@ -1313,9 +1307,10 @@ version = "0.1.0"
 dependencies = [
  "borsh",
  "strata-asm-common",
- "strata-checkpoint-types",
+ "strata-checkpoint-types-ssz",
  "strata-codec",
  "strata-codec-utils",
+ "strata-identifiers",
  "strata-msg-fmt",
  "strata-predicate",
  "strata-primitives",
@@ -1450,9 +1445,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "strata-checkpoint-types-ssz"
+version = "0.1.0"
+dependencies = [
+ "borsh",
+ "serde",
+ "ssz",
+ "ssz_codegen",
+ "ssz_derive",
+ "ssz_primitives",
+ "ssz_types",
+ "strata-identifiers",
+ "strata-ol-chain-types-new",
+ "thiserror 2.0.17",
+ "tree_hash",
+ "tree_hash_derive",
+]
+
+[[package]]
 name = "strata-codec"
 version = "0.1.0"
-source = "git+https://github.com/alpenlabs/strata-common#b504cf613eacb861629fbedb36b17a7e54d229f1"
+source = "git+https://github.com/alpenlabs/strata-common#4223a23387fd632876964bc8391ef2becee10090"
 dependencies = [
  "strata-codec-derive",
  "thiserror 2.0.17",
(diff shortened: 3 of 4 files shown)


crates/checkpoint-types-ssz/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+strata-identifiers.workspace = true
+
+borsh.workspace = true
+serde.workspace = true
+ssz.workspace = true
+ssz_derive.workspace = true
+ssz_types.workspace = true
+thiserror.workspace = true
+tree_hash.workspace = true
+tree_hash_derive.workspace = true
```
storopoli, line 18: Let's move the `ssz-*` and `tree_hash*` deps to the `strata-*`/`alpen-*` block? They are both under the `alpenlabs` GH org and pointing towards git repos.

Ref: https://github.com/alpenlabs/alpen/pull/1204#discussion_r2619479852

crates/asm/txs/checkpoint/src/errors.rs
```diff
@@ -21,9 +21,15 @@ pub enum CheckpointTxError {
...
-    Deserialization(#[source] BorshIoError),
+    /// Failed to deserialize SSZ checkpoint payload.
+    #[error("failed to deserialize checkpoint payload: {0:?}")]
+    SszDecode(DecodeError),
+}
+
+impl From<DecodeError> for CheckpointTxError {
+    fn from(err: DecodeError) -> Self {
+        Self::SszDecode(err)
+    }
```
storopoli: You can just use `thiserror` `#[from]` here and avoid the manual conversion.

Image

From: https://docs.rs/thiserror/latest/thiserror/ (no pun intended)

crates/checkpoint-types-ssz/ssz/payload.ssz
```diff
@@ -0,0 +1,103 @@
...
+###
+### NOTE: This type is defined inline rather than using strata_identifiers::L1BlockCommitment
+### due to a type mismatch issue when the `bitcoin` feature is enabled:
+### - strata_identifiers::L1BlockCommitment uses `absolute::Height` (custom type)
+### - strata_identifiers::L1BlockCommitmentRef::to_owned() returns the SSZ-generated type (u32)
+### - This causes type mismatches in generated code when using `external_kind: container`
+###
+### TODO: Fix ssz_codegen to use trait method (ToOwnedSsz::to_owned) instead of inherent
+### method, which would allow proper type conversion via the From impl in strata_identifiers.
+### Once fixed, this can be replaced with: strata_identifiers.L1BlockCommitment
```
storopoli: This is not a docstring `###` and should be a comment `#`.

crates/checkpoint-types-ssz/ssz/state.ssz
```diff
@@ -0,0 +1,20 @@
...
+    #~# external_kind: container
+    terminal: strata_identifiers.OLBlockCommitment
+    ### The previous epoch's terminal block (for chain continuity).
+    ### Zero if this is the genesis epoch.
+    #~# external_kind: container
+    prev_terminal: strata_identifiers.OLBlockCommitment
+    ### The final L1 block viewed by this epoch's terminal block.
+    new_l1: payload.L1Commitment
+    ### The final state root of the epoch.
+    ### TODO: Validate - is this state root BEFORE or AFTER terminal block execution?
```
storopoli: This is not a docstring `###` and should be a comment `#`.

crates/asm/logs/src/checkpoint.rs
```diff
@@ -1,3 +1,13 @@
+//! Legacy checkpoint update log type.
+//!
+//! This module provides the `CheckpointUpdate` log type using the legacy checkpoint types
+//! from `strata-checkpoint-types` for compatibility with csm-worker.
+//!
+//! # TODO: cleanup
+//!
+//! Remove this module when csm-worker is deprecated after OL STF migration.
+//! At that point, rename `CheckpointUpdateSsz` (in `checkpoint_ssz.rs`) to `CheckpointUpdate`.
```
storopoli: Not docstring, but comment please

crates/asm/logs/src/checkpoint_ssz.rs
```diff
@@ -0,0 +1,93 @@
+//! SSZ-compatible checkpoint update log type.
+//!
+//! This module provides the `CheckpointUpdateSsz` log type for the checkpoint subprotocol,
+//! using the SPS-62 SSZ-based checkpoint types.
+//!
+//! # TODO: cleanup
+//!
+//! Rename `CheckpointUpdateSsz` to `CheckpointUpdate` and remove the
+//! `CHECKPOINT_UPDATE_SSZ_LOG_TYPE` constant when the legacy `CheckpointUpdate` (in
+//! `checkpoint.rs`) is removed after csm-worker deprecation.
```
storopoli: Not docstring, but comment please

crates/asm/logs/src/checkpoint_ssz.rs
```diff
@@ -0,0 +1,93 @@
...
+use strata_identifiers::EpochCommitment;
+use strata_msg_fmt::TypeId;
+use strata_primitives::l1::BitcoinTxid;
+
+use crate::constants::CHECKPOINT_UPDATE_SSZ_LOG_TYPE;
+
+/// Checkpoint update log details using SSZ types.
+///
+/// This struct uses the new `BatchInfo` and `BatchTransition` types from
+/// `checkpoint-types-ssz` for the checkpoint subprotocol.
```
storopoli: Is this relevant information? When "new" becomes "old". I mean do we have an expiration date on this docstring?

crates/asm/txs/checkpoint/src/errors.rs
irnb: Can't use `#[from]` here unfortunately, `ssz::DecodeError `doesn't `impl std::error::Error`, which `thiserror` requires. Could open an issue on `ssz-gen` to add it.

storopoli: Yes open an issue in `ssz-gen` please

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -0,0 +1,146 @@
...
+use strata_identifiers::{Buf32, CredRule, Epoch, OLBlockCommitment};
+use strata_predicate::PredicateKey;
+
+/// Checkpoint subprotocol state.
+#[derive(Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct CheckpointState {
+    /// Credential rule for sequencer signature verification.
+    /// Updated via `UpdateSequencerKey` message from admin subprotocol.
+    sequencer_cred: CredRule,
+
```
prajwolrg, line 14: I wonder if this should also be replaced with `PredicateKey`. Not sure if it'd add more confusion.

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// Predicate for checkpoint ZK proof verification.
+    /// Updated via `UpdateCheckpointPredicate` message from admin subprotocol.
+    checkpoint_predicate: PredicateKey,
+
+    /// Summary of the last verified checkpoint epoch.
+    /// `None` before the first checkpoint is verified.
+    verified_epoch_summary: Option<EpochSummary>,
```
prajwolrg: I think this should be some genesis value instead of being Option.

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    /// Last L1 commitment covered by checkpoints (seeded from genesis).
+    last_covered_l1: L1Commitment,
```
prajwolrg: I think we should be able to get away with just having height for the last covered L1. The OL should not submit L1BlockHash. It can be inferred from the L1ViewState though.

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -0,0 +1,146 @@
...
+            batch_info.epoch,
+            batch_info.l2_range.end, // terminal: this epoch's final L2 block
+            prev_terminal,           // prev_terminal: from current state
+            batch_info.l1_range.end, // new_l1: this epoch's final L1 block
+            transition.post_state_root, // final_state: post-execution state
+        );
+
+        self.verified_epoch_summary = Some(epoch_summary);
+        self.last_covered_l1 = batch_info.l1_range.end;
+
```
prajwolrg: EpochSummary already seems to have `l1_range.end`. Having `last_covered_l1` seems like a duplication.

prajwolrg (changes requested):
Mostly looked only on the `state`. Have a few comments.

crates/asm/txs/checkpoint/src/errors.rs
storopoli: FYI @irnb
Done in https://github.com/alpenlabs/ssz-gen/pull/53 it's now in a new tagged version `v0.11.1`

Cc @krsnapaudel

crates/asm/subprotocols/checkpoint/src/handler.rs
```diff
@@ -0,0 +1,216 @@
...
+    utils::{compute_manifest_hashes_commitment, get_manifest_hashes},
+};
+
+/// Process a checkpoint transaction.
+///
+/// Steps:
+/// 1. Extract signed checkpoint from envelope
+/// 2. Verify signature
+/// 3. Validate state transitions (epoch, L1/L2 progression)
+/// 4. Validate start values match expected state
```
barakshani: These are reversed below.

crates/asm/subprotocols/checkpoint/src/handler.rs
```diff
@@ -0,0 +1,216 @@
...
+    cred_rule: &CredRule,
+) -> bool {
+    match cred_rule {
+        CredRule::SchnorrKey(pk) => verify_schnorr_sig(signature, payload_hash, pk),
+        _ => false,
+    }
+}
+
+/// Validate that checkpoint start values match expected state.
+fn validate_start_values(state: &CheckpointState, batch_info: &BatchInfo) -> CheckpointResult<()> {
```
barakshani, line 92: I'm not sure what is the practical implication of the following, but it seems to me more correct to have this function based on checkpoint index rather than last_{l1,l2}. That is, I propose we check that checkpoint j+1 is consistent with checkpoint j.

crates/asm/subprotocols/checkpoint/src/handler.rs
```diff
@@ -0,0 +1,216 @@
...
+    Ok(())
+}
+
+/// Validate state transitions: epoch sequence, L1 height progression, L2 slot progression.
+fn validate_state_transitions(
+    state: &CheckpointState,
+    batch_info: &BatchInfo,
+) -> CheckpointResult<()> {
+    // Epoch must be sequential
+    let expected_epoch = state.expected_next_epoch();
```
barakshani, line 130: as above

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// Update state with a verified checkpoint.
+    ///
+    /// Called after the checkpoint signature and proof have been verified.
+    pub fn update_with_checkpoint(&mut self, checkpoint: &CheckpointPayload) {
+        let batch_info = &checkpoint.commitment.batch_info;
+        let transition = &checkpoint.commitment.transition;
+
+        // prev_terminal comes from current state (the terminal of the previous epoch).
+        // For the first checkpoint (epoch 0), this will be null/zero.
+        let prev_terminal = self
```
barakshani: this is really the previous state we had, right? In that case I won't use the word terminal -- it's reserved to L2 context (an L2 block) while ASM should only be aware of batches.

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -0,0 +1,146 @@
...
+            .verified_epoch_summary
+            .as_ref()
+            .map(|s| *s.terminal())
+            .unwrap_or_else(OLBlockCommitment::null);
+
+        let epoch_summary = EpochSummary::new(
+            batch_info.epoch,
+            batch_info.l2_range.end, // terminal: this epoch's final L2 block
+            prev_terminal,           // prev_terminal: from current state
```
barakshani, line 162: as above, if this is the state, let's just call it state (note also the prev/current conflict you have here)

barakshani (commented):
Overall looks good!

crates/asm/subprotocols/checkpoint/src/handler.rs
irnb: We don't store the raw `CheckpointPayload` in the checkpoint subprotocol state we extract `EpochSummary` (epoch, terminal, prev_terminal, l1_end, final_state) which is the minimal data needed to validate the next checkpoint. 

The epoch validation in `validate_state_transitions` (`expected_next_epoch`) already ensures checkpoint j+1 follows checkpoint j. The L1/L2 start validations in validate_start_values are additional defense in depth checks that ensure the ranges connect without gaps

crates/asm/subprotocols/checkpoint/src/state.rs
irnb: The `prev_terminal` variable holds the end block commitment of the previous epoch it's an `OLBlockCommitment` we read from the currently stored epoch_summary before updating. 
This is needed for `EpochSummary` to maintain chain continuity between epochs. 
You're right that "terminal" is L2-specific terminology. The local variable name prev_terminal matches the EpochSummary field name for clarity, but we could rename it to something more batch-oriented like `prev_epoch_end_block` if preferred.

crates/asm/subprotocols/checkpoint/src/state.rs
irnb: These are not state roots `prev_terminal` and `terminal` are L2 block commitments stored in `EpochSummary`. 

The flow in `update_with_checkpoint`:
We have the currently stored epoch_summary (from the previous valid checkpoint, e.g., epoch N-1)
When processing a new valid checkpoint (epoch N), we read `self.epoch_summary.terminal()`  this is epoch N-1's final L2 block we store it as `prev_terminal` in the new `EpochSummary` for epoch N
The rest of the data (terminal, l1_end, final_state) comes from the new checkpoint transaction.


So the comment "from current state" means read from the currently stored epoch summary before we overwrite it.

crates/asm/subprotocols/checkpoint/src/state.rs
barakshani: > we read from the currently stored epoch_summary

Ok, are you saying that the epoch summary (is this _the same_ as checkpoint? or is it that checkpoint is a tx that contains summery + proof?) contains a commitment (just block id?) to the terminal block? If so, why is this needed -- to prove L2 is a proper chain?

crates/checkpoint-types-ssz/ssz/payload.ssz
```diff
@@ -0,0 +1,103 @@
...
+OUTPUT_MSG_MAX_SIZE = 1 << 14
+
+### ZK proof max size (4 KiB)
+### SP1 Groth16 proofs are ~260 bytes, using 4 KiB for safety margin with metadata.
+### TODO: Validate this limit based on actual checkpoint proof format.
+MAX_PROOF_LEN = 1 << 12
+
+### L1 block commitment (height + block ID).
+###
+# NOTE: This type is defined inline rather than using strata_identifiers::L1BlockCommitment
```
krsnapaudel: What is the plan here? We should try to put L1BlockCommitment in one place, right?

storopoli:
> **Blocked On**
> Legacy code deprecation (discussed [here](https://alpen-labs.slack.com/archives/C08G7PL1RBK/p1765460301573599)). Once CSM is removed, this PR can merge. Fixing CSM to handle new log type is throwaway work since it's being deprecated. cc: @prajwolrg

Is this fixed by #1223 or #1225?

crates/asm/subprotocols/checkpoint/src/handler.rs
```diff
@@ -26,13 +28,14 @@ use crate::{
 /// Steps:
 /// 1. Extract signed checkpoint from envelope
 /// 2. Verify signature
```
barakshani:
```suggestion
/// 2. Verify signature using predicate framework.
```
for consistency

crates/asm/subprotocols/checkpoint/src/handler.rs
```diff
@@ -26,13 +28,14 @@ use crate::{
...
-/// 3. Validate state transitions (epoch, L1/L2 progression)
-/// 4. Validate start values match expected state
-/// 5. Get manifest hashes from auxiliary data
-/// 6. Construct claim and verify proof
-/// 7. Update state with verified checkpoint
-/// 8. Forward withdrawal intents to bridge
-/// 9. Emit checkpoint update log
+/// 3. Validate start values match expected state
+/// 4. Validate state transitions (epoch, L1/L2 progression)
+/// 5. Validate pre-state root matches expected state
```
barakshani: You added the pre-state root verification, but it's in the same step.
```suggestion
/// 4. Validate state transitions (epoch, L1/L2 progression, pre-state root)
```
change the following to start from 5.

crates/asm/subprotocols/checkpoint/src/handler.rs
```diff
@@ -164,6 +157,18 @@ fn validate_state_transitions(
         }
     }
 
+    // Pre-state root in signed payload must match expected state.
```
barakshani: This works, but here's a question: why do we even need the pre-state in the checkpoint payload?

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -2,42 +2,63 @@
...
-    /// `None` before the first checkpoint is verified.
-    verified_epoch_summary: Option<EpochSummary>,
-
-    /// Last L1 commitment covered by checkpoints (seeded from genesis).
-    last_covered_l1: L1Commitment,
+    /// Summary of the current epoch state.
+    ///
+    /// Before the first checkpoint is verified, this contains a genesis summary with:
+    /// - `epoch` = 0 (first expected epoch)
+    /// - `terminal` = null (sentinel indicating no epoch verified yet)
```
barakshani, line 23: EpochSummary also has a field prev_terminal -- what would that be?

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -2,42 +2,63 @@
...
+    /// - `l1_end` = genesis L1 commitment
```
barakshani, line 24: I don't see l1_end, but new_l1.

crates/asm/subprotocols/checkpoint/src/state.rs
barakshani, line 23: Can you explain what you mean by that? Since this epoch isn't sealed yet, `terminal` (as well as the field `final_state`) should be "empty", but I'm not sure why this would indicate no epoch verified yet (unless combined with epoch number).

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -2,42 +2,63 @@
...
 
-    /// Returns the sequencer credential rule.
-    pub fn sequencer_cred(&self) -> &CredRule {
-        &self.sequencer_cred
+    /// Returns true if no checkpoint has been verified yet (genesis state).
+    ///
+    /// This is determined by checking if the terminal block commitment is null.
+    pub fn is_genesis(&self) -> bool {
+        Buf32::from(*self.epoch_summary.terminal().blkid()).is_zero()
+    }
```
barakshani: see my comment above -- is it that we won't create a new `EpochSummary` for epoch `k+1` when sealing epoch `k` (but before `k+1` is sealed)? It make sense because we don't have a summary yet, however would there be a different struct that represent the current (unsealed) epoch?

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -46,21 +67,31 @@ impl CheckpointState {
...
+    ///
+    /// Returns `None` if no checkpoint has been verified yet (genesis state).
     pub fn verified_epoch_summary(&self) -> Option<&EpochSummary> {
-        self.verified_epoch_summary.as_ref()
+        if self.is_genesis() {
+            None
+        } else {
+            Some(&self.epoch_summary)
+        }
     }
```
barakshani, line 84: Can you provide more info on this function?
I'd assume that we store "locally" the (non-verified) EpochSummary and separately, after reading it from L1 store an indicator that the EpochSummary is verified (if valid), but I dee that now we no longer have it in `CheckpointState`.

Do I understand correctly that because it's part of the checkpoint subprotocol, we only get access to the checkpoint, here, after it's on L1?

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -46,21 +67,31 @@ impl CheckpointState {
...
     /// Get the expected next epoch number.
     pub fn expected_next_epoch(&self) -> Epoch {
-        self.verified_epoch_summary
-            .as_ref()
-            .map(|s| s.epoch() + 1)
-            .unwrap_or(0)
+        if self.is_genesis() {
+            // Genesis: epoch field holds the first expected epoch
+            self.epoch_summary.epoch()
+        } else {
```
barakshani, line 91: Is this better than setting the hardcoded genesis 'epoch' to `-1` or MAX value?

crates/checkpoint-types-ssz/ssz/payload.ssz
irnb: Agreed! We should consolidate `L1BlockCommitment` in one place. However, when we try to use s`trata_identifiers::L1BlockCommitment` with `external_kind: container`, we get the following type mismatch errors:

```

  error[E0308]: mismatched types
     --> .../out/generated.rs:605:28
      |
  605 |                     start: self.start().expect("valid view").to_owned(),
      |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `strata_identifiers::L1BlockCommitment`, found `L1BlockCommitment`

  error[E0308]: mismatched types
     --> .../out/generated.rs:606:26
      |
  606 |                     end: self.end().expect("valid view").to_owned(),
      |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `strata_identifiers::L1BlockCommitment`, found `L1BlockCommitment`

  error[E0308]: mismatched types
      --> .../out/generated.rs:2350:29
       |
  2350 |                     l1_end: self.l1_end().expect("valid view").to_owned(),
       |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `strata_identifiers::L1BlockCommitment`, found `L1BlockCommitment`
```

 **Root cause:** ssz_codegen generates code that calls the inherent `.to_owned()` method on the SSZ-generated `L1BlockCommitmentRef`, which returns the SSZ-generated `L1BlockCommitment` (with height: u32). But the expected type is `strata_identifiers::L1BlockCommitment` (with height: `absolute::Height` when bitcoin feature is enabled).

Fix: ssz_codegen needs to be updated to use the `ToOwnedSsz::to_owned()` trait method instead of the inherent method. This would allow proper type conversion via the `From impl` in strata_identifiers. Once that's fixed, we can replace the inline `L1Commitment` with `strata_identifiers.L1BlockCommitment`.

crates/asm/txs/checkpoint/src/errors.rs
irnb: Thanks @storopoli! Waiting on #1260 (MagicBytes/MMR type migration) to merge first, then I'll adopt #[from] for DecodeError in this PR.

crates/asm/subprotocols/checkpoint/src/state.rs
irnb: `prev_terminal` stores the terminal L2 block commitment from the **previous** epoch. For genesis (epoch 0), it's null/zero. For epoch N (N > 0), `prev_terminal` = epoch N-1's terminal. This enables chain continuity verification.

In genesis state, `epoch_summary` holds **expected** next epoch's context (not a verified epoch):
- `epoch = 0`: First expected epoch
- `terminal = null`: Sentinel indicating no epoch verified
- `l1_end = genesis_l1`: Starting point
- `final_state = genesis_state`: Pre-state for first checkpoint

After verification, `epoch_summary` represents the **last verified** epoch. The `is_genesis()` check uses `terminal.blkid().is_zero()` as sentinel.

crates/asm/subprotocols/checkpoint/src/state.rs
irnb: Correct; we don't create a new EpochSummary until the checkpoint for that epoch is verified.

  There's no separate struct for the "current unsealed epoch." Instead, we derive the expectations on-demand via accessor methods:
  - `expected_next_epoch() `→ last_epoch + 1
  - `last_covered_l1_height() + 1` → expected L1 start
  - `last_l2_terminal_slot() + 1` → expected L2 start
  - `pre_state_root()` → expected pre-state (equals last epoch's post-state)
  
  This avoids storing speculative data that could become stale.

crates/asm/subprotocols/checkpoint/src/state.rs
irnb: Correct! The checkpoint subprotocol operates **reactively** it only processes checkpoints after they appear on L1. 

Flow:
1. Sequencer builds `CheckpointPayload` locally
2. Sequencer signs and posts to L1
3. ASM scans L1, finds checkpoint tx
4. Checkpoint subprotocol validates (signature, state transitions, ZK proof)
5. If valid, updates `epoch_summary` with verified data

crates/asm/subprotocols/checkpoint/src/state.rs
irnb: Since we already store the epoch at genesis and have sentinel logic for detecting genesis (via null terminal), we can reuse that here. Cleaner than introducing a separate magic value for epoch.

crates/asm/subprotocols/checkpoint/src/state.rs
irnb: EpochSummary (what we store) is derived from `checkpoint tx + checkpoint subprotocol state` and just the fields needed for validating the next checkpoint.

 The terminal commitment (OLBlockCommitment = slot + blkid) is used exactly for what you said - proving L2 chain continuity. In validate_start_values, we check that the new checkpoint's l2_range.start.slot() == last_terminal.slot() + 1. This ensures epochs link together without gaps or overlaps.

crates/checkpoint-types-ssz/ssz/payload.ssz
irnb: https://github.com/alpenlabs/ssz-gen/pull/58

crates/asm/subprotocols/checkpoint/src/handler.rs
```diff
@@ -0,0 +1,682 @@
...
+    // Epoch must be sequential
+    let expected_epoch = state.expected_next_epoch();
+    if batch_info.epoch != expected_epoch {
+        return Err(CheckpointError::InvalidEpoch {
+            expected: expected_epoch,
+            actual: batch_info.epoch,
+        });
+    }
+
+    // L1 end height must progress beyond last covered L1
```
barakshani, line 138: why is this a requirement? What if we seal 2 batches, say because of DA limits, in a short time -- say in 5 minutes, but more generally, the second batch is sealed before the first batch gets recorded on chain? In this case the two batches will have the same L1 block height as _last_covered_. 
If we follow this check, I think that it effectively means that we'll need to stop tx inclusion (maybe even block production) on the OL.

crates/asm/subprotocols/checkpoint/src/handler.rs
```diff
@@ -0,0 +1,682 @@
...
+    let input_msgs_commitment = compute_manifest_hashes_commitment(&manifest_hashes)?;
+    let claim =
+        CheckpointClaim::from_payload(checkpoint_payload, pre_state_root, input_msgs_commitment);
+
+    state
+        .checkpoint_predicate()
+        .verify_claim_witness(&claim.to_bytes(), &checkpoint_payload.proof)?;
+
+    // 7. Update state with verified checkpoint
+    state.update_with_checkpoint(checkpoint_payload);
```
barakshani, line 76: just double checking: if we have multiple checkpoints in the same bitcoin block (ordered correctly), we will process the first, then update the state, and _only then_ start process the next checkpoint? So the states values will be updated and the verifications in steps 4,5 will pass for the second checkpoint?


what happens if they are not ordered correctly, say checkpoint `n+1` appears before checkpoint `n`? we will discard `n+1` and only process `n`? (if so, we may want to chain txns in this case, though it's not related to the work in this PR).

crates/asm/subprotocols/checkpoint/src/handler.rs
```diff
@@ -0,0 +1,682 @@
...
+        assert!(relayer.logs.is_empty());
+    }
+
+    #[test]
+    fn test_invalid_epoch_sequence() {
+        let keypair = SequencerKeypair::random();
+        let genesis_l1 = genesis_l1();
+        let mut generator = CheckpointGenerator::new(genesis_l1);
+        let mut payload = generator.gen_payload(1, 1, vec![]);
+        payload.commitment.batch_info.epoch = 1; // state expects 0
```
barakshani, line 368: isn't epoch 0 transition is special, so this only check some part of the logic?

crates/checkpoint-types-ssz/src/claim.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use ssz::Encode;
+use strata_identifiers::{Buf32, Epoch};
+
+use crate::{
+    BatchTransition, CheckpointPayload, L1BlockRange, L2BlockRange,
+    ssz_generated::ssz::claim::CheckpointClaim,
+};
+
+impl CheckpointClaim {
```
barakshani, line 11: what do you mean by "claim"? that it is not verified yet?

crates/sequencer/src/checkpoint/helper.rs
```diff
@@ -1,12 +1,151 @@
...
+
+/// Verify signature on a SignedCheckpointPayload.
+pub fn verify_signed_checkpoint_payload(
+    signed_payload: &SignedCheckpointPayload,
+    params: &Params,
+) -> bool {
+    verify_checkpoint_payload_sig(&signed_payload.inner, &signed_payload.signature, params)
+}
+
+/// Converts an `Checkpoint-v0` to the SSZ `CheckpointPayload` format.
```
barakshani: v0?

crates/sequencer/src/checkpoint/helper.rs
```diff
@@ -1,12 +1,151 @@
...
+    // Get proof bytes
+    let proof_bytes = checkpoint.proof().as_bytes().to_vec();
+
+    CheckpointPayload::new(commitment, sidecar, proof_bytes)
+        .expect("checkpoint payload construction should succeed")
+}
+
+/// Extracts withdrawal intents from chainstate and converts them to SSZ-encoded OL logs.
+///
+/// The checkpoint-v0 format stores a serialized `Chainstate` in the sidecar, which contains
```
barakshani: what do we need this for?

crates/asm/subprotocols/checkpoint/src/state.rs
barakshani: is this field called `terminal` or `prev_terminal`?

irnb: EpochSummary has both fields:
 - `terminal`: Last L2 block of this epoch -> null in genesis as a sentinel for "no epoch verified"
 - `prev_terminal`: Last L2 block of previous epoch -> null in genesis because there's no previous epoch


 Sorry for the confusion in my previous reply I started by explaining `prev_terminal` but the comment and sentinel logic are about terminal.
 
I'll update the comment to clarify this.

crates/asm/subprotocols/checkpoint/src/handler.rs
irnb: The L1 progression requirement is correct for the current **checkpoint-driven** architecture.

Current flow: Epoch `N` can only seal when Epoch `N-1's` checkpoint is observed on L1 (see `has_expected_checkpoint()` in `block_assembly.rs:259-299`). 

This means:

  1. Epoch `N-1` completes → checkpoint posted to L1 in block B
  2. OL scans L1, sees checkpoint ack in block B → seals Epoch N with l1_end ≥ B
  3. Since B > Epoch N-1's l1_end, progression is guaranteed

The scenario you describe (**DA-driven batching**) would require a different architecture where batches seal independently of checkpoint confirmation. The current design intentionally ties epoch sealing to checkpoint acknowledgment.

If we wanted DA-driven batching, we'd need to revisit both the OL sealing logic and this ASM validation. Is that a requirement we should support?

cc: @storopoli

crates/checkpoint-types-ssz/ssz/claim.ssz
```diff
@@ -0,0 +1,23 @@
...
+### Checkpoint claim used as public input for proof verification.
+### Reconstructed from checkpoint payload, local state and aux data for verification.
+#~# derive: Serialize, Deserialize
+class CheckpointClaim(Container):
+    ### Epoch number.
+    epoch: strata_identifiers.Epoch
+    ### L1 block range.
+    l1_range: payload.L1BlockRange
+    ### L2 block range.
+    l2_range: payload.L2BlockRange
```
prajwolrg, line 13: nit: This seems to be encapsulated by `BatchInfo`.

crates/checkpoint-types-ssz/ssz/claim.ssz
```diff
@@ -0,0 +1,23 @@
...
+    ### State transition (pre/post state roots).
+    ### Pre comes from local state, post from payload.
+    transition: payload.BatchTransition
```
prajwolrg, line 16: I wonder if we really need `BatchTransition` separately. Since we have both the `L1BlockRange` and `L2BlockRange`, is BatchTransition necessary?

crates/asm/subprotocols/checkpoint/src/handler.rs
irnb: Yes, the code handles this correctly each checkpoint is processed individually via the loop in `subprotocol.rs:68-95`, with `state.update_with_checkpoint()` called immediately after each successful validation (line 76).

However, multiple checkpoints in the same Bitcoin block shouldn't happen in practice due to the checkpoint-driven architecture.

This means checkpoint N+1 would always be in a later Bitcoin block than checkpoint N.

**Out-of-order checkpoints**

Correct checkpoint n+1 would fail with `InvalidEpoch` and be discarded. 
But given the architecture above, this scenario shouldn't occur: the sequencer should not produce checkpoint n+1 until it observes checkpoint n's confirmation on L1.

crates/checkpoint-types-ssz/ssz/payload.ssz
```diff
@@ -0,0 +1,88 @@
...
+    pre_state_root: strata_identifiers.Buf32
+    ### Chainstate root after batch execution.
+    post_state_root: strata_identifiers.Buf32
+
+### Core commitment data combining batch metadata and post-state.
+### Pre-state is not included as it's derived from ASM state during verification.
+#~# derive: Copy, Serialize, Deserialize
+class CheckpointCommitment(Container):
+    ### Batch metadata (epoch, L1/L2 ranges).
+    batch_info: BatchInfo
```
prajwolrg, line 59: `BatchInfo` has the pre-state as well (both the L1 prestate and L2 prestate). We should be able to getaway without having duplicate information that is already in state.

crates/asm/subprotocols/checkpoint/src/handler.rs
irnb: Thanks, good catch! 
The test only covers the genesis case (`is_genesis() = true` branch of `expected_next_epoch()`).

Added `test_invalid_epoch_after_first_checkpoint`

crates/checkpoint-types-ssz/src/claim.rs
irnb: `CheckpointPayload`  -> the pruned checkpoint committed on L1
`CheckpointClaim` -> the full checkpoint, reconstructed from:
- the pruned payload (CheckpointPayload),
- checkpoint subprotocol state, and
- auxiliary data

`CheckpointClaim` would then serve as the public parameters for proof verification.

The term "claim" is aligned with the terminology from the predicate framework.

crates/checkpoint-types-ssz/Cargo.toml
prajwolrg: I'm slightly in favor of separating ssz deps, instead of having it in the same block as `strata-*`/`alpen-*`. It just looks nice together. 

Since we have all the same deps everywhere (where ssz-gen is used), does it make sense to instead combine into a same crate and expose it from there?

crates/sequencer/src/checkpoint/helper.rs
irnb: The existing sequencer produces Checkpoint which stores raw `chainstate` (with `pending_withdraws` queue). The new checkpoint subprotocol expects withdrawal intents as SSZ-encoded OL logs in the sidecar. This extracts and converts between the two formats.

 We need this to pass the existing functional tests.

 Once we migrate the sequencer to produce CheckpointPayload directly, this helper can be removed.

crates/checkpoint-types-ssz/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "strata-checkpoint-types-ssz"
```
krsnapaudel, line 2: Should this be called "strata-checkpoint-types-new", following other examples. Not a strong preference.

crates/asm/common/src/aux/provider.rs
```diff
@@ -115,7 +115,8 @@ impl VerifiedAuxData {
                     hash: *item.hash(),
                 });
             }
-            manifest_hashes.insert(item.proof().index(), *item.hash());
+            // Index by L1 block height, not MMR index, since consumers query by height
+            manifest_hashes.insert(item.height(), *item.hash());
```
prajwolrg, line 119: This is wrong. The height is not bounded in the verification. We can prove any log was emitted at any height in this way.

One way to solve this is when we want to fetch the hash at the height, ASM converts that to proper index and query on the index.

crates/asm/subprotocols/checkpoint/src/handler.rs
prajwolrg: Related: https://github.com/alpenlabs/alpen/pull/1209/changes#r2671718534.

We don't strictly need to have starting information in the batch info. Having it helps in debugging, but is not required for security purposes.

crates/asm/subprotocols/checkpoint/src/state.rs
prajwolrg: I don't think we need both fields now. It just adds confusion and doesn't serve any useful purpose anymore. We can treat that the genesis epoch is the verified one.

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -0,0 +1,186 @@
...
+
+impl CheckpointState {
+    /// Create initial state from configuration.
+    ///
+    /// Initializes with a genesis epoch summary where:
+    /// - `epoch` = 0 (first expected epoch)
+    /// - `terminal` = null (indicates no epoch has been verified yet)
+    /// - `previous_terminal` = null (no previous terminal for genesis)
+    /// - `l1_end` = genesis L1 commitment from config
+    /// - `final_state` = genesis OL state root (pre-state for first checkpoint)
```
prajwolrg, line 38: I can't think of a reason to have a separate `final_state` field.

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    /// Returns the checkpoint predicate for proof verification.
+    pub fn checkpoint_predicate(&self) -> &PredicateKey {
+        &self.checkpoint_predicate
+    }
+
+    /// Returns the verified epoch summary, if any.
+    ///
+    /// Returns `None` if no checkpoint has been verified yet (genesis state).
+    pub fn verified_epoch_summary(&self) -> Option<&EpochSummary> {
```
prajwolrg, line 78: The genesis epoch should be the verified epoch summary in case of no checkpoint.

crates/asm/txs/checkpoint/src/test_utils/mmr.rs
```diff
@@ -0,0 +1,400 @@
...
+//! Provides a full MMR implementation that stores all nodes in memory,
+//! enabling proof generation for unit tests without a database backend.
+//!
+//! ## Production Semantics
+//!
+//! This module models the production MMR semantics where:
+//! - Genesis block has **NO entry** in the MMR (it's a boundary, not a processable block)
+//! - The first MMR leaf (index 0) contains the manifest hash for `genesis_height + 1`
+//! - Formula: `mmr_index = l1_height - genesis_height - 1`
+//! - Reverse: `l1_height = genesis_height + mmr_index + 1`
```
prajwolrg, line 12: This should be part of the main logic itself. 

Related: https://github.com/alpenlabs/alpen/pull/1209/changes#r2675427474

crates/asm/txs/checkpoint/src/test_utils/mmr.rs
```diff
@@ -0,0 +1,400 @@
...
+
+impl TestMmr {
+    /// Creates a new empty test MMR.
+    pub fn new() -> Self {
+        Self {
+            nodes: HashMap::new(),
+            metadata: MmrMetadata::empty(),
+            standard_mmr: AsmMmr::new(ASM_MMR_CAP_LOG2),
+        }
+    }
```
prajwolrg, line 54: This should be defined somewhere else. Somewhere close to the MMR implementation.

crates/sequencer/src/duty/extractor.rs
```diff
@@ -85,7 +86,11 @@ async fn extract_batch_duties(
 
         // Need to wait for a proof.  Also avoid generating a duty if it's already in the pipe
         if publish_ready {
-            let duty = CheckpointDuty::new(ckpt.into());
+            let old_checkpoint: Checkpoint = ckpt.into();
+            // CheckpointDuty only carries the old checkpoint format for RPC
+            // serialization. The SSZ CheckpointPayload is reconstructed on-demand using
+            // convert_checkpoint_to_payload().
+            let duty = CheckpointDuty::new(old_checkpoint);
```
prajwolrg, line 93: We need to do the follow up task to handle this properly i.e. to remove the old checkpoint completely.

crates/checkpoint-types-ssz/ssz/claim.ssz
irnb: - Block ranges (L1BlockRange, L2BlockRange)
 - BatchTransition: Contains state roots

crates/checkpoint-types-ssz/ssz/payload.ssz
irnb: I think there might be a terminology confusion here. `BatchInfo` doesn't contain pre-state.
 
The existing comment on CheckpointCommitment is correct: pre-state is omitted because ASM already knows it from its own state tracking. Only `post_state_root` needs to be in the on-chain payload.

crates/checkpoint-types-ssz/Cargo.toml
irnb: > Why do we need the conversion between ssz checkpoint types and existing types?

https://github.com/alpenlabs/alpen/pull/1209#issuecomment-3681208449

crates/asm/subprotocols/checkpoint/src/handler.rs
barakshani: > Is that a requirement we should support?

Eventually yes: https://alpenlabs.atlassian.net/browse/STR-1958

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -0,0 +1,186 @@
...
+    epoch_summary: EpochSummary,
+}
+
+impl CheckpointState {
+    /// Create initial state from configuration.
+    ///
+    /// Initializes with a genesis epoch summary where:
+    /// - `epoch` = 0 (first expected epoch)
+    /// - `terminal` = null (indicates no epoch has been verified yet)
+    /// - `previous_terminal` = null (no previous terminal for genesis)
```
barakshani, line 36: For non-genesis epochs, what's the difference between these fields?

crates/asm/subprotocols/checkpoint/src/error.rs
```diff
@@ -0,0 +1,69 @@
...
+use strata_predicate::PredicateError;
+use thiserror::Error;
+
+/// Result type for checkpoint subprotocol operations.
+pub(crate) type CheckpointResult<T> = Result<T, CheckpointError>;
+
+/// Errors that can occur during checkpoint processing.
+#[derive(Debug, Error)]
+pub(crate) enum CheckpointError {
+    /// Failed to parse checkpoint transaction.
```
delbonis, line 16: Stylistic comments:

* `error` is redundant in error messages in all cases
* elsewhere we've been putting the expected/actual fields in parens to avoid suggesting that there's a new level to the error hierarchy (ie. `invalid epoch (expected {expected}, got {got})`)

crates/asm/subprotocols/checkpoint/src/error.rs
```diff
@@ -0,0 +1,69 @@
...
+    #[error("L1 height did not progress: previous {previous}, new {new}")]
+    InvalidL1Progression { previous: u32, new: u32 },
+
+    /// L2 slot did not progress correctly.
+    #[error("L2 slot did not progress: previous {previous}, new {new}")]
+    InvalidL2Progression { previous: u64, new: u64 },
+
+    /// Failed to retrieve manifest hashes from auxiliary data.
+    #[error("auxiliary data error: {0}")]
+    AuxData(#[from] AuxError),
```
delbonis, line 56: This is a bit of an obtuse error variant, it's not clear what's really meant here without extra context.  I think this might be better described by saying "insufficient aux data provided" or something.

crates/asm/subprotocols/checkpoint/src/state.rs
delbonis: I don't think we need the full `EpochSummary`, no.  That was originally meant primarily as a database bookkeeping structure.  We can probably get away with just the epoch index, the last epoch's terminal, and the last updated L1 view block info (to do the manifests checks for the next checkpoint).

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// - `l1_end` = genesis L1 commitment from config
+    /// - `final_state` = genesis OL state root (pre-state for first checkpoint)
+    pub fn new(config: &CheckpointConfig) -> Self {
```
delbonis, line 39: This isn't like a more typical one-for-one constructor, so `new` is maybe a misleading name.  I would probably call this `from_config`.

crates/asm/subprotocols/checkpoint/src/subprotocol.rs
delbonis, line 1: Put `use logging::*` at the top instead of doing `logging::foo` everywhere.

Also in both branches, we use the txid in the log, so we can compute that upfront instead of in the log expression so we can just do `%txid`.

crates/asm/subprotocols/checkpoint/src/subprotocol.rs
```diff
@@ -0,0 +1,113 @@
...
+                match extract_signed_checkpoint_from_envelope(tx) {
+                    Ok(signed_checkpoint) => {
+                        let batch_info = &signed_checkpoint.inner.commitment.batch_info;
+                        let (start_height, end_height) = get_manifest_hash_range(state, batch_info);
+                        collector.request_manifest_hashes(start_height, end_height);
+                    }
+                    Err(e) => {
+                        logging::warn!(
+                            txid = ?tx.tx().compute_txid(),
+                            error = ?e,
```
delbonis, line 59: `err = %e`

crates/asm/subprotocols/checkpoint/src/subprotocol.rs
```diff
@@ -0,0 +1,113 @@
...
+                    Ok(()) => {
+                        logging::info!(
+                            txid = %tx.tx().compute_txid(),
+                            "Successfully processed checkpoint transaction"
+                        );
+                    }
+                    Err(e) => {
+                        logging::error!(
+                            txid = %tx.tx().compute_txid(),
+                            error = %e,
```
delbonis, line 88: same

crates/asm/subprotocols/checkpoint/src/subprotocol.rs
```diff
@@ -0,0 +1,113 @@
...
+        _params: &Self::Params,
+    ) {
+        for tx in txs {
+            if tx.tag().tx_type() == OL_STF_CHECKPOINT_TX_TYPE {
+                match handle_checkpoint_tx(state, tx, verified_aux_data, relayer) {
+                    Ok(()) => {
+                        logging::info!(
+                            txid = %tx.tx().compute_txid(),
+                            "Successfully processed checkpoint transaction"
+                        );
```
delbonis, line 83: Should just be `debug` or maybe even just `trace`.

crates/asm/subprotocols/checkpoint/src/subprotocol.rs
```diff
@@ -0,0 +1,113 @@
...
+        // ASM design assumes subprotocols are not adversarial against each other,
+        // so no additional validation is performed on incoming messages.
+        for msg in msgs {
+            match msg {
+                CheckpointIncomingMsg::UpdateSequencerKey(new_key) => {
+                    logging::info!(%new_key, "Updating sequencer predicate");
+                    state.update_sequencer_predicate(new_key.as_ref());
+                }
+                CheckpointIncomingMsg::UpdateCheckpointPredicate(new_predicate) => {
+                    logging::info!("Updating checkpoint predicate");
```
delbonis, line 107: Should just be debug.

delbonis (changes requested):
I'm worried a little bit about the error handling here, but I am not sure if I'm identifying this correctly.

I *think* we *might* be conflating errors relating to an invalid checkpoint being seen with errors relating to insufficient aux data.  If the aux data is insufficient then we MUST fail the state transition, but invalid checkpoints are normal and we MUST NOT fail the state transition.  I'm not sure if I'm reading all that code properly because there's kinda a lot there, but I wanted to raise it to make sure that we are doing it properly here.

crates/asm/common/src/aux/provider.rs
prajwolrg: Handled this in https://github.com/alpenlabs/alpen/pull/1284.

PR #1211, opened 2025-12-15
update deposit request and deposit transaction parsing and validation

prajwolrg:
## Description

This PR updates the parsing and validation of Deposit Request Transaction (DRT) and Deposit Transaction (DT). It also makes a lot of refactors to simplify auxiliary data handling, extract validation logic into separate modules, and improve test utilities for bridge transaction flows.

### Refactoring
- **Validation module**: Created `validation/` directory with dedicated modules for validating specific transaction.
- **Improved test utilities**: Enhanced bridge transaction test helpers with better ergonomics, documentation, and removed unwraps
- **Better error handling**: Unified error types and improved error messages across bridge transaction parsing
- **Code cleanup**: Removed unused functions, improved naming clarity, and added comprehensive documentation

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [x] Dependency Update

## Notes to Reviewers

I've intentionally used `[u8;20]` for the EE address. This will be replaced with proper destination address based on [SPS-deposit-spec](https://www.notion.so/SPS-deposit-desc-295901ba000f80fd8175c7d0a49ef43e) as part of [STR-1950](https://alpenlabs.atlassian.net/browse/STR-1950).

I used Codex for some of the refactors and docstring.

## Checklist

## Related Issues

[STR-1950]: https://alpenlabs.atlassian.net/browse/STR-1950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/asm/txs/bridge-v1/src/constants.rs
+++ b/crates/asm/txs/bridge-v1/src/constants.rs
@@ -64,8 +64,9 @@ impl std::fmt::Display for BridgeTxType {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
     use proptest::prelude::*;
+
+    use super::*;
 
     impl Arbitrary for BridgeTxType {
         type Parameters = ();
--- a/bin/strata-test-cli/src/bridge/dt.rs
+++ b/bin/strata-test-cli/src/bridge/dt.rs
@@ -68,8 +68,8 @@ pub(crate) fn create_deposit_transaction_cli(
     .ok_or_else(|| Error::TxBuilder("Missing takeback script merkle root".to_string()))?;
 
     // Use canonical OP_RETURN construction from asm/txs/bridge-v1
-    let dt_tag = DepositTxHeaderAux::new(dt_index, *drt_data.header_aux().ee_address())
-        .build_tag_data();
+    let dt_tag =
+        DepositTxHeaderAux::new(dt_index, *drt_data.header_aux().ee_address()).build_tag_data();
     let sps50_script = ParseConfig::new(*MAGIC_BYTES)
         .encode_script_buf(&dt_tag.as_ref())
         .map_err(|e| Error::TxBuilder(e.to_string()))?;


crates/asm/txs/bridge-v1/src/deposit_request/lock.rs
```diff
@@ -0,0 +1,60 @@
...
+///    Transaction by spending this output.
+///
+/// 2. **Recovery path (tapscript)**: Allows the depositor to reclaim their funds after a timeout
+///    period if the bridge operators fail to process the deposit.
+///
+/// # Parameters
+///
+/// * `recovery_pk` - The depositor's x-only public key (32 bytes) for the recovery path. The
+///   depositor can use the corresponding secret key to reclaim funds after the timeout if deposit
+///   transaction is not created.
```
storopoli, line 23: ".... and the deposit request transaction is not spent."

crates/asm/txs/bridge-v1/src/deposit_request/lock.rs
```diff
@@ -0,0 +1,60 @@
...
+///
+/// # Panics
+///
+/// Panics if taproot finalization fails, which should not occur with valid inputs.
+pub fn create_deposit_request_locking_script(
+    recovery_pk: &[u8; 32],
+    internal_key: XOnlyPublicKey,
+    recovery_delay: u32,
+) -> ScriptBuf {
+    let secp = Secp256k1::new();
```
storopoli: I think we can get away using the [global static `SECP256K1`](https://docs.rs/secp256k1/latest/secp256k1/static.SECP256K1.html) which should be more efficient given that is instantiated once per the lifetime of the program.

crates/asm/txs/bridge-v1/src/deposit_request/lock.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    let tapscript = Builder::new()
+        .push_slice(recovery_pk)
+        .push_opcode(OP_CHECKSIGVERIFY)
+        .push_int(recovery_delay as i64)
+        .push_opcode(OP_CSV)
```
storopoli, line 58: I think you need to add `OP_DROP` here no?
Ref: https://github.com/rust-bitcoin/rust-bitcoin/issues/1530#issuecomment-1373841266

And I've done this is the past in a personal project of a Bitcoin Escrow service: https://github.com/storopoli/scrow/blob/d9d97787fc895db726dbf7661b8d23633a759c82/src/scripts.rs#L181-L195

crates/asm/txs/bridge-v1/src/test_utils/deposit.rs
```diff
@@ -137,3 +143,48 @@ pub fn build_deposit_transaction(
...
+        submit_transaction_with_keys_blocking(&bitcoind, &client, operator_keys, &mut drt).unwrap();
+
+    let mut dt = create_test_deposit_tx_new(dt_header_aux, nn_address);
+    dt.input[0].previous_output = OutPoint {
+        txid: drt_txid,
+        vout: 1,
+    };
+
+    let _ =
+        submit_transaction_with_keys_blocking(&bitcoind, &client, operator_keys, &mut dt).unwrap();
```
storopoli: I know this is a testing library/module but at least replace the `unwrap`s with `expect`s?

crates/asm/txs/bridge-v1/src/deposit_request/lock.rs
storopoli: More context here.

You do need `OP_DROP` (see https://github.com/bitcoin/bips/blob/master/bip-0112.mediawiki and Ctrl+F for `DROP`)
Because `OP_CSV` checks the timelock and then pushes 1 in the stack or fails the script.

bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -124,28 +125,31 @@ pub async fn deposit(
...
         .try_into()
         .expect("magic_bytes validated to be 4 bytes");
-    let op_return_data = drt_metadata
-        .op_return_data(magic_bytes)
-        .internal_error("Failed to generate DRT metadata")?;
-    let push_bytes = PushBytesBuf::try_from(op_return_data)
-        .expect("conversion should succeed after length check");
+
+    // Construct the DRT metadata using the canonical builder
+    let drt_sps50_tag = DrtHeaderAux::new(
```
barakshani: Are you sure _DRT sps50 tag_ is the right wording?

bin/strata-test-cli/src/bridge/withdrawal.rs
```diff
@@ -59,7 +60,12 @@ fn create_withdrawal_fulfillment_inner(
     let amount = Amount::from_sat(amount);
 
     // Create withdrawal metadata
```
barakshani:
```suggestion
    // Create withdrawal fulfillment metadata

File: crates/asm/subprotocols/bridge-v1/src/handler.rs [hunk 7]
```diff
@@ -15,16 +23,51 @@ use crate::{SlashValidationError, errors::BridgeSubprotocolError, state::BridgeV
...
+            // Verify the DRT output script matches what we expect
+            if actual_script != &expected_drt_script {
+                return Err(DepositValidationError::DrtOutputScriptMismatch(Mismatch {
+                    expected: expected_drt_script,
+                    got: actual_script.clone(),
+                })
+                .into());
+            }
+
+            // Verify the deposit amount matches the bridge's expected amount
```
barakshani: if we fail here then it must be that the bridge operators didn't do their job correctly, right? I mean, the previous checks already confirm that the funds in the bridge were processed by the bridge operators and nobody else.

crates/asm/subprotocols/bridge-v1/src/handler.rs
```diff
@@ -196,4 +243,44 @@ mod tests {
...
+
+        let (drt, dt) = create_connected_drt_and_dt(
+            drt_aux,
+            dt_aux,
+            (*state.denomination()).into(),
+            &operators,
+        );
+
+        // 3. Prepare ParsedTx
+        // We need to re-parse the slash tx to get the correct SlashInfo with updated input
```
barakshani: slash?

crates/asm/subprotocols/bridge-v1/src/handler.rs
```diff
@@ -196,4 +243,44 @@ mod tests {
...
+        // (create_connected_stake_and_slash_txs updates the input to point to stake_tx)
+        let dt_input = parse_sps50_tx(&dt);
+        let parsed_dt = parse_deposit_tx(&dt_input).expect("Should parse slash tx");
```
barakshani:
```suggestion
        let parsed_dt = parse_deposit_tx(&dt_input).expect("Should parse deposit tx");

File: crates/asm/subprotocols/bridge-v1/src/handler.rs [hunk 10]
```diff
@@ -196,4 +243,44 @@ mod tests {
...
+        let parsed_tx = ParsedTx::Deposit(parsed_dt);
+
+        // 4. Prepare VerifiedAuxData containing the stake transaction
```
barakshani: stake?

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -199,11 +199,6 @@ impl AssignmentEntry {
         self.deposit_entry.idx()
     }
 
-    /// Returns the deposit txid associated with this assignment.
-    pub fn deposit_txid(&self) -> BitcoinTxid {
-        self.deposit_entry.output().outpoint().txid.into()
-    }
-
```
barakshani, line 206: Why removing this?

crates/asm/txs/bridge-v1/src/constants.rs
```diff
@@ -1,14 +1,63 @@
...
-pub const UNSTAKE_TX_TYPE: TxType = 5;
+/// Bridge V1 transaction types.
+///
+/// This enum represents all valid transaction types for the Bridge V1 subprotocol.
+/// Each variant corresponds to a specific transaction type with its associated u8 value.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[repr(u8)]
+pub enum BridgeTxType {
+    /// Deposit request transaction - user initiates a deposit
+    DepositRequest = 0,
```
barakshani, line 17: not used, I think

crates/asm/txs/bridge-v1/src/deposit_request/lock.rs
prajwolrg: I think this is intentionally not using `OP_DROP`. After the execution of the script, we need 1 in the stack for it to be valid. If we drop it, it will be invalid. Since this is the last executed opcode, it's alright to keep 1 in the stack. CC: @Rajil1213 for confirmation.

storopoli: Ok you're right...
I had to take a look at both BIP-342 (taproot script spending rules) and BIP-141 (segwit v0 script spending rules).
And yes, 342 defers to 141 for script rules (wrt to everything apart from schnorr sigs instead of ecdsa sigs), and 141 has:

> The script must not fail, and result in exactly a single TRUE on the stack.

All the examples that I've pointed above have a final `OP_CHECKSIG` which pushes `TRUE` in the stack, hence when combined with `OP_CSV` that also pushes `TRUE` in the stack makes the final stack having 2 elements `TRUE`s that we need to drop one at least.

However, `OP_CHECKSIGVERIFY` pops the two elements (pk and sig, or the other way around not relevant for this argument) and verifies. It fails the script if incorrect sig against pk or otherwise DOES NOTHING.

My bad to mislead you. Thanks!

crates/asm/txs/bridge-v1/src/test_utils/unstake.rs
```diff
@@ -115,16 +107,17 @@ fn stake_connector_tapproot_addr(
...
         .unwrap()
-        .finalize(SECP256K1, *UNSPENDABLE_PUBLIC_KEY)
+        .finalize(&secp, *UNSPENDABLE_PUBLIC_KEY)
         .unwrap();
 
     let merkle_root = spend_info.merkle_root();
 
     let address = Address::p2tr(
-        SECP256K1,
+        &secp,
```
storopoli, line 129: I mean this is the exactly opposite of what I suggested.
Just use the global `SECP256K1`.

bin/alpen-cli/src/cmd/deposit.rs
prajwolrg: I think so. Is there any confusion this is causing?

crates/asm/subprotocols/bridge-v1/src/handler.rs
prajwolrg: Yes, that is correct.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
prajwolrg: This is not used anywhere.

crates/asm/txs/bridge-v1/src/constants.rs
prajwolrg: `Bridgetx::DepositRequest` is parsed when handling `BridgeTxType::Deposit`.

bin/alpen-cli/src/cmd/deposit.rs
barakshani: DRTs don't have sps50 tag

crates/asm/subprotocols/bridge-v1/src/handler.rs
barakshani: Ok, it's a question what shall we do in this case.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
barakshani: won't we need it at some future point? Especially once we include batch deposits (see [here](https://www.notion.so/Glock-based-Strata-Bridge-277901ba000f80b2b59fddb682a41a77?source=copy_link#2ab901ba000f80059472e181d7113aeb))?

crates/asm/txs/bridge-v1/src/constants.rs
barakshani: what do we need it for?

crates/asm/txs/bridge-v1/src/commit/info.rs
```diff
@@ -10,25 +10,25 @@ pub struct CommitInfo {
     /// Parsed SPS-50 auxiliary data.
     header_aux: CommitTxHeaderAux,
 
-    /// The outpoint spent by the first input.
+    /// The outpoint spent by the stake connector input.
     /// Must be validated that it spends from an N/N-locked output during transaction validation.
-    first_input_outpoint: BitcoinOutPoint,
+    stake_inpoint: BitcoinOutPoint,
```
barakshani: stake? why this is part of the commit info?

crates/asm/txs/bridge-v1/src/commit/parse.rs
```diff
@@ -3,49 +3,55 @@ use strata_codec::decode_buf_exact;
...
     commit::{CommitInfo, aux::CommitTxHeaderAux},
-    errors::{CommitParseError, Mismatch},
+    constants::BridgeTxType,
+    errors::TxStructureError,
 };
 
-/// Expected number of inputs in a commit transaction.
-const EXPECTED_COMMIT_TX_INPUT_COUNT: usize = 1;
+/// Index of the commit transaction input that spends the stake transaction output.
+const STAKE_INPUT_INDEX: usize = 0;
```
barakshani: stake for commit?

crates/asm/txs/bridge-v1/src/commit/parse.rs
```diff
@@ -3,49 +3,55 @@ use strata_codec::decode_buf_exact;
...
-        return Err(CommitParseError::InvalidInputCount(Mismatch {
-            expected: EXPECTED_COMMIT_TX_INPUT_COUNT,
-            got: tx.tx().input.len(),
-        }));
-    }
+    let header_aux: CommitTxHeaderAux = decode_buf_exact(tx.tag().aux_data())
+        .map_err(|e| TxStructureError::invalid_auxiliary_data(BridgeTxType::Commit, e))?;
+
+    // Extract the previous outpoint from the first (and only) input
+    let stake_inpoint = tx
```
barakshani: ditto

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
```diff
@@ -0,0 +1,58 @@
...
+use arbitrary::Arbitrary;
+use strata_codec::{Codec, encode_to_vec};
+use strata_l1_txfmt::TagData;
+
+use crate::{
+    constants::{BRIDGE_V1_SUBPROTOCOL_ID, BridgeTxType},
+    errors::TagDataError,
+};
+
+/// Auxiliary data in the SPS-50 header for [`BridgeTxType::DepositRequest`].
```
barakshani, line 9: but DRTs don't have an SPS-50 header

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// # Errors
+    ///
+    /// Returns [`TagDataError`] if:
+    /// - Encoding the auxiliary data fails
+    /// - The encoded auxiliary data exceeds the maximum allowed size (74 bytes)
+    pub fn build_tag_data(&self) -> Result<TagData, TagDataError> {
+        let aux_data = encode_to_vec(self)?;
+        let tag = TagData::new(
+            BRIDGE_V1_SUBPROTOCOL_ID,
+            BridgeTxType::DepositRequest as u8,
```
barakshani, line 49: In particular, I don't see where we need this type, as indicated [here](https://github.com/alpenlabs/alpen/pull/1211/files#r2620229763).

crates/asm/txs/bridge-v1/src/deposit_request/info.rs
```diff
@@ -0,0 +1,55 @@
+use arbitrary::Arbitrary;
+use strata_primitives::l1::{BitcoinAmount, BitcoinTxOut};
+
+use crate::deposit_request::DrtHeaderAux;
+
+/// Information extracted from a deposit request transaction.
+#[derive(Debug, Clone, PartialEq, Eq, Arbitrary)]
+pub struct DepositRequestInfo {
+    /// Parsed SPS-50 auxiliary data.
```
barakshani, line 9: see above on DRT and header

crates/asm/txs/bridge-v1/src/deposit_request/parse.rs
```diff
@@ -1,266 +1,46 @@
...
+    let aux_data: DrtHeaderAux = decode_buf_exact(tag.aux_data())
+        .map_err(|e| TxStructureError::invalid_auxiliary_data(BridgeTxType::DepositRequest, e))?;
+
+    // Extract the deposit request output (second output at index 1)
+    let drt_output = tx
+        .output
+        .get(DRT_OUTPUT_INDEX)
+        .ok_or_else(|| {
+            TxStructureError::missing_output(
+                BridgeTxType::DepositRequest,
```
barakshani, line 36: ditto

crates/asm/txs/bridge-v1/src/deposit_request/parse.rs
```diff
@@ -1,266 +1,46 @@
...
-        assert_eq!(result_20.address.len(), 20);
-
-        // Test with 32-byte address
-        let ee_address_32 = [0x07; 32];
-        let tx_32 = create_test_drt_tx(recovery_pk, &ee_address_32, 1_000_000);
-        let tx_input_32 = parse_tx(&tx_32);
-        let result_32 = parse_drt(&tx_input_32).unwrap();
-        assert_eq!(result_32.address.len(), 32);
-    }
+/// Returns [`TxStructureError`] if the SPS-50 format cannot be parsed, the auxiliary
```
barakshani, line 20: ditto

crates/asm/txs/bridge-v1/src/deposit/parse.rs
```diff
@@ -1,156 +1,172 @@
...
+//         let mut tx = create_test_tx(&info);
+
+//         // Remove the deposit output (keep only OP_RETURN at index 0)
+//         tx.output.truncate(1);
+
+//         let tx_input = parse_sps50_tx(&tx);
+//         let err = parse_deposit_tx(&tx_input).unwrap_err();
+//         assert!(matches!(err, DepositTxParseError::MissingDepositOutput));
+//     }
+// }
```
storopoli: Why are we commenting this whole section? Can we just remove it?

barakshani: I think this should return now (or soon) after this https://github.com/alpenlabs/alpen/pull/1211/commits/0db3b258df6b30b6516fb47e254f0f20c3c85918

storopoli: Yes, makes sense. Just that whenever a PR is "green" (non-draft) it is "ready to be merged". Wanted to flag this hunk of commented code so that it does not get merged yet.

bin/alpen-cli/src/cmd/deposit.rs
prajwolrg: DRTs need to have some identifier. Right now we are using OP_RETURN `b"ALPN"` (or same as magic defined in the SPS-50) on first output UTXO. . We can skip the `SubprotoId` and `TxId`, but I think this is good for consistency.

barakshani: why not calling it `drt_tag` or something?

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
prajwolrg: Even for batching this should not be required. For the bridge subprotocol, all we rely on is the `deposit_idx`.

crates/asm/txs/bridge-v1/src/constants.rs
prajwolrg: To extract out the user destination address on L2. Right now, we're forced to pass the user destination address of L2 onto DT to make it compatible with the existing logic, but it should not be required.

crates/asm/txs/bridge-v1/src/deposit/parse.rs
prajwolrg: Sorry, my bad, I meant to fix this.

prajwolrg: Done in [b14df71](https://github.com/alpenlabs/alpen/pull/1211/commits/b14df7198e94aa7327c40e79de15babc5d37f4b5).

crates/asm/subprotocols/bridge-v1/src/handler.rs
prajwolrg: In that case, we don't accept the deposit. It is (kindof) equivalent as funds being stolen by N/N.

crates/asm/txs/bridge-v1/src/commit/parse.rs
prajwolrg: Yes, based on old design. It might be better to remove commit transaction as a whole if this is causing confusion.

prajwolrg:
> Let's agree on DRT and SPS50 headers, we seem to have some misunderstanding here.

DRTs need to have some identifier as this is user-initiated. Right now we are using OP_RETURN b"ALPN" (or same as magic defined in the SPS-50) on first output UTXO. We can skip the SubprotoId and TxId, but making is SPS-50 compatible is good for consistency.

crates/asm/txs/bridge-v1/src/constants.rs
barakshani: I understand that we need to "call" the DRT once we see a DRT. What I don't understand is we do we need a special tx type. Seems to me that we'll call the DRT, then we do some processing, then we discard it.

crates/asm/txs/bridge-v1/src/commit/parse.rs
barakshani: As far as I can tell, we're not using stake anywhere for the purpose of committing. I'd rather either reflecting this in the code, or indeed remove the entire thing. But I don't want us to work on, or show, things that aren't relevant anymore.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
Rajil1213: This is in a fuzzy territory now that the ASM is also technically dealing with the DRT. We have this agreement that any transaction relevant to the ASM needs this tag because otherwise, it wouldn't even be parsed. That is not the case here since this is being fetched directly. However, the main issue here is I think of aesthetics and not functionality. It's weird for the ASM to be parsing a transaction but using a different mechanism to do so for a particular transaction. I guess we _can_ add a header here which would make things consistent. The bridge does not really care about this -- we were already fine with having the user pay for the 32 bytes of x-only recovery key twice (once in DRT and and then in DT) and now that it is only going to be present in the DRT, the net cost is still low. The bridge already expects some magic bytes whose length would now be 6 bytes instead of 4 if we make it SPS-50 compatible.

Rajil1213: ~I'd have to check the actual implementation but I think the ASM does not even need the transaction. It just needs the utxo being spent. So it doesn't even have to parse the entire transaction. If this is the case, then I don't think there are any aesthetic issues either other than the fact that there are some transactions that have a different OP_RETURN header format in the overall system.~

barakshani: I don't have a strong opinion on this, but to me, given that DRT is fetched directly, and given that DT has already passed the 'validation', I don't see any reason -- aesthetic or not -- to have this header; I just don't find it useful.

Rajil1213: I think the reason is that it needs to be parsed to extract the recovery key. If you're saying we shouldn't have any header, then we _can_ do that too but there may be more false positives in the bridge.

barakshani: Why would ASM need to recovery key?
We should have _some_ header, for recovery and where to mint, but it's not sps50 compatible.

Rajil1213: I meant the deposit descriptor (evm address). I think what @prajwolrg is getting at is why not make it compatible with SPS 50 and therefore consistent if we are adding a header anyway.

crates/asm/txs/bridge-v1/src/deposit/aux.rs
```diff
@@ -9,17 +9,27 @@ use crate::{BRIDGE_V1_SUBPROTOCOL_ID, constants::BridgeTxType, errors::TagDataEr
 pub struct DepositTxHeaderAux {
     /// idx of the deposit as given by the N/N multisig.
     deposit_idx: u32,
+    // TODO:PG- This is not really required, we are adding it here just to make sure that the
+    // existing functional tests pass.
+    ee_address: [u8; 20],
 }
```
storopoli, line 15: Can you please move this to a ticket so we don't lose this in a deep buried `TODO` comment in the codebase?

storopoli (dismissed):
ACK b14df7198e94aa7327c40e79de15babc5d37f4b5

crates/asm/txs/bridge-v1/src/test_utils/deposit_request.rs
barakshani, line 1: we seem to already have a DRT "builder" here crates/asm/txs/bridge-v1/src/deposit_request/build.rs

prajwolrg: I don't think we have that anymore. We agreed to move all the transaction building to test-utils.

bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -124,28 +125,28 @@ pub async fn deposit(
...
-        .expect("conversion should succeed after length check");
+
+    // Construct the DRT metadata using the canonical builder
+    let drt_sps50_tag = DrtHeaderAux::new(recovery_public_key.serialize(), alpen_address.into())
+        .build_tag_data()
+        .expect("tag data should be created");
+
+    let sps50_script = ParseConfig::new(magic_bytes)
+        .encode_script_buf(&drt_sps50_tag.as_ref())
+        .expect("drt metadata should be created");
```
Rajil1213: Since the ASM now has a parser for DRT. If possible, we should add a unit test here that checks the DRT constructed by the `alpen-cli` against the parser.

crates/asm/subprotocols/bridge-v1/src/validation/deposit.rs
```diff
@@ -0,0 +1,154 @@
...
+        return Err(DepositValidationError::WrongOutputLock(Mismatch {
+            expected: expected_script.clone(),
+            got: info.locked_script().clone(),
+        }));
+    }
+
+    let expected_drt_script = create_deposit_request_locking_script(
+        drt_info.header_aux().recovery_pk(),
+        state.operators().agg_key().to_xonly_public_key(),
+        RECOVER_DELAY,
```
Rajil1213, line 39: Does the ASM accept configurable parameters? If it does, then it might be a good idea to not hardcode this value. This value is mostly a product decision and can change. So if it does, not hardcoding this would allow us to reset this without requiring a recompilation.

crates/asm/txs/bridge-v1/src/deposit/aux.rs
```diff
@@ -1,116 +1,52 @@
...
+    /// Builds a `TagData` instance from this auxiliary data.
+    ///
+    /// This method encodes the auxiliary data and constructs the tag data for inclusion
+    /// in the SPS-50 OP_RETURN output.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`TagDataError`] if:
+    /// - Encoding the auxiliary data fails
+    /// - The encoded auxiliary data exceeds the maximum allowed size (74 bytes)
```
Rajil1213: Is this limit still relevant? (given that this PR is already massive, I don't recommend making that change here though)

crates/asm/txs/bridge-v1/src/deposit/info.rs
```diff
@@ -1,58 +1,77 @@
...
     header_aux: DepositTxHeaderAux,
 
-    /// The amount of Bitcoin deposited.
-    amt: BitcoinAmount,
+    /// The deposit output containing the deposited amount and its locking script.
+    deposit_output: BitcoinTxOut,
 
-    /// The outpoint of the deposit transaction.
-    outpoint: BitcoinOutPoint,
+    /// Previous outpoint referenced by the DRT input.
```
Rajil1213:
```suggestion
    /// Previous outpoint referenced by the DT input.
```

crates/asm/txs/bridge-v1/src/constants.rs
```diff
@@ -1,14 +1,63 @@
...
+    WithdrawalFulfillment = 2,
+    /// Commit transaction - operator commits to a game
+    Commit = 3,
+    /// Slash transaction - penalize misbehaving operator
+    Slash = 4,
+    /// Unstake transaction - operator exits the bridge
+    Unstake = 5,
+}
+
+impl From<BridgeTxType> for u8 {
```
Rajil1213, line 30: You should a roundtrip test for this conversion to make sure that the conversion to and from `u8` is right.

crates/asm/txs/bridge-v1/src/deposit_request/info.rs
```diff
@@ -0,0 +1,51 @@
...
+    header_aux: DrtHeaderAux,
+
+    /// The deposit request output containing the amount and its locking script.
+    ///
+    /// The amount must be larger than the bridge denomination amount to cover
+    /// the mining fee required to settle the deposit transaction.
```
Rajil1213: I don't think this doc comment belongs here unless the construction of `DepositRequestInfo` is not possible if this is violated (which does not seem to be the case).

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
```diff
@@ -82,8 +82,8 @@ impl Codec for DrtHeaderAux {
 impl<'a> Arbitrary<'a> for DrtHeaderAux {
     fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
         let recovery_pk = <[u8; 32]>::arbitrary(u)?;
-        // Generate address between 20 and 64 bytes (reasonable range for EE addresses)
-        let addr_len = u.int_in_range(20..=64)?;
+        // Generate address between 20 and 42 bytes
+        let addr_len = u.int_in_range(20..=42)?;
```
Rajil1213: What is the reason for this diff?

crates/asm/txs/bridge-v1/src/test_utils/deposit.rs
```diff
@@ -47,16 +50,42 @@ pub fn create_connected_drt_and_dt(
...
+    }
+    dbg!(&drt);
+    dbg!(&drt.compute_txid());
+    dbg!(&drt.compute_wtxid());
 
-    let _ =
-        submit_transaction_with_keys_blocking(&bitcoind, &client, operator_keys, &mut dt).unwrap();
+    dbg!(&dt);
+    dbg!(&dt.compute_txid());
+    dbg!(&dt.compute_wtxid());
```
Rajil1213: Have these been left out intentionally?

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
```diff
@@ -387,9 +387,10 @@ mod tests {
 
         privkeys.pop();
 
-        let err = bridge_state.add_deposit(&deposit_info).unwrap_err();
+        let _err = bridge_state.add_deposit(&deposit_info).unwrap_err();
 
-        assert!(matches!(err, DepositValidationError::DrtSignature(_)));
+        // FIXME:
+        // assert!(matches!(err, DepositValidationError::DrtSignature(_)));
```
Rajil1213: Is this fixed in later commits?

crates/asm/txs/bridge-v1/src/commit/info.rs
Rajil1213: Looks like the structure was copy-pasted from the `SlashInfo`?

Rajil1213 (dismissed):
Approving preemptively since most of my review comments are nits or questions.

As Barak said, this is too big of a PR with too many unrelated changes, so reviewing it was a pain. There are probably whole chunks of this PR that I did not review properly.

bin/alpen-cli/src/cmd/deposit.rs
prajwolrg: We are keeping SPS-50 header on DRT. So keeping the name as is.

crates/asm/txs/bridge-v1/src/constants.rs
prajwolrg: Discussed in detail [here](https://github.com/alpenlabs/alpen/pull/1211#discussion_r2626677070).

crates/asm/txs/bridge-v1/src/commit/info.rs
prajwolrg: Removing the commit transaction as a whole.

crates/asm/txs/bridge-v1/src/commit/parse.rs
prajwolrg: Removing the whole commit thing.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
prajwolrg: Yes, Rajil put it perfectly. Making DRT SPS-50 compatible for now.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
prajwolrg: Discussed https://github.com/alpenlabs/alpen/pull/1211#discussion_r2626677070.

crates/asm/txs/bridge-v1/src/commit/parse.rs
prajwolrg: Removing commit tx.

crates/asm/txs/bridge-v1/src/deposit_request/info.rs
prajwolrg: Discussed https://github.com/alpenlabs/alpen/pull/1211#discussion_r2626677070.

crates/asm/txs/bridge-v1/src/deposit_request/parse.rs
prajwolrg: Discussed https://github.com/alpenlabs/alpen/pull/1211#discussion_r2626677070.

crates/asm/txs/bridge-v1/src/deposit/aux.rs
prajwolrg: This can now be done after https://github.com/alpenlabs/alpen/pull/1222. I've updated [this ticket](https://alpenlabs.atlassian.net/browse/STR-1950) to handle this as well.

crates/asm/subprotocols/bridge-v1/src/validation/deposit.rs
prajwolrg: Created a separate ticket for this [here](https://alpenlabs.atlassian.net/browse/STR-2003).

crates/asm/subprotocols/bridge-v1/src/state/bridge.rs
prajwolrg: Yes, it's done as part of https://github.com/alpenlabs/alpen/pull/1211/changes/658a4d235d63d3d78bb0c137fc2f8cb87f3283a8.

crates/asm/txs/bridge-v1/src/test_utils/deposit.rs
prajwolrg: This is cleaned up in later commit.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
prajwolrg: Originally it was 64 bytes, which was causing some flaky tests since the aux data cannot be more than 74 bytes.

prajwolrg: Originally it was 64 bytes, which was causing some flaky tests since the aux data cannot be more than 74 bytes. This should no longer be an issue after [this](https://alpenlabs.atlassian.net/browse/STR-1950).

crates/asm/txs/bridge-v1/src/deposit/aux.rs
prajwolrg: It's not that relevant now. This should never happen. It makes more sense to directly panic instead and add test to show that doesn't happen.

Doing it here since, we added the `build_tag_data` in this PR.

PR #1213, opened 2025-12-17
STR-1971: update logger configuration 

voidash:
## Description

Changes and addition in Structured logging via tracing, OpenTelemetry integration for distributed tracing,  Proper resource semantic conventions, Configurable span limits to prevent pollution, Graceful shutdown with span flushing , JSON and compact output formats, Trace context propagation

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

- [] I have performed a self-review of my code.
- [] I have commented my code where necessary.
- [] I have updated the documentation if needed.
- [] My changes do not introduce new warnings.
- [] I have added (where necessary) tests that prove my changes are effective or that my feature works.

## Related Issues

https://alpenlabs.atlassian.net/browse/STR-1971

### DIFF
--- a/crates/common/src/logging/manager.rs
+++ b/crates/common/src/logging/manager.rs
@@ -1,6 +1,6 @@
 //! Logging initialization and shutdown management.
 
-use std::{sync::OnceLock, time::Duration};
+use std::sync::OnceLock;
 
 use opentelemetry::trace::TracerProvider;
 use opentelemetry_otlp::WithExportConfig;


bin/strata-sequencer-client/src/config.rs
```diff
@@ -6,13 +6,41 @@ use crate::args::Args;
 
 const DEFAULT_DUTY_POLL_INTERVAL: u64 = 1000;
 
+/// Logging configuration for the sequencer client.
+#[derive(Debug, Clone, Deserialize, Default)]
+pub(crate) struct LoggingConfig {
```
bewakes, line 11: Can we reuse LoggingConfig over multiple binaries/components?

bin/strata-sequencer-client/src/main.rs
```diff
@@ -85,23 +86,53 @@ fn get_config(args: Args) -> Result<Config> {
 
 /// Sets up the logging system given a handle to a runtime context to possibly
 /// start the OTLP output on.
-fn init_logging(rt: &Handle) {
-    let mut lconfig = logging::LoggerConfig::with_base_name("strata-sequencer");
+fn init_logging(rt: &Handle, config: &Config) {
```
bewakes, line 89: Also just wondering, can we reuse the init_logging function as well? Seems to have similar patterns across different binaries.

bewakes (commented):
On a quick glance, this looks good. Could you also add the setup for new strata binary?

bin/strata-sequencer-client/src/config.rs
storopoli: I think this is the most important review comment!
Can we have "Logging As A Service" similar to how we did with the Prover in https://github.com/alpenlabs/alpen/pull/1149?

bin/strata-sequencer-client/src/main.rs
voidash: I don't think reusing init_logging across different binaries is a good idea. Main reason being, the separation of configuration options we want `common/logging` to handle and some binary specific configurations that might exist. For example in this case, before you start logging you have to enter the runtime.

crates/common/src/logging/manager.rs
```diff
@@ -0,0 +1,152 @@
...
+        debug!("no tracer provider to shut down");
+    }
+
+    // Shutdown global text map propagator
+    opentelemetry::global::shutdown_tracer_provider();
+}
+
+/// Shuts down with a custom timeout for span flushing
+pub fn finalize_with_timeout(timeout: Duration) {
+    info!(?timeout, "shutting down logging with custom timeout");
```
bewakes: The timeout is only being used in `info!`. The name `finalize_with_timeout` suggests that it does something more than this. Is this intended?

crates/common/src/logging/service.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    if let Some(file_config) = &file_logging_config {
+        lconfig = lconfig.with_file_logging(file_config.clone());
+    }
+
+    // Configure JSON format if specified
+    if let Some(json_format) = config.json_format {
+        lconfig = lconfig.with_json_logging(json_format);
+    }
+
```
bewakes, line 60: nit: Might be a good idea to separate this into a function something like `init_config_to_logging_config(LoggingInitCofnig) -> LoggingConfig`.

bewakes (dismissed):
Only some nits, feel free to address them. Otherwise looks good.

PR #1214, opened 2025-12-18
add types for Batch and Chunk

sapinb:
## Description
- [x] Add ~EeUpdate~ Batch and Chunk types
- [x] Add storage trait for persisting above types
- [x] Track account's inbox index during block assembly

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
This adds types and storage trait to unblock parallel work on ee prover implementation.
The account inbox index is also required during proving.
Best to review by commit as the third one touches a lot of places.

## Checklist

### DIFF
--- a/crates/alpen-ee/common/src/traits/storage/batch.rs
+++ b/crates/alpen-ee/common/src/traits/storage/batch.rs
@@ -10,8 +10,8 @@ use crate::{Batch, BatchId, BatchStatus, Chunk, ChunkId, ChunkStatus, StorageErr
 /// Chunks are subsets of blocks in a batch, and units for proving ee chain execution, typically
 /// limited by prover constraints.
 ///
-/// An Batch will consist of 1 or more chunks.
-/// These chunks can be generated before or after updates, but the chunks must fit inside Batch
+/// A batch will consist of 1 or more chunks.
+/// These chunks can be generated before or after updates, but the chunks must fit inside batch
 /// boundaries.
 ///
 /// ... | ------------ Batch 1 ------------- | --------- Batch 2 ------- | ...
--- a/crates/alpen-ee/common/src/types/batch.rs
+++ b/crates/alpen-ee/common/src/types/batch.rs
@@ -6,7 +6,7 @@ use strata_identifiers::L1BlockCommitment;
 
 use crate::ProofId;
 
-/// Unique, deterministic identifier for an Batch
+/// Unique, deterministic identifier for a batch.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct BatchId {
     prev_block: Hash,
@@ -56,19 +56,18 @@ pub enum BatchStatus {
 /// Represents a sequence of blocks that are treated as a unit for DA and posting updates to OL.
 #[derive(Debug)]
 pub struct Batch {
-    /// Sequential update index, also used in
+    /// Sequential update index.
     idx: u64,
-    /// last block of (idx - 1)th update
+    /// Last block of (idx - 1)th update.
     prev_block: Hash,
-    /// last block in this update
+    /// Last block in this update.
     last_block: Hash,
-    /// rest of the blocks in this update.
-    /// cached here for easier processing.
+    /// Rest of the blocks in this update, cached here for easier processing.
     inner_blocks: Vec<Hash>,
 }
 
 impl Batch {
-    /// Create a new Batch.
+    /// Create a new batch.
     pub fn new(idx: u64, prev_block: Hash, last_block: Hash, inner_blocks: Vec<Hash>) -> Self {
         debug_assert_ne!(prev_block, last_block);
         Self {
@@ -85,22 +84,21 @@ impl Batch {
     }
 
     /// Get sequential index.
-    /// This should equal the sequence number in account update sent to OL.
     pub fn idx(&self) -> u64 {
         self.idx
     }
 
-    /// last block of the previous Batch.
+    /// last block of the previous batch.
     pub fn prev_block(&self) -> Hash {
         self.prev_block
     }
 
-    /// last block of this Batch
+    /// last block of this batch.
     pub fn last_block(&self) -> Hash {
         self.last_block
     }
 
-    /// Iterate over all blocks in range of this Batch.
+    /// Iterate over all blocks in range of this batch.
     pub fn blocks_iter(&self) -> impl Iterator<Item = Hash> + '_ {
         self.inner_blocks
             .iter()
--- a/crates/alpen-ee/common/src/types/chunk.rs
+++ b/crates/alpen-ee/common/src/types/chunk.rs
@@ -34,13 +34,13 @@ impl ChunkId {
 /// Represents a sequence of blocks that are processed together as a unit during proving.
 #[derive(Debug)]
 pub struct Chunk {
-    /// Sequential chunk index
+    /// Sequential chunk index.
     idx: u64,
-    /// last block of (idx - 1)th chunk that this chunk extends
+    /// Last block of (idx - 1)th chunk that this chunk extends.
     prev_block: Hash,
-    /// last block of this chunk. A chunk cannot be empty.
+    /// Last block of this chunk. A chunk cannot be empty.
     last_block: Hash,
-    /// rest of the blocks in the chunk.
+    /// Rest of the blocks in the chunk.
     inner_blocks: Vec<Hash>,
 }
 
@@ -56,22 +56,22 @@ impl Chunk {
         }
     }
 
-    /// Deterministic chunk id
+    /// Deterministic chunk id.
     pub fn id(&self) -> ChunkId {
         ChunkId::new(self.prev_block, self.last_block)
     }
 
-    /// Sequential chunk index
+    /// Sequential chunk index.
     pub fn idx(&self) -> u64 {
         self.idx
     }
 
-    /// last block of (idx - 1)th chunk that this chunk extends
+    /// Last block of (idx - 1)th chunk, that this chunk extends.
     pub fn prev_block(&self) -> Hash {
         self.prev_block
     }
 
-    /// last block of this chunk.
+    /// Last block of this chunk.
     pub fn last_block(&self) -> Hash {
         self.last_block
     }
--- a/crates/alpen-ee/sequencer/src/block_builder/task.rs
+++ b/crates/alpen-ee/sequencer/src/block_builder/task.rs
@@ -212,7 +212,7 @@ async fn build_next_block(
         .await
         .context("build_next_block: failed to get best exec block")?;
 
-    // check if last local block is not as expected from previous block building cycle
+    // Check if last local block is not as expected from previous block building cycle
     // This shouldn't happen in a single sequencer case, but checking for sanity anyway.
     if last_local_block.parent_blockhash() != expected_block_target.parent {
         warn!(


crates/alpen-ee/common/src/traits/prover.rs
```diff
@@ -0,0 +1,39 @@
...
+#[derive(Debug)]
+pub enum ProofGenerationStatus {
+    /// Proof generation started or proof is getting generated
+    Pending,
+    /// Proof is ready and can be fetched using proof_id
+    Ready { proof_id: ProofId },
+    /// Proof generation has not been triggered for provided ee_update_id.
+    Invalid,
+    /// Cannot generate proof for some reason. All retries exhausted, etc.
+    Failed { reason: String },
```
storopoli: I think that "not started" being `Invalid` gives the wrong semantics.
Why not `NotStarted` or something else?

crates/alpen-ee/common/src/traits/prover.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/// Interface between Prover and Batch assembly
+/// Note: In case of reorgs, `EeUpdateId` is unique enough to not conflict between forks.
+/// Note2: eyre::Result<> is here mainly to pass back db related and other errors.
```
storopoli: This does not belong to the docstring.
Or at least you could say something like

"Most of the return types will have transparent errors from ..."

crates/alpen-ee/common/src/traits/prover.rs
```diff
@@ -0,0 +1,39 @@
...
+    /// Check if proof is generated for ee_update_id.
+    ///
+    /// The generated proof is expected to be persisted, available to be fetched at any time
+    /// afterwards with the returned proof_id.
+    async fn check_proof_ready(
+        &self,
+        ee_update_id: EeUpdateId,
+    ) -> eyre::Result<ProofGenerationStatus>;
+
+    /// Get a persisted proof
```
storopoli: "Persisted proof" is strange.

```suggestion
    /// Get a proof from persistence storage.
```

crates/alpen-ee/common/src/types/chunk.rs
```diff
@@ -0,0 +1,83 @@
...
+
+/// Lifecycle states for chunk
+#[derive(Debug)]
+pub enum ChunkStatus {
+    /// Proving has not started yet.
+    Unproven,
+    /// Proving started. Pending proof generation.
+    PendingProof(String),
+    /// Valid proof ready.
+    /// TODO: correct proof type
```
storopoli: Not a docstring but a comment

crates/alpen-ee/common/src/traits/prover.rs
storopoli: You could use the `Unproven` from `ChunkStatus` to make these more similar.

crates/alpen-ee/common/src/types/chunk.rs
```diff
@@ -0,0 +1,83 @@
...
+#[derive(Debug)]
+pub struct Chunk {
+    /// Sequential chunk index
+    idx: u64,
+    /// last block of (idx - 1)th chunk that this chunk extends
+    prev_block: Hash,
+    /// last block of this chunk. A chunk cannot be empty.
+    last_block: Hash,
+    /// rest of the blocks in the chunk.
+    blocks: Vec<Hash>,
```
storopoli: Do we need to grow this dynamically?
Otherwise, we could do

```suggestion
    /// rest of the blocks in the chunk.
    blocks: &[Hash],
```

crates/alpen-ee/common/src/types/ee_update.rs
```diff
@@ -0,0 +1,87 @@
...
+
+/// EeUpdate lifecycle states
+#[derive(Debug)]
+pub enum EeUpdateStatus {
+    /// Newly created
+    Init,
+    /// DA started, waiting for block
+    DaPending { txids: Vec<Descriptor> },
+    /// DA complete
+    DaComplete { da: Vec<L1DaRef> },
```
storopoli: Do we need this to grow dynamically?
```suggestion
    DaComplete { da: &[L1DaRef] },
```

crates/alpen-ee/common/src/types/ee_update.rs
```diff
@@ -0,0 +1,87 @@
...
+    // inclusion merkle proof ?
+}
+
+/// EeUpdate lifecycle states
+#[derive(Debug)]
+pub enum EeUpdateStatus {
+    /// Newly created
+    Init,
+    /// DA started, waiting for block
+    DaPending { txids: Vec<Descriptor> },
```
storopoli: How are we representing a txid with a BOSD `Descriptor`?
That is to represent an address or more precisely a `scriptPubkey`.

crates/alpen-ee/common/src/types/ee_update.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// DA complete
+    DaComplete { da: Vec<L1DaRef> },
+    /// Proving started, waiting for proof
+    ProofPending {
+        da: Vec<L1DaRef>,
```
storopoli: Do we need this to grow dynamically?

```suggestion
        da: &[L1DaRef],
```

crates/alpen-ee/common/src/types/ee_update.rs
```diff
@@ -0,0 +1,87 @@
...
+        proof_job_id: String,
+    },
+    /// Proof ready
+    /// TODO: correct proof type
+    ProofReady { da: Vec<L1DaRef>, proof: Vec<u8> },
```
storopoli: Do we need this to grow dynamically?

```suggestion
    ProofReady { da: &[L1DaRef], proof: Vec<u8> },
```

storopoli: Also for the `proof`

```suggestion
    ProofReady { da: Vec<L1DaRef>, proof: &[u8] },
```

crates/alpen-ee/common/src/types/ee_update.rs
```diff
@@ -0,0 +1,87 @@
...
+pub struct EeUpdate {
+    /// Sequential update index, also used in
+    idx: u64,
+    /// last block of (idx - 1)th update
+    prev_block: Hash,
+    /// last block in this update
+    last_block: Hash,
+    /// rest of the blocks in this update.
+    /// cached here for easier processing.
+    blocks: Vec<Hash>,
```
storopoli: Do we need this to grow dynamically?

```suggestion
    blocks: &[Hash],
```

crates/alpen-ee/common/src/types/prover.rs
```diff
@@ -0,0 +1,9 @@
+use strata_acct_types::Hash;
+
+/// Unique identifier to a persisted proof.
+pub type ProofId = Hash;
+
+/// TODO: proper proof type
+#[expect(dead_code, reason = "wip")]
+#[derive(Debug)]
+pub struct Proof(Vec<u8>);
```
storopoli, line 9: Do we need this to grow dynamically?

```suggestion
pub struct Proof(&[u8]);
```

crates/alpen-ee/sequencer/src/ol_chain_tracker/state.rs
```diff
@@ -5,53 +5,90 @@ use strata_identifiers::{OLBlockCommitment, OLBlockId};
 use strata_snark_acct_types::MessageEntry;
 use tracing::warn;
 
+#[derive(Debug)]
+pub(crate) struct BlockInfo {
+    pub messages: Vec<MessageEntry>,
```
storopoli: Do we need this to grow dynamically?
Also does it need to be `pub`?

```suggestion
    messages: &[MessageEntry],
```

crates/alpen-ee/sequencer/src/ol_chain_tracker/state.rs
```diff
@@ -5,53 +5,90 @@ use strata_identifiers::{OLBlockCommitment, OLBlockId};
...
+            next_inbox_msg_idx,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct InboxMessages {
+    pub messages: Vec<MessageEntry>,
+    pub next_inbox_msg_idx: u64,
+}
```
storopoli, line 42: Do we need this to grow dynamically?
And does it need to be `pub`?

```suggestion
#[derive(Debug)]
pub struct InboxMessages {
    messages: &[MessageEntry],
    next_inbox_msg_idx: u64,
}
```

storopoli (changes requested):
Almost all types only implement `Debug`.
Is that deliberate?
Should we also implement `Clone`,`PartialEq` and `Eq` (note that some types that have unique IDs can have a manual `PartialEq`/`Eq` implementation by just comparing the ID)?

crates/alpen-ee/common/src/types/chunk.rs
sapinb: This needs to be an owned type afaik.
So the choice is Vec<Hash> of Box<[Hash]>.
I think I prefer Vec 😄

crates/alpen-ee/common/src/traits/prover.rs
```diff
@@ -15,8 +15,6 @@ pub enum ProofGenerationStatus {
 }
 
 /// Interface between Prover and Batch assembly
-/// Note: In case of reorgs, `EeUpdateId` is unique enough to not conflict between forks.
```
storopoli: This docstring (the "Note:") was valuable.
The one who as not was the "Note2: " 😂

crates/alpen-ee/common/src/types/chunk.rs
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+
+    pub fn set_proof_pending(&mut self, proof_task_id: String) {
+        self.proof_status = ChunkStatus::PendingProof(proof_task_id);
+    }
+
+    pub fn set_proof(&mut self, proof: ProofId) {
+        self.proof_status = ChunkStatus::ProofReady(proof);
+    }
```
delbonis: This pattern involving a data structure that sometimes stores a proof and sometimes doesn't existed before with checkpoints and led to a ton of interesting bugs.  We should store the proofs separately and have this `Chunk` type strictly be a description of the things going into the thing.

Also should have a builder instead of directly constructing it to improve the ergonomics around the weirdness of always needing to have at least one block.

crates/alpen-ee/common/src/types/chunk.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// Valid proof ready.
+    ProofReady(ProofId),
+}
+
+/// Unique identifier for a chunk
+#[derive(Debug)]
+pub struct ChunkId {
+    pub prev_block: Hash,
+    pub last_block: Hash,
+}
```
delbonis: I don't think we need `prev_block` here do we?  We can reconstruct the chunk data itself just from the last block and knowing the list of chunks.  Or if this is really meant to be a generic description of the range then I think `ChunkId` is a misleading name for this, maybe `ChunkBounds`?

Also should we define a `ExecBlkid` instead of using plain `Hash` to represent this?

crates/alpen-ee/common/src/types/ee_update.rs
```diff
@@ -0,0 +1,88 @@
...
+pub struct EeUpdateId {
+    pub prev_block: Hash,
+    pub last_block: Hash,
+}
+
+/// DA related data relevant to EeUpdate
+#[derive(Debug)]
+pub struct L1DaRef {
+    pub block: L1BlockCommitment,
+    pub txns: Vec<Txid>,
```
delbonis: Should be wtxids.

crates/alpen-ee/common/src/types/ee_update.rs
```diff
@@ -0,0 +1,88 @@
...
+use strata_identifiers::L1BlockCommitment;
+
+use crate::ProofId;
+
+/// Unique identifier for an EeUpdate
+#[derive(Debug)]
+pub struct EeUpdateId {
+    pub prev_block: Hash,
+    pub last_block: Hash,
+}
```
delbonis: see other comment about ids vs bounds

crates/alpen-ee/common/src/types/ee_update.rs
```diff
@@ -0,0 +1,88 @@
...
+    idx: u64,
+    /// last block of (idx - 1)th update
+    prev_block: Hash,
+    /// last block in this update
+    last_block: Hash,
+    /// rest of the blocks in this update.
+    /// cached here for easier processing.
+    blocks: Vec<Hash>,
+    /// Status
+    status: EeUpdateStatus,
```
delbonis: see other comment about proving state

crates/alpen-ee/sequencer/src/block_builder/task.rs
```diff
@@ -124,21 +141,20 @@ pub async fn block_builder_task<
 }
 
 async fn block_builder_task_inner<TEngine: PayloadBuilderEngine>(
+    next_block_target: &BlockTarget,
     config: &BlockBuilderConfig,
     exec_chain_handle: &ExecChainHandle,
     ol_chain_handle: &OLChainTrackerHandle,
     payload_builder: &TEngine,
     storage: &impl ExecBlockStorage,
     clock: &impl Clock,
```
delbonis, line 150: Can some of this be bundled up and hidden behind some `BuilderContext`/`BuilderContextProvider` trait?

crates/alpen-ee/sequencer/src/ol_chain_tracker/state.rs
```diff
@@ -5,53 +5,105 @@ use strata_identifiers::{OLBlockCommitment, OLBlockId};
 use strata_snark_acct_types::MessageEntry;
 use tracing::warn;
 
+/// Inbox messages within some range of blocks and the next expected inbox message idx of the last
+/// block of the range
+#[derive(Debug)]
+pub struct InboxMessages {
+    messages: Vec<MessageEntry>,
+    next_inbox_msg_idx: u64,
```
delbonis, line 13: I would think this field might make more sense to be `first_inbox_msg_idx` so that it relates to validation better and we compute the *new* value of the "next msg idx" from that, but I'm not sure how this used.

crates/alpen-ee/sequencer/src/ol_chain_tracker/state.rs
```diff
@@ -5,53 +5,105 @@ use strata_identifiers::{OLBlockCommitment, OLBlockId};
...
+    pub fn new_empty(next_inbox_msg_idx: u64) -> Self {
+        Self {
+            messages: vec![],
+            next_inbox_msg_idx,
+        }
+    }
+
+    pub fn messages(&self) -> &[MessageEntry] {
+        &self.messages
+    }
```
delbonis, line 26: *might* want to have an iterator fn that returns `(u64, &MessageEntry)` so it's easier to deal with the indexes

crates/alpen-ee/common/src/traits/prover.rs
```diff
@@ -0,0 +1,37 @@
...
+use crate::{EeUpdateId, Proof, ProofId};
+
+#[derive(Debug)]
+pub enum ProofGenerationStatus {
+    /// Proof generation started or proof is getting generated
+    Pending,
+    /// Proof is ready and can be fetched using proof_id
+    Ready { proof_id: ProofId },
+    /// Proof generation has not been triggered for provided ee_update_id.
+    Unproven,
```
barakshani:
```suggestion
    NotStarted,
```
everything not in Ready state is technically unproven

crates/alpen-ee/common/src/traits/prover.rs
```diff
@@ -0,0 +1,37 @@
...
+pub trait EeUpdateProver: Sized {
+    /// Request proof generation for ee_update_id.
+    /// Ok(()) -> proof generation has been queued
+    async fn begin_proof_generation(&self, ee_update_id: EeUpdateId) -> eyre::Result<()>;
+
+    /// Check if proof is generated for ee_update_id.
+    ///
+    /// The generated proof is expected to be persisted, available to be fetched at any time
+    /// afterwards with the returned proof_id.
+    async fn check_proof_ready(
```
barakshani:
```suggestion
    async fn check_proof_status(
```
this doesn't return true/false on being ready -- but the proof status -- right?

crates/alpen-ee/common/src/traits/prover.rs
```diff
@@ -0,0 +1,37 @@
...
+    /// Cannot generate proof for some reason. All retries exhausted, etc.
+    Failed { reason: String },
+}
+
+/// Interface between Prover and Batch assembly
+#[async_trait]
+pub trait EeUpdateProver: Sized {
+    /// Request proof generation for ee_update_id.
+    /// Ok(()) -> proof generation has been queued
+    async fn begin_proof_generation(&self, ee_update_id: EeUpdateId) -> eyre::Result<()>;
```
barakshani:
```suggestion
    async fn request_proof_generation(&self, ee_update_id: EeUpdateId) -> eyre::Result<()>;
```

crates/alpen-ee/common/src/traits/storage/update_chunk.rs
```diff
@@ -0,0 +1,63 @@
+use async_trait::async_trait;
+
+use crate::{Chunk, ChunkId, ChunkStatus, EeUpdate, EeUpdateId, EeUpdateStatus, StorageError};
+
+/// Storage for EeUpdates and Chunks
+///
+/// EeUpdates are units for DA and posting updates to OL, typically limited by DA size constraints.
```
barakshani:
```suggestion
/// EeUpdates are units for DA and posting proven state updates to OL, typically limited by DA size constraints.

File: crates/alpen-ee/common/src/types/chunk.rs [hunk 27]
```diff
@@ -0,0 +1,84 @@
...
+
+use strata_acct_types::Hash;
+
+use crate::ProofId;
+
+/// Lifecycle states for chunk
+#[derive(Debug)]
+pub enum ChunkStatus {
+    /// Proving has not started yet.
+    Unproven,
```
barakshani:
```suggestion 
    NotStarted,
```
for consistency with above.

crates/alpen-ee/common/src/types/chunk.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// Proving started. Pending proof generation.
+    PendingProof(String),
+    /// Valid proof ready.
+    ProofReady(ProofId),
```
barakshani: why different language than the above?

crates/alpen-ee/common/src/types/chunk.rs
```diff
@@ -0,0 +1,84 @@
...
+#[derive(Debug)]
+pub struct Chunk {
+    /// Sequential chunk index
+    idx: u64,
+    /// last block of (idx - 1)th chunk that this chunk extends
+    prev_block: Hash,
+    /// last block of this chunk. A chunk cannot be empty.
+    last_block: Hash,
+    /// rest of the blocks in the chunk.
+    blocks: Vec<Hash>,
```
barakshani: inner blocks?

crates/alpen-ee/common/src/types/ee_update.rs
```diff
@@ -0,0 +1,88 @@
...
+use crate::ProofId;
+
+/// Unique identifier for an EeUpdate
+#[derive(Debug)]
+pub struct EeUpdateId {
+    pub prev_block: Hash,
+    pub last_block: Hash,
+}
+
+/// DA related data relevant to EeUpdate
```
barakshani:
```suggestion
/// EeUpdate DA-related data
```

crates/alpen-ee/common/src/types/ee_update.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/// EeUpdate lifecycle states
+#[derive(Debug)]
+pub enum EeUpdateStatus {
+    /// Newly created
+    Init,
+    /// DA started, waiting for block
+    DaPending { txids: Vec<Txid> },
+    /// DA complete
```
barakshani: please explain better, I don't understand what _data availability started/complete_ means.

crates/alpen-ee/common/src/types/ee_update.rs
```diff
@@ -0,0 +1,88 @@
...
+pub struct EeUpdate {
+    /// Sequential update index, also used in
+    idx: u64,
+    /// last block of (idx - 1)th update
+    prev_block: Hash,
+    /// last block in this update
+    last_block: Hash,
+    /// rest of the blocks in this update.
+    /// cached here for easier processing.
+    blocks: Vec<Hash>,
```
barakshani: inner?

crates/alpen-ee/common/src/types/exec_record.rs
```diff
@@ -21,6 +21,8 @@ struct ExecPackageMetadata {
     /// 2. This does not uniquely identify a package or exec block. One `ol_block` can be linked
     ///    with multiple records.
     ol_block: OLBlockCommitment,
+    /// Next inbox index at this ol_block.
```
barakshani:
```suggestion
    /// Inbox next message index at this ol_block.

File: crates/alpen-ee/common/src/types/ee_update.rs (hunk 7 above)
Comment (sapinb) [reply crates/alpen-ee/common/src/types/ee_update.rs]: fixed

File: crates/alpen-ee/common/src/types/chunk.rs (hunk 15 above)
Comment (sapinb) [reply crates/alpen-ee/common/src/types/chunk.rs]: Afaik, the previous "interesting" bugs we because we used `proof: Vec<u8>`, with empty representing both "not proven" and "proven with empty proof" states. I represented these as different enum variants to avoid this exactly.
Also, saving an id to the proof here, and not the actual proof bytes, but that probably does not make much difference.

Comment (sapinb) [reply crates/alpen-ee/common/src/types/chunk.rs]: Regarding construction ergonomics, I dont really see the issues. 
Outside tests, it will be constructed in exactly 1 location in code and stored to db immediately. 
Updates will be load from db -> update -> save to db.
Although needing atleast 1 block is a bit awkward, it does enforce the requirement in the type itself, instead of runtime checks.

File: crates/alpen-ee/common/src/types/chunk.rs (hunk 16 above)
Comment (sapinb) [reply crates/alpen-ee/common/src/types/chunk.rs]: This is meant to represent a unique, deterministic id to the chunk across reorgs. So it is a function of last blockhash of previous chunk and the last blockhash of current chunk.
I can use something like sha256([prev_block, last_block]) as the id, but this works just as well. 
The fields likely dont need to be pub, so they can serve as as an opaque identifier outside the storage module.

File: crates/alpen-ee/common/src/types/ee_update.rs (hunk 18 above)
Comment (sapinb) [reply crates/alpen-ee/common/src/types/ee_update.rs]: Same case as my reply for chunk id above.

File: crates/alpen-ee/sequencer/src/ol_chain_tracker/state.rs (hunk 21 above)
Comment (sapinb) [reply crates/alpen-ee/sequencer/src/ol_chain_tracker/state.rs:13]: Due to need to handle reorgs OL, the inbox messages are used (included during EE block building) as groups of messages per OL block instead of as individual messages. So this idx isnt really used being in the block assembly process. 
But it is still needed during proving and posting EE updates according to current design, so fetching and caching this along with the block metadata.
This is only being cached, so using the same names as in ol and updates to avoid confusion.

File: crates/alpen-ee/common/src/types/ee_update.rs (hunk 31 above)
Comment (sapinb) [reply crates/alpen-ee/common/src/types/ee_update.rs]: updated comments.

File: crates/alpen-ee/common/src/traits/prover.rs (hunk 2 above)
Comment (sapinb) [reply crates/alpen-ee/common/src/traits/prover.rs]: updated.

File: crates/alpen-ee/common/src/traits/prover.rs (hunk 14 above)
Comment (sapinb) [reply crates/alpen-ee/common/src/traits/prover.rs]: Added this comment in https://github.com/alpenlabs/alpen/pull/1214/changes#diff-bde12f64bba3bb73a5ec5f448fed6ef328895e81028787a91af780ad0ce39f10R20-R21

File: crates/alpen-ee/common/src/traits/prover.rs [hunk 34]
```diff
@@ -0,0 +1,34 @@
...
+}
+
+/// Interface between Prover and Batch assembly
+#[async_trait]
+pub trait BatchProver: Sized {
+    /// Request proof generation for batch_id.
+    /// Ok(()) -> proof generation has been queued
+    async fn request_proof_generation(&self, batch_id: BatchId) -> eyre::Result<()>;
+
+    /// Check if proof is generated for batch_id.
```
barakshani, line 24: this is to specific, no? We just fetch the status afaict

crates/alpen-ee/common/src/traits/storage/batch.rs
```diff
@@ -0,0 +1,80 @@
+use async_trait::async_trait;
+
+use crate::{Batch, BatchId, BatchStatus, Chunk, ChunkId, ChunkStatus, StorageError};
+
+/// Storage for Batches and Chunks
+///
+/// A batch is a contiguous group of blocks. One batch corresponds to one Account Update Operation
+/// sent to OL, and batch size should typically be limited by DA size constraints.
```
barakshani, line 8: >and batch size should typically be limited by DA size constraints

what size constraints?

crates/alpen-ee/common/src/traits/storage/batch.rs
```diff
@@ -0,0 +1,80 @@
...
+///
+/// Chunks are subsets of blocks in a batch, and units for proving ee chain execution, typically
+/// limited by prover constraints.
+///
+/// An Batch will consist of 1 or more chunks.
```
barakshani:
```suggestion
/// A batch consists of 1 or more chunks.

File: crates/alpen-ee/common/src/traits/storage/batch.rs [hunk 37]
```diff
@@ -0,0 +1,80 @@
...
+/// These chunks can be generated before or after updates, but the chunks must fit inside Batch
```
barakshani:
```suggestion
/// These chunks can be generated before or after updates, but the chunks must fit inside batch
```
Can you clarify the relation to update here?

crates/alpen-ee/common/src/types/batch.rs
```diff
@@ -0,0 +1,110 @@
...
+/// Batch lifecycle states
+#[derive(Debug, Clone)]
+pub enum BatchStatus {
+    /// Newly created
+    Init,
+    /// DA txn(s) posted, waiting for inclusion in block.
+    DaPending { txns: Vec<(Txid, Wtxid)> },
+    /// DA txn(s) included in block(s).
+    DaComplete { da: Vec<L1DaBlockRef> },
+    /// Proving started, waiting for proof generation.
```
barakshani, line 44: there's a hybrid scenario where some txns are in blocks while other are not. how do you suggest we handle this?

crates/alpen-ee/common/src/types/batch.rs
```diff
@@ -0,0 +1,110 @@
...
+    ProofReady {
+        da: Vec<L1DaBlockRef>,
+        proof: ProofId,
+    },
+}
+
+/// Represents a sequence of blocks that are treated as a unit for DA and posting updates to OL.
+#[derive(Debug)]
+pub struct Batch {
+    /// Sequential update index, also used in
```
barakshani: in?

crates/alpen-ee/common/src/types/batch.rs
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    /// Get deterministic id.
+    pub fn id(&self) -> BatchId {
+        BatchId::new(self.prev_block, self.last_block)
+    }
+
+    /// Get sequential index.
+    /// This should equal the sequence number in account update sent to OL.
```
barakshani: mmm..... this makes me wonder. Would we never have any other tx targeting a snark account which is not an "update" (batch update + proof)?
If not, why do we even need sequence number?

In addition, if at any point in the future we would also invalid updates (e.g. update where the snark fails, but sequencer still collects fee), then this index will not necessarily equal the sequence number.

crates/alpen-ee/common/src/types/batch.rs
```diff
@@ -0,0 +1,110 @@
...
+    pub fn idx(&self) -> u64 {
+        self.idx
+    }
+
+    /// last block of the previous Batch.
```
barakshani: why not "batch"?

crates/alpen-ee/common/src/types/batch.rs
```diff
@@ -0,0 +1,110 @@
...
+        proof: ProofId,
+    },
+}
+
+/// Represents a sequence of blocks that are treated as a unit for DA and posting updates to OL.
+#[derive(Debug)]
+pub struct Batch {
+    /// Sequential update index, also used in
+    idx: u64,
+    /// last block of (idx - 1)th update
```
barakshani:
```suggestion
    /// last block of (idx - 1)th batch, that this batch extends.

File: crates/alpen-ee/common/src/types/chunk.rs [hunk 43]
```diff
@@ -0,0 +1,86 @@
...
+        }
+    }
+}
+
+/// Represents a sequence of blocks that are processed together as a unit during proving.
+#[derive(Debug)]
+pub struct Chunk {
+    /// Sequential chunk index
+    idx: u64,
+    /// last block of (idx - 1)th chunk that this chunk extends
```
barakshani:
```suggestion
    /// last block of (idx - 1)th chunk, that this chunk extends.

File: crates/alpen-ee/common/src/types/chunk.rs [hunk 44]
```diff
@@ -0,0 +1,86 @@
...
+        debug_assert_ne!(prev_block, last_block);
+        Self {
+            idx,
+            prev_block,
+            last_block,
+            inner_blocks,
+        }
+    }
+
+    /// Deterministic chunk id
```
barakshani: for consistency, add periods or remove them at the end of the sentence.

crates/alpen-ee/sequencer/src/block_builder/task.rs
```diff
@@ -198,7 +212,17 @@ async fn build_next_block(
         .await
         .context("build_next_block: failed to get best exec block")?;
 
-    // validate blocktime constraint
+    // check if last local block is not as expected from previous block building cycle
```
barakshani:
```suggestion
    // Check if last local block is not as expected from previous block building cycle.

File: crates/alpen-ee/common/src/types/prover.rs [hunk 46]
```diff
@@ -0,0 +1,9 @@
+use strata_acct_types::Hash;
+
+/// Unique identifier to a proof.
+pub type ProofId = Hash;
+
+// TODO: proper proof type
+#[expect(dead_code, reason = "wip")]
+#[derive(Debug)]
```
delbonis, line 8: Shouldn't be `Debug` imo, it's just entropy.  Or make it just emit `Proof(...)` or something.

PR #1112, opened 2025-10-29
fix!(ol): remove leftover bridge bookeeping

storopoli:
## Description

Remove all bridge related leftover bookeeping.

Needs #1219

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The bridge bookkeeping has been successfully moved to the ASM subprotocols
in #1072, but there's one remaining artifact: the `total_btc_bridged` field in
`OLState`. This field appears to be dead code that was left behind during the
migration

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [x] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [x] New and existing tests pass with my changes.
- [x] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/.github/CONTRIBUTING.md#ai-assistance-notice)
      in the body of this PR.

## Related Issues

STR-1315

PR #1219, opened 2025-12-19
feat!: Remove old dataflow and migrate to ASM based Block building

storopoli:
## Description

Removes the Old `btcio`-driven Dataflow (`L1BlockManifest`) for the new ASM-drive Dataflow (`AsmManifest`).

This is effectively replacing the old EE-OL bundled code with the new EE <> OL separation code.

It also removes all the bridge bookkeeping from OL (migrated over #1112).

TODO:

- [ ] functional tests are a mess

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

PR assisted by Cursor AI with Claude Code.

PSA: DON'T TRY TO REVIEW THIS COMMIT BY COMMIT. It's a mess. just review the whole set of changes. 😢 

## Checklist

## Related Issues

STR-1796
STR-1315

### DIFF
--- a/crates/sequencer/src/block_template/block_assembly.rs
+++ b/crates/sequencer/src/block_template/block_assembly.rs
@@ -178,7 +178,7 @@ fn prepare_l1_segment(
     // everything we know about.
     let cur_safe_height = prev_chstate.l1_view().safe_height();
     let cur_next_exp_height = prev_chstate.l1_view().next_expected_height();
-    let l1_verified_block = prev_chstate.l1_view().header_vs().last_verified_block;
+    let l1_verified_block = prev_chstate.l1_view().safe_blkid();
     debug!(%target_height, %cur_safe_height, %cur_next_exp_height, "figuring out which blocks to include in L1 segment");
     debug!(?l1_verified_block, "last verified L1 block");
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14126,7 +14126,6 @@ dependencies = [
  "borsh",
  "rand 0.8.5",
  "strata-asm-common",
- "strata-asm-types",
  "strata-asm-worker",
  "strata-bridge-types",
  "strata-btc-types",
@@ -14685,9 +14684,7 @@ name = "strata-ol-chainstate-types"
 version = "0.3.0-alpha.1"
 dependencies = [
  "arbitrary",
- "bitcoin",
  "borsh",
- "strata-asm-types",
  "strata-bridge-types",
  "strata-identifiers",
  "strata-primitives",
--- a/crates/chaintsn/Cargo.toml
+++ b/crates/chaintsn/Cargo.toml
@@ -20,11 +20,13 @@ strata-predicate.workspace = true
 strata-primitives.workspace = true
 strata-state.workspace = true
 
-bitcoin.workspace = true
 rand_chacha.workspace = true
 rand_core.workspace = true
 thiserror.workspace = true
 tracing = { workspace = true, optional = true }
+
+[dev-dependencies]
+bitcoin.workspace = true
 
 [features]
 default = ["fullstd"]
--- a/crates/chaintsn/src/checkin.rs
+++ b/crates/chaintsn/src/checkin.rs
@@ -8,7 +8,7 @@ use strata_asm_logs::{
 use strata_bridge_types::DepositIntent;
 use strata_ol_chain_types::L1Segment;
 use strata_params::RollupParams;
-use strata_primitives::l1::BitcoinAmount;
+use strata_primitives::l1::{BitcoinAmount, L1BlockCommitment};
 
 use crate::{
     context::{AuxProvider, ProviderError, ProviderResult, StateAccessor},
@@ -70,6 +70,9 @@ pub fn process_l1_view_update<'s, S: StateAccessor>(
 
     // If there's no new blocks we can abort.
     let new_tip_height = prov.get_l1_tip_height();
+    dbg!(&new_tip_height);
+    dbg!(&l1v.safe_height());
+
     if prov.get_l1_tip_height() == l1v.safe_height() {
         return Ok(false);
     }
@@ -93,6 +96,11 @@ pub fn process_l1_view_update<'s, S: StateAccessor>(
         // We don't need to validate headers here anymore
 
         process_asm_logs(state, &mf)?;
+
+        // Advance the verified L1 tip to the latest manifest we've processed.
+        let verified_blk = L1BlockCommitment::from_height_u64(mf.height(), *mf.blkid())
+            .expect("height should be valid");
+        state.update_verified_blk(verified_blk);
     }
 
     // If prev_finalized_epoch is null, i.e. this is the genesis batch, it is
@@ -150,9 +158,9 @@ fn process_l1_checkpoint<'s, S: StateAccessor>(
     state: &mut FauxStateCache<'s, S>,
     ckpt_update: &CheckpointUpdate,
 ) -> Result<(), OpError> {
+    debug!(?ckpt_update, "observed l1 checkpoint");
     let new_fin_epoch = ckpt_update.epoch_commitment();
     state.inner_mut().set_finalized_epoch(new_fin_epoch);
-    trace!(?new_fin_epoch, "observed finalized checkpoint");
     Ok(())
 }
 
--- a/crates/chaintsn/src/legacy.rs
+++ b/crates/chaintsn/src/legacy.rs
@@ -1,8 +1,7 @@
 //! Legacy routines extracted from `StateCache`.
 
-use bitcoin::block::Header;
-use strata_asm_types::L1VerificationError;
 use strata_bridge_types::DepositIntent;
+use strata_identifiers::L1BlockCommitment;
 use strata_ol_chainstate_types::Chainstate;
 
 use crate::context::StateAccessor;
@@ -34,11 +33,8 @@ impl<'s, S: StateAccessor> FauxStateCache<'s, S> {
     }
 
     /// Update HeaderVerificationState
-    pub fn update_header_vs(&mut self, header: &Header) -> Result<(), L1VerificationError> {
-        self.state_mut()
-            .l1_view_mut()
-            .header_vs_mut()
-            .check_and_update(header)
+    pub fn update_verified_blk(&mut self, blk: L1BlockCommitment) {
+        self.state_mut().l1_view_mut().update_verified_blk(blk);
     }
 
     /// Writes a deposit intent into an execution environment's input queue.
--- a/crates/consensus-logic/Cargo.toml
+++ b/crates/consensus-logic/Cargo.toml
@@ -8,7 +8,6 @@ workspace = true
 
 [dependencies]
 strata-asm-common.workspace = true
-strata-asm-types.workspace = true
 strata-asm-worker.workspace = true
 strata-bridge-types.workspace = true
 strata-btc-types.workspace = true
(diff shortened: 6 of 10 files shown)


bin/strata-client/src/rpc_server.rs
```diff
@@ -110,6 +112,23 @@ impl StrataRpcImpl {
...
             .await
     }
+
+    fn fetch_bridge_state_from_asm(&self) -> Result<BridgeV1State, Error> {
+        let opt = self
+            .storage
+            .asm()
+            .fetch_most_recent_state()
+            .map_err(Error::Db)?;
+        let (_blk, asm_state) =
```
barakshani, line 122: out of curiosity, why not `(_, asm_state)`?

bin/strata-client/src/rpc_server.rs
```diff
@@ -396,56 +415,51 @@ impl StrataApiServer for StrataRpcImpl {
...
+        let dep = bridge
+            .deposits()
             .get_deposit(deposit_id)
             .ok_or(Error::UnknownIdx(deposit_id))
-            .map(RpcDepositEntry::from_deposit_entry)?)
+            .map_err(to_jsonrpsee_error("deposit not found"))?;
+
+        // Convert ASM notary bitmap to operator idx vec expected by bridge-types
+        let notary_ops: Vec<OperatorIdx> = dep.notary_operators().active_indices().collect();
+        let shim = DepositEntry::new(dep.idx(), *dep.output(), notary_ops, dep.amt(), None);
```
barakshani: Can you explain to me, despite the existing comment, why this is needed? `dep` is already `Option<&DepositEntry>`, why can't you unwrap (or whatever you find more suitable) and clone(?) and apply `RpcDepositEntry::from_deposit_entry` on it?

bin/strata-client/src/rpc_server.rs
```diff
@@ -396,56 +415,51 @@ impl StrataApiServer for StrataRpcImpl {
...
+        Ok(RpcDepositEntry::from_deposit_entry(&shim))
     }
 
     async fn get_cur_withdrawal_assignments(&self) -> RpcResult<Vec<RpcWithdrawalAssignment>> {
```
barakshani:
```suggestion
    async fn get_current_withdrawal_assignments(&self) -> RpcResult<Vec<RpcWithdrawalAssignment>> {
```
for consistency, or change the one above.

crates/asm/manifest-types/src/manifest.rs
```diff
@@ -8,14 +10,25 @@ use crate::{
 
 impl AsmManifest {
     /// Creates a new ASM manifest.
-    pub fn new(blkid: L1BlockId, wtxids_root: WtxidsRoot, logs: Vec<AsmLogEntry>) -> Self {
+    pub fn new(
```
barakshani, line 13: Are manifest extracted from a single block of multiple blocks?
I think that previously it was a single manifest per BTC scan (i.e. a range of L1 blocks), but now it's per L1 block, but would like to double check.
Asking because I want to see if it makes sense to make it not a single height/blkid, but vector instead.

crates/btcio/src/reader/query.rs
```diff
@@ -247,17 +226,9 @@ async fn poll_for_new_blocks<R: Reader>(
...
-                if let Some(checkpt) = find_checkpoint_in_event(&ev) {
-                    // if we have a checkpoint in this block, update filterconfig based on this
-                    let chainstate = borsh::from_slice(checkpt.checkpoint().sidecar().chainstate())
-                        .expect("deserialize chainstate");
-
-                    state
-                        .filter_config_mut()
-                        .update_from_chainstate(&chainstate);
-                }
+                // Note: checkpoint detection is now handled by the ASM STF via logs,
```
barakshani:
```suggestion
                // Note: Checkpoint detection is now handled by the ASM STF via logs,
```
consistency

bin/strata-dbtool/src/cmd/client_state.rs
```diff
@@ -30,10 +30,33 @@ pub(crate) fn get_client_state_update(
...
-        .ok_or(DisplayedError::InternalError("".to_string(), Box::new(())))?
-        .into();
+    let block_mf = get_l1_block_manifest(db, block_id)?.ok_or(DisplayedError::InternalError(
+        "Block manifest not found".to_string(),
+        Box::new(()),
+    ))?;
+
+    // Find the height for this block ID by searching the canonical chain
+    let (tip_height, _) = get_l1_chain_tip(db)?;
+    let height = (0..=tip_height)
```
barakshani: I'm not so happy with this, but can we at least traverse from (up to, actually) the Strata genesis block?

bin/strata-dbtool/src/cmd/l1.rs
```diff
@@ -92,24 +93,27 @@ pub(crate) fn get_l1_manifest(
...
-    // Create the output data structure
-    let block_info = L1BlockInfo {
-        block_id: &block_id,
-        transactions: l1_block_manifest.txs(),
-        height: l1_block_manifest.height(),
-        transaction_infos,
-    };
+    // Find the height for this block ID by searching the canonical chain
+    let (tip_height, _) = get_l1_chain_tip(db)?;
+    let height = (0..=tip_height)
```
barakshani: as above

crates/chaintsn/src/transition.rs
```diff
@@ -115,25 +92,8 @@ fn process_proto_op(
...
-            process_withdrawal_fulfillment(state, info)?;
-        }
-
-        ProtocolOperation::DepositSpent(info) => {
-            process_deposit_spent(state, info)?;
-        }
-
-        // Other operations we don't do anything with for now.
-        _ => {}
+    if let ProtocolOperation::Checkpoint(ckpt) = &op {
```
barakshani, line 95: will this be removed at some point?

crates/consensus-logic/src/asm_worker_context.rs
```diff
@@ -44,15 +44,30 @@ impl AsmWorkerCtx {
 
 impl WorkerContext for AsmWorkerCtx {
     fn get_l1_block(&self, blockid: &L1BlockId) -> WorkerResult<bitcoin::Block> {
-        let l1_mf = self
-            .l1man
-            .get_block_manifest(blockid)
-            .map_err(conv_db_err)?
-            .ok_or(WorkerError::MissingL1Block(*blockid))?;
+        // With ASM manifests, we don't store height in the manifest anymore
+        // We need to search the canonical chain to find the height
```
barakshani, line 48: why don't we? I mean, why not adding it?

crates/sequencer/src/block_template/block_assembly.rs
```diff
@@ -251,54 +252,54 @@ fn prepare_l1_segment(
...
             continue;
         };
 
-        // Must get expected checkpoint.
-        if expected.commitment() != checkpoint.commitment() {
-            warn!(got = ?checkpoint, ?expected, "got unexpected checkpoint");
-            continue;
+        // Check if the ack epoch matches our expected checkpoint
+        if ack_data.epoch().epoch() == expected.batch_info().epoch() {
+            // Found checkpoint ack for expected epoch. Should end current epoch.
```
barakshani, line 289: >Should end current epoch.

What do you mean by that? And given that we (only) return `true`, what is expected to happen next?

bin/strata-client/src/rpc_server.rs
storopoli: It is the same result, but `_blk` is more informative.

crates/consensus-logic/src/asm_worker_context.rs
storopoli: Yes, that was my initial reaction in 54ccb38c1b12d43b71f37d0c0fd96b8dbcc39baa but I've reverted in b788d50b9113a929ef65f06d8716d231d6f51e1e. Now dropping it so that ew have the `height` field in `AsmManifest`.

bin/strata-client/src/rpc_server.rs
storopoli: Nice catch done in 4c5fc8d38540cddd7a2e0a3fc10b8ddcff0b1bde

bin/strata-dbtool/src/cmd/client_state.rs
storopoli: This is removed now that we have the `height` back in `AsmManifest`.

bin/strata-dbtool/src/cmd/l1.rs
storopoli: This is removed now that we have the `height` back in `AsmManifest`.

crates/sequencer/src/block_template/block_assembly.rs
storopoli: When this returns `true`, the sequencer knows it's safe to end the current epoch because the previous epoch's checkpoint has been confirmed on L1.

I was trying to make the old `sequencer` work with the new ASM stuff.
I know that this is "old" code, but I don't have a better idea.

crates/chaintsn/src/transition.rs
storopoli: Probably, but maybe not in this PR.

crates/sequencer/src/block_template/block_assembly.rs
barakshani: Is it necessary? I would rather not follow this path where we force batch sealing on previous epoch finalisation, but if this is the last thing we need to make us remove the old code I won't stand in our way. We already have a ticket to define OL batch sealing criteria.

bin/strata-client/src/rpc_server.rs
storopoli: Another good catch.
I have removed the old `DepositEntry`s from bridge types and used the `DepositEntry`s from ASM types.

Done in 093c5a5ce608fb032fab8ef5ea49ba10d3d615d5

crates/btcio/src/reader/query.rs
storopoli: I think I've done that change in commit 093c5a5ce608fb032fab8ef5ea49ba10d3d615d5

crates/consensus-logic/src/asm_worker_context.rs
barakshani: I'm confused now. You seem to have agreed to include height in AsmManifest, but here we still traverse over all blocks.

crates/rpc/types/src/types.rs
```diff
@@ -313,29 +313,26 @@ pub struct RpcDepositEntry {
     /// The outpoint that this deposit entry references.
     output: BitcoinOutPoint,
 
-    /// List of notary operators, by their indexes.
-    // TODO convert this to a windowed bitmap or something
-    notary_operators: Vec<OperatorIdx>,
+    /// Historical set of operators that formed the N/N multisig for this deposit.
```
barakshani, line 316:
```suggestion
    /// The set of operators that formed the N/N multisig for this deposit.
```
I know it's the language used in ASM, but I don't really like it.

crates/rpc/types/src/types.rs
```diff
@@ -313,29 +313,26 @@ pub struct RpcDepositEntry {
...
+    ///
+    /// This preserves the specific operators who controlled the multisig when the
+    /// deposit was locked, since the active operator set may change over time.
+    /// Any one honest operator from this set can process user withdrawals.
```
barakshani, line 320:
```suggestion
    /// Any honest operator from this set can process user withdrawals.

PR #1185, opened 2025-12-03
STR-1803 OL Mempool implementation

krsnapaudel:
## Description

Implements the OL mempool for storing pending transactions before block inclusion. Follows a worker-handle pattern with async service framework integration. In some parts, it is inspired by Reth's mempool design.

Updated based on [this design doc](https://www.notion.so/Strata-OL-Mempool-2af901ba000f80fd81a3cbbdb1d850a2?source=copy_link).

### Architecture

**Service Pattern**
- Worker-handle pattern: Internal `MempoolService` runs as async task, external code interacts via `MempoolHandle`
- Two-layer persistence: In-memory `HashMap<OLTxId, MempoolEntry>` for fast access + Sled database backup for crash recovery
- Service framework: Integrates with strata-service for consistent lifecycle management

**Transaction Ordering**
`SnarkAccountUpdate`: Ordered by sequence number (`seq_no`), with `insertion_id` as tiebreaker
`GenericAccountMessage`: FIFO using `first_seen_slot`, with `insertion_id` as tiebreaker

**Per-Account Tracking**
- `account_state`: HashMap<AccountId, AccountMempoolState>` index for efficient per-account operations (including txids and seqnos)

### Validation & State Integration

**`StateAccessor` Integration** 
- Fetches `OLState` from storage for current chain tip
- Per-tip caching: State accessor cached and invalidated on chain updates
- Async API: `get_toplevel_ol_state_async()` via storage manager

**Validation Rules** 
Sequence number validation: `tx.seq_no == account.current_seq_no` (basic validation)
Transaction replacement and gap rejection: `tx.seq_no` must be in the range [`min_seq_no`, `max_seq_no +1`] for account
- **Slot bounds**: Validates `min_slot` and `max_slot` constraints
- **Account existence**: Verifies target account exists in state
- **Account type matching**: `SnarkAccountUpdate` must target Snark account

### Chain Updates & Reorg Handling

**Chain Update Processing**
- `chain_update(new_tip)` handles FCM notifications
- Removes expired txs (`max_slot <= current_slot`)
- Invalidates state accessor cache on tip changes
- Revalidates all transactions

This PR was created with help from Claude Code and Cursor AI.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
[...]

PR #1221, opened 2025-12-21
STR-1999: OL block database and manager

krsnapaudel:
## Description

OL Block Database and manager. Mostly modeled after `L2BlockDatabase`. It's a blocker for both OL Mempool #1185  and OL Block Assembly.

This PR was created with help from Cursor AI.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


crates/db/store-sled/src/ol/db.rs
```diff
@@ -0,0 +1,124 @@
...
+            let valid = blocks
+                .into_iter()
+                .filter_map(|blkid| match self.get_block_status(blkid) {
+                    Ok(Some(BlockStatus::Valid)) => Some(Ok(blkid)),
+                    Ok(_) => None,
+                    Err(e) => Some(Err(e)),
+                })
+                .collect::<Result<Vec<_>, _>>()?;
+
+            // Return the first valid block at the highest slot as the tip.
```
barakshani: what is this function being used for? why is it ok to return the first block?

krsnapaudel: This is meant to return the tip block id. The idea of vec for block ids at slot is to support forks. For now, we are assuming the vec will have length 1.

crates/db/store-sled/src/ol/db.rs
```diff
@@ -0,0 +1,127 @@
...
+            // This is fine because we expect only one valid block per slot (no reorgs in OL chain).
+            if let Some(id) = valid.first().cloned() {
+                return Ok(id);
+            }
+
+            if slot == 0 {
```
storopoli: If `put_block_data` is called with the same commitment twice, it overwrites the status back to `Unchecked` even if the block was previously marked as `Valid` or `Invalid`. This could cause issues if a block is re-inserted after validation.

We should check existing status before overwriting

storopoli: It might be even possible to reproduce on tests.

crates/db/store-sled/src/ol/db.rs
```diff
@@ -0,0 +1,127 @@
...
+                    bht.insert(&slot, &block_height_data)?;
+                    Ok(())
+                },
+            )
+            .map_err(to_db_error)?;
+        Ok(())
+    }
+
+    fn set_block_status(&self, id: OLBlockId, status: BlockStatus) -> DbResult<()> {
+        // Check if block exists by trying to find it via commitment
```
storopoli: Either implement the check or remove the misleading comment

crates/db/store-sled/src/ol/db.rs
krsnapaudel: put logic is identical to what L2 db does, but I see your point. So not updating status if block exists.

storopoli (approved):
ACK af54211829532f28dd0dc17477aead78189bcbba

crates/db/store-sled/src/ol/db.rs
barakshani: gotta say, I don't see how `put_block_data` is related to this function -- can you explain?
was just interested because the initial comment, but got lost...

crates/db/types/src/traits.rs
```diff
@@ -480,3 +482,34 @@ pub trait OLStateDatabase: Send + Sync + 'static {
...
+
+    /// Sets the block's validity status.
+    fn set_block_status(&self, commitment: OLBlockCommitment, status: BlockStatus) -> DbResult<()>;
+
+    /// Gets the OL block IDs that we have at some slot, in case there's more
+    /// than one on competing forks.
+    fn get_blocks_at_height(&self, slot: u64) -> DbResult<Vec<OLBlockId>>;
+
+    /// Gets the validity status of a block.
+    fn get_block_status(&self, commitment: OLBlockCommitment) -> DbResult<Option<BlockStatus>>;
```
delbonis: Most of these that operate in terms of `OLBlockCommitment` can just use `OLBlockId`, which is more pure key.  Especially `put_block_data`, since the block contains the header which contains the slot already.

Is there some reason otherwise?  I can see it making `get_blocks_at_height` a little bit easier to implement.

crates/db/store-sled/src/ol/db.rs
delbonis: I agree that maybe we should be just talking about it in terms of determining the highest slot and then we can let the FCM pick the best tip from the blocks near the tip.

delbonis (changes requested):
Implementation-wise looks fine but have some questions about exactly the the interface we're exposing.

crates/db/store-sled/src/ol/schemas.rs
```diff
@@ -0,0 +1,41 @@
...
+    define_table_with_default_codec, define_table_with_integer_key, define_table_without_codec,
+    impl_borsh_key_codec, impl_borsh_value_codec,
+};
+
+define_table_without_codec!(
+    /// A table to store OL Block data. Maps block commitment to Block
+    (OLBlockSchema) OLBlockCommitment => OLBlock
+);
+
+impl_borsh_key_codec!(OLBlockSchema, OLBlockCommitment);
```
bewakes: I am not sure why there is two steps for table definition. I believe this can be done with `define_table_with_default_codec!`. I can see `impl_borsh_key_codec!` has better error handling, but I believe we can just use `impl_borsh_key_codec!` inside `define_table_with_default_codec!`.

Also, since `OLBlockCommitment` contains an integer field, we should most likely define this with `define_table_with_seek_key_codec`.

crates/db/store-sled/src/ol/schemas.rs
```diff
@@ -0,0 +1,41 @@
...
+
+define_table_with_default_codec!(
+    /// A table to store OL Block status. Maps block commitment to BlockStatus
+    (OLBlockStatusSchema) OLBlockCommitment => BlockStatus
```
bewakes: We won't most likely have to iterate lexicographically on this. But to be safe, we can use `define_table_with_seek_key_codec!`.

crates/db/store-sled/src/ol/schemas.rs
bewakes: I see below that `OLBlock` has a ssz based ValueCodec and thus two steps of table definition. Still might need to consider "seekable" key encoding.

crates/db/store-sled/src/ol/db.rs
```diff
@@ -0,0 +1,139 @@
...
+                    bht.insert(&slot, &blocks_at_slot)?;
+                    Ok(true)
+                },
+            )
+            .map_err(to_db_error)
+    }
+
+    fn set_block_status(&self, commitment: OLBlockCommitment, status: BlockStatus) -> DbResult<()> {
+        // Check if block exists before setting status
+        if self.get_block_data(commitment)?.is_none() {
```
bewakes: Should we error out here? Or at least change the return value to indicate whether the status update succeeded or not. Because now we won't know whether we were actually able to update the status or not.

crates/storage/src/managers/ol.rs
```diff
@@ -0,0 +1,433 @@
...
+        commitment: OLBlockCommitment,
+        status: BlockStatus,
+    ) -> DbResult<()> {
+        self.ops.set_block_status_blocking(commitment, status)?;
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
bewakes, line 127: Great work here! Haven't seen db managers being tested before.

crates/storage/src/managers/ol.rs
```diff
@@ -0,0 +1,433 @@
...
+        get_mock_block_with_slot_and_id(slot, 0)
+    }
+
+    fn get_mock_block_with_slot_and_id(slot: u64, id_byte: u8) -> OLBlock {
+        // Use id_byte to create different body_root to ensure different block IDs
+        let mut body_root_bytes = [0u8; 32];
+        body_root_bytes[0] = id_byte;
+        let body_root = Buf32::from(body_root_bytes);
+
+        let header = OLBlockHeader::new(
```
bewakes: Looks like we can use `Arbitrary` here.

crates/db/types/src/traits.rs
krsnapaudel: Not really. I will update to use `OLBlockId`.

crates/db/store-sled/src/ol/schemas.rs
krsnapaudel: Probably not relevant with `OLBlockId`.

crates/db/store-sled/src/ol/db.rs
krsnapaudel: Adding return bool.

crates/storage/src/managers/ol.rs
krsnapaudel: Yes but we need a lot of `Arbitrary` impls I think. Instead reusing test strategies from `ol/chain-types`.

crates/db/types/src/traits.rs
```diff
@@ -473,3 +475,35 @@ pub trait OLStateDatabase: Send + Sync + 'static {
...
+
+    /// Gets the OL block IDs that we have at some slot, in case there's more
+    /// than one on competing forks.
+    fn get_blocks_at_height(&self, slot: u64) -> DbResult<Vec<OLBlockId>>;
+
+    /// Gets the validity status of a block.
+    fn get_block_status(&self, id: OLBlockId) -> DbResult<Option<BlockStatus>>;
+
+    /// Returns the latest valid OL block ID, or an error at genesis or when no valid block exists.
+    fn get_tip_block(&self) -> DbResult<OLBlockId>;
```
delbonis: As discussed, should just return a slot and let the FCM pick the block.  As-is, this makes the database have to have some way to decide what the tip block canonically is, but that's not its responsibility.

storopoli (approved):
ACK f1919fa7bda3d5ef3a32fc9c428e47e1fe86a33f

delbonis (approved):
I'm not sure if it makes sense for the database layer to care about if blocks are marked as valid or not when determining what the tip slot is, but approving this anyways.

PR #1053, opened 2025-09-24
refactor: move asm-related types to asm-types

krsnapaudel:
## Description

A separate crate for `asm-types`.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

WIP

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have commented my code where necessary.
- [ ] I have updated the documentation if needed.
- [ ] My changes do not introduce new warnings.
- [ ] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [ ] New and existing tests pass with my changes.
- [ ] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/.github/CONTRIBUTING.md#ai-assistance-notice)
      in the body of this PR.

PR #1222, opened 2025-12-21
feat!: Remove old dataflow and migrate to ASM based Block building

prajwolrg:
## Description

Removes the Old `btcio`-driven Dataflow (`L1BlockManifest`) for the new ASM-drive Dataflow (`AsmManifest`).

This is effectively replacing the old EE-OL bundled code with the new EE <> OL separation code.

It also removes all the bridge bookkeeping from OL (migrated over #1112).

ASM Worker has to do some special handling for the Genesis block manifest.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

PR assisted by Cursor AI with Claude Code.

## Checklist

## Related Issues

STR-1796
STR-1315

### DIFF
--- a/functional-tests/utils/utils.py
+++ b/functional-tests/utils/utils.py
@@ -17,8 +17,13 @@ import base58
 from bitcoinlib.services.bitcoind import BitcoindClient
 
 from factory.config import BitcoindConfig
-from factory.seqrpc import JsonrpcClient
+from factory.seqrpc import JsonrpcClient, RpcError
 from utils.constants import *
+
+# RPC error codes from crates/rpc/types/src/errors.rs
+RPC_ERROR_MISSING_ASM_STATE = -32619  # ASM state not found
+RPC_ERROR_MISSING_BRIDGE_V1_SECTION = -32620  # BridgeV1 section not found in ASM state
+RPC_ERROR_BRIDGE_V1_DECODE_ERROR = -32621  # Failed to decode BridgeV1 state
 
 
 def generate_jwt_secret() -> str:
@@ -857,7 +862,6 @@ def retry_rpc_with_asm_backoff(
     rpc_fn: Callable[[], T],
     timeout: int = 30,
     step: float = 1.0,
-    asm_error_messages: list[str] | None = None,
 ) -> T:
     """
     Retry an RPC call with backoff when ASM state is not yet available.
@@ -870,7 +874,6 @@ def retry_rpc_with_asm_backoff(
         rpc_fn: The RPC function to call
         timeout: Maximum time to retry (seconds)
         step: Time between retries (seconds)
-        asm_error_messages: List of error messages that indicate ASM is not ready
 
     Returns:
         The result of the RPC call
@@ -879,24 +882,23 @@ def retry_rpc_with_asm_backoff(
         AssertionError: If the timeout is reached without success
         Exception: Any non-ASM-related exceptions from the RPC call
     """
-    if asm_error_messages is None:
-        asm_error_messages = [
-            "ASM state not found",
-            "BridgeV1 section not found",
-            "failed to load BridgeV1 state",
-            "No ASM state found",
-        ]
 
     def predicate():
         try:
             return rpc_fn()
-        except Exception as e:
-            error_str = str(e)
-            # Check if this is an ASM-not-ready error
-            if any(msg in error_str for msg in asm_error_messages):
-                logging.debug(f"ASM not ready, will retry: {error_str}")
+        except RpcError as e:
+            # Check if this is an ASM-not-ready error using specific error codes
+            if e.code in (
+                RPC_ERROR_MISSING_ASM_STATE,
+                RPC_ERROR_MISSING_BRIDGE_V1_SECTION,
+                RPC_ERROR_BRIDGE_V1_DECODE_ERROR,
+            ):
+                logging.debug(f"ASM not ready (error code {e.code}), will retry: {e}")
                 return None  # Signal to keep retrying
             # Re-raise if it's a different error
+            raise
+        except Exception:
+            # Re-raise non-RPC errors
             raise
 
     return wait_until_with_value(
--- a/functional-tests/utils/wait/strata.py
+++ b/functional-tests/utils/wait/strata.py
@@ -3,6 +3,33 @@ from typing import Any
 from factory.seqrpc import RpcError
 from utils.wait.base import RpcWaiter
 
+# RPC error codes from crates/rpc/types/src/errors.rs
+RPC_ERROR_MISSING_ASM_STATE = -32619  # ASM state not found
+RPC_ERROR_MISSING_BRIDGE_V1_SECTION = -32620  # BridgeV1 section not found in ASM state
+RPC_ERROR_BRIDGE_V1_DECODE_ERROR = -32621  # Failed to decode BridgeV1 state
+
+
+def is_asm_not_ready_error(error: RpcError) -> bool:
+    """
+    Check if an RPC error indicates that ASM state is not ready.
+
+    ASM state errors are returned with specific error codes:
+    - -32619: ASM state not found
+    - -32620: BridgeV1 section not found in ASM state
+    - -32621: Failed to decode BridgeV1 state
+
+    Args:
+        error: The RPC error to check
+
+    Returns:
+        True if the error indicates ASM is not ready, False otherwise
+    """
+    return error.code in (
+        RPC_ERROR_MISSING_ASM_STATE,
+        RPC_ERROR_MISSING_BRIDGE_V1_SECTION,
+        RPC_ERROR_BRIDGE_V1_DECODE_ERROR,
+    )
+
 
 class StrataWaiter(RpcWaiter):
     def wait_until_genesis(self, message: str | None = None):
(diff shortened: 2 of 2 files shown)


crates/chaintsn/src/checkin.rs
```diff
@@ -70,6 +70,9 @@ pub fn process_l1_view_update<'s, S: StateAccessor>(
 
     // If there's no new blocks we can abort.
     let new_tip_height = prov.get_l1_tip_height();
+    dbg!(&new_tip_height);
+    dbg!(&l1v.safe_height());
```
storopoli: I think we can remove these now no?

crates/asm/logs/src/deposit.rs
```diff
@@ -12,16 +12,16 @@ pub struct DepositLog {
     /// Amount in satoshis.
     pub amount: u64,
     /// Serialized address for the operation.
-    pub addr: VarVec<u8>,
+    pub addr: [u8; 20],
```
storopoli, line 15: We will have to change this in the future to support multiple EEs, but it's fine for now for Alpen's EVM only support.

barakshani: I'd say let's address this soon. The bridge deposits and ASM are going to support the general format from day 1, so it's better not to forget this.

crates/btc-types/src/block.rs
barakshani, line 1: Are you sure this isn't needed? I see that it was added by @krsnapaudel in [here](https://github.com/alpenlabs/alpen/pull/1053/changes#diff-816ac8d7bb9363749f06f4ea64708cf33c17eaa7f679e4b91821739825043a9b) -- maybe he can shed more light.

crates/sequencer/src/block_template/block_assembly.rs
```diff
@@ -221,6 +222,8 @@ fn prepare_l1_segment(
         let Some(rec) = try_fetch_manifest(height, l1man)? else {
             // If we are missing a record, then something is weird, but it would
             // still be safe to abort.
+            // NOTE: This is now expected because the target height is written by btcio handler, but
```
barakshani: Does this note correspond with the previous comment??

crates/btc-types/src/block.rs
storopoli: I think it was a shim, a compatibility layer for the ASM types when they were introduced. As far as I understand, the whole idea behind the new data flow was to remove the old `L1BlockManifest` for `AsmManifest`.

crates/asm/logs/src/deposit.rs
storopoli: @barakshani, I've spoken with @prajwolrg and I'll take this over to finish the merge into #1211 (which is the `target` for this PR). He is finishing up some things in functional tests.

About the multi-EE address support, this is being tracked here: https://alpenlabs.atlassian.net/browse/STR-1950

Hence, we can merge this as it is and updated this later.

I'll do a final pass in this PR and address the current comments in a bit.

crates/sequencer/src/block_template/block_assembly.rs
storopoli: No, I've updated the comment in 8d3114848358525260bffbca9648e3ac45e70519 and moved the log to `DEBUG`

barakshani: >This is expected

What is "this"? The _else_ path?

crates/chaintsn/src/checkin.rs
storopoli: Removed in ba08037b607ab7962be5eae83049e216907fd52a

crates/asm/logs/src/deposit.rs
storopoli: Solving this

crates/btc-types/src/block.rs
storopoli: Solving this for now

crates/asm/manifest-types/ssz/manifest.ssz
```diff
@@ -9,6 +9,9 @@ MAX_LOGS_PER_MANIFEST = 1 << 10
 ### containing the essential commitments and execution logs needed.
 #~# derive: Serialize, Deserialize
 class AsmManifest(Container):
+    ### The L1 block height
+    height: uint64
```
delbonis, line 13: Wanted to avoid including this in the manifest type itself since then we have a bunch of redundant data in the terminal blocks, since all of the manifests are all strictly sequential and we already know the height the first one should be.  Most cases where we're using the manifest we should *already* know the height it corresponds to.

It's not the end of the world but it's a little bit of a data modelling smell and adds another set of checks we probably have to make when processing blocks.

Maybe we could split it up somehow but putting the data and the wrapper with the height in separate places?

crates/sequencer/src/block_template/block_assembly.rs
```diff
@@ -219,9 +220,11 @@ fn prepare_l1_segment(
...
     for height in cur_next_exp_height..=target_height {
         let Some(rec) = try_fetch_manifest(height, l1man)? else {
-            // If we are missing a record, then something is weird, but it would
-            // still be safe to abort.
-            warn!(%height, "missing expected L1 block during assembly");
+            // This is expected: the btcio handler updates the canonical chain tip immediately
+            // when new L1 blocks arrive, but the ASM worker processes blocks asynchronously
+            // to generate manifests. We may be ahead of manifest generation, so just stop here
+            // and include only the blocks we have manifests for.
+            debug!(%height, "L1 manifest not yet available, ASM worker still processing");
```
delbonis: I don't think this is worthy of logging about.  Is it really an *aberrant*/unexpected condition?

crates/sequencer/src/block_template/block_assembly.rs
```diff
@@ -177,9 +178,9 @@ fn prepare_l1_segment(
     // everything we know about.
     let cur_safe_height = prev_chstate.l1_view().safe_height();
     let cur_next_exp_height = prev_chstate.l1_view().next_expected_height();
-    let l1_verified_block = prev_chstate.l1_view().header_vs().last_verified_block;
-    trace!(%target_height, %cur_safe_height, %cur_next_exp_height, "figuring out which blocks to include in L1 segment");
-    trace!(?l1_verified_block, "last verified L1 block");
+    let l1_verified_block = prev_chstate.l1_view().safe_blkid();
+    debug!(%target_height, %cur_safe_height, %cur_next_exp_height, "figuring out which blocks to include in L1 segment");
+    debug!(?l1_verified_block, "last verified L1 block");
```
delbonis: Could we put all this together?

crates/sequencer/src/block_template/block_assembly.rs
delbonis, line 1: Not sure if this exists and I'm just missing it, but it would be a good idea to add logging spans to help with visibility, at least one around the main block assembly function.  That might help remove a couple of logging params.

functional-tests/utils/wait/strata.py
```diff
@@ -321,3 +321,88 @@ def wait_until_latest_checkpoint_at(self, idx: int, timeout: int | None = None):
...
+                error_str = str(e)
+                # Check for ASM/Bridge-related errors
+                if any(
+                    phrase in error_str
+                    for phrase in [
+                        "ASM state",
+                        "BridgeV1 section",
+                        "failed to load BridgeV1",
+                    ]
+                ):
```
delbonis: Lmao I love the ingenuity with this but maybe should split this check out into a helper fn.  Especially since it exists in the other fn too.

bin/strata-client/src/rpc_server.rs
```diff
@@ -110,6 +112,23 @@ impl StrataRpcImpl {
             .map_err(Error::Db)
             .await
     }
+
+    fn fetch_bridge_state_from_asm(&self) -> Result<BridgeV1State, Error> {
```
bewakes: We need to have these in the new binary as well. Would it make more sense to have these in the new binary since we will be getting rid of this old one?

functional-tests/utils/wait/strata.py
```diff
@@ -321,3 +321,88 @@ def wait_until_latest_checkpoint_at(self, idx: int, timeout: int | None = None):
...
+                    f"Bridge operational: {len(deposits)} deposits, "
+                    f"{len(withdrawals)} withdrawals, {len(operators)} operators"
+                )
+                return True
+            except RpcError as e:
+                error_str = str(e)
+                # Check for ASM/Bridge-related errors
+                if any(
+                    phrase in error_str
+                    for phrase in [
```
bewakes: Is it possible to have error codes instead of phrases here?

functional-tests/utils/utils.py
```diff
@@ -851,3 +851,58 @@ def compile_solidity(source: str, contract_name: str | None = None) -> tuple[lis
...
+
+    This helper function wraps RPC calls that depend on ASM state being ready.
+    It will retry the call if it fails with an ASM-related error, allowing
+    time for the ASM worker to process L1 blocks and build state.
+
+    Args:
+        rpc_fn: The RPC function to call
+        timeout: Maximum time to retry (seconds)
+        step: Time between retries (seconds)
+        asm_error_messages: List of error messages that indicate ASM is not ready
```
bewakes: Similar to above, maybe we can use rpc error codes here?

bewakes (commented):
Always good to see LoC removed. Great work!
Ideally we would have these changes for the new binary but looks like it won't be hard to use them/migrate over to the new binary.

storopoli:
Closing in favor of #1223

crates/asm/manifest-types/ssz/manifest.ssz
storopoli: Sorry, let's move this to a ticket. Please feel free to open one if you still think that we should do this in a different way.

PR #1226, opened 2025-12-24
Str 2015 snark msg mmr db

bewakes:
## Description

This PR started off with creating a non-generic mmr db for snark messages for velocity reasons, as it is blocking mempool and block assembly. However, as I kept on enhancing it(which was the initial plan to do in subsequent tickets), this came to be very close to the Generic MMR DB ticket(STR-1202) except for supporting snapshots and variable hash sizes.

On a high level, this adds what's called `UnifiedMmrManager` in `NodeStorage` and stores all the mmr related data differentiated by `MmrId`. To use it for let's say an account in OL stf we can now do something like:
```rust
let mmr_handle: MmrHandle<SnarkMessageEntry> =  
    storage.unified_mmr_manager.get_data_handle(MmrId::Snark(account_id));
mmr_handle.append_blocking(snark_msg); // this will atomically store the entry as well as mmr entry
``` 

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This PR was constructed with the aid of Claude Code.

## Checklist

### DIFF
--- a/crates/consensus-logic/src/asm_worker_context.rs
+++ b/crates/consensus-logic/src/asm_worker_context.rs
@@ -138,25 +138,10 @@ impl WorkerContext for AsmWorkerCtx {
     }
 
     fn get_manifest_hash(&self, index: u64) -> WorkerResult<Option<Hash>> {
-        let num_leaves = self.mmr_handle.get_num_leaves_blocking().map_err(|e| {
-            error!(?e, "Failed to get MMR num_leaves");
+        self.mmr_handle.get_node_blocking(index).map_err(|e| {
+            error!(?e, index, "Failed to get leaf hash from MMR");
             WorkerError::DbError
-        })?;
-
-        if index >= num_leaves {
-            return Err(WorkerError::ManifestIndexOutOfBound {
-                index,
-                max: num_leaves,
-            });
-        }
-
-        self.mmr_handle
-            .get_leaf_hash_blocking(index)
-            .map(Some)
-            .map_err(|e| {
-                error!(?e, index, "Failed to get leaf hash from MMR");
-                WorkerError::DbError
-            })
+        })
     }
 }
 
--- a/crates/db/store-sled/src/global_mmr/db.rs
+++ b/crates/db/store-sled/src/global_mmr/db.rs
@@ -10,7 +10,7 @@ use strata_primitives::buf::Buf32;
 use typed_sled::tree::SledTransactionalTree;
 
 use super::schemas::{
-    GlobalMmrHashIndexSchema, GlobalMmrMetaSchema, GlobalMmrNodeSchema, GlobalMmrPreimageSchema,
+    GlobalMmrHashPositionSchema, GlobalMmrMetaSchema, GlobalMmrNodeSchema, GlobalMmrPreimageSchema,
 };
 use crate::define_sled_database;
 
@@ -18,7 +18,7 @@ define_sled_database!(
     pub struct GlobalMmrDb {
         node_tree: GlobalMmrNodeSchema,
         meta_tree: GlobalMmrMetaSchema,
-        hash_index_tree: GlobalMmrHashIndexSchema,
+        hash_pos_tree: GlobalMmrHashPositionSchema,
         preimage_tree: GlobalMmrPreimageSchema,
     }
 );
@@ -39,18 +39,13 @@ impl GlobalMmrDb {
         })
     }
 
-    fn get_mmr_node(&self, mmr_id: &[u8], pos: u64) -> DbResult<Hash> {
-        self.node_tree.get(&(mmr_id.to_vec(), pos))?.ok_or_else(|| {
-            DbError::Other(format!(
-                "MMR node not found at pos {} for mmr_id {:?}",
-                pos, mmr_id
-            ))
-        })
+    fn get_mmr_node(&self, mmr_id: &[u8], pos: u64) -> DbResult<Option<Hash>> {
+        Ok(self.node_tree.get(&(mmr_id.to_vec(), pos))?)
     }
 
     /// Get the position of a leaf by its hash (reverse lookup)
     pub fn get_leaf_position(&self, mmr_id: &[u8], hash: Hash) -> DbResult<Option<u64>> {
-        Ok(self.hash_index_tree.get(&(mmr_id.to_vec(), hash))?)
+        Ok(self.hash_pos_tree.get(&(mmr_id.to_vec(), hash))?)
     }
 
     fn append_leaf_in_transaction<A: MmrAlgorithm>(
@@ -58,7 +53,7 @@ impl GlobalMmrDb {
         hash: Hash,
         nt: SledTransactionalTree<GlobalMmrNodeSchema>,
         mt: SledTransactionalTree<GlobalMmrMetaSchema>,
-        hit: SledTransactionalTree<GlobalMmrHashIndexSchema>,
+        hpt: SledTransactionalTree<GlobalMmrHashPositionSchema>,
     ) -> typed_sled::error::Result<u64> {
         let metadata = mt
             .get(&mmr_id)?
@@ -73,10 +68,10 @@ impl GlobalMmrDb {
 
         for (pos, node_hash) in &result.nodes_to_write {
             nt.insert(&(mmr_id.clone(), *pos), &Buf32(*node_hash))?;
-            hit.insert(&(mmr_id.clone(), Buf32(*node_hash)), pos)?;
+            hpt.insert(&(mmr_id.clone(), Buf32(*node_hash)), pos)?;
         }
 
-        hit.insert(&(mmr_id.clone(), hash), &result.leaf_index)?;
+        hpt.insert(&(mmr_id.clone(), hash), &result.leaf_index)?;
         mt.insert(&mmr_id, &result.new_metadata)?;
 
         Ok(result.leaf_index)
@@ -90,20 +85,20 @@ impl GlobalMmrDatabase for GlobalMmrDb {
         self.ensure_mmr_metadata(&mmr_id)?;
 
         self.config.with_retry(
-            (&self.node_tree, &self.meta_tree, &self.hash_index_tree),
-            |(nt, mt, hit)| {
+            (&self.node_tree, &self.meta_tree, &self.hash_pos_tree),
+            |(nt, mt, hpt)| {
                 Ok(Self::append_leaf_in_transaction::<Self::MmrAlgorithm>(
                     mmr_id.clone(),
                     hash,
                     nt,
                     mt,
-                    hit,
+                    hpt,
                 )?)
             },
         )
     }
 
-    fn get_node(&self, mmr_id: RawMmrId, pos: u64) -> DbResult<Hash> {
+    fn get_node(&self, mmr_id: RawMmrId, pos: u64) -> DbResult<Option<Hash>> {
         self.get_mmr_node(&mmr_id, pos)
     }
 
@@ -146,8 +141,8 @@ impl GlobalMmrDatabase for GlobalMmrDb {
         self.ensure_mmr_metadata(&mmr_id)?;
 
         self.config.with_retry(
-            (&self.node_tree, &self.meta_tree, &self.hash_index_tree),
-            |(nt, mt, hit)| {
+            (&self.node_tree, &self.meta_tree, &self.hash_pos_tree),
+            |(nt, mt, hpt)| {
                 let metadata = mt
                     .get(&mmr_id)?
                     .expect("MMR metadata must exist after ensure_mmr_metadata");
@@ -170,7 +165,7 @@ impl GlobalMmrDatabase for GlobalMmrDb {
                 mt.insert(&mmr_id, &result.new_metadata)?;
 
                 // Remove hash -> position mapping for the popped leaf
-                hit.remove(&(mmr_id.clone(), Buf32(result.leaf_hash)))?;
+                hpt.remove(&(mmr_id.clone(), Buf32(result.leaf_hash)))?;
 
                 Ok(Some(result.leaf_hash.into()))
             },
@@ -189,16 +184,16 @@ impl GlobalMmrDatabase for GlobalMmrDb {
             (
                 &self.node_tree,
                 &self.meta_tree,
-                &self.hash_index_tree,
+                &self.hash_pos_tree,
                 &self.preimage_tree,
             ),
-            |(nt, mt, hit, pit)| {
+            |(nt, mt, hpt, pit)| {
                 let leaf_index = Self::append_leaf_in_transaction::<Self::MmrAlgorithm>(
                     mmr_id.clone(),
                     hash,
                     nt,
                     mt,
-                    hit,
+                    hpt,
                 )?;
                 pit.insert(&(mmr_id.clone(), leaf_index), &preimage)?;
                 Ok(leaf_index)
@@ -242,7 +237,7 @@ mod tests {
         assert_eq!(db.get_mmr_size(mmr_id.clone()).unwrap(), 1);
 
         // Can retrieve the node
-        let node = db.get_node(mmr_id, 0).unwrap();
+        let node = db.get_node(mmr_id, 0).unwrap().unwrap();
         assert_eq!(node, hash);
     }
 
@@ -282,10 +277,22 @@ mod tests {
         // Leaves:   [0, 1, x, 2, 3, x, x]
 
         // Verify leaf positions
-        assert_eq!(db.get_node(mmr_id.clone(), 0).unwrap(), [0u8; 32].into());
-        assert_eq!(db.get_node(mmr_id.clone(), 1).unwrap(), [1u8; 32].into());
-        assert_eq!(db.get_node(mmr_id.clone(), 3).unwrap(), [2u8; 32].into());
-        assert_eq!(db.get_node(mmr_id.clone(), 4).unwrap(), [3u8; 32].into());
+        assert_eq!(
+            db.get_node(mmr_id.clone(), 0).unwrap().unwrap(),
+            [0u8; 32].into()
+        );
+        assert_eq!(
+            db.get_node(mmr_id.clone(), 1).unwrap().unwrap(),
+            [1u8; 32].into()
+        );
+        assert_eq!(
+            db.get_node(mmr_id.clone(), 3).unwrap().unwrap(),
+            [2u8; 32].into()
+        );
+        assert_eq!(
+            db.get_node(mmr_id.clone(), 4).unwrap().unwrap(),
+            [3u8; 32].into()
+        );
 
         // Internal nodes exist
         assert!(db.get_node(mmr_id.clone(), 2).is_ok());
--- a/crates/db/store-sled/src/global_mmr/schemas.rs
+++ b/crates/db/store-sled/src/global_mmr/schemas.rs
@@ -17,7 +17,7 @@ define_table_with_seek_key_codec!(
 define_table_with_seek_key_codec!(
     /// Global MMR hash position: (mmr_id, hash) -> position
     /// Enables reverse lookup from node hash to node position
-    (GlobalMmrHashIndexSchema) (RawMmrId, Buf32) => u64
+    (GlobalMmrHashPositionSchema) (RawMmrId, Buf32) => u64
 );
 
 define_table_with_seek_key_codec!(
--- a/crates/db/types/src/mmr_helpers.rs
+++ b/crates/db/types/src/mmr_helpers.rs
@@ -426,8 +426,7 @@ impl MmrAlgorithm for BitManipulatedMmrAlgorithm {
                 current_hash = parent_hash;
                 current_height = next_height;
             } else {
-                // Current is a left sibling (will be merged when right sibling comes)
-                // or we've reached a peak - stop here
+                // Current is a left sibling (will be merged when right sibling comes) - stop here
                 break;
             }
         }
@@ -479,7 +478,7 @@ impl MmrAlgorithm for BitManipulatedMmrAlgorithm {
         // Read the leaf hash before deletion
         let leaf_hash = get_node(leaf_pos)?;
 
-        // Calculate the new MMR size (before the last leaf was added)
+        // Calculate the new MMR size (after the last leaf was popped)
         let new_mmr_size = if metadata.num_leaves == 1 {
             0 // Empty MMR
         } else {
--- a/crates/db/types/src/traits.rs
+++ b/crates/db/types/src/traits.rs
@@ -368,7 +368,7 @@ pub trait GlobalMmrDatabase: Send + Sync + 'static {
     fn append_leaf(&self, mmr_id: RawMmrId, hash: Hash) -> DbResult<u64>;
 
     /// Get a node hash by MMR position
-    fn get_node(&self, mmr_id: RawMmrId, pos: u64) -> DbResult<Hash>;
+    fn get_node(&self, mmr_id: RawMmrId, pos: u64) -> DbResult<Option<Hash>>;
 
     /// Get the total MMR size
     fn get_mmr_size(&self, mmr_id: RawMmrId) -> DbResult<u64>;
--- a/crates/storage/src/managers/global_mmr.rs
+++ b/crates/storage/src/managers/global_mmr.rs
@@ -2,7 +2,7 @@ use std::{marker::PhantomData, sync::Arc};
 
 use borsh::{BorshDeserialize, BorshSerialize};
 use strata_db_types::{
-    mmr_helpers::{leaf_index_to_pos, BitManipulatedMmrAlgorithm, MmrAlgorithm},
+    mmr_helpers::{BitManipulatedMmrAlgorithm, MmrAlgorithm},
     traits::GlobalMmrDatabase,
     DbError, DbResult,
 };
@@ -100,7 +100,7 @@ impl MmrHandle {
     }
 
     /// Get a node at a specific position (blocking)
-    pub fn get_node_blocking(&self, pos: u64) -> DbResult<Hash> {
+    pub fn get_node_blocking(&self, pos: u64) -> DbResult<Option<Hash>> {
         self.ops.get_node_blocking(self.mmr_id.to_bytes(), pos)
     }
 
@@ -118,18 +118,22 @@ impl MmrHandle {
     pub fn generate_proof(&self, index: u64) -> DbResult<MerkleProof> {
         let mmr_size = self.get_mmr_size_blocking()?;
 
-        BitManipulatedMmrAlgorithm::generate_proof(index, mmr_size, |pos| {
-            self.get_node_blocking(pos).map(|x| x.0)
-        })
+        BitManipulatedMmrAlgorithm::generate_proof(index, mmr_size, self.node_getter())
     }
 
     /// Generate Merkle proofs for a range of leaf positions
     pub fn generate_proofs(&self, start: u64, end: u64) -> DbResult<Vec<MerkleProof>> {
         let mmr_size = self.get_mmr_size_blocking()?;
 
-        BitManipulatedMmrAlgorithm::generate_proofs(start, end, mmr_size, |pos| {
-            self.get_node_blocking(pos).map(|x| x.0)
-        })
+        BitManipulatedMmrAlgorithm::generate_proofs(start, end, mmr_size, self.node_getter())
+    }
+
+    fn node_getter(&self) -> impl Fn(u64) -> DbResult<[u8; 32]> + '_ {
+        |pos| {
+            self.get_node_blocking(pos)?
+                .map(|x| x.0)
+                .ok_or(DbError::MmrLeafNotFound(pos))
+        }
     }
 
     /// Remove and return the last leaf from the MMR (async version)
@@ -145,14 +149,6 @@ impl MmrHandle {
     /// Get the MmrId for this handle
     pub fn mmr_id(&self) -> &MmrId {
         &self.mmr_id
-    }
-
-    /// Get a leaf hash by leaf index (blocking)
-    ///
-    /// Converts the leaf index to MMR position and retrieves the hash.
-    pub fn get_leaf_hash_blocking(&self, leaf_index: u64) -> DbResult<Hash> {
-        let pos = leaf_index_to_pos(leaf_index);
-        self.get_node_blocking(pos)
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2847,7 +2847,7 @@ checksum = "7fb42c12f538927ccce999e0227b1ac1b911065d2d1086f6b7451f1e217c8c32"
 dependencies = [
  "base64 0.22.1",
  "bitcoin",
- "corepc-types 0.10.1",
+ "corepc-types",
  "hex-conservative",
  "reqwest 0.12.26",
  "secp256k1 0.29.1",
@@ -3612,12 +3612,12 @@ dependencies = [
 
 [[package]]
 name = "corepc-client"
-version = "0.8.0"
+version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "565724b3d7556c06827b27c468333bbf97e2173097aaed576b5da73faf7a1e04"
+checksum = "7755b8b9219b23d166a5897b5e2d8266cbdd0de5861d351b96f6db26bcf415f3"
 dependencies = [
  "bitcoin",
- "corepc-types 0.8.0",
+ "corepc-types",
  "jsonrpc",
  "log",
  "serde",
@@ -3626,9 +3626,9 @@ dependencies = [
 
 [[package]]
 name = "corepc-node"
-version = "0.8.0"
+version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "596e4ee1ff3616ae3f36e89e0b847604034ba689243cc5e7dd08da395b88ec58"
+checksum = "768391062ec3812e223bb3031c5b2fcdd6e0e60b816157f21df82fd3e6617dc0"
 dependencies = [
  "anyhow",
  "corepc-client",
@@ -3636,17 +3636,6 @@ dependencies = [
  "serde_json",
  "tempfile",
  "which 3.1.1",
-]
-
-[[package]]
-name = "corepc-types"
-version = "0.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5a7f2faedffc7c654e348e2da6c6416090525c6072979fee9681d620d1d398a4"
-dependencies = [
- "bitcoin",
- "serde",
- "serde_json",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -402,7 +402,7 @@ bytes = "1.6.0"
 cfg-if = "1.0.0"
 chrono = "0.4.38"
 clap = { version = "4.5.39", default-features = false, features = ["derive"] }
-corepc-node = { version = "0.8.0", features = ["29_0"] }
+corepc-node = { version = "0.10.0", features = ["29_0"] }
 criterion = { version = "0.7.0", features = ["html_reports"] }
 deadpool = "0.12.1"
 digest = "0.10"
--- a/crates/asm/worker/src/errors.rs
+++ b/crates/asm/worker/src/errors.rs
@@ -53,6 +53,6 @@ pub enum WorkerError {
     #[error("MMR proof generation failed for index {index}")]
     MmrProofFailed { index: u64 },
 
-    #[error("Mahifest hash out of bound (max {max}, requested {index})")]
+    #[error("Manifest hash out of bound (max {max}, requested {index})")]
     ManifestIndexOutOfBound { index: u64, max: u64 },
 }
--- a/crates/storage/src/ops/global_mmr.rs
+++ b/crates/storage/src/ops/global_mmr.rs
@@ -9,7 +9,7 @@ inst_ops_simple! {
         append_leaf_with_preimage(mmr_id: RawMmrId, hash: Hash, preimage: Vec<u8>) => u64;
         get_preimage(mmr_id: RawMmrId, index: u64) => Option<Vec<u8>>;
         pop_leaf(mmr_id: RawMmrId) => Option<Hash>;
-        get_node(mmr_id: RawMmrId, pos: u64) => Hash;
+        get_node(mmr_id: RawMmrId, pos: u64) => Option<Hash>;
         get_mmr_size(mmr_id: RawMmrId) => u64;
         get_num_leaves(mmr_id: RawMmrId) => u64;
     }


crates/db/store-sled/src/lib.rs
```diff
@@ -143,9 +148,18 @@ impl DatabaseBackend for SledBackend {
 }
 
 impl SledBackend {
-    /// Get the MMR database
-    ///
-    /// Returns the ASM database which implements the MmrDatabase trait.
+    /// Get the ASM MMR database
+    pub fn asm_mmr_db(&self) -> Arc<AsmDBSled> {
+        self.asm_db.clone()
```
bewakes: Should probably separate out the asm mmr from the asm db in another ticket?

crates/consensus-logic/src/asm_worker_context.rs
```diff
@@ -20,8 +20,8 @@ pub struct AsmWorkerCtx {
     bitcoin_client: Arc<Client>,
     l1man: Arc<L1BlockManager>,
     asmman: Arc<AsmStateManager>,
-    /// MMR manager for proof generation
-    mmr_manager: Arc<MmrManager>,
+    /// MMR handle for ASM manifest MMR
+    mmr_handle: MmrHandle,
```
barakshani, line 25: handler?

crates/consensus-logic/src/asm_worker_context.rs
```diff
@@ -124,32 +124,38 @@ impl WorkerContext for AsmWorkerCtx {
...
-            .map(|opt| opt.map(|buf| buf.0))
-            .map_err(conv_db_err)
+        let num_leaves = self.mmr_handle.num_leaves_blocking().map_err(|e| {
+            tracing::error!(?e, "Failed to get MMR num_leaves");
+            WorkerError::DbError
+        })?;
+
+        if index >= num_leaves {
+            return Ok(None);
+        }
```
barakshani: shouldn't this be an error?

crates/db/store-sled/src/unified_mmr/mod.rs
barakshani, line 1: what makes this unified? is this a db all MMR instances are supposed to use?

crates/db/store-sled/src/unified_mmr/schemas.rs
```diff
@@ -0,0 +1,26 @@
+use strata_db_types::mmr_helpers::{MmrId, MmrMetadata};
+use strata_primitives::buf::Buf32;
+
+use crate::{define_table_with_seek_key_codec, define_table_without_codec, impl_borsh_value_codec};
+
+define_table_with_seek_key_codec!(
+    /// Unified MMR node storage: (mmr_id, position) -> hash
+    (UnifiedMmrNodeSchema) (MmrId, u64) => Buf32
```
barakshani: this is the reverse of `UnifiedMmrHashIndexSchema`, right? Why not calling it `UnifiedMmrIndexHashSchema` for consistent language?

Also, _node_ usually refers to a tree internal node, while here I think we mean (only) leaves.

crates/db/types/src/mmr_helpers.rs
```diff
@@ -224,10 +243,330 @@ pub fn find_peak_for_pos(pos: u64, mmr_size: u64) -> DbResult<u64> {
...
+    {
+        let leaf_index = metadata.num_leaves;
+        let leaf_pos = leaf_index_to_pos(leaf_index);
+
+        let mut nodes_to_write = Vec::new();
+
+        // Store the leaf
+        nodes_to_write.push((leaf_pos, hash));
+
+        // Merge along the path to create internal nodes
```
barakshani: I don't understand the merge terminology. What are we merging?

crates/db/types/src/mmr_helpers.rs
```diff
@@ -224,10 +243,330 @@ pub fn find_peak_for_pos(pos: u64, mmr_size: u64) -> DbResult<u64> {
...
+                // Store parent
+                nodes_to_write.push((next_pos, parent_hash));
+
+                // Move up to parent
+                current_pos = next_pos;
+                current_hash = parent_hash;
+                current_height = next_height;
+            } else {
+                // Current is a left sibling (will be merged when right sibling comes)
+                // or we've reached a peak - stop here
```
barakshani: even if we reached a peak, if a node isn't a right sibling then it is a potential left sibling (that will be merged when, and if, a right sibling comes), so you can safely remove the second case.

crates/db/types/src/mmr_helpers.rs
```diff
@@ -224,10 +243,330 @@ pub fn find_peak_for_pos(pos: u64, mmr_size: u64) -> DbResult<u64> {
...
+
+        // Keep merging as long as we have a left sibling
+        loop {
+            let next_pos = current_pos + 1;
+            let next_height = pos_height_in_tree(next_pos);
+
+            // If next position is higher, current is a right sibling - we should merge
+            if next_height > current_height {
+                // Current is right sibling, get left sibling
+                let sibling_position = sibling_pos(current_pos, current_height);
```
barakshani, line 415: maybe not worth the optimisation, but since you already know it's a right sibling you can directly get the correct sibling position by `current_pos - sibling_offset(current_height)`.

crates/db/types/src/mmr_helpers.rs
```diff
@@ -224,10 +243,330 @@ pub fn find_peak_for_pos(pos: u64, mmr_size: u64) -> DbResult<u64> {
...
+                    get_node(pos).map(Buf32)
+                }
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let new_metadata = MmrMetadata {
+            num_leaves: new_num_leaves,
+            mmr_size: new_mmr_size,
+            peak_roots: new_peak_roots,
+        };
```
barakshani, line 456: Would this be cleaner?
```suggestion
        let new_metadata = MmrMetadata {
            num_leaves: metadata.num_leaves + 1,
            mmr_size: leaf_index_to_mmr_size(metadata.num_leaves),
            peak_roots: new_peak_roots,
        };
```

crates/db/types/src/mmr_helpers.rs
```diff
@@ -224,10 +243,330 @@ pub fn find_peak_for_pos(pos: u64, mmr_size: u64) -> DbResult<u64> {
...
+                // Current is a left sibling (will be merged when right sibling comes)
+                // or we've reached a peak - stop here
+                break;
+            }
+        }
+
+        // Update metadata
+        let new_num_leaves = metadata.num_leaves + 1;
+        let peak_count = new_num_leaves.count_ones() as u64;
+        let new_mmr_size = 2 * new_num_leaves - peak_count;
```
barakshani: it's natural that code readers won't understand where this is coming from. Either refer to `leaf_index_to_mmr_size` in this file above, or use the suggested alternative below and just remove this.

crates/db/store-sled/src/unified_mmr/schemas.rs
```diff
@@ -0,0 +1,26 @@
...
+);
+
+define_table_with_seek_key_codec!(
+    /// Unified MMR metadata schema: mmr_id -> metadata
+    (UnifiedMmrMetaSchema) MmrId => MmrMetadata
+);
+
+define_table_with_seek_key_codec!(
+    /// Unified MMR hash index: (mmr_id, hash) -> position
+    /// Enables reverse lookup from hash to leaf position
```
barakshani: What do you mean by hash index?
This suppose to be the reverse to `UnifiedMmrNodeSchema`, right? but there we store hashes for all positions (nodes), not just leaves. See also my comment above in `append_leaf()`.

crates/db/store-sled/src/unified_mmr/db.rs
```diff
@@ -0,0 +1,515 @@
...
+                .map_err(typed_sled::error::Error::abort)?;
+
+                for (pos, node_hash) in &result.nodes_to_write {
+                    nt.insert(&(mmr_id.clone(), *pos), &Buf32(*node_hash))?;
+                }
+
+                mt.insert(&mmr_id, &result.new_metadata)?;
+
+                // Store hash -> position mapping for the leaf
+                hit.insert(&(mmr_id.clone(), Buf32(hash)), &result.leaf_index)?;
```
barakshani: shouldn't we map hashes for all nodes, not just the (added) leaf? see also my comment below, on the schema.

crates/db/store-sled/src/unified_mmr/db.rs
```diff
@@ -0,0 +1,515 @@
...
+            .collect();
+
+        CompactMmr64 {
+            entries: metadata.num_leaves,
+            cap_log2: 64,
+            roots: roots_vec.into(),
+        }
+    }
+
+    fn pop_leaf(&self, mmr_id: MmrId) -> DbResult<Option<[u8; 32]>> {
```
barakshani: What are we going to use this for??

crates/db/types/src/mmr_helpers.rs
```diff
@@ -26,7 +28,24 @@
...
+#[derive(Debug, Clone, Error)]
+pub enum MmrError {
+    #[error("MMR leaf {0} not found")]
+    LeafNotFound(u64),
+
+    #[error("Invalid MMR range: start={start}, end={end}")]
+    InvalidRange { start: u64, end: u64 },
+
+    #[error("Position {pos} out of bounds for MMR size {mmr_size}")]
+    PositionOutOfBounds { pos: u64, mmr_size: u64 },
```
delbonis: `invalid mmr range (start {start}, end {end})`

`mmr index {pos} out of bounds (max {max})`

crates/db/types/src/mmr_helpers.rs
```diff
@@ -224,10 +243,330 @@ pub fn find_peak_for_pos(pos: u64, mmr_size: u64) -> DbResult<u64> {
...
+///
+/// Each variant represents a different MMR type, with optional scoping
+/// within that type (e.g., per-account MMRs).
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum MmrId {
+    /// ASM manifest MMR (singleton, no account scope)
+    Asm,
+    /// Snark message MMR (per-account scope)
+    SnarkMsg(strata_identifiers::AccountId),
+}
```
delbonis: [The ticket](https://alpenlabs.atlassian.net/browse/STR-1202) specifies that IDs must be opaque `Buf32`s.  It's the job of consumers to determine what the opaque identifiers mean.

crates/db/types/src/mmr_helpers.rs
```diff
@@ -224,10 +243,330 @@ pub fn find_peak_for_pos(pos: u64, mmr_size: u64) -> DbResult<u64> {
...
+
+/// Metadata for an MMR instance
+#[derive(Debug, Clone, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct MmrMetadata {
+    pub num_leaves: u64,
+    pub mmr_size: u64,
+    pub peak_roots: Vec<Buf32>,
```
delbonis: What is `mmr_size` here?

The `peak_roots` isn't necessary with the design improvements from @evgenyzdanovich , afaik.

crates/db/types/src/mmr_helpers.rs
```diff
@@ -224,10 +243,330 @@ pub fn find_peak_for_pos(pos: u64, mmr_size: u64) -> DbResult<u64> {
...
+    pub new_metadata: MmrMetadata,
+}
+
+/// Pure MMR algorithm logic separated from storage
+///
+/// This struct provides the core MMR algorithms (append, pop) as pure functions
+/// that compute what changes need to be made without actually making them.
+/// Storage implementations apply these changes within their own transaction context.
+#[derive(Debug)]
+pub struct MmrAlgorithm;
```
delbonis: Why is this a struct?  This shouldn't be, this is a Java-like pattern.  When I skimmed this initially I thought it was a trait so that consumers could plug in their own hashing functions, but this doesn't make sense.

Try to rely on the functionality from the upstream crates, if possible.

crates/storage/Cargo.toml
```diff
@@ -23,6 +23,7 @@ strata-storage-common.workspace = true
 
 anyhow.workspace = true
 async-trait.workspace = true
+borsh.workspace = true
```
delbonis, line 27: why?

crates/db/types/src/traits.rs
```diff
@@ -359,11 +350,67 @@ pub trait L1BroadcastDatabase: Send + Sync + 'static {
...
+    fn mmr_size(&self, mmr_id: MmrId) -> DbResult<u64>;
+
+    /// Get the total number of leaves
+    fn num_leaves(&self, mmr_id: MmrId) -> DbResult<u64>;
+
+    /// Get the individual peak roots
+    fn peak_roots(&self, mmr_id: MmrId) -> Vec<[u8; 32]>;
+
+    /// Get a compact representation of the MMR
+    fn to_compact(&self, mmr_id: MmrId) -> CompactMmr64B32;
```
delbonis: `to_` implies that this is a conversion, but it's really doing database fetches.

crates/db/types/src/traits.rs
```diff
@@ -359,11 +350,67 @@ pub trait L1BroadcastDatabase: Send + Sync + 'static {
...
+    fn get_node(&self, mmr_id: MmrId, pos: u64) -> DbResult<[u8; 32]>;
+
+    /// Get the total MMR size
+    fn mmr_size(&self, mmr_id: MmrId) -> DbResult<u64>;
+
+    /// Get the total number of leaves
+    fn num_leaves(&self, mmr_id: MmrId) -> DbResult<u64>;
+
+    /// Get the individual peak roots
+    fn peak_roots(&self, mmr_id: MmrId) -> Vec<[u8; 32]>;
```
delbonis: These are not verby names which implies that they're all accessors.

Shouldn't `peak_roots` return a result?

crates/db/types/src/traits.rs
```diff
@@ -359,11 +350,67 @@ pub trait L1BroadcastDatabase: Send + Sync + 'static {
...
+        Err(DbError::Unimplemented)
+    }
+
+    /// Get pre-image data by leaf index (optional operation)
+    ///
+    /// Returns None if no pre-image exists at the given index.
+    /// Default implementation returns Unimplemented error.
+    fn get_preimage(&self, _mmr_id: MmrId, _index: u64) -> DbResult<Option<Vec<u8>>> {
+        Err(DbError::Unimplemented)
+    }
```
delbonis, line 407: Do not have default implementations on database traits, we should always implement all of these.

crates/db/types/src/traits.rs
```diff
@@ -359,11 +350,67 @@ pub trait L1BroadcastDatabase: Send + Sync + 'static {
...
+// =============================================================================
+// MMR Database Traits
+// =============================================================================
+
+/// Unified MMR database trait using MmrId for identification
+///
+/// This trait provides a simpler, more direct API where the MMR instance
+/// is identified by an `MmrId` enum rather than a generic scope parameter.
+/// This is the preferred interface for new code.
+pub trait UnifiedMmrDatabase: Send + Sync + 'static {
```
delbonis: How does this relate to the other `MmrDatabase` trait?

crates/db/types/src/traits.rs
```diff
@@ -359,11 +350,67 @@ pub trait L1BroadcastDatabase: Send + Sync + 'static {
...
+
+    /// Append a leaf with its pre-image data (optional operation)
+    ///
+    /// Atomically stores both the MMR hash and the original data.
+    /// Default implementation returns Unimplemented error.
+    fn append_leaf_with_preimage(
+        &self,
+        _mmr_id: MmrId,
+        _hash: [u8; 32],
+        _preimage: Vec<u8>,
```
delbonis: Should only need one of these if the underlying database assumes a MMR node hash function.

crates/db/types/src/mmr_helpers.rs
```diff
@@ -224,10 +243,330 @@ pub fn find_peak_for_pos(pos: u64, mmr_size: u64) -> DbResult<u64> {
...
+    /// Compute the result of popping the last leaf from the MMR
+    ///
+    /// # Arguments
+    ///
+    /// * `metadata` - Current MMR metadata
+    /// * `get_node` - Closure to read existing nodes (for leaf hash and new peaks)
+    ///
+    /// # Returns
+    ///
+    /// `Some(PopLeafResult)` if there was a leaf to pop, containing:
```
barakshani: reading this sentence is a bit ambiguous because it's not strictly clear if "containing" refers to the result or to the leaf.
I advise to change the order s.t. ` None if the MMR is empty` is first, then `Some(PopLeafResult)` and just delete "if there was a leaf to pop,".

crates/db/types/src/mmr_helpers.rs
```diff
@@ -224,10 +243,330 @@ pub fn find_peak_for_pos(pos: u64, mmr_size: u64) -> DbResult<u64> {
...
+            return Ok(None);
+        }
+
+        let leaf_index = metadata.num_leaves - 1;
+        let leaf_pos = leaf_index_to_pos(leaf_index);
+
+        // Read the leaf hash before deletion
+        let leaf_hash = get_node(leaf_pos)?;
+
+        // Calculate the old MMR size (before the last leaf was added)
```
barakshani: I think that _new_ is better than _old_.

crates/db/types/src/mmr_helpers.rs
```diff
@@ -224,10 +243,330 @@ pub fn find_peak_for_pos(pos: u64, mmr_size: u64) -> DbResult<u64> {
...
+        } else {
+            leaf_index_to_mmr_size(metadata.num_leaves - 2)
+        };
+
+        // Collect nodes to remove: [old_mmr_size, current_mmr_size)
+        let nodes_to_remove: Vec<u64> = (old_mmr_size..metadata.mmr_size).collect();
+
+        // Update metadata
+        let new_num_leaves = metadata.num_leaves - 1;
+        let new_mmr_size = old_mmr_size;
```
barakshani: ...and here you call it _new_, so why not start like this and save some LoC?

crates/db/store-sled/src/global_mmr/db.rs
```diff
@@ -0,0 +1,515 @@
...
+                };
+
+                for pos in &result.nodes_to_remove {
+                    nt.remove(&(mmr_id.clone(), *pos))?;
+                }
+
+                mt.insert(&mmr_id, &result.new_metadata)?;
+
+                // Remove hash -> position mapping for the popped leaf
+                hit.remove(&(mmr_id.clone(), Buf32(result.leaf_hash)))?;
```
barakshani: same as https://github.com/alpenlabs/alpen/pull/1226/files#r2647280836, in case we end up changing something.

crates/db/store-sled/src/unified_mmr/db.rs
```diff
@@ -0,0 +1,515 @@
...
+                Ok(Some(result.leaf_hash))
+            },
+        )
+    }
+
+    fn append_leaf_with_preimage(
+        &self,
+        mmr_id: MmrId,
+        hash: [u8; 32],
+        preimage: Vec<u8>,
```
barakshani: why do we need to provide the hash, and if we do so, we should verify their consistency.

crates/db/store-sled/src/global_mmr/db.rs
```diff
@@ -0,0 +1,515 @@
...
+        preimage: Vec<u8>,
+    ) -> DbResult<u64> {
+        self.ensure_mmr_metadata(&mmr_id)?;
+
+        self.config.with_retry(
+            (
+                &self.node_tree,
+                &self.meta_tree,
+                &self.hash_index_tree,
+                &self.preimage_tree,
```
barakshani, line 188: given we have `append_leaf`, that doesn't take a preimage, this means that `preimage_tree` may not include items for which we actually we do a have leaf.
1. I guess it's fine, but want to double check. Perhaps in implementation we only call this function(?)
2. shouldn't we at least update it to indicate that a leaf exists but we don't know the preimage? say with _None_, if it makes sense, or some distinguished value?

crates/db/store-sled/src/unified_mmr/db.rs
```diff
@@ -0,0 +1,515 @@
...
+
+        self.config.with_retry(
+            (
+                &self.node_tree,
+                &self.meta_tree,
+                &self.hash_index_tree,
+                &self.preimage_tree,
+            ),
+            |(nt, mt, hit, pit)| {
+                let metadata = mt
```
barakshani: can't we just call `self.append_leaf()` and spare these Loc? (any change to that function will need to be applied here as well)

crates/db/store-sled/src/unified_mmr/db.rs
```diff
@@ -0,0 +1,515 @@
...
+        let account1 = AccountId::zero();
+        let account2 = AccountId::from([1u8; 32]);
+
+        // Append to ASM MMR
+        let asm_hash = [10u8; 32];
+        db.append_leaf(MmrId::Asm, asm_hash).unwrap();
+
+        // Append to account1 MMR
+        let acc1_hash = [20u8; 32];
+        db.append_leaf(MmrId::SnarkMsg(account1), acc1_hash)
```
barakshani: SnarkMsg?? do you mean _inbox_?

crates/db/types/src/mmr_helpers.rs
```diff
@@ -224,10 +243,330 @@ pub fn find_peak_for_pos(pos: u64, mmr_size: u64) -> DbResult<u64> {
...
+
+        if end >= num_leaves {
+            return Err(MmrError::LeafNotFound(end).into());
+        }
+
+        let mut proofs = Vec::with_capacity((end - start + 1) as usize);
+        for index in start..=end {
+            let proof = Self::generate_proof(index, mmr_size, num_leaves, &get_node)?;
+            proofs.push(proof);
+        }
```
barakshani, line 558: I would have assumed we'll have something more sophisticated than individual proofs. Since the provided leaves are sequential, at some (rather early) point the proofs will contain the same suffix (for example, if we have 3 leaves in the range, we are guaranteed to have the same final hash (excluding the peak) appearing in two different proofs -- this is true even if leaves are not neighbours).

It's a side quest I encourage you to take at some point.

crates/db/types/src/traits.rs
```diff
@@ -359,11 +350,67 @@ pub trait L1BroadcastDatabase: Send + Sync + 'static {
...
+/// Unified MMR database trait using MmrId for identification
+///
+/// This trait provides a simpler, more direct API where the MMR instance
+/// is identified by an `MmrId` enum rather than a generic scope parameter.
+/// This is the preferred interface for new code.
+pub trait UnifiedMmrDatabase: Send + Sync + 'static {
+    /// Append a new leaf to the specified MMR
+    fn append_leaf(&self, mmr_id: MmrId, hash: [u8; 32]) -> DbResult<u64>;
+
+    /// Get a node hash by MMR position
```
barakshani, line 370: nit: I think that "node label" is more common; one may confuse to think that we hash the (already hashed) node label.

crates/db/types/src/traits.rs
```diff
@@ -359,11 +350,67 @@ pub trait L1BroadcastDatabase: Send + Sync + 'static {
...
+    fn get_node(&self, mmr_id: MmrId, pos: u64) -> DbResult<[u8; 32]>;
+
+    /// Get the total MMR size
+    fn mmr_size(&self, mmr_id: MmrId) -> DbResult<u64>;
+
+    /// Get the total number of leaves
+    fn num_leaves(&self, mmr_id: MmrId) -> DbResult<u64>;
```
barakshani: why not `get_`?

crates/db/types/src/traits.rs
barakshani: _peak_(s) and _roots_ are the same, and not sure what "individual" adds here.

crates/storage/src/managers/global_mmr.rs
```diff
@@ -0,0 +1,250 @@
...
+    pub fn new(pool: ThreadPool, db: Arc<impl UnifiedMmrDatabase + 'static>) -> Self {
+        let ops = Arc::new(ops::unified_mmr::Context::new(db).into_ops(pool));
+        Self { ops }
+    }
+
+    /// Get a handle for a specific MMR instance (hash-only operations)
+    ///
+    /// The handle captures the MmrId and provides methods that don't require
+    /// repeatedly passing the mmr_id parameter.
+    pub fn get_handle(&self, mmr_id: MmrId) -> MmrHandle {
```
barakshani, line 39: handler? as in https://github.com/alpenlabs/alpen/pull/1226/files#r2647185663

barakshani (commented):
generally looks good

crates/db/types/src/mmr_helpers.rs
barakshani: Actually, in this case where leaves are neighbours, if there are 3 leaves in the range then since 2 must be siblings, we end up with exactly the same proof apart from the first item.

crates/db/store-sled/src/unified_mmr/db.rs
evgenyzdanovich: for example can be useful in btc "reorg" cases. let's say asm accepted a block that's no longer belongs to the longest chain - then, instead of abandoning (or reconstructing) the corresponding mmr entries, we can just drop the last leaf and continue with appending. 

Right now it's a nice-to-have (that was initially broguth by me, according to the feedback in #1193 )

crates/db/store-sled/src/unified_mmr/db.rs
```diff
@@ -0,0 +1,515 @@
...
+        Ok(metadata.mmr_size)
+    }
+
+    fn num_leaves(&self, mmr_id: MmrId) -> DbResult<u64> {
+        self.ensure_mmr_metadata(&mmr_id)?;
+        let metadata = self.load_mmr_metadata(&mmr_id)?;
+        Ok(metadata.num_leaves)
+    }
+
+    fn peak_roots(&self, mmr_id: MmrId) -> Vec<[u8; 32]> {
```
evgenyzdanovich: could we also please prettify bare arrays into `Buf32` for example? both in the return args across this file (here and in get_node) and in params (for example, `append_leaf`). I guess it's a leftover from my PR :)

storopoli: Actually this should be `Hash` which is an alias to `Buf32` implemented in #1207.

crates/db/store-sled/src/unified_mmr/db.rs
barakshani: I have assumed we have MMR snapshots up to some depth, so we can just "jump" to the MMR state at the right block.
Given that we have ASM snapshots, and that ASM state includes the ASM logs MMR, doesn't it imply that we store MMR snapshots, at least for the ASM one? Or do we only store the MMR peaks, but not all data?

crates/db/store-sled/src/unified_mmr/mod.rs
evgenyzdanovich: I also don't like the name.
 
maybe a better name could be like `MmrCollection` and its correspondence in other places (like `MmrCollectionManager` / or just `MmrManager` and just `MmrHandle` for an access to a speficic, keyed by `MmrId`, instance)

crates/db/types/src/mmr_helpers.rs
```diff
@@ -1,6 +1,8 @@
-//! MMR position calculation helpers
+//! MMR position calculation helpers and algorithm implementation
```
evgenyzdanovich, line 1: happy to see it going in this direction, thank you sir!

given it doesnt contain any storage related routines - this allows a very easy in-memory impl (for example, for the asm tests) **while still reusing** the exact same algo AND bitwise tricks behind that.

crates/db/types/src/mmr_helpers.rs
evgenyzdanovich: Agree here with Trey, it's better to abstract it here completely as a trait or (likely easier & better) make it opaque bytes.

crates/consensus-logic/src/asm_worker_context.rs
```diff
@@ -115,37 +115,47 @@ impl WorkerContext for AsmWorkerCtx {
...
         })
     }
 
     fn get_manifest_hash(&self, index: u64) -> WorkerResult<Option<[u8; 32]>> {
-        self.asmman
-            .get_manifest_hash(index)
-            .map(|opt| opt.map(|buf| buf.0))
-            .map_err(conv_db_err)
+        let num_leaves = self.mmr_handle.num_leaves_blocking().map_err(|e| {
+            tracing::error!(?e, "Failed to get MMR num_leaves");
```
storopoli: import `error!`

crates/consensus-logic/src/asm_worker_context.rs
storopoli: No, this is a `Handle` as in "a part of an object designed for holding, moving, or carrying the object easily"

All the API things that you do with the MMR instance you do it from the "handle"

as the docstrings for the `MmrHandle` portrays:

```
/// Handle for a specific MMR instance
///
/// Provides ergonomic async/blocking APIs for a single MMR instance.
/// The MmrId is captured at creation time, so methods don't need to
/// repeatedly pass it.
```

crates/db/types/src/errors.rs
```diff
@@ -144,3 +153,17 @@ impl From<OpsError> for DbError {
         }
     }
 }
+
+impl From<crate::mmr_helpers::MmrError> for DbError {
+    fn from(value: crate::mmr_helpers::MmrError) -> Self {
+        use crate::mmr_helpers::MmrError;
```
storopoli: import these `crate::Foo` stuff

crates/db/types/src/mmr_helpers.rs
storopoli: Please use `Hash` which is an alias for `Buf32` that was implemented in #1207. It is the same thing code wise but semantically it is much better.

crates/db/types/src/mmr_helpers.rs
storopoli: It you keep `peak_roots` please use `Hash` (#1207):

```suggestion
    pub mmr_size: u64,
    pub peak_roots: Vec<Hash>,
```

crates/db/types/src/mmr_helpers.rs
```diff
@@ -224,10 +243,330 @@ pub fn find_peak_for_pos(pos: u64, mmr_size: u64) -> DbResult<u64> {
...
+            peak_roots: Vec::new(),
+        }
+    }
+}
+
+/// Result of appending a leaf to the MMR
+#[derive(Debug)]
+pub struct AppendLeafResult {
+    pub leaf_index: u64,
+    pub nodes_to_write: Vec<(u64, [u8; 32])>,
```
storopoli, line 264: Can we use

```suggestion
    pub nodes_to_write: Vec<(u64, Hash)>,
```

crates/db/types/src/mmr_helpers.rs
```diff
@@ -224,10 +243,330 @@ pub fn find_peak_for_pos(pos: u64, mmr_size: u64) -> DbResult<u64> {
...
+    pub new_metadata: MmrMetadata,
+}
+
+/// Result of popping a leaf from the MMR
+#[derive(Debug)]
+pub struct PopLeafResult {
+    pub leaf_hash: [u8; 32],
```
storopoli, line 271: Can we use

```suggestion
    pub leaf_hash: Hash,
```

crates/db/types/src/traits.rs
```diff
@@ -359,11 +350,67 @@ pub trait L1BroadcastDatabase: Send + Sync + 'static {
...
+// =============================================================================
+
+/// Unified MMR database trait using MmrId for identification
+///
+/// This trait provides a simpler, more direct API where the MMR instance
+/// is identified by an `MmrId` enum rather than a generic scope parameter.
+/// This is the preferred interface for new code.
+pub trait UnifiedMmrDatabase: Send + Sync + 'static {
+    /// Append a new leaf to the specified MMR
+    fn append_leaf(&self, mmr_id: MmrId, hash: [u8; 32]) -> DbResult<u64>;
```
storopoli:
```suggestion
    fn append_leaf(&self, mmr_id: MmrId, hash: Hash) -> DbResult<u64>;
```

crates/db/store-sled/src/unified_mmr/mod.rs
bewakes: > what makes this unified? is this a db all MMR instances are supposed to use?

This is called unified because yes all mmr instances are going to use the same underlying database differentiated by some mmr id.

>I also don't like the name.

Hmm. What do you think of `GlobalMmr`?

crates/db/types/src/mmr_helpers.rs
bewakes: I think just having an opaque `Buf32` will not suffice here. We need some more bytes to differentiate the id. For example if say we need to have account specific mmrs for some other things than snark messages then `Buf32` won't be sufficient.
I agree it's better to have something more opaque on the database level. Perhaps just some `Vec<u8>` ?

crates/db/types/src/traits.rs
bewakes: Yes, updated.

crates/db/types/src/traits.rs
bewakes: This was done because the `*_preimage` methods were supposed to be optional. Don't know if we should enforce implementation of optional methods.

crates/db/types/src/traits.rs
bewakes: Actually this has similar interface to the `MmrDatabase` which is not required now. Removed.

crates/db/store-sled/src/unified_mmr/db.rs
bewakes: Even though we might have snapshots, we would either need to have `pop_leaf` or some sort of `init_mmr_db_from_snapshot` mechanism because database is the ultimate source of truth. It's still more mature to have this `pop_leaf` method.

crates/db/types/src/mmr_helpers.rs
bewakes: done.

crates/db/store-sled/src/unified_mmr/db.rs
bewakes: Yes, it's best to make this a total reverse of the other mapping. Updated accordingly.

crates/db/store-sled/src/unified_mmr/schemas.rs
bewakes: Updated.

crates/db/types/src/traits.rs
bewakes: updated.

crates/db/types/src/mmr_helpers.rs
```diff
@@ -372,9 +372,7 @@ impl MmrAlgorithm {
         }
 
         // Update metadata
-        let new_num_leaves = metadata.num_leaves + 1;
-        let peak_count = new_num_leaves.count_ones() as u64;
-        let new_mmr_size = 2 * new_num_leaves - peak_count;
+        let new_mmr_size = leaf_index_to_mmr_size(metadata.num_leaves);
```
barakshani: `num_leaves + 1`, no?

crates/db/types/src/mmr_helpers.rs
bewakes: Converted to trait and its implementation. Right now both are in the same place. when we actually move the algorithm to `strata-merkle` crate, we can clean that up.

crates/db/types/src/mmr_helpers.rs
bewakes: I am not sure if we should use Hash here because this will be a library module and we don't want the library to be dependent on `strata::identifiers`. The  rest of the interfaces i.e. db traits, impls, and whatever strata specific can be `Hash`.

crates/db/store-sled/src/unified_mmr/db.rs
bewakes: Yep, good point. We can't directly reuse `self.append_leaf` because of atomicity required between adding leaf and adding preimage. Created a helper that eliminates the duplication.
Thanks!

crates/db/types/src/mmr_helpers.rs
bewakes: the `leaf_index_to_mmr_size` does `+1` internally.

crates/db/types/src/mmr_helpers.rs
bewakes: Yes barak, this can be improved which is better left for the subsequent PRs.

crates/asm/worker/src/service.rs
```diff
@@ -133,12 +132,7 @@ impl<W: WorkerContext + Send + Sync + 'static> SyncService for AsmWorkerService<
                     state.context.store_l1_manifest(manifest)?;
 
                     // Append manifest hash to MMR database
-                    let leaf_index = state.context.append_manifest_to_mmr(manifest_hash)?;
-
-                    // Store manifest hash for fast lookup by aux resolver
-                    state
-                        .context
-                        .store_manifest_hash(leaf_index, manifest_hash)?;
+                    let leaf_index = state.context.append_manifest_to_mmr(manifest_hash.into())?;
```
prajwolrg, line 135: I feel like this should be atomic at this level. 

`store_l1_manifest` should also append it to MMR. 

CC: @evgenyzdanovich

crates/consensus-logic/src/asm_worker_context.rs
```diff
@@ -116,36 +117,46 @@ impl WorkerContext for AsmWorkerCtx {
...
+
+        self.mmr_handle
+            .get_leaf_hash_blocking(index)
+            .map(Some)
+            .map_err(|e| {
+                error!(?e, index, "Failed to get leaf hash from MMR");
+                WorkerError::DbError
+            })
     }
 }
```
prajwolrg, line 146: A few thoughts here.

1. Since we check the error condition everytime, should we directly return `WorkerResult<Hash>`? instead of `WorkerResult<Option<Hash>>`?
2. Do we need to log the errors here? Otherwise we could just propagate the errors to the caller.

CC: @evgenyzdanovich

crates/db/types/src/traits.rs
bewakes: The underlying database does not assume any hash function, it just accepts id as raw bytes and nodes as Hash.

crates/db/store-sled/src/unified_mmr/db.rs
bewakes: The database layer does not assume any hash function. This needs to be and is enforced in the manager layer.

crates/storage/Cargo.toml
bewakes: The mmr handle expects the type(leaf entry) to be borsh serializable. This is explicitly required for mmr manager because all other managers are based on concrete types which have already implemented Borsh de/serialization but for mmr entries we store them as raw bytes in the database and need to be serialized in the manager layer.

crates/db/store-sled/src/unified_mmr/schemas.rs
```diff
@@ -0,0 +1,26 @@
+use strata_db_types::mmr_helpers::MmrMetadata;
+use strata_primitives::buf::Buf32;
+
+use crate::{define_table_with_seek_key_codec, define_table_without_codec, impl_borsh_value_codec};
+
+define_table_with_seek_key_codec!(
+    /// Unified MMR node storage: (mmr_id, position) -> hash
+    (UnifiedMmrNodeSchema) (Vec<u8>, u64) => Buf32
```
prajwolrg: Should probably create a new type for `mmr_id` that wraps `Vec<u8>`. Since we already have `MmrId` type defined, something like `RawMmrId`?

crates/db/store-sled/src/unified_mmr/mod.rs
bewakes: Used `GlobalMmr`.

crates/db/types/src/mmr_helpers.rs
prajwolrg: Agree with @bewakes. If we want to go with opaque Buf32, we might have to do some hashing to ensure that the key size is 32 bytes.

crates/db/types/src/mmr_helpers.rs
```diff
@@ -214,20 +233,554 @@ pub fn get_peaks(mmr_size: u64) -> Vec<u64> {
...
+    ///
+    /// Uses bincode with big-endian encoding for compatibility with existing data.
+    pub fn to_bytes(&self) -> Vec<u8> {
+        use bincode::Options;
+        let options = bincode::options()
+            .with_fixint_encoding()
+            .with_big_endian();
+        options.serialize(self).expect("MmrId serialization should not fail")
+    }
+}
```
prajwolrg: I don't think we need to worry about compatibility yet. Are we importing `bincode` just for this?

crates/db/types/src/mmr_helpers.rs
```diff
@@ -214,20 +233,554 @@ pub fn get_peaks(mmr_size: u64) -> Vec<u64> {
...
+    /// The caller provides a closure to read nodes and applies the result
+    /// within their own transaction mechanism.
+    fn append_leaf<F, E>(
+        hash: [u8; 32],
+        metadata: &MmrMetadata,
+        get_node: F,
+    ) -> Result<AppendLeafResult, E>
+    where
+        F: Fn(u64) -> Result<[u8; 32], E>,
+        E: From<MmrError>,
```
prajwolrg: If we are using `Hash` type as suggested in a few places, we might also make this `Hash`.

crates/db/types/src/mmr_helpers.rs
```diff
@@ -214,20 +233,554 @@ pub fn get_peaks(mmr_size: u64) -> Vec<u64> {
...
+        num_leaves: u64,
+        get_node: F,
+    ) -> Result<MerkleProof, E>
+    where
+        F: Fn(u64) -> Result<[u8; 32], E>,
+        E: From<MmrError>,
+    {
+        if leaf_index >= num_leaves {
+            return Err(MmrError::LeafNotFound(leaf_index).into());
+        }
```
prajwolrg: Do we need this check? We are passing `num_leaves` as argument just for this. Since `get_node` returns an error, can't we just use that error?

prajwolrg (approved):
Brilliant work! 

I just have a few nits and questions.

crates/db/types/src/mmr_helpers.rs
bewakes: Actually not. This will be caught from multiple other places like `get_node`, and `find_peak_for_pos`. Good catch!

crates/db/types/src/mmr_helpers.rs
bewakes: yes, actually we can just get away with borsh here which is already being used.

crates/db/store-sled/src/unified_mmr/schemas.rs
bewakes: Good idea! Done.

crates/db/store-sled/src/global_mmr/db.rs
bewakes: I don't understand fully what you mean here. How this would work is that where we use the mmr db, we would use either `append_leaf` or `append_leaf_with_preimage` based on the usecase.

barakshani: We have 2 mappings, `GlobalMmrNodeSchema` and `GlobalMmrPreimageSchema`. For a _leaf_ position, if we don't have the preimage, only `GlobalMmrNodeSchema` will be updated -- I am proposing that we also update `GlobalMmrPreimageSchema` with some "default" value that indicates that this leaf has been inserted/ is occupied, though we don't know what the pre-image is.
I am not necessarily saying this is better, that is _that we need this_, but at least for consistency reasons it seems reasonable (consistency of the two mappings).

bewakes: Hmm. I don't think that's necessary because `get_preimage` returns `None` if the preimage does not exist. If we try to explicitly insert something(null value like you suggested) then we would just use up the disk without much benefit.

crates/db/store-sled/src/global_mmr/schemas.rs
```diff
@@ -0,0 +1,27 @@
...
+
+define_table_with_seek_key_codec!(
+    /// Global MMR metadata schema: mmr_id -> metadata
+    (GlobalMmrMetaSchema) RawMmrId => MmrMetadata
+);
+
+define_table_with_seek_key_codec!(
+    /// Global MMR hash position: (mmr_id, hash) -> position
+    /// Enables reverse lookup from node hash to node position
+    (GlobalMmrHashIndexSchema) (RawMmrId, Buf32) => u64
```
barakshani: `GlobalMmrHashPositionSchema`?

storopoli (approved):
ACK af03995b21d737c1bb3f8d3108eda6170378b091

crates/db/types/src/mmr_helpers.rs
```diff
@@ -214,20 +223,343 @@ pub fn get_peaks(mmr_size: u64) -> Vec<u64> {
...
+            return Ok(None);
+        }
+
+        let leaf_index = metadata.num_leaves - 1;
+        let leaf_pos = leaf_index_to_pos(leaf_index);
+
+        // Read the leaf hash before deletion
+        let leaf_hash = get_node(leaf_pos)?;
+
+        // Calculate the new MMR size (before the last leaf was added)
```
barakshani:
```suggestion
        // Calculate the new MMR size (after the last leaf was popped)

crates/db/types/src/mmr_helpers.rs
barakshani: https://alpenlabs.atlassian.net/browse/STR-2037

barakshani (approved):
Overall state looks good; there are a few early comments that I (or others) left that I see are unaddressed (either no comments or not marked as resolved) -- can you go over them and address them (no need for a code change if you disagree with the suggestion, but either write something or close to the conversation, so I know you have seen all comments)?

crates/db/types/src/mmr_helpers.rs
bewakes: I suggest we do this in the next optimization ticket for MMR. In this PR, I am not trying to modify the core algorithm itself as it was in the codebase.

crates/db/types/src/mmr_helpers.rs
bewakes: Same as above. Thank you for the suggestion.

crates/db/types/src/mmr_helpers.rs
bewakes: Great, thanks!

crates/db/types/src/traits.rs
bewakes: The return type is a `Hash` so either way this might cause some confusion. leaving as is for now.

prajwolrg (dismissed):

crates/db/types/src/mmr_helpers.rs
barakshani: I was saying to remove the comment "or we've reached a peak".

barakshani (approved):

crates/db/types/src/mmr_helpers.rs
bewakes: Ohh, I see. Done.

barakshani: we still need the "stop here" command.

crates/db/types/src/mmr_helpers.rs
barakshani: @bewakes is this in or out?

crates/db/store-sled/src/global_mmr/schemas.rs
barakshani: @bewakes please check and decide.

bewakes: oops, my bad. I had for some reason stashed the changes to switch branch and forgot to re-apply them. Done now. Phew!

crates/db/types/src/mmr_helpers.rs
bewakes: This is in now.

PR #1228, opened 2025-12-25
ASM admin subprotocol tests

evgenyzdanovich:
## Description

ASM integration tests:
- minimal regtest + asm worker harness
- basic admin subprotocol tests
- basic admin to checkpoint (cross subprotocol) tests

Opus 4.5 was used with the heavy context of sps-65 (admin subprotocol).

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/tests/Cargo.toml
+++ b/tests/Cargo.toml
@@ -7,13 +7,6 @@ version = "0.3.0-alpha.1"
 workspace = true
 
 [dependencies]
-anyhow.workspace = true
-bitcoin.workspace = true
-bitcoind-async-client.workspace = true
-borsh.workspace = true
-corepc-node.workspace = true
-rand.workspace = true
-rand_chacha.workspace = true
 strata-asm-common.workspace = true
 strata-asm-manifest-types.workspace = true
 strata-asm-proto-administration.workspace = true
@@ -32,6 +25,14 @@ strata-state.workspace = true
 strata-tasks.workspace = true
 strata-test-utils-btcio.workspace = true
 strata-test-utils-l2.workspace = true
+
+anyhow.workspace = true
+bitcoin.workspace = true
+bitcoind-async-client.workspace = true
+borsh.workspace = true
+corepc-node.workspace = true
+rand.workspace = true
+rand_chacha.workspace = true
 tokio.workspace = true
 
 [dev-dependencies]


delbonis (approved):
This looks pretty good, I'm happy that we're doing something with this.

However the tests themselves are a bit huge.  I'm wondering if there's something that we could do to structure these better and reduce the duplication.

Also also, is that a redundant `#[cfg(tests)] mod tests` in `tests/asm/core.rs`?

crates/asm/spec/src/lib.rs
```diff
@@ -82,10 +90,33 @@ impl StrataAsmSpec {
             operator_fee: BitcoinAmount::ZERO,
         };
 
+        // For now, use the same operator config for admin roles
+        // TODO: Add proper admin config to RollupParams
```
prajwolrg: I created a [ticket](https://alpenlabs.atlassian.net/browse/STR-2024) so we don't forget it. Let's add the ticket directly

crates/test-utils/l2/src/lib.rs
```diff
@@ -143,6 +143,16 @@ fn make_dummy_operator_pubkeys_with_seed(seed: u64) -> OperatorPubkeys {
     OperatorPubkeys::new(pk.into(), pk.into())
 }
 
+/// Gets the operator secret key for testing.
+/// This matches the key generation in `make_dummy_operator_pubkeys_with_seed(0)`.
```
prajwolrg, line 144: I forgot why we needed same rollup parameter generation everytime. But I'm not sure if we need this anymore.

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+    let admin_state = section.try_to_state::<AdministrationSubprotocol>()?;
+    Ok(admin_state)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_admin_sequencer_update() {
```
prajwolrg: The test here, I think needs to also see if the interaction between the subprotocol is handled properly. It needs to check whether the sequencer on the checkpoint subprotocol was indeed updated. A lot of the check here can be done as a unit test inside of the checkpoint subprotocol directly.

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+        assert_eq!(
+            admin_state.next_update_id(),
+            0,
+            "Update ID should not change for rejected tx"
+        );
+
+        println!("✓ Verified admin tx with insufficient signatures was rejected");
+        println!("✓ No state changes occurred");
+        println!("Block processed - threshold failure should be rejected by ASM");
+        println!("Multisig threshold test completed");
```
prajwolrg: I wonder the value of printing this at the end.

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+        );
+
+        // No queued updates (sequencer updates apply immediately)
+        assert_eq!(admin_state.queued().len(), 0, "No updates should be queued");
+
+        println!("✓ Verified second tx with duplicate seqno was rejected");
+        println!("✓ Only first transaction was processed");
+        println!("Both blocks processed - replay should be rejected by ASM");
+        println!("Sequence number replay test completed");
+    }
```
prajwolrg: Same as above.

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_admin_operator_set_update() {
+        // Test updating the operator set (add/remove operators)
```
prajwolrg: Similar to above, this should check if the operator set was indeed added removed in the bridge subprotocol.

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+                        && out.script_pubkey.as_bytes().len() >= 7
+                        && &out.script_pubkey.as_bytes()[2..6] == b"ALPN"
+                })
+            })
+            .count();
+
+        println!(
+            "Block {} contains {} admin transactions",
+            block_hash, admin_tx_count
+        );
```
prajwolrg: We can use `strata_l1_xfmt::ParseConfig` to parse the transaction directly.

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+            processed_count, admin_tx_count);
+
+        println!(
+            "✓ Verified {} admin transactions in block, {} processed successfully",
+            admin_tx_count, processed_count
+        );
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_admin_sequencer_update_end_to_end() {
```
prajwolrg: Not sure if this is end to end. Sequencer update must be updated in the checkpoint state.

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+        let admin_tx = harness
+            .build_funded_admin_tx(admin_payload, tx_type, fee)
+            .await
+            .expect("Failed to build funded admin tx");
+
+        println!("✓ Built funded admin transaction");
+        println!("  - TXID: {}", admin_tx.compute_txid());
+        println!("  - Inputs: {}", admin_tx.input.len());
+        println!("  - Outputs: {}", admin_tx.output.len());
+        println!("  - Witness items: {}", admin_tx.input[0].witness.len());
```
prajwolrg: Same as above. I'm not sure of the value of printing these.

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+        println!("✓ Commit transaction funded taproot address");
+        println!("✓ Reveal transaction submitted to mempool");
+        println!("✓ Block mined containing reveal transaction");
+        println!("✓ ASM processed block and updated state");
+        println!("\nThis test demonstrates the complete end-to-end flow:");
+        println!("  1. Create signed admin payload (action + multisig signatures)");
+        println!("  2. Build funded Bitcoin transaction with envelope");
+        println!("  3. Submit to regtest and mine block");
+        println!("  4. ASM extracts and processes admin action");
+    }
```
prajwolrg: Same as above. Printing only at the end doesn't help in debugging. I'm not sure of adding all these.

evgenyzdanovich:
@prajwolrg yes thanks for the feedback - cleaning up tests themselves, there are a lot of useless prints that do nothing, I agree, and some tests are a bit misleading. 

the pr is in draft, I primarily released it (in a messy state) to show _in which direction I see the asm tests moving_, and how the complete test infra would look like.

crates/asm/subprotocols/admin/src/lib.rs
```diff
@@ -10,7 +10,8 @@ mod config;
 mod error;
 mod handler;
 mod queued_update;
-mod state;
+pub mod state;
```
storopoli: This is the only `pub` here. Can't you re-export whatever you need from `state` below like all other stuff?
We are trying to avoid `pub mod` and direct implementers to use the re-exported stuff only.

crates/test-utils/l2/src/lib.rs
```diff
@@ -143,6 +143,16 @@ fn make_dummy_operator_pubkeys_with_seed(seed: u64) -> OperatorPubkeys {
...
+pub fn get_test_operator_secret_key() -> SecretKey {
+    let mut rng = StdRng::seed_from_u64(0);
+    let sk = SecretKey::new(&mut rng);
+    // Ensure the key has even parity for taproot compatibility
+    let even_sk = EvenSecretKey::from(sk);
+    SecretKey::from(even_sk)
+}
```
storopoli, line 148: I think we should return an `EvenSecretKey` and propagate that forwards to anything that we need. Otherwise we'll keep having to wrangle all over the place to make sure that the SKs are even downstream.

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+use strata_params::OperatorConfig;
+use strata_primitives::buf::Buf32;
+use strata_test_utils_l2::get_test_operator_secret_key;
+
+/// Helper to create test admin multisig configurations from test params.
+/// Extracts the operator XOnly keys from params and reconstructs them with even parity,
+/// matching how the ASM spec initializes authorities.
+fn create_test_admin_config_from_params(
+    params: &strata_params::Params,
+) -> (ThresholdConfig, Vec<SecretKey>) {
```
storopoli:
```suggestion
) -> (ThresholdConfig, Vec<EvenSecretKey>) {
```

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+};
+use strata_bridge_types::OperatorPubkeys;
+use strata_crypto::threshold_signature::{CompressedPublicKey, ThresholdConfig};
+use strata_params::OperatorConfig;
+use strata_primitives::buf::Buf32;
+use strata_test_utils_l2::get_test_operator_secret_key;
+
+/// Helper to create test admin multisig configurations from test params.
+/// Extracts the operator XOnly keys from params and reconstructs them with even parity,
+/// matching how the ASM spec initializes authorities.
```
storopoli: If you take an `EvenSecretKey` you don't need to worry with parity on `XOnlyPublicKey`s.

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+        let tx_type = action.tx_type();
+        let sighash = action.compute_sighash(seqno);
+        let signature_set = create_signature_set(&admin_privkeys, &signer_indices, sighash);
+        let signed_payload = SignedPayload::new(action, signature_set);
+        let admin_payload = borsh::to_vec(&signed_payload).expect("Failed to serialize");
+
+        println!("Created admin transaction successfully");
+        println!("Admin tx has 1 inputs, 1 outputs");
+        println!("Admin transaction creation and setup verified");
+        let fee = bitcoin::Amount::from_sat(1000);
```
storopoli: import `Amount`

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+            .submit_and_mine_admin_tx(&admin_tx)
+            .await
+            .expect("Failed to submit and mine admin tx");
+
+        println!("Mined and submitted block: {}", block_hash);
+
+        // Wait for ASM worker to process
+        let target_height = harness.genesis_height + 1;
+        harness
+            .wait_for_height(target_height, std::time::Duration::from_secs(5))
```
storopoli: import `Duration`

tests/harness/test_harness.rs
```diff
@@ -0,0 +1,616 @@
...
+    /// or times out after the specified duration.
+    pub async fn wait_for_height(
+        &self,
+        target_height: u64,
+        timeout: Duration,
+    ) -> anyhow::Result<()> {
+        let start = std::time::Instant::now();
+        loop {
+            if start.elapsed() > timeout {
+                anyhow::bail!("Timeout waiting for height {}", target_height);
```
storopoli:
```suggestion
                anyhow::bail!("Timeout waiting for height {target_height}");
```

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+            *queued_update.id(),
+            0,
+            "First queued update should have ID 0"
+        );
+
+        println!("✓ Verified operator set update was queued (ID=0)");
+        println!("✓ Update will activate after confirmation depth");
+        println!("Operator set update submitted - should be queued pending confirmation depth");
+        println!("Admin queuing test completed");
+    }
```
storopoli: avoid `println!` and use `debug!` or `trace!` or `warn!` whatever
@prajwolrg already pointed out the printing stuff here

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+        // Test that signatures from unauthorized keys are rejected
+        let harness = create_test_harness()
+            .await
+            .expect("Failed to create test harness");
+
+        println!("Testing admin signature authorization");
+
+        // Create a transaction signed with WRONG key (not the operator key)
+        let secp = Secp256k1::new();
+        let wrong_privkey = SecretKey::new(&mut OsRng);
```
storopoli: `EvenSecretKey`

storopoli: If needed you can also add a `EvenSecretKey::new(&mut OsRng)` method.

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+
+        // Create signed payload with wrong key
+        let tx_type = action.tx_type();
+        let sighash = action.compute_sighash(seqno);
+        let signature_set = create_signature_set(&wrong_privkeys, &signer_indices, sighash);
+        let signed_payload = SignedPayload::new(action, signature_set);
+        let admin_payload = borsh::to_vec(&signed_payload).expect("Failed to serialize");
+
+        println!("Created admin tx signed with unauthorized key");
+        let fee = bitcoin::Amount::from_sat(1000);
```
storopoli: import `Amount`

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+            .expect("Failed to build second admin tx");
+
+        let block_hash_2 = harness
+            .submit_and_mine_admin_tx(&admin_tx_2)
+            .await
+            .expect("Failed to submit and mine second admin tx");
+
+        println!(
+            "Submitted second admin tx (replay) in block: {}",
+            block_hash_2
```
storopoli:
```suggestion
            "Submitted second admin tx (replay) in block: {block_hash_2}"
```

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+        );
+
+        // Wait for ASM to process second transaction
+        let target_height_2 = harness.genesis_height + 2;
+        harness
+            .wait_for_height(target_height_2, std::time::Duration::from_secs(5))
```
storopoli: import `Duration`

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+        let cancel_signature_set = create_signature_set(&privkeys, &signer_indices, cancel_sighash);
+        let cancel_tx_type = cancel_multisig_action.tx_type();
+        let cancel_signed_payload =
+            SignedPayload::new(cancel_multisig_action, cancel_signature_set);
+        let cancel_admin_payload =
+            borsh::to_vec(&cancel_signed_payload).expect("Failed to serialize");
+
+        println!(
+            "Created cancel action tx with seqno=1 (targets update_id={})",
+            target_update_id
```
storopoli:
```suggestion
            "Created cancel action tx with seqno=1 (targets update_id={target_update_id})"
```

tests/asm/admin.rs
```diff
@@ -0,0 +1,1161 @@
...
+        );
+
+        let _cancel_block_hash = harness
+            .submit_and_mine_admin_tx(&cancel_tx)
+            .await
+            .expect("Failed to submit and mine cancel tx");
+
+        let target_height = harness.genesis_height + 2; // Two blocks
+        harness
+            .wait_for_height(target_height, std::time::Duration::from_secs(5))
```
storopoli: import `Duration`

storopoli (changes requested):
very good minor nits.
I am annoyed with the absolute paths and I've created https://github.com/alpenlabs/alpen/pull/1251

tests/asm/admin.rs
evgenyzdanovich: removed misleading names, added some cross subproto validation (for now admin -> checkpoint)

tests/asm/admin.rs
evgenyzdanovich: pending work to create `admin_to_bridge` (similarly to `admin_to_checkpoint`).

tests/asm/admin.rs
prajwolrg: I don't think this is correct now. For the admin subprotocol we don't require `EvenSecretKey` or even parity. Since we are using ECDSA signature instead of Schnorr signature, I don't think this is needed anymore.

crates/asm/subprotocols/admin/src/handler.rs
```diff
@@ -238,19 +241,26 @@ mod tests {
         }
     }
 
-    fn create_test_params() -> (AdministrationSubprotoParams, Vec<SecretKey>, Vec<SecretKey>) {
+    fn create_test_params() -> (
+        AdministrationSubprotoParams,
+        Vec<EvenSecretKey>,
+        Vec<EvenSecretKey>,
+    ) {
```
prajwolrg: We don't need to enforce EvenSecretKey for the admin. The signature is ECDSA based instead of Schnorr based. Even parity was required for taproot.

tests/harness/test_harness.rs
```diff
@@ -0,0 +1,597 @@
...
+    /// 3. Has SPS-50 compliant OP_RETURN tag
+    ///
+    /// # Arguments
+    /// * `subprotocol_id` - SPS-50 subprotocol ID (0=admin, 1=checkpoint, 2=bridge)
+    /// * `tx_type` - Transaction type within the subprotocol
+    /// * `payload` - Serialized payload to embed in witness
+    pub async fn build_envelope_tx(
+        &self,
+        subprotocol_id: u8,
+        tx_type: u8,
```
prajwolrg: nit: Use `SubprotocolId`, `TxType` instead of u8.

tests/harness/test_harness.rs
```diff
@@ -0,0 +1,597 @@
...
+        // Build SPS-50 compliant OP_RETURN tag
+        let mut sps50_tag = Vec::with_capacity(6);
+        sps50_tag.extend_from_slice(self.params.rollup().magic_bytes.as_bytes()); // 4 bytes
+        sps50_tag.push(subprotocol_id); // 1 byte
+        sps50_tag.push(tx_type); // 1 byte
+
+        let op_return_output = TxOut {
+            value: Amount::ZERO,
+            script_pubkey: ScriptBuf::new_op_return(PushBytesBuf::try_from(sps50_tag)?),
+        };
```
prajwolrg: nit: we can use `TagData` type together with `ParseConfig` to get the `ScriptBuf`.

prajwolrg (approved):
Brilliant work! Thank you! 

I just have a few nits for the tests.

tests/harness/test_harness.rs
evgenyzdanovich: done

tests/harness/test_harness.rs
evgenyzdanovich: done

crates/asm/subprotocols/admin/src/handler.rs
evgenyzdanovich: brought back the logic parts to work with `SecretKey` as before.

tests/asm/admin.rs
evgenyzdanovich: thanks, brought back

tests/Cargo.toml
```diff
@@ -7,9 +7,55 @@ version = "0.3.0-alpha.1"
...
+strata-l1-txfmt.workspace = true
+strata-merkle.workspace = true
+strata-params.workspace = true
+strata-predicate.workspace = true
+strata-primitives.workspace = true
+strata-state.workspace = true
+strata-tasks.workspace = true
+strata-test-utils-btcio.workspace = true
+strata-test-utils-l2.workspace = true
+tokio.workspace = true
```
delbonis: Dep grouping?

delbonis (changes requested):
Functionally this looks pretty good and extensive.  I'm wondering if there's something we could do to reduce some the duplication between the various test cases without making it really convoluted.

However I'm a little bit sad to see us deepening the requirement on having bitcoind available to be able to run the `cargo test` tests.  We already have this so technically it's not making us take a qualitative step back but it does add *more* we'd have to change.  I'm not going to say we should rework this because this does do a lot that would be annoying to rewrite.

tests/Cargo.toml
evgenyzdanovich: done, thanks!

PR #836, opened 2025-05-29
chore: add workspace lints

storopoli:
## Description

This PR introduces a comprehensive set of workspace-level lints to improve code quality, maintainability, and adherence to Rust best practices across all crates in the workspace.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Configured Lints:

- Code Quality & Debugging:
  - `rust.missing_debug_implementations` (warn) - Ensures all public types implement `Debug`, improving debuggability and developer experience.
  - `rust.unused_must_use` (deny) - Prevents accidentally ignoring important return values that should be handled

- Modern Rust Practices:
  - `rust.rust_2018_idioms` (deny) - Enforces modern Rust 2018+ idioms and patterns, keeping the codebase up-to-date with current best practices
  - ~`clippy.missing_const_for_fn` (deny) - Identifies functions that could be marked as `const`, enabling compile-time evaluation and better optimization~

- API Design & Visibility:
  - `rust.unreachable_pub` (warn) - Identifies public items that are never used externally, helping maintain clean public APIs
  - `rust.unused_crate_dependencies` (deny) - Prevents bloat by catching unused dependencies in `Cargo.toml` files

- Documentation:
  - `rustdoc.all` (warn) - Enables comprehensive documentation linting to ensure high-quality docs
  - `rust.missing_docs` (commented) - Reserved for future enablement to enforce documentation coverage

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [x] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [x] New and existing tests pass with my changes.

## Related Issues

STR-207 but I also went overboard

PR #1229, opened 2025-12-26
feat!: Use AccountSerial and SubjectId for DRT destination addressing

prajwolrg:
## Description

This PR refactors the Deposit Request Transaction (DRT) auxiliary data and Deposit Transaction (DT)  to use `AccountSerial` and `SubjectId` instead of fixed-length byte arrays. Relevant changes include:
-  Introduces `SubjectBytes`-Variable-length subject identifier (0-32 bytes) that converts to `SubjectId`.
- Remove EE Address from DT: Since we fetch DRT in DT handling, we can directly use it's address without requiring a separate destination address in DT.
- `DepositLog` now uses `AccountSerial` and `SubjectId`.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [x] Dependency Update

## Checklist

### DIFF
--- a/.github/scripts/grafana/pyproject.toml
+++ b/.github/scripts/grafana/pyproject.toml
@@ -4,6 +4,4 @@ version = "0.1.0"
 description = "Grafana dashboard provisioning and metrics collection scripts"
 requires-python = ">=3.10"
 
-dependencies = [
-  "requests>=2.31.0",
-]
+dependencies = ["requests>=2.31.0"]


bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -111,7 +112,13 @@ fn prepare_deposit_request(
             .address
     };
 
-    let header_aux = DrtHeaderAux::new(recovery_public_key.serialize(), alpen_address.into());
+    let alpen_subject_bytes =
+        SubjectBytes::try_new(alpen_address.to_vec()).expect("must be valid subject bytes");
+    let header_aux = DrtHeaderAux::new(
+        recovery_public_key.serialize(),
+        AccountSerial::zero(),
```
storopoli: This deserves a comment that all accounts have `0` and reserved (special) accounts are `1.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
```diff
@@ -1,25 +1,49 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct DrtHeaderAux {
     recovery_pk: [u8; 32],
-    // TODO:PG - Intentionally using 20 bytes for now. Will be properly handled as part of https://alpenlabs.atlassian.net/browse/STR-1950
-    ee_address: [u8; 20],
+
+    /// [`AccountSerial`] of the destination account.
+    ///
+    /// We use [`AccountSerial`] instead of [`AccountId`](strata_identifiers::AccountId) to
+    /// minimize onchain cost. Since account serials are immutable and uniquely identify accounts,
```
storopoli: Sorry `AccountSerial` is not unique it is either 0 or 1 and most accounts will be 0. This docstring is very wrong.
I think you've meant `deposit_request::SubjectBytes` no which is the field below.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
```diff
@@ -29,8 +53,12 @@ impl DrtHeaderAux {
...
 
     /// Returns the execution environment address
-    pub fn ee_address(&self) -> &[u8; 20] {
-        &self.ee_address
+    pub fn dest_acct_serial(&self) -> &AccountSerial {
+        &self.dest_acct_serial
+    }
+
+    pub fn dest_subject(&self) -> &SubjectBytes {
+        &self.dest_subject
```
storopoli: I think these functions can be `const` no?

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
```diff
@@ -53,29 +81,91 @@ impl DrtHeaderAux {
     }
 }
 
+impl Codec for DrtHeaderAux {
```
storopoli: Why do we need the manual impl?
Can we just keep using the `#[derive(Codec)]`?
Is this because one of the new fields does not impl `Codec`?
If so, should we add the derive to this new field, and then keep the derive here?

crates/asm/txs/bridge-v1/src/deposit_request/parse.rs
```diff
@@ -59,7 +60,17 @@ mod tests {
...
+    ///
+    /// This is the size of the recovery public key (32 bytes) plus the size of
+    /// [`AccountSerial`] (4 bytes), when the subject bytes are empty.
+    const MIN_AUX_LEN: usize = 32 + std::mem::size_of::<AccountSerial>();
+
+    /// Maximum encoded auxiliary data length in bytes.
+    ///
+    /// This is [`MIN_AUX_LEN`] plus the maximum subject bytes length ([`SUBJ_ID_LEN`]),
+    /// representing the case where the subject identifier uses all 32 bytes.
+    const MAX_AUX_LEN: usize = MIN_AUX_LEN + SUBJ_ID_LEN;
```
storopoli: Suggest we add a very stupid "canary in the coal mine" test that validates that `MIN_AUX_LEN` is `36` and `MAX_AUX_LEN` is `64` bytes. It will be helpful to avoid nasty lengthy debugging sessions in infra environments in the future.

crates/asm/txs/bridge-v1/src/deposit_request/subject_bytes.rs
```diff
@@ -0,0 +1,131 @@
...
+    TooLong(usize, usize),
+}
+
+/// Variable-length [`SubjectId`] bytes.
+///
+/// Subject IDs are canonically [`SUBJ_ID_LEN`] bytes per the account system specification, but in
+/// practice many subject IDs are shorter. This type stores the variable-length byte representation
+/// to optimize DA costs by avoiding unnecessary zero padding in the on-chain deposit descriptor.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SubjectBytes(Vec<u8>);
```
storopoli: I would prefer to call it `SubjectIdBytes` to make it explicit that it is the `Bytes` of `SubjectId`.
There might be probably a more ergonomic way to do this. I don't know, maybe @delbonis might have some ideas. Trait extensions etc...

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/subject_id_bytes.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    /// Returns the length of the subject bytes.
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    /// Returns true if the subject bytes are empty.
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
```
storopoli: Probably a `const` fn

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/subject_id_bytes.rs
```diff
@@ -0,0 +1,131 @@
...
+        if bytes.len() > SUBJ_ID_LEN {
+            return Err(SubjectBytesError::TooLong(bytes.len(), SUBJ_ID_LEN));
+        }
+        Ok(Self(bytes))
+    }
+
+    /// Returns the raw, unpadded subject bytes.
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.0
+    }
```
storopoli: probably a `const` fn

crates/bridge-types/src/bridge_ops.rs
```diff
@@ -75,21 +76,21 @@ pub struct DepositIntent {
...
+    /// Destination subject identifier within the execution environment.
+    /// For EVM, the first 20 bytes hold the address and the remaining bytes must be zero.
+    dest_ident: SubjectId,
 }
 
 impl DepositIntent {
-    pub fn new(amt: BitcoinAmount, dest_ident: Vec<u8>) -> Self {
+    pub fn new(amt: BitcoinAmount, dest_ident: SubjectId) -> Self {
         Self { amt, dest_ident }
     }
```
storopoli: `const` fn

crates/bridge-types/src/bridge_ops.rs
```diff
@@ -75,21 +76,21 @@ pub struct DepositIntent {
...
 
     pub fn amt(&self) -> u64 {
         self.amt.to_sat()
     }
 
-    pub fn dest_ident(&self) -> &[u8] {
+    pub fn dest_ident(&self) -> &SubjectId {
         &self.dest_ident
     }
```
storopoli, line 94: `const` fn

crates/eectl/src/errors.rs
```diff
@@ -13,7 +14,7 @@ pub enum EngineError {
     AmountConversion(u64),
 
     #[error("invalid address {0:?}")]
-    InvalidAddress(Vec<u8>),
+    InvalidAddress(SubjectId),
```
storopoli: I think we can do `Display` now in this error?
#[error("invalid address {0}")]

crates/evmexec/src/engine.rs
```diff
@@ -31,9 +32,22 @@ use crate::{
     http_client::EngineRpc,
 };
 
-fn address_from_slice(slice: &[u8]) -> Option<Address> {
-    let slice: Option<[u8; 20]> = slice.try_into().ok();
-    slice.map(Address::from)
+const EVM_ADDR_LEN: usize = 20;
```
storopoli: should we move this to `strata-identifiers`?

crates/evmexec/src/engine.rs
```diff
@@ -31,9 +32,22 @@ use crate::{
...
+    let mut address_bytes = [0u8; EVM_ADDR_LEN];
+    address_bytes.copy_from_slice(&bytes[..EVM_ADDR_LEN]);
+    Some(Address::from(address_bytes))
+}
+
+fn address_to_subject(address: Address) -> SubjectId {
+    let mut buf = [0u8; SUBJ_ID_LEN];
+    buf[0..EVM_ADDR_LEN].copy_from_slice(address.as_slice());
+    SubjectId::new(buf)
 }
```
storopoli: These functions need to be `proptest!`ed.
Also I think that a valid `SubjectId` non-EVM address might have one of the remaining 12 bytes as the `0` byte no?
Then we're screwed with this

```rs
if bytes[EVM_ADDR_LEN..].iter().any(|&byte| byte != 0) {
    return None;
}
```

crates/identifiers/src/acct.rs
```diff
@@ -153,10 +154,13 @@ type RawSubjectId = [u8; SUBJ_ID_LEN];
...
     Decode,
     Encode,
+    Serialize,
+    Deserialize,
+    Arbitrary,
     BorshSerialize,
     BorshDeserialize,
 )]
-pub struct SubjectId(RawSubjectId);
+pub struct SubjectId(#[serde(with = "hex::serde")] RawSubjectId);
```
storopoli, line 178: Why not add `Codec` here?

crates/proof-impl/evm-ee-stf/src/utils.rs
```diff
@@ -51,3 +53,11 @@ const fn gwei_to_sats(gwei: u64) -> u64 {
     // 1 BTC = 10^8 sats = 10^9 gwei
     gwei / 10
 }
+
+const EVM_ADDR_LEN: usize = 20;
```
storopoli: move `strata_identifiers`?
This is the second time you are defining the same `const` in this PR in different files.
And both modules are in crates that already use `strata-identifiers`.

crates/evmexec/src/engine.rs
storopoli: Given that `SubjectId` is 32 bytes and can be a hash. And given that cryptographic secure hashes are uniform distributed, then there's a 1 in 256 chance of a byte being the `0` byte no?

barakshani: Isn't it that this implementation is tailored for Alpen EE?
I mean, it's an interesting hybrid where one the one hand we do allow 32 bytes, and not only 20, but on the other hand, we restrict to the first 20 bytes.
Seems to me we only consider EVM subjects for the time being.

bin/alpen-cli/src/cmd/deposit.rs
prajwolrg: Addressed in [5c1f624](https://github.com/alpenlabs/alpen/pull/1229/commits/5c1f6242b956d3a29ef2e08d2989cbd87890dd2b). 

In the current implementation of `Chainstate`, we only allow for a single AccountId, but that will change with the new OL we are building.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
prajwolrg: It mentions [here](https://www.notion.so/SPS-acct-sys-Strata-OL-account-system-1e4901ba000f803aa54cf6834883f466) that the serials are unique.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
prajwolrg: Yes. They satisfy the constraints of `const` functions. 

Fixed in [584300e](https://github.com/alpenlabs/alpen/pull/1229/commits/584300e64e93dd295484645a32545afcb0343144).

I do wonder why they haven't been used in other places in our codebase.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
prajwolrg: That's not the reason. Manual implementation minimize onchain cost.

Since `dest_subject` has variable length and is the last field, we can save bytes by not encoding its length explicitly—instead reading all remaining bytes during decoding after the fixed-size fields (`recovery_pk` and `dest_acct_serial`).

I've added the docstring in [e7c22d1](https://github.com/alpenlabs/alpen/pull/1229/commits/e7c22d1ab9243165bba3815600e6529ec419601d).

crates/asm/txs/bridge-v1/src/deposit_request/parse.rs
prajwolrg: I don't understand.

crates/asm/txs/bridge-v1/src/deposit_request/subject_bytes.rs
prajwolrg: Renamed in [56d87433](https://github.com/alpenlabs/alpen/pull/1229/commits/56d87433).

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/subject_id_bytes.rs
prajwolrg: cannot perform non-const deref coercion on `std::vec::Vec<u8>` in constant functions attempting to deref into `[u8]`.

crates/evmexec/src/engine.rs
prajwolrg: I'm averse to moving this to strata-identifiers. If `strata-identifiers` is the place we put it, that means everytime we add a new execution environment in the future, we need to update `strata-identifiers` with the new constants.

I believe each execution environment should have it's own constant for things like these. But right now it is in two different places, and I'm not very happy about it. Maybe @sapinb can suggest a good place where this should live.

crates/evmexec/src/engine.rs
prajwolrg: Agree with Barak. This is specific to EVM. 

With multi-EE, with different addresses (even with same 20 byte length), `AccountSerial` provides the information of where it should be routed to.

We can do a better job here with how we are creating the engine payload, where this functions is actually used, by maybe logging that we somehow passed in the wrong information here, but I'm not sure if this PR should do that.

crates/identifiers/src/acct.rs
prajwolrg: See https://github.com/alpenlabs/alpen/pull/1229#discussion_r2648304941.

crates/proof-impl/evm-ee-stf/src/utils.rs
prajwolrg: See https://github.com/alpenlabs/alpen/pull/1229#discussion_r2648320760.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
storopoli: There is a `clippy` lint (`clippy.missing_const_for_fn`) and was initially proposed in  https://github.com/alpenlabs/alpen/pull/836 but overruled by @delbonis (reason in https://github.com/alpenlabs/alpen/pull/836#pullrequestreview-2882391202).

Hence, I do keep in mind that some functions can (and should be) `const` in reviews.

crates/asm/txs/bridge-v1/src/deposit_request/parse.rs
storopoli: So right now we don't have a way to assert in the codebase that is robust to changes, that the `MIN_AUX_LEN` is 36 bytes and the `MAX_AUX_LEN` is 64 bytes, which is what the dock strings say.

Also, these values can change under the hood in future PRs and can wreak havoc in the future in a development or staging environment.
Adding a "canary", simple test that validates those values, is a way to safely keep that those values won't change accidentally. Except when it was intended to.

I know that you probably won't change this accidentally, but having those simple, very stupid tests is a good idea, I think. 

Does that clear my idea for the suggestion?

crates/evmexec/src/engine.rs
storopoli: Can you open a ticket so that we don't forget this issue when we move to a multi-EE codebase?

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
```diff
@@ -53,29 +81,96 @@ impl DrtHeaderAux {
...
+        let dest_acct_serial = AccountSerial::decode(dec)?;
+
+        // Read remaining bytes as address - we need to read from a buffer
+        // Since Decoder doesn't provide a way to read all remaining bytes,
+        // this decode assumes the input has already been sized correctly
+        let mut dest_subject_bytes = Vec::new();
+        // Try to read bytes until we hit end of buffer
+        while let Ok(byte) = dec.read_arr::<1>() {
+            dest_subject_bytes.push(byte[0]);
+        }
```
delbonis: Ah this is kinda a hack, I wonder if it's not worth using `Codec` for this?

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
```diff
@@ -1,36 +1,64 @@
...
+
+    /// [`AccountSerial`] of the destination account.
+    ///
+    /// We use [`AccountSerial`] instead of [`AccountId`](strata_identifiers::AccountId) to
+    /// minimize onchain cost. Since account serials are immutable and uniquely identify accounts,
+    /// they provide the same identification guarantees as `AccountId` while being more
+    /// space-efficient on-chain.
+    dest_acct_serial: AccountSerial,
+
+    /// [`SubjectId`](`strata_identifiers::SubjectId`) within the destination account.
```
barakshani: are you sure you don't mean to write `SubjectIdBytes`?

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
delbonis, line 1: I had wrote this [deposit descriptor spec](https://www.notion.so/SPS-deposit-desc-295901ba000f80fd8175c7d0a49ef43e) and shared it in Slack some weeks ago when this was a point of discussion.  It has considerations for variable length integers to save bytes and flag fields to be future-compatible.  That should probably be implemented in the strata-common repo.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
```diff
@@ -1,36 +1,64 @@
...
+    ///
+    /// We use [`SubjectBytes`] instead of `SubjectId` to minimize onchain cost.
+    dest_subject: SubjectIdBytes,
 }
```
delbonis, line 45: This is sorta mixing in bridge protocol related fields (`recovery_pk`) with orchestration-layer fields that describe "what we do with the funds" (`dest_acct_serial`, `dest_subject`).  To match the structure of the data represented in a deposit desc, we'd split these out into their own type.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
```diff
@@ -1,36 +1,64 @@
...
         &self.recovery_pk
     }
 
     /// Returns the execution environment address
-    pub fn ee_address(&self) -> &[u8; 20] {
-        &self.ee_address
+    pub const fn dest_acct_serial(&self) -> &AccountSerial {
+        &self.dest_acct_serial
+    }
+
```
barakshani, line 82:
```suggestion
    /// Returns the destination address within the execution environment

File: crates/asm/txs/bridge-v1/src/deposit_request/aux.rs [hunk 20]
```diff
@@ -1,36 +1,64 @@
...
         }
     }
 
     /// Returns the recovery public key
-    pub fn recovery_pk(&self) -> &[u8; 32] {
+    pub const fn recovery_pk(&self) -> &[u8; 32] {
         &self.recovery_pk
     }
 
     /// Returns the execution environment address
```
barakshani:
```suggestion
    /// Returns the execution environment serial number

**Review by delbonis — CHANGES_REQUESTED** [review]

File: crates/asm/txs/bridge-v1/src/deposit_request/aux.rs [hunk 21]
```diff
@@ -1,36 +1,64 @@
...
+    ///
+    /// We use [`AccountSerial`] instead of [`AccountId`](strata_identifiers::AccountId) to
+    /// minimize onchain cost. Since account serials are immutable and uniquely identify accounts,
+    /// they provide the same identification guarantees as `AccountId` while being more
+    /// space-efficient on-chain.
+    dest_acct_serial: AccountSerial,
+
+    /// [`SubjectId`](`strata_identifiers::SubjectId`) within the destination account.
+    ///
+    /// We use [`SubjectBytes`] instead of `SubjectId` to minimize onchain cost.
```
barakshani: id bytes?

crates/evmexec/src/engine.rs
```diff
@@ -31,9 +32,22 @@ use crate::{
...
+        return None;
+    }
+    let mut address_bytes = [0u8; EVM_ADDR_LEN];
+    address_bytes.copy_from_slice(&bytes[..EVM_ADDR_LEN]);
+    Some(Address::from(address_bytes))
+}
+
+fn address_to_subject(address: Address) -> SubjectId {
+    let mut buf = [0u8; SUBJ_ID_LEN];
+    buf[0..EVM_ADDR_LEN].copy_from_slice(address.as_slice());
```
barakshani: This is lazy, you should also check if last bytes of the given address are `0`.

Go figure, maybe swap providers or frequent bridge users will create vanity address trailed by zeros to reduce onchain footprint and pay less.

crates/evmexec/src/engine.rs
barakshani: It depends on this [comment](https://github.com/alpenlabs/alpen/pull/1229/files#r2648320760), no? Right now this lives inside evm-related code.

crates/evmexec/src/engine.rs
prajwolrg: > This is lazy, you should also check if last bytes of the given address are 0.

I don't understand what you mean. Why should we need to check if the last bytes of the given address are `0` in `address_to_subject`? There are only 20 bytes in `revm_primitives::Address`. 

> maybe swap providers or frequent bridge users will create vanity address trailed by zeros to reduce onchain footprint and pay less

I hadn't realize you could do that. Is this something we want to prevent? or is alright to have?

crates/asm/txs/bridge-v1/src/deposit_request/parse.rs
prajwolrg: But these values are only part of the test. We don't use this anywhere else.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
prajwolrg: > I wonder if it's not worth using `Codec` for this?

What do you mean by using `Codec` for this? We are implementing a custom `Codec` implementation. We want to avoid encoding the length explicitly since this information is in the SPS-50 header. I couldn't find a way in the `strata_codec::Decoder` to read all the remaining bytes.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
```diff
@@ -53,29 +81,96 @@ impl DrtHeaderAux {
...
+
+        Ok(Self {
+            recovery_pk,
+            dest_acct_serial,
+            dest_subject,
+        })
+    }
+}
+
+impl<'a> Arbitrary<'a> for DrtHeaderAux {
```
barakshani: Do we use this in tests anywhere?

crates/evmexec/src/engine.rs
barakshani: Maybe I misunderstood where this is being used -- I see now that we first initiate 32 bytes (SUBJ_ID_LEN) to zero, and then copy the first 20 bytes (EVM_ADDR_LEN) -- while I thought that we take the address, as is, and convert it to the SubjectId that will present in the DRT. In the latter case, the point is that if the last byte(s) are `0`, we don't even have to include it in the DRT.
This should actually be in the Alpen-CLI, not here, and here we seem to always initiate with zeros and then copy the bytes we are given, so I think it's fine, and I'm taking back this comment (actually here we already use the given types (Address, SubjectId) which have fixed length, so I think the real place where this is meaningful in this PR is crates/asm/txs/bridge-v1/src/deposit_request/subject_id_bytes.rs, which seems to be fine as well -- however see my comments on DrtHeaderAux's arbitrary).

>I hadn't realize you could do that. Is this something we want to prevent? or is alright to have?

No real reason to prevent from bridge users to reduce their onchain cost, however vanity addresses are generally not advised (see https://blog.1inch.com/a-vulnerability-disclosed-in-profanity-an-ethereum-vanity-address-tool/).

prajwolrg: > This should actually be in the Alpen-CLI, not here.

I think it makes sense to be here. Having the conversion from SubjectId to Address and vice versa should in the same place, allows us to test the roundtrip. I think Alpen-CLI should instead have a dependency on where this is implemented and use that.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
prajwolrg: Since it depends upon the `strata-identifiers`, which is in alpen repo, I don't want to move it to `strata-common` yet.

I did implemented it based on the [deposit descriptor spec](https://www.notion.so/SPS-deposit-desc-295901ba000f80fd8175c7d0a49ef43e) in [8c67e01](https://github.com/alpenlabs/alpen/pull/1229/commits/8c67e01b8431b1b709a6c7275cb4d341ad38089e), but I see an issue:

Since we don't encode the length of the subject bytes, we have to do the hack as referenced in https://github.com/alpenlabs/alpen/pull/1229#discussion_r2648385613. This also means that when used with Codec, the `DepositDescriptor` can only ever be last field of wherever it is used. This is brittle.

I wonder if we should encode the length of the subject bytes in control block instead of serial MSB. We can do the following format:

bit | (value) | meaning
-- | -- | --
7 | 128 | reserved future extension bit
6 | 64 |  serial length bit 0
5 | 32 |  serial length bit 1
4 | 16 |  subject bytes length bit 0
3 | 8 | subject bytes length bit 1
2 | 4 | subject bytes length bit 2
1 | 2 | subject bytes length bit 3
0 | 1 | subject bytes length bit 4

Since `SUBJ_ID_LEN=32` 5 bits can be used to encode it. 

Edit: This wouldn't work. 5 bits can only represent from 0..=31 or 1..=32. 
We can use 6 bits for subject bytes length if we take away reserved future extension bit.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
prajwolrg: Fixed this as part of [8c67e01](https://github.com/alpenlabs/alpen/pull/1229/commits/8c67e01b8431b1b709a6c7275cb4d341ad38089e), but I do wonder if we should pass the `recovery_pk` to the OL. This can be useful to fix the issue as described [here](https://alpenlabs.atlassian.net/browse/STR-2021). We pass the recovery_pk (or make bosd from this) as part of the Log. If it can't be minted, we then create a new withdrawal request. Similar approach can be used to solve the issue of not being able to mint on EE.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
prajwolrg: We use it in a some places in the bridge subprotocol tests like [here](https://github.com/alpenlabs/alpen/blob/main/crates/asm/subprotocols/bridge-v1/src/validation/deposit.rs#L77).

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
prajwolrg: Outdated after [8c67e01](https://github.com/alpenlabs/alpen/pull/1229/commits/8c67e01b8431b1b709a6c7275cb4d341ad38089e).

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
prajwolrg: Outdated after [8c67e01](https://github.com/alpenlabs/alpen/pull/1229/commits/8c67e01b8431b1b709a6c7275cb4d341ad38089e).

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
```diff
@@ -0,0 +1,355 @@
+//! Deposit descriptor wire format used by bridge deposits.
+//!
+//! A descriptor says "route this deposit to account serial X and subject bytes Y". The encoding is
+//! compact: it avoids zero padding in the subject and uses fewer bytes for small serial values.
+//!
+//! Layout (all values are big-endian):
+//! `[control:1][serial:1..=3][subject:0..=32]`
```
barakshani: >[control:1]

Do I understand correctly that this is _control + serial_, 4 bits each? Some of the comments below are based on this understanding.

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
```diff
@@ -0,0 +1,355 @@
...
+//!
+//! Control byte (bit 7 is MSB):
+//! - bits 7..6: reserved, must be 0
+//! - bits 5..4: serial length (00=1 byte, 01=2 bytes, 10=3 bytes, 11 reserved)
```
barakshani: see the next comment: it could be that we don't need any of these, because serial fit in 4 bits (this is what I'd expect for quite some time) -- seems to me we'd have to use `00` here just to have a full-zero byte.

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
```diff
@@ -0,0 +1,355 @@
...
+//! - bits 3..0: the 4 bits immediately above the encoded serial bytes
```
barakshani: above?
these bits are already used in the serial encoding, right?

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
```diff
@@ -0,0 +1,355 @@
...
+//!
+//! Serial encoding:
+//! - A serial is a `u32`, but only 28 bits are encodable (max `0x0FFF_FFFF`).
+//! - Pick the smallest length that fits: `<= 0x0FFF` -> 1 byte, `<= 0x0FFFFF` -> 2 bytes, `<=
```
barakshani: >`<= 0x0FFF` -> 1 byte

as stated above, iiuc we are always forced to use 12 bits -- that's quite a lot don't you think?

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
```diff
@@ -0,0 +1,355 @@
...
+//!   0x0FFF_FFFF` -> 3 bytes.
+//! - Write the least-significant `serial_len` bytes to the buffer; store the next 4 bits in
+//!   `control[3..0]`.
+//! - To decode, rebuild a 4-byte big-endian buffer, insert the serial bytes at the end, then OR in
+//!   the nibble from `control[3..0]`.
+//!
+//! Subject encoding:
+//! - The subject has no explicit length; it is "the rest of the bytes" after the serial.
+//! - Length must be `<= 32`.
+//! - When expanded to a 32-byte `SubjectId`, the bytes are left-padded with zeros.
```
barakshani: >the bytes are left-padded with zeros.

right padded?
**"with zeros"**: this, in principle, should be subject to the serial account formatting.

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
```diff
@@ -0,0 +1,355 @@
...
+//!
+//! Because the subject length is implicit, callers must pass the exact descriptor slice (the
+//! surrounding container must be length-delimited).
+
```
barakshani: I feel this encoding scheme is not optimised enough. Wouldn't the scheme here would leave a smaller onchain footprint, because it's on bit level rather than byte level: https://alpenlabs.atlassian.net/browse/STR-1763?focusedCommentId=22802?

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
```diff
@@ -0,0 +1,355 @@
...
+//! - bits 5..4: serial length (00=1 byte, 01=2 bytes, 10=3 bytes, 11 reserved)
+//! - bits 3..0: the 4 bits immediately above the encoded serial bytes
+//!
+//! Serial encoding:
+//! - A serial is a `u32`, but only 28 bits are encodable (max `0x0FFF_FFFF`).
+//! - Pick the smallest length that fits: `<= 0x0FFF` -> 1 byte, `<= 0x0FFFFF` -> 2 bytes, `<=
+//!   0x0FFF_FFFF` -> 3 bytes.
+//! - Write the least-significant `serial_len` bytes to the buffer; store the next 4 bits in
+//!   `control[3..0]`.
+//! - To decode, rebuild a 4-byte big-endian buffer, insert the serial bytes at the end, then OR in
```
barakshani: then OR in?

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
```diff
@@ -0,0 +1,355 @@
...
+//!   the nibble from `control[3..0]`.
+//!
+//! Subject encoding:
+//! - The subject has no explicit length; it is "the rest of the bytes" after the serial.
+//! - Length must be `<= 32`.
```
barakshani: why do we need this restriction actually? if we need to expend, we can just take the next power of 2, no?

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
prajwolrg: Not really. See https://www.notion.so/SPS-deposit-desc-295901ba000f80fd8175c7d0a49ef43e?source=copy_link#295901ba000f807c9bf3db8884d70e75. As I understand it, control is 1 byte. But we use 4 bits of the control byte for serial. Since the serial don't fit in 4 bits we need more information.

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
prajwolrg: serial doesn't fit it 4 bits. It's 4 bytes. But we are trying to optimize it. See https://www.notion.so/SPS-deposit-desc-295901ba000f80fd8175c7d0a49ef43e?source=copy_link#295901ba000f8072b0c6c2ed990c5dc0.

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
prajwolrg: See https://www.notion.so/SPS-deposit-desc-295901ba000f80fd8175c7d0a49ef43e?source=copy_link#295901ba000f807c9bf3db8884d70e75.

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
prajwolrg: I'm not so sure. I feel it's plenty for now. But if we ever go the route where a more general accounting is supported (apart from the SnarkAccount type), similar to Ethereum accounts, it might not be enough. The way OL seems to be implemented now seems to be with intention to move forward in that direction.

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
prajwolrg: No the [spec](https://www.notion.so/SPS-deposit-desc-295901ba000f80fd8175c7d0a49ef43e) said it is left padded. I also went with left padded as the ethereum address in left padded when seen as bytes32.

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
prajwolrg: I'm not sure if I fully follow. I think this is as optimized as it can be. I even feel if this more-optimized than it needs to be.

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
prajwolrg: It's how the subject is defined [here](https://github.com/alpenlabs/alpen/blob/main/crates/identifiers/src/acct.rs#L9).

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
prajwolrg: OR here meant the bitwise OR. OR with the nibble..

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
barakshani: Isn't that what I was saying? The first byte is divided into 4 bits _control_ and 4 bits _serial_ (the serial can use more bytes)?

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
barakshani: I'm saying that if a given account serial is anything in 0-15, we only need 4 bits for it. But it seems to me we'd use 12 bits. Isn't that the case?
(there would be 4 more bite for control, summing into 2 bytes)

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
barakshani: the sentence seems incomplete

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
barakshani: It's too crude as we operate in byte level, while we could operate in bit level.

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
prajwolrg: It's on line 21.

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
prajwolrg: The first 128 serial numbers are reserved right now. So we need to start with at least 8 bits.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
barakshani: 1..=32 should be fine. We will only be missing subject bytes length 0. We can exclude this option -- subject byte = 0 -- from the system, which generally seems reasonable (all EE's will have to comply).

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
```diff
@@ -0,0 +1,355 @@
...
+        }
+
+        let control = bytes[0];
+        if control & CONTROL_RESERVED_MASK != 0 {
+            return Err(DepositDescriptorError::ReservedControlBits(control));
+        }
+
+        let len_bits =
+            (((control & CONTROL_LEN_BIT0) >> 5) << 1) | ((control & CONTROL_LEN_BIT1) >> 4);
+        if len_bits == 3 {
```
barakshani:
```suggestion
        // checking if the len bits are `11`, which is `2^0 + 2^1 = 3`
        if len_bits == 3 {

File: crates/asm/txs/bridge-v1/src/deposit_request/parse.rs (hunk 5 above)
Comment (storopoli) [reply crates/asm/txs/bridge-v1/src/deposit_request/parse.rs]: 🤦🏻

File: crates/evmexec/src/engine.rs [hunk 33]
```diff
@@ -31,9 +32,22 @@ use crate::{
...
+
+fn subject_to_address(subject: &SubjectId) -> Option<Address> {
+    let bytes = subject.inner();
+    if bytes[EVM_ADDR_LEN..].iter().any(|&byte| byte != 0) {
+        return None;
+    }
+    let mut address_bytes = [0u8; EVM_ADDR_LEN];
+    address_bytes.copy_from_slice(&bytes[..EVM_ADDR_LEN]);
+    Some(Address::from(address_bytes))
+}
```
storopoli: `SubjectIdBytes::to_subject_id()` uses left-padding (zeros at start, data at end):

https://github.com/alpenlabs/alpen/blob/c97482ca2a88c6f3fc78af569b826e7fd93a4515/crates/asm/txs/bridge-v1/src/deposit_request/descriptor/subject_id_bytes.rs#L41-L57

But `subject_to_address()` expects right-padding: `address_bytes.copy_from_slice(&bytes[..EVM_ADDR_LEN]);  // Reads from START`

This mismatch means a 20-byte EVM address converted through `SubjectIdBytes -> SubjectId -> address` will extract the wrong bytes (likely zeros instead of the actual address). Deposits could be misdirected.

crates/evmexec/src/engine.rs
storopoli: We have the same function here: https://github.com/alpenlabs/alpen/blob/c97482ca2a88c6f3fc78af569b826e7fd93a4515/crates/proof-impl/evm-ee-stf/src/utils.rs#L59-L63

crates/evmexec/src/engine.rs
storopoli, line 42: We have the same function here: https://github.com/alpenlabs/alpen/blob/c97482ca2a88c6f3fc78af569b826e7fd93a4515/crates/proof-impl/evm-ee-stf/src/utils.rs#L52-L55

crates/evmexec/src/engine.rs
```diff
@@ -226,12 +239,16 @@ impl<T: EngineRpc> RpcExecEngineInner<T> {
...
             .filter_map(|op| match op {
-                Op::Deposit(deposit_data) => Some(Withdrawal {
-                    index: deposit_data.intent_idx(),
-                    address: address_from_slice(deposit_data.dest_addr())?,
-                    amount: sats_to_gwei(deposit_data.amt())?,
-                    validator_index: 0,
-                }),
+                Op::Deposit(deposit_data) => {
+                    let address = subject_to_address(deposit_data.dest_addr())?;
+                    let amount = sats_to_gwei(deposit_data.amt())?;
```
storopoli: No logging when `subject_to_address` returns `None` or `sats_to_gwei` overflows (this one highly unlikely).

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    fn decode(dec: &mut impl Decoder) -> Result<Self, CodecError> {
+        let mut bytes = Vec::new();
+        while let Ok(byte) = dec.read_arr::<1>() {
+            bytes.push(byte[0]);
+        }
+        Self::decode_from_slice(&bytes)
+            .map_err(|_| CodecError::MalformedField("deposit descriptor"))
+    }
```
storopoli: This greedy decoding will consume all remaining bytes - fragile if `DepositDescriptor` is ever not the last field in a struct.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
prajwolrg: 1..=32 is enough for now.

But this does mean that if we have OL account that doesn't have any `SubjectId`, we would have to add an extra dummy byte and set length to 1. We can interpret the reserved bit to somehow handle this properly, or just use the full 6 bits for encoding length.

barakshani: Just fixed a typo in a previous message, but was saying there that EE will have to comply, that is, they cannot have no SubjectId.

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
delbonis: I agree we should not be using Codec for this case, it's not what it's designed for.

crates/asm/txs/bridge-v1/src/deposit_request/aux.rs
delbonis: I'm not sure what to do here, it's not really what it was designed for.

crates/asm/txs/bridge-v1/src/deposit_request/descriptor/format.rs
prajwolrg: I've simplified the implementation in [d91bb3f](https://github.com/alpenlabs/alpen/pull/1229/commits/d91bb3f808dbdda23ae3900bd436ab19c8102456). I'm not sure if adding test vectors will be more helpful.

crates/evmexec/src/engine.rs
prajwolrg: Moved to `alpen-reth-evm` in [d8b5b2b](https://github.com/alpenlabs/alpen/pull/1229/commits/d8b5b2b3f224ddf6ee31a1c0448632b08f0ef0f8).

crates/evmexec/src/engine.rs
prajwolrg: Nice catch, I've fixed in [d8b5b2b](https://github.com/alpenlabs/alpen/pull/1229/commits/d8b5b2b3f224ddf6ee31a1c0448632b08f0ef0f8) by having a single source of truth for the conversion. Also added proptest to test it properly.

crates/evmexec/src/engine.rs
prajwolrg: Resolved this in [d8b5b2b](https://github.com/alpenlabs/alpen/pull/1229/commits/d8b5b2b3f224ddf6ee31a1c0448632b08f0ef0f8).

crates/evmexec/src/engine.rs
prajwolrg: Addressed in [dced3a0](https://github.com/alpenlabs/alpen/pull/1229/commits/dced3a0f23c863c1490395be3beaf7f3f0cbe480). I've created a separate ticket to handle this properly [here](https://alpenlabs.atlassian.net/browse/STR-2125).

crates/identifiers/src/deposit.rs
```diff
@@ -0,0 +1,424 @@
...
+        self.0 & Self::MSB_NIBBLE_MASK
+    }
+
+    fn set_serial_len(&mut self, len: u8) {
+        self.0 |= len << 4
+    }
+
+    fn set_serial_msb(&mut self, msb: u8) {
+        self.0 |= msb
+    }
```
delbonis, line 71: Maybe these should panic if the bits passed don't comply with the masks?

crates/identifiers/src/deposit.rs
```diff
@@ -0,0 +1,424 @@
...
+/// a variable-length encoding of an account serial number (12-28 bits).
+///
+/// This representation makes the encoding logic clearer by working directly
+/// with the big-endian byte representation.
+struct EncodedSerial {
+    /// Control byte containing length encoding and MSB nibble.
+    control: ControlByte,
+
+    /// Additional 1-3 bytes containing the remaining serial bits.
+    additional_bytes: Vec<u8>,
```
delbonis: We could probably make this a `[u8; 3]` tbh, returning a slice or even the decoded u32 based on the value of the bitfield in `control`.

crates/identifiers/src/deposit.rs
prajwolrg: Added a `debug_assert!` in [d75b65b](https://github.com/alpenlabs/alpen/pull/1229/commits/d75b65ba50c02bf0401ca9e1967ba26576f39ed5).

crates/identifiers/src/deposit.rs
prajwolrg: Used a `[u8; 3]` in [d75b65b](https://github.com/alpenlabs/alpen/pull/1229/commits/d75b65ba50c02bf0401ca9e1967ba26576f39ed5).

crates/chaintsn/src/checkin.rs
```diff
@@ -166,8 +167,10 @@ fn process_l1_deposit<'s, S: StateAccessor>(
...
     let amt = BitcoinAmount::from_sat(deposit.amount);
-    let dest_ident = deposit.addr.to_vec();
+    let descriptor = DepositDescriptor::decode_from_slice(&deposit.destination)?;
+    let ee_id = descriptor.dest_acct_serial();
+    let dest_ident = descriptor.dest_subject().to_subject_id();
     let intent = DepositIntent::new(amt, dest_ident);
-    state.insert_deposit_intent(0, intent);
+    state.insert_deposit_intent(*ee_id, intent);
     Ok(())
 }
```
storopoli, line 176: `process_l1_deposit` decodes `DepositDescriptor` and passes `ee_id` directly into `insert_deposit_intent`, but both `FauxStateCache` and `StateCache` still `assert_eq!(ee_id, AccountSerial::zero())`, which panics on non-zero values.

https://github.com/alpenlabs/alpen/blob/353f6b1e01ab540db3f569c4710f863d26c7aa84/crates/chaintsn/src/legacy.rs#L40-L51

A malicious DRT can embed a non-zero serial (ASM does not validate descriptor semantics), causing a panic when processing L1 logs.

crates/alpen-ee/block-assembly/src/payload.rs
```diff
@@ -1,33 +1,37 @@
...
+                        // FIXME: https://alpenlabs.atlassian.net/browse/STR-2125
+                        error!(
+                            subject_id = %data.dest(),
+                            "Skipping deposit with invalid destination address (non-zero padding in first 12 bytes)"
+                        );
+                        None
+                    }
+                }
             }
         })
```
storopoli, line 23: `extract_deposits` now `filter_maps` (skips) invalid destinations but `processed_inputs` is still set to deposits.len().

https://github.com/alpenlabs/alpen/blob/353f6b1e01ab540db3f569c4710f863d26c7aa84/crates/alpen-ee/block-assembly/src/payload.rs#L41-L54

In EE update verification, `processed_inputs` must match the number of inputs actually consumed/removed from the pending queue.

https://github.com/alpenlabs/alpen/blob/353f6b1e01ab540db3f569c4710f863d26c7aa84/crates/ee-acct-runtime/src/update_processing.rs#L232-L236

Skipping entries can cause:
* Invalid updates (`EnvError::InvalidBlock`) due to mismatch
* Or incorrect queue advancement (dropping the wrong pending inputs)

crates/chaintsn/src/checkin.rs
prajwolrg: This is a different issue that needs to be handled appropriately with https://alpenlabs.atlassian.net/browse/STR-2021. Since this is legacy codebase, I don't want to make a lot of changes here.

crates/alpen-ee/block-assembly/src/payload.rs
prajwolrg: I created a new function in [8922682](https://github.com/alpenlabs/alpen/pull/1229/commits/89226822ccdc363a2d4d97b30c9d0d5ccd5f224c) and used that in [db1aaee](https://github.com/alpenlabs/alpen/pull/1229/commits/db1aaeee588ef8616c75284429419d2904c86b5a) to avoid this scenario.

crates/chaintsn/src/checkin.rs
storopoli: Yeah I've imagined so.

storopoli (approved):
ACK 2f709c8eab0b227168a715c990e45710fb6d96b7

PR #1231, opened 2025-12-29
ee batch builder

sapinb:
## Description
This PR adds batch builder and update submitter for alpen ee. 

AI usage: test boilerplate and comments.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
Depends on the following traits that have not yet been implemented. Integration is deferred until these are ready:
1. BatchStorage
2. BatchProver

Additionally, current update_submitter lacks access to OL mempool inspection, so it simply resends everything that is not already in OL chain. Once OL mempool can be queried, this should be updated. 

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15092,8 +15092,8 @@ version = "0.1.0"
 dependencies = [
  "hex",
  "serde",
- "ssz",
  "serde_json",
+ "ssz",
  "strata-acct-types",
  "strata-identifiers",
  "strata-ol-chain-types-new",
--- a/bin/strata/src/rpc.rs
+++ b/bin/strata/src/rpc.rs
@@ -17,7 +17,8 @@ use strata_ledger_types::{
 use strata_ol_mempool::{MempoolHandle, OLMempoolError, OLMempoolTransaction};
 use strata_rpc_api_new::OLClientRpcServer;
 use strata_rpc_types_new::{
-    RpcAccountBlockSummary, RpcAccountEpochSummary, RpcOLChainStatus, RpcOLTransaction,
+    OLBlockOrTag, RpcAccountBlockSummary, RpcAccountEpochSummary, RpcOLChainStatus,
+    RpcOLTransaction, RpcSnarkAccountState,
 };
 use strata_snark_acct_types::ProofState;
 use strata_status::StatusChannel;
@@ -328,6 +329,111 @@ impl OLClientRpcServer for OLRpcServer {
             .map_err(map_mempool_error_to_rpc)?;
 
         Ok(txid)
+    }
+
+    async fn get_snark_account_state(
+        &self,
+        account_id: AccountId,
+        block_or_tag: OLBlockOrTag,
+    ) -> RpcResult<Option<RpcSnarkAccountState>> {
+        // Resolve block_or_tag to a block commitment
+        let block_commitment = match block_or_tag {
+            OLBlockOrTag::Latest => {
+                let chain_sync_status = self
+                    .status_channel
+                    .get_chain_sync_status()
+                    .ok_or_else(|| internal_error("Chain sync status not available"))?;
+                chain_sync_status.tip
+            }
+            OLBlockOrTag::Confirmed => {
+                let chain_sync_status = self
+                    .status_channel
+                    .get_chain_sync_status()
+                    .ok_or_else(|| internal_error("Chain sync status not available"))?;
+                chain_sync_status.prev_epoch.to_block_commitment()
+            }
+            OLBlockOrTag::Finalized => {
+                let chain_sync_status = self
+                    .status_channel
+                    .get_chain_sync_status()
+                    .ok_or_else(|| internal_error("Chain sync status not available"))?;
+                chain_sync_status.finalized_epoch.to_block_commitment()
+            }
+            OLBlockOrTag::OLBlockId(block_id) => {
+                let block = self
+                    .storage
+                    .ol_block()
+                    .get_block_data_async(block_id)
+                    .await
+                    .map_err(|e| {
+                        error!(?e, %block_id, "Failed to get block data");
+                        db_error(e)
+                    })?
+                    .ok_or_else(|| not_found_error(format!("Block {block_id} not found")))?;
+                OLBlockCommitment::new(block.header().slot(), block_id)
+            }
+            OLBlockOrTag::Slot(slot) => {
+                let block_ids = self
+                    .storage
+                    .ol_block()
+                    .get_blocks_at_height_async(slot)
+                    .await
+                    .map_err(|e| {
+                        error!(?e, slot, "Failed to get blocks at slot");
+                        db_error(e)
+                    })?;
+                let block_id = block_ids
+                    .first()
+                    .copied()
+                    .ok_or_else(|| not_found_error(format!("No block found at slot {slot}")))?;
+                OLBlockCommitment::new(slot, block_id)
+            }
+        };
+
+        // Get OL state at the resolved block
+        let ol_state = self
+            .storage
+            .ol_state()
+            .get_toplevel_ol_state_async(block_commitment)
+            .await
+            .map_err(|e| {
+                error!(?e, %block_commitment, "Failed to get OL state");
+                db_error(e)
+            })?
+            .ok_or_else(|| {
+                not_found_error(format!("No OL state found for block {block_commitment}"))
+            })?;
+
+        // Get account state
+        let account_state = match ol_state.get_account_state(account_id) {
+            Ok(Some(state)) => state,
+            Ok(None) => return Ok(None), // Account doesn't exist
+            Err(e) => {
+                error!(?e, %account_id, "Failed to get account state");
+                return Err(internal_error(format!("Account error: {e}")));
+            }
+        };
+
+        // Try to get snark account state; return None if not a snark account
+        match account_state.as_snark_account() {
+            Ok(snark_state) => {
+                // Manually construct RpcSnarkAccountState from the native state
+                // Note: update_vk is not available from NativeSnarkAccountState (it's stored
+                // as account metadata, not runtime state), so we return an empty vec for now
+                let seq_no: u64 = *snark_state.seqno().inner();
+                let inner_state = snark_state.inner_state_root().0.into();
+                let next_inbox_msg_idx = snark_state.get_next_inbox_msg_idx();
+                let update_vk = vec![].into(); // Not available from native state
+
+                Ok(Some(RpcSnarkAccountState::new(
+                    seq_no,
+                    inner_state,
+                    next_inbox_msg_idx,
+                    update_vk,
+                )))
+            }
+            Err(_) => Ok(None), // Not a snark account
+        }
     }
 }
 
--- a/crates/rpc-new/types/Cargo.toml
+++ b/crates/rpc-new/types/Cargo.toml
@@ -18,3 +18,6 @@ hex.workspace = true
 serde.workspace = true
 ssz.workspace = true
 thiserror.workspace = true
+
+[dev-dependencies]
+serde_json.workspace = true


crates/alpen-ee/common/src/types/exec_record.rs
```diff
@@ -30,13 +31,16 @@ struct ExecPackageMetadata {
 pub struct ExecBlockRecord {
     /// Additional metadata associated with this block.
     metadata: ExecPackageMetadata,
+    /// OL Account messages processed in this block.
+    messages: Vec<MessageEntry>,
```
prajwolrg, line 37: I wonder if this can be part of `ExecBlockPackage::inputs`.

crates/alpen-ee/sequencer/src/batch_builder/traits.rs
prajwolrg, line 1: Wow! This is so simple and nice.

crates/alpen-ee/block-assembly/src/block.rs
```diff
@@ -12,12 +12,12 @@ use crate::{package::build_block_package, payload::build_exec_payload};
...
 /// All inputs that control the next built block.
 #[derive(Debug)]
-pub struct BlockAssemblyInputs {
+pub struct BlockAssemblyInputs<'a> {
     /// EeAccountState of last block.
     pub account_state: EeAccountState,
     /// New inbox messages to be included in this block.
     /// Can be empty.
-    pub inbox_messages: Vec<MessageEntry>,
+    pub inbox_messages: &'a [MessageEntry],
```
storopoli, line 20: Yes! Less allocs!

crates/alpen-ee/sequencer/src/batch_builder/accumulator.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    /// Drain blocks for batch creation.
+    ///
+    /// Returns `(inner_blocks, last_block)` where `inner_blocks` excludes `last_block`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the accumulator is empty.
+    pub fn drain_for_batch(&mut self) -> (Vec<Hash>, Hash) {
```
storopoli: I generally don't like returning tuples that you need to explain what the tuple is in the docstring.
Maybe we can do this with a custom type?

crates/alpen-ee/sequencer/src/batch_builder/ctx.rs
```diff
@@ -0,0 +1,42 @@
...
+    /// Storage for exec blocks.
+    pub block_storage: Arc<ES>,
+    /// Storage for batches.
+    pub batch_storage: Arc<BS>,
+    /// Handle to query canonical chain status.
+    pub exec_chain: Arc<ExecChainHandle>,
+    /// Sender to notify about latest batch updates (new batch sealed or reorg).
+    pub latest_batch_tx: watch::Sender<Option<BatchId>>,
+    /// Marker for the policy type.
+    pub _policy: std::marker::PhantomData<P>,
```
storopoli: Can you import this the same way you imported `std::sync::Arc` above?

crates/alpen-ee/sequencer/src/batch_builder/handle.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /// Returns the current latest batch ID, if any.
+    pub fn latest_batch_id(&self) -> Option<BatchId> {
+        *self.latest_batch_rx.borrow()
+    }
+}
+
+/// Create batch builder task.
+#[expect(clippy::too_many_arguments, reason = "todo builder")]
```
storopoli: I'm fine with keeping this lint but please update the reason?

crates/alpen-ee/sequencer/src/batch_builder/handle.rs
```diff
@@ -0,0 +1,77 @@
...
+    let ctx = BatchBuilderCtx {
+        genesis_hash,
+        preconf_rx,
+        block_data_provider,
+        sealing_policy,
+        block_storage,
+        batch_storage,
+        exec_chain,
+        latest_batch_tx,
+        _policy: std::marker::PhantomData,
```
storopoli: import above same way you did with `use std::{future::Future, sync::Arc};`

crates/alpen-ee/common/src/traits/ol_client.rs
```diff
@@ -45,6 +45,12 @@ pub struct OLBlockData {
     pub next_inbox_msg_idx: u64,
 }
 
+#[derive(Debug, Clone)]
+pub struct OLAccountState {
+    pub seq_no: Seqno,
+    pub proof_state: ProofState,
+}
+
```
storopoli, line 56: Can't we use an `ol-foo` type?
It is strange that we have an `OLAccountState` defined in an `alpen-ee-common` crate...

crates/alpen-ee/database/src/serialization_types/exec_block.rs
```diff
@@ -58,6 +62,39 @@ impl TryFrom<DBExecBlockRecord> for ExecBlockRecord {
...
             value.parent_blockhash,
             value.next_inbox_msg_idx,
+            value.messages.into_iter().map(Into::into).collect(),
         ))
     }
 }
+
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize, PartialEq)]
+struct DBMessageEntry {
+    source: [u8; 32],
```
storopoli, line 72: Should this be a `Hash`?

crates/alpen-ee/exec-chain/src/task.rs
```diff
@@ -28,34 +28,71 @@ pub(crate) enum ChainTrackerError {
...
+        },
+    )
 }
 
 /// Main task loop for the execution chain tracker.
 ///
 /// Processes incoming messages to update chain state, handle new blocks, and respond to queries.
 /// The task exits if the preconf head channel is closed, as this is considered a fatal error.
+///
+/// Message priority (highest to lowest): NewBlock -> OLConsensusUpdate -> Query
```
storopoli:
```suggestion
/// Message priority (highest to lowest): `NewBlock` -> `OLConsensusUpdate` -> `Query`
```

crates/alpen-ee/sequencer/src/batch_builder/state.rs
```diff
@@ -0,0 +1,179 @@
...
+use eyre::Result;
+use strata_acct_types::Hash;
+
+use super::{Accumulator, BatchPolicy};
+
+/// State for the batch builder task.
+///
+/// This tracks the current position in the chain and the accumulator for
+/// the pending batch. This state is not persisted; it is rebuilt from
+/// `BatchStorage` on restart.
```
storopoli:
```suggestion
/// [`BatchStorage`] on restart.
```

crates/alpen-ee/sequencer/src/batch_builder/task.rs
```diff
@@ -0,0 +1,311 @@
...
+                }
+                let new_tip = *ctx.preconf_rx.borrow_and_update();
+                handle_new_tip(&mut state, &ctx, new_tip).await
+            }
+
+            // Branch 2: Process pending blocks when first block's data is ready
+            Some(_) = check_first_pending_block_data(&state, ctx.block_data_provider.as_ref()) => {
+                process_pending_blocks(&mut state, &ctx).await
+            }
+        };
```
storopoli, line 142: Do we need for this `select!` to be also `biased;`?

crates/alpen-ee/sequencer/src/batch_builder/traits.rs
```diff
@@ -0,0 +1,54 @@
...
+/// The `BatchPolicy` trait allows different batching strategies to be implemented
+/// by specifying the types for block data and accumulated values, along with
+/// the accumulation logic.
+pub trait BatchPolicy: Send + Sync + 'static {
+    /// Data collected per block, used for sealing decisions.
+    /// This is provided by [`BlockDataProvider`](super::BlockDataProvider).
+    type BlockData: Send + Sync + Clone;
+
+    /// Accumulated value across blocks (e.g., count, DA size).
+    /// Must implement `Default` for initialization and reset.
```
storopoli:
```suggestion
    /// Must implement [`Default`] for initialization and reset.
```

crates/alpen-ee/sequencer/src/update_submitter/task.rs
```diff
@@ -0,0 +1,154 @@
...
+use crate::update_submitter::update_builder::build_update_from_batch;
+
+/// Maximum number of entries in the update cache.
+const UPDATE_CACHE_MAX_SIZE: usize = 64;
+
+/// Cache for built updates, keyed by BatchId.
+/// Stores (batch_idx, update) to allow eviction based on sequence number.
+struct UpdateCache {
+    entries: HashMap<BatchId, (u64, SnarkAccountUpdate)>,
+}
```
storopoli, line 23: More tuples that we need to explain in docstrings.
Can't we have a type for `(u64, SnarkAccountUpdate)`?

crates/alpen-ee/sequencer/src/update_submitter/task.rs
```diff
@@ -0,0 +1,154 @@
...
+            }
+            // Branch 2: OL chain status update
+            changed = ol_status_rx.changed() => {
+                if changed.is_err() {
+                    warn!("ol_status_rx closed; exiting");
+                    return;
+                }
+                process_ready_batches(ol_client.as_ref(), batch_storage.as_ref(), exec_storage.as_ref(), prover.as_ref(), &mut update_cache).await
+            }
+        };
```
storopoli, line 93: Should this `select!` be `biased;`?

crates/alpen-ee/sequencer/src/update_submitter/update_builder.rs
```diff
@@ -0,0 +1,107 @@
...
+use strata_acct_types::Hash;
+use strata_codec::encode_to_vec;
+use strata_ee_acct_types::UpdateExtraData;
+use strata_snark_acct_types::{
+    LedgerRefs, OutputMessage, OutputTransfer, ProofState, SnarkAccountUpdate, UpdateOperationData,
+    UpdateOutputs,
+};
+use tree_hash::{Sha256Hasher, TreeHash};
+
+/// Build a SnarkAccountUpdate from a batch in ProofReady state.
```
storopoli:
```suggestion
/// Build a [`SnarkAccountUpdate`] from a batch in `ProofReady` state.
```

crates/alpen-ee/sequencer/src/update_submitter/update_builder.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    // Should we re-check that proof is valid ?
+
+    Ok(SnarkAccountUpdate::new(
+        update_operation,
+        update_proof.to_vec(),
+    ))
+}
+
+/// Build a [`UpdateOperationData`] from data in a batch.
```
storopoli:
```suggestion
/// Build an [`UpdateOperationData`] from data in a batch.
```

crates/alpen-ee/sequencer/src/update_submitter/task.rs
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
+    /// Get a cached update by BatchId.
+    fn get(&self, batch_id: &BatchId) -> Option<&SnarkAccountUpdate> {
+        self.entries.get(batch_id).map(|(_, update)| update)
+    }
+
+    /// Insert an update into the cache if there is room.
+    /// If the cache is at max capacity, the entry is not inserted.
```
bewakes, line 38: Shouldn't this error out? Because the caller can't know if the update made to cache or not. At least a boolean indicating success or failure.

crates/alpen-ee/common/src/traits/ol_client.rs
sapinb: This is the view of OLAccountState with the fields needed by EE. There isnt really a concise type for this on OL afaik (OL has this hidden behind multiple state accessor traits) and I didnt want to use the rpc types here. Maybe a different name ? OLAccountStateView ? suggestions please

crates/alpen-ee/database/src/serialization_types/exec_block.rs
sapinb: Its the parallel of AccountId, which doesn't use Hash, but wraps [u8; 32] directly.

crates/alpen-ee/sequencer/src/update_submitter/task.rs
sapinb: Thinking up new names for these small types is hard 🥲

crates/alpen-ee/sequencer/src/batch_builder/traits.rs
```diff
@@ -0,0 +1,54 @@
...
+pub trait BatchSealingPolicy<P: BatchPolicy>: Send + Sync {
+    /// Check if adding a block would exceed the batch threshold.
+    ///
+    /// If this returns `true`, the current batch should be sealed before
+    /// adding this block. The block will then become the first block of
+    /// the next batch.
+    ///
+    /// This is called with the current accumulator state and the data for
+    /// the block about to be added.
+    fn would_exceed(&self, accumulator: &Accumulator<P>, block_data: &P::BlockData) -> bool;
```
bewakes, line 43: Can a more generic name for this be `should_seal`? `would_exceed` as a trait method sounds too restricted for some policy that's only based on a threshold.

crates/alpen-ee/sequencer/src/batch_builder/task.rs
```diff
@@ -0,0 +1,311 @@
...
+                if changed.is_err() {
+                    warn!("preconf_rx channel closed; exiting");
+                    return;
+                }
+                let new_tip = *ctx.preconf_rx.borrow_and_update();
+                handle_new_tip(&mut state, &ctx, new_tip).await
+            }
+
+            // Branch 2: Process pending blocks when first block's data is ready
+            Some(_) = check_first_pending_block_data(&state, ctx.block_data_provider.as_ref()) => {
```
bewakes: Won't this run in some sort of a "hot loop"?

crates/alpen-ee/sequencer/src/batch_builder/task.rs
```diff
@@ -0,0 +1,311 @@
...
+        idx -= 1;
+    }
+}
+
+/// Check for reorgs and handle them.
+///
+/// Returns `Some(new_latest_batch_id)` if a deep reorg was detected and batches were reverted.
+/// The inner `Option<BatchId>` is `None` if all batches were reverted, `Some(id)` otherwise.
+/// Returns `None` if no deep reorg occurred (including shallow reorgs).
+async fn check_and_handle_reorg<P: BatchPolicy>(
```
bewakes: Reorg logic is always a tricky one. This needs a test.

bewakes (commented):
Great one! My little concern is over missing testcases for reorg handling, which can be tackled on a separate ticket if not feasible here. The rests are questions and nit.

crates/alpen-ee/common/src/types/exec_record.rs
sapinb: There are a lot of things that ExecBlockPackage does not have that are needed by the sequencer, which are saved under ExecBlockRecord.
I am not sure if these sequencer specifics should go inside  ExecBlockPackage. Need to rethink what exactly ExecBlockPackage will be used for.

crates/alpen-ee/common/src/traits/ol_client.rs
storopoli: I like `OLAccountStateView`.
Should we add this type in an `ol-foo` crate?
I think it makes more sense to be there instead of `alpen-ee-common`.

crates/alpen-ee/sequencer/src/update_submitter/task.rs
storopoli: As the wise men say:

> We have 3 issues in computer science:
>
> 1. Naming
> 2. off by 1 errors

crates/alpen-ee/sequencer/src/batch_builder/traits.rs
sapinb: Its `would_exceed` because adding this next block would exceed the accumulator and so it should be sealed without this next block. A `should_seal` does not convey the same meaning.

crates/alpen-ee/sequencer/src/batch_builder/task.rs
sapinb: Yes it will. Thanks for the catch. Fixed.

crates/alpen-ee/sequencer/src/update_submitter/task.rs
sapinb: the cache is mostly an optimization. even if it is not cached, the update can and will be recomputed deterministically if needed next time. 
The slightly weird way the cache works here (not caching instead of replacing an old entry directly) is because of the repeated sequential nature of the update generation. This avoids a total cache failure when the number of pending updates to be created exceeds the cache size if this was a LRU.

crates/alpen-ee/sequencer/src/update_submitter/task.rs
sapinb: doesnt matter in this case. both branches trigger the same code.

crates/alpen-ee/sequencer/src/batch_builder/task.rs
sapinb: reorg logic is updated and tests added.

crates/alpen-ee/sequencer/src/batch_builder/task.rs
sapinb: there isnt much practical difference here.

crates/alpen-ee/sequencer/src/batch_builder/traits.rs
bewakes: I think the meaning should be conveyed from the implementing struct's name and not from the trait method which should be generic enough. The sealing condition might change in the future for example from "size exceeds N" to "every N batches" or something else.

PR #1236, opened 2025-12-31
feat: implement new `chain-worker` based on the new STF

storopoli:
## Description

Implements the new chain worker for the OL <> EE decoupling.
This is a new crate called `chain-worker-new`.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

There is very few TODOs that I don’t know if I leave them as it is or implement it. Please advise.

The commit history is very good and linear. I encourage to review this PR commit by commit.

Done with the aid of Claude Code (spent tons of time in "plan mode" to have EVERY detail nailed before starting the AI-assisted coding)

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-1922

### DIFF
--- a/crates/chain-worker-new/src/context.rs
+++ b/crates/chain-worker-new/src/context.rs
@@ -70,6 +70,32 @@ impl ChainWorkerContext for ChainWorkerContextImpl {
         Ok(block_opt.map(|block| block.header().clone()))
     }
 
+    fn fetch_chain_tip(&self) -> WorkerResult<Option<OLBlockCommitment>> {
+        // Get the highest slot with a block
+        let tip_slot = self.ol_block_mgr.get_tip_slot_blocking()?;
+
+        // Slot 0 with no blocks means no chain yet
+        if tip_slot == 0 {
+            let blocks = self.fetch_blocks_at_slot(0)?;
+            if blocks.is_empty() {
+                return Ok(None);
+            }
+        }
+
+        // Get blocks at the tip slot
+        let block_ids = self.fetch_blocks_at_slot(tip_slot)?;
+
+        // Return the first block at the tip slot
+        // If there are multiple (forks), we just pick one - the caller can
+        // use fork choice logic if needed
+        let blkid = match block_ids.first() {
+            Some(id) => *id,
+            None => return Ok(None),
+        };
+
+        Ok(Some(OLBlockCommitment::new(tip_slot, blkid)))
+    }
+
     fn fetch_ol_state(&self, commitment: OLBlockCommitment) -> WorkerResult<Option<OLState>> {
         let state_opt = self
             .ol_state_mgr
@@ -147,13 +173,34 @@ impl ChainWorkerContext for ChainWorkerContextImpl {
     }
 
     fn merge_finalized_epoch(&self, epoch: &EpochCommitment) -> WorkerResult<()> {
-        // Get the epoch summary to find the block range
         let summary = self.fetch_summary(epoch)?;
-
-        // Get the current finalized state (or start from prev_terminal's state)
+        let terminal = *summary.terminal();
         let prev_terminal = *summary.prev_terminal();
+
+        // Collect canonical chain by walking backwards from terminal via parent pointers.
+        // This ensures we only apply write batches for blocks in the canonical chain,
+        // not fork blocks that may also have write batches stored.
+        let mut chain: Vec<OLBlockCommitment> = Vec::new();
+        let mut current = terminal;
+
+        while current != prev_terminal && !current.is_null() {
+            chain.push(current);
+            // Get header to find parent
+            let header = self
+                .fetch_header(current.blkid())?
+                .ok_or(WorkerError::MissingOLBlock(*current.blkid()))?;
+            let parent_blkid = header.parent_blkid();
+            if parent_blkid.is_null() {
+                break;
+            }
+            current = OLBlockCommitment::new(current.slot().saturating_sub(1), *parent_blkid);
+        }
+
+        // Reverse to get forward order (excluding prev_terminal which is already finalized)
+        chain.reverse();
+
+        // Get base state from prev_terminal (or genesis)
         let mut current_state = if prev_terminal.is_null() {
-            // Genesis case - get genesis state
             self.fetch_ol_state(OLBlockCommitment::null())?
                 .ok_or(WorkerError::MissingPreState(OLBlockCommitment::null()))?
         } else {
@@ -161,29 +208,16 @@ impl ChainWorkerContext for ChainWorkerContextImpl {
                 .ok_or(WorkerError::MissingPreState(prev_terminal))?
         };
 
-        // Walk through all blocks from prev_terminal to terminal and apply write batches
-        let terminal = *summary.terminal();
-        let mut current_slot = prev_terminal.slot();
-
-        // Process blocks slot by slot until we reach the terminal
-        while current_slot < terminal.slot() {
-            current_slot += 1;
-
-            // Get blocks at this slot
-            let block_ids = self.fetch_blocks_at_slot(current_slot)?;
-
-            // For each block, check if it's in our chain and apply its write batch
-            for blkid in block_ids {
-                let commitment = OLBlockCommitment::new(current_slot, blkid);
-
-                // Try to get the write batch - if it exists, this block was executed
-                if let Some(wb) = self.fetch_write_batch(commitment)? {
-                    // Apply the write batch to advance the state
-                    current_state.apply_write_batch(wb).map_err(|e| {
-                        WorkerError::Unexpected(format!("failed to apply batch: {e}"))
-                    })?;
-                }
-            }
+        // Apply write batches in canonical order.
+        // Every block in the canonical chain must have a write batch - a missing one
+        // indicates data corruption or a bug, so we error out rather than skip.
+        for commitment in chain {
+            let wb = self
+                .fetch_write_batch(commitment)?
+                .ok_or(WorkerError::MissingWriteBatch(commitment))?;
+            current_state
+                .apply_write_batch(wb)
+                .map_err(|e| WorkerError::Unexpected(format!("failed to apply batch: {e}")))?;
         }
 
         // Store the final merged state at the terminal commitment
--- a/crates/chain-worker-new/src/state.rs
+++ b/crates/chain-worker-new/src/state.rs
@@ -141,37 +141,34 @@ impl<W: ChainWorkerContext + Send + Sync + 'static> ChainWorkerServiceState<W> {
     }
 
     /// Waits for genesis and resolves the initial tip commitment.
+    ///
+    /// This first checks the database for an existing chain tip (highest executed block).
+    /// If found, it resumes from there. Otherwise, it waits for genesis and starts fresh.
     pub(crate) fn wait_for_genesis_and_resolve_tip(&self) -> WorkerResult<OLBlockCommitment> {
+        // First, check if we have an existing chain tip in the database.
+        // This allows us to resume from where we left off after a restart,
+        // including unfinalized blocks.
+        if let Some(db_tip) = self.deps.context.fetch_chain_tip()? {
+            info!(slot = db_tip.slot(), %db_tip, "resuming from database chain tip");
+            return Ok(db_tip);
+        }
+
+        // No existing chain - wait for genesis
         info!("waiting until genesis");
 
-        let init_state = self
+        let _init_state = self
             .deps
             .runtime_handle
             .block_on(self.deps.status_channel.wait_until_genesis())
             .map_err(|_| WorkerError::ShutdownBeforeGenesis)?;
 
-        // Get the last finalized epoch from the checkpoint, if any
-        let finalized_epoch = init_state
-            .get_last_finalized_checkpoint()
-            .map(|ckpt| ckpt.batch_info.get_epoch_commitment());
+        // Start from genesis block
+        let genesis_block_ids = self.deps.context.fetch_blocks_at_slot(0)?;
+        let genesis_blkid = *genesis_block_ids
+            .first()
+            .ok_or(WorkerError::MissingGenesisBlock)?;
 
-        let cur_tip = match finalized_epoch {
-            Some(epoch) => {
-                // Resume from the last finalized epoch's block
-                let l2bc = epoch.to_block_commitment();
-                OLBlockCommitment::new(l2bc.slot(), *l2bc.blkid())
-            }
-            None => {
-                // No finalized checkpoint yet - start from genesis
-                let genesis_block_ids = self.deps.context.fetch_blocks_at_slot(0)?;
-                let genesis_blkid = *genesis_block_ids
-                    .first()
-                    .ok_or(WorkerError::MissingGenesisBlock)?;
-                OLBlockCommitment::new(0, genesis_blkid)
-            }
-        };
-
-        Ok(cur_tip)
+        Ok(OLBlockCommitment::new(0, genesis_blkid))
     }
 
     /// Initializes the worker with the given tip commitment.
@@ -272,13 +269,12 @@ impl<W: ChainWorkerContext + Send + Sync + 'static> ChainWorkerServiceState<W> {
         let (write_batch, indexer_writes) =
             Self::run_stf_verification(&parent_state, block, parent_header)?;
 
-        // Use the state root from the header (verify_block validated it)
+        // Use the state root from the header (verify_block validated it).
+        // Note: logs are validated internally by verify_block via the logs_root commitment.
         let computed_state_root = *block.header().state_root();
-        let logs = Vec::new(); // TODO: Collect logs from execution context when available
 
         Ok(OLBlockExecutionOutput::new(
             computed_state_root,
-            logs,
             write_batch,
             indexer_writes,
         ))
--- a/crates/chain-worker-new/src/traits.rs
+++ b/crates/chain-worker-new/src/traits.rs
@@ -28,6 +28,13 @@ pub trait ChainWorkerContext: Send + Sync + 'static {
     /// Fetches a block's header by its ID.
     fn fetch_header(&self, blkid: &OLBlockId) -> WorkerResult<Option<OLBlockHeader>>;
 
+    /// Fetches the current chain tip from the database.
+    ///
+    /// Returns the highest slot block that has been stored. If there are multiple
+    /// blocks at the tip slot (forks), returns one of them.
+    /// Returns `None` if no blocks have been stored yet.
+    fn fetch_chain_tip(&self) -> WorkerResult<Option<OLBlockCommitment>>;
+
     // =========================================================================
     // State access
     // =========================================================================
--- a/crates/chain-worker-new/src/output.rs
+++ b/crates/chain-worker-new/src/output.rs
@@ -1,7 +1,6 @@
 //! Output types for block execution.
 
 use strata_identifiers::Buf32;
-use strata_ol_chain_types_new::OLLog;
 use strata_ol_state_support_types::IndexerWrites;
 use strata_ol_state_types::{NativeAccountState, WriteBatch};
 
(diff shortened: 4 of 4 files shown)


crates/chain-worker-new/src/service.rs
```diff
@@ -0,0 +1,352 @@
...
+    /// Shared state for the worker.
+    #[expect(unused, reason = "will be used later for shared state access")]
+    shared: Arc<Mutex<WorkerShared>>,
+
+    /// Parameters for the chain.
+    #[expect(unused, reason = "params will be used for chain configuration")]
+    params: Arc<Params>,
+
+    /// Context for the worker.
+    context: W,
```
bewakes: Can we move static things that don't actually contribute to the state like `context`, `runtime_handle`, etc. to the service struct and just keep in this state only the things that actually consist of a state?
Basically, my idea of state here is something that is in memory which we can write out to db or elsewhere via the context or any other handle.
Right now, the state does everything like in an OOP style. Most of the out-of-memory actions should be done by the service struct itself imo.

crates/chain-worker-new/src/service.rs
```diff
@@ -0,0 +1,352 @@
...
+        Ok(Response::Continue)
+    }
+}
+
+/// Service state for the chain worker.
+#[expect(
+    missing_debug_implementations,
+    reason = "Some inner types don't have Debug impl"
+)]
+pub struct ChainWorkerServiceState<W> {
```
bewakes: Can we split this out to  a separate module?

crates/chain-worker-new/src/service.rs
```diff
@@ -0,0 +1,352 @@
...
+        info!(%blkid, "initializing chain worker");
+
+        self.cur_tip = cur_tip;
+        self.initialized = true;
+
+        Ok(())
+    }
+
+    /// Tries to execute a block using the new OL STF.
+    fn try_exec_block(&mut self, block_commitment: &OLBlockCommitment) -> WorkerResult<()> {
```
bewakes: This function is doing a lot of things. Can this be split sensibly?

crates/chain-worker-new/src/service.rs
```diff
@@ -0,0 +1,352 @@
...
+
+    /// Parameters for the chain.
+    #[expect(unused, reason = "params will be used for chain configuration")]
+    params: Arc<Params>,
+
+    /// Context for the worker.
+    context: W,
+
+    /// Handle for the execution controller.
+    exec_ctl_handle: ExecCtlHandle,
```
bewakes: We don't need this. This is legacy.

crates/chain-worker-new/src/service.rs
```diff
@@ -0,0 +1,352 @@
...
+                    .fetch_header(parent_commitment.blkid())?
+                    .ok_or(WorkerError::MissingOLBlock(*parent_commitment.blkid()))?,
+            )
+        };
+
+        // 2. Execute EE payload (validates the EL block)
+        self.exec_ctl_handle
+            .try_exec_el_payload_blocking(*block_commitment)
+            .map_err(|_| WorkerError::InvalidExecPayload(*block_commitment))?;
+
```
bewakes: This should not be needed.

crates/chain-worker-new/src/service.rs
```diff
@@ -0,0 +1,352 @@
...
+            logs,
+            write_batch,
+            indexer_writes.clone(),
+        );
+
+        self.context
+            .store_block_output(*block_commitment, &output)?;
+        self.context
+            .store_auxiliary_data(*block_commitment, &indexer_writes)?;
+
```
bewakes: Like I mentioned earlier, the ideal design would be to have these "side-effect" stuffs outside the state, inside the service itself. Feel free to disagree though.

crates/chain-worker-new/src/service.rs
```diff
@@ -0,0 +1,352 @@
...
+        self.exec_ctl_handle
+            .update_safe_tip_blocking(tip)
+            .map_err(WorkerError::ExecEnvEngine)?;
+
+        Ok(())
+    }
+
+    /// Finalizes an epoch, updating the EE and internal state.
+    fn finalize_epoch(&mut self, epoch: EpochCommitment) -> WorkerResult<()> {
+        self.exec_ctl_handle
```
bewakes: This should not be needed.

crates/chain-worker-new/src/traits.rs
```diff
@@ -0,0 +1,79 @@
...
+use strata_primitives::epoch::EpochCommitment;
+
+use crate::{OLBlockExecutionOutput, WorkerResult};
+
+/// Context trait for a worker to interact with the database.
+///
+/// This trait abstracts the database access layer, allowing the worker to be
+/// tested with mock implementations. All methods should be blocking operations
+/// as the worker runs on a dedicated thread pool.
+pub trait WorkerContext: Send + Sync + 'static {
```
bewakes: Is `ChainWorkerContext` a better name? But I can sense that this will collide with the actual implementing struct's name.

crates/chain-worker-new/src/traits.rs
```diff
@@ -0,0 +1,79 @@
...
+    // =========================================================================
+
+    /// Stores the block execution output (write batch, state root).
+    fn store_block_output(
+        &self,
+        commitment: OLBlockCommitment,
+        output: &OLBlockExecutionOutput,
+    ) -> WorkerResult<()>;
+
+    /// Stores auxiliary data for indexing (inbox messages, manifests, snark updates).
```
bewakes: I don't think we write snark updates.

crates/chain-worker-new/src/handle.rs
```diff
@@ -0,0 +1,115 @@
...
+use crate::{WorkerError, WorkerResult, message::ChainWorkerMessage};
+
+/// Handle for interacting with the chain worker service.
+///
+/// This provides an ergonomic API for sending commands to the worker
+/// and waiting for results.
+#[derive(Debug)]
+pub struct ChainWorkerHandle {
+    #[expect(unused, reason = "will be used later for shared state")]
+    shared: Arc<Mutex<WorkerShared>>,
```
bewakes: I know this `shared` thing comes as a legacy, but isn't the point of having handles like this is to communicate via message passing. I suggest to remove this field unless actually required.

crates/chain-worker-new/src/service.rs
```diff
@@ -0,0 +1,352 @@
...
+
+    fn wait_for_genesis_and_resolve_tip(&self) -> WorkerResult<OLBlockCommitment> {
+        info!("waiting until genesis");
+
+        let init_state = self
+            .runtime_handle
+            .block_on(self.status_channel.wait_until_genesis())
+            .map_err(|_| WorkerError::ShutdownBeforeGenesis)?;
+
+        let cur_tip = match init_state.get_declared_final_epoch() {
```
bewakes: I might be wrong but we might not need the concept of declared final epoch now. This is already confusing in that what other sorts of final epochs are there, is there an undeclared final epoch, etc.

crates/chain-worker-new/src/service.rs
bewakes: Just had a look into the service framework and this is not supported by the service framework at the moment. But we should definitely move towards separating those out in the future.

crates/chain-worker-new/src/handle.rs
storopoli: removed in eac1801a768898f32beffa06a9d719f468d5d9ad

crates/chain-worker-new/src/traits.rs
storopoli: Corrected in 33e11e3ba163e331910d0a6afdee46682ef45849

crates/chain-worker-new/src/service.rs
storopoli: Removed in a2b3a3c111b054d06ad079bbaa7e78357378b799

crates/chain-worker-new/src/service.rs
storopoli: Removed in a2b3a3c111b054d06ad079bbaa7e78357378b799

crates/chain-worker-new/src/service.rs
storopoli: Removed in a2b3a3c111b054d06ad079bbaa7e78357378b799

crates/chain-worker-new/src/service.rs
storopoli: Done in 9e05b2674cf0f573cccb860753240b83f95f76ab

crates/chain-worker-new/src/service.rs
storopoli: Done in 1ad6b55f99db846d115ae87e3b8a7995cd4ada48

crates/chain-worker-new/src/traits.rs
storopoli: Done in 95c2fc3e7ec30651a4b974bed59159ab5a15d8d4

crates/chain-worker-new/src/service.rs
storopoli: I don't want to touch the service framework in the `strata-common` repo in this PR or have this PR depend on a change there.

So I did a refactoring proposal into better types so that we don't rely on `// NOTE` comments.

See 575830b7cf7edebd103d767fa3d9e61933bda752

crates/chain-worker-new/src/service.rs
storopoli: See comment above wrt to 575830b7cf7edebd103d767fa3d9e61933bda752

crates/chain-worker-new/src/service.rs
storopoli: Check 80180c4245270bf6a29d6f033eaf751fc63b2ead

crates/chain-worker-new/src/handle.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Finalize an epoch, making whatever database changes necessary (blocking).
+    pub fn finalize_epoch_blocking(&self, epoch: EpochCommitment) -> WorkerResult<()> {
+        self.command_handle
+            .send_and_wait_blocking(|completion| {
+                ChainWorkerMessage::FinalizeEpoch(epoch, completion)
+            })
+            .map_err(convert_service_error)?
+    }
+
+    /// Update the safe tip, informing the EE of the new tip (async).
```
bewakes: I am not sure if we actually require this. I am pretty sure we don't need to inform the EE. You can just remove the comment mentioning EE here.

crates/chain-worker-new/src/handle.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub async fn update_safe_tip(&self, safe_tip: OLBlockCommitment) -> WorkerResult<()> {
+        self.command_handle
+            .send_and_wait(|completion| ChainWorkerMessage::UpdateSafeTip(safe_tip, completion))
+            .await
+            .map_err(convert_service_error)?
+    }
+
+    /// Update the safe tip, informing the EE of the new tip (blocking).
```
bewakes: Same as above.

crates/chain-worker-new/src/message.rs
```diff
@@ -0,0 +1,21 @@
...
+#[derive(Debug)]
+pub enum ChainWorkerMessage {
+    /// Try to execute a block at the given commitment.
+    TryExecBlock(OLBlockCommitment, CommandCompletionSender<WorkerResult<()>>),
+
+    /// Finalize an epoch, updating database state accordingly.
+    FinalizeEpoch(EpochCommitment, CommandCompletionSender<WorkerResult<()>>),
+
+    /// Update the safe tip, informing the EE of the new tip.
+    UpdateSafeTip(OLBlockCommitment, CommandCompletionSender<WorkerResult<()>>),
```
bewakes, line 20: Same as above, no need to mention EE.

crates/chain-worker-new/src/state.rs
```diff
@@ -0,0 +1,392 @@
...
+            .fetch_ol_state(parent_commitment)?
+            .ok_or(WorkerError::MissingPreState(parent_commitment))?;
+
+        // Execute and extract outputs
+        let (write_batch, indexer_writes) =
+            Self::run_stf_verification(&parent_state, block, parent_header)?;
+
+        // Use the state root from the header (verify_block validated it)
+        let computed_state_root = *block.header().state_root();
+        let logs = Vec::new(); // TODO: Collect logs from execution context when available
```
bewakes: Do we need logs here? If so then calling the `verify_block` function in stf will not be sufficient because it doesn't emit the logs, just verifies it. You need to either modify that function or call other functions, or create/update functions if not existent.

crates/chain-worker-new/src/traits.rs
```diff
@@ -0,0 +1,79 @@
...
+use strata_primitives::epoch::EpochCommitment;
+
+use crate::{OLBlockExecutionOutput, WorkerResult};
+
+/// Context trait for a worker to interact with the database.
+///
+/// This trait abstracts the database access layer, allowing the worker to be
+/// tested with mock implementations. All methods should be blocking operations
+/// as the worker runs on a dedicated thread pool.
+pub trait ChainWorkerContext: Send + Sync + 'static {
```
bewakes, line 17: I do not see this trait implemented anywhere. Is this supposed to be implemented in another ticket? I was under the assumption that since this is a chain worker PR, the context would be implemented here itself.

crates/chain-worker-new/src/state.rs
```diff
@@ -0,0 +1,392 @@
...
+            .context
+            .fetch_block(blkid)?
+            .ok_or(WorkerError::MissingOLBlock(*blkid))?;
+
+        let parent_blkid = block.header().parent_blkid();
+        let parent_commitment = if parent_blkid.is_null() {
+            OLBlockCommitment::null()
+        } else {
+            // We need to figure out the parent slot. For now, assume slot-1.
+            // TODO: Properly track parent slot
```
bewakes: What do you mean by "properly track parent slot"? If there is no parent then would it not mean this is a genesis block?

crates/chain-worker-new/src/handle.rs
storopoli: Done in https://github.com/alpenlabs/alpen/commit/c225a60850f2b930d7184b3a9e1f01ca3f16e789. Removed the 'informing the EE of the new tip' from all three locations.

crates/chain-worker-new/src/handle.rs
storopoli: Done in https://github.com/alpenlabs/alpen/commit/c225a60850f2b930d7184b3a9e1f01ca3f16e789. Removed the 'informing the EE of the new tip' from all three locations.

crates/chain-worker-new/src/message.rs
storopoli: Done in https://github.com/alpenlabs/alpen/commit/c225a60850f2b930d7184b3a9e1f01ca3f16e789. Removed the 'informing the EE of the new tip' from all three locations.

crates/chain-worker-new/src/state.rs
storopoli: Good catch. The `verify_block` function doesn't emit logs - it only validates the block. For this initial implementation, I've left logs as empty (`Vec::new()`). If we need logs in the future (e.g., forindexing or debugging), we'll need to either:

1. Modify the STF to return logs as part of execution output, or
2. Add a separate execution path that captures logs

For now, is it acceptable to leave this as empty and address it when there's a concrete use case for logs?
 
 Cc @prajwolrg

crates/chain-worker-new/src/state.rs
storopoli: Done in f8c5826a18fd4ab384b9c03794765efa5e9cc670

crates/chain-worker-new/src/traits.rs
storopoli: Done in d2519296df2555b5e6ec498446b0640b6d3c19d2

crates/chain-worker-new/src/state.rs
bewakes: Yeah, we can leave this empty for now.

crates/chain-worker-new/src/context.rs
```diff
@@ -0,0 +1,195 @@
...
+
+        // Process blocks slot by slot until we reach the terminal
+        while current_slot < terminal.slot() {
+            current_slot += 1;
+
+            // Get blocks at this slot
+            let block_ids = self.fetch_blocks_at_slot(current_slot)?;
+
+            // For each block, check if it's in our chain and apply its write batch
+            for blkid in block_ids {
```
bewakes: This doesn't feel quite right because it is trying to apply write batch from all the blocks in a slot. Only one block's changes should be applied per slot.
There doesn't seem to be a way to find the canonical block at given slot. But I believe, we can traverse back from the current terminal block tracing the previous block upto the `prev_terminal` and just get that block and its writebatches and apply them.

bewakes: Also, if the writebatch doesn't exist, it should error out. Because for each slot, there must be a writebatch that we can apply.

crates/chain-worker-new/src/state.rs
```diff
@@ -0,0 +1,391 @@
...
+            .runtime_handle
+            .block_on(self.deps.status_channel.wait_until_genesis())
+            .map_err(|_| WorkerError::ShutdownBeforeGenesis)?;
+
+        // Get the last finalized epoch from the checkpoint, if any
+        let finalized_epoch = init_state
+            .get_last_finalized_checkpoint()
+            .map(|ckpt| ckpt.batch_info.get_epoch_commitment());
+
+        let cur_tip = match finalized_epoch {
```
bewakes: I don't think this should be the finalized tip. Won't this mean that every time the node (re)starts, it will discard the non finalized blocks and starts executing from the last finalized tip?

This should be just the valid tip block in the current database(if there are many blocks for the tip slot, we can chose one of them).

bewakes (changes requested):
Noticed a couple of potential issues mostly around writebatch application and initialization of the worker with chain tip.

crates/chain-worker-new/src/state.rs
storopoli: Removed logs from OLBlockExecutionOutput. The STF's verify_block validates logs internally via logs_root comparison - we don't need to store them separately. See commit 69d9cde20.

crates/chain-worker-new/src/traits.rs
storopoli: The trait is implemented by ChainWorkerContextImpl in context.rs (lines 58-195).

crates/chain-worker-new/src/context.rs
storopoli: Fix `merge_finalized_epoch` to follow canonical chain in e45ee78c6560f86c885af718693c2cd92e31f0f0

storopoli: Added an error if `WriteBatch` is not available.

storopoli: Crap forgot to push 2cdd81645f571a8717983cda26b72c6a7677dac5

crates/chain-worker-new/src/state.rs
storopoli: I've fixed the chain tip initialization to resume from actual database tip instead of finalized tip in 2e8cf389b98d18157045b0cb683ebfdbeacb8949

crates/chain-worker-new/src/context.rs
storopoli: Crap, this one 6e779d813eb8bd84755e0f69a5037cea125ff2af

crates/chain-worker-new/src/state.rs
storopoli: Correction it became this one 6e779d813eb8bd84755e0f69a5037cea125ff2af

bewakes (approved):

PR #1250, opened 2026-01-02
Str 1759 replace buf32 bitcointxid

purusang:
## Description
In this pr we refactor L1Status type from `csm-types`. `last_published_txid` now uses `Option<BitcoinTxid>` instead of `Option<Buf32>`. I have refactored the Debug and Display trait to reduce repetition and added the unit tests to ensure the refactor does not introduce error or change in behavior.

### AI usage declaration
I have used Claude code to generate unit tests and analyse the code before refactoring.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues
STR-1759

### DIFF
--- a/crates/csm-types/src/status.rs
+++ b/crates/csm-types/src/status.rs
@@ -1,12 +1,11 @@
-use std::{fmt, str};
+use std::fmt;
 
 use arbitrary::Arbitrary;
-use const_hex as hex;
 use serde::{Deserialize, Serialize};
 use strata_btc_types::BitcoinTxid;
 
 /// Data that reflects what's happening around L1
-#[derive(Clone, Serialize, Deserialize, Default, Arbitrary)]
+#[derive(Clone, Serialize, Deserialize, Default, Arbitrary, Debug)]
 pub struct L1Status {
     /// If the last time we tried to poll the client (as of `last_update`)
     /// we were successful.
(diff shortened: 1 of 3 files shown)


crates/csm-types/src/status.rs
```diff
@@ -23,7 +23,7 @@ pub struct L1Status {
     pub cur_tip_blkid: String,
 
     /// Last published txid where L2 blob was present
-    pub last_published_txid: Option<Buf32>,
+    pub last_published_txid: Option<BitcoinTxid>,
```
bewakes, line 25: If we are using `BitcoinTxId` then the default string representation should already be in little endian order. Can you look into if we can just simply derive `Debug` on this struct? Or at the very least just use `format!("{:?}", self.last_published_txid)` instead of doing all this reversing and unsafe code.

bewakes (commented):
Looks alright, but I still think there's a lot going on for just doing debug impl on a fairly simple struct.

crates/csm-types/src/status.rs
purusang: Thanks, will correct it.

bewakes (approved):

PR #1252, opened 2026-01-02
ci: add `miri` to the lints

storopoli:
## Description

Miri is an [Undefined Behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) detection tool for Rust.

See https://github.com/rust-lang/miri

This PR adds `cargo miri test` to all crates that have `unsafe` code. I had to exclude couple crates that use directly `secp256k1` FFI bindings because `miri` does not work with these. Also `miri` does not support `tokio::test` macros. Excluded:

- `strata-primitives`
- `alpen-cli`
- `strata-key-derivation`
- `strata-btcio`
- `strata-test-utils-btcio`

Included:

- `strata-state` (BUG: found, more details below)
- `strata-identifiers`
- `strata-dbtool`
- `alpen-client`
- `strata-csm-types`
- `strata-acct-types`

## TODO

- [ ] Find a way to avoid `proptest!` macro tests with `miri` because they take way more than 1h.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The bug found in `strata-state` is the following.
The original code in `pop_front_n_arr` had undefined behavior due to a Stacked Borrows violation:

```rs
// BEFORE (buggy):
let slice_ptr = slice_box.as_ptr();  // Creates SharedReadOnly borrow
std::mem::forget(slice_box);          // Invalidates the SharedReadOnly borrow  
Box::<[T; N]>::from_raw(slice_ptr as *mut [T; N])  // Uses invalidated pointer!
```

The issue is that `as_ptr()` creates a shared reference (`SharedReadOnly` tag in Miri's Stacked Borrows model), and then `forget()` moves the `Box` which invalidates that shared reference. When we later try to use `slice_ptr` with `from_raw()`, we're using an invalidated pointer.

The fix: Use `Box::into_raw()` which properly consumes the `Box` and returns an owned raw pointer:

```rs
// AFTER (correct):
Box::<[T; N]>::from_raw(Box::into_raw(slice_box) as *mut [T; N])
```

Miri found the bug and I fixed with the aid of Claude Code.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-1992

### DIFF
--- a/crates/state/src/state_queue.rs
+++ b/crates/state/src/state_queue.rs
@@ -214,14 +214,12 @@ impl<T> StateQueue<T> {
             new_entries
         };
 
-        // TODO verify that this is safe
         let slice_box = out.into_boxed_slice();
-        let slice_ptr = slice_box.as_ptr();
         assert_eq!(slice_box.len(), N);
-        let arr_box = unsafe {
-            mem::forget(slice_box);
-            Box::<[T; N]>::from_raw(slice_ptr as *mut [T; N])
-        };
+        // SAFETY: We've verified the slice has exactly N elements, so the pointer
+        // cast from [T] to [T; N] is valid. Box::into_raw consumes the Box and
+        // transfers ownership of the allocation to the new Box.
+        let arr_box = unsafe { Box::<[T; N]>::from_raw(Box::into_raw(slice_box) as *mut [T; N]) };
 
         // Now just copy it to the stack.
         self.base_idx += N as u64;


voidash (dismissed):
There is a conflict so might need rebase

voidash (approved):

PR #1254, opened 2026-01-04
Str 1532 Validate checkpoint vk matches with the loaded elf 

purusang:
## Description
In this pr we add an startup check that validates the vk committed in params.json matches with the elf file that is used by prover to generate proofs. This check enable finding mismatch between proving and verifying key the moment we start/deploy prover.

### AI usage declaration
I have used Claue-code as a copilot to plan and suggest code.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
I have tested this function locally by adding a test by using the actual elf and params. For CI I am not sure if we want to have that test as that requires release build.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues
STR-1532

### DIFF
--- a/bin/prover-client/src/main.rs
+++ b/bin/prover-client/src/main.rs
@@ -82,8 +82,8 @@ fn main_inner(args: Args) -> anyhow::Result<()> {
     {
         // The checkpoint elf lazily initializes after this call and later
         // the checkpoint proving task utilizes the same.
-        let checkpoint_vk = get_checkpoint_groth16_vk()
-            .context("Failed to get checkpoint verification key")?;
+        let checkpoint_vk =
+            get_checkpoint_groth16_vk().context("Failed to get checkpoint verification key")?;
         let params_vk = rollup_params
             .checkpoint_predicate
             .as_buf_ref()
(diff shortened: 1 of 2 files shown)


bin/prover-client/src/main.rs
```diff
@@ -72,6 +77,11 @@ fn main_inner(args: Args) -> anyhow::Result<()> {
         .resolve_and_validate_rollup_params()
         .context("Failed to resolve and validate rollup parameters")?;
 
+    // Validate checkpoint VK matches between params and loaded ELF
+    #[cfg(feature = "sp1")]
+    validate_checkpoint_vk(&rollup_params)
+        .context("Checkpoint verification key validation failed")?;
+
```
prajwolrg: Instead of having this function also load the ELF, maybe we can split this into two separate functions:
1. Load ELF.
2. Check that the vk matches. We should not need to pass the whole `rollup_params` for that.

Since we are loading the ELF somewhere else as well, it might be better to do the test there as well.

storopoli:
I've discussed this with @purusang and he mentioned that @prajwolrg thought that the code is fine but should be agnostic of zkVM backend. I asked @purusang to crate a new tech debt ticket for that so that we can merge the code as it is.

PR #686, opened 2025-02-24
STR-1084: Introduce retries for transient failures of tasks in prover-client

evgenyzdanovich:
## Description

 - Implement retry mechanism in the `prover-client`. From now on, failed tasks with RPC errors are retried. 
 - Implement exponential backoff retry strategy (15 retries, back-off exponentially, across 1hr)
 - Some unit tests that cover new/modified functionality.
 - Re-purpose an existing `prover_checkpoint_runner` test to check for normal checkpoint proving when sequencer is unreliable.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [x] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [x] New and existing tests pass with my changes.

PR #1255, opened 2026-01-04
test fix: prover checkpoint runner

barakshani:
## Description

A fix for a flaky test: the test is flaky because after we restart the sequencer and connect to it again (with `create_rpc()`, I believe) we _sometimes_ get an error (`WARNING:root:caught exception <class 'ConnectionRefusedError'>, will still wait for timeout: [Errno 111] Connection refused`), which probably retries and then succeeds.
In the current state this causes us to "miss" the checkpoint we are waiting for, as the sequencer already posts a new checkpoint.

Actually, the test was "broken": after the restart we were merely checking if an old checkpoint is on chain, but what we need to test is the prover keep on producing proofs. This is how it was in the original test (https://github.com/alpenlabs/alpen/pull/686/), but then changed to the current state (https://github.com/alpenlabs/alpen/pull/872).

This is 'reverted' and a more detailed test description is added.

The current modification should already fix the test flakiness, however there are few more things I want to look at, mainly the "weird implementation" part, but also the fact that if we miss the specific epoch we were waiting for (i.e. more advanced epochs are checkpointed), we don't end the test.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

### DIFF
--- a/functional-tests/tests/prover/prover_checkpoint_runner.py
+++ b/functional-tests/tests/prover/prover_checkpoint_runner.py
@@ -76,6 +76,7 @@ class ProverCheckpointRunnerTest(testenv.StrataTestBase):
         time.sleep(10)
         sequencer.start()
         sequencer_rpc = sequencer.create_rpc()
+        seq_waiter.wait_until_client_ready(timeout=15)
 
         # Wait for target epoch to be confirmed; for debug use the same function as above,
         # it has more logs.


functional-tests/tests/prover/prover_checkpoint_runner.py
```diff
@@ -84,9 +89,9 @@ def consecutive_checkpoints_verified():
         def _ck2():
             ckpt_idx = sequencer_rpc.strata_getLatestCheckpointIndex(False)
             logging.info(f"cur checkpoint idx: {ckpt_idx}")
-            return ckpt_idx == self.checkpoint_settings.consecutive_proofs_required
+            return ckpt_idx == self.checkpoint_settings.consecutive_proofs_required * 2
```
storopoli: Maybe we should use `>=`. Using `==` means if checkpoint advances past 6 before the check, the test times out.

barakshani: yeah, that was part of the problem. As I write in the description, I am looking at alternatives, as we already have better methods to check the same thing.

functional-tests/tests/prover/prover_checkpoint_runner.py
```diff
@@ -14,12 +14,16 @@ class ProverCheckpointSettings:
     """Test configuration for checkpoint-based prover"""
 
     consecutive_proofs_required: int = 3
-    prover_timeout_seconds: int = 600
+    # waiting time for inner functions should end after 20 seconds, so 60 is more than enough
+    prover_timeout_seconds: int = 60
```
bewakes, line 18: This is great!!

PR #1266, opened 2026-01-06
test fix: failing on dbtool revert call

barakshani:
## Description

The specific flakiness this PR addresses happens when we try to revert to a target block that is already in a (sealed) checkpoint, even thought the checkpoint isn't final (may be that it hasn't been posted yet, or potentially also confirmed but not "final" (not sure about the latter, maybe in some edge case)).

I resolve this in 2 ways:
1. The prover is stopped to ensure that we don't end up sealing a new batch (epoch) after we will already checked what was the latest final slot. This, I believe, should already fix the problem (in addition to have a few more blocks produced).
2. (used only in the non-dbtool test) We have a flag that allows the block revert tool reverting (non-final) checkpointed blocks, which is now set to true. This should not be an issue for this particular test.

**
These tests are likely to be changed quite a bit with the new design.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-2046](https://alpenlabs.atlassian.net/browse/STR-2046), [STR-2049](https://alpenlabs.atlassian.net/browse/STR-2049)

[STR-2046]: https://alpenlabs.atlassian.net/browse/STR-2046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/functional-tests/tests/dbtool/revert_chainstate_fn.py
+++ b/functional-tests/tests/dbtool/revert_chainstate_fn.py
@@ -70,14 +70,15 @@ class RevertChainstateFnTest(FullnodeDbtoolMixin):
                 f"EL={old_fn_el_block_number}"
             )
 
+        # extra buffer time to let latest checkpoint get final
+        time.sleep(2)
+
         # Stop services to use dbtool
         self.seq.stop()
         self.reth.stop()
         self.follower_1_node.stop()
         self.follower_1_reth.stop()
 
-        # extra buffer time to let latest checkpoint get final
-        time.sleep(2)
         # Get checkpoint info and target block
         checkpt = get_latest_checkpoint(self)
         if not checkpt:
--- a/functional-tests/utils/dbtool.py
+++ b/functional-tests/utils/dbtool.py
@@ -17,7 +17,7 @@ def send_tx(web3: Web3):
         }
     )
     print("txid", txid.to_0x_hex())
-    web3.eth.wait_for_transaction_receipt(txid, timeout=10)
+    web3.eth.wait_for_transaction_receipt(txid, timeout=5)
 
 
 def run_dbtool_command(datadir: str, subcommand: str, *args) -> tuple[int, str, str]:


functional-tests/tests/cl_reorg_resume_blockproduction.py
```diff
@@ -52,17 +41,20 @@ def main(self, ctx: flexitest.RunContext):
...
 
+        # stop prover -- additional safety to make sure we don't produce checkpoints too quickly
+        prover.stop()
+
         orig_blocknumber = seqrpc.strata_syncStatus()["tip_height"]
 
         # ensure there are some blocks more than our tip height
         wait_until(
-            lambda: int(rethrpc.eth_blockNumber(), base=16) > orig_blocknumber + 1,
+            lambda: int(rethrpc.eth_blockNumber(), base=16) > orig_blocknumber + 5,
```
krsnapaudel, line 51: What is the idea for +5 here? We are not tracking this.

functional-tests/tests/dbtool/revert_chainstate_delete_blocks.py
```diff
@@ -29,6 +29,21 @@ def __init__(self, ctx: flexitest.InitContext):
...
         setup_revert_chainstate_test(self)
+        prover = ctx.get_service("prover_client")
+        # stop prover -- additional safety to make sure we don't produce checkpoints too quickly
+        prover.stop()
+
+        cur_block = int(self.rethrpc.eth_blockNumber(), base=16)
+
+        # ensure there are some blocks more than our tip height
+        wait_until(
+            lambda: int(self.rethrpc.eth_blockNumber(), base=16) > cur_block + 3,
```
krsnapaudel, line 40: Same here. Do we just want some blocks past the strata tip?

functional-tests/tests/dbtool/revert_chainstate_fn.py
```diff
@@ -28,6 +31,18 @@ def main(self, ctx: flexitest.RunContext):
         # Setup: generate blocks and finalize epoch
         setup_revert_chainstate_test(self, web3_attr="web3")
 
+        cur_block = int(self.rethrpc.eth_blockNumber(), base=16)
+
+        # ensure there are some blocks more than our tip height
+        wait_until(
+            lambda: int(self.rethrpc.eth_blockNumber(), base=16) > cur_block + 3,
```
krsnapaudel, line 38: Same here. Is there a reason we used 5 in one test vs 3 here?

functional-tests/tests/dbtool/revert_chainstate_seq.py
```diff
@@ -29,6 +29,20 @@ def __init__(self, ctx: flexitest.InitContext):
...
         # Setup: generate blocks and finalize epoch
         setup_revert_chainstate_test(self)
+        prover = ctx.get_service("prover_client")
+        # stop prover -- additional safety to make sure we don't produce checkpoints too quickly
+        prover.stop()
+
+        cur_block = int(self.rethrpc.eth_blockNumber(), base=16)
+        # ensure there are some blocks more than our tip height
+        wait_until(
+            lambda: int(self.rethrpc.eth_blockNumber(), base=16) > cur_block + 3,
```
krsnapaudel, line 39: Same as above.

functional-tests/tests/dbtool/revert_checkpointed_block_fn.py
```diff
@@ -28,6 +31,18 @@ def main(self, ctx: flexitest.RunContext):
         # Setup: generate blocks, finalize epoch 1, and wait for checkpoint 2
         setup_revert_chainstate_test(self, web3_attr="web3")
 
+        cur_block = int(self.rethrpc.eth_blockNumber(), base=16)
+
+        # ensure there are some blocks more than our tip height
+        wait_until(
+            lambda: int(self.rethrpc.eth_blockNumber(), base=16) > cur_block + 3,
+            error_with="not building blocks",
```
krsnapaudel, line 39: Same as above.

functional-tests/tests/dbtool/revert_checkpointed_block_seq.py
```diff
@@ -44,10 +52,11 @@ def main(self, ctx: flexitest.RunContext):
             )
 
         # Stop services to use dbtool
-        self.seq_signer.stop()
         self.seq.stop()
         self.reth.stop()
 
+        # extra buffer time to let latest checkpoint get final
```
krsnapaudel, line 58: Should we be using waiters instead of sleep?

functional-tests/tests/dbtool/revert_chainstate_seq.py
```diff
@@ -55,14 +68,6 @@ def main(self, ctx: flexitest.RunContext):
 
         target_block_id, target_slot = target_end_of_epoch(checkpt["l2_range"])
 
-        # Ensure we have blocks outside checkpointed range
-        sync_info = self.get_syncinfo()
```
krsnapaudel, line 59: Why are we deleting this? Did we really ensure that we will have blocks outside of checkpointed range?

functional-tests/tests/dbtool/revert_checkpointed_block_seq.py
```diff
@@ -29,6 +31,12 @@ def __init__(self, ctx: flexitest.InitContext):
     def main(self, ctx: flexitest.RunContext):
         # Setup: generate blocks, finalize epoch 1, and wait for checkpoint 2
         setup_revert_chainstate_test(self)
+        prover = ctx.get_service("prover_client")
```
krsnapaudel, line 34: Why don't we do wait for additional reth blocks here?

functional-tests/tests/cl_reorg_resume_blockproduction.py
barakshani: see pr description, we need to "buy" more time before we check for latest checkpoint.

functional-tests/tests/dbtool/revert_chainstate_fn.py
barakshani: still experimenting what's the flakiness "threshold".

functional-tests/tests/dbtool/revert_chainstate_seq.py
barakshani: see [commit](https://github.com/alpenlabs/alpen/pull/1266/commits/0efa61cc449c3673f67b5378f2d3b032222420c3) comment/description -- it already addresses this question:
1. we should not skip tests and return true.
2. we stop the prover and then create more blocks, so yes, we will have blocks outside the checkpoint range.

functional-tests/tests/dbtool/revert_checkpointed_block_seq.py
barakshani: as in the test description, time delay should be enough, so still want to check if we have flakiness in one of the approaches.

functional-tests/tests/dbtool/revert_checkpointed_block_seq.py
barakshani: what would you suggest to wait for?

krsnapaudel: We already wait for epoch 1 finalization at setup. The goal here is to let another checkpoint be created but not finalized. This is tricky to do with sleep or waits.

barakshani: The prover is stopped -- exactly for the reason to make sure we don't create another checkpoint -- so we cannot have another checkpoint.
Ideally we would stop the communication with btcio, so even if we have more checkpoints, they don't get broadcast, but I wasn't able to stop it.

functional-tests/tests/cl_reorg_resume_blockproduction.py
delbonis: Should probably make this a constant at the top of the file so that we know what it represents and can maybe more easily tweak it.

functional-tests/tests/dbtool/revert_chainstate_fn.py
delbonis, line 38: Can we use the `RethWaiter` or something else in this vein to avoid using the low-level wait fns directly?

functional-tests/tests/cl_reorg_resume_blockproduction.py
barakshani: This would be overoptimization if you ask me, as we use this value only once. It would have made sense to have a general variable used across all tests, but this is not in the scope of this PR.

functional-tests/tests/dbtool/revert_chainstate_fn.py
barakshani: Almost all tests here use these low-level functions, but also note that as we decouple OL and Reth, for this revert tool tests we won't need waiters for reth blocks, but for OL blocks. This restructure is not part of this PR -- the purpose is just to see if we are able to decrease the overall flakiness, and in particular to make dbtool flakiness stop.

functional-tests/tests/dbtool/revert_checkpointed_block_seq.py
krsnapaudel: My point is: for `*checkpointed_block_*` tests, we do need non-finalized checkpoint.

barakshani: I don't change the target block we are reverting to (if it was a final block, revert would fail).

The issue with these tests was that we called get_chkp too early: we seal and post a checkpoint `j` and then call get checkpoint. Checkpoint `j` may not be final yet on L1, and therefore our node is not aware of it and we get back from get_chkp checkpoint `j-1`, and we ask to revert to the first block after checkpoint `j-1`, which is the first block of `j`. By the time we call revert, `j` becomes final, so we _were trying_ to revert to a final block, and so the test failed. This is where the flakiness came from.

What I'm doing here is waiting a bit before we call get checkpoint, to allow checkpoint `j` become final. Then we revert to the next block after checkpoint `j`, which is the first block of `j+1`, however because the prover is stopped, there's no risk that `j+1` will be submitted to L1 and become final.

krsnapaudel: I understand but how does waiting with prover stopped help us create a non-finalized checkpoint? I do acknowledge the limitation of the test, but not convinced about the fix.

functional-tests/tests/dbtool/revert_chainstate_fn.py
krsnapaudel: TBH we should revisit them when we have new OL blocks, checkpoints and related dbs in place.

functional-tests/tests/dbtool/revert_checkpointed_block_seq.py
barakshani: 1. the point is not to create a non-final checkpoint, it is to make sure we wait sufficient time before we ask for the last onchain checkpoint. In this time, that checkpoint becomes final.
2. by stopping the prover and waiting, we are guaranteed to have more blocks _after_ the checkpoint, so there are blocks to revert.

barakshani:
> > @krsnapaudel do you have any other comments? I would like to push this in to be able to see if we get any of these failures on new PRs, where the CI kicks in.
> 
> @barakshani No comments except the one for reverting to the start of (non-finalized) checkpointed epoch. This is tricky.

Last chance to reply on the comments before I merge this. The review process has been going too slow, and having these fixes in the codebase would help us see if the problem is resolved.

PR #1269, opened 2026-01-07
feat(alpen-ee-sequencer): implement `finalized_blocknum` using OL tracker

storopoli:
## Description

Implement the `CanonicalChainReader::finalized_blocknum()` method for the batch builder's reorg detection. This enables the batch builder to distinguish between batches that can be safely reverted (above finalized height) and batches that are finalized (at or below finalized height).

The implementation adds a new query to `ExecChainHandle` that exposes the finalized block number directly from `ExecChainState`, providing a single-call API instead of requiring separate hash lookup and storage query.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The finalized block is guaranteed to exist in `ExecChainState` - if it doesn't, something has gone seriously wrong and the application should crash. This is why we use a direct query through `ExecChainHandle` rather than a fallible storage lookup.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-2055

### DIFF
--- a/crates/alpen-ee/sequencer/src/batch_builder/canonical.rs
+++ b/crates/alpen-ee/sequencer/src/batch_builder/canonical.rs
@@ -19,14 +19,13 @@ pub(crate) trait CanonicalChainReader: Send + Sync {
     async fn finalized_blocknum(&self) -> Result<u64>;
 }
 
-/// Implementation of `CanonicalChainReader` using `ExecChainHandle` and `ExecBlockStorage`.
+/// Implementation of [`CanonicalChainReader`] using [`ExecChainHandle`] and [`ExecBlockStorage`].
 pub(crate) struct ExecChainCanonicalReader<S> {
     exec_chain: Arc<ExecChainHandle>,
     block_storage: Arc<S>,
 }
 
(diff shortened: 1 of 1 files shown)


crates/alpen-ee/sequencer/src/batch_builder/canonical.rs
```diff
@@ -52,6 +59,18 @@ impl<S: ExecBlockStorage> CanonicalChainReader for ExecChainCanonicalReader<S> {
...
+        // Look up block number in storage
+        let block = self
+            .block_storage
+            .get_exec_block(finalized_hash)
+            .await?
+            .ok_or_else(|| {
+                eyre::eyre!("finalized block not found in storage: {finalized_hash:?}")
+            })?;
+
+        Ok(block.blocknum())
```
sapinb: The finalized height can be fetched from exec_chain directly, but need to add a method to the handler.
ExecChainState -> unfinalized -> finalized -> blocknum

storopoli: Thanks, check now

PR #1272, opened 2026-01-08
STR-1974: Service Level Instrumentation for ASM and DB operations

voidash:
## Description
Modified the inst_ops_generic! macro to automatically add #[instrument] attributes with component fields to all storage operations. Establishes <layer>:<domain> convention (e.g., storage:l1, db:sled:transaction)and now each operation now emits spans with timing metrics (time.busy, time.idle)
  This provides observability into:
  - ASM worker operations: State transitions, manifest storage, block processing
  - Database transactions: Retry behavior, conflicts, transaction lifecycle

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

### DIFF
--- a/crates/db/store-sled/src/lib.rs
+++ b/crates/db/store-sled/src/lib.rs
@@ -8,7 +8,7 @@ pub mod client_state;
 mod config;
 pub mod global_mmr;
 mod init;
-pub mod instrumentation;
+mod instrumentation;
 pub mod l1;
 pub mod l2;
 pub mod macros;
--- a/crates/storage/src/lib.rs
+++ b/crates/storage/src/lib.rs
@@ -2,7 +2,7 @@
 
 mod cache;
 mod exec;
-pub mod instrumentation;
+mod instrumentation;
 mod managers;
 pub mod ops;
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1164,7 +1164,6 @@ dependencies = [
  "strata-ee-chain-types",
  "strata-identifiers",
  "strata-snark-acct-types",
- "strata-storage",
  "strata-storage-common",
  "thiserror 2.0.17",
  "threadpool",
--- a/crates/alpen-ee/database/Cargo.toml
+++ b/crates/alpen-ee/database/Cargo.toml
@@ -15,7 +15,6 @@ strata-ee-acct-types.workspace = true
 strata-ee-chain-types.workspace = true
 strata-identifiers.workspace = true
 strata-snark-acct-types.workspace = true
-strata-storage.workspace = true
 strata-storage-common.workspace = true
 
 async-trait.workspace = true
--- a/crates/alpen-ee/database/src/database.rs
+++ b/crates/alpen-ee/database/src/database.rs
@@ -5,10 +5,9 @@ use alpen_ee_common::{
 use strata_acct_types::Hash;
 use strata_ee_acct_types::EeAccountState;
 use strata_identifiers::{EpochCommitment, OLBlockId};
-use strata_storage::instrumentation::components;
 use strata_storage_common::inst_ops_generic;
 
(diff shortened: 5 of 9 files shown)


crates/asm/worker/src/service.rs
```diff
@@ -53,6 +53,12 @@ impl<W: WorkerContext + Send + Sync + 'static> SyncService for AsmWorkerService<
 
         // Traverse back the chain of l1 blocks until we find an l1 block which has AnchorState.
         // Remember all the blocks along the way and pass it (in the reverse order) to process.
+        let pivot_span = info_span!("asm.pivot_lookup",
+            target_height = height.to_consensus_u32(),
+            target_block = %incoming_block.blkid()
+        );
+        let _pivot_guard = pivot_span.enter();
```
delbonis: This binding get used in the `drop` call so it's technically not correct to use the `_` prefix.  It'd also call it `pivot_span_guard` to be clear about what it is.

But maybe it'd be even better to split out the more complex pivot code into a separate function so we can just create the guard there and then it naturally gets dropped.

crates/asm/worker/src/state.rs
```diff
@@ -115,6 +130,9 @@ impl<W: WorkerContext + Send + Sync + 'static> AsmWorkerServiceState<W> {
         };
 
         // Asm transition.
+        let stf_span = tracing::debug_span!("asm.stf.compute");
```
delbonis: To be consistent with `pre_process` this might be better to be called `process` maybe?

crates/common/src/instrumentation.rs
delbonis, line 1: Not sure how I feel about enumerating these services/components in the common crate, we'd like to keep it generic.  I think it's pretty normal to put these in the crates that actually use them.  Unless there's some reason for it I'm understanding?

Also the doc comment here includes stuff that seems like it's referring to things that were moved somewhere else.

crates/storage/src/managers/l1.rs
```diff
@@ -49,6 +66,14 @@ impl L1BlockManager {
     /// Append [`L1BlockId`] to tracked canonical chain at the specified height.
     // Note: In the new architecture, btcio stores chain tracking data first,
     // then the ASM worker stores manifests asynchronously.
+    #[instrument(
+        skip(self),
+        fields(
+            component = components::STORAGE_L1,
+            block_id = %blockid,
+            height = height,
```
delbonis: Can this just be `%height`?

crates/storage/src/managers/l1.rs
delbonis, line 1: In other logs we use `blkid` as the key for block IDs, but this is using `block_id`.  I realize this isn't a *big* deal.

delbonis (changes requested):
Comments aside, this is ***awesome*** to see being done.  I'm really looking forward to getting interesting real-world database performance metrics out of this and seeing what we learn.

crates/db/store-sled/src/lib.rs
```diff
@@ -8,6 +8,7 @@ pub mod client_state;
 mod config;
 pub mod global_mmr;
 mod init;
+pub mod instrumentation;
```
delbonis: I think we can avoid exporting this publicly, right?

crates/storage/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 
 mod cache;
 mod exec;
+pub mod instrumentation;
```
delbonis: same

storopoli (approved):
ACK 8cd218103e7a2a63443143deab76154ff3976f5c

delbonis (approved):

PR #1278, opened 2026-01-12
(rpc-new): Add block apis in OL rpc

bewakes:
## Description

This PR adds two new rpc methods `getRawBlocks(start, end)` and `getRawBlockById(blkid)` under the namespace strata. These fetch OL blocks from db and serve raw serialized OL blocks which will be used for syncing blocks by fullnodes.

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist


bin/strata/src/rpc.rs
```diff
@@ -331,6 +330,50 @@ impl OLClientRpcServer for OLRpcServer {
...
+        // Fetch blocks in backward order to ensure a valid chain.
+        for _ in (start_height..=end_height).rev() {
+            let blk = self.get_block(cur_blk).await?;
+            cur_blk = blk.header().parent_blkid;
+            blocks.push(blk);
+        }
+        // Reverse back to get chronological sequence.
+        blocks.reverse();
+        let blks: VariableList<_, MAX_RAW_BLOCKS_RANGE> = VariableList::new(blocks)
+            .map_err(|e| internal_error(format!("cannot collect OL blocks: {e}")))?;
```
delbonis: This is effectively defining a new SSZ schema type and also has an issue in that it will break when we support skipped slots.

There should be a type like this which provides more generally useful information that doesn't require the consumer to know how to parse the block to make sense of it:

```rs
struct RpcBlockRangeEntry {
  slot: Slot, // maybe also epoch?
  blkid: Buf32, // or OLBlockId
  raw_block: HexBytes,
}
```

and then return a `Vec<RpcBlockRangeEntry>`.

crates/rpc-new/api/src/lib.rs
```diff
@@ -41,3 +42,17 @@ pub trait OLClientRpc {
...
 }
+
+/// OL RPC methods served by block executing nodes.
+#[cfg_attr(not(feature = "client"), rpc(server, namespace = "strata"))]
+#[cfg_attr(feature = "client", rpc(server, client, namespace = "strata"))]
+pub trait OLFullNodeRpc {
+    /// Get blocks in range as raw bytes of serialized `Vec<OLBlock>`.
+    /// `start_height` and `end_height` are inclusive.
+    #[method(name = "getRawBlocks")]
+    async fn get_raw_blocks(&self, start_height: u64, end_height: u64) -> RpcResult<HexBytes>;
```
delbonis: Should be `get_raw_block_range`/`getRawBlockRange` to be explicit about what it's doing.

bin/strata/src/rpc.rs
bewakes: > This is effectively defining a new SSZ schema type

yeah, because the underlying block is ssz serializable and I wanted not to make this dependent on another serialization scheme. Basically there would be nested serialization: outside(borsh/codec) and the internal ssz. What do you suggest here regarding the serializations?

bewakes: >  that doesn't require the consumer to know how to parse the block to make sense of it

Assuming you are refering to the returned `Hexbyte`, this is how block range rpc was implemented in the older binary. Perhaps this transfers fast? maybe not.

But again in this case also it has to parse those hexbytes for the raw_block right?

bewakes: Done in https://github.com/alpenlabs/alpen/pull/1278/commits/5b89a96bdc27a3c3d26c3b75e6f4e7e47f119fc0

delbonis (approved):

bin/strata/src/rpc.rs
```diff
@@ -50,6 +53,29 @@ impl OLRpcServer {
...
     }
+
+    async fn get_canonical_block_at_height(&self, height: u64) -> RpcResult<Option<OLBlockId>> {
+        let blkid = self
+            .storage
+            .ol_block()
+            .get_blocks_at_height_async(height)
+            .await
+            .map_err(db_error)?
+            .first() // TODO: Assumes the canonical is the first one, but need to define it
```
sapinb, line 64: need to remember to handle this todo.
either all operations should ensure .first() remains the canonical block
or use some other mechanism to store canonical chain.

bewakes: Created a ticket for this. https://alpenlabs.atlassian.net/browse/STR-2105

PR #1283, opened 2026-01-13
Str 2011 Witness generation for a range of alpen blocks

purusang:
## Description
In this work we create the base for generating witness required to prove an arbitrary range a alpen blocks. [EE-prover](https://alpenlabs.atlassian.net/browse/STR-1782) should utilize this crate to orchestrate inner-outer proofs. The output of this crate will be the witness to inner (chunk) proof.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

There is a [docs](https://www.notion.so/Plan-Block-Range-Witness-Extraction-for-Proving-2e3901ba000f80859be5ebff1a51f831?source=copy_link) where this work was planned. There is minor implementation drift for good.

Also I will add a functional tests in ee-prover pr which should/will validate the correctness of current pr.

- [x]  Yes
- [ ]  No

[Block Range Witness Extraction for Proving](https://www.notion.so/Plan-Block-Range-Witness-Extraction-for-Proving-2e3901ba000f80859be5ebff1a51f831?source=copy_link)
[EE prover implementation plan](https://notion.so/EE-Prover-Implementation-Plan-2e6901ba000f80eb836fc3a7cf0323a8?pvs=25)
## Checklist

## Related Issues
STR-2011

### DIFF
--- a/crates/reth/witness/src/range_witness_extractor.rs
+++ b/crates/reth/witness/src/range_witness_extractor.rs
@@ -73,6 +73,14 @@ where
         let start_block_num = start_block.number;
         let end_block_num = end_block.number;
 
+        if start_block_num > end_block_num {
+            return Err(eyre!(
+                "invalid block range: start {} > end {}",
+                start_block_num,
+                end_block_num
+            ));
+        }
+
         debug!(start_block_num, end_block_num, %start_block_hash, %end_block_hash, "extracting range witness");
 
         // Fetch previous block using parent hash
@@ -194,21 +202,15 @@ where
             targets,
         )?;
 
-        // Extract account proofs, track pre-existing accounts
+        // Extract account proofs
         let mut pre_proofs =
             HashMap::with_capacity_and_hasher(touched.len(), DefaultHashBuilder::default());
         let mut post_proofs =
             HashMap::with_capacity_and_hasher(touched.len(), DefaultHashBuilder::default());
-        let mut accessed_accounts = HashSet::new();
 
         for (addr, keys) in &touched {
             pre_proofs.insert(*addr, proof_pre.account_proof(*addr, keys)?);
             post_proofs.insert(*addr, proof_post.account_proof(*addr, keys)?);
-
-            // Only accounts existing at (start_block - 1) are "pre-existing"
-            if pre_state.basic_account(addr)?.is_some() {
-                accessed_accounts.insert(*addr);
-            }
         }
 
         let state =


crates/reth/witness/src/range_witness_extractor.rs
```diff
@@ -0,0 +1,264 @@
+//! Range witness extraction for arbitrary block ranges.
```
evgenyzdanovich, line 1: My current understanding.

this does pretty much the same stuff as before, using the same state provider from `reth` (that does state DB reads under the hood). 
So, I assume, to make it work in an efficient way, we **will still maintain a separate exex** (that was designed exactly for that - to be a background sidecar alongside "core" reth) for witness generation.

Given this is potentially a point of disagreement - lets settle on the using/not using exex here.
+CC @prajwolrg @sapinb @purusang

crates/reth/exex/src/cache_db_provider.rs
```diff
@@ -34,28 +34,32 @@ impl fmt::Debug for CacheDBProvider {
     }
 }
 #[derive(Debug)]
-pub(crate) struct AccessedState {
+pub struct AccessedState {
     accessed_accounts: HashMap<Address, Vec<Uint<256, 4>>>,
     accessed_contracts: Vec<Bytecode>,
     accessed_block_idxs: HashSet<u64>,
 }
```
delbonis: Should the bytecodes also be a `HashMap` of codehash->bytecode?  If not, then I would think we would want to make the others be flat (sorted?) vecs as well.

Also what is the value represented by `Vec<Uint<256, 4>>`?  That's unclear, at least add a `type Foo = ...;` alias, but it would be better to at least make it a `struct Foo(...);` newtype.

Also fix the module organization.  Impl blocks should be next to the types they're impls for.

crates/reth/witness/src/range_witness_extractor.rs
purusang: **Information from a follow up call (for reviewers 😃 )**
We decided to continue witness extraction as in this pr and utilize exex for DA.

crates/reth/exex/src/cache_db_provider.rs
```diff
@@ -6,13 +6,41 @@ use reth_revm::{
...
+
+/// A single storage slot key.
+pub type StorageKey = U256;
+
+/// Accessed state captured during block execution for witness generation.
+#[derive(Debug)]
+pub struct AccessedState {
+    accessed_accounts: HashMap<Address, Vec<StorageKey>>,
+    accessed_contracts: Vec<Bytecode>,
+    accessed_block_idxs: HashSet<u64>,
```
delbonis: See previous comment, I think it got mangled by github.

crates/reth/witness/src/range_witness_extractor.rs
```diff
@@ -0,0 +1,264 @@
...
+            .collect()
+    }
+}
+
+#[derive(Debug, Default)]
+struct AccumulatedState {
+    accounts: HashMap<Address, HashSet<U256>>,
+    bytecodes: HashSet<Bytecode>,
+    block_idxs: HashSet<u64>,
+}
```
delbonis, line 252: Or maybe here?

delbonis (changes requested):
not sure what got juggled around by github here

crates/reth/witness/src/range_witness_extractor.rs
```diff
@@ -0,0 +1,268 @@
...
+            TrieInput::from_state(HashedPostState::from_bundle_state::<KeccakKeyHasher>([])),
+            targets,
+        )?;
+
+        // Extract account proofs, track pre-existing accounts
+        let mut pre_proofs =
+            HashMap::with_capacity_and_hasher(touched.len(), DefaultHashBuilder::default());
+        let mut post_proofs =
+            HashMap::with_capacity_and_hasher(touched.len(), DefaultHashBuilder::default());
+        let mut accessed_accounts = HashSet::new();
```
MdTeach: Any thing missing here ?
I see `accessed_accounts` created and inserted but it doesn’t seem to be used anywhere afterward.

crates/reth/witness/src/range_witness_extractor.rs
```diff
@@ -0,0 +1,268 @@
...
+        let prev_block_hash = start_block.header.parent_hash;
+        let prev_block = self
+            .provider_factory
+            .block_by_hash(prev_block_hash)?
+            .ok_or_else(|| eyre!("previous block not found for hash {}", prev_block_hash))?;
+        let prev_block_num = prev_block.number;
+        let start_state_root = prev_block.header.state_root;
+
+        // 1. Execute all blocks to discover accessed state
+        let accessed = self.execute_blocks_for_accessed_state(start_block_num, end_block_num)?;
```
storopoli, line 96: No validation that `start_block_num <= end_block_num`. A reversed range silently yields an empty execution pass but still builds proofs from mismatched state; this can produce invalid witnesses. Consider explicit range validation early in `extract_range_witness`.

crates/reth/witness/src/range_witness_extractor.rs
storopoli: I have the same concern as @MdTeach

MdTeach (approved):

PR #1284, opened 2026-01-13
fix(asm-aux): index verifiable manifest hash by block height

prajwolrg:
## Description

When requesting the auxiliary data for manifest hash, we are querying by height, but when we are providing the data, we were providing it basis of the index. This PR fixes that mismatch by making things consistent..

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/tests/asm/core.rs
+++ b/tests/asm/core.rs
@@ -87,6 +87,16 @@ async fn test_multiple_block_processing() {
     let harness = create_test_harness()
         .await
         .expect("Failed to create test harness");
+    let (l1, state) = harness.get_latest_asm_state().unwrap().unwrap();
+    assert_eq!(l1, state.state().chain_view.pow_state.last_verified_block);
+    assert_eq!(
+        l1.height().to_consensus_u32() as u64,
+        state
+            .state()
+            .chain_view
+            .history_accumulator
+            .last_inserted_height()
+    );
 
     let block_hashes = harness.mine_blocks(3).await.expect("Failed to mine blocks");
     assert_eq!(block_hashes.len(), 3);
@@ -96,4 +106,13 @@ async fn test_multiple_block_processing() {
         .await
         .expect("Failed to get chain tip");
     assert_eq!(tip_height, harness.genesis_height + 3);
+    assert_eq!(l1, state.state().chain_view.pow_state.last_verified_block);
+    assert_eq!(
+        l1.height().to_consensus_u32() as u64,
+        state
+            .state()
+            .chain_view
+            .history_accumulator
+            .last_inserted_height()
+    );
 }


crates/asm/common/src/aux/errors.rs
```diff
@@ -52,10 +51,10 @@ pub enum AuxError {
...
     },
 
-    /// Manifest hash not found at the given MMR index.
-    #[error("manifest hash not found at index {index}")]
+    /// Manifest hash not found at the given L1 Block height.
+    #[error("manifest hash not found for height {height}")]
     ManifestHashNotFound {
         /// The requested MMR index
-        index: u64,
+        height: u64,
```
barakshani, line 58: Do we index leaves by block height? Current bitcoin block is almost 1,000,000 - what would be with these MMR leaves/indices?

barakshani: I see below we have offset. good. should this be reflected here? if not in code, at least clarify in the comment.

crates/asm/worker/src/aux_resolver.rs
```diff
@@ -195,9 +196,11 @@ impl<'a> AuxDataResolver<'a> {
...
 
             // Calculate MMR indices from L1 heights
-            // MMR index 0 = genesis height, index 1 = genesis + 1, etc.
-            let start_index = start_height.saturating_sub(genesis_height);
-            let end_index = end_height - genesis_height;
+            // MMR index 0 = genesis height + 1, index 1 = genesis + 2, etc.
+            // offset = genesis_height + 1 (height of first block with manifest)
+            let offset = genesis_height + 1;
+            let start_index = start_height.saturating_sub(offset);
+            let end_index = end_height - offset;
```
barakshani: why not `saturating_sub` as well for end index?

barakshani (dismissed):
some small comments

crates/asm/common/src/mmr.rs
```diff
@@ -1,30 +1,69 @@
...
+/// // offset will be 800001 internally
+/// mmr.add_leaf(hash1)?; // MMR index 0, L1 block height 800001
+/// mmr.add_leaf(hash2)?; // MMR index 1, L1 block height 800002
+/// ```
+#[derive(Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize, Serialize, Deserialize)]
+pub struct AsmManifestMmr {
+    mmr: CompactMmr64<Hash32>,
+    /// Height offset for mapping MMR indices to L1 block heights.
+    /// Equal to `genesis_height + 1` since manifests start after genesis.
+    offset: u64,
```
delbonis, line 49: Shouldn't this be a fixed parameter taken from the `AsmSpec` or something?  Or maybe we *do* want it to be a fixed state field. :thinking:

crates/asm/common/src/state.rs
```diff
@@ -41,12 +41,12 @@ pub struct ChainViewState {
     ///
     /// Each leaf represents the root hash of an [`AsmManifest`](crate::AsmManifest) for the
     /// corresponding block, enabling efficient historical proofs of ASM state transitions.
-    pub manifest_mmr: AsmCompactMmr,
+    pub manifest_mmr: AsmManifestMmr,
```
delbonis: Maybe should call this `asm_manifest_mmr_state` (and update type name) to be clear it's not *just* the MMR.  Or `asm_history_accumulator_state`?

crates/asm/common/src/mmr.rs
prajwolrg: We could do it that way, but that means we'd have to pass part of `AsmSpec` everytime. I like it when it's closer together.

delbonis: Sure, but not if the alternative makes us have an extra state field.

prajwolrg: Created a separate [ticket](https://alpenlabs.atlassian.net/browse/STR-2111) to discuss this.

prajwolrg: Since the `Params` are defined by the subprotocol and the accumulator lives outside of the subprotocols, it is not exactly clear how to do this.

PR #1285, opened 2026-01-14
STR-1880: EE statediffs realigned to use da-framework

evgenyzdanovich:
## Description

- statediffs MVP revisited, fully aligned to use `da-framework` and `Codec`
- proper diffs across blocks (previously it was set-value only)
- **additional optimizations**: trimming zeroes from smart contract storage values, storing balance as diff supported by the custom schme for the counter.
- bugfix of `new_from_spec` method
- basic fntests to validate it's not awfully broken
- some structuring of the crate, separating persistence layer (`/block`) from actual "diffing and da" layer (`/batch`)

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Contains changes from https://github.com/alpenlabs/alpen/pull/1299, to be rebased once it hits main

- [ ]  Yes
- [x]  No

## Checklist

### DIFF
--- a/crates/reth/rpc/src/rpc.rs
+++ b/crates/reth/rpc/src/rpc.rs
@@ -59,7 +59,8 @@ where
         let mut state = StateReconstructor::from_chain_spec("dev")
             .map_err(to_jsonrpsee_error("Can't initialize reconstructed state"))?;
 
-        // Apply each block's diff sequentially
+        // Aggregate all block diffs into a single BatchStateDiff
+        let mut builder = BatchBuilder::new();
         for i in 1..=block_number {
             let block_diff = self
                 .db
@@ -67,13 +68,7 @@ where
                 .map_err(to_jsonrpsee_error("Failed fetching block state diff"))?;
 
             match block_diff {
-                Some(diff) => {
-                    // Convert BlockStateChanges -> BatchStateDiff for apply_diff
-                    let batch_diff = BatchStateDiff::from(&diff);
-                    state
-                        .apply_diff(&batch_diff)
-                        .map_err(to_jsonrpsee_error("Error while applying state diff"))?;
-                }
+                Some(diff) => builder.apply_block(&diff),
                 None => {
                     return RpcResult::Err(to_jsonrpsee_error_object(
                         Some("missing_diff"),
@@ -82,6 +77,12 @@ where
                 }
             }
         }
+
+        // Apply aggregated diff once
+        let batch_diff = builder.build();
+        state
+            .apply_diff(&batch_diff)
+            .map_err(to_jsonrpsee_error("Error while applying state diff"))?;
 
         RpcResult::Ok(Some(state.state_root()))
     }
--- a/crates/reth/statediff/src/batch/account.rs
+++ b/crates/reth/statediff/src/batch/account.rs
@@ -19,12 +19,20 @@ use crate::{
 /// - `nonce`: Counter (signed delta, varint-encoded)
 /// - `code_hash`: Register (only changes on contract creation)
 ///
+/// # Delta vs Full Value Replacement
+///
+/// **Full value replacement** (register): Store the new value directly.
+/// - Example: balance changes from 1000 to 1005 → encode `1005` (full 32 bytes for U256)
+///
+/// **Delta encoding** (counter): Store only the difference between old and new.
+/// - Example: balance changes from 1000 to 1005 → encode `+5` (1 byte for sign + 1 byte for value)
+///
 /// # Why signed deltas?
 ///
-/// **Balance**: Using delta encoding instead of full value replacement provides significant
-/// space savings for typical transactions where balance changes by small amounts (gas fees,
-/// small transfers) compared to the total balance. The signed delta supports both increases
-/// (deposits, rewards) and decreases (transfers, fees).
+/// **Balance**: Delta encoding provides significant space savings for typical transactions
+/// where balance changes by small amounts (gas fees, small transfers) compared to the total
+/// balance. The signed delta supports both increases (deposits, rewards) and decreases
+/// (transfers, fees).
 ///
 /// **Nonce**: Post-Shanghai, account nonces can effectively decrease via the selfdestruct +
 /// recreate pattern: when a contract selfdestructs and is recreated in the same block (or batch),
@@ -71,7 +79,7 @@ fn balance_delta_to_counter(old: U256, new: U256) -> DaCounter<CtrU256BySignedU2
     if old == new {
         return DaCounter::new_unchanged();
     }
-    let delta = if new >= old {
+    let delta = if new > old {
         SignedU256Delta::positive(new - old)
     } else {
         SignedU256Delta::negative(old - new)
@@ -101,7 +109,7 @@ impl AccountDiff {
         }
     }
 
-    /// Creates a diff from block-level account states.
+    /// Creates a diff from two point-in-time account state ([`AccountSnapshot`]).
     ///
     /// If `original` is None, all fields are treated as changed (account creation).
     /// Returns None if no fields changed or if the nonce delta is invalid.
@@ -224,7 +232,7 @@ mod tests {
 
         // Balance delta should be +1000 (from 0)
         let balance_diff = decoded.balance.diff().unwrap();
-        assert!(balance_diff.is_positive());
+        assert!(balance_diff.is_nonnegative());
         assert_eq!(balance_diff.magnitude(), U256::from(1000));
         assert_eq!(decoded.nonce.diff().map(|v| v.inner()), Some(1));
         assert_eq!(
@@ -346,7 +354,7 @@ mod tests {
 
         // Balance delta should be negative (decrease of 1000)
         let balance_delta = diff.balance.diff().unwrap();
-        assert!(!balance_delta.is_positive());
+        assert!(!balance_delta.is_nonnegative());
         assert_eq!(balance_delta.magnitude(), U256::from(1000));
 
         // Verify encoding roundtrip
@@ -354,7 +362,7 @@ mod tests {
         let decoded: AccountDiff = decode_buf_exact(&encoded).unwrap();
 
         let decoded_delta = decoded.balance.diff().unwrap();
-        assert!(!decoded_delta.is_positive());
+        assert!(!decoded_delta.is_nonnegative());
         assert_eq!(decoded_delta.magnitude(), U256::from(1000));
 
         // Verify apply works correctly
--- a/crates/reth/statediff/src/batch/builder.rs
+++ b/crates/reth/statediff/src/batch/builder.rs
@@ -56,6 +56,32 @@ impl<T> TrackedState<T> {
 /// builder.apply_block(&block2_diff);
 /// let diff = builder.build();
 /// ```
+///
+/// # Limitations
+///
+/// ## No block ordering validation
+///
+/// Currently, [`apply_block`](Self::apply_block) does not validate that blocks are applied
+/// in consecutive order. Callers must ensure blocks are applied sequentially (block N, then
+/// block N+1, etc.). Applying blocks out of order will produce incorrect diffs.
+///
+/// **TODO**: Add block number tracking and validation to reject non-consecutive blocks.
+///
+/// ## No pre-state provider
+///
+/// The builder assumes the first block diff for each account/slot contains the correct
+/// "original" value. If an account is touched in block 5 but not in blocks 1-4, the builder
+/// has no way to know the account's state before block 1.
+///
+/// This works correctly when:
+/// - Building a diff for a single block (original comes from that block's diff)
+/// - All touched accounts/slots appear in at least the first block where they're modified
+///
+/// This may produce incorrect "Created vs Updated" classification when:
+/// - An account existed before the batch but is first touched in a later block
+///
+/// **TODO**: Add an optional pre-state provider trait that the builder can query for the
+/// true original state of any account/slot when first encountered within the batch.
 #[derive(Clone, Debug, Default)]
 pub struct BatchBuilder {
     /// Account states: address -> tracked state (original is None if account didn't exist).
--- a/crates/reth/statediff/src/codec.rs
+++ b/crates/reth/statediff/src/codec.rs
@@ -253,11 +253,17 @@ impl SignedU256Delta {
     }
 
     /// Creates a positive delta (balance increase).
+    ///
+    /// Note: Zero magnitude is normalized - the returned delta will have `is_nonnegative() == true`
+    /// regardless, since zero has no meaningful sign.
     pub fn positive(magnitude: U256) -> Self {
         Self::new(true, magnitude)
     }
 
     /// Creates a negative delta (balance decrease).
+    ///
+    /// Note: Zero magnitude is normalized to a nonnegative zero - if `magnitude.is_zero()`,
+    /// the returned delta will have `is_nonnegative() == true` since zero has no meaningful sign.
     pub fn negative(magnitude: U256) -> Self {
         Self::new(false, magnitude)
     }
@@ -267,8 +273,8 @@ impl SignedU256Delta {
         self.magnitude.is_zero()
     }
 
-    /// Returns the sign (true = positive, false = negative).
-    pub fn is_positive(&self) -> bool {
+    /// Returns true if this delta is non-negative (positive or zero).
+    pub fn is_nonnegative(&self) -> bool {
         self.positive
     }
 
@@ -491,7 +497,7 @@ mod tests {
     fn test_signed_u256_delta_zero() {
         let delta = SignedU256Delta::default();
         assert!(delta.is_zero());
-        assert!(delta.is_positive()); // Zero is normalized to positive
+        assert!(delta.is_nonnegative()); // Zero is normalized to positive
 
         let encoded = encode_to_vec(&delta).unwrap();
         assert_eq!(encoded, vec![0x00]); // Single byte for zero
@@ -504,7 +510,7 @@ mod tests {
     fn test_signed_u256_delta_positive_small() {
         let delta = SignedU256Delta::positive(U256::from(100u8));
         assert!(!delta.is_zero());
-        assert!(delta.is_positive());
+        assert!(delta.is_nonnegative());
         assert_eq!(delta.magnitude(), U256::from(100u8));
 
         let encoded = encode_to_vec(&delta).unwrap();
@@ -513,14 +519,14 @@ mod tests {
 
         let decoded: SignedU256Delta = decode_buf_exact(&encoded).unwrap();
         assert_eq!(decoded.magnitude(), U256::from(100u8));
-        assert!(decoded.is_positive());
+        assert!(decoded.is_nonnegative());
     }
 
     #[test]
     fn test_signed_u256_delta_negative_small() {
         let delta = SignedU256Delta::negative(U256::from(50u8));
         assert!(!delta.is_zero());
-        assert!(!delta.is_positive());
+        assert!(!delta.is_nonnegative());
         assert_eq!(delta.magnitude(), U256::from(50u8));
 
         let encoded = encode_to_vec(&delta).unwrap();
@@ -529,7 +535,7 @@ mod tests {
 
         let decoded: SignedU256Delta = decode_buf_exact(&encoded).unwrap();
         assert_eq!(decoded.magnitude(), U256::from(50u8));
-        assert!(!decoded.is_positive());
+        assert!(!decoded.is_nonnegative());
     }
 
     #[test]
@@ -547,7 +553,7 @@ mod tests {
 
         let decoded: SignedU256Delta = decode_buf_exact(&encoded).unwrap();
         assert_eq!(decoded.magnitude(), large);
-        assert!(decoded.is_positive());
+        assert!(decoded.is_nonnegative());
     }
 
     #[test]
@@ -555,7 +561,7 @@ mod tests {
         // Creating a "negative zero" should normalize to positive zero
         let delta = SignedU256Delta::new(false, U256::ZERO);
         assert!(delta.is_zero());
-        assert!(delta.is_positive()); // Normalized
+        assert!(delta.is_nonnegative()); // Normalized
     }
 
     // ==================== CtrU256BySignedU256 Tests ====================
@@ -566,7 +572,7 @@ mod tests {
         let b = U256::from(150u8);
 
         let delta = CtrU256BySignedU256::compare(a, b).unwrap();
-        assert!(delta.is_positive());
+        assert!(delta.is_nonnegative());
         assert_eq!(delta.magnitude(), U256::from(50u8));
     }
 
@@ -576,7 +582,7 @@ mod tests {
         let b = U256::from(100u8);
 
         let delta = CtrU256BySignedU256::compare(a, b).unwrap();
-        assert!(!delta.is_positive());
+        assert!(!delta.is_nonnegative());
         assert_eq!(delta.magnitude(), U256::from(50u8));
     }
 
--- a/crates/reth/statediff/src/serde_impl.rs
+++ b/crates/reth/statediff/src/serde_impl.rs
@@ -44,7 +44,7 @@ impl From<&AccountDiff> for AccountDiffSerde {
     fn from(diff: &AccountDiff) -> Self {
         Self {
             balance_delta: diff.balance.diff().map(|d| BalanceDeltaSerde {
-                positive: d.is_positive(),
+                positive: d.is_nonnegative(),
                 magnitude: d.magnitude(),
             }),
             nonce_delta: diff.nonce.diff().map(|v| v.inner()),
@@ -177,7 +177,7 @@ mod tests {
         // Convert back
         let roundtrip: AccountDiff = serde.into();
         let roundtrip_delta = roundtrip.balance.diff().unwrap();
-        assert!(roundtrip_delta.is_positive());
+        assert!(roundtrip_delta.is_nonnegative());
         assert_eq!(roundtrip_delta.magnitude(), U256::from(1000));
         assert_eq!(roundtrip.nonce.diff().map(|v| v.inner()), Some(5));
         assert_eq!(
--- a/functional-tests/tests/el_stateroot_from_state_diffs.py
+++ b/functional-tests/tests/el_stateroot_from_state_diffs.py
@@ -24,8 +24,13 @@ class ElBlockStateDiffDataGenerationTest(testenv.StrataTestBase):
         block = get_latest_eth_block_number(rethrpc)
         self.info(f"Latest reth block={block}")
 
+        # Wait for state diff to be available (race: block may exist before diff is stored)
+        block_info = rethrpc.eth_getBlockByNumber(hex(block), False)
+        block_hash = block_info["hash"]
+        reth_waiter.wait_until_state_diff_at_blockhash(block_hash)
+
         reconstructed_root = rethrpc.strataee_getStateRootByDiffs(block)
-        actual_root = rethrpc.eth_getBlockByNumber(hex(block), False)["stateRoot"]
+        actual_root = block_info["stateRoot"]
         self.info(f"reconstructed state root = {reconstructed_root}")
         self.info(f"actual state root = {actual_root}")
 


crates/reth/statediff/src/batch/account.rs
```diff
@@ -0,0 +1,270 @@
...
+///
+/// - `balance`: Register (can change arbitrarily)
+/// - `nonce`: Stored as `Option<u8>` (nonces typically increment by small amounts)
+/// - `code_hash`: Register (only changes on contract creation)
+#[derive(Clone, Debug, Default)]
+pub struct AccountDiff {
+    /// Balance change (full replacement if changed).
+    pub balance: DaRegister<CodecU256>,
+    /// Nonce increment (None = unchanged, Some(n) = increment by n).
+    pub nonce_incr: Option<u8>,
```
delbonis: This would be a counter DA type.

crates/reth/statediff/src/batch/account.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    /// Returns true if no changes are recorded.
+    pub fn is_unchanged(&self) -> bool {
+        DaWrite::is_default(&self.balance)
+            && self.nonce_incr.is_none()
+            && DaWrite::is_default(&self.code_hash)
+    }
+}
+
+impl Codec for AccountDiff {
```
delbonis: This is kinda manually reimplementing what the compound macros were for.

functional-tests/tests/el_batch_state_diff.py
```diff
@@ -0,0 +1,51 @@
...
+
+        assert actual_root == reconstructed_root, (
+            f"State root mismatch! Actual: {actual_root}, Reconstructed: {reconstructed_root}"
+        )
+
+        # Validate batch diff is not empty
+        assert batch_diff is not None, "Batch diff should not be None"
+        assert len(batch_diff.get("accounts", {})) > 0, "Batch diff should have account changes"
+
+        self.info("Batch state diff test passed!")
```
delbonis: We're getting rid of these `self.info` logging fns, it turned out not to be as good of an idea as it was thought to be, just use `logging.info`.

delbonis (changes requested):
So this looks *reasonably* okay, although it's kinda bodging the DA primitives in a few places like with accounts.

However, I didn't expect the DA framework to be used for the EVM EE, it was only designed with the OL state in mind.  This isn't to say that it *can't* be used, although I didn't make any considerations for it, so you shouldn't shy away from doing a little more iteration/extension on it if needed.

crates/reth/rpc/src/lib.rs
```diff
@@ -25,13 +25,21 @@ pub trait StrataRpcApi {
...
+    /// Returns the state diff for the block in DA-optimized format.
     #[method(name = "getBlockStateDiff")]
-    fn get_block_state_diff(&self, block_hash: B256) -> RpcResult<Option<BlockStateDiff>>;
+    fn get_block_state_diff(&self, block_hash: B256) -> RpcResult<Option<BatchStateDiffSerde>>;
 
     /// Returns the state root for the block_number as reconstructured from the state diffs.
     #[method(name = "getStateRootByDiffs")]
     fn get_state_root_via_diffs(&self, block_number: u64) -> RpcResult<Option<B256>>;
+
+    /// Returns the aggregated state diff for a range of blocks in DA-optimized format.
```
barakshani: Do we have any other format than this one? If not, I'd say that specifying "DA-optimized" is not really needed, and adds a bit of confusion: I am assuming that the format is the actual bit difference than the previous value, but that's because I already have prior knowledge, otherwise it won't be clear to me.

crates/reth/rpc/src/rpc.rs
```diff
@@ -44,39 +44,81 @@ where
...
-    fn get_block_state_diff(&self, block_hash: B256) -> RpcResult<Option<BlockStateDiff>> {
-        self.db
+    fn get_block_state_diff(&self, block_hash: B256) -> RpcResult<Option<BatchStateDiffSerde>> {
+        let block_diff = self
+            .db
             .get_state_diff_by_hash(block_hash)
-            .map_err(to_jsonrpsee_error("Failed fetching block state diff"))
+            .map_err(to_jsonrpsee_error("Failed fetching block state diff"))?;
+
+        // Convert BlockStateDiff -> BatchStateDiff -> BatchStateDiffSerde
```
barakshani: ~ok, so from here it seems that we do have a "non-optimised" format ?~

crates/reth/rpc/src/lib.rs
```diff
@@ -25,13 +25,21 @@ pub trait StrataRpcApi {
         json: Option<bool>,
     ) -> RpcResult<Option<BlockWitness>>;
 
-    /// Returns the state diff for the block.
+    /// Returns the state diff for the block in DA-optimized format.
     #[method(name = "getBlockStateDiff")]
-    fn get_block_state_diff(&self, block_hash: B256) -> RpcResult<Option<BlockStateDiff>>;
+    fn get_block_state_diff(&self, block_hash: B256) -> RpcResult<Option<BatchStateDiffSerde>>;
```
barakshani: get_**block** function returns a batch output.

crates/reth/rpc/src/rpc.rs
```diff
@@ -44,39 +44,81 @@ where
...
+        Ok(block_diff.map(|diff| BatchStateDiffSerde::from(BatchStateDiff::from(&diff))))
     }
 
     fn get_state_root_via_diffs(&self, block_number: u64) -> RpcResult<Option<B256>> {
```
barakshani, line 57: It's not clear from the function name that we apply state diffs on genesis state, and I also assume that we'll have a more generic function that allows us to apply it on _any_ state?
Can we change the function name to something like _apply_diffs_on_genesis_? Or maybe just have a general function that takes a state, and we can indicate that the state should be genesis (when applicable)?

crates/reth/statediff/src/batch/account.rs
```diff
@@ -0,0 +1,270 @@
...
+    /// Nonce increment (None = unchanged, Some(n) = increment by n).
+    pub nonce_incr: Option<u8>,
+    /// Code hash change (only on contract creation).
+    pub code_hash: DaRegister<CodecB256>,
+}
+
+/// Converts a nonce delta to `Option<u8>`, panicking if it exceeds `u8::MAX`.
+///
+/// In practice, a single batch should never have more than 255 transactions
+/// from the same account.
```
barakshani: what do you mean by that? Like _unlikely_? I can see cases (see arbitrage bots, oracles) that have frequent updates.

barakshani: I'm uncomfortable with panicking here. Can we make this variable length?

crates/reth/statediff/src/batch/storage.rs
```diff
@@ -0,0 +1,130 @@
...
+use serde::{Deserialize, Serialize};
+use strata_codec::{Codec, CodecError, Decoder, Encoder};
+
+/// Diff for storage slots of an account.
+///
+/// Uses a sorted map for deterministic encoding.
+/// Each slot value is encoded as a register (full replacement).
+#[derive(Clone, Debug, Default)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct StorageDiff {
```
barakshani, line 18: is this what we'll be using for changes in contract values?

crates/reth/statediff/src/batch/account.rs
evgenyzdanovich: yup, agree. dev is cooking. tho requires a bit of work and ideally proper support in `da-framework` crate. I'm exactly on it right now.

crates/reth/statediff/src/batch/account.rs
evgenyzdanovich: agree, requires a bit of changes in `da-framework` tho, as you pointed out. 
I'm doing a "da-framework-enhancements" PR as a prereq work, will ask for review soon. then I'll use it here.

crates/reth/statediff/src/batch/account.rs
```diff
@@ -0,0 +1,270 @@
...
+fn checked_nonce_incr(delta: u64, addr: Address) -> Option<u8> {
+    if delta == 0 {
+        return None;
+    }
+    assert!(
+        delta <= u8::MAX as u64,
+        "nonce delta {delta} exceeds u8::MAX for account {addr}",
+    );
```
storopoli: Code panics if nonce `delta > 255` within a batch.
Very unlikely, but this can be triggered.
Would crash the program during state diff processing

crates/reth/rpc/src/rpc.rs
```diff
@@ -44,39 +44,81 @@ where
...
-        let mut builder = BatchStateDiffBuilder::new();
+        // Initialize state from genesis
+        let mut state = StateReconstructor::from_chain_spec("dev")
```
storopoli, line 59: Hardcoded `"dev"` Chain Spec

crates/reth/rpc/src/rpc.rs
```diff
@@ -44,39 +44,81 @@ where
...
+
+            match block_diff {
+                Some(diff) => builder.apply_block(&diff),
+                None => {
+                    return RpcResult::Err(to_jsonrpsee_error_object(
+                        Some("missing_diff"),
+                        &format!("state diff missing for block {i}"),
+                    ));
+                }
+            }
```
storopoli, line 119: DoS vector: No limit on block range in RPC can be from `0` to `u64::MAX`

crates/reth/statediff/src/batch/builder.rs
```diff
@@ -0,0 +1,322 @@
...
+                slot_entry.1 = *current;
+            }
+        }
+
+        // Collect deployed contract code hashes
+        for hash in &block_diff.deployed_code_hashes {
+            if !self.deployed_code_hashes.contains(hash) {
+                self.deployed_code_hashes.push(*hash);
+            }
+        }
```
storopoli: The field `deployed_code_hashes` is a `Vec<B256>` (line 37). When checking for duplicates with `Vec::contains()`, it does a linear scan through all existing elements - O(n) per lookup. If you're adding `m` hashes across all blocks, and there are already `n` unique hashes, this becomes O(n*m) in the worst case.

A `HashSet<B256>` would give O(1) lookups, making the overall dedup O(m).

Do we really need `deployed_code_hashes` to be a `Vec`?

I also don't like the deduplication is only enforced through a docstring, I would much prefer to have it enforced through the type system.

evgenyzdanovich:
@delbonis I think da-framework is general enough to be used in EE statediffs and I'm willing to extend it a bit: will create a separate PR shortly, then we continue there (with this one in mind).

crates/reth/statediff/src/batch/account.rs
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    /// Creates a diff representing a new account creation.
+    pub fn new_created(balance: U256, nonce: u64, code_hash: B256) -> Self {
+        Self {
+            balance: DaRegister::new_set(CodecU256(balance)),
+            nonce_incr: if nonce > 0 { Some(nonce as u8) } else { None },
+            code_hash: DaRegister::new_set(CodecB256(code_hash)),
+        }
+    }
```
storopoli, line 110: `new_created` truncates `nonce >255` without a check (`nonce as u8`). Use the same checked path as `checked_nonce_incr`.

crates/reth/statediff/src/batch/account.rs
```diff
@@ -0,0 +1,270 @@
...
+        let (orig_balance, orig_nonce, orig_code_hash) = original
+            .map(|o| (Some(o.balance), o.nonce, Some(o.code_hash)))
+            .unwrap_or((None, 0, None));
+
+        let balance = match orig_balance {
+            Some(ob) if ob == current.balance => DaRegister::new_unset(),
+            _ => DaRegister::new_set(CodecU256(current.balance)),
+        };
+
+        let nonce_delta = current.nonce.saturating_sub(orig_nonce);
```
storopoli: NOTE: This is impossible post‑Shanghai (EIP‑6780). I don't know if we should check for that or how we're going to do it.

Nonce decreases are silently dropped. This uses `saturating_sub`, so if an account’s nonce decreases (e.g., selfdestruct + recreate in the same batch), the delta becomes `0` and the change is omitted; reconstruction then keeps the old nonce, yielding a wrong state root.

crates/reth/rpc/src/rpc.rs
```diff
@@ -44,39 +44,81 @@ where
...
 
     fn get_state_root_via_diffs(&self, block_number: u64) -> RpcResult<Option<B256>> {
-        let mut builder = BatchStateDiffBuilder::new();
+        // Initialize state from genesis
+        let mut state = StateReconstructor::from_chain_spec("dev")
+            .map_err(to_jsonrpsee_error("Can't initialize reconstructed state"))?;
 
-        // First construct the batch state diff consisting of all the changes so far.
+        // Apply each block's diff sequentially
         for i in 1..=block_number {
```
storopoli, line 64: `for i in 1..=block_number { ... }` means a caller can pass a very large `block_number` and force O(n) DB reads + reconstruction.

storopoli (changes requested):
Another pass

evgenyzdanovich:
@storopoli @barakshani @delbonis 
I squeezed changes to `da-framework` in a separate PR according to the feedback here: https://github.com/alpenlabs/alpen/pull/1299

let's review that PR first, then I rebase and we continue here

crates/reth/statediff/src/batch/account.rs
evgenyzdanovich: addressed in the newest revisions

crates/reth/statediff/src/batch/account.rs
evgenyzdanovich: resolved, PTAL

crates/reth/statediff/src/batch/account.rs
evgenyzdanovich: addressed, PTAL

crates/reth/statediff/src/batch/account.rs
evgenyzdanovich: addressed, PTAL

crates/reth/rpc/src/lib.rs
evgenyzdanovich: removed the emphasis on DA-optimized, I think the only mention is in lib.rs where it's applicable and the context is clear.

crates/reth/rpc/src/rpc.rs
evgenyzdanovich: yes, correct, in DB we store "originals" to support proper diffs over a range of blocks.

functional-tests/tests/el_batch_state_diff.py
evgenyzdanovich: done

crates/reth/rpc/src/rpc.rs
evgenyzdanovich: I documented in the `lib.rs` that those methods are intended to be used for testing now (applicable to this comment and other similar comments in `rpc.rs`). 

The purpose of this PR is not to polish state_root reconstruction from diffs, it's just a gate for fntests for now. 
With that in mind it actually makes sense to have a hardcoded dev (this is what we use there by default and nothing else at the moment). 

Lastly. Previously, it was broken because no genesis spec has been taken into account and we were reconstructing from an empty state. So in a way, the hardcoded value repairs it and it makes more sense for now.

crates/reth/rpc/src/rpc.rs
evgenyzdanovich: I documented in the `lib.rs` that those methods are intended to be used for testing now (applicable to this comment and other similar comments in `rpc.rs`).

crates/reth/rpc/src/rpc.rs
evgenyzdanovich: I documented in the `lib.rs` that those methods are intended to be used for testing now (applicable to this comment and other similar comments in `rpc.rs`).

crates/reth/rpc/src/rpc.rs
evgenyzdanovich: addressed it.

rn: 
- getStateDiffForBlock -> BatchStateDiff (da diff for a single block)
- getStateDiffForRange -> BatchStateDiff
- getStateRootByDiffs -> state root bytes 

Note these methods are now used only in tests and will be reworked soon(-ish).

crates/reth/statediff/src/batch/storage.rs
evgenyzdanovich: yes, correct. 

btw, I plan to push one more optimization for those values (trimming zeroes) soon.

evgenyzdanovich: Related, I propose that this is what we are gonna store.

[feat(statediff): implement trimming zeroes encoding for u256 storage …](https://github.com/alpenlabs/alpen/pull/1285/commits/261d1a960e432274dbf2e153c0265f202d7cd22e) 
[261d1a9](https://github.com/alpenlabs/alpen/pull/1285/commits/261d1a960e432274dbf2e153c0265f202d7cd22e)

delbonis: I had considered a version where if only a few bytes change we can encode it with a u32 bitmap of the bytes that changed and followed by *just* those bytes.  Similarly we can special-case things like where the lower 20 bytes are set, like storage slots that just contain an address, so we can avoid overheads there too.

Probably out of scope for this PR.

crates/reth/rpc/src/rpc.rs
```diff
@@ -44,39 +44,81 @@ where
         res.map_err(to_jsonrpsee_error("Failed fetching witness"))
     }
 
-    fn get_block_state_diff(&self, block_hash: B256) -> RpcResult<Option<BlockStateDiff>> {
-        self.db
+    fn get_state_diff_for_block(&self, block_hash: B256) -> RpcResult<Option<BatchStateDiffSerde>> {
```
barakshani, line 47: This should be BlockStateDiffSerde (which I don't think we have), no? If you end up using BatchStateDiffSerde, maybe comment why.

crates/reth/rpc/src/rpc.rs
barakshani: >getStateDiffForBlock -> BlockStateDiff (da diff for a single block)

Why not BatchStateDiff?

crates/reth/statediff/src/batch/account.rs
barakshani: I think this is only nonce diff _calculation_: diff = new_nonce minus orig_nonce.

crates/reth/statediff/src/block/diff.rs
```diff
@@ -0,0 +1,138 @@
...
+use super::{AccountSnapshot, BlockAccountChange, BlockStorageDiff};
+
+/// Per-block state diff stored in DB by exex.
+///
+/// Contains both original and current values to enable proper batch aggregation
+/// with revert detection when building [`BatchStateDiff`](crate::batch::BatchStateDiff).
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct BlockStateDiff {
+    /// Account changes with original values for tracking.
+    pub accounts: BTreeMap<Address, BlockAccountChange>,
```
barakshani, line 18: I'm not sure where's the right place to ask, so I'll ask here: BlockAccountChange is a tuple (before, after), and I wonder why do we need in BlockStateDiff to have the previous value, and not just a map from account to last value.

crates/reth/statediff/src/batch/builder.rs
```diff
@@ -0,0 +1,322 @@
...
+/// # Example
+///
+/// ```ignore
+/// let mut builder = BatchBuilder::new();
+/// builder.apply_block(&block1_diff);
+/// builder.apply_block(&block2_diff);
+/// let diff = builder.build();
+/// ```
+#[derive(Clone, Debug, Default)]
+pub struct BatchBuilder {
```
barakshani, line 86: I'm not sure how exactly this is being accessed, but wouldn't you want to add a field for block number so that each time `apply_block` is called we can first check that we build on the correct (previous) state?

crates/reth/statediff/src/batch/builder.rs
```diff
@@ -0,0 +1,322 @@
...
+    /// Account states: address -> (original_before_batch, current_state).
+    /// Original is None if account didn't exist before the batch started.
+    accounts: BTreeMap<Address, (Option<AccountSnapshot>, Option<AccountSnapshot>)>,
```
barakshani: Do we initiate a new BatchBuilder per batch? I wonder when do we set, or change, the original_before_batch value.

crates/reth/statediff/src/lib.rs
```diff
@@ -1,311 +1,50 @@
...
+//! - [`batch`]: DA-optimized batch diff types (compact, no originals), actually posted on-chain
+//! - `reconstruct`: State reconstruction from diffs (see [`StateReconstructor`]), currently
+//!   experimental and used only in tests - will be adjusted later (for syncing from diffs).
+//!
+//! # Key Types
+//!
+//! | Type | Module | Purpose |
+//! |------|--------|---------|
+//! | [`BlockStateDiff`] | `block` | Per-block diff for DB storage |
+//! | [`BatchStateDiff`] | `batch` | Aggregated diff for DA |
```
barakshani, line 28: How does BatchStateDiffSerde fit in here, if at all?

crates/reth/statediff/src/batch/account.rs
```diff
@@ -0,0 +1,260 @@
...
+        original: Option<&AccountSnapshot>,
+        _addr: Address,
+    ) -> Option<Self> {
+        let (orig_balance, orig_nonce, orig_code_hash) = original
+            .map(|o| (Some(o.balance), o.nonce, Some(o.code_hash)))
+            .unwrap_or((None, 0, None));
+
+        let balance = match orig_balance {
+            Some(ob) if ob == current.balance => DaRegister::new_unset(),
+            _ => DaRegister::new_set(CodecU256(current.balance)),
```
barakshani: should we take `current.balance - ob`?

crates/reth/statediff/src/reconstruct.rs
```diff
@@ -0,0 +1,204 @@
...
+                        .get_rlp(&acc_info_trie_path)
+                        .unwrap_or_default();
+
+                    // Build snapshot from current state and apply diff
+                    let mut snapshot = current
+                        .as_ref()
+                        .map(AccountSnapshot::from)
+                        .unwrap_or_default();
+
+                    account_diff.apply(&mut snapshot)?;
```
barakshani, line 97: here we should only apply the delta right -- can you direct me to this function?

crates/reth/statediff/src/codec.rs
```diff
@@ -0,0 +1,350 @@
...
+        enc.write_buf(&[len])?;
+        if len > 0 {
+            enc.write_buf(&bytes[start..])?;
+        }
+        Ok(())
+    }
+
+    fn decode(dec: &mut impl Decoder) -> Result<Self, CodecError> {
+        let mut len_buf = [0u8; 1];
+        dec.read_buf(&mut len_buf)?;
```
delbonis: I thought I had a `read_arr` you could call like `dec.read_arr::<1>()?`?

crates/reth/statediff/src/reconstruct.rs
```diff
@@ -0,0 +1,204 @@
...
+    block::AccountSnapshot,
+};
+
+/// Error that may occur during state reconstruction.
+#[derive(Debug, ThisError)]
+pub enum ReconstructError {
+    #[error("MPT error: {0}")]
+    Mpt(#[from] strata_mpt::Error),
+    #[error("DA apply error: {0}")]
+    Da(#[from] strata_da_framework::DaError),
```
delbonis: `error` is redundant in these error strings

crates/reth/statediff/src/batch/builder.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    /// Storage states: address -> slot -> (original_value, current_value).
+    storage: BTreeMap<Address, BTreeMap<U256, (U256, U256)>>,
```
delbonis: For this (and `accounts`) I'd think it'd improve clarify to have a type that captures "the old state and the new state" being tracked in these tables.

crates/reth/statediff/src/batch/diff.rs
```diff
@@ -0,0 +1,147 @@
...
+        // Encode deployed code hashes
+        (self.deployed_code_hashes.len() as u32).encode(enc)?;
+        for hash in &self.deployed_code_hashes {
+            enc.write_buf(hash.as_slice())?;
+        }
+
+        Ok(())
+    }
+
+    fn decode(dec: &mut impl Decoder) -> Result<Self, CodecError> {
```
delbonis, line 46: This probably doesn't matter since I can't imagine what could go wrong, but do you think it might make sense to enforce that entries are ordered by key?

Also, this and in `encode`, we could probably define helpers for encoding/decoding btreemaps that contain codecable keys/values.  Also would be helpful in a bunch of other places too probably.

crates/reth/statediff/src/batch/account.rs
evgenyzdanovich: Jose is exactly right on this point, thanks for catching it: im fixing the counter schema for nonces to include (potentially) negative increments

crates/reth/rpc/src/rpc.rs
evgenyzdanovich: Here and elsewhere, I see some confusion with the naming:
serde structs are only used to return something from RPC endpoints (returrn type should use `serde_json`). 
Serde counterparts are equal to non-serde (e.g. BatchStateDiff vs BatchStateDiff**Serde**), represent the same data, but only used in RPC in tests at the moment. 

On the original question: BatchStateDiff is a "public" representation of state diff for several blocks. It's also perfectly valid to have batch of size=1 (single block) and still can be represented as batched diff.  Here, the function just returns the state diff for a single block.

crates/reth/statediff/src/reconstruct.rs
evgenyzdanovich: this is literally the line you are at, you found it!

`account_diff.apply(&mut snapshot)?;`

`account_diff` contains "diffed" values (in the public representation, **this is what we actually deserialize into from "chunks"**) and diffs are applied "into" the account `snapshot` (which is just point-in-time-view of account state).
In the end (after applying diffs iteratively, potentially from several batches) this way, diffs are accumuted into this account snapshot struct and will contain the actual reconstructed account state.

crates/reth/statediff/src/lib.rs
evgenyzdanovich: Here and elsewhere, I see some confusion with the naming:
serde structs are only used to return something from RPC endpoints (returrn type should use serde_json).
Serde counterparts are equal to non-serde (e.g. BatchStateDiff vs BatchStateDiffSerde), represent the same data, but only used in RPC in tests at the moment.

I'll mention it here (given the feedback on the types being confusing).

crates/reth/statediff/src/batch/account.rs
evgenyzdanovich: This is a "register", not a "counter", so I guess the impl is correct and this is the intention for now.

Note: in the future, I envision balances to be encoded as a ` Counter < Delta > `, where Delta represents the counter schema: for such a case `Delta` can be backed by "zigzag" encoding ([HELLO HATED PROTOBUF](https://protobuf.dev/programming-guides/encoding/#signed-ints)) + VarInt and we can take diffs. 

The idea behind that is that all the balance changes within batch(es) (_excluding fees_) can be represented as "subbing from one balance and adding to another", so in a way we'd like to encode positive and negative deltas "symmetrically". 

**But it's to be researched, agreed upon, compared vs other approaches and done at a later stage, as a part of optimizing EE DA costs.**

crates/reth/statediff/src/reconstruct.rs
barakshani: haha, yeah I know it's here, but my IDE can't find declaration of the `apply` function for some reason, so that is what I was asking for. I am interested to understand what diff "type" we apply (like I mention elsewhere).

crates/reth/statediff/src/batch/account.rs
barakshani: I am trying to ask about the diff "format" (or "type", as I just called in a different comment): I thought we are taking the actual diffs between values, so here I meant saying current_balance - orig_balance, regardless of types. It seemed to me that here we just that the current balance.

crates/reth/statediff/src/reconstruct.rs
evgenyzdanovich: aha, I now understand you fully.

>but my IDE can't find declaration of the apply function for some reason

I guess I know the reason. it is because of `macro_rule` in the code path you are navigating (we even have a [TODO](https://github.com/alpenlabs/alpen/blob/main/crates/da-framework/src/compound.rs#L166) to fix that, with proc macro there shouldnt be any issues in the IDE).

On your question: the `apply` is a part of da-framework and it allows for `DaWrite` (diff in the context of conversation) type to be "applied" on top of `Target` type. The "Account" is declared as a compound type (over `da-framework` primitives such as counters and registers), so each field (being a `DaWrite` themselves) will be "applied" correspondingly (register have its own apply, so as counter).
 
N.B. The logic is defined exclusively in da-framework, not here. Particularly, the whole definition of an account (`Target` type together with diff type) can be found at batch/account.rs.

crates/reth/statediff/src/block/diff.rs
evgenyzdanovich: `block/` submodule is internal stuff, it's not exposed to the caller. That's how we store block state changes (name can be misleading)

crates/reth/statediff/src/batch/builder.rs
evgenyzdanovich: Yes, we initiate batch builder per batch. 

> I wonder when do we set, or change, the original_before_batch value.

This piece is currently missing and IMO better to be defined later (at the integration).

I confirm we'll need some thin providers trait that EE sequencer would need to implement, so that batch builder uses those providers and original_before_batch is set _dynamically in a lazy fashion_.

crates/reth/statediff/src/batch/account.rs
evgenyzdanovich: done

crates/reth/statediff/src/batch/account.rs
evgenyzdanovich: done

crates/reth/statediff/src/lib.rs
evgenyzdanovich: done

crates/reth/statediff/src/codec.rs
evgenyzdanovich: yeah, done

crates/reth/statediff/src/batch/diff.rs
evgenyzdanovich: >do you think it might make sense to enforce that entries are ordered by key?

I think iterating over `BTreeMap` specifically guarantees that iterated entries are ordered by key, am I mistaken?

Added helpers to `da-framework/codec.rs` (I think it makes the most sense there) and used it here.

crates/reth/statediff/src/batch/builder.rs
evgenyzdanovich: well, this block number is likely only be used to validate this invariant, so I guess it's redundant.

crates/reth/statediff/src/batch/builder.rs
evgenyzdanovich: done via btreeset, should be good enough.

crates/reth/statediff/src/batch/account.rs
evgenyzdanovich: made it a counter (with "diffing" - unlike register), so I assume it's done?

crates/reth/statediff/src/batch/builder.rs
barakshani: What would be the other way to making sure that we apply block on the previous block's state?

crates/reth/statediff/src/batch/account.rs
```diff
@@ -0,0 +1,366 @@
...
+
+/// Diff for a single account using DA framework primitives.
+///
+/// - `balance`: Counter (signed U256 delta, trimmed encoding)
+/// - `nonce`: Counter (signed delta, varint-encoded)
+/// - `code_hash`: Register (only changes on contract creation)
+///
+/// # Why signed deltas?
+///
+/// **Balance**: Using delta encoding instead of full value replacement provides significant
```
barakshani: I find value in, therefore also suggest, defining what "delta" and "full value replacement" means. Maybe add a small example (eg. previous=3, new=5), and how the diff looks like in each of those.

crates/reth/statediff/src/codec.rs
```diff
@@ -0,0 +1,679 @@
...
+        } else {
+            Self {
+                positive,
+                magnitude,
+            }
+        }
+    }
+
+    /// Creates a positive delta (balance increase).
+    pub fn positive(magnitude: U256) -> Self {
```
barakshani, line 259: nit: maybe call this non_nagative?

crates/reth/statediff/src/batch/account.rs
```diff
@@ -0,0 +1,366 @@
...
+}
+
+/// Converts a balance delta to `DaCounter<CtrU256BySignedU256>`.
+///
+/// Computes the signed difference between old and new balance.
+fn balance_delta_to_counter(old: U256, new: U256) -> DaCounter<CtrU256BySignedU256> {
+    if old == new {
+        return DaCounter::new_unchanged();
+    }
+    let delta = if new >= old {
```
barakshani: nit: `>` is sufficient here, also if `=` then "non_nagative" is a better term (as I commented below)

crates/reth/statediff/src/batch/account.rs
```diff
@@ -0,0 +1,366 @@
...
+        // For new accounts, balance delta is the full balance (from 0)
+        let balance_delta = SignedU256Delta::positive(balance);
+        Self {
+            balance: DaCounter::new_changed(balance_delta),
+            nonce: DaCounter::new_changed(nonce_incr),
+            code_hash: DaRegister::new_set(CodecB256(code_hash)),
+        }
+    }
+
+    /// Creates a diff from block-level account states.
```
barakshani: What do you mean by block level here? Seems to me that the function has no real context, just some start and end value.

However, since we work with delta, we must have some context here (the original value). I suppose that the "right way" to implement this is on a _batch_ level, where `delta = batch_end_value - batch_start_value`, while on every block we may just carry _block_end_ until the end (which is what we had so far, as far as I am aware).

Note: I am _not_ asking to change the implementation. Just commenting for the sake of discussion/alignment, and for more clarity.

crates/reth/statediff/src/batch/builder.rs
barakshani: If it's a tech debt, let's add a TODO/open ticket/both on this.

crates/reth/statediff/src/block/diff.rs
barakshani: I understand, but the question was more about _what justifies having the "before" value?_ Do we need to use it anywhere? I understand it's an implementation choice, I'm just curios about the necessity.

crates/reth/statediff/src/codec.rs
barakshani: I am actually not sure if this is always positive or can also take 0

crates/reth/statediff/src/reconstruct.rs
barakshani, line 1: The logic here deserves some tests. Especially I'd like to see that we apply the _delta_ state diffs correctly.

barakshani: I see that the two tests below use reconstruct (via rpc get_state_diff_for_block), but if I understand correctly rpc works on block level and not batch level.

crates/reth/rpc/src/rpc.rs
```diff
@@ -44,39 +44,81 @@ where
...
+            .db
             .get_state_diff_by_hash(block_hash)
-            .map_err(to_jsonrpsee_error("Failed fetching block state diff"))
+            .map_err(to_jsonrpsee_error("Failed fetching block state diff"))?;
+
+        // Convert BlockStateDiff -> BatchStateDiff -> BatchStateDiffSerde
+        Ok(block_diff.map(|diff| BatchStateDiffSerde::from(BatchStateDiff::from(&diff))))
     }
 
     fn get_state_root_via_diffs(&self, block_number: u64) -> RpcResult<Option<B256>> {
```
barakshani, line 57: I am not sure if you wrote it elsewhere, but is this only for tests? Computing this from genesis block can be very costly.

crates/reth/statediff/src/batch/diff.rs
delbonis: I mean when we decode them, ie enforcing that the bytes are ordered how `encode` orders it, enforcing that it's bijective with the in-memory form.

Again, probably not necessary but maybe a good idea for encoding stability reasons or something.

crates/reth/statediff/src/batch/builder.rs
evgenyzdanovich: It should be ensured by the caller at the moment, which can be done trivially: we persist block number in our EE DA tables for every `BlockStateChanges`. 
In fact, those block changes are keyed by block_hash/number in the db, but it's not a part of `crates/statediff` types.
 
Later, we'll likely define some minimal sanity checks here to validate `BatchBuilder` usage is correct. It makes sense to revise when it's actually being used in EE sequencer properly.

crates/reth/statediff/src/batch/builder.rs
evgenyzdanovich: it is not a tech debt, but rather the integration into EESequencer's reconstruction flow.

crates/reth/statediff/src/batch/builder.rs
barakshani: It seems like we agree on a need for a sanity check on this. Can you expand more about your preference not introducing it right away? I'm mostly curious about your way of thinking on this. Since we agree that it should be done, it's mainly about sequencing, so not a big deal, but I'd like to learn.

crates/reth/statediff/src/block/diff.rs
evgenyzdanovich: the necessity and goal is simple and clear: optimizing resulting onchain footprint by detecting "reverts" across changes within several blocks.

Context
---------------

Lets say we need to construct a `EEDaBatch` for three blocks (blocks number 2,3 and 4) with the following changes for simplicity (only one acc):

[1 acc.balance=4]   || (START of DaBatch)   [2 acc.balance=2] [3 acc.balance=5] [4 acc.balance=4]  (END of DaBatch) ||

here acc obviously has changed several times within a batch, but has been "reverted" to a state  _right-before-the-batch_. So for such a case, our resulting EEDaBatch should be empty.


Impl
-----------------

now, it's implemented by just storing the original values for all the necessary fields (account has its original_before_batch representation, so does every storage slot should.
With such an impl, we effectively doubled the size of persisted DA data, but its fast and simple. 

Alternative implementation could have done it without tracking original _for every change in every block_, but would require a layer of indexing over stored data to allow for queries like: "give me the exact state of this piece of data (account/storage slot) **right-before-the-batch**". 
This approach would be arguably harder to implement and the runtime efficiency would be worse (but we'd persist less metadata to supports EE DA construction).

Lastly, the impl choice can change after integration with EESequencer and reconstruction flow is complete. I do hope it makes some sense now.

evgenyzdanovich: @barakshani leaving this one for you (or somebody else) to resolve if an approach chosen is fine and the impl decision Ive taken is sane.

barakshani: Yes, that's clear and thanks for the explanation. What I had in mind is, presumably(?), a third option: we already have the original _batch_ value within BatchBuilder (how it gets there is, of course, a question, that's why the "presumably"), but here we "carry" the original value in each _block_ (I have assumed that original for block j is the same as current (last) value in block j-1, but even if not it doesn't really matter right now). So as long as we have original_before_batch, for each block we can just store the last value and in the end compute the delta. 

Note only that it's not only for cases where the original value hasn't change (end value = start value), we need to original value to compute the delta.

Anyway, I think we're good on this. I just wanted to know what's the benefit in choosing this implementation.

crates/reth/statediff/src/batch/account.rs
evgenyzdanovich: quite relevant to another comment (it contains substr "I understand, but the question was more about what justifies having the "before" value?").

it seems to me now that we converge on what's required: the fact that we ultimately need to know the value **right-before-the-batch** (for _any piece of data_, be it acc fields or storage slots). 
The means by which to achieve it are described in that comment (as well as some reasoning on why I guess another approach would be harder to impl).

barakshani: Actually this was more about the docstring than chosen logic. Writing "block level" seems to me to be too concrete. If I understand correctly the function has no notion of block, or slot, or even batch: it gets _before_ and _after_ and computes the diff. So we can apply it in other levels, not only block.

When I wrote this comment I wasn't sure if the way to compute delta-based diffs is in block level [say (account in) block 1 ends with balance 4, in block 2 balance 6 and block 3 balance 5 -- we can have vector of deltas (2,-1) so in the end we sum and gets delta of 1] or in batch level [in the same example delta is computed as 5-4=1], but after going over the rest of the code here, I have the impression we indeed do it in batch level.

crates/reth/rpc/src/rpc.rs
evgenyzdanovich: it is currently only for tests, correct, basically to validate the happy path and be able to make any changes to statediff impl (without worrying that it never worked).

"Proper" reconstruction **from pre-batch-state** is out-of-scope of this PR, as well as more test coverage for the reconstruction path.

crates/reth/statediff/src/codec.rs
evgenyzdanovich: the struct itself can carry "zero value", it's a valid representation of the struct. But `DaCounter` (a type that drtives `SignedU256Delta`) would never instantiate zero delta (`Incr=SignedU256Delta` type in the code) directly, so the code is sound.

On the question: I decided to leave this method as is, but adjust docstrings and rename `is_positive` -> `is_nonnegative` instead.

crates/reth/statediff/src/batch/account.rs
evgenyzdanovich: done

crates/reth/statediff/src/batch/account.rs
evgenyzdanovich: > Writing "block level" seems to me to be too concrete. If I understand correctly the function has no notion of block, or slot, or even batch: it gets before and after and computes the diff. So we can apply it in other levels, not only block.

got it.

/// Creates a diff from block-level account states.

renamed to 

/// Creates a diff from two point-in-time account states.

crates/reth/statediff/src/batch/account.rs
evgenyzdanovich: agree done

crates/reth/statediff/src/batch/builder.rs
evgenyzdanovich: IMO it'd seem somewhat artificial _**for now**_ to define, carry and check block_num, esp given the caller logic is not yet present.

Added an extensive section describing 2 current limitations I can see and left 2 todos. short TODO summary:
1. no block ordering validation
2. no pre-state provider

crates/reth/statediff/src/block/diff.rs
evgenyzdanovich: >  So as long as we have original_before_batch, for each block we can just store the last value and in the end compute the delta.

I totally agree. it's just tricky to do without complications and premature over-engineering without fetching and EESeq integration in-place.

> how it gets there is, of course, a question

that's indeed a big part of why I've chosen a simpler approach for now.

crates/reth/statediff/src/reconstruct.rs
evgenyzdanovich: reconstruct will be heavily refactored, I don't see much value putting the effort into it. In fact it's a "stub"/"toy" impl part of future recostruction (belongs to EE seq / node, not here)

>but if I understand correctly rpc works on block level and not batch level.

You are right, thanks. I adjusted the impl of `get_state_root_via_diffs` to properly construct batch individual blocks changes.

Generally, we now have a basic happy path that validates a `DaBatch` (that's constructed from several blocks that have some ETH txns) applied to some pre-state actually results in the same state root as canonical chain. 
IMO, at this stage, we are fine solely with this test as it is.

delbonis (approved):
looks good for now

PR #1287, opened 2026-01-15
refactor(ee-acct-runtime): chunk-based architecture

purusang:
## Description
  - Refactors `ee-acct-runtime` to support chunk-based processing for EE account updates                                 
  - Renames `UpdateX` → `UpdateTransitionX` to avoid confusion with existing `UpdateX` data types and to keep            
  chunk-specific terminology out of the runtime crate                                                                    
  - Updates `ee-acct-types` and `snark-acct-types` with required type changes                                            
  - Adds verification state and update processing logic  

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
This is the first split pr of [https://github.com/alpenlabs/alpen/pull/1198](https://github.com/alpenlabs/alpen/pull/1198)

- [x]  Yes
- [ ]  No

[inner/outer proving](https://www.notion.so/EE-Account-Update-Proof-Inner-Outer-Split-Design-2d2901ba000f804b88e0e5597983a9c1?source=copy_link)
Last section describes we do not want terms "chunks" within ee-acct-runtime.

## AI usage
Have used AI for planning extensively and for code generation as well.

## Checklist

## Related Issues
STR-1918


crates/ee-acct-runtime/src/update_builder.rs
```diff
@@ -219,3 +221,29 @@ impl UpdateBuilder {
...
+    // Compute the inner_state_root using tree_hash_root
+    let computed_hash = tree_hash::TreeHash::<tree_hash::Sha256Hasher>::tree_hash_root(state);
+    let inner_state_root = Hash::from(computed_hash.0);
+
+    // For now, assume prev_next_msg_read_idx = 0
+    // TODO: Need to pass prev_next_msg_read_idx to UpdateBuilder::new()
+    let next_msg_read_idx = processed_messages_count as u64;
+
+    ProofState::new(inner_state_root, next_msg_read_idx)
+}
```
delbonis, line 249: I don't like this function.  This is kinda doing spooky action at a distance.  The `compute_` prefix implies that it's a pure function, but it's really not.  It's doing 2 unrelated things.  It's updating the EE account's inner state and separately computing the proof state that we check against the pub params.

crates/ee-acct-runtime/src/update_builder.rs
```diff
@@ -180,10 +180,12 @@ impl UpdateBuilder {
...
-        // Compute the new state by simulating the update
-        // For now we just use a placeholder state
-        // TODO: compute actual state root
-        let new_state = ProofState::new(Hash::new([0; 32]), 0);
+        // Compute the new ProofState
+        let new_state = compute_new_proof_state(
+            &mut self.current_state,
+            new_tip_blkid,
+            self.processed_messages.len(),
+        );
```
delbonis, line 188: We're consuming `self` in this function so we *can* just modify `self.current_state` if needed, to compute the new inner state root.

crates/ee-acct-runtime/src/update_processing.rs
```diff
@@ -59,9 +48,93 @@ impl<'v> SharedData<'v> {
...
+        Vec<u8>,
+    ) {
+        (
+            self.prev_state,
+            self.new_state,
+            self.processed_messages,
+            self.outputs,
+            self.extra_data,
+        )
+    }
```
delbonis, line 132: Should avoid this pattern, especially when there's so many fields.  It's one thing when it's like two closely coupled fields that are being unpacked, but it's another when it's very heterogeneous.

Make the fields `pub(crate)` if needed.

crates/ee-acct-runtime/src/update_processing.rs
```diff
@@ -59,9 +48,93 @@ impl<'v> SharedData<'v> {
...
+    new_state: ProofState,
+
+    /// Expected messages to be processed in this transition
+    processed_messages: Vec<MessageEntry>,
+
+    /// Expected outputs (messages + transfers) from this transition
+    outputs: UpdateOutputs,
+
+    /// Application-specific extra data
+    extra_data: Vec<u8>,
```
delbonis, line 69: Does this not duplicate the `UpdateExtraData` field in `SharedData`?  Be *very careful* about the modelling here.

crates/ee-acct-runtime/tests/common/mod.rs
delbonis, line 1: This module structure is weird.  Just make a `common.rs` in the above dir?  Or does that not work with these separate integration tests?

crates/ee-acct-runtime/src/update_processing.rs
```diff
@@ -59,9 +48,93 @@ impl<'v> SharedData<'v> {
     }
 }
 
+/// Data describing a state transition for verification within proofs.
+///
+/// This represents a state transition over one or more blocks.
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
+pub struct UpdateTransitionData {
```
delbonis, line 55: Can you explain what the motivation for defining this type is?  It's very similar to other ones that already exist, but also sorta supersceded the role of `UpdateOperationData`.  It's been a while since I looked at this so I may be misremembering things, but was there something there where it wasn't well-suited for the role it was being used for in the runtime?

crates/snark-acct-types/src/proof_interface.rs
```diff
@@ -1,33 +1,29 @@
...
+        extra_data: Vec<u8>,
+    ) -> Self {
+        Self {
+            cur_state,
+            new_state,
+            message_inputs: message_inputs.into(),
+            ledger_refs,
+            outputs,
+            extra_data: extra_data.into(),
+        }
```
delbonis, line 24: > `.into()`

We got dinged in an audit just now because we weren't careful about SSZ conversions.  We should expose these as whatever types they need to to shift the validation burden up to the caller instead of silently breaking if they're oversized.

PR #1295, opened 2026-01-18
feat: checkpoint subprotocol

prajwolrg:
## Description

Implements the new checkpoint subprotocol. 

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [X] New or updated tests
- [X] Dependency Update

## Notes to Reviewers

- [X]  Yes
- [ ]  No

https://www.notion.so/SPS-62-OL-checkpoint-structure-and-verification-225901ba000f80cfb3cff69d0c916054

## Checklist

- [X] I have updated the documentation if needed.
- [X] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

### DIFF
--- a/crates/asm/subprotocols/checkpoint/src/verification.rs
+++ b/crates/asm/subprotocols/checkpoint/src/verification.rs
@@ -130,6 +130,20 @@ fn construct_full_claim(
     ))
 }
 
+/// Computes the ASM manifests hash for a range of L1 blocks.
+///
+/// Returns an error if the manifest hashes cannot be retrieved from aux data.
+fn compute_asm_manifests_hash_for_checkpoint(
+    start_height: u32,
+    end_height: u32,
+    verified_aux_data: &VerifiedAuxData,
+) -> CheckpointValidationResult<FixedBytes<32>> {
+    let manifest_hashes =
+        verified_aux_data.get_manifest_hashes(start_height as u64, end_height as u64)?;
+
+    Ok(compute_asm_manifests_hash(&manifest_hashes))
+}
+
 /// Extracts and validates withdrawal intent logs from OL logs.
 ///
 /// Filters OL logs from the bridge gateway account, validates that withdrawal intent
@@ -165,20 +179,6 @@ fn extract_and_validate_withdrawal_intents(
     }
 
     Ok(withdrawal_intents)
-}
-
-/// Computes the ASM manifests hash for a range of L1 blocks.
-///
-/// Returns an error if the manifest hashes cannot be retrieved from aux data.
-fn compute_asm_manifests_hash_for_checkpoint(
-    start_height: u32,
-    end_height: u32,
-    verified_aux_data: &VerifiedAuxData,
-) -> CheckpointValidationResult<FixedBytes<32>> {
-    let manifest_hashes =
-        verified_aux_data.get_manifest_hashes(start_height as u64, end_height as u64)?;
-
-    Ok(compute_asm_manifests_hash(&manifest_hashes))
 }
 
 #[cfg(test)]
@@ -325,6 +325,28 @@ mod tests {
                 InvalidCheckpointPayload::L1HeightGoesBackwards { .. }
             )
         ))
+    }
+
+    #[test]
+    fn test_new_l1_tip_same_as_last_verified() {
+        let (state, harness) = test_setup();
+
+        let mut new_tip = harness.gen_new_tip();
+        new_tip.l1_height = state.verified_tip().l1_height;
+        let payload = harness.build_payload_with_tip(new_tip);
+
+        let verified_aux_data = harness.gen_verified_aux(&new_tip);
+        let signed_payload = harness.sign_payload(payload);
+
+        let current_l1_height = state.verified_tip().l1_height + 1;
+
+        let res = validate_checkpoint_and_extract_withdrawal_intents(
+            &state,
+            current_l1_height,
+            &signed_payload,
+            &verified_aux_data,
+        );
+        assert!(res.is_ok());
     }
 
     #[test]
--- a/crates/test-utils/l2/src/checkpoint.rs
+++ b/crates/test-utils/l2/src/checkpoint.rs
@@ -86,7 +86,7 @@ impl CheckpointTestHarness {
     /// - Increment the epoch by 1
     /// - Process 1-100 random L1 blocks
     /// - Process 1-200 random L2 blocks
-    fn gen_new_tip(&self) -> CheckpointTip {
+    pub fn gen_new_tip(&self) -> CheckpointTip {
         let mut rng = thread_rng();
         let mut arb = ArbitraryGenerator::new();
         let l1_blocks_processed: u32 = rng.gen_range(1..=100);
@@ -101,6 +101,11 @@ impl CheckpointTestHarness {
         let new_ol_block_commitment = OLBlockCommitment::new(new_ol_slot, new_ol_blkid);
 
         CheckpointTip::new(new_epoch, new_covered_l1_height, new_ol_block_commitment)
+    }
+
+    /// Updates the verified tip to reflect a newly accepted checkpoint.
+    pub fn update_verified_tip(&mut self, new_tip: CheckpointTip) {
+        self.verified_tip = new_tip
     }
 
     /// Generates deterministic manifest leaves for L1 blocks between verified tip and new tip.
@@ -151,19 +156,23 @@ impl CheckpointTestHarness {
         VerifiedAuxData::try_new(&data, &asm_accumulator_state).unwrap()
     }
 
+    /// Generates a valid checkpoint payload with a randomly generated tip.
+    ///
+    /// Convenience wrapper around [`Self::build_payload_with_tip`] that automatically
+    /// generates a new checkpoint tip advancing from the current verified tip.
+    pub fn build_payload(&self) -> CheckpointPayload {
+        let new_tip = self.gen_new_tip();
+        self.build_payload_with_tip(new_tip)
+    }
+
     /// Generates a valid checkpoint payload signed by the checkpoint predicate.
     ///
     /// Creates a complete checkpoint payload including:
-    /// - New checkpoint tip advancing from the current verified tip
-    /// - Random state diff and OL logs in the sidecar
-    /// - Properly constructed checkpoint claim
+    /// - Random state diff and empty OL logs in the sidecar
+    /// - Properly constructed checkpoint claim with manifest hashes
     /// - Valid checkpoint proof signature
-    pub fn build_payload(&self) -> CheckpointPayload {
-        let mut arb = ArbitraryGenerator::new();
-        let verified_tip = self.verified_tip;
-        let new_tip = self.gen_new_tip();
-
-        let state_diff: Vec<u8> = arb.generate();
+    pub fn build_payload_with_tip(&self, new_tip: CheckpointTip) -> CheckpointPayload {
+        let state_diff: Vec<u8> = ArbitraryGenerator::new().generate();
         let ol_logs = Vec::new();
         let sidecar = CheckpointSidecar::new(state_diff.clone(), ol_logs.clone()).unwrap();
 
@@ -173,7 +182,7 @@ impl CheckpointTestHarness {
         let manifest_hashes = self.gen_manifest_leaves(&new_tip);
         let asm_manifests_hash = compute_asm_manifests_hash(&manifest_hashes);
 
-        let l2_range = L2BlockRange::new(verified_tip.l2_commitment, new_tip.l2_commitment);
+        let l2_range = L2BlockRange::new(self.verified_tip.l2_commitment, new_tip.l2_commitment);
         let claim = CheckpointClaim::new(
             new_tip.epoch,
             l2_range,


crates/asm/subprotocols/checkpoint/src/errors.rs
```diff
@@ -0,0 +1,54 @@
...
+    /// advance forward or stay at the same L1 height, never go backwards.
+    #[error(
+        "checkpoint goes backwards in L1 height: new checkpoint covers up to L1 height {new_height}, but previous checkpoint covered up to L1 height {prev_height}"
+    )]
+    L1HeightGoesBackwards { prev_height: u32, new_height: u32 },
+
+    /// Checkpoint claims L1 blocks that don't exist yet.
+    ///
+    /// This error occurs when a checkpoint claims to have processed L1 blocks
+    /// up to a height that is greater than or equal to the current L1 chain tip.
```
barakshani:
```suggestion
    /// up to a height that is greater than or equal to the L1 block the checkpoint tx is in.

File: crates/asm/subprotocols/checkpoint/src/verification.rs [hunk 2]
```diff
@@ -0,0 +1,392 @@
...
+    let l1_height_covered_in_last_checkpoint = verified_tip.l1_height();
+    let l1_height_covered_in_new_checkpoint = new_tip.l1_height();
+
+    if l1_height_covered_in_new_checkpoint >= current_l1_height {
+        return Err(InvalidCheckpointPayload::CheckpointBeyondL1Tip {
+            checkpoint_height: l1_height_covered_in_new_checkpoint,
+            current_height: current_l1_height,
+        }
+        .into());
+    }
```
storopoli: Off-by-one L1 tip check likely rejects valid checkpoints that cover the last verified L1 block. `current_l1_height` is taken from the pre-state chain view (`last_verified_block`) in the handler

https://github.com/alpenlabs/alpen/blob/25faa7e03542fce3cfdca261455250076438b7ac/crates/asm/subprotocols/checkpoint/src/handler.rs#L41

but `compute_asm_manifests_hash_for_checkpoint` errors on `new_tip >= current_l1_height`, which means a checkpoint in block $H$ cannot cover height $H‑1$. If the intent is "no more than the current L1 tip", the comparison should be `>`. Consider adding a test for the equality case maybe...

crates/asm/subprotocols/checkpoint/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
```
evgenyzdanovich: what's the plan for this with regards to v0? can v0 be removed soon?

crates/asm/subprotocols/checkpoint/src/verification.rs
```diff
@@ -0,0 +1,392 @@
...
+    // 2. Validate epoch progression
+    let expected_epoch = state.verified_tip().epoch + 1;
+    if payload.inner().new_tip().epoch != expected_epoch {
+        return Err(InvalidCheckpointPayload::InvalidEpoch {
+            expected: expected_epoch,
+            actual: payload.inner().new_tip().epoch,
+        }
+        .into());
+    }
+
```
barakshani, line 48:
```suggestion
    // 3. Validate proof

File: crates/asm/subprotocols/checkpoint/src/handler.rs [hunk 5]
```diff
@@ -0,0 +1,110 @@
...
+) {
+    let Ok(payload) = extract_signed_checkpoint_from_envelope(tx) else {
+        logging::warn!("failed to extract checkpoint payload from envelope, ignoring");
+        return;
+    };
+    let epoch = payload.inner().new_tip().epoch;
+
+    logging::debug!(epoch, "processing checkpoint transaction");
+
+    let current_l1_height = l1view.pow_state.last_verified_block.height_u32();
```
barakshani: this should be the block where the checkpoint is in

crates/asm/subprotocols/checkpoint/src/verification.rs
```diff
@@ -0,0 +1,392 @@
...
+/// Computes the ASM manifests hash for a checkpoint transition.
+///
+/// Validates L1 height progression between the previous and new checkpoint tips:
+/// - Returns an error if the new checkpoint goes backwards in L1 height
+/// - Returns an error if the new checkpoint exceeds the current L1 tip
+/// - Returns a zero hash if no new L1 blocks were processed (L1 height unchanged)
+/// - Otherwise computes and returns the hash of all ASM manifests from the L1 block range
+fn compute_asm_manifests_hash_for_checkpoint(
+    verified_tip: &CheckpointTip,
+    new_tip: &CheckpointTip,
```
barakshani: I prefer calling this chkp_last_(l1_)block rather than new_tip

crates/asm/subprotocols/checkpoint/src/errors.rs
```diff
@@ -0,0 +1,58 @@
...
+pub(crate) type CheckpointValidationResult<T> = Result<T, CheckpointValidationError>;
+
+#[derive(Debug, Error)]
+pub enum CheckpointValidationError {
+    #[error("invalid checkpoint payload: {0}")]
+    InvalidPayload(#[from] InvalidCheckpointPayload),
+
+    /// Failed to retrieve manifest hashes from auxiliary data.
+    #[error("auxiliary data error: {0}")]
+    InvalidAux(#[from] AuxError),
```
barakshani, line 16: where do we emit this error? I can't find it (only in tests)

crates/asm/subprotocols/checkpoint/src/verification.rs
```diff
@@ -0,0 +1,392 @@
...
+    errors::{CheckpointValidationResult, InvalidCheckpointPayload},
+    state::CheckpointState,
+};
+
+/// Validates a checkpoint payload by verifying the sequencer signature, epoch progression,
+/// and checkpoint proof.
+///
+/// The full [`CheckpointClaim`] is reconstructed from the current subprotocol state and payload
+/// for the full proof verification.
+pub fn validate_checkpoint_payload(
```
barakshani: shouldn't we check correct L2 progression within this function?

crates/asm/subprotocols/checkpoint/src/handler.rs
```diff
@@ -0,0 +1,110 @@
...
+            continue;
+        };
+
+        // Parse destination descriptor; panic on malformed descriptors
+        let Ok(destination) = Descriptor::from_bytes(withdrawal_data.dest()) else {
+            // CRITICAL: User funds are destroyed on L2 but cannot be withdrawn on L1.
+            logging::error!(
+                "Failed to parse withdrawal destination descriptor - user funds may be lost"
+            );
+            panic!("malformed withdrawal destination descriptor");
```
barakshani: I believe this should not be the right behaviour, it opens the door to any EE to halt the OL by haveing a withdrawal request with malformed descriptor. In Alpen, we should perhaps have a check on the withdrawal precompile (if we don't have it already), but we cannot guarantee such a check on other EE's.

crates/asm/subprotocols/checkpoint/src/verification.rs
prajwolrg: > `current_l1_height` is taken from the pre-state chain view (`last_verified_block`) in the handler.

That is correct, but we update the `pow_state` before processing any subprotocol transactions as you can see [here](https://github.com/alpenlabs/alpen/blob/main/crates/asm/stf/src/stf.rs#L50-L53). That means that the `current_l1_height` is the height of the block of which this checkpoint transaction is part of. 

Edit: But we're still passing `prestate` to the `ProcessStage`, so you're right. Nice catch!

crates/asm/subprotocols/checkpoint/src/errors.rs
prajwolrg: We emit this from `compute_asm_manifests_hash`. There is implicit conversion from `AuxError` to `CheckpointValidationError`.

crates/asm/subprotocols/checkpoint/src/verification.rs
prajwolrg: Fixed in [e5d75a7](https://github.com/alpenlabs/alpen/pull/1295/commits/e5d75a7dc1ec77f1e44ec03c6422851daf8f8f7c).

crates/asm/subprotocols/checkpoint/src/errors.rs
prajwolrg: Fixed in [e5d75a7](https://github.com/alpenlabs/alpen/pull/1295/commits/e5d75a7dc1ec77f1e44ec03c6422851daf8f8f7c).

crates/asm/subprotocols/checkpoint/src/handler.rs
prajwolrg: Fixed in [e5d75a7](https://github.com/alpenlabs/alpen/pull/1295/commits/e5d75a7dc1ec77f1e44ec03c6422851daf8f8f7c).

crates/asm/subprotocols/checkpoint/src/verification.rs
prajwolrg: Correct L2 progression is verified with the proof itself. Since we construct the `CheckpointClaim` with the last verified l2 block, we rely on the OL STF guest code to have the appropriate logic for handling progression correctly.

prajwolrg: Added a check to see if the slot is progressing in [54ea7bf](https://github.com/alpenlabs/alpen/pull/1295/commits/54ea7bf3303f14ac07c8d664418af850f0228c94).

crates/asm/subprotocols/checkpoint/src/handler.rs
prajwolrg: The `OLLogs` are emitted by the OL and should do validation on the EE Logs such that there is no way for the EE to be able to halt this. Still ASM subprotocol should properly handle this without panicking.

crates/asm/subprotocols/checkpoint/src/subprotocol.rs
```diff
@@ -0,0 +1,113 @@
...
+        collector: &mut AuxRequestCollector,
+        _anchor_pre: &AnchorState,
+        _params: &Self::Params,
+    ) {
+        for tx in txs {
+            if tx.tag().tx_type() == OL_STF_CHECKPOINT_TX_TYPE {
+                match extract_signed_checkpoint_from_envelope(tx) {
+                    Ok(signed_checkpoint) => {
+                        let start_height = state.verified_tip().l1_height + 1;
+                        let end_height = signed_checkpoint.inner().new_tip().l1_height;
```
storopoli, line 54: This is dangerous. What happens if we somehow call this and a checkpoint has no L1 progress (`end_height == state.verified_tip().l1_height`), then we get that `start_height > end_height` which is unreasonable.
We should do a check here that `start_height <= end_height`.

crates/asm/subprotocols/checkpoint/src/handler.rs
```diff
@@ -0,0 +1,69 @@
+use strata_asm_bridge_msgs::BridgeIncomingMsg;
+use strata_asm_common::{MsgRelayer, TxInputRef, VerifiedAuxData, logging};
+use strata_asm_proto_checkpoint_txs::parser::extract_signed_checkpoint_from_envelope;
```
storopoli, line 3: This is still using v0 see this permalink from this branch:

https://github.com/alpenlabs/alpen/blob/ca75dfbe6548287ce0732e206de5477dc9ad1dce/crates/asm/txs/checkpoint/src/lib.rs#L5-L10

Shouldn't you be using the v1 in `mod parser`? (I don't know).

storopoli (changes requested):

crates/asm/subprotocols/checkpoint/src/handler.rs
prajwolrg: I'm using `parser` instead of `parser_v0`. I don't think I understand your comment here.

crates/asm/subprotocols/checkpoint/src/subprotocol.rs
prajwolrg: This isn't dangerous. If L1 has no progress then in `aux_resolver`, it won't resolve to any hash, which is fine because when processing we do check this condition and if indeed L1 hasn't progressed, we use empty manifest hash.

crates/asm/subprotocols/checkpoint/src/handler.rs
storopoli: Oh crap, yeah sorry. I didn't read that.
But anyways, shouldn't we remove the `pub use` in `parser_v0` in

https://github.com/alpenlabs/alpen/blob/ca75dfbe6548287ce0732e206de5477dc9ad1dce/crates/asm/txs/checkpoint/src/lib.rs#L10

that will cause confusion in the future.

crates/asm/subprotocols/checkpoint/src/handler.rs
barakshani: Conceptually, about funds getting stuck, I don't see a difference between user sending to a wrong address on the EE and a wrong address (descriptor) on L1.

crates/asm/subprotocols/checkpoint/src/handler.rs
prajwolrg: Yes, but I didn't want to touch a lot of files because of this.

crates/asm/subprotocols/checkpoint/src/handler.rs
prajwolrg: There's two types of wrong addresses:
1. Address that isn't controlled by the user, but is still valid.
2. Address that isn't valid at all.

We should handle these cases separately. We can't do anything about 1, but should try to handle 2 appropriately. I've updated the implementation such that we reject the update if the descriptor is invalid. OL should also reject the `EEUpdate` if destination is referring is invalid.

delbonis: I think it's possibly to just drop impossible withdrawal requests, as a safety behavior.  We shouldn't just assume that the OL will only ever emit properly formed withdrawal logs.  If we want to, then maybe we could add to a counter to track funds "destroyed due to protocol misuse" so that maybe we can do something with them in the future.

crates/asm/subprotocols/checkpoint/src/errors.rs
```diff
@@ -0,0 +1,73 @@
...
+/// Result type for checkpoint subprotocol operations.
+pub(crate) type CheckpointValidationResult<T> = Result<T, CheckpointValidationError>;
+
+#[derive(Debug, Error)]
+pub enum CheckpointValidationError {
+    #[error("invalid checkpoint payload: {0}")]
+    InvalidPayload(#[from] InvalidCheckpointPayload),
+
+    /// Failed to retrieve manifest hashes from auxiliary data.
+    #[error("auxiliary data error: {0}")]
```
delbonis: redundant `error` in error message

crates/asm/subprotocols/checkpoint/src/errors.rs
```diff
@@ -0,0 +1,73 @@
...
+    SequencerPredicateVerification(PredicateError),
+
+    /// Predicate verification failed.
+    #[error("checkpoint predicate verification failed: {0}")]
+    CheckpointPredicateVerification(PredicateError),
+
+    /// Checkpoint epoch does not match expected progression.
+    ///
+    /// Each checkpoint must advance the epoch by exactly 1.
+    #[error("invalid epoch: expected {expected}, got {actual}")]
```
delbonis: `invalid epoch (expected {expected}, got {got})`

crates/asm/subprotocols/checkpoint/src/errors.rs
```diff
@@ -0,0 +1,73 @@
...
+    )]
+    L2SlotDoesNotAdvance { prev_slot: u64, new_slot: u64 },
+
+    /// Invalid log data.
+    ///
+    /// This error occurs when a withdrawal intent log contains a malformed
+    /// destination descriptor. Since user funds have been destroyed on L2,
+    /// this prevents the funds from being withdrawn on L1.
+    #[error("invalid log: malformed withdrawal destination descriptor")]
+    InvalidLog,
```
delbonis: Should have a `MalformedWithdrawalDestDesc` message instead of calling it `InvalidLog`.  `InvalidLog` implies it's emitted because of other kinds of logs being malformed, but the error message says specifically that it's because a descriptor was malformed.

Also, log message can be just `malformed withdrawal dest descriptor`.

crates/asm/subprotocols/checkpoint/src/verification.rs
```diff
@@ -0,0 +1,478 @@
...
+/// Validates a checkpoint payload and extracts withdrawal intents.
+///
+/// This is the single validation function for checkpoint payloads. Once validation succeeds,
+/// the payload can be safely acted upon. Returns extracted withdrawal intents on success.
+pub fn validate_checkpoint_and_extract_withdrawal_intents(
+    state: &CheckpointState,
+    current_l1_height: u32,
+    payload: &SignedCheckpointPayload,
+    verified_aux_data: &VerifiedAuxData,
+) -> CheckpointValidationResult<Vec<WithdrawOutput>> {
```
delbonis, line 29: This function is doing a lot of different things and has a few issues because of it.

* Verifying the "provenance" of the checkpoint (sequencer's signature), which requires re-encoding the checkpoint after we just decoded it before.
  * We should verify this signature *before* even trying to parse the checkpoint structure.
  * Using the predicate spec also makes it harder to piggyback on the envelope's signature, the spec only specifies schnorr sigs anyways.
* Verifying the checkpoint follows from the epoch in the current state.
* Verifying the snark proof on the checkpoint based on current state.
* Extracting the logs from the checkpoint.
* Parsing just the withdrawal intents from the logs.

[The spec](https://www.notion.so/SPS-62-OL-checkpoint-structure-and-verification-225901ba000f80cfb3cff69d0c916054) describes a very different relationship to these steps that more closely aligns with epoch processing by the sequencer.  I would have written the checkpoint verification state machine as a separate crate independent of the ASM logic, which we perhaps could also use in the client's checkpoint sync.

The general flow should be, as distinct high-level steps:

* Verify sequencer sig.
* Verify checkpoint, updating checkpoint tracking state.
* Process logs.

crates/asm/subprotocols/checkpoint/src/verification.rs
```diff
@@ -0,0 +1,478 @@
...
+        *verified_tip.l2_commitment(),
+        payload.new_tip().l2_commitment,
+    );
+
+    let asm_manifests_hash = compute_asm_manifests_hash_for_checkpoint(
+        verified_tip,
+        payload.new_tip(),
+        current_l1_height,
+        verified_aux_data,
+    )?;
```
delbonis: This is also a bit of a weird structure, the function is doing 2 separate things: verifying/processing the manifests and constructing the claim.  We should do this outside and then pass it into `construct_full_claim` as a pre-verified argument.

crates/asm/txs/checkpoint/src/lib.rs
```diff
@@ -3,7 +3,8 @@
 pub mod constants;
 pub mod errors;
 pub mod parser;
+pub mod parser_v0;
```
delbonis: Avoid exporting these?

crates/checkpoint-types-ssz/src/lib.rs
```diff
@@ -75,3 +77,16 @@ impl_borsh_via_ssz!(CheckpointSidecar);
...
+///
+/// Concatenates the manifest hashes for all L1 blocks in the range
+/// and returns a single hash commitment over them.
+pub fn compute_asm_manifests_hash(manifest_hashes: Vec<[u8; 32]>) -> FixedBytes<32> {
+    let mut data = Vec::with_capacity(manifest_hashes.len() * 32);
+    for h in manifest_hashes {
+        data.extend_from_slice(h.as_ref());
+    }
+    let hash = hash::raw(&data);
+    hash.into()
```
delbonis: Don't need to copy the hashes, we can write them directly into the digest in multiple calls.

Maybe we need a variant of `hash::raw` like `hash::slice_iter_raw` that accepts a `impl IntoIterator<Item = &[u8]>`.

crates/test-utils/l2/Cargo.toml
delbonis, line 1: Getting really worried about this test-utils crate making dependency pseudo-cycles again!  I don't want to have to rebuild the whole workspace when I'm iterating on the unit tests in some crate high up in the dependency tree.

crates/asm/subprotocols/checkpoint/src/handler.rs
storopoli: Can you open a ticket please?

prajwolrg: Opened [this](https://alpenlabs.atlassian.net/browse/STR-2132) ticket.

crates/asm/subprotocols/checkpoint/src/handler.rs
prajwolrg: > We shouldn't just assume that the OL will only ever emit properly formed withdrawal logs

Why not? If the destination account is the bridge gateway account, can't we enforce that withdrawal destination be a well-formed `bitcoin-bosd`?

crates/asm/subprotocols/checkpoint/src/verification.rs
prajwolrg: > Verifying the "provenance" of the checkpoint 

Yes, re-encoding again after decoding is a waste. I was thinking of using the pubkey from the envelope transaction directly. The extraction logic would extract the checkpoint payload and the pubkey. Then for `provenance`, we construct the `SchnorrPredicateKey` and verify it matches the sequencer predicate in the state.

I don't understand the difference between the two:
1. Verifying the checkpoint follows from the epoch in the current state.
2. Verifying the snark proof on the checkpoint based on current state.

I mostly agree with your take and flow, except on the way we handle malformed logs https://github.com/alpenlabs/alpen/pull/1295#discussion_r2724078171.

crates/asm/subprotocols/checkpoint/src/verification.rs
prajwolrg: Yes, the structure was weird. I've refactored this in [74b88fa](https://github.com/alpenlabs/alpen/pull/1295/commits/74b88fa75232e0ea17af821c5f0e1a3e7e9676be).

crates/test-utils/l2/Cargo.toml
prajwolrg: I don't think it's psedo-cycles anywhere yet. But I do see the point. Should I create a new crate for this? or make Checkpoint harness part of some other crate?

crates/checkpoint-types-ssz/src/lib.rs
prajwolrg: That's a good idea. Done in [73c8b60](https://github.com/alpenlabs/alpen/pull/1295/commits/73c8b608231a8c5ec2b7e3cb5fffa1a3bd9be944).

crates/asm/subprotocols/checkpoint/src/verification.rs
```diff
@@ -0,0 +1,459 @@
...
+    if l1_height_covered_in_new_checkpoint >= current_l1_height {
+        return Err(InvalidCheckpointPayload::CheckpointBeyondL1Tip {
+            checkpoint_height: l1_height_covered_in_new_checkpoint,
+            current_height: current_l1_height,
+        }
+        .into());
+    }
+
+    // 3b. Invalid: checkpoint goes backwards in L1 height
+    if l1_height_covered_in_last_checkpoint > l1_height_covered_in_new_checkpoint {
```
barakshani, line 63: you seem to be comparing tips. we should compare last L1 block (tip) in previous checkpoint with earliest L1 block that this (chkp being verified) covers

crates/asm/subprotocols/checkpoint/src/verification.rs
```diff
@@ -0,0 +1,459 @@
...
+            prev_height: l1_height_covered_in_last_checkpoint,
+            new_height: l1_height_covered_in_new_checkpoint,
+        }
+        .into());
+    }
+
+    // 4. Validate L2 progression
+    let prev_slot = state.verified_tip().l2_commitment().slot();
+    let new_slot = payload.inner().new_tip().l2_commitment().slot();
+    if new_slot <= prev_slot {
```
barakshani, line 74: same issue I think

crates/asm/subprotocols/checkpoint/src/verification.rs
```diff
@@ -0,0 +1,459 @@
...
+        let state = CheckpointState::new(
+            harness.sequencer_predicate(),
+            harness.checkpoint_predicate(),
+            *harness.verified_tip(),
+        );
+        (state, harness)
+    }
+
+    #[test]
+    fn test_validate_checkpoint_success() {
```
barakshani, line 211: Let's have one more (success) test for the edge case where checkpoint L1 block is the same as the previous checkpoint last L1 block (the range of this checkpoint's L1 blocks is a single block)

crates/test-utils/l2/src/checkpoint.rs
```diff
@@ -0,0 +1,206 @@
...
+        let mut arb = ArbitraryGenerator::new();
+        let verified_tip = self.verified_tip;
+        let new_tip = self.gen_new_tip();
+
+        let state_diff: Vec<u8> = arb.generate();
+        let ol_logs = Vec::new();
+        let sidecar = CheckpointSidecar::new(state_diff.clone(), ol_logs.clone()).unwrap();
+
+        let state_diff_hash = hash::raw(&state_diff).into();
+        let ol_logs_hash = hash::raw(&ol_logs.as_ssz_bytes()).into();
```
barakshani: what exactly do we do here? generate some bytes and making sure they appear in the claim as the state diffs? So in the test above (test_invalid_state_diff) we have different bytes in the sidecar than what appear in the claim?

crates/test-utils/l2/src/checkpoint.rs
```diff
@@ -0,0 +1,206 @@
...
+        let claim = CheckpointClaim::new(
+            new_tip.epoch,
+            l2_range,
+            asm_manifests_hash,
+            state_diff_hash,
+            ol_logs_hash,
+        );
+
+        let proof = self
+            .checkpoint_predicate
```
barakshani, line 195: this means either real SNARK proof or a signature?

crates/asm/subprotocols/checkpoint/src/verification.rs
prajwolrg: > we should compare last L1 block (tip) in previous checkpoint with earliest L1 block that this (chkp being verified) covers

Why should we do this? We are deliberately not sending the earliest L1 block that this checkpoint payload covers.

crates/asm/subprotocols/checkpoint/src/verification.rs
prajwolrg: Same as in https://github.com/alpenlabs/alpen/pull/1295#discussion_r2731134561. The validity of the progression is ultimately done by the constructed `CheckpointClaim`. This is just an early redundant check as mentioned in https://github.com/alpenlabs/alpen/pull/1295#discussion_r2716196498.

crates/test-utils/l2/src/checkpoint.rs
prajwolrg: Yes.

crates/test-utils/l2/src/checkpoint.rs
prajwolrg: This is meant to be a real SNARK proof. We are mocking it here as a signature.

crates/asm/subprotocols/checkpoint/src/verification.rs
barakshani: in that case, it's not clear to me why do we even have this comparison.

what is checkpoint `k` covers L1 range [1,10] and checkpoint `k+1` covers range [5,13]? This check would pass. Now, I'm sure you are aware of it and we have this covered elsewhere (in particular, we must check that the range coverage is consecutive), but if we do have it elsewhere, what is the benefit in comparing tips here?

prajwolrg: The main reason for doing this check is to make sure we're requesting the right auxiliary data. ASM STF needs auxiliary data to proceed. If anyone running the ASM STF cannot provide the right auxiliary data, it must panic (i.e. the proof of that STF should not be generated). Since the payload can be posted onchain by ~~anyone~~ malicious sequencer, they can post the payload such that auxiliary data cannot exist. For checkpoint subprotocol it can happen in two cases:
1. The L1 exceeds the current L1 tip - We request `AsmManifest` that hasn't been yet produced by the ASM.
2. L1 range goes backwards - We request `AsmManifest` such that `end_height` < `start_height`. In this case the aux_resolver returns empty verified manifests. 

We can be sure that the AsmManifest were consumed in the right order at the construction of the `CheckpointClaim`. In your example, if the checkpoint covers range `[5,13]`, and last range covered `[1,10]`, the manifest would be constructed for `[11,13]`, and therefore a different hash would be calculated, making it a different public parameter than what the proof was generated against.

barakshani: >Since the payload can be posted onchain by anyone

Don't we check the signature first?

>the proof of that STF should not be generated

We are at the checkpoint subprotocol so some alleged proof is already given. Won't we fail the proof verification even if we don't explicitly make these checks?

> returns empty verified [manifests]

In particular for the L2 check (which is commented below), we should never accept an empty block-delta (from previous checkpoint); isn't that a more explicit check than comparing tips?

prajwolrg: > Don't we check the signature first?

Yes, we do. Saying "posted onchain by anyone" is wrong. I meant to say, a malicious sequencer can.

> Won't we fail the proof verification even if we don't explicitly make these checks?

We will fail the proof verification if we don't explicitly make these checks. We don't have to maliciously do these checks. These cheap checks allows us to do catch some errors without running the proof verification (just in case). The only check that we **really** need to have apart from proof verification is the case of `CheckpointBeyondL1Tip`. If we don't do that, a malicious sequencer can stop the ASM STF since we can't run the STF without the required aux data. 

> for the L2 check (which is commented below), we should never accept an empty block-delta

For L2 blocks, we don't accept empty L2 blocks in the checkpoint. If the L1 doesn't progress, that is allowed. That allows us to create epochs of < 10 minutes (if needed). The OL proof statements can decide whether L1 needs to progress or not. We can add a check similar to [this](https://github.com/alpenlabs/alpen/pull/1309/changes#diff-a35c92ffb2fe997445d7f01715eecb638721faec796548bc03afdebd42eb81dbR97-R101) for L1. In that case if the L1 doesn't progress, the proof itself will be invalid.

> isn't that a more explicit check than comparing tips?

Comparing tips in my opinion is an implementation detail. How do you suggest we do that without comparing tips?

crates/asm/subprotocols/checkpoint/src/verification.rs
prajwolrg: I added a test with no L1 progression in [bc35ce2](https://github.com/alpenlabs/alpen/pull/1295/commits/bc35ce287536f7b09f3b8110c8861e9c5aa85ef9).

crates/asm/subprotocols/checkpoint/src/verification.rs
barakshani: If I understood you correctly, for the previous example I gave, your point was that we can't distinguish a valid (L1) empty block delta (where 2 checkpoints cover the same range) from invalid block delta, where the tip of chkp k+1 is smaller than the tip of chkp k.

But for L2 range, we should never end with empty block delta, so I am suggesting that we check explicitly if the block delta is empty, for the L2 range.

prajwolrg: Is the check you're suggesting different than the one done [here](https://github.com/alpenlabs/alpen/pull/1295#discussion_r2725660579)?

crates/asm/subprotocols/checkpoint/src/handler.rs
delbonis: Assuming this creates a coupling between ASM and OL code, especially to ensure robustness.  It's plausible that we can ship a bad update that allows malformed logs to be emitted.  Panicking like this means that we wouldn't be able to recover since the ASM would be stalled and unable to process manual upgrade operations to replace the OL checkpoint vk with a fixed one.

crates/asm/subprotocols/checkpoint/src/verification.rs
delbonis: We can separate these checks into two that are treated mostly independently to better organize the code.

* 1 is a stateful check, that the checkpoint we're reading builds on the chain we already trust.

* 2 is a state*less* check, that the checkpoint is correct assuming *some* previous state.

We can treat the state needed to verify both as a "checkpoint sync state machine" and build it independently from the ASM itself so that we can use it in other places.

crates/asm/subprotocols/checkpoint/src/handler.rs
prajwolrg: Yes, panicking here was indeed a bad idea. Now instead of panicking, the checkpoint is considered malformed if it's invalid descriptor.

crates/asm/subprotocols/checkpoint/src/verification.rs
prajwolrg: I don't fully follow here. Both the verification checks are done based on what's on the subprotocol state.

PR #1296, opened 2026-01-19
Str 2064 fullnode sync

bewakes:
## Description

This is basically a copy-and-modify of the existing `sync` crate, but catered for the new OL related types. This also adds a method called `load_unfinalized_ol_blocks` to `UnfinalizedTracker` to make it work with both older `L2Block` and newer `OLBlock`.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist


crates/primitives/src/sync_block_header.rs
delbonis, line 1: Why do we need this?  We don't expect to have many types of OL blocks.

This should not be in the primitives crate.

crates/ol/chain-types/src/block.rs
```diff
@@ -5,6 +5,7 @@ use ssz_types::VariableList;
 use strata_asm_common::AsmManifest;
 use strata_crypto::hash;
 use strata_identifiers::{Buf32, Buf64, Epoch, OLBlockId, Slot};
+use strata_primitives::SyncBlockHeader;
```
delbonis, line 8: Eventually we want the primitives crate to depend on this crate instead of the other way around.  It's supposed to be moving towards being used by external consumers, so internal consumers depend on the actual upstream crates they need.

crates/consensus-logic/src/unfinalized_tracker.rs
```diff
@@ -344,6 +344,68 @@ impl UnfinalizedBlockTracker {
         Ok(())
     }
 
+    /// Loads the unfinalized OL blocks into the tracker which are already in the DB
+    pub fn load_unfinalized_ol_blocks(
+        &mut self,
+        ol_block_manager: &OLBlockManager,
```
delbonis, line 350: Should move towards defining a provider trait.  I know `load_unfinalized_blocks` uses this type directly, but since we're here we should bring it in-line with the current conventions.

crates/ol-sync/src/client.rs
```diff
@@ -0,0 +1,126 @@
...
+use strata_ol_chain_types_new::OLBlock;
+use strata_ol_state_types as _;
+use strata_primitives as _;
+use strata_rpc_api_new::{OLClientRpcClient, OLFullNodeRpcClient};
+use tracing::error;
+
+#[derive(Debug, thiserror::Error)]
+pub enum ClientError {
+    #[error("missing block: {0}")]
+    MissingBlock(OLBlockId),
```
delbonis, line 15: Add `use thiserror::Error`.

And don't need the colon here since it's not implying another level of error hierarchy, just do `missing block {0}` or `missing block '{0}'`.

crates/ol-sync/src/client.rs
```diff
@@ -0,0 +1,126 @@
...
+    }
+}
+
+#[async_trait::async_trait]
+pub trait SyncClient {
+    async fn get_sync_status(&self) -> Result<PeerSyncStatus, ClientError>;
+
+    fn get_blocks_range(&self, start_height: u64, end_height: u64) -> impl Stream<Item = OLBlock>;
+
+    async fn get_block_by_id(&self, block_id: &OLBlockId) -> Result<Option<OLBlock>, ClientError>;
```
delbonis, line 47: Inconsistent async vs sync here might be a red flag, I realize it's `Stream`.

We should probably avoid `async_trait` here since we can used async fns in traits now and it's not providing ergnomic improvements with that return type being generic.  We should go farther and make it `async ... -> Result<impl Stream<Item = Result<OLBlock, ClientError>>, ClientError>` to expose errors properly.  Or just avoid the complexity and not use streams yet if we don't know that we have an impl for `SyncClient` that allows streaming blocks.

crates/ol-sync/src/client.rs
```diff
@@ -0,0 +1,126 @@
...
+            .map(move |s| (s, min(self.download_batch_size as u64 + s - 1, end_height)));
+
+        stream::unfold(block_ranges, |mut block_ranges| async {
+            let (start_height, end_height) = block_ranges.next()?;
+            match self.get_blocks(start_height, end_height).await {
+                Ok(blocks) => Some((stream::iter(blocks), block_ranges)),
+                Err(err) => {
+                    error!("failed to get blocks: {err}");
+                    None
+                }
```
delbonis, line 109: Gobbling the error here isn't good and would probably lead to flapping if there's an issue with the RPC, see other comment.

crates/ol-sync/src/state.rs
```diff
@@ -0,0 +1,97 @@
...
+    let tracker = Handle::current()
+        .spawn_blocking(move || {
+            let mut tracker = UnfinalizedBlockTracker::new_empty(finalized_epoch);
+            tracker
+                .load_unfinalized_ol_blocks(&ol_tracker)
+                .map(|_| tracker)
+        })
+        .await
+        .map_err(|err| OLSyncError::LoadUnfinalizedFailed(err.to_string()))?
+        .map_err(|err| OLSyncError::LoadUnfinalizedFailed(err.to_string()))?;
```
delbonis, line 85: Can we not consolidate this somehow?

crates/ol-sync/src/state.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    pub(crate) fn finalized_epoch(&self) -> &EpochCommitment {
+        self.tracker.finalized_epoch()
+    }
+}
+
+pub(crate) async fn initialize_from_db(
+    finalized_epoch: EpochCommitment,
+    storage: &NodeStorage,
+) -> Result<OLSyncState, OLSyncError> {
```
delbonis, line 71: Can we hide the storage behind a provider trait?

crates/ol-sync/src/worker.rs
```diff
@@ -0,0 +1,284 @@
...
+                        continue;
+                    }
+                };
+                handle_finalized_epoch(finalized_epoch, &mut state, context).await;
+            }
+
+            _ = interval.tick() => {
+                do_tick(&mut state, context).await?;
+            }
+            // maybe subscribe to new blocks on client instead of polling?
```
delbonis, line 122: This should use the service framework for this instead of half reimplementing it.  You can implement this "ticking" with a new input type that supports an intermittent idle wakeup message if you need it.

But really we should plan for having a stream of new blocks being fed from the upstream RPC.  The sync worker should not be aware of any polling happening.  This couples fetching inputs with handling them and complicates testing/tracing.

crates/ol-sync/src/worker.rs
```diff
@@ -0,0 +1,284 @@
...
+        if state.has_block(&block_id) {
+            warn!(block_id = ?block_id, "block already known");
+            return Ok(());
+        }
+
+        let height = block.header().slot();
+        if height <= state.finalized_height() {
+            // got block on different fork than one we're finalized on
+            // log error and ignore received blocks
+            error!(height = height, block_id = ?block_id, "got block on incompatible fork");
```
delbonis, line 218: `%slot`, `?blkid`

crates/ol-sync/src/worker.rs
```diff
@@ -0,0 +1,284 @@
...
+) -> Result<(), OLSyncError> {
+    let mut block = block;
+    let mut fetched_blocks = vec![];
+
+    loop {
+        let block_id = block.header().compute_blkid();
+        debug!(block_id = ?block_id, height = block.header().slot(), "received new block");
+
+        if state.has_block(&block_id) {
+            warn!(block_id = ?block_id, "block already known");
```
delbonis, line 210: `?blkid`

crates/ol-sync/src/worker.rs
```diff
@@ -0,0 +1,284 @@
...
+    context: &OLSyncContext<T>,
+    block: OLBlock,
+) -> Result<(), OLSyncError> {
+    let mut block = block;
+    let mut fetched_blocks = vec![];
+
+    loop {
+        let block_id = block.header().compute_blkid();
+        debug!(block_id = ?block_id, height = block.header().slot(), "received new block");
+
```
delbonis, line 208: Wrap this function with a span that includes the ~`height`~ `slot` and `blkid` fields so that we don't have to include it in every log.

crates/ol-sync/src/worker.rs
```diff
@@ -0,0 +1,284 @@
...
+    while let Some(block) = fetched_blocks.pop() {
+        state.attach_block(block.signed_header())?;
+        // TODO: Update this to use OL storage when available
+        context
+            .storage
+            .ol_block()
+            .put_block_data_async(block.clone())
+            .await?;
+        let block_idx = block.header().slot();
+        debug!(%block_idx, "ol sync: sending chain tip msg");
```
delbonis, line 251: `%slot`

crates/ol-sync/src/worker.rs
```diff
@@ -0,0 +1,284 @@
...
+}
+
+async fn handle_finalized_epoch<T: SyncClient>(
+    fin_epoch: EpochCommitment,
+    state: &mut OLSyncState,
+    _context: &OLSyncContext<T>,
+) {
+    if let Err(e) = handle_block_finalized(state, fin_epoch).await {
+        error!(?fin_epoch, err = %e, "failed to handle newly finalized block");
+    }
```
delbonis, line 134: Is it valid to gobble this error if `do_tick` returning an error propagates up and triggers the worker to exit?

crates/ol-sync/src/worker.rs
```diff
@@ -0,0 +1,284 @@
...
+
+#[expect(
+    missing_debug_implementations,
+    reason = "Some inner types don't have Debug implementation"
+)]
+pub struct OLSyncContext<T: SyncClient> {
+    client: T,
+    storage: Arc<NodeStorage>,
+    status_channel: Arc<StatusChannel>,
+    fcm_tx: Sender<ForkChoiceMessage>,
```
delbonis, line 32: I thought we created a FCM handle that uses the service framework's command handle?  Can we use this here instead of the underlying `Sender` directly?

PR #1299, opened 2026-01-19
da-framework enhancements

evgenyzdanovich:
## Description

General `da-framework` enhancements:
- Counter support in the compound
- Coersion support in the compound
- Additional semantic in the compound macro for coersion 
- Counter by VarInt schema
- Some general nits

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

These changes are required for cleaner ee statediffs: 
https://github.com/alpenlabs/alpen/pull/1285

- [ ]  Yes
- [x]  No

## Checklist

### DIFF
--- a/crates/da-framework/src/compound.rs
+++ b/crates/da-framework/src/compound.rs
@@ -59,8 +59,7 @@ impl<T: Bitmap> BitSeqReader<T> {
     }
 
     /// Returns the next bit, if possible.
-    #[allow(clippy::should_implement_trait)]
-    pub fn next(&mut self) -> bool {
+    pub fn next_bit(&mut self) -> bool {
         if self.off >= T::BITS {
             panic!("bitqueue: out of bits");
         }
@@ -76,7 +75,7 @@ impl<T: Bitmap> BitSeqReader<T> {
         &mut self,
         dec: &mut impl Decoder,
     ) -> CodecResult<C> {
-        let set = self.next();
+        let set = self.next_bit();
         if set {
             C::decode_set(dec)
         } else {
@@ -146,10 +145,28 @@ impl<T: Bitmap> Default for BitSeqWriter<T> {
 ///     }
 /// }
 /// ```
+///
+/// # Limitations
+///
+/// - **Type coercion is only supported for registers**, not counters. Counter target types are
+///   determined by the [`CounterScheme::Base`](crate::CounterScheme::Base) type.
+///
+/// - **Context is ignored for primitive fields.** While the macro accepts a context type parameter
+///   (`DiffType<ContextType> ...`), primitive fields (`DaRegister`, `DaCounter`) have `Context =
+///   ()` and the macro passes `&()` to their `apply` and `poll_context` methods. The compound's
+///   context parameter is currently unused.
+///
+/// # Context-aware resolution
+///
+/// For cases where field resolution requires context (e.g., resolving compact serials
+/// to full IDs using a lookup table), implement [`DaWrite`](crate::DaWrite) manually instead of
+/// using the macro. This gives full control over how context is used during apply.
+///
+/// See the `context` test module for an example of manual `DaWrite` implementation.
 // TODO turn this into a proc macro
 #[macro_export]
 macro_rules! make_compound_impl {
-    // Entry point without context type - delegates to version with () context
+    // Entry point without context type - uses () context
     (
         $tyname:ident $maskty:ident => $target:ty {
             $( $fname:ident : $daty:ident $fspec:tt ),* $(,)?
@@ -169,7 +186,6 @@ macro_rules! make_compound_impl {
         }
     ) => {
         // Compile-time check: ensure bitmap has enough bits for all fields.
-        // Equal to const_assert! from static_assertions, but doesn't bring the dependency.
         const _: () = {
             const FIELD_COUNT: usize = [$(stringify!($fname)),*].len();
             const MASK_BITS: usize = <$maskty>::BITS as usize;
@@ -193,8 +209,6 @@ macro_rules! make_compound_impl {
 
                 bitw.mask().encode(enc)?;
 
-                // This goes through them in the same order as the above, which
-                // is why this is safe.
                 $(
                     if !$crate::CompoundMember::is_default(&self.$fname) {
                         $crate::CompoundMember::encode_set(&self.$fname, enc)?;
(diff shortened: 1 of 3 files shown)


crates/da-framework/src/compound.rs
```diff
@@ -59,6 +59,7 @@ impl<T: Bitmap> BitSeqReader<T> {
     }
 
     /// Returns the next bit, if possible.
+    #[allow(clippy::should_implement_trait)]
```
storopoli:
```suggestion
    #[expect(clippy::should_implement_trait, reason="change me")]
```

crates/da-framework/src/counter.rs
```diff
@@ -269,12 +274,83 @@ pub mod counter_schemes {
...
+        type Base = u64;
+        type Incr = VarintIncr;
+
+        fn is_zero(incr: &Self::Incr) -> bool {
+            incr.0.inner() == 0
+        }
+
+        fn update(base: &mut Self::Base, incr: &Self::Incr) {
+            *base += incr.0.inner() as u64;
+        }
```
storopoli, line 337: `CtrU64ByVarint`'s docs suggest avoiding overflow

https://github.com/alpenlabs/alpen/blob/31e4a661b458336b757df93fcaf08de769cb249d/crates/da-framework/src/counter.rs#L315-L325

but `update` still uses unchecked addition.

crates/da-framework/src/compound.rs
```diff
@@ -184,17 +221,16 @@ macro_rules! make_compound_impl {
                 v
             }
 
-            fn poll_context(&self, target: &Self::Target, context: &Self::Context) -> Result<(), $crate::DaError> {
-                $(
-                    $crate::DaWrite::poll_context(&self.$fname, &target.$fname, context)?;
-                )*
+            fn poll_context(&self, _target: &Self::Target, _context: &Self::Context) -> Result<(), $crate::DaError> {
+                // Note: poll_context is skipped for coercion fields since the types don't match.
+                // This is fine since poll_context is mainly used for context validation.
```
storopoli: This change might not be ok.

The macro now skips `poll_context` for all fields, not just coercion ones. `poll_context` is implemented as a no‑op at the compound level, so even non‑coercion members never get their context validation called.

crates/da-framework/src/compound.rs
```diff
@@ -184,17 +221,16 @@ macro_rules! make_compound_impl {
...
                 Ok(())
             }
 
-            fn apply(&self, target: &mut Self::Target, context: &Self::Context) -> Result<(), $crate::DaError> {
+            fn apply(&self, target: &mut Self::Target, _context: &Self::Context) -> Result<(), $crate::DaError> {
                 // Depends on all the members being accessible.
                 $(
-                    $crate::DaWrite::apply(&self.$fname, &mut target.$fname, context)?;
+                    $crate::_mct_field_apply!(self target; $fname $daty $fspec);
```
storopoli: `apply` now uses `ContextlessDaWrite` for all fields.

https://github.com/alpenlabs/alpen/blob/31e4a661b458336b757df93fcaf08de769cb249d/crates/da-framework/src/compound.rs#L264-L285

That only works if every member's `Context = ()`. If any member expects a real context, the macro either won't compile (trait bound fails) or you've effectively removed its validation path.

storopoli (changes requested):

crates/da-framework/src/compound.rs
```diff
@@ -203,10 +239,48 @@ macro_rules! make_compound_impl {
...
+    ($self:ident $target:ident; $fname:ident register { $fty:ty }) => {
+        $crate::ContextlessDaWrite::apply(&$self.$fname, &mut $target.$fname)?
+    };
+    // Counter - use standard DaWrite::apply (counter targets scheme's Base type)
+    ($self:ident $target:ident; $fname:ident counter ( $fty:ty )) => {
+        $crate::ContextlessDaWrite::apply(&$self.$fname, &mut $target.$fname)?
+    };
+    ($self:ident $target:ident; $fname:ident counter { $fty:ty }) => {
+        $crate::ContextlessDaWrite::apply(&$self.$fname, &mut $target.$fname)?
     };
```
delbonis: Inconsistency here with `$fname` vs `ContextlessDaWrite`, is this correct?

crates/da-framework/src/compound.rs
delbonis: On this context feature, the purpose was to enable us to use account serials in an encoding but actually resolve them to account IDs when applying the diff.  This would also make sense to leverage for the EVM DA as well.  I didn't fully figure out everything needed to make this work, so this might still need some tweaks/extension when we get around to that.

delbonis (changes requested):
This is very interesting.  Only some very minor comments.

crates/da-framework/src/compound.rs
evgenyzdanovich: Jose, it's equally broken on the main in case you want to define a custom context for compound (like below) with primitives as fields.

For example,

```
 make_compound_impl! {                                                                                                                                                                                            
      MyDiff<MyContext> u8 => MyTarget {                                                                                                                                                                           
          field: register i32,                                                                                                                                                                                     
      }                                                                                                                                                                                                            
  } 
 ```
 
 `DaRegister<i32>` has type `Context = ()`, so it doesn't compile at all (apply for the type passes MyContext while register expects `&()`).
 
-----------------------------

I'm trying to find a proper middleground: to not complicate macro too much, but still make it somewhat useful.

1. **For now, I documented the usage pattern together with current limitations.**
**Main limitation IMO**: its unclear what to do when we have different context types required for fields (`field1` requires `Context1`, `field2` requires `Context2`, compound `struct` defines `MyContext`). Both solutions aren't ideal (current on main and modified on this branch) and suffer from this problem in one way or another.

2. For now at least, I propose to use manual impl for `DaWrite` and `Codec` for more complicated cases and use macro for simple ones (such as contextless with coersion support). 
This implies we can forbid "contextful" explicitly in the macro and simplify it, until we establish usage pattern and think about proper "context resolution" (like the problem I mentioned above). 

3. I've also added a small test illustrating the AccountSerial usage and manual impl of `DaWrite` for such a case. It can be extended once we have proper context resolution implemented.

Thoughts?

crates/da-framework/src/compound.rs
evgenyzdanovich: I _think_ this is fine, `$fname` part just ensures coersion support (via `apply_into` non-trait impl method **for the register**), the rest uses standard `apply` fn from the trait.

crates/da-framework/src/compound.rs
evgenyzdanovich: adjusted.

crates/da-framework/src/counter.rs
evgenyzdanovich: thanks, done.

crates/da-framework/src/compound.rs
evgenyzdanovich: renamed, the method name was a bit misleading in the first place.

crates/da-framework/src/compound.rs
delbonis: ok I see, that makes sense

crates/da-framework/src/compound.rs
delbonis: Yeah I had a hard time figuring out how to make the context and compounds play together nicely.  I think where we need it will be near the "top level" of the diff and we would have to define some of those types manually *anyways*, so it's fine to leave it not completely integrated.

storopoli: I like your middleground. Let's go with that.

PR #1301, opened 2026-01-20
Alpen client e2e tests

voidash:
## Description

 This PR enables P2P networking for alpen-client so multiple nodes can gossip blocks to each other. Before this PR, discovery was disabled. There was a bug in `gossip` message encoding which didn't allow propagating blocks. This is also fixed by the PR 

There are tests for 
- `admin_addPeer` method where we can add peers after startup
- `bootnodes + discovery` method .i.e using `discv5` and `bootnodes`

Topology tested
- Star topology where all full nodes connect to sequencer
- relay style where one of the fullnode is a bootnode and sequencer and other fullnodes connect with it. 
- late fullnode joiner 

when peers discover each other they connect with RLPx where we share protocol information. WE share the type of protocol we support and it's ID. the MESSGE ID is pretty important. The bug was that when we were encoding message our message ID, the prefix 0xf9 was present in every message and internally reth lib adds 0x10 which makes it 0x109 but  
we have u8 data type. You can see that it becomes 0x09. And 0x09 is not a valid message ID for our alpen gossip

```
cdjk@cdjks-MacBook-Pro ~/g/a/v/f/t/alpen_client> tree -L 1 .                 alpen-client-e2e-tests!?
├── __init__.py
├── __pycache__
├── test_block_propagation.py
├── test_bootnode_discovery.py
├── test_fullnode_relay.py
├── test_fullnode_sync.py
├── test_multi_fullnode.py
├── test_node_info.py
└── test_pure_bootnode_discovery.py
```

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-1929

### DIFF
--- a/bin/alpen-client/src/dummy_ol_client.rs
+++ b/bin/alpen-client/src/dummy_ol_client.rs
@@ -114,3 +114,36 @@ fn u64_to_256(v: u64) -> [u8; 32] {
     result[24..32].copy_from_slice(&v.to_le_bytes());
     result
 }
+
+#[cfg(test)]
+mod tests {
+    use proptest::prelude::*;
+
+    use super::*;
+
+    proptest! {
+        #[test]
+        fn u64_to_256_is_deterministic_and_reversible(v: u64) {
+            let bytes = u64_to_256(v);
+
+            // Verify structure: [1u64 LE][0u64][0u64][v LE]
+            let prefix = u64::from_le_bytes(bytes[0..8].try_into().unwrap());
+            let mid1 = u64::from_le_bytes(bytes[8..16].try_into().unwrap());
+            let mid2 = u64::from_le_bytes(bytes[16..24].try_into().unwrap());
+            let suffix = u64::from_le_bytes(bytes[24..32].try_into().unwrap());
+
+            prop_assert_eq!(prefix, 1u64);
+            prop_assert_eq!(mid1, 0u64);
+            prop_assert_eq!(mid2, 0u64);
+            prop_assert_eq!(suffix, v);
+        }
+
+        #[test]
+        fn u64_to_256_produces_unique_outputs(v1: u64, v2: u64) {
+            prop_assume!(v1 != v2);
+            let bytes1 = u64_to_256(v1);
+            let bytes2 = u64_to_256(v2);
+            prop_assert_ne!(bytes1, bytes2);
+        }
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1072,6 +1072,7 @@ dependencies = [
  "clap",
  "eyre",
  "jsonrpsee",
+ "proptest",
  "reth-chainspec",
  "reth-cli-commands",
  "reth-cli-runner",
--- a/bin/alpen-client/Cargo.toml
+++ b/bin/alpen-client/Cargo.toml
@@ -67,3 +67,4 @@ tokio.workspace = true
 tracing.workspace = true
 
 [dev-dependencies]
+proptest.workspace = true


functional-tests-new/envconfigs/alpen_client.py
```diff
@@ -0,0 +1,134 @@
...
+        fn1_enode = fullnode_1.get_enode()
+
+        # Connect via admin_addPeer:
+        # - sequencer ↔ fullnode_0
+        # - fullnode_0 ↔ fullnode_1
+        fn0_rpc = fullnode_0.create_rpc()
+        seq_rpc = sequencer.create_rpc()
+
+        seq_rpc.admin_addPeer(fn0_enode)
+        fn0_rpc.admin_addPeer(fn1_enode)
```
sapinb: we dont want to do addPeer on each node.
The discovery should handle this.

crates/reth/node/src/gossip/package.rs
```diff
@@ -20,6 +20,10 @@ use strata_primitives::{
 /// Size of the sequence number in bytes.
 const SEQ_NO_SIZE: usize = mem::size_of::<u64>();
 
+/// Message ID for gossip packages
+/// This is prepended to each Message
+const GOSSIP_PACKAGE_MSG_ID: u8 = 0x00;
+
```
storopoli, line 26: Damn, that was the bug you've told me?!
Nice catch!

crates/reth/node/src/gossip/package.rs
```diff
@@ -150,9 +161,21 @@ impl AlpenGossipPackage {
...
     pub(crate) fn try_decode(buf: &mut &[u8]) -> Result<Self> {
         ensure!(!buf.is_empty(), "buffer is empty");
 
+        // Read and verify message ID
+        let msg_id = buf.get_u8();
+        ensure!(
+            msg_id == GOSSIP_PACKAGE_MSG_ID,
+            "unexpected gossip message type: expected {}, got {}",
+            GOSSIP_PACKAGE_MSG_ID,
+            msg_id
```
storopoli: You can do

```suggestion
            "unexpected gossip message type: expected {GOSSIP_PACKAGE_MSG_ID}, got {msg_id}"    
```

functional-tests-new/tests/alpen_client/test_pure_bootnode_discovery.py
```diff
@@ -0,0 +1,95 @@
...
+        logger.info(f"Fullnode enode: {fn_info.get('enode', 'N/A')}")
+
+        # Wait for peer discovery - this is the key test!
+        # Without admin_addPeer, nodes must discover each other via discv5
+        logger.info("Waiting for nodes to discover each other via discv5 (no admin_addPeer)...")
+
+        try:
+            # Give discovery more time since it's organic
+            sequencer.wait_for_peers(1, timeout=90)
+            fullnode.wait_for_peers(1, timeout=90)
```
storopoli: Do we really need 90 seconds here?

functional-tests-new/common/services/alpen_client.py
```diff
@@ -0,0 +1,199 @@
...
+            return []
+
+    def get_peer_count(self) -> int:
+        """Get number of connected peers."""
+        rpc = self.create_rpc()
+        try:
+            result = rpc.net_peerCount()
+            return int(result, 16)
+        except Exception:
+            return 0
```
storopoli, line 132: `get_peers()`/`get_peer_count(`) return `[]`/`0` on any exception. Logging the exception at least at `debug` (or `warning` once) will make test failures much easier to debug.

functional-tests-new/tests/alpen_client/test_fullnode_sync.py
```diff
@@ -0,0 +1,85 @@
...
+            fullnode_1.wait_for_peers(1, timeout=30)
+
+            # Verify historical sync
+            fullnode_1.wait_for_block(target_block, timeout=120)
+            fn1_hash = fullnode_1.get_block_by_number(target_block)["hash"]
+            assert expected_hash == fn1_hash, "Historical block hash mismatch"
+
+            # Verify new block relay
+            new_target = target_block + 5
+            fullnode_1.wait_for_block(new_target, timeout=60)
```
storopoli, line 79: This can be flaky and hard to debug.
You should make sure that the `sequencer` have produced this block first.

You are only waiting the sequencer for the original `target_block` not the new `new_target`.

bin/alpen-client/src/dummy_ol_client.rs
```diff
@@ -0,0 +1,103 @@
...
+
+fn slot_to_block_commitment(slot: u64) -> OLBlockCommitment {
+    OLBlockCommitment::new(slot, Buf32::new(u64_to_256(slot)).into())
+}
+
+fn u64_to_256(v: u64) -> [u8; 32] {
+    // SAFETY: This is a simple bit reinterpretation for generating deterministic test data.
+    // The resulting array is used only for creating mock block IDs.
+    unsafe { mem::transmute([1u64, 0u64, 0u64, v]) }
+}
```
storopoli, line 116: The current `mem::transmute` is platform-endianness dependent. For deterministic behavior across architectures, build the `[u8; 32]` with explicit `to_le_bytes()` or `to_be_bytes()` concatenation.

bin/alpen-client/src/dummy_ol_client.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    async fn epoch_summary(&self, epoch: Epoch) -> Result<OLEpochSummary, OLClientError> {
+        let commitment = EpochCommitment::new(
+            epoch,
+            epoch as u64,
+            self.slot_to_block_commitment(epoch as u64).blkid,
+        );
+        Ok(OLEpochSummary::new(commitment, commitment, vec![]))
+    }
+}
```
storopoli, line 66: Dummy OL epoch chaining may trip OL tracker if chain status advances.

The OL tracker expects each `epoch_summary` to chain to the previous epoch commitment. It initializes `expected_prev` from the local best epoch, then for each fetched epoch it checks `epoch_summary.prev_epoch() == expected_prev`. If that fails on the first fetched epoch, it forces a reorg. If it fails later, it stops fetching and reorgs on the next cycle.

https://github.com/alpenlabs/alpen/blob/7ce7307e84d8a47b0bef0eebf0f16722ac70111b/crates/alpen-ee/ol-tracker/src/task.rs#L128-L167

The dummy OL client currently constructs an `OLEpochSummary` where `prev` is set to the same epoch commitment.

That means for epoch `N`, `prev_epoch == epoch N` instead of `epoch N-1`, so the continuity check fails the moment the tracker tries to advance past its local best epoch.

Why “may” this happen? Because with the dummy client, `chain_status()` always returns the same genesis epoch for latest/confirmed/finalized, so in normal dummy-only usage the tracker won’t try to advance and you won’t hit the check.

But if anything causes the tracker to ask for epoch summaries beyond the local epoch (e.g., if chain status were to move forward, or if `epoch_summary` is called directly in another context), the continuity check will detect the mismatch and trigger a reorg loop.

---

TLDR; so far we're ok for the current fn-tests. We might have future flakiness if the dummy ever advances: If a later change makes the dummy `chain_status` move forward (or any code path calls `epoch_summary` for `epoch > 0`), the tracker will immediately detect a discontinuity and trigger reorg loops. That would show up as hangs or test failures that are hard to diagnose.

storopoli (changes requested):
Well you brought `DummyOLClient` back from the dead, now it is your responsibility...

EDIT: You should also add that fact to the "Notes to reviewers" in the PR body please.

bin/alpen-client/src/dummy_ol_client.rs
```diff
@@ -97,7 +107,10 @@ fn slot_to_block_commitment(slot: u64) -> OLBlockCommitment {
...
 fn u64_to_256(v: u64) -> [u8; 32] {
-    // SAFETY: This is a simple bit reinterpretation for generating deterministic test data.
-    // The resulting array is used only for creating mock block IDs.
-    unsafe { mem::transmute([1u64, 0u64, 0u64, v]) }
+    // Use explicit little-endian byte order for deterministic cross-platform behavior.
+    let mut result = [0u8; 32];
+    result[0..8].copy_from_slice(&1u64.to_le_bytes());
+    // bytes 8..16 and 16..24 are already zero
+    result[24..32].copy_from_slice(&v.to_le_bytes());
+    result
```
storopoli, line 115: We should do a sanity check test on this?
Ideally a `proptest!` property test.

PR #1303, opened 2026-01-20
STR-2120: Add OlState Update to status_channel

bewakes:
## Description

Resolves - [STR-2120](https://alpenlabs.atlassian.net/browse/STR-2120)

This is a quick PR to add `OLState` Updates to `StatusChannel` which only contains `Chainstate` updates. This adds a new attribute to the status channel for ol state similar to what was there for chain state. The older attribute will be removed later keeping just the newer channel for ol state updates.

This is required for the work that both @krsnapaudel  and I are working on.
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-2120]: https://alpenlabs.atlassian.net/browse/STR-2120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/status/src/status_manager.rs
+++ b/crates/status/src/status_manager.rs
@@ -197,6 +197,14 @@ impl StatusChannel {
         }
     }
 
+    /// Sends the updated `OLSyncStatusUpdate` to the chain state receiver. Logs a warning if the
+    /// receiver is dropped.
+    pub fn update_ol_sync_status(&self, update: OLSyncStatusUpdate) {
+        if self.sender.ols.send(Some(update)).is_err() {
+            warn!("chain state receiver dropped");
+        }
+    }
+
     /// Sends the updated `ClientState` to the client state receiver. Logs a warning if the receiver
     /// is dropped.
     pub fn update_client_state(&self, post_state: ClientState, post_block: L1BlockCommitment) {


krsnapaudel:
@bewakes Thanks for sharing. Certainly much less diff than my changes. Check `STR-2120` branch.

crates/status/src/chain.rs
```diff
@@ -58,25 +58,25 @@ impl ChainSyncStatus {
 
 /// Published to the FCM status including chainstate.
 #[derive(Debug, Clone)]
-pub struct ChainSyncStatusUpdate {
+pub struct ChainSyncStatusUpdate<C = Chainstate> {
     new_status: ChainSyncStatus,
-    new_tl_chainstate: Arc<Chainstate>,
+    new_tl_state: Arc<C>,
 }
```
delbonis, line 68: The state will be too large to keep in memory.  Anything that's consuming state updates will be expected to be able to make state queries directly to the database if they need it.  They should just depend on the fields from `ChainSyncStatusUpdate` to know when/how to make queries to the updated database state.

Also, using default type parameters like this creates secret dependencies/complexity.  They're for things like making a container generic over the memory allocator but defaulting to using the system allocator.

bewakes: > The state will be too large to keep in memory

This is the toplevel state. Isn't the design that we have supposed to make that in-memory state small? Right now we keep accounts in memory(for the single ED) but then later we will just have account state root in the state which will still be smaller. I don't think the other fields are that big to keep in memory. Or do we expect memory-heavy things to be included in the toplevel state?

Having said that I agree whoever requires the state will be able to fetch from db using the fields from `ChainSyncStatusUpdate`.

I also agree with the point on defualt generic param, at least in this case.

bewakes: Changed in https://github.com/alpenlabs/alpen/pull/1303/commits/fb5789d56b994b987bb199f9cdcf3f88aee15941

delbonis: I want to avoid us getting in the habit of using `OLState` "as a whole" since it'll be broken up as we expand the design in this area.  Yes sure there's certain fields we can (and should) take out of it to be able to have available more freely where they're needed (like, without having to make a database query), but we should be doing that on a selective basis to avoid creating unintentional dependencies between components.

bewakes: Yeah, that's fair. I've removed the OLState as a whoe from the update sync which only has the existing `ChainSyncStatus` with tip, last epoch, l1 safe block etc. PTAL.

PR #1310, opened 2026-01-22
feat(ol): Constructing DA (Epoch `DAWrite` Accumulation + DA Blob Assembly)

storopoli:
## Description

Implements OL DA construction per STR-1366 and spec alignment (see links below).

Includes targeted tests for determinism and consistency.

### Type of Change

- [x] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- The serialization is done using `strata-codec` and I've tried to minimize as much as possible the on-chain footprint.
- Done with the aid of Cursor using Opus 4.5 (reasoning) and GPT-5.2 codex high reasoning.

- [x]  Yes
- [ ]  No

- [SPS-da-primitives](https://www.notion.so/SPS-da-primitives-Data-Availability-Primitives-2e9901ba000f804f94c4e0d26282fe64)
- [SPS-ol-da-structure](https://www.notion.so/SPS-ol-da-structure-Orchestration-Layer-Data-Availability-Structure-2ea901ba000f800f80e5d249b08a82c4)
- [Data Availability: Data Flow & Plan](https://www.notion.so/Data-Availability-Data-Flow-Plan-2ac901ba000f803eb42bfaf996a0616b)

## Checklist

## Related Issues

STR-1366

### DIFF
--- a/crates/checkpoint-types-ssz/Cargo.toml
+++ b/crates/checkpoint-types-ssz/Cargo.toml
@@ -25,6 +25,7 @@ thiserror.workspace = true
 
 [dev-dependencies]
 proptest.workspace = true
+
 strata-identifiers = { workspace = true, features = ["test-utils"] }
 strata-ol-chain-types-new = { workspace = true, features = ["test-utils"] }
 strata-test-utils-ssz.workspace = true
--- a/crates/checkpoint-types-ssz/src/test_utils.rs
+++ b/crates/checkpoint-types-ssz/src/test_utils.rs
@@ -1,14 +1,16 @@
 //! Test utilities and proptest strategies for checkpoint types.
 
 use proptest::prelude::*;
-use strata_identifiers::test_utils::{
-    buf64_strategy, epoch_strategy, fixed_bytes_32_strategy, ol_block_commitment_strategy,
+use strata_identifiers::{
+    AccountSerial,
+    test_utils::{
+        buf64_strategy, epoch_strategy, fixed_bytes_32_strategy, ol_block_commitment_strategy,
+    },
 };
-use strata_ol_chain_types_new::test_utils::ol_log_strategy;
 
 use crate::{
     CheckpointClaim, CheckpointPayload, CheckpointSidecar, CheckpointTip, L2BlockRange,
-    SignedCheckpointPayload,
+    MAX_LOG_PAYLOAD_BYTES, SignedCheckpointPayload,
 };
 
 /// Strategy for generating random [`CheckpointTip`] values.
@@ -35,7 +37,14 @@ fn proof_strategy() -> impl Strategy<Value = Vec<u8>> {
 
 /// Strategy for generating random OL logs of varying sizes.
 fn ol_logs_strategy() -> impl Strategy<Value = Vec<crate::OLLog>> {
-    prop::collection::vec(ol_log_strategy(), 0..10)
+    prop::collection::vec(
+        (
+            any::<u32>().prop_map(AccountSerial::from),
+            prop::collection::vec(any::<u8>(), 0..=MAX_LOG_PAYLOAD_BYTES),
+        )
+            .prop_map(|(account_serial, payload)| crate::OLLog::new(account_serial, payload)),
+        0..10,
+    )
 }
 
 /// Strategy for generating random [`CheckpointSidecar`] values.


crates/ledger-types/src/state_accessor.rs
```diff
@@ -107,4 +108,18 @@ pub trait IStateAccessor {
         // Default implementation does nothing - only implementations that manage
         // the legacy chainstate need to implement this
     }
+
+    /// Enables or disables DA tracking on a state accessor.
+    ///
+    /// Default implementation is a no-op for accessors that do not track DA writes.
+    fn set_da_tracking_enabled(&mut self, _enabled: bool) {
+        // Default implementation does nothing.
+    }
```
delbonis: DA tracking is a property of the `IStateAccessor` impl, it shouldn't be something the STF enables/disables explicitly.

crates/ledger-types/src/state_accessor.rs
```diff
@@ -107,4 +108,18 @@ pub trait IStateAccessor {
...
+
+    /// Records output logs for inclusion in the DA blob.
+    ///
+    /// Default implementation is a no-op for accessors that do not track DA logs.
+    fn record_output_logs(&mut self, _logs: &[OLLog]) {
+        // Default implementation does nothing.
+    }
```
delbonis: Logs are not part of state, this should not exist.

crates/ol/state-types/src/snark_account.rs
```diff
@@ -8,8 +9,15 @@ use crate::ssz_generated::ssz::state::{OLSnarkAccountState, ProofState};
...
 impl OLSnarkAccountState {
     /// Creates an account instance with specific values.
-    pub(crate) fn new(seqno: Seqno, proof_state: ProofState, inbox_mmr: Mmr64) -> Self {
+    pub(crate) fn new(
+        seqno: Seqno,
+        proof_state: ProofState,
+        inbox_mmr: Mmr64,
+        update_vk: Vec<u8>,
+    ) -> Self {
+        let update_vk = VariableList::new(update_vk).expect("ol/state: update vk exceeds limit");
```
delbonis: This should be passed in pre-constructed for safety.

crates/ol/stf/src/assembly.rs
```diff
@@ -110,6 +110,10 @@ pub fn execute_block_inputs<S: IStateAccessor>(
         &tx_ctx,
     )?;
 
+    // Record output logs for the DA payload before any epoch-sealing updates.
+    let log_snapshot = output.snapshot_logs();
+    state.record_output_logs(&log_snapshot);
```
delbonis: What is this needed for?  Logs are already emitted as part of the return values.

crates/ol/stf/src/assembly.rs
```diff
@@ -120,7 +124,9 @@ pub fn execute_block_inputs<S: IStateAccessor>(
     let post_state_roots = if let Some(manifest_container) = block_exec_input.manifest_container() {
         // Terminal block, with manifests.
         let term_ctx = tx_ctx.basic_context();
+        state.set_da_tracking_enabled(false);
         manifest_processing::process_block_manifests(state, manifest_container, term_ctx)?;
+        state.set_da_tracking_enabled(true);
```
delbonis: We'd want to restructure the block exec functions for when doing DA tracking a bit to go through the non-sealing parts of the epoch so that this function doesn't even get called.  We're *just* doing the DA tracking for steps 2 and 3, not even step 1 actually.  See the ["processing flows" section of the spec](https://www.notion.so/SPS-ol-stf-Orchestration-layer-STF-v1-268901ba000f80b68179dc2e4b83fba0?source=copy_link#27a901ba000f80338a53ec87643642b5).

It'd be great if we could find a way to split out the different phases and checks into a more compositional structure, but as-written it's to more closely align with the spec.

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+    fn encode(&self, enc: &mut impl Encoder) -> Result<(), CodecError> {
+        enc.write_buf(&self.0.to_be_bytes())
+    }
+
+    fn decode(dec: &mut impl Decoder) -> Result<Self, CodecError> {
+        let mut buf = [0u8; 2];
+        dec.read_buf(&mut buf)?;
+        Ok(Self(u16::from_be_bytes(buf)))
+    }
+}
```
delbonis: Codec uses big-endian for [all integers](https://github.com/alpenlabs/strata-common/blob/main/crates/codec/src/types.rs#L62).

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+// ============================================================================
+
+/// Versioned OL DA payload containing state diff and withdrawal intents.
+#[derive(Debug)]
+pub struct OLDaBlobV1 {
+    /// State diff for the epoch.
+    pub state_diff: StateDiff,
+
+    /// Withdrawal intents encoded into the blob.
+    pub withdrawal_intents: WithdrawalIntents,
```
delbonis: **Withdrawal intents are just a type of log**, this doesn't exist anywhere in the specs afaik.  If it is somewhere, then please point it out to me because that needs to be corrected.

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+
+    /// Ordered output logs emitted during execution.
+    pub output_logs: OutputLogs,
```
delbonis: Logs are not part of DA, they're part of the checkpoint structure.

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+            output_logs,
+        })
+    }
+}
+
+/// Withdrawal intents included in the DA payload.
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct WithdrawalIntents {
+    /// Collected intents for the epoch.
+    entries: Vec<SimpleWithdrawalIntentLogData>,
```
delbonis: > `LogData`

see?

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+        self.ledger.encode(enc)?;
+        Ok(())
+    }
+
+    fn decode(dec: &mut impl Decoder) -> Result<Self, CodecError> {
+        let global = GlobalStateDiff::decode(dec)?;
+        let ledger = LedgerDiff::decode(dec)?;
+        Ok(Self { global, ledger })
+    }
+}
```
delbonis: Use the codec macros.

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+
+// ============================================================================
+// Top-level DA payload
+// ============================================================================
+
+/// Versioned OL DA payload containing state diff and withdrawal intents.
+#[derive(Debug)]
+pub struct OLDaBlobV1 {
```
delbonis: "Blob" is technically wrong for this since it's decoded at this point and it's no longer a "binary large object".

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+        } else {
+            DaQueue::new()
+        };
+        // Return the new global state diff.
+        Ok(Self {
+            cur_slot,
+            pending_withdraws,
+        })
+    }
+}
```
delbonis: Use the codec compound macro.  Evgeniy's PR is merged (or close to it) which generalizes this and makes it better.

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+
+/// Target for applying a global state diff.
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct GlobalStateTarget {
+    /// Current slot value.
+    pub cur_slot: u64,
+
+    /// Pending withdrawal queue target.
+    pub pending_withdraws: PendingWithdrawQueue,
+}
```
delbonis: I think we can make the `DaWrite` trait impl generic over `IStateAccessor` (or at least the materialized SSZ global state type) so that we can apply it directly without needing this "target" that we need to do a secondary apply step with.

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+
+    fn decode(dec: &mut impl Decoder) -> Result<Self, CodecError> {
+        let new_accounts = U16LenList::<NewAccountEntry>::decode(dec)?;
+        let account_diffs = U16LenList::<AccountDiffEntry>::decode(dec)?;
+        Ok(Self {
+            new_accounts,
+            account_diffs,
+        })
+    }
+}
```
delbonis: This is a bit of a weird case because you are actually kinda defining your own DA type.  Should use bitmasks like the compound does though.

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+
+// ============================================================================
+// Ledger diff entries
+// ============================================================================
+
+/// New account initialization entry.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct NewAccountEntry {
+    /// Assigned account serial number.
+    pub serial: AccountSerial,
```
delbonis: This is purely redundant, the serial is always inferred from context because we always assign them monotonically.

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+        let serial = AccountSerial::decode(dec)?;
+        let account_id = AccountId::decode(dec)?;
+        let init = AccountInit::decode(dec)?;
+        Ok(Self {
+            serial,
+            account_id,
+            init,
+        })
+    }
+}
```
delbonis: Use the macros, Luke.

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+    }
+
+    fn decode(dec: &mut impl Decoder) -> Result<Self, CodecError> {
+        let balance = BitcoinAmount::decode(dec)?;
+        let raw_type_id = u8::decode(dec)?;
+        let type_state = match raw_type_id {
+            0 => AccountTypeInit::Empty,
+            1 => AccountTypeInit::Snark(SnarkAccountInit::decode(dec)?),
+            _ => return Err(CodecError::InvalidVariant("account_type_id")),
+        };
```
delbonis: If you impl codec on the type state then you can use the macros.

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+        self.update_vk.encode(enc)?;
+        Ok(())
+    }
+
+    fn decode(dec: &mut impl Decoder) -> Result<Self, CodecError> {
+        let initial_state_root = Hash::decode(dec)?;
+        let update_vk = U16LenBytes::decode(dec)?;
+        if update_vk.as_slice().len() > MAX_VK_BYTES {
+            return Err(CodecError::OverflowContainer);
+        }
```
delbonis: Here's actually a case where *maybe* we could use varint lengths instead of u16.

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+
+    fn decode(dec: &mut impl Decoder) -> Result<Self, CodecError> {
+        let account_serial = AccountSerial::decode(dec)?;
+        let diff = AccountDiff::decode(dec)?;
+        Ok(Self {
+            account_serial,
+            diff,
+        })
+    }
+}
```
delbonis: :)

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+// ============================================================================
+
+/// Per-account diff (balance + snark state).
+#[derive(Debug)]
+pub struct AccountDiff {
+    /// Balance register diff.
+    pub balance: DaRegister<BitcoinAmount>,
+
+    /// Snark state diff.
+    pub snark_state: SnarkAccountDiff,
```
delbonis: Make this an enum with a type ID so that we don't have to horribly break the DA format when we add new account types.

*Technically* we could get away with saving a byte if codec had a context param that let us query the account type, but that would complicate our ability to change account types.

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+            update_vk,
+        })
+    }
+}
+
+/// Per-account diff entry keyed by account serial.
+#[derive(Debug)]
+pub struct AccountDiffEntry {
+    /// Account serial number.
+    pub account_serial: AccountSerial,
```
delbonis: We'd want to pass this to the DA context in `poll_context` so that we can prefetch the accounts we need.

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+            SnarkAccountDiff::decode(dec)?
+        } else {
+            <SnarkAccountDiff as Default>::default()
+        };
+        Ok(Self {
+            balance,
+            snark_state,
+        })
+    }
+}
```
delbonis: compounds!

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+
+/// Target for applying an account diff.
+#[derive(Debug, Default)]
+pub struct AccountDiffTarget {
+    /// Current balance value.
+    pub balance: BitcoinAmount,
+
+    /// Snark account target.
+    pub snark: SnarkAccountTarget,
+}
```
delbonis: Same as above.  See if we can make it generic over `IAccountState` or at least `OLAccountState` or whatever the types are.

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+        Self::decode(dec)
+    }
+
+    fn encode_set(&self, enc: &mut impl Encoder) -> Result<(), CodecError> {
+        if CompoundMember::is_default(self) {
+            return Err(CodecError::InvalidVariant("snark_account_diff"));
+        }
+        self.encode(enc)
+    }
+}
```
delbonis: ahhhh

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+
+    fn decode(dec: &mut impl Decoder) -> Result<Self, CodecError> {
+        let inner_state_root = Hash::decode(dec)?;
+        let next_msg_read_idx = u64::decode(dec)?;
+        Ok(Self {
+            inner_state_root,
+            next_msg_read_idx,
+        })
+    }
+}
```
delbonis: :)

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+
+/// Proof state snapshot used in DA diffs.
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct DaProofState {
+    /// Inner state root commitment.
+    pub inner_state_root: Hash,
+
+    /// Next message read index.
+    pub next_msg_read_idx: u64,
```
delbonis: should be a counter

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+
+// ============================================================================
+// Inbox accumulator types
+// ============================================================================
+
+/// DA-encoded snark inbox message entry.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct DaMessageEntry {
+    /// Source account for the message.
+    pub source: AccountId,
```
delbonis: Should be encoded in DA as the serial, and then use the context to lookup the account ID when actually converting to `MessageEntry` in `apply`.

crates/ol/da/src/types.rs
```diff
@@ -0,0 +1,1055 @@
...
+
+impl LinearAccumulator for InboxAccumulator {
+    type InsertCnt = u16;
+    type EntryData = DaMessageEntry;
+    const MAX_INSERT: Self::InsertCnt = u16::MAX;
+
+    fn insert(&mut self, entry: &Self::EntryData) {
+        self.entries.push(entry.clone());
+    }
+}
```
delbonis: This name is misleading, it isn't actually the accumulator, it's a buffer for messages inserted into the accumulator.

As with the other cases, we might be able to implement this over `ISnarkAccountState` to be able to insert it into the accumulator with normal message insertion function.  It shouldn't be needed to have a second type like this that effectively just copies the data out of the DA types.

crates/ol/da/src/types.rs
delbonis, line 1: This module is kinda huge and it seems like it could make sense to break it apart into separate modules.  Like `snark_acct_types`, for example, maybe.

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -0,0 +1,774 @@
...
+        self.touched_accounts.clear();
+        self.pre_states.clear();
+        self.inbox_messages.clear();
+        self.snark_updates.clear();
+        self.withdrawal_intents.clear();
+        self.pending_withdraw_source = PendingWithdrawQueue::default();
+        self.pending_withdraw_front_incr = 0;
+        self.output_logs.clear();
+        self.last_error = None;
+    }
```
delbonis: Eek this feels like C++, is this necessary?  Can't we just naturally throw this instance away and construct a new one?

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -0,0 +1,774 @@
...
+    fn finalize_epoch_blob(&mut self) {
+        match self.epoch_acc.finalize(&self.inner) {
+            Ok(blob) => self.epoch_acc.completed_blobs.push_back(blob),
+            Err(err) => {
+                #[cfg(feature = "tracing")]
+                error!(error = %err, "failed to finalize epoch DA blob");
+                self.epoch_acc.last_error = Some(err);
+            }
+        }
+    }
```
delbonis: Why does this track multiple blobs?  We should have a "finish" operation that unwraps the layer and returns the materialized DA structure that a consumer can then (first sanity check by reapplying it to the state in a write batch layer, and then) encode as they need it.

Also, why if there was an error it should return it right?

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -0,0 +1,774 @@
...
+
+        if self.da_tracking_enabled {
+            self.epoch_acc.record_output_logs(logs);
+        }
+    }
+
+    fn set_da_tracking_enabled(&mut self, enabled: bool) {
+        if self.da_tracking_enabled && !enabled {
+            self.finalize_epoch_blob();
+        }
```
delbonis: This is *really* scary that this kind of blob cutting is being done with contextual hint calls like this!

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -0,0 +1,774 @@
...
+    pending_withdraw_front_incr: u16,
+
+    /// Output logs emitted during the epoch.
+    output_logs: Vec<OLLog>,
+
+    /// Completed DA blobs waiting to be consumed.
+    completed_blobs: VecDeque<Vec<u8>>,
+
+    /// Last error encountered while building a blob.
+    last_error: Option<DaAccumulationError>,
```
delbonis: [this feels like](https://man7.org/linux/man-pages/man3/errno.3.html)

crates/ledger-types/src/state_accessor.rs
storopoli: Removed in a913ca0e96e2b78a40d0e8b45f3e17cddd5e27b7

crates/ledger-types/src/state_accessor.rs
storopoli: Removed in a913ca0e96e2b78a40d0e8b45f3e17cddd5e27b7

crates/ol/da/src/types.rs
storopoli: Removed in a913ca0e96e2b78a40d0e8b45f3e17cddd5e27b7

crates/ol/da/src/types.rs
storopoli: Removed in a913ca0e96e2b78a40d0e8b45f3e17cddd5e27b7

crates/ol/da/src/types.rs
storopoli: Removed in a913ca0e96e2b78a40d0e8b45f3e17cddd5e27b7

crates/ol/stf/src/assembly.rs
storopoli: Removed in a913ca0e96e2b78a40d0e8b45f3e17cddd5e27b7

crates/ol/state-support-types/src/da_accumulating_layer.rs
storopoli: Done in 27fc93f67b327bc59307dbd3034591773d128418
We now panic

crates/ol/state-support-types/src/da_accumulating_layer.rs
storopoli: Done in 27fc93f67b327bc59307dbd3034591773d128418

crates/ol/state-support-types/src/da_accumulating_layer.rs
storopoli: Done in 27fc93f67b327bc59307dbd3034591773d128418

crates/ol/stf/src/assembly.rs
storopoli: Done in 27fc93f67b327bc59307dbd3034591773d128418

crates/ol/state-types/src/snark_account.rs
storopoli: Done in e5b6d16c3289067854b14eac7cbe21d695e01096

crates/ol/da/src/types.rs
storopoli: Ok now it's `OLDaPayloadV1`.

Check 498033cd7f6b3e97d72ebdb4be9918d5db20cbd6

Feel free to bikeshed during this PR we can always trivially rename it with `rust-analyzer`.

crates/ol/da/src/types.rs
storopoli: Done in 81fc64722bd32467b27d72febf3d6568215d9a57

crates/ol/da/src/types.rs
storopoli: Done in 81fc64722bd32467b27d72febf3d6568215d9a57

crates/ol/da/src/types.rs
storopoli: Done in 81fc64722bd32467b27d72febf3d6568215d9a57

See https://github.com/alpenlabs/alpen/blob/81fc64722bd32467b27d72febf3d6568215d9a57/crates/ol/da/src/types.rs#L228-L259

crates/ol/state-support-types/src/da_accumulating_layer.rs
storopoli: Suggestions?

crates/ol/da/src/types.rs
storopoli: Broke down starting from 9fbd6bf95ac62e14b85c1055347d6649628c5995

crates/ol/da/src/types.rs
storopoli: Done in 9fbd6bf95ac62e14b85c1055347d6649628c5995

crates/ol/da/src/types.rs
storopoli: Done 81eb8d9b53a63a1e575c4ae87c0478679408bd04

crates/ol/da/src/types.rs
storopoli: Done 81eb8d9b53a63a1e575c4ae87c0478679408bd04

crates/ol/da/src/types.rs
storopoli: Done 81eb8d9b53a63a1e575c4ae87c0478679408bd04

crates/ol/da/src/types.rs
storopoli: Done 81eb8d9b53a63a1e575c4ae87c0478679408bd04

crates/ol/da/src/types.rs
storopoli: Done in 81eb8d9b53a63a1e575c4ae87c0478679408bd04

crates/ol/da/src/types.rs
storopoli: 81eb8d9b53a63a1e575c4ae87c0478679408bd04

crates/ol/da/src/types.rs
storopoli: Done in 00476932082219af80c2ed75c50d2676bb70386a

crates/ol/da/src/types.rs
storopoli: Done in f4dde338543e94bc84ee5e6d8e35c793b805e7ec

crates/ol/da/src/types.rs
storopoli: Done 3a53db99cbc9d73f1f1684b0ca306d47f76d8c42

crates/ol/state-support-types/src/da_accumulating_layer.rs
storopoli: Ok I've removed this in 3a53db99cbc9d73f1f1684b0ca306d47f76d8c42 but I have to give up the `make_compound_impl!` macro

crates/ol/da/src/types.rs
storopoli: Ok done in f5bf7ee0e41ac51a3d4069910521ba53a9986cf8

crates/da-framework/src/compound.rs
```diff
@@ -273,6 +273,13 @@ macro_rules! _mct_field_decode {
     ($reader:ident $dec:ident; counter { $fty:ty }) => {
         $reader.decode_next_member::<$crate::DaCounter<$fty>>($dec)?
     };
+    // Compound member - type implements CompoundMember directly
+    ($reader:ident $dec:ident; compound ( $fty:ty )) => {
+        $reader.decode_next_member::<$fty>($dec)?
+    };
+    ($reader:ident $dec:ident; compound { $fty:ty }) => {
+        $reader.decode_next_member::<$fty>($dec)?
+    };
```
delbonis, line 371: I might be misunderstanding something but why do we need two of these if they have the same inputs and expand to the same syntax?

Same for the others...

crates/ledger-types/src/state_accessor.rs
```diff
@@ -33,6 +33,12 @@ pub trait IStateAccessor {
     /// Sets the current epoch.
     fn set_cur_epoch(&mut self, epoch: u32);
 
+    /// Signals that epoch sealing (L1-derived updates) is about to begin.
+    ///
+    /// Implementations that track DA-covered writes can use this to finalize
+    /// the preseal diff and suspend tracking for sealing updates.
+    fn begin_epoch_sealing(&mut self) {}
```
delbonis: This makes the abstractions leakier.  We can avoid needing hint functions like this by just not doing the sealing phase when accumulating DA related changes.

crates/ol/da/src/types/payload.rs
```diff
@@ -0,0 +1,313 @@
...
+            apply_account_diff(target, account_id, &entry.diff)?;
+        }
+        Ok(())
+    }
+}
+
+fn validate_ledger_entries(
+    pre_state_next_serial: AccountSerial,
+    diff: &StateDiff,
+) -> Result<(), DaError> {
```
delbonis, line 197: What property is being validated here?  I'm not sure I understand what's going on.

crates/ol/da/src/types/payload.rs
```diff
@@ -0,0 +1,313 @@
...
+    let mut last_serial: Option<u32> = None;
+    for entry in diff.ledger.account_diffs.entries() {
+        let serial: u32 = entry.account_serial.into();
+        if serial >= pre_serial {
+            return Err(DaError::InsufficientContext);
+        }
+        if let Some(prev) = last_serial
+            && serial <= prev
+        {
+            return Err(DaError::InsufficientContext);
```
delbonis: How can we error with `InsufficientContext` here when there is no context arg being used by this function that can be insufficient?  Should we add a new error variant?

crates/ol/da/src/types/ledger.rs
```diff
@@ -0,0 +1,194 @@
...
+}
+
+/// Snark account initialization data.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct SnarkAccountInit {
+    /// Initial inner state root.
+    pub initial_state_root: Hash,
+
+    /// Update verification key bytes.
+    pub update_vk: U16LenBytes,
```
delbonis: Can this just be `VarVec`?

crates/ol/da/src/types/payload.rs
```diff
@@ -0,0 +1,313 @@
...
+        Self { new_account_data }
+    }
+
+    pub fn new_account_data(
+        &self,
+        init: &AccountInit,
+    ) -> Result<NewAccountData<S::AccountState>, DaError> {
+        (self.new_account_data)(init)
+    }
+}
```
delbonis: Can you expand on what this is doing?  I don't see a reason why we would *necessarily* need to be generic over how accounts are constructed *like this*.  There's [`IAccountStateConstructible`](https://github.com/alpenlabs/alpen/blob/main/crates/ledger-types/src/account.rs#L173) to help with this, and honestly maybe we could move some of that functionality into `IAccountState` itself in the first place.

But what *is* missing here is the kind of thing I *would* expect, a mapping of serial->ID.  That's the *only* thing I expect would be provided.  And all `poll_context` would need to do is call a function to inspect those.

crates/ol/da/src/types/payload.rs
```diff
@@ -0,0 +1,313 @@
...
+        acct.add_balance(coin);
+    } else if new_balance < current {
+        let delta = current
+            .checked_sub(new_balance)
+            .ok_or(DaError::InsufficientContext)?;
+        acct.take_balance(delta)
+            .map_err(|_| DaError::InsufficientContext)?;
+    }
+    Ok(())
+}
```
delbonis, line 278: Okay maybe we're trying a little too hard here and we should expose more appropriate accessors on the account types to make this work.

Also, again, using `InsufficientContext` for this doesn't make any sense.

crates/ol/da/src/types/payload.rs
```diff
@@ -0,0 +1,313 @@
...
+fn apply_snark_diff<T: IAccountStateMut>(
+    acct: &mut T,
+    diff: &SnarkAccountDiff,
+) -> Result<(), DaError> {
+    if diff.seq_no.diff().is_none()
+        && diff.proof_state.new_value().is_none()
+        && diff.inbox.new_entries().is_empty()
+    {
+        return Ok(());
+    }
```
delbonis: This conditional should be unnecessary.

crates/ol/da/src/types/snark.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+
+/// Target for applying a snark account diff.
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct SnarkAccountTarget {
+    pub seq_no: u64,
+    pub proof_state: DaProofState,
+    pub inbox: InboxBuffer,
+}
```
delbonis: Is there no way to get rid of this?

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -0,0 +1,761 @@
...
+}
+
+/// Per-epoch accumulator of DA writes before encoding.
+#[derive(Default, Debug)]
+struct EpochDaAccumulator {
+    /// Slot value at the start of the epoch.
+    slot_base: Option<u64>,
+
+    /// Final slot value seen during the epoch.
+    slot_final: Option<u64>,
```
delbonis: This is just duplicating the work that the register/counter builder type does.

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -0,0 +1,761 @@
...
+        };
+
+        Ok(GlobalStateDiff::new(cur_slot))
+    }
+
+    /// Builds the ledger diff for the epoch.
+    fn build_ledger_diff<S: IStateAccessor>(
+        &self,
+        state: &S,
+    ) -> Result<LedgerDiff, DaAccumulationError> {
```
delbonis, line 353: This is doing a huge amount of work that I would not expect it would need to do if we kept better bookkeeping while we were processing the writes.

This seems like it's going back over the set of touched accounts and then trying to reconstruct the diff retroactively instead of tracking the changes as they're being made, which is more like how the builder types behave.

(also applies to `build_snark_diff`)

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -0,0 +1,761 @@
...
+}
+
+/// State accessor that accumulates DA-covered writes for a single epoch.
+#[derive(Debug)]
+pub struct DaAccumulatingState<S: IStateAccessor> {
+    /// Wrapped state accessor.
+    inner: S,
+
+    /// Toggle for recording DA-covered writes.
+    da_tracking_enabled: bool,
```
delbonis: As a heuristic, any design that needs a flag like this (unless it's just a sanity check thing, not for correctness) should be iterated on.

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -0,0 +1,761 @@
...
+    #[cfg(test)]
+    pub fn post_seal_writes_detected(&self) -> bool {
+        self.post_seal_writes
+    }
+
+    fn record_post_seal_write(&mut self) {
+        if !self.da_tracking_enabled {
+            self.post_seal_writes = true;
+        }
+    }
```
delbonis: Why do we need to record this?

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -0,0 +1,761 @@
...
+        }
+
+        #[cfg(feature = "tracing")]
+        if !self.post_seal_writes_reported {
+            tracing::warn!(context, "post-seal writes detected");
+        }
+        self.post_seal_writes_reported = true;
+
+        Err(DaAccumulationError::PostSealWrites { context })
+    }
```
delbonis: We don't want to ensure this, we *expect* this to happen, for things like deposits.

crates/da-framework/src/compound.rs
storopoli: The don't have the same inputs!

The problem is that `make_compound_impl!` syntax already allows field types to be wrapped either way; e.g. `register (Foo)` or `register {Foo}`.

Hence we need this to normalize either form in the inner type.

I've refactored this in 9a95ed87dbdfffb3f59fc0adc3735c24830b3adf to make it easier, but the behavior is the same.

crates/ledger-types/src/state_accessor.rs
storopoli: Removed in b2a9073bdc116f610bbf9c68a73d6a88a70f768e

crates/ol/da/src/types/payload.rs
storopoli: See 7975e07edbfdb199aeb5bc0666b748e22a0e846e. It has more error variants with proper error messages.

crates/ol/da/src/types/payload.rs
storopoli: Added more error variants in 7975e07edbfdb199aeb5bc0666b748e22a0e846e: `InvalidStateDiff` and `InvalidLedgerDiff`

crates/ol/da/src/types/ledger.rs
storopoli: Kept `U16LenBytes` for `update_vk` since SPS‑ol‑da‑structure specifies a `u16` length prefix; added an inline doc note. See 9a95ed87dbdfffb3f59fc0adc3735c24830b3adf

crates/ol/da/src/types/payload.rs
storopoli: See 7975e07edbfdb199aeb5bc0666b748e22a0e846e. It has more error variants with proper error messages.

crates/ol/da/src/types/payload.rs
storopoli: Removed in 7975e07edbfdb199aeb5bc0666b748e22a0e846e

crates/ol/da/src/types/snark.rs
storopoli: It is hard. `SnarkAccountTarget` remains the `DaWrite::Target` needed by the compound macro; doc expanded to clarify usage in 9a95ed87dbdfffb3f59fc0adc3735c24830b3adf

crates/ol/state-support-types/src/da_accumulating_layer.rs
storopoli: Removed for a single `slot_builder: Option<DaCounterBuilder<CtrU64ByU16>>` in b2a9073bdc116f610bbf9c68a73d6a88a70f768e

crates/ol/state-support-types/src/da_accumulating_layer.rs
storopoli: Refactored in b2a9073bdc116f610bbf9c68a73d6a88a70f768e

crates/ol/state-support-types/src/da_accumulating_layer.rs
storopoli: Removed in b2a9073bdc116f610bbf9c68a73d6a88a70f768e

crates/ol/state-support-types/src/da_accumulating_layer.rs
storopoli: Removed in b2a9073bdc116f610bbf9c68a73d6a88a70f768e

crates/ol/state-support-types/src/da_accumulating_layer.rs
storopoli: Removed in b2a9073bdc116f610bbf9c68a73d6a88a70f768e

crates/ol/da/src/types/payload.rs
storopoli: Ok this was done in:
- 90d25fa08c2f4d878df5e0a94b27dcd30e1e69c8
- e8221c6f3a0b9c7e3da1aaee2d14e2dabdba3ad4
- ec2e3e09d16f9f5e44601fb7785e0111581429d3
- 0c8a183b978c85a17bdb8635f13a734495a99055

crates/ol/da/src/types/ledger.rs
delbonis: ah shit!

storopoli: We can change the spec if you think that we can go with variable length.

crates/da-framework/src/compound.rs
evgenyzdanovich: I think we can remove either `{` or `(` since the semantic is the same.

crates/da-framework/src/errors.rs
```diff
@@ -7,6 +7,12 @@ use thiserror::Error;
 pub enum DaError {
```
evgenyzdanovich, line 7: this is a struct that's intended to describe Errors from logic inside `da-framework`. 
You defined error variants, but nothing has changed in da-framework (nothing throws those variants) and I see error variants are used in ol-da :( 

Let's keep it clean and not introduce "business logic" errors to the "framework". OL DA likely would need to create its own errors and return it (and ofc handle framework error(s) -> business logic erros conversion).

crates/ol/da/src/types/account.rs
```diff
@@ -0,0 +1,57 @@
...
+use super::snark::{SnarkAccountDiff, SnarkAccountTarget};
+
+/// Per-account diff keyed by account type.
+///
+/// The account type is implied by pre-state; the snark field is only populated
+/// for snark accounts.
+#[derive(Debug)]
+pub struct AccountDiff {
+    /// Balance register diff.
+    pub balance: DaRegister<BitcoinAmount>,
```
evgenyzdanovich, line 15: here (and mentioned in the doc) - it's likely better to make it counter and have values "diffed".

crates/ol/da/src/types/payload.rs
evgenyzdanovich: (mentioned elsewhere) it's probably better to define ol-da specific error, and return it (instead of da-framework error).

crates/da-framework/src/errors.rs
storopoli: Yes, makes total sense. But I had to change `DaWrite` to have generic `Error`. Done in b3396623a610e3ad4b05144be14daa0a7fe70fee

crates/ol/da/src/types/payload.rs
storopoli: Done in b3396623a610e3ad4b05144be14daa0a7fe70fee. Had to make `DaWrite` error generic.

crates/da-framework/src/compound.rs
storopoli: Removed in b3396623a610e3ad4b05144be14daa0a7fe70fee

crates/ol/da/src/types/account.rs
storopoli: Done in 3cbffb0bb1cc9d83cbc4a9916335a4dc2d22f498 using a new `CtrU64BySignedVarint`

storopoli: Should we drop that commit and use the new `CtrU256BySignedU256` that you are proposing in #1285 and move it to `da-framework/src/counter.rs`?

Cc @delbonis

storopoli: Ok pretty much you're already implemented `CtrU64BySignedVarint` in 2d4c8d0d7b40fb0647559f7cfe967de09101a922 in #1285
I think it makes sense to drop my commit in this PR rebase and use it once #1285 is merged

delbonis: I don't think it'd be a problem to put `CtrU256BySignedU256` or similar in the codec crate, although it would make sense to use here since we can only use 64 bits for sats.

Like discussed in the spec doc, I don't think using varints makes sense here since most balance changes are >32k (or would it be 16k with signed?).  It might be worth using an `i48`-like type since that would allow us to express balance changes of 1.4M BTC with only 6 bytes, which is a comfortable range and saves us 2 bytes.

The only other thing in this part of the spec is that it would be *great* to consolidate the bitmaps for the different parts of account state into the same bitmap.

storopoli: Reverted back the commit

crates/checkpoint-types-ssz/src/error.rs
```diff
@@ -8,9 +8,15 @@ pub enum CheckpointPayloadError {
     #[error("state diff too large: {provided} bytes (max {max})")]
     StateDiffTooLarge { provided: u64, max: u64 },
 
-    #[error("OL logs too large: {provided} bytes (max {max})")]
+    #[error("OL logs count too large: {provided} (max {max})")]
     OlLogsTooLarge { provided: u64, max: u64 },
```
krsnapaudel: `OLLogsTooLarge`

crates/checkpoint-types-ssz/src/error.rs
```diff
@@ -8,9 +8,15 @@ pub enum CheckpointPayloadError {
...
 
+    #[error("OL log payload too large: {provided} bytes (max {max})")]
+    OlLogPayloadTooLarge { provided: u64, max: u64 },
```
krsnapaudel: `OLLogPayloadTooLarge`

crates/checkpoint-types-ssz/src/error.rs
```diff
@@ -8,9 +8,15 @@ pub enum CheckpointPayloadError {
...
+
+    #[error("OL logs total payload too large: {provided} bytes (max {max})")]
+    OlLogsTotalPayloadTooLarge { provided: u64, max: u64 },
```
krsnapaudel: `OLLogsTotalPayloadTooLarge`

crates/checkpoint-types-ssz/src/payload.rs
```diff
@@ -110,3 +139,40 @@ impl SignedCheckpointPayload {
...
+
+#[cfg(test)]
+mod tests {
+    use strata_identifiers::AccountSerial;
+    use strata_ol_chain_types_new::OLLog;
+
+    use super::*;
+
+    #[test]
+    fn test_checkpoint_sidecar_rejects_oversize_log_payload() {
```
krsnapaudel, line 157: Should these tests use proptest?

crates/ol/da/src/types/snark.rs
```diff
@@ -0,0 +1,255 @@
...
+        let next_msg_read_idx = u64::decode(dec)?;
+        Ok(Self::new(inner_state_root, next_msg_read_idx))
+    }
+}
+
+/// Diff for proof state (inner state root + next inbox read index).
+#[derive(Clone, Debug)]
+pub struct DaProofStateDiff {
+    pub inner_state: DaRegister<Hash>,
+    pub next_inbox_msg_idx: DaCounter<counter_schemes::CtrU64ByUnsignedVarint>,
```
krsnapaudel, line 69: This is different from the spec: `CtrU64ByVarint`

crates/ol/da/src/types/snark.rs
```diff
@@ -0,0 +1,255 @@
...
+    fn default() -> Self {
+        Self {
+            inner_state: DaRegister::new_unset(),
+            next_inbox_msg_idx: DaCounter::new_unchanged(),
+        }
+    }
+}
+
+impl Codec for DaProofStateDiff {
+    fn decode(dec: &mut impl Decoder) -> Result<Self, CodecError> {
```
krsnapaudel, line 94: Can we use this macro version?
```
`make_compound_impl! {
    DaProofStateDiff < (), crate::DaError > u8 => DaProofState {
        inner_state: register (Hash),
        next_inbox_msg_idx: counter (counter_schemes::CtrU64ByUnsignedVarint),
    }
}
```

crates/ol/da/src/types/payload.rs
```diff
@@ -0,0 +1,325 @@
...
+        snark
+            .insert_inbox_message(msg)
+            .map_err(|_| DaError::InvalidStateDiff("failed to insert inbox message"))?;
+    }
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
```
krsnapaudel, line 312: Can we improve test coverage for any of these?

`OLStateDiff::poll_context()` - validates diff before application
`OLStateDiff::apply()` - applies diff to state
`validate_ledger_entries()` - validates ordering/uniqueness rules
`apply_account_diff() / apply_balance() / apply_snark_diff()` - applies per-account diffs

crates/ol/state-types/ssz/state.ssz
```diff
@@ -6,6 +6,8 @@ import strata_snark_acct_types
 
 MAX_LEDGER_ACCOUNTS = 1 << 24
 MAX_ACCOUNT_SERIALS = 1 << 24
+### Max VK bytes encoded with a u16 length prefix (65_535 bytes).
+MAX_VK_BYTES = 65535
```
irnb: nit: `MAX_VK_BYTES` looks like dead code here, can we remove it?  

The Rust equivalent in `crates/ol/da/src/types/mod.rs` is used for DA payload validation.

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -0,0 +1,709 @@
...
+    #[error("da accumulator builder error: {0}")]
+    Builder(#[from] strata_da_framework::BuilderError),
+
+    /// Error while encoding DA blob.
+    #[error("da accumulator codec error: {0}")]
+    Codec(#[from] CodecError),
+
+    /// Account state missing when assembling diffs.
+    #[error("da accumulator missing account {0}")]
+    MissingAccount(AccountId),
```
irnb: nit: MissingAccount → MissingAccountState?

irnb (approved):
LGTM! 🚀

crates/checkpoint-types-ssz/src/error.rs
storopoli: Check c30acb62fc8918c5aba34bc1e9aa570d69adf6fe

crates/checkpoint-types-ssz/src/error.rs
storopoli: c30acb62fc8918c5aba34bc1e9aa570d69adf6fe

crates/ol/da/src/types/snark.rs
storopoli: Yes, we should update the spec. This is different because we have unsigned and signed varints increments now.

crates/ol/da/src/types/payload.rs
storopoli: Check 1638a2f838371ca05bb5f747367a6b632c9c7b25

crates/ol/state-support-types/src/da_accumulating_layer.rs
storopoli: See 817845dab7148f0c44ee689dd51f1998b11d36c8

crates/ol/state-types/ssz/state.ssz
storopoli: See 8edfb51f1785e1829c4b16c4bd7efa20f783a190

crates/checkpoint-types-ssz/src/error.rs
storopoli: c30acb62fc8918c5aba34bc1e9aa570d69adf6fe

crates/checkpoint-types-ssz/src/payload.rs
storopoli: I kept these as boundary unit tests since they only validate max-size rejection; proptest felt like overkill here. Happy to switch if you prefer property coverage.

crates/ol/da/src/types/snark.rs
storopoli: I kept the manual impl because `DaProofState` is a wrapper without field access, so `make_compound_impl!` can’t apply cleanly without introducing a new target struct or refactor.

crates/ol/da/src/types/snark.rs
storopoli: Updated the spec

PR #1312, opened 2026-01-23
STR-1918 EE chunk runtime and snark account runtime refactors

delbonis:
## Description

The goal of this PR is to do some of the hairy refactors to support a chunk-based execution domain snark account runtime.

There are two parts:

- [x] Create new ee-chunk-runtime crate to support this pattern.
- [x] Refactor existing `ee-acct-runtime` crate to support its interfaces.

I used Claude for some refactoring to split out the functionality in `snark-acct-runtime` from `ee-acct-runtime` and retarget it to use it via the trait system.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

- [x] I have [disclosed my (lack of) use of AI](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

### DIFF
--- a/crates/alpen-ee/block-assembly/src/package.rs
+++ b/crates/alpen-ee/block-assembly/src/package.rs
@@ -16,7 +16,7 @@ use tracing::warn;
 /// Builds [`ExecInputs`] from parsed input messages.
 ///
 /// Only `Deposit` messages are processed; other message types are logged and ignored.
-// TODO convert this to do it based on extracting pending inputs from the
+// TODO(STR-2583) convert this to do it based on extracting pending inputs from the
 // current EE account inner state
 pub(crate) fn build_block_inputs(
     parsed_inputs: Vec<InputMessage<DecodedEeMessageData>>,


crates/ee-acct-runtime/src/traits.rs
```diff
@@ -0,0 +1,106 @@
...
+pub enum ProgramError {
+    #[error("invalid coinput")]
+    InvalidCoinput,
+
+    #[error("malformed coinput")]
+    MalformedCoinput,
+
+    #[error("obligations unsatisfied after update finished processing")]
+    UnsatisfiedObligations,
+}
```
storopoli: Why errors in a `traits.rs` file? Should we have an `errors.rs`?

crates/ee-acct-runtime/src/traits.rs
```diff
@@ -0,0 +1,106 @@
...
+
+pub type ProgramResult<T> = Result<T, ProgramError>;
+
+pub trait SnarkAccountProgram {
+    /// Account inner state.
+    type State: IInnerState;
+
+    /// Temporary state that can be modified while processing coinputs but isno .
```
storopoli: But what?

crates/ee-acct-runtime/src/traits.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// Recognized messages.
+    type Msg: IAcctMsg;
+
+    /// Update extra data.
+    type ExtraData: IExtraData;
+
+    /// Starts an update, also producing a verification state we use while
+    /// processing coinputs.
+    ///
+    /// The result is discarded if we're not in a proof.
```
storopoli: How would you enforce that in the trait itself? Or are you just relying on wishful thinking of implementors?

crates/ee-acct-runtime/src/traits.rs
```diff
@@ -0,0 +1,106 @@
...
+    #[error("malformed coinput")]
+    MalformedCoinput,
+
+    #[error("obligations unsatisfied after update finished processing")]
+    UnsatisfiedObligations,
+}
+
+pub type ProgramResult<T> = Result<T, ProgramError>;
+
+pub trait SnarkAccountProgram {
```
storopoli: I really like the distinction between "state" (`State`) and "temporary state" (`VState`) here and the interplay with the different methods that take/consume (mutable) references on them depending on the context and the expected behavior.

Amazing work with abstracting and encapsulating behavior.

crates/ee-acct-runtime/src/traits.rs
```diff
@@ -0,0 +1,106 @@
...
+        state: &mut Self::State,
+        extra_data: Self::ExtraData,
+    ) -> ProgramResult<()>;
+}
+
+/// Trait describing the program state.
+pub trait IInnerState: Clone + Codec + 'static {
+    /// Computes a commitment to the inner state.
+    fn compute_state_root(&self) -> Hash;
+}
```
storopoli: Shouldn't you just extend `IAccountState`?

https://github.com/alpenlabs/alpen/blob/c14212fa2dc826b103e732812ecd3b2efaaade05/crates/ledger-types/src/account.rs#L7-L26

crates/ee-acct-runtime/src/traits.rs
```diff
@@ -0,0 +1,106 @@
...
+///
+/// This should probably be implemented on an enum.
+pub trait IAcctMsg: Clone + Codec + 'static {
+    // TODO
+}
+
+/// Trait describing the extra data processed by the snark account.
+pub trait IExtraData: Clone + Codec + 'static {
+    // TODO
+}
```
storopoli: If you decide to extend `IAccountState` as questioned above, then you should move these also to somewhere in `crates/ledger-types/src`?

crates/ee-chain-types/ssz/io.ssz
```diff
@@ -0,0 +1,35 @@
+import strata_acct_types
+
+MAX_SUBJECT_DEPOSITS = 1 << 16
+MAX_OUTPUT_TRANSFERS = 1 << 16
+MAX_OUTPUT_MESSAGES = 1 << 16
```
storopoli, line 7: Where does these come from? If from an SPS add it as a comment. Otherwise add the logic behind this as a comment.

crates/ee-chain-types/src/sequence_tracker.rs
```diff
@@ -0,0 +1,216 @@
...
+#[derive(Debug, Error)]
+pub enum SeqError {
+    #[error("inconsistent entry at index {0}")]
+    Mismatch(usize),
+
+    #[error("consumed all entries but checked another")]
+    Overrun,
+}
+
+pub type SeqResult<T> = Result<T, SeqError>;
```
storopoli, line 16: Should we move this to `errors.rs`? Or `error.rs`?

crates/ee-chain-types/src/sequence_tracker.rs
```diff
@@ -0,0 +1,216 @@
...
+
+/// Tracks a list of entries that we want to compare a sequence of them against.
+///
+/// We use this for processing pending inputs while processing notpackages.
+#[derive(Debug)]
+pub struct SequenceTracker<'a, T> {
+    expected_inputs: &'a [T],
```
storopoli, line 23: We should either guarantee somehow that this list will always have the right sorting; or clarify in the docstrings.

crates/ee-chunk-runtime/src/runtime.rs
```diff
@@ -0,0 +1,56 @@
...
+use strata_ee_acct_types::{EnvError, EnvResult, ExecutionEnvironment};
+
+use crate::{ArchivedPrivateInput, Chunk, ChunkBlock, verify_chunk_transition};
+
+/// Verifies the private input's consistency using the provided execution
+/// environment.
+pub fn verify_input<E: ExecutionEnvironment>(
+    ee: &E,
+    input: &ArchivedPrivateInput,
+) -> EnvResult<()> {
```
storopoli, line 12: Should we also validate intermediate header linkage or per-block state roots?

crates/ee-acct-runtime/src/traits.rs
delbonis: These are not part of the OL state, this is strictly internal to the snark account program.

crates/ee-acct-runtime/src/traits.rs
delbonis: It's just a note for the trait implementor.  It shouldn't actually matter.  Actually I should clarify the language slightly better.

crates/ee-acct-runtime/src/traits.rs
delbonis: I was throwing this together very fast at the end of the day on a friday. :)

crates/ee-chain-types/ssz/io.ssz
delbonis: I think it's from the spec?  I just copied these from the other file when I was generalizing these types.

crates/ee-chain-types/src/sequence_tracker.rs
delbonis: Actually since these are closely coupled to this specific bookkeeping data structure I wouldn't normally put it in with the more general crate-level errors.

crates/ee-chain-types/src/sequence_tracker.rs
delbonis: I'm not sure how we'd do that.  This is meant to be a helper type where the caller just dictates the ordering.

crates/ee-chunk-runtime/src/runtime.rs
delbonis: Maybe, I haven't fully thought through those details yet.

There was some discussion and it seemed that we actually don't want to verify the per-block state roots because that's expensive to do, we just want to verify the one at the end since that's the only one that *really* matters.

We can mitigate the kinds of games a misbehaving sequencer can play later on when we have sequencer rotation and PoS by actually validating the state roots when deciding what the next block is and not voting for blocks that are invalid by the full rules.

crates/ee-acct-runtime/src/traits.rs
storopoli: Ok makes sense

crates/ee-chunk-runtime/src/runtime.rs
storopoli: Ok if you find that important, please open a ticket.

delbonis:
Ok this is still in a *pretty* rough state, and there's some trait/lifetime issues I was partway through working out, but it captures the general idea well enough that I wanted to share it so that others can comment on it.

This makes the (new) snark account runtime more general purpose and provides a more narrow structure for the EE logic to slot into and be focused on processing chunk proofs in a structured way.  There's still some organizational issues with where modules life, but this does capture the general approach that we need to go with here.

I figured it would becoming untenably complex to keep modifying the integrated EE account runtime to support the multiple proofs involved in separate proving of chunks, so formalizing the architecture and cleanly separating things with a trait really helps with ensuring that things behave correctly.

(cc @storopoli @purusang)

crates/ee-acct-runtime/src/errors.rs
```diff
@@ -0,0 +1,39 @@
...
+    #[error("malformed extradata")]
+    MalformedExtraData,
+
+    #[error("malformed extradata")]
+    InvalidExtraData,
+
+    /// When we reach the end of processing and still have some unsatisfied
+    /// obligation to verify.
+    #[error("obligations unsatisfied after update finished processing")]
+    UnsatisfiedObligations,
```
storopoli: These are really opaque errors, should they carry any relevant data for debugging?

delbonis: We can emit more structured errors from the snark account program when we have that context but yeah these are supposed to be fairly generic/opaque at this level.  I've separated out snark account handling errors from execution environment validation errors, which are more specific for the EE use case and get exposed under the `Internal` variant.

crates/ee-acct-types/src/state.rs
```diff
@@ -2,17 +2,21 @@
...
 use strata_acct_types::BitcoinAmount;
 use strata_identifiers::Hash;
+use strata_snark_acct_runtime::IInnerState;
+use tree_hash::{Sha256Hasher, TreeHash};
 
 use crate::ssz_generated::ssz::state::{EeAccountState, PendingFinclEntry, PendingInputEntry};
 
 impl EeAccountState {
     pub fn new(
+        chunk_predicate_key: Vec<u8>,
```
storopoli: Forgot to add this to `into_parts`.
And also to add to the DB serialization in https://github.com/alpenlabs/alpen/blob/5607349f2cfb9d25e358e652c8daa80c5ea2cb4c/crates/alpen-ee/database/src/serialization_types/account_state.rs#L35-L55

crates/ee-chunk-runtime/src/chunk_processing.rs
```diff
@@ -0,0 +1,124 @@
...
+    state: &mut E::PartialState,
+    chunk: &Chunk<'_, E>,
+) -> EnvResult<()> {
+    // 1. Make sure the parent block ID we have that we're extending from
+    // matches the chunk transition.
+    let computed_prev_blkid = prev_header.compute_block_id();
+    if computed_prev_blkid != tsn.parent_exec_blkid() {
+        // TODO better error type?
+        return Err(EnvError::MismatchedChainSegment);
+    }
```
storopoli, line 148: We are just checking the parent tip and final tip/state root, however we need per-block parent-child continuity check inside the chunk loop.

crates/snark-acct-runtime/src/ledger.rs
```diff
@@ -0,0 +1,35 @@
+//! IO related types from proof interface.
```
storopoli: Is this module docstring up to date for `ledger.rs`?

crates/snark-acct-runtime/src/message.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+
+/// Represents a parsed message.
+// TODO does this make sense to just be a struct with the `M` as an Option?
+#[derive(Clone, Debug)]
+pub enum InputMessage<M: IAcctMsg> {
+    Valid(MsgMeta, M),
+    Unknown(MsgMeta),
+}
```
storopoli, line 44: Yes it does make sense to be an optional struct.

crates/ee-chunk-runtime/src/chunk_processing.rs
delbonis: oh shit good catch!

crates/ee-chunk-runtime/src/chunk_processing.rs
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        // Verify the block itself.
+        process_block(ee, state, cb)?;
+
+        // Check the block's IO.
+        io_tracker.check_update(cb.inputs(), cb.outputs())?;
+
+        cur_verified_tip_blkid = cb_parent;
+    }
```
storopoli, line 126: The new intra-chunk continuity check is wrong. Inside the loop it verifies `cb_parent == cur_verified_tip_blkid`, but then updates `cur_verified_tip_blkid = cb_parent` instead of the current block’s own ID. That means after the first block the tracker never advances to the first block, so any valid chunk with more than one block will fail on the second iteration.

crates/ee-acct-runtime/src/verification_state.rs
```diff
@@ -134,195 +205,93 @@ impl UpdateVerificationState {
...
+                .map_err(|_| EnvError::InvalidChunkProof)?;
+
+            // Decode the transition for linkage, input matching, and outputs.
+            let transition = chunk
+                .try_decode_chunk_transition()
+                .map_err(|_| EnvError::MalformedChainSegment)?;
+
+            // Process the decoded transition.
+            self.process_decoded_transition(&transition, &mut pending_inp_tracker)?;
+        }
```
storopoli, line 293: During verification, `EeVerificationState::process_chunks_on_acct()` iterates over every proven chunk transition.

For each transition, `process_decoded_transition()` validates linkage and input ordering, then calls `merge_new_outputs(transition.outputs())`.

https://github.com/alpenlabs/alpen/blob/4fa8cb7b5ff7d9eb64f3ff65e621b79c40da2a49/crates/ee-acct-runtime/src/verification_state.rs#L237-L238

In the builder path, `UpdateBuilder::accept_chunk_transition()` does the same kind of accumulation into `self.inner.outputs_mut()`.

https://github.com/alpenlabs/alpen/blob/4fa8cb7b5ff7d9eb64f3ff65e621b79c40da2a49/crates/ee-acct-runtime/src/verification_state.rs#L159-L192

Both sites call `try_extend_transfers(...)` and `try_extend_messages(...)`. Those APIs already model "too many outputs" as a normal `Err`. Instead of propagating that error as an `EnvError`/`BuilderError`, the PR converts it into `panic!`.

That means the system treats “this chunk/update exceeds protocol limits” as “abort the process” instead of “reject the update”.

I think we should reject the update instead of `panic`ing.

storopoli (changes requested):
More changes requested.

crates/ee-acct-runtime/src/verification_state.rs
delbonis: Yeah that makes sense.  I think this something Claude copied from somewhere else it makes more sense.

storopoli: You should try GPT-5.4 :)

crates/ee-chunk-runtime/src/chunk_processing.rs
delbonis: I should write a test for this!

crates/ee-acct-runtime/src/verification_state.rs
```diff
@@ -1,124 +1,172 @@
...
 
-    // Recorded DA.
-    #[expect(dead_code, reason = "for future use")]
-    l1_da_blob_hashes: Vec<Hash>,
+    /// Chunk transitions to verify.
+    input_chunks: &'a [ArchivedChunkInput],
+
+    /// Partial pre-state corresponding to the previous header.
+    // TODO do something with this
+    raw_partial_pre_state: &'a [u8],
```
purusang: Why pre_state may be needed or be utilized in account-proof? I am just trying to understand your thought process on this.

purusang: I think we validate pre_state linkages through chunk proof so I'm not sure if its needed here.

crates/ee-acct-runtime/src/verification_state.rs
```diff
@@ -132,197 +180,97 @@ impl UpdateVerificationState {
...
+        if self.expected_outputs != self.accumulated_outputs {
+            return Err(EnvError::UnsatisfiedObligations(
+                "expected and accumulated outputs mismatch",
+            ));
+        }
 
-        assert!(tracker.consume_input(&"bar".to_string()).is_ok());
-        assert_eq!(tracker.consumed(), 2);
+        // TODO check more of these obligations!  this is all being refactored
+        // for chunk/multiproof updates so there might need to be more here
```
purusang: I am not sure but do we need to check if accumulated sum is less than account's balance? I know its checked in OL but just as a preventive measure.

crates/ee-acct-runtime/src/block_assembly.rs
```diff
@@ -1,24 +1,40 @@
...
     let mut parsed_messages = Vec::with_capacity(msgs.len());
-    for (i, inp) in msgs.iter().enumerate() {
-        let Some(msg) = MsgData::from_entry(inp).ok() else {
+
+    for entry in msgs.iter() {
+        let meta = MsgMeta::new(entry.source(), entry.incl_epoch(), entry.payload_value());
+
+        // Try to decode the message; skip if it fails.
+        let Ok(decoded) = DecodedEeMessageData::decode_raw(entry.payload_buf()) else {
             continue;
```
purusang: Unknown msgs are skipped here (no `add_tracked_balance`), but in `ee_program.rs` (`process_message`), value is credited for all msgs. Feels like these two paths might result in divergent state even though it may not ultimately impact.

crates/ee-acct-runtime/src/ee_program.rs
```diff
@@ -0,0 +1,180 @@
...
+        &self,
+        state: &Self::State,
+        mut vstate: Self::VState<'a>,
+        extra_data: &Self::ExtraData,
+    ) -> ProgramResult<(), Self::Error> {
+        // Process and verify all chunks sequentially.
+        vstate.process_chunks_on_acct(state, extra_data)?;
+
+        // Make sure the state matches the extra data.
+        if state.last_exec_blkid() != *extra_data.new_tip_blkid() {
```
purusang: This code will always pass since we set `last_exec_blkid` two steps back through `pre_finalize_state(state, extra_data)`.

crates/snark-acct-runtime/src/update_builder.rs
```diff
@@ -0,0 +1,426 @@
...
+    ) -> ProgramResult<(), P::Error> {
+        self.messages.push(msg);
+        self.provide_coinput(coinput)
+    }
+
+    /// Returns the current inner state of the snark account, accounting for the
+    /// messages processed.
+    pub fn cur_state(&self) -> &P::State {
+        &self.current_state
+    }
```
purusang, line 116: This duplicates `current_state()`

crates/ee-chain-types/src/sequence_tracker.rs
```diff
@@ -0,0 +1,235 @@
...
+        self.check_inputs(inputs)?;
+        self.advance_unchecked(inputs.len());
+        Ok(())
+    }
+
+    /// Checks inputs without actually consuming them.
+    ///
+    /// This is provided so that multiple sequence trackers can be checked
+    /// before advancing any of them.
+    pub fn check_inputs(&mut self, inputs: &[T]) -> SeqResult<()> {
```
purusang:
```suggestion
    pub fn check_inputs(&self, inputs: &[T]) -> SeqResult<()> {
```

crates/ee-acct-runtime/src/lib.rs
```diff
@@ -1,24 +1,23 @@
 //! Basic EE account runtime framework.
-//!
-//! This is expected to be used within the SNARK proof program used as a snark
-//! account in order to implement an execution environment host account on the
-//! Strata orchestration layer.  There are a collection of utilities in order to
-//! help manipulate EE accounts like by building chain segments and updates.
 
 #![cfg_attr(test, expect(unused_crate_dependencies, reason = "test weirdness"))]
+#![expect(unused, reason = "lots of stuff being refactored")]
```
purusang, line 4:
```suggestion
// TODO: remove this suppression before mainnet?
#![expect(unused, reason = "lots of stuff being refactored")]
```

crates/ee-chain-types/src/sequence_tracker.rs
```diff
@@ -0,0 +1,235 @@
...
+        // Bounds check early so we can skip it on each iteration.
+        if inputs.len() > self.remaining().len() {
+            return Err(SeqError::Overrun);
+        }
+
+        for (i, input) in inputs.iter().enumerate() {
+            // SAFETY: we already did bounds checking
```
purusang, line 105:
```suggestion
            //INVARIANT: we already did bounds checking
```

crates/ee-acct-runtime/src/verification_state.rs
delbonis: It's eventually for DA, unless we decide to have a separate DA proof.

crates/ee-acct-runtime/src/block_assembly.rs
delbonis: Actually this is important to do, good catch.

crates/snark-acct-runtime/src/update_builder.rs
delbonis: I blame claude.

crates/ee-chain-types/src/sequence_tracker.rs
storopoli: @purusang `SAFETY` is something that we put on `unsafe` code and sometimes in `.unwraps` to communicate that it is a safe operation and we know what we're are doing.

crates/ee-acct-runtime/src/lib.rs
storopoli: Can we move this all of the TODO's highlighted in your review to a ticket. We'll make it low priority and add it to a tech debt sprint please?

crates/ee-acct-runtime/src/ee_program.rs
delbonis: Reworked the checks to make it equivalent but make more semantic sense.

delbonis: Reworked the checks to make it equivalent but make more semantic sense.

crates/ee-acct-runtime/src/verification_state.rs
delbonis: I improved the way we do these checks and added some more balance bookkeeping.

crates/ee-acct-runtime/src/ee_program.rs
delbonis: Also reworked these checks.

delbonis: Also reworked these checks.

crates/ee-chain-types/src/sequence_tracker.rs
delbonis: I'll leave it as-is for now.

crates/ee-acct-runtime/src/lib.rs
delbonis: will follow up on this later

storopoli (dismissed):
ACK 333d8e7c416c12dbe80be8da7af86f62a07bebc4

purusang (dismissed):

crates/alpen-ee/block-assembly/src/block.rs
```diff
@@ -68,8 +68,9 @@ pub async fn build_next_exec_block<E: PayloadBuilderEngine>(
     .await?;
 
     // 3. update account state based on built payload and consumed inputs
-    apply_final_update_changes(&mut account_state, &update_extra_data)
-        .context("build_next_exec_block: failed to apply final update changes")?;
+    account_state.set_last_exec_blkid(*update_extra_data.new_tip_blkid());
+    account_state.remove_pending_inputs(*update_extra_data.processed_inputs() as usize);
+    account_state.remove_pending_fincls(*update_extra_data.processed_fincls() as usize);
```
sapinb, line 73: It would be nice if the EeSnarkAccountProgram<E> could be reusable here, without depending on the `ExecutionEnvironment` generic. Maybe something to refactor later.

crates/ee-acct-types/ssz/state.ssz
```diff
@@ -20,6 +21,9 @@ class PendingFinclEntry(Container):
 
 ### EE account internal state
 class EeAccountState(Container):
+    ### Predicate key used for chunk verification.
+    chunk_predicate_key: List[uint8, MAX_CHUNK_PRED_LEN]
```
sapinb: Does this represent the VK for the chunk proof ?
I don't understand why this this required to be persisted here.

Chunk proofs are just implementation details for the ee update proof, an optimization measure. This proof exists only to be used internally to generate the update proof. I do not see any case where this chunk proof vk is independent of the ee proof.
AFAIK, this vk is included at compile time in the update proof.
How will this key be initialized first ?
How will this key be updated ?

crates/ee-acct-runtime/src/verification_state.rs
```diff
@@ -1,328 +1,318 @@
...
+        &mut self,
+        state: &EeAccountState,
+        extra_data: &UpdateExtraData,
+    ) -> EnvResult<()> {
+        let mut pending_inp_tracker = SequenceTracker::new(state.pending_inputs());
+
+        // Loop through all the chunks and verify them.
+        for chunk in self.input_chunks {
+            // Verify the proof against the chunk predicate key.
+            let predicate_key = PredicateKeyBuf::try_from(state.chunk_predicate_key())
```
sapinb: chunk predicate key should just be passed in. It should be known statically by the guest code during compile time, as our update and chunk prover guest codes should be considered coupled.

delbonis: I can see the argument for this, will adapt things.

crates/alpen-ee/block-assembly/src/block.rs
delbonis: I tried to make it more generic but I didn't want to interfere with the existing code too much, this falls into the goal about "generic client infra" in the repo decoupling doc.

crates/ee-acct-types/ssz/state.ssz
delbonis: Yes it's the vk for the chunk proof.  It's persisted so that we could change it without having to change the snark account key itself.  But now that I think about it, since we do do DA stuff in this proof, it probably doesn't make sense for it to be changeable dynamically.  Will update.

Initialized first by how we construct the initial state of the snark account.  Updated hasn't been decided yet.

crates/alpen-ee/block-assembly/src/block.rs
delbonis: created ticket STR-2575

crates/alpen-ee/block-assembly/src/package.rs
```diff
@@ -1,47 +1,56 @@
...
+use strata_snark_acct_runtime::InputMessage;
 use tracing::warn;
 
-/// Builds [`BlockInputs`] from parsed input messages.
+/// Builds [`ExecInputs`] from parsed input messages.
 ///
 /// Only `Deposit` messages are processed; other message types are logged and ignored.
-pub(crate) fn build_block_inputs(parsed_inputs: Vec<MsgData>) -> BlockInputs {
-    let mut inputs = BlockInputs::new_empty();
+// TODO convert this to do it based on extracting pending inputs from the
```
evgenyzdanovich: could you please attach the ticket `TODO(STR-ticketid): todo text`

delbonis: added ticket ref

PR #1315, opened 2026-01-27
STR 2107 New OL Checkpoint db types and database

krsnapaudel:
## Description

Database and manager for new checkpoint types. Includes tests for `checkpoint-types-ssz`.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-2107](https://alpenlabs.atlassian.net/browse/STR-2107)

[STR-2107]: https://alpenlabs.atlassian.net/browse/STR-2107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/storage/src/managers/ol_checkpoint.rs
+++ b/crates/storage/src/managers/ol_checkpoint.rs
@@ -1,0 +1,630 @@
+//! High-level OL checkpoint interface.
+
+use std::{num::NonZeroUsize, sync::Arc};
+
+use strata_checkpoint_types::EpochSummary;
+use strata_db_types::{traits::OLCheckpointDatabase, types::OLCheckpointEntry, DbResult};
+use strata_identifiers::Epoch;
+use strata_primitives::epoch::EpochCommitment;
+use threadpool::ThreadPool;
+
+use crate::{
+    cache::CacheTable,
+    ops::ol_checkpoint::{Context, OLCheckpointOps},
+};
+
+/// Default cache capacity for OL checkpoint entries.
+const DEFAULT_CACHE_CAPACITY: NonZeroUsize = NonZeroUsize::new(64).expect("64 is non-zero");
+
+#[expect(
+    missing_debug_implementations,
+    reason = "Some inner types don't have Debug implementation"
+)]
+pub struct OLCheckpointManager {
+    ops: OLCheckpointOps,
+    checkpoint_cache: CacheTable<Epoch, Option<OLCheckpointEntry>>,
+}
+
+impl OLCheckpointManager {
+    pub fn new<D: OLCheckpointDatabase + Sync + Send + 'static>(
+        pool: ThreadPool,
+        db: Arc<D>,
+    ) -> Self {
+        let ops = Context::new(db.clone()).into_ops(pool);
+        let checkpoint_cache = CacheTable::new(DEFAULT_CACHE_CAPACITY);
+        Self {
+            ops,
+            checkpoint_cache,
+        }
+    }
+
+    /// Stores an OL checkpoint entry for a given epoch.
+    pub async fn put_checkpoint_async(
+        &self,
+        epoch: Epoch,
+        entry: OLCheckpointEntry,
+    ) -> DbResult<()> {
+        self.ops.put_checkpoint_async(epoch, entry.clone()).await?;
+        self.checkpoint_cache.insert_async(epoch, Some(entry)).await;
+        Ok(())
+    }
+
+    /// Inserts an epoch summary retrievable by its epoch commitment.
+    pub async fn insert_epoch_summary_async(&self, summary: EpochSummary) -> DbResult<()> {
+        self.ops.insert_epoch_summary_async(summary).await
+    }
+
+    /// Inserts an epoch summary retrievable by its epoch commitment.
+    pub fn insert_epoch_summary_blocking(&self, summary: EpochSummary) -> DbResult<()> {
+        self.ops.insert_epoch_summary_blocking(summary)
+    }
+
+    /// Gets an epoch summary given an epoch commitment.
+    pub async fn get_epoch_summary_async(
+        &self,
+        epoch: EpochCommitment,
+    ) -> DbResult<Option<EpochSummary>> {
+        self.ops.get_epoch_summary_async(epoch).await
+    }
+
+    /// Gets an epoch summary given an epoch commitment.
+    pub fn get_epoch_summary_blocking(
+        &self,
+        epoch: EpochCommitment,
+    ) -> DbResult<Option<EpochSummary>> {
+        self.ops.get_epoch_summary_blocking(epoch)
+    }
+
+    /// Gets all commitments for an epoch.
+    pub async fn get_epoch_commitments_at_async(
+        &self,
+        epoch: u64,
+    ) -> DbResult<Vec<EpochCommitment>> {
+        self.ops.get_epoch_commitments_at_async(epoch).await
+    }
+
+    /// Gets all commitments for an epoch.
+    pub fn get_epoch_commitments_at_blocking(&self, epoch: u64) -> DbResult<Vec<EpochCommitment>> {
+        self.ops.get_epoch_commitments_at_blocking(epoch)
+    }
+
+    /// Gets the index of the last epoch that we have a summary for, if any.
+    pub async fn get_last_summarized_epoch_async(&self) -> DbResult<Option<u64>> {
+        self.ops.get_last_summarized_epoch_async().await
+    }
+
+    /// Gets the index of the last epoch that we have a summary for, if any.
+    pub fn get_last_summarized_epoch_blocking(&self) -> DbResult<Option<u64>> {
+        self.ops.get_last_summarized_epoch_blocking()
+    }
+
+    /// Deletes an epoch summary given an epoch commitment.
+    pub async fn del_epoch_summary_async(&self, epoch: EpochCommitment) -> DbResult<bool> {
+        self.ops.del_epoch_summary_async(epoch).await
+    }
+
+    /// Deletes an epoch summary given an epoch commitment.
+    pub fn del_epoch_summary_blocking(&self, epoch: EpochCommitment) -> DbResult<bool> {
+        self.ops.del_epoch_summary_blocking(epoch)
+    }
+
+    /// Deletes all epoch summaries from the specified epoch onwards (inclusive).
+    pub async fn del_epoch_summaries_from_epoch_async(
+        &self,
+        start_epoch: u64,
+    ) -> DbResult<Vec<u64>> {
+        self.ops
+            .del_epoch_summaries_from_epoch_async(start_epoch)
+            .await
+    }
+
+    /// Deletes all epoch summaries from the specified epoch onwards (inclusive).
+    pub fn del_epoch_summaries_from_epoch_blocking(&self, start_epoch: u64) -> DbResult<Vec<u64>> {
+        self.ops
+            .del_epoch_summaries_from_epoch_blocking(start_epoch)
+    }
+
+    /// Stores an OL checkpoint entry for a given epoch.
+    pub fn put_checkpoint_blocking(&self, epoch: Epoch, entry: OLCheckpointEntry) -> DbResult<()> {
+        self.ops.put_checkpoint_blocking(epoch, entry.clone())?;
+        self.checkpoint_cache.insert_blocking(epoch, Some(entry));
+        Ok(())
+    }
+
+    /// Retrieves an OL checkpoint entry for a given epoch.
+    pub async fn get_checkpoint_async(&self, epoch: Epoch) -> DbResult<Option<OLCheckpointEntry>> {
+        self.checkpoint_cache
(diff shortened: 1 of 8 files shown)


crates/checkpoint-types-ssz/Cargo.toml
```diff
@@ -21,5 +21,16 @@ borsh.workspace = true
 serde.workspace = true
 thiserror.workspace = true
 
+[dependencies.proptest]
+optional = true
+workspace = true
```
storopoli:
```suggestion
proptest = {optional = true, workspace = true}
```

crates/db/store-sled/src/lib.rs
```diff
@@ -153,6 +159,10 @@ impl DatabaseBackend for SledBackend {
         self.checkpoint_db.clone()
     }
 
+    fn ol_checkpoint_db(&self) -> Arc<impl OLCheckpointDatabase> {
+        self.ol_checkpoint_db.clone()
+    }
+
     fn writer_db(&self) -> Arc<impl traits::L1WriterDatabase> {
         self.writer_db.clone()
     }
```
storopoli: I know that's not relevant to this pr, but can you import it?

```suggestion
fn writer_db(&self) -> Arc<impl L1WriterDatabase> {
self.writer_db.clone()
}
```

crates/storage/src/managers/ol_checkpoint.rs
```diff
@@ -0,0 +1,364 @@
...
+    pub fn del_checkpoints_from_epoch_blocking(&self, start_epoch: Epoch) -> DbResult<Vec<Epoch>> {
+        let deleted = self.ops.del_checkpoints_from_epoch_blocking(start_epoch)?;
+        self.checkpoint_cache
+            .purge_if_blocking(|epoch| *epoch >= start_epoch);
+        Ok(deleted)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
storopoli, line 206: All of the tests here are for the `blocking` functions. Should we make an effort to add `async` `proptest!`s?

crates/checkpoint-types-ssz/src/test_utils.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/// Strategy for generating random [`CheckpointSidecar`] values.
+///
+/// Note: OL logs are currently empty as OLLog strategy would need to be imported
+/// from ol-chain-types-new test-utils.
+pub fn checkpoint_sidecar_strategy() -> impl Strategy<Value = CheckpointSidecar> {
+    state_diff_strategy().prop_map(|state_diff| {
+        // Use empty logs for now; full OLLog strategy can be added if needed
+        CheckpointSidecar::new(state_diff, vec![]).expect("valid sidecar")
```
storopoli: We should maybe make an effort to generate non-empty `OLLog`s

crates/checkpoint-types-ssz/src/test_utils.rs
storopoli, line 1: Maybe we should refactor the following strategies to a more shared/common `test_utils` somewhere because they are not exclusively for `checkpoint-types`

- `buf32_strategy` (`strata-identifiers` feature-gated to `test-utils`)
- `buf64_strategy` (`strata-identifiers` feature-gated to `test-utils`)
- `ol_block_id_strategy` (`strata-identifiers` feature-gated to `test-utils`)
- `epoch_strategy` (`strata-identifiers` feature-gated to `test-utils`)
- `ol_block_commitment_strategy` (`strata-identifiers` feature-gated to `test-utils`)
- `fixed_bytes_32_strategy` (don't know where to put it)

crates/db/types/src/traits.rs
```diff
@@ -236,6 +242,45 @@ pub trait CheckpointDatabase: Send + Sync + 'static {
     fn get_next_unproven_checkpoint_idx(&self) -> DbResult<Option<u64>>;
 }
 
+/// Database for OL checkpoint data.
+pub trait OLCheckpointDatabase: Send + Sync + 'static {
```
storopoli, line 247: Should OL checkpoint epoch summaries support deletion/pruning similar to `CheckpointDatabase`?

https://github.com/alpenlabs/alpen/blob/3b4a7b4fcab1fcea02cfe2ba8bb238cb16a4f40f/crates/db/types/src/traits.rs#L205-L214

crates/checkpoint-types-ssz/src/test_utils.rs
bewakes: Yes, these could probably go to primitives or identifiers. Ahh, one thing I hate about rust is that these need to be feature gated.

krsnapaudel:
> Looks good, bunch of minor comments and the missing `EpochSummary`s deletion feature in `OLCheckpointDatabase` compared to `CheckpointDatabase`.
> 
> Also what's the difference from `CheckpointDatabase` and `OLCheckpointDatabase`? See
> 
> https://github.com/alpenlabs/alpen/blob/3b4a7b4fcab1fcea02cfe2ba8bb238cb16a4f40f/crates/db/types/src/traits.rs#L189-L190
> 
> and
> 
> https://github.com/alpenlabs/alpen/blob/3b4a7b4fcab1fcea02cfe2ba8bb238cb16a4f40f/crates/db/types/src/traits.rs#L245-L246
> 
> Shouldn't we be using `CheckpointDatabase` instead of trivially cloning it and naming `OLCheckpointDatabase`?

@storopoli I think we need a new db because checkpoint types have changed.

crates/db/types/src/traits.rs
krsnapaudel: Yes `strata-dbtool` will need them.

crates/checkpoint-types-ssz/src/test_utils.rs
krsnapaudel: `test-utils-identifiers`? Also now there are some refactor commits. If you prefer, I can create a separate PR for them.

crates/storage/src/managers/ol_checkpoint.rs
krsnapaudel: Apparently `async` and `proptest` don't get along well. Using `Arbitrary`.

crates/checkpoint-types-ssz/src/test_utils.rs
bewakes: If the tests/test-utils are not directly tied to this PR in particular, it would be great to have a separate PR because this PR is already getting bigger.

krsnapaudel: Some refactor commits are not directly tied, but the diffs are small.

storopoli: > Yes, these could probably go to primitives or identifiers. Ahh, one thing I hate about rust is that these need to be feature gated.

Yes, but that means that you are giving the implementer the option to compile or not a bunch of macros and extra dependencies, right?

Also, `strata-primitives` just re-exports `strata-identifiers`

https://github.com/alpenlabs/alpen/blob/53dc34a4f025306c3623517f34dcdf0cf1ee63eb/crates/primitives/src/lib.rs#L6-L11

thus, this should go into `strata-identifiers`.

storopoli: > `test-utils-identifiers`? Also now there are some refactor commits. If you prefer, I can create a separate PR for them.

No, add them to the `strata-identifiers` crate feature-gated to `test-utils`, then when people use the `strata-identifiers` without that feature they won't bring in `proptest` crate or macros as dependencies and more stuff to compile.

crates/storage/src/managers/ol_checkpoint.rs
storopoli: Ok can you move this to a ticket? There must be a way to handle them. Also include all other database-related `proptest`s that only test `blocking` functions and does not test `async` functions.

crates/checkpoint-types-ssz/src/test_utils.rs
krsnapaudel: @storopoli updated in latest commit.

storopoli (dismissed):
ACK 58fd5f95c90122e0d3655595d675c550873179c9

bewakes (dismissed):

crates/identifiers/src/test_utils.rs
prajwolrg, line 1: This is nice. Thanks!

crates/db/types/src/traits.rs
```diff
@@ -236,6 +242,56 @@ pub trait CheckpointDatabase: Send + Sync + 'static {
...
+    /// Inserts an epoch summary retrievable by its epoch commitment.
+    ///
+    /// Fails if there's already an entry there.
+    fn insert_epoch_summary(&self, epoch: EpochSummary) -> DbResult<()>;
+
+    /// Gets an epoch summary given an epoch commitment.
+    fn get_epoch_summary(&self, epoch: EpochCommitment) -> DbResult<Option<EpochSummary>>;
+
+    /// Gets all commitments for an epoch. This makes no guarantees about ordering.
+    fn get_epoch_commitments_at(&self, epoch: u64) -> DbResult<Vec<EpochCommitment>>;
```
prajwolrg, line 257: I wonder if we need this right now. This might come in handy when there are reorgs. But having this here means we need to have some logic to choose the canonical commitment which the trait doesn't provide. I think we will use `first()` or `last()` in the implementation.

crates/db/types/src/traits.rs
```diff
@@ -236,6 +242,56 @@ pub trait CheckpointDatabase: Send + Sync + 'static {
...
+    /// Delete a specific epoch summary by epoch commitment.
+    ///
+    /// Returns true if the epoch summary existed and was deleted, false otherwise.
+    fn del_epoch_summary(&self, epoch: EpochCommitment) -> DbResult<bool>;
+
+    /// Delete epoch summaries from the specified epoch onwards (inclusive).
+    ///
+    /// This method deletes all epoch summaries with epoch index >= start_epoch.
+    /// Returns a vector of deleted epoch indices.
+    fn del_epoch_summaries_from_epoch(&self, start_epoch: u64) -> DbResult<Vec<u64>>;
```
prajwolrg, line 271: I think it makes sense to have it the other way around. To support pruning, we might need deleting the old epoch summaries instead of the new one.

crates/db/types/src/traits.rs
```diff
@@ -236,6 +242,56 @@ pub trait CheckpointDatabase: Send + Sync + 'static {
...
+
+    /// Store an [`OLCheckpointEntry`] by epoch.
+    fn put_checkpoint(&self, epoch: Epoch, entry: OLCheckpointEntry) -> DbResult<()>;
```
prajwolrg, line 274: Should we have a method to update the checkpoint status?

crates/db/types/src/types.rs
```diff
@@ -344,6 +345,38 @@ impl MempoolTxData {
...
+}
+
+/// Signing status of an OL checkpoint.
+#[derive(Debug, Clone, PartialEq, BorshSerialize, BorshDeserialize, Serialize)]
+pub enum OLCheckpointStatus {
+    /// Not signed yet.
+    Unsigned,
+    /// Signed and stored as L1PayloadIntent with given index.
+    Signed(L1PayloadIntentIndex),
+}
```
prajwolrg, line 378: I wonder if we need more status here. Like the checkpoint transaction has been created/submitted/mined.

bewakes: The actual payload to be submitted to L1 is tracked by other database entries like `BundledPayloadEntry`(which already exists). So these two status should suffice here.

crates/db/types/src/traits.rs
bewakes: Since the status is tied with the OLCheckpointEntry unlike OLBlock status which is tracked separate, status update should be handled by this method itself.

crates/db/types/src/traits.rs
krsnapaudel: This is to support delete epoch summary when using `strata-dbtool`. Reverting checkpointed blocks also requires deleting checkpoints and epoch summaries.

crates/storage/src/managers/ol_checkpoint.rs
krsnapaudel: What do you mean? Remove `async` api tests from here? Add and then remove? One way to handle was `tokio` runtime with `block_on`. Not sure if that is a good approach.

@storopoli I had a chat with @bewakes. I will remove `Arbitrary` impls and using the `tokio::Runtime` `block_on`.

crates/db/types/src/traits.rs
krsnapaudel: Actually not sure what we should do here. Following the design for `checkpoint_db`. Yes callers need to decide how to pick the canonical one.

bewakes (approved):

PR #1321, opened 2026-01-29
refactor(identifers): make AccountSerial Varint-backed

prajwolrg:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/identifiers/src/acct.rs
+++ b/crates/identifiers/src/acct.rs
@@ -1,10 +1,17 @@
-use std::{fmt, mem};
+use std::{
+    fmt,
+    hash::{Hash, Hasher},
+    io::{self, Read, Write},
+    mem,
+};
 
 use arbitrary::Arbitrary;
 use borsh::{BorshDeserialize, BorshSerialize};
 use int_enum::IntEnum;
 use serde::{Deserialize, Serialize};
+use ssz::view;
 use ssz_derive::{Decode, Encode};
+use strata_codec::{Codec, CodecError, VARINT_MAX, Varint};
 use thiserror::Error;
 
 const ACCT_ID_LEN: usize = 32;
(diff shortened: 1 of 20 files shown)


crates/ol/stf/src/constants.rs
```diff
@@ -6,12 +8,14 @@ const BRIDGE_GATEWAY_REF: u8 = 0x10;
 pub const BRIDGE_GATEWAY_ACCT_ID: AccountId = AccountId::special(BRIDGE_GATEWAY_REF);
 
 /// Serial of the bridge gateway account.
-pub const BRIDGE_GATEWAY_ACCT_SERIAL: AccountSerial = AccountSerial::reserved(BRIDGE_GATEWAY_REF);
+pub static BRIDGE_GATEWAY_ACCT_SERIAL: LazyLock<AccountSerial> =
+    LazyLock::new(|| AccountSerial::reserved(BRIDGE_GATEWAY_REF));
```
delbonis, line 12: maybe we could make these constructors const?

crates/identifiers/src/acct.rs
```diff
@@ -85,75 +92,213 @@ impl fmt::Display for AccountId {
 
 impl_ssz_transparent_byte_array_wrapper!(AccountId, 32);
 
-type RawAccountSerial = u32;
+type RawAccountSerial = Varint;
 
-/// Size of RawAccountSerial (u32) in bytes
-const RAW_ACCOUNT_SERIAL_LEN: usize = mem::size_of::<RawAccountSerial>();
+/// Size of AccountSerial's SSZ/Borsh representation (u32) in bytes.
+const RAW_ACCOUNT_SERIAL_LEN: usize = mem::size_of::<u32>();
```
delbonis, line 99: I'm a little confused about how this shakes out in practice.  This kinda breaks a little bit of the consistency things because the Codec and Borsh encodings are different than the SSZ representation.

Would it make more sense just to leave this as a u32 and do the varint conversion when we do the Codec/Borsh encoding?

crates/identifiers/src/acct.rs
```diff
@@ -85,75 +92,213 @@ impl fmt::Display for AccountId {
...
-    pub const fn reserved(b: u8) -> Self {
+    pub fn reserved(b: u8) -> Self {
+        let value = b as u32;
         assert!(
-            (b as RawAccountSerial) < SYSTEM_RESERVED_ACCTS,
+            value < SYSTEM_RESERVED_ACCTS,
             "acct: out of bounds reserved serial"
         );
-        Self(b as RawAccountSerial)
+        AccountSerial::new(value).expect("acct: reserved serial within varint bounds")
```
delbonis, line 147: this panic is pointless since we already have the `assert`

crates/identifiers/src/acct.rs
storopoli: Also in `strata-codec` we are defining `Varint` to be represented internally as a `u32`: https://github.com/alpenlabs/strata-common/blob/197c4fb95d52f81ccece68b3f1185fa583f648d2/crates/codec/src/varint.rs#L18-L19

crates/ol/stf/src/constants.rs
prajwolrg: We couldn't make these constructors const since there are no const constructor functions defined [here](https://github.com/alpenlabs/strata-common/blob/main/crates/codec/src/varint.rs).

crates/identifiers/src/acct.rs
prajwolrg: > Would it make more sense just to leave this as a u32 and do the varint conversion when we do the Codec/Borsh encoding?

Doing this means that some `AccountSerial` with value larger than `Varint::MAX` and smaller than `u32::MAX` can't be serializable.

prajwolrg:
> I'm a little bit unsure of how much value this is providing, at least over just changing the encodings. It seems like it complicates the ergonomics for a lot of things.

It does complicate the ergonomics on creation of the type. The value would be in the L1 payload size. Since we need to post `OLLog` that includes `AccountSerial` in the checkpoint payload, using Varint comes in handy there.

crates/identifiers/src/acct.rs
storopoli: Yes probably. @evgenyzdanovich pointed that out in the DA framework in [STR-2146](https://alpenlabs.atlassian.net/browse/STR-2146) which prompted me to create a property testing task for the da primitives in [STR-2147](https://alpenlabs.atlassian.net/browse/STR-2147)

storopoli: You can try doing a `proptest!` here in this PR to see if you get round-trip invariant sustained or errors.

crates/ol/stf/src/constants.rs
delbonis: we could add those...?

prajwolrg: Sure, created a PR https://github.com/alpenlabs/strata-common/pull/70.

delbonis:
Yeah I mean maybe we gotta change how varint works.  I'm a little dissatisfied that it caps out at 1G, maybe we gotta add one more bitpattern to support 62 bit (or maybe we only need 46 bit?) integers.

---

Okay maybe like this, with a little bit of a lopsided distribution

```
0bbbbbbb
10bbbbbb_bbbbbbbb
110bbbbb_bbbbbbbb_bbbbbbbb_bbbbbbbb
111bbbbb_bbbbbbbb_bbbbbbbb_bbbbbbbb_bbbbbbbb_bbbbbbbb
```

so that's actually 45 bits for a max of 35T.

Also, do we need to formally define a signed varint?

prajwolrg:
> Yeah I mean maybe we gotta change how varint works. I'm a little dissatisfied that it caps out at 1G (billion), maybe we gotta add one more bitpattern to support 62 bit (or maybe we only need 46 bit?) integers.

I think having 1B (billion) for account serials is sufficient for now. We would need to rethink this altogether once we support permissionless accounts on OL. Till now I think 1B is sufficient.

> Also, do we need to formally define a signed varint?

I remember there being discussion about this in the spec session, but I'm not sure where we are with that. It's definitely something that'd be super handy. CC: @evgenyzdanovich @storopoli .

PR #1322, opened 2026-01-29
fix: use proptest for `ol_state` db tests

krsnapaudel:
## Description

Fix tests for `ol_state` db to use proptest strategies.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

### DIFF
--- a/crates/db/store-sled/Cargo.toml
+++ b/crates/db/store-sled/Cargo.toml
@@ -12,7 +12,7 @@ strata-checkpoint-types.workspace = true
 strata-codec.workspace = true
 strata-csm-types.workspace = true
 strata-db-types.workspace = true
-strata-identifiers.workspace = true
+strata-identifiers = { workspace = true, features = ["test-utils"] }
 strata-ol-chain-types.workspace = true
 strata-ol-chain-types-new.workspace = true
 strata-ol-chainstate-types.workspace = true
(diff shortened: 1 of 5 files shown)


barakshani:
> > @delbonis Did you check the questions for reviewers? Do we need preseal state persisted in db at all? cc: @bewakes, @prajwolrg, @sapinb, @barakshani
> 
> I don't think we need the preseal state for M2. We can re‑execute the epoch from pre‑epoch state and stop before manifest processing (or otherwise derive a preseal diff). In that flow, we don’t need to persist preseal; we can recompute it.
> 
> It is a very cheap operation for now. It might be interesting to persist the preseal state later and we should work on this post-M2.

This discussion should have been taken before the first line of code was written. Now that we have the code, the tactical questions are less relevant. 

> Also trying to figure out if we will need preseal state for anything else: e.g. proving or state diff generation.

I am not sure what goes into the checkpoint, but you can check this [thread](https://alpen-labs.slack.com/archives/C08G7PL1RBK/p1764105673391389?thread_ts=1764046078.825479&cid=C08G7PL1RBK) (seems like we prove the prestate ony). We are not proving state diff generation. Validity of state diffs is inferred from the checkpoint proof.

storopoli (approved):
ACK 9679c592510de81348cccd8f3f3b9ed006d125e4

Nice repurposing

PR #1323, opened 2026-01-29
STR 2110 Create a simple OL checkpoint building service

krsnapaudel:
## Description

OL checkpoint builder service for Mainnet 1 (v1).

`crates/ol/checkpoint/` - Builder service with provider-based extensibility
`bin/strata/src/services.rs` - Wires checkpoint service into strata binary
### V1 Design Choices
Uses simplified providers:

**DA**: Empty DA.
**Logs**: Empty (production will aggregate from execution)
**Proof**: Placeholder (production will use ZK proofs)
Provider traits allow swapping to production implementations without changing core service logic.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [x]  Yes
- [ ]  No

1. [Checkpointing/reconstruction flow](https://www.notion.so/Full-chain-checkpointing-reconstruction-flow-15b901ba000f807db4cbd1452fee90d8?source=copy_link)
2. [DA data flow and plan](https://www.notion.so/Data-Availability-Data-Flow-Plan-2ac901ba000f803eb42bfaf996a0616b?source=copy_link)

## Checklist

## Related Issues

[STR-2110](https://alpenlabs.atlassian.net/browse/STR-2110)

[STR-2110]: https://alpenlabs.atlassian.net/browse/STR-2110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/strata/src/run_context.rs
+++ b/bin/strata/src/run_context.rs
@@ -10,6 +10,7 @@ use strata_consensus_logic::FcmServiceHandle;
 use strata_csm_worker::CsmWorkerStatus;
 use strata_node_context::{CommonContext, NodeContext};
 use strata_ol_block_assembly::BlockasmHandle;
+use strata_ol_checkpoint::OLCheckpointWorkerHandle;
 use strata_ol_mempool::MempoolHandle;
 use strata_params::Params;
 use strata_service::ServiceMonitor;
@@ -81,6 +82,7 @@ pub(crate) struct ServiceHandles {
     csm_monitor: Arc<ServiceMonitor<CsmWorkerStatus>>,
     mempool_handle: Arc<MempoolHandle>,
     chain_worker_handle: Arc<ChainWorkerHandle>,
+    checkpoint_handle: Arc<OLCheckpointWorkerHandle>,
     fcm_handle: Arc<FcmServiceHandle>,
     /// Sequencer-specific handles (None for fullnodes)
     sequencer_handles: Option<SequencerServiceHandles>,
@@ -92,6 +94,7 @@ impl ServiceHandles {
         csm_monitor: Arc<ServiceMonitor<CsmWorkerStatus>>,
         mempool_handle: Arc<MempoolHandle>,
         chain_worker_handle: Arc<ChainWorkerHandle>,
+        checkpoint_handle: Arc<OLCheckpointWorkerHandle>,
         fcm_handle: Arc<FcmServiceHandle>,
         sequencer_handles: Option<SequencerServiceHandles>,
     ) -> Self {
@@ -100,6 +103,7 @@ impl ServiceHandles {
             csm_monitor,
             mempool_handle,
             chain_worker_handle,
+            checkpoint_handle,
             fcm_handle,
             sequencer_handles,
         }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13664,6 +13664,7 @@ dependencies = [
  "strata-node-context",
  "strata-ol-block-assembly",
  "strata-ol-chain-types-new",
+ "strata-ol-checkpoint",
  "strata-ol-mempool",
  "strata-ol-state-types",
  "strata-ol-stf",
--- a/bin/strata/Cargo.toml
+++ b/bin/strata/Cargo.toml
@@ -20,6 +20,7 @@ strata-ledger-types.workspace = true
 strata-node-context.workspace = true
 strata-ol-block-assembly.workspace = true
 strata-ol-chain-types-new.workspace = true
+strata-ol-checkpoint.workspace = true
 strata-ol-mempool.workspace = true
 strata-ol-state-types.workspace = true
 strata-ol-stf.workspace = true
--- a/bin/strata/src/services.rs
+++ b/bin/strata/src/services.rs
@@ -21,6 +21,7 @@ use strata_node_context::NodeContext;
 use strata_ol_block_assembly::{
     BlockasmBuilder, BlockasmHandle, FixedSlotSealing, MempoolProviderImpl,
 };
+use strata_ol_checkpoint::OLCheckpointBuilder;
 use strata_ol_mempool::{MempoolBuilder, MempoolHandle, OLMempoolConfig};
 use strata_rpc_api_new::OLClientRpcServer;
 use strata_status::StatusChannel;
@@ -67,6 +68,15 @@ pub(crate) fn start_strata_services(nodectx: NodeContext) -> Result<RunContext> 
     // Start Chain worker
     let chain_worker_handle = Arc::new(start_chain_worker_service_from_ctx(&nodectx)?);
 
+    // Start OL checkpoint service
+    let epoch_summary_rx = chain_worker_handle.subscribe_epoch_summaries();
+    let checkpoint_handle = Arc::new(
+        OLCheckpointBuilder::new()
+            .with_node_context(&nodectx)
+            .with_epoch_summary_receiver(epoch_summary_rx)
+            .launch(nodectx.executor())?,
+    );
+
     // Sequencer-specific tasks
     let sequencer_handles = if nodectx.config().client.is_sequencer {
         let broadcast_handle = Arc::new(start_broadcaster(&nodectx));
@@ -95,6 +105,7 @@ pub(crate) fn start_strata_services(nodectx: NodeContext) -> Result<RunContext> 
         csm_monitor,
         mempool_handle,
         chain_worker_handle,
+        checkpoint_handle,
         fcm_handle,
         sequencer_handles,
     );


crates/ol/checkpoint/src/worker.rs
```diff
@@ -0,0 +1,63 @@
...
+/// This worker monitors chain sync status updates and triggers checkpoint
+/// building when new epoch summaries are available.
+///
+/// Designed to be spawned via `TaskExecutor::spawn_critical`:
+/// ```ignore
+/// executor.spawn_critical("ol_checkpoint_worker", |shutdown| {
+///     ol_checkpoint_worker(shutdown, status_ch.clone(), handle.clone(), rt.clone())
+/// });
+/// ```
+pub fn ol_checkpoint_worker(
```
bewakes: Please use the `Input` structure from the service framework. With the framework, we shouldn't be needing ad-hoc workers like this.

krsnapaudel:
@bewakes @prajwolrg Updated to use empty checkpoint sidecar.

crates/ol/checkpoint/src/service.rs
```diff
@@ -0,0 +1,50 @@
...
+impl SyncService for OLCheckpointService {
+    fn on_launch(state: &mut Self::State) -> anyhow::Result<()> {
+        state.initialize()?;
+        Ok(())
+    }
+
+    fn process_input(state: &mut Self::State, input: &Self::Msg) -> anyhow::Result<Response> {
+        match input {
+            OLCheckpointMessage::NewEpochSummary(commitment) => {
+                let _ = state.on_epoch_summary(*commitment);
```
storopoli: Errors from on_epoch_summary are swallowed (`let _ = ...`), so DB failures or missing summaries fail silently and the service keeps running in a bad state.

storopoli (changes requested):

crates/ol/checkpoint/src/service.rs
```diff
@@ -0,0 +1,73 @@
...
+                        }
+                        // Should not occur from on_epoch_summary
+                        OLCheckpointError::MissingDependency(_) => {
+                            error!(epoch, %err, "unexpected error type");
+                        }
+                        // Expected errors from this code path
+                        OLCheckpointError::Database(_)
+                        | OLCheckpointError::NotInitialized
+                        | OLCheckpointError::Unexpected(_) => {
+                            error!(epoch, %err, "failed to process epoch summary");
```
storopoli: Minor nit. I prefer to be explicit with what is `Debug` (`?`) versus what is `Display` (`%`) in logging statements.

I don't know what is the default from `tracing`, hence let's not leave that to chance:

```suggestion
                            warn!(%epoch, %err, "epoch data not yet available");
                        }
                        // Should not occur from on_epoch_summary
                        OLCheckpointError::MissingDependency(_) => {
                            error!(%epoch, %err, "unexpected error type");
                        }
                        // Expected errors from this code path
                        OLCheckpointError::Database(_)
                        | OLCheckpointError::NotInitialized
                        | OLCheckpointError::Unexpected(_) => {
                            error!(%epoch, %err, "failed to process epoch summary");
```

crates/ol/checkpoint/src/builder.rs
```diff
@@ -0,0 +1,131 @@
+//! Builder pattern for launching the OL checkpoint service.
+
```
bewakes, line 2: I do not see much value of having a wrapper service builder here, because we already have `ServiceBuilder` that we can use for almost all cases.

How about just having a function that starts the service which takes in the deps/ctx for the service and initializes the service inner state. Something like:
```rust
pub fn start_checkpoint_builder_service(ctx: Ctx/Dep) -> Result<ServiceMonitor<_>> {
    let service_state = CheckpointServiceState::new(ctx);
    let input = ... initialize input
    let monitor = ServiceBuilder::new().with_state(service_state).with_input(input).launch(...);
    Ok(monitor)
}
```
This saves us from having the redundant builder wrapper which is doing just what the above function does, all with redundant boilerplate.

crates/ol/checkpoint/src/state.rs
```diff
@@ -0,0 +1,328 @@
...
+        self.last_processed_epoch
+    }
+
+    pub fn initialize(&mut self) {
+        self.init_cursor_from_db();
+        self.initialized = true;
+    }
+
+    /// Process epoch summary notification, catching up from last checkpoint to latest summary.
+    pub fn on_epoch_summary(&mut self, _target: EpochCommitment) -> OLCheckpointResult<()> {
```
bewakes: `on_epoch_summary` sounds like it takes a callback function. Better rename this to `handle_epoch_summary` or even better `handle_epoch_commitment` to be consistent with the argument?

crates/ol/checkpoint/src/state.rs
```diff
@@ -0,0 +1,328 @@
...
+        if !self.initialized {
+            return Err(OLCheckpointError::NotInitialized);
+        }
+
+        let ol_checkpoint = self.deps.storage.ol_checkpoint();
```
bewakes: nit: The name `ol_checkpoint` is really not descriptive of what it is. :D  It is a db.

crates/ol/checkpoint/src/state.rs
```diff
@@ -0,0 +1,328 @@
...
+    /// Returns error if the epoch index cannot be processed (missing data).
+    /// Checkpoints must be built sequentially, so caller should stop on error.
+    fn process_epoch(&mut self, epoch_index: u64) -> OLCheckpointResult<()> {
+        let ol_checkpoint = self.deps.storage.ol_checkpoint();
+
+        // Get commitments for this epoch index - must exist to proceed
+        let commitments = ol_checkpoint.get_epoch_commitments_at_blocking(epoch_index)?;
+        let commitment = commitments
+            .first()
+            .copied()
```
bewakes: This looks like we are fetching the canonical checkpoint commitment, which we assume to be the first one. It might make more sense to have a method in the manager like `get_canonical_epoch_commitment_at_*` that hides the logic of fetching the canonical commitment. This way, when we change the logic, we can change at one place.

crates/ol/checkpoint/src/state.rs
```diff
@@ -0,0 +1,328 @@
...
+            return;
+        };
+
+        let Ok(Some(last_summarized_index)) = ol_checkpoint.get_last_summarized_epoch_blocking()
+        else {
+            return;
+        };
+
+        for epoch_index in (0..=last_summarized_index).rev() {
+            let Ok(commitments) = ol_checkpoint.get_epoch_commitments_at_blocking(epoch_index)
```
bewakes: yeah, here also we can use the manager's `get_canonical_commitment` that I mentioned earlier.

crates/ol/checkpoint/src/builder.rs
```diff
@@ -0,0 +1,131 @@
...
+    reason = "Some inner types don't have Debug implementation"
+)]
+/// Builder for constructing and launching an OL checkpoint service.
+pub struct OLCheckpointBuilder {
+    storage: Option<Arc<NodeStorage>>,
+    runtime_handle: Option<Handle>,
+    epoch_summary_rx: Option<watch::Receiver<Option<EpochCommitment>>>,
+    da_provider: Option<Arc<dyn DaProvider>>,
+    log_provider: Option<Arc<dyn LogProvider>>,
+    proof_provider: Option<Arc<dyn ProofProvider>>,
```
delbonis: This is sorta mixing up a lot of stuff with different patterns.  Since we're developing this as a new component, these providers and the node storage should *probably* be integrated into a "checkpoint builder service context" trait.  That would also simplify this builder type.

crates/ol/checkpoint/src/input.rs
```diff
@@ -0,0 +1,54 @@
...
+use tracing::trace;
+
+use crate::message::OLCheckpointMessage;
+
+/// Input source for the checkpoint service.
+#[derive(Debug)]
+pub struct OLCheckpointInput {
+    summary_rx: watch::Receiver<Option<EpochCommitment>>,
+    closed: bool,
+}
```
delbonis: I thought we had a generic service input that wraps an arbitrary `watch::Receiver`?  We shouldn't have to define a custom version of this.

crates/ol/checkpoint/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+//! OL checkpoint builder service.
+
+pub mod builder;
+pub mod errors;
+pub mod handle;
+pub mod input;
+pub mod message;
+pub mod providers;
+pub mod service;
+pub mod state;
```
delbonis: should avoid making these modules `pub`

crates/ol/checkpoint/src/message.rs
```diff
@@ -0,0 +1,12 @@
...
+
+use strata_primitives::epoch::EpochCommitment;
+
+/// Input messages for the OL checkpoint service.
+#[derive(Debug)]
+pub enum OLCheckpointMessage {
+    /// New epoch summary commitment is available.
+    NewEpochSummary(EpochCommitment),
+    /// Input channel closed.
+    Abort,
```
delbonis: This doesn't need to be explicitly signalled, there's a function called when the service shuts down due to the input closing.

crates/ol/checkpoint/src/providers.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/// Provides proof bytes for checkpoint payload.
+///
+/// V1 implementation returns empty placeholder.
+/// Production implementation will integrate with prover client.
+pub trait ProofProvider: Send + Sync {
+    /// Get proof bytes for the checkpoint.
+    fn get_proof(&self, summary: &EpochSummary) -> OLCheckpointResult<Vec<u8>>;
+}
```
delbonis: Yeah these could all be integrated into a common context trait specifically for the checkpoint service.

crates/ol/checkpoint/src/state.rs
```diff
@@ -0,0 +1,328 @@
...
+        let payload = build_checkpoint_payload(
+            &summary,
+            self.deps.da_provider.as_ref(),
+            self.deps.log_provider.as_ref(),
+            self.deps.proof_provider.as_ref(),
+        )?;
+        let entry = OLCheckpointEntry::new_unsigned(payload);
+        ol_checkpoint.put_checkpoint_blocking(epoch, entry)?;
+
+        info!(epoch, "stored OL checkpoint entry");
```
delbonis: Can probably be made `debug!`.

crates/ol/checkpoint/src/state.rs
```diff
@@ -0,0 +1,328 @@
...
+            return Ok(());
+        };
+
+        // Determine starting epoch index (last processed + 1, or 0 if none)
+        let start_epoch_index = self.last_processed_epoch_index.map(|e| e + 1).unwrap_or(0);
+
+        // Process all epochs from start to target (inclusive)
+        for epoch_index in start_epoch_index..=target_epoch_index {
+            self.process_epoch(epoch_index)?;
+        }
```
delbonis: So this is interesting because we're passing in an `EpochCommitment` but we're not using it to figure out which epochs to process, even as a sanity check.

bin/strata/src/run_context.rs
```diff
@@ -92,6 +94,7 @@ impl ServiceHandles {
         csm_monitor: Arc<ServiceMonitor<CsmWorkerStatus>>,
         mempool_handle: Arc<MempoolHandle>,
         chain_worker_handle: Arc<ChainWorkerHandle>,
+        checkpoint_handle: Arc<OLCheckpointHandle>,
```
delbonis: OLCheckpoint*Worker*Handle?

bewakes (commented):
Looks good overall. Just some comments and suggestions.

crates/ol/checkpoint/src/builder.rs
krsnapaudel: I have left the builder pattern in place. With the context added (see @delbonis' comments above), the gain is minimal.

crates/ol/checkpoint/src/state.rs
krsnapaudel: Added a check to make sure we process up to the target.

crates/ol/checkpoint/src/worker.rs
krsnapaudel: No longer relevant.

crates/ol/checkpoint/src/builder.rs
delbonis: I think it's fine to have a "second-order" builder that creates and launches the lower-level service builder in its `launch` function.

crates/ol/checkpoint/src/builder.rs
```diff
@@ -0,0 +1,92 @@
...
+            .ok_or(OLCheckpointError::MissingDependency("storage"))?;
+        let runtime_handle = self
+            .runtime_handle
+            .ok_or(OLCheckpointError::MissingDependency("runtime_handle"))?;
+        let epoch_summary_rx = self
+            .epoch_summary_rx
+            .ok_or(OLCheckpointError::MissingDependency("epoch_summary_rx"))?;
+
+        let input = TokioWatchInput::from_receiver(epoch_summary_rx);
+        let input = SyncAsyncInput::new(input, runtime_handle);
```
delbonis: Also now that I'm thinking about this a little more, it would be more "technically correct" to get the runtime handle from the `TaskExecutor` in this `launch` fn instead of the node context in `with_...`.  Then we can make the builder just take the storage and input.  This reduces the scope of the interdependencies.

crates/ol/checkpoint/src/context.rs
```diff
@@ -0,0 +1,62 @@
...
+/// with mock implementations and future production providers.
+pub(crate) trait CheckpointWorkerContext: Send + Sync + 'static {
+    /// Get access to checkpoint database manager.
+    fn checkpoint_db_manager(&self) -> &Arc<OLCheckpointManager>;
+
+    /// Computes the data availability payload for the epoch.
+    fn compute_da(&self, summary: &EpochSummary) -> OLCheckpointResult<Vec<u8>>;
+
+    /// Gets aggregated OL logs for the epoch.
+    fn get_epoch_logs(&self, summary: &EpochSummary) -> OLCheckpointResult<Vec<OLLog>>;
```
delbonis: I think these only need to take `EpochCommitment` as a reference to the epoch.  The summary is more of a "description" type, not something we'd use as a key in a query like this.

Also, I'm a bit uncertain about `compute_da` being the model.  I would expect that the checkpoint worker is the component that itself would be responsible doing the DA generation, and this kinda punts on doing one of the "hard parts" to this lower-level context impl.  It's fine to keep it stubbed for now, since that's still being developed by Jose.

(cc: @storopoli thoughts?)

crates/ol/checkpoint/src/context.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use crate::errors::OLCheckpointResult;
+
+/// Context providing dependencies for the checkpoint worker.
+///
+/// This trait abstracts storage and data providers, enabling testing
+/// with mock implementations and future production providers.
+pub(crate) trait CheckpointWorkerContext: Send + Sync + 'static {
+    /// Get access to checkpoint database manager.
+    fn checkpoint_db_manager(&self) -> &Arc<OLCheckpointManager>;
```
delbonis: Instead of exposing the manager we should expose the highest-level basic database operations that we use the manager for in this context trait.  Then the impl is defined around the manager without actually exposing it.

As-is, there's no way to run the checkpoint worker without also running a bunch of the database infrastructure, since there's no way  to define the context trait without that.  This lets us write more isolated tests since it doesn't assume that we have all the database infrastructure existing.

crates/ol/checkpoint/src/context.rs
delbonis, line 1: Also, it might be fine to just use anyhow errors instead of defining a new error/result type since this is a high-level/service-level trait.

crates/ol/checkpoint/src/service.rs
```diff
@@ -0,0 +1,80 @@
...
+
+/// OL checkpoint builder service implementation.
+///
+/// Generic over the context type to support different storage/provider implementations.
+#[derive(Debug)]
+pub(crate) struct OLCheckpointService<C: CheckpointWorkerContext>(PhantomData<C>);
+
+impl<C: CheckpointWorkerContext> Service for OLCheckpointService<C> {
+    type State = OLCheckpointServiceState<C>;
+    type Msg = Option<EpochCommitment>;
```
delbonis, line 28: Should probably capture this with some kind of "event" type instead of mapping the concepts onto `Option`.  What does `None` really signify here?

crates/ol/checkpoint/src/state.rs
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    /// Handle epoch commitment notification, catching up from last checkpoint to latest summary.
+    ///
+    /// The `target` commitment is used as a sanity check - we verify that after processing,
+    /// we've caught up to at least the target epoch.
+    pub(crate) fn handle_epoch_commitment(
+        &mut self,
+        target: EpochCommitment,
+    ) -> OLCheckpointResult<()> {
```
delbonis: Like what does `handle_epoch_commitment` really mean here?  What is this being called in response to?

I can tell this is being called when we get a new input to the service, but I am not 100% clear of what that is intended to correspond to.  Is this called when we newly finish an epoch and want to generate the checkpoint from it?  If so, then a better name for this function would be something like `handle_complete_epoch`, where the argument is merely a identifier for the epoch that was completed that we want to generate a checkpoint for.

crates/ol/checkpoint/src/service.rs
```diff
@@ -0,0 +1,80 @@
...
+                // Transient: data may not be available yet
+                OLCheckpointError::MissingEpochCommitment(_)
+                | OLCheckpointError::MissingEpochSummary(_) => {
+                    warn!(%epoch, %err, "epoch data not yet available");
+                }
+                // Should not occur from on_epoch_summary
+                OLCheckpointError::MissingDependency(_) => {
+                    error!(%epoch, %err, "unexpected error type");
+                    return Err(err.into());
+                }
```
delbonis: This means these should be different error types.

crates/ol/checkpoint/src/context.rs
krsnapaudel: Removed `OLCheckpointError/OLCheckpointResult`.

crates/ol/checkpoint/src/context.rs
krsnapaudel: Trait now exposes 8 individual high-level operations.

crates/ol/checkpoint/src/context.rs
krsnapaudel: All methods take `EpochCommitment`. Moved `compute_da` off the trait to a standalone function

crates/ol/checkpoint/src/service.rs
krsnapaudel: Kept `Option<EpochCommitment>` since the watch channel sender determines the type. Added doc comment explaining `None` = initial state (no epoch completed yet)

crates/ol/checkpoint/src/builder.rs
krsnapaudel: Removed `runtime_handle` from builder fields.

crates/ol/checkpoint/src/service.rs
krsnapaudel: `CheckpointNotReady (thiserror)` for transient "data not ready" cases, `anyhow::Error` for everything else. Service distinguishes via downcast

delbonis: I meant that we shouldn't be using `OLCheckpointError` for both init/builder errors and running service errors.

But will re-review shortly.

PR #1329, opened 2026-02-01
feat(strata): add essential services in client

purusang:
## Description
Add btcio reader task, broadcaster, writer, and block assembly services to new strata binary.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues
STR-2109

### DIFF
--- a/bin/strata/src/run_context.rs
+++ b/bin/strata/src/run_context.rs
@@ -66,7 +66,7 @@ impl RunContext {
 /// envelope signing, and block assembly. Stored as `Option` in [`ServiceHandles`]
 /// since fullnodes don't run these services.
 #[expect(
-    unused, 
+    unused,
     reason = "fields will be accessed when sequencer RPC is implemented"
 )]
 pub(crate) struct SequencerServiceHandles {


bin/strata/src/context.rs
```diff
@@ -22,15 +22,25 @@ pub(crate) fn load_config_early(args: &Args) -> Result<Config, InitError> {
     get_config(args.clone())
 }
 
-pub(crate) fn init_storage(config: &Config) -> Result<Arc<NodeStorage>, InitError> {
+pub(crate) fn init_storage(
```
bewakes: Let's keep the pool and db in the node storage itself and expose it via the node storage instead via node context. This will keep node context cleaner and also keeps the backend and pool where they belong which is inside the node storage

bin/strata/src/run_context.rs
```diff
@@ -65,22 +67,36 @@ pub(crate) struct ServiceHandles {
...
+        clippy::too_many_arguments,
+        reason = "Service handles aggregates all service handles"
+    )]
     pub(crate) fn new(
         asm_handle: Arc<AsmWorkerHandle>,
         csm_monitor: Arc<ServiceMonitor<CsmWorkerStatus>>,
         mempool_handle: Arc<MempoolHandle>,
         chain_worker_handle: Arc<ChainWorkerHandle>,
         fcm_handle: Arc<FcmServiceHandle>,
+        broadcast_handle: Option<Arc<L1BroadcastHandle>>,
```
bewakes: It might be a better idea to group these into something called `SequencerServiceHandles` since all of these are either all None or all some. This way, we can also just pass a single argument that's None or Some.

bin/strata/src/services.rs
```diff
@@ -69,11 +95,112 @@ pub(crate) fn start_strata_services(nodectx: NodeContext) -> Result<RunContext>
...
+fn start_writer(
+    nodectx: &NodeContext,
+    broadcast_handle: Arc<L1BroadcastHandle>,
+) -> Result<Arc<EnvelopeHandle>> {
+    let sequencer_address = nodectx
+        .task_manager()
+        .handle()
+        .block_on(generate_sequencer_address(nodectx.bitcoin_client()))?;
+
+    let writer_db = nodectx.db().writer_db();
```
bewakes: Relates to above, this is better if we access it from node storage and not from nodectx directly.

bin/strata/src/services.rs
```diff
@@ -51,8 +62,23 @@ pub(crate) fn start_strata_services(nodectx: NodeContext) -> Result<RunContext>
     // Start Chain worker
     let chain_worker_handle = Arc::new(start_chain_worker_service_from_ctx(&nodectx)?);
 
-    // TODO: Start other tasks like l1writer, broadcaster, btcio reader, etc. all as
-    // service, returning the monitors.
+    // Start btcio reader task (polls Bitcoin for new blocks and submits to ASM)
+    start_btcio_reader(&nodectx, asm_handle.clone());
```
bewakes: This needs to be started before genesis init/check because genesis requires ASM to be ready with the genesis manifest which will be only after the btcio reader provides the L1 block to ASM. So right now, the sequencer would be stuck waiting for genesis.

When you do so, could you please also add comment on this ordering of services?

bewakes: This is also why the functional-tests-new must be failing right now.

bin/strata/src/run_context.rs
```diff
@@ -58,13 +60,25 @@ impl RunContext {
     }
 }
 
+/// Sequencer-specific service handles.
+///
+/// These handles are only present when running as a sequencer (not for fullnodes).
+#[expect(unused, reason = "will be used later")]
```
prajwolrg: Now that this is used in `ServiceHandles`, can we update the docstring?

purusang: handles in the struct are still not accessed so we will need it to pass clippy errors.

bewakes:
>  Need to fix new functional tests before this can go into the codbase.

@barakshani , func tests new are passing now.

PR #1331, opened 2026-02-02
feat(ol): add OL genesis parameters crate

prajwolrg:
## Description

Introduce the `strata-ol-params` crate (crates/ol/params/) providing JSON-serializable genesis configuration for OL, including header params for the parent of the genesis block, genesis account definitions, and the initial L1 block commitment.

Add `OLState::from_genesis_params()` and `TsnlLedgerAccountsTable::from_genesis_params()` to construct initial OL state directly from the new params, wiring up epoch state, global state, and ledger accounts

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/ol/state-types/src/toplevel.rs
+++ b/crates/ol/state-types/src/toplevel.rs
@@ -44,7 +44,7 @@ impl OLState {
         let checkpointed_epoch = params.checkpointed_epoch();
         let manifests_mmr = Mmr64::from_generic(&CompactMmr64::new(64));
 
-        let ledger = TsnlLedgerAccountsTable::from_genesis_params(params)?;
+        let ledger = TsnlLedgerAccountsTable::from_genesis_account_params(&params.accounts)?;
         let total_ledger_funds = ledger.calculate_total_funds();
 
         let global = GlobalState::new(params.header.slot);
--- a/crates/ol/state-types/src/ledger.rs
+++ b/crates/ol/state-types/src/ledger.rs
@@ -5,7 +5,7 @@
 use strata_acct_types::{
     AccountId, AccountSerial, AcctError, AcctResult, BitcoinAmount, SYSTEM_RESERVED_ACCTS,
 };
-use strata_ol_params::OLParams;
+use strata_ol_params::AccountParams;
 
 use crate::ssz_generated::ssz::state::{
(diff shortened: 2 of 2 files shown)


crates/ol/params/src/header.rs
```diff
@@ -0,0 +1,111 @@
...
+
+/// Header parameters for the parent of the genesis block.
+///
+/// These describe the block immediately preceding genesis, not the genesis
+/// block itself. All fields have sensible defaults. If not provided,
+/// `timestamp`, `slot`, and `epoch` default to 0, while `parent_blkid`,
+/// `body_root`, and `logs_root` default to their zero/null values.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
+pub struct HeaderParams {
```
bewakes, line 16: Why do we need this? Shouldn't this be hardcoded/implied?

crates/ol/state-types/src/toplevel.rs
```diff
@@ -38,6 +39,29 @@ impl OLState {
         }
     }
 
+    /// Creates initial OL state from genesis parameters.
+    pub fn from_genesis_params(params: &OLParams) -> AcctResult<Self> {
+        let checkpointed_epoch = params.checkpointed_epoch();
+        let manifests_mmr = Mmr64::from_generic(&CompactMmr64::new(64));
+
+        let ledger = TsnlLedgerAccountsTable::from_genesis_params(params)?;
```
bewakes: I suggest passing just the account related params instead of whole params because ledger table uses the "accounts" section of the params and name it `from_genesis_account_params`.

crates/ol/params/src/header.rs
prajwolrg: This can be useful for times when we explicitly want to set something else as the genesis. We already use [`OLState::new_at`](https://github.com/alpenlabs/alpen/blob/24b9d6fcc65a9c34e2f523d330f01937766ff0ba/crates/ol/state-types/src/toplevel.rs#L42). This is meant to be a more generic version of that.

PR #1334, opened 2026-02-02
refactor: move genesis out of strata binary into crates

storopoli:
## Description

Moves `strata-node-context` from `crates/node-context` to `crates/ol/node` as `strata-ol-node` to match the OL/EE split and avoid the generic "node" name. The former `lib.rs` is now `strata_ol_node::context`, and genesis lives in `strata_ol_node::genesis`.

Genesis now writes the epoch 0 summary into the OL checkpoint DB (`NodeStorage::ol_checkpoint()`), so "previous epoch" lookups always have data.

### Type of Change

- [x] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Used Cursor with GPT-5.2 Codex for fixing the annoying changes post renaming/refactors.

- Genesis previously used `NodeStorage::checkpoint()`; this is updated to `NodeStorage::ol_checkpoint()` (post-#1315).
- I removed null-epoch checks in the new OL stack that only existed because epoch 0 summaries were missing. If you spot any remaining in new sequencer or `chain-worker-new`, please flag.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-2156

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,7 +97,6 @@ members = [
   "crates/simple-ee",
   "crates/snark-acct-sys",
   "crates/snark-acct-types",
-  "crates/snark-acct-sys",
   "crates/state",
   "crates/status",
   "crates/storage",


Cargo.toml
```diff
@@ -97,7 +97,6 @@ members = [
   "crates/simple-ee",
   "crates/snark-acct-sys",
   "crates/snark-acct-types",
-  "crates/snark-acct-sys",
```
storopoli, line 100: This was duplicated and not my doing.

bewakes:
@storopoli should we keep the node-context and genesis logic in the same crate? We should probably not because `NodeContext` is supposed to be used by any other "service" crate which shouldn't be able to trigger genesis logic. 
We could pass the service params from the binary, but it's much easier to just pass in the `&NodeContext` and let the service crate handle the rest. What do you think?

PR #1338, opened 2026-02-03
refactor(btcio): add L1 block info to tx status and decouple from strata-params

irnb:
## Description

- Add `block_hash` and `block_height` fields to `L1TxStatus::Confirmed` and `L1TxStatus::Finalized` variants to provide L1 block context for transaction status tracking
- Decouple btcio from `strata-params` by introducing `BtcioParams` struct containing only the parameters btcio needs (`l1_reorg_safe_depth`, `magic_bytes`, `genesis_l1_height`)
- Add `params_to_btcio_params` helper function in strata-client for on-demand conversion

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This PR is part of a series of sequential PRs that build on each other:

These PRs should be merged sequentially. The end goal is implementing chunked envelope support for EE data availability and integrating it with the EE sequencer.

Is this PR addressing any specification, design doc or external reference document?

- [ ]  Yes
- [x]  No

If yes, please add relevant links:

## Checklist

- [x] I have performed a self-review of my code.
- [ ] I have commented my code where necessary.
- [ ] I have updated the documentation if needed.
- [ ] My changes do not introduce new warnings.
- [ ] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [ ] New and existing tests pass with my changes.
- [ ] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

PR #1340, opened 2026-02-03
feat(ee-da): add chunked envelope storage layer and btcio chunked envelope writer

irnb:
## Description

- Add DA chunk encoding types (`DaChunkHeader`, `encode_da_blob_chunks`) in `alpen-ee-common` for splitting large DA blobs into 390KB chunks with headers containing version, blob hash, chunk index, and total chunks
- Add `DaBlobProvider` trait for abstracting DA blob sources
- Implement chunked envelope storage layer with `ChunkedEnvelopeEntry`, `RevealTxMeta`, and `ChunkedEnvelopeStatus` types in db-types
- Add `L1ChunkedEnvelopeDatabase` trait and SledDB implementation for persisting chunked envelopes
- Add `ChunkedEnvelopeOps` in storage crate for async database operations
- Implement btcio chunked envelope writer module with builder, signer, handle, and context for constructing commit + N reveal transactions with wtxid-chain linking
- Add `EeDatabases` container in `alpen-ee-database` for unified database initialization (ee_node_db, broadcast_db, chunked_envelope_db)

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

> ⚠️ **This PR must be merged after https://github.com/alpenlabs/alpen/pull/1338**

PR #1338 introduces the `BtcioParams` struct that decouples `btcio` from `strata-params`. This PR builds on that decoupling.

- [x]  Yes
- [ ]  No

[EE-DA-Chunked-Envelope-Payload](https://www.notion.so/EE-DA-Chunked-Envelope-Payload-2c3901ba000f80d9ac7ccde65a2738cf
)

## Checklist

### DIFF
--- a/crates/db/store-sled/src/lib.rs
+++ b/crates/db/store-sled/src/lib.rs
@@ -48,8 +48,8 @@ use strata_db_types::{
     traits::{
         AccountGenesisDatabase, AsmDatabase, CheckpointDatabase, ClientStateDatabase,
         DatabaseBackend, L1BroadcastDatabase, L1ChunkedEnvelopeDatabase, L1Database,
-        L1WriterDatabase, L2BlockDatabase, MempoolDatabase, OLBlockDatabase,
-        OLCheckpointDatabase, OLStateDatabase, ProofDatabase,
+        L1WriterDatabase, L2BlockDatabase, MempoolDatabase, OLBlockDatabase, OLCheckpointDatabase,
+        OLStateDatabase, ProofDatabase,
     },
 };
 use typed_sled::SledDb;


crates/alpen-ee/common/src/types/da.rs
```diff
@@ -0,0 +1,266 @@
...
+/// 33      2     chunk_index
+/// 35      2     total_chunks
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Codec)]
+struct DaChunkHeader {
+    version: u8,
+    blob_hash: BlobHash,
+    chunk_index: u16,
+    total_chunks: u16,
+}
```
storopoli, line 111: According to the [design doc](https://www.notion.so/EE-DA-Chunked-Envelope-Payload-2c3901ba000f80d9ac7ccde65a2738cf?source=copy_link#2c3901ba000f802790eaff8c2592ac9f), this is omitting `prev_chunk_wtxid` and it is 37 bytes instead of 69 bytes.

Are you sure we don't need `prev_chunk_wtxid`?

crates/alpen-ee/common/src/types/da.rs
```diff
@@ -0,0 +1,266 @@
...
+//   - Chunk header: 37 wu
+//
+// Total overhead: ~665 wu. Remaining: ~399,335 wu for payload.
+// Using 390,000 to keep a safe margin.
+
+/// Maximum chunk payload size in bytes.
+const MAX_CHUNK_PAYLOAD: usize = 390_000;
+
+/// Serialized size of [`DaChunkHeader`] in bytes.
+const DA_CHUNK_HEADER_SIZE: usize = 37;
```
storopoli, line 79: Not including `prev_chunk_wtxid` which would push this to 69 bytes.
See the [design doc](https://www.notion.so/EE-DA-Chunked-Envelope-Payload-2c3901ba000f80d9ac7ccde65a2738cf?source=copy_link#2c3901ba000f802790eaff8c2592ac9f)

crates/alpen-ee/common/src/types/da.rs
```diff
@@ -0,0 +1,266 @@
...
+
+    for (i, (header, _)) in decoded.iter().enumerate() {
+        if header.chunk_index != i as u16 {
+            return None;
+        }
+    }
+
+    let payloads: Vec<Vec<u8>> = decoded.iter().map(|(_, p)| p.to_vec()).collect();
+    reassemble_from_payloads(&payloads, expected_hash)
+}
```
storopoli: In the linked spec we should:

1. check the header version to be `==0` (not doing this here)

    Image

2. check the expected hash of ALL chunks (you're just checking the first one):

    Image

crates/alpen-ee/common/src/types/da.rs
```diff
@@ -0,0 +1,266 @@
...
+//   - Schnorr signature: 65 wu
+//   - Control block: 33 wu
+//   - Script overhead (opcodes, pushdata headers): ~50 wu
+//   - Chunk header: 37 wu
+//
+// Total overhead: ~665 wu. Remaining: ~399,335 wu for payload.
+// Using 390,000 to keep a safe margin.
+
+/// Maximum chunk payload size in bytes.
+const MAX_CHUNK_PAYLOAD: usize = 390_000;
```
storopoli: In the spec you say ~`330_000` bytes. Either update there or fix here:

Image

crates/btcio/src/writer/chunked_envelope/builder.rs
```diff
@@ -0,0 +1,431 @@
...
+}
+
+/// `OP_RETURN <magic_bytes(4)> <prev_wtxid(32)>`.
+fn build_linking_tag(magic_bytes: &[u8; 4], prev_wtxid: &Buf32) -> ScriptBuf {
+    script::Builder::new()
+        .push_opcode(OP_RETURN)
+        .push_slice(magic_bytes)
+        .push_slice(prev_wtxid.as_ref())
+        .into_script()
+}
```
storopoli, line 176: My understanding from the spec is that we're not using `OP_RETURN`s (SPS-50 style) here and we'd put the linking tag in the witness. Please clarify

Image

crates/btcio/src/writer/chunked_envelope/builder.rs
```diff
@@ -0,0 +1,431 @@
...
+    TxIn {
+        previous_output: OutPoint { txid, vout },
+        script_sig: script::Builder::new().into_script(),
+        witness: Witness::new(),
+        sequence: Sequence::ENABLE_RBF_NO_LOCKTIME,
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
storopoli, line 254: We should have a full flow tested here or somewhere (can also be moved to a separate PR) that ends with a `test_mempool_accept` or `submit_transaction` into a regtest `bitcoind-async-client` setup that verifies that this whole thing works.

crates/alpen-ee/common/src/traits/da.rs
```diff
@@ -37,3 +37,15 @@ pub trait BatchDaProvider: Send + Sync {
...
+
+/// Provides the encoded DA blob bytes for a batch.
+///
+/// Separates blob preparation (fetching state diffs, aggregating, encoding)
+/// from blob publication (chunking, posting to Bitcoin, tracking).
+#[cfg_attr(feature = "test-utils", mockall::automock)]
+pub trait DaBlobProvider: Send + Sync {
+    /// Returns the strata-codec encoded `BatchStateDiff` bytes for the given batch.
+    ///
+    /// Returns `None` if the batch has no state diffs.
```
evgenyzdanovich: why do we need to return None for empty statediff (instead of serialized bytes representation for the empty `BatchStateDiff`)? 
It can be useful and intended (well, depending on the usage patterns by the caller), but the semantic of returning `None` for the empty batch (_instead of indicating "non-existence" of the DA for the batch with BatchId_) is a bit surprising.

crates/storage/src/ops/chunked_envelope.rs
```diff
@@ -0,0 +1,15 @@
+//! Operations for reading/writing chunked envelope data from/to database.
+
+use strata_db_types::{traits::L1ChunkedEnvelopeDatabase, types::ChunkedEnvelopeEntry};
+
+use crate::{exec::*, instrumentation::components};
+
+inst_ops_simple! {
+    (<D: L1ChunkedEnvelopeDatabase> => ChunkedEnvelopeOps, component = components::STORAGE_CHUNKED_ENVELOPE) {
+        put_chunked_envelope_entry(idx: u64, entry: ChunkedEnvelopeEntry) => ();
```
evgenyzdanovich, line 9: IIUC, the `idx` is global key for all `Entries` across all ee da batches. 

For now, I assume it's alright. I guess the intention was _to allow a very simple loop for `Entries` to progress_: first oldest index of unfinalized by scanning, progress one-by-one from there.

this, however, means, that the caller (EE sequencer) should be responsible for juggling with idx, maintaining `BatchId => Vec<Idx>` and (potentially) deleting unnecessary entries.

In the end, this may be complicated to track and likely should be rewritten during the sequencer DA integration work.

crates/alpen-ee/common/src/types/da.rs
irnb: The `prev_chunk_wtxid` was moved from the witness to an `OP_RETURN` output to break a build order dependency. 

The commit tx must know all reveal P2TR addresses upfront (one output per reveal), which means all reveal scripts and therefore their taproot spend info must be finalized before the commit tx is constructed. 

If `prev_chunk_wtxid` were part of the chunk header in the witness, `reveal_script[N]` would depend on `reveal_tx[N-1]'s` `wtxid`, which requires the signed tx, which requires the commit tx, which requires `reveal_script[N]'s` P2TR address. 

Moving the linking to `OP_RETURN` (non-witness, doesn't affect the reveal script or P2TR address) lets us build all reveal scripts first, then the commit, then sign reveals sequentially with the `wtxid` chain in the outputs.
 
We don't use `SPS-50` here since that format carries ASM-specific fields (subprotocol_id, tx_type) that the EE DA layer doesn't need  we use a minimal `magic_bytes ++ prev_wtxid` tag instead. 

I'll update the Notion doc to document this.

crates/alpen-ee/common/src/types/da.rs
irnb: https://github.com/alpenlabs/alpen/pull/1340#discussion_r2778671269

crates/alpen-ee/common/src/types/da.rs
irnb: The spec's ~330,000 was an early conservative estimate. The overhead calculation above shows the budget supports more, I've bumped this to 395,000 in my next commit. Will update the Notion doc to match.

crates/btcio/src/writer/chunked_envelope/builder.rs
irnb: https://github.com/alpenlabs/alpen/pull/1340#discussion_r2778671269

crates/btcio/src/writer/chunked_envelope/builder.rs
irnb: Agreed. 
The next PR (`STR-1368-ee-da-integration`) adds a full EE DA flow functional test (`test_da_pipeline.py`) that runs against regtest  
it submits chunked envelopes, verifies mempool acceptance, confirms on-chain, and reassembles the DA blob from the committed chunks.   

The PR will be up by end of today.

crates/alpen-ee/common/src/traits/da.rs
irnb: Good callout, I've already reworked this locally. 

`get_blob` now returns `eyre::Result<DaBlob>` instead of `Option<Vec<u8>>`. The `DaBlob` struct wraps `batch_id (prev/last block hashes)`, `last_block_num`, and the `BatchStateDiff`. Even batches with no state changes return a `DaBlob` with an empty diff.
the metadata is still needed on L1 so a future sequencer can reconstruct the batch boundaries and resume the chain from DA alone. 

Non-existence is surfaced as an error through the Result, removing the ambiguity between "no state changes" and "batch not found.

irnb:
```
/// Compact summary of the last EVM block header in a batch.
///
/// Captures the subset of the EVM block header needed to build the next
/// block during DA-only chain reconstruction. A new sequencer recovering
/// from L1 DA has the [`BatchStateDiff`] (account/storage changes) but
/// **not** the block headers themselves,  these fields fill that gap.
///
/// - `base_fee`, `gas_used`, `gas_limit` feed the EIP-1559 base-fee
///   calculation and gas-limit adjustment for the next block.
/// - `timestamp` enforces monotonicity (`next > parent`).
/// - `block_num` identifies where the chain continues.
#[derive(Debug, Clone, Copy, PartialEq, Eq, Codec)]
pub struct EvmHeaderDigest {
    /// Block number of the last EVM block in this batch.
    pub block_num: u64,
    /// Unix timestamp (seconds) of the last EVM block.
    pub timestamp: u64,
    /// Base fee per gas (EIP-1559) of the last EVM block.
    pub base_fee: u64,
    /// Total gas consumed by the last EVM block.
    pub gas_used: u64,
    /// Gas limit of the last EVM block.
    pub gas_limit: u64,
}

/// DA blob containing batch metadata and state diff.
///
/// This is the top-level structure that gets encoded and posted to L1.
/// It wraps the batch state diff with identification metadata needed for
/// L1 sync and chain reconstruction.
#[derive(Debug, Clone, Codec)]
pub struct DaBlob {
    /// Batch identifier (prev_block_hash, last_block_hash)
    pub batch_id: BatchId,
    /// EVM header context of the last block in this batch.
    pub evm_header: EvmHeaderDigest,
    /// Aggregated state diff for the batch (can be empty for batches with no state changes)
    pub state_diff: BatchStateDiff,
}
```

@prajwolrg @storopoli @evgenyzdanovich
Guys, do you think we need any other fields here?

The way I see it: a new sequencer syncing purely from DA needs to
  (1) reconstruct state via the `state_diff` available in EE DA L1 txs,
  (2) know where to resume the chain  `batch_id` gives the prev/last block hashes for batch boundary verification, and
  (3) have enough parent block header context to build the next EVM block.

The EIP-1559 base-fee calculation requires the parent block's `base_fee`, `gas_used`, and
`gas_limit`, and these live in the block header, not in state tries. `timestamp` is also needed to enforce monotonicity. 
Without these fields a new sequencer would be unable to construct a valid next block.

Am I missing a case that would require additional metadata?

crates/storage/src/ops/chunked_envelope.rs
irnb: You're right that the global idx is batch-agnostic, that's intentional.

The envelope entry is basically a FIFO record for the `btcio chunked_envelope` watcher, so it doesn't need to know about batches. The plan is to handle the mapping on the batch side. 

On the EE sequencer layer, batches go through a lifecycle (`Sealed` → `DaPending` → `DaComplete` → ...)

When we post DA for a batch it transitions to `DaPending`, and once all DA txs finalize on L1
it moves to `DaComplete { da: Vec<L1DaBlockRef> }` with the permanent L1 references.                                        

What I'm thinking is to store the envelope index directly in the batch status: 
`DaPending { envelope_idx: u64 }`. 

So when the sequencer calls `submit_entry()` and gets back idx `37`, it writes `BatchStatus::DaPending { envelope_idx: 37 }` into the batch DB.  

On restart the batch itself tells you where to look **no in-memory HashMap** or external bookkeeping needed.

No changes needed in this PR for this, the chunked envelope DB provides the right primitives already. I'll carry the `BatchStatus` change in the integration PR.

irnb:
Aligned the spec documentation with the changes implemented in this PR.

@storopoli @evgenyzdanovich

crates/alpen-ee/common/src/types/da.rs
irnb: Fixed 
https://github.com/alpenlabs/alpen/pull/1340/changes/4943c6382e9c3311410105346f7719f43cf661aa

irnb:
FYI: I split the work by flow into separate commits and tried to make the commit messages review-friendly.

While doing the integration, I ran into a few issues/ improvements opportunity and fixed them in the later commits.

crates/alpen-ee/common/src/types/da.rs
```diff
@@ -0,0 +1,268 @@
...
+use strata_codec::{BufDecoder, Codec, CodecError};
+use strata_crypto::hash;
+use strata_identifiers::Buf32;
+
+// Bitcoin standardness limit is 400,000 weight units per tx. Reveal tx
+// overhead (weight units):
+//
+// - Non-witness (x4 multiplier):
+//   - Version, locktime, counts, outpoint, sequence: ~40 bytes -> 160 wu
+//   - OP_RETURN output (36-byte tag + script): ~45 bytes -> 180 wu
```
prajwolrg: It will be helpful to explain where does 36-byte come from here.

crates/alpen-ee/common/src/types/da.rs
```diff
@@ -0,0 +1,268 @@
...
+
+use strata_codec::{BufDecoder, Codec, CodecError};
+use strata_crypto::hash;
+use strata_identifiers::Buf32;
+
+// Bitcoin standardness limit is 400,000 weight units per tx. Reveal tx
+// overhead (weight units):
+//
+// - Non-witness (x4 multiplier):
+//   - Version, locktime, counts, outpoint, sequence: ~40 bytes -> 160 wu
```
prajwolrg: It might be worth to break this down as well. no of bytes for the version, locktime, counts, outpoint and sequence.

crates/db/types/src/types.rs
```diff
@@ -387,6 +387,84 @@ pub enum OLCheckpointStatus {
...
+///
+/// Used for posting large DA blobs that exceed single-transaction limits.
+/// Each reveal contains a chunk of the original blob with header metadata
+/// for reassembly.
+#[derive(Debug, Clone, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct ChunkedEnvelopeEntry {
+    /// Opaque witness data per reveal, ordered by output index.
+    pub chunk_data: Vec<Vec<u8>>,
+    /// OP_RETURN tag magic bytes.
+    pub magic_bytes: [u8; 4],
```
prajwolrg: nit: can use `MagicBytes` type here. 

edit: `MagicBytes` doesn't implement the borsh

crates/btcio/src/writer/chunked_envelope/builder.rs
prajwolrg: Linking https://github.com/alpenlabs/alpen/pull/1349/. Does it also test the scenario when the account doesn't have enough UTXOs to cover for the L1 tx fee?

prajwolrg: I think it might be worth doing some tests that can be done using `corepc`. It can be handled separately. It also avoids having to manually create dummy `ListUnspectItem`.

crates/btcio/src/writer/chunked_envelope/handle.rs
```diff
@@ -0,0 +1,377 @@
...
+                    Ok(updated) => {
+                        ops.put_chunked_envelope_entry_async(curr, updated).await?;
+                        debug!("entry signed successfully");
+                    }
+                    Err(EnvelopeError::NotEnoughUtxos(need, have)) => {
+                        error!(%need, %have, "waiting for sufficient utxos");
+                    }
+                    Err(e) => return Err(e.into()),
+                }
+            }
```
prajwolrg, line 154: There's a lot of things happening in the watcher task. We should test all these behavior thoroughly. Let's create tickets to test these if they aren't planned for upcoming PRs.

crates/alpen-ee/common/src/types/da.rs
```diff
@@ -156,111 +182,160 @@ pub fn prepare_da_chunks(blob: &[u8]) -> Result<Vec<Vec<u8>>, CodecError> {
...
+    decode_buf_exact(&bytes).ok()
+}
+
+/// Reassembles raw bytes from encoded chunks (header ++ payload each).
+///
+/// Performs the full pipeline: decode headers, reject unknown versions,
+/// order by `chunk_index`, concatenate payloads, compute the blob hash,
+/// and verify every chunk's claimed hash against the computed value.
+/// Returns `None` on any validation failure.
+fn reassemble_from_da_chunks(encoded_chunks: &[Vec<u8>]) -> Option<Vec<u8>> {
```
prajwolrg: nit: It might be worth having a version of this that returns an error on what went wrong. We can then have this function call the new function. This can be useful in debugging and testing.

prajwolrg (commented):
The PR is entangling EE side of things with L1 side of things. Although EE is what will be using the chunked envelope, they are two distinct things. We should try to make the PR more focussed on one thing. 

One example can be instead of having the (functional) tests for chunked envelope payload in a separate PR, it'd make more sense for it to be here. The payload doesn't really need to be be EE DA, and can just be random data to see if the chunking works as expected.

crates/db/types/src/types.rs
prajwolrg: I'll add the support as part of https://github.com/alpenlabs/strata-common/pull/73.

crates/alpen-ee/common/src/traits/da.rs
irnb: this shape was introduced in [4943c638](https://github.com/alpenlabs/alpen/commit/4943c638) (`feat(alpen-ee)introduce DaBlob envelope with batch metadata`).

crates/alpen-ee/common/src/types/da.rs
irnb: [Fixed](https://github.com/alpenlabs/alpen/pull/1340/changes/d6c0694b50251483773e156556c82901f6dc0d1d)

crates/alpen-ee/common/src/types/da.rs
irnb: Fixed

crates/btcio/src/writer/chunked_envelope/builder.rs
irnb: Addressed in https://github.com/alpenlabs/alpen/pull/1340/changes/8812d91a61e014519e042e272dab57b278206c9a#diff-e62579aa94fb3bb7fcaaa3fd2fd52a42907e13b43eb323a1c845b03772ccb0bfR419

crates/btcio/src/writer/chunked_envelope/handle.rs
irnb: [Fixed](https://github.com/alpenlabs/alpen/commit/8812d91a61e014519e042e272dab57b278206c9a)

Thanks for calling this out.

crates/alpen-ee/common/src/types/da.rs
irnb: [Fixed](https://github.com/alpenlabs/alpen/commit/d6c0694b50251483773e156556c82901f6dc0d1d)

crates/db/types/src/types.rs
irnb: [Fixed](https://github.com/alpenlabs/alpen/pull/1340/changes/b688d98334122721391680a87e95736a6b7f5d63)

crates/alpen-ee/common/src/traits/da.rs
```diff
@@ -37,3 +37,24 @@ pub trait BatchDaProvider: Send + Sync {
...
+    /// to ensure L1 chain continuity.
+    fn get_blob(&self, batch_id: BatchId) -> eyre::Result<DaBlob>;
+
+    /// Returns `true` if state diffs are ready for all blocks in the given batch.
+    ///
+    /// Used by the batch lifecycle to ensure state diffs have been written
+    /// by the Reth exex before attempting to post DA. This prevents race
+    /// conditions where DA posting is attempted before state diffs are ready.
+    fn are_state_diffs_ready(&self, batch_id: BatchId) -> eyre::Result<bool>;
+}
```
prajwolrg, line 61: nit: We are calling the trait `DaBlobProvider`, but it's also related with state diff. Should we split the trait or rename the trait to better reflect what this is.

crates/alpen-ee/common/src/types/da.rs
```diff
@@ -0,0 +1,385 @@
...
+/// block during DA-only chain reconstruction. A new sequencer recovering
+/// from L1 DA has the [`BatchStateDiff`] (account/storage changes) but
+/// **not** the block headers themselves — these fields fill that gap.
+///
+/// - `base_fee`, `gas_used`, `gas_limit` feed the EIP-1559 base-fee calculation and gas-limit
+///   adjustment for the next block.
+/// - `timestamp` enforces monotonicity (`next > parent`).
+/// - `block_num` identifies where the chain continues.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Codec)]
+pub struct EvmHeaderDigest {
```
prajwolrg: nit: Digest usually mean something related hash, but this seems to be opposite. It might be better to rename this to something else instead.

crates/alpen-ee/common/src/types/da.rs
```diff
@@ -0,0 +1,385 @@
...
+//   - OP_RETURN linking output: 48 B (39-byte script + value + script_len)
+//   - P2TR sequencer output: 43 B
+// - witness bytes = marker/flag (2) + stack item framing + sig + control block
+//   + witness script = 397,461 -> 397,461 wu
+//
+// Total reveal weight is 398,029 wu.
+// This is 1,971 wu below the 400,000 wu standardness limit.
+// Using 395,000 to keep a safe margin.
+/// Maximum chunk payload size in bytes.
+const MAX_CHUNK_PAYLOAD: usize = 395_000;
```
prajwolrg, line 75: Thanks. This is really nice. Maybe we could also add this information as part of the [spec](https://www.notion.so/SPS-envelope-1d1901ba000f807f9b36e491d765e786) here as well as in [envelope-fmt](https://github.com/alpenlabs/strata-common/tree/main/crates/l1proto/envelope-fmt) crate.

crates/btcio/src/writer/chunked_envelope/builder.rs
```diff
@@ -0,0 +1,473 @@
...
+/// a sequential chain: reveal 0 references `prev_tail_wtxid`, reveal 1
+/// references reveal 0's wtxid, etc.
+pub(crate) fn build_chunked_envelope_txs(
+    config: &EnvelopeConfig,
+    chunks: &[Vec<u8>],
+    magic_bytes: &[u8; 4],
+    prev_tail_wtxid: &Buf32,
+    utxos: Vec<ListUnspentItem>,
+) -> Result<ChunkedEnvelopeTxs, EnvelopeError> {
+    assert!(!chunks.is_empty(), "chunks must not be empty");
```
storopoli: should this be an error or a panic, like currently is?

storopoli (approved):
ACK c1df490ceff8cbcb93b5753efda8a8c8621848e3

crates/alpen-ee/common/src/types/da.rs
irnb: Yes, absolutely.

crates/alpen-ee/common/src/types/da.rs
irnb: Agreed, renamed to `EvmHeaderSummary`.

crates/alpen-ee/common/src/traits/da.rs
irnb: Renamed to `DaBlobSource`. "Source" better captures both readiness checking and data retrieval.

crates/btcio/src/writer/chunked_envelope/builder.rs
irnb: The panic was intentional since the only caller feeds chunks from `prepare_da_chunks` which guarantees non empty output  so hitting this would be a bug. 
But you're right, returning an error is nicer and easier to trace.  Switched it to `Err(EnvelopeError::EmptyPayload)`.

storopoli (dismissed):
ACK 41de079d9a31015ca17a2e43bcc965b75a61d7e2

crates/alpen-ee/common/src/types/da.rs
```diff
@@ -0,0 +1,385 @@
...
+/// 1       32    blob_hash
+/// 33      2     chunk_index
+/// 35      2     total_chunks
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Codec)]
+struct DaChunkHeader {
+    version: u8,
+    blob_hash: BlobHash,
+    chunk_index: u16,
+    total_chunks: u16,
```
bewakes, line 110: If this is going to L1 then maybe we can save a byte here by using a boolean or a `u8` indicating whether this is the terminal chunk or not.

bewakes: Not entirely sure, but if we ensure ordering of chunks(which perhaps from the commit - reveal chain, it should be), we can also possibly get away without the chunk index saving 2 more bytes.

crates/btcio/src/writer/chunked_envelope/handle.rs
```diff
@@ -0,0 +1,882 @@
...
+        L1TxStatus::InvalidInputs => {
+            unreachable!("InvalidInputs is handled before aggregation")
+        }
+    }
+}
+
+/// Returns `true` if `a` has made less broadcast progress than `b`.
+fn is_less_progressed(a: &L1TxStatus, b: &L1TxStatus) -> bool {
+    progress_ordinal(a) < progress_ordinal(b)
+}
```
bewakes, line 323: Does it make sense for `L1TxStatus` to implement `Ord`/`PartialOrd` ?

crates/alpen-ee/common/src/types/da.rs
irnb: Thanks for the suggestion, it’s a really interesting improvement. I created a separate ticket to track it.

https://alpenlabs.atlassian.net/browse/STR-2317

crates/btcio/src/writer/chunked_envelope/handle.rs
irnb: `Ord/PartialOrd` would still need to decide how to compare inner fields of `Confirmed/Finalized`. Since this ordering is specifically about "broadcast progress" and only needed here, keeping it local avoids baking a domain-specific interpretation into a shared type.

crates/alpen-ee/common/src/types/da.rs
```diff
@@ -0,0 +1,385 @@
...
+/// from L1 DA has the [`BatchStateDiff`] (account/storage changes) but
+/// **not** the block headers themselves — these fields fill that gap.
+///
+/// - `base_fee`, `gas_used`, `gas_limit` feed the EIP-1559 base-fee calculation and gas-limit
+///   adjustment for the next block.
+/// - `timestamp` enforces monotonicity (`next > parent`).
+/// - `block_num` identifies where the chain continues.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Codec)]
+pub struct EvmHeaderSummary {
+    /// Block number of the last EVM block in this batch.
```
prajwolrg, line 26: Don't we need the `parent_blockid` here?

crates/alpen-ee/common/src/traits/da.rs
irnb: https://alpenlabs.atlassian.net/browse/STR-2320

crates/alpen-ee/common/src/types/da.rs
prajwolrg: Discussed offchain with @irnb on this. This will be handled as part of https://alpenlabs.atlassian.net/browse/STR-2320.

PR #1324, opened 2026-01-30
Str 2112 new sequencer crate

bewakes:
## Description

This also contains code for new OL RPCs(STR-2108) for sequencer duties and checkpoint/block signing which is required for M2 and sequencer binary.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Is this PR addressing any specification, design doc or external reference document?

- [ ]  Yes
- [ ]  No

If yes, please add relevant links:

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have commented my code where necessary.
- [ ] I have updated the documentation if needed.
- [ ] My changes do not introduce new warnings.
- [ ] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [ ] New and existing tests pass with my changes.
- [ ] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

PR #1343, opened 2026-02-03
feat: new sequencer binary

storopoli:
## Description

This follows the same approach as the Alpen EE sequencer in `bin/alpen-client` which is to have the same binary be the full node or full node + sequencer given a feature flag.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

I've found a bug in the sequencer crate code that was recently merged #1324 see 071efa7a8fe6541ce00a5666393a9038dd7ef3d7

This PR was done with the help of Codex GPT-5.2 High.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-2113

### DIFF
--- a/bin/strata/src/services.rs
+++ b/bin/strata/src/services.rs
@@ -1,51 +1,186 @@
 //! Service spawning and lifecycle management.
 
 use std::sync::Arc;
-#[cfg(feature = "sequencer")]
-use std::time::Duration;
 
 use anyhow::{Result, anyhow};
 use strata_btcio::reader::query::bitcoin_data_reader_task;
-#[cfg(feature = "sequencer")]
-use strata_btcio::{
-    broadcaster::{L1BroadcastHandle, spawn_broadcaster_task},
-    writer::{EnvelopeHandle, start_envelope_task},
-};
 use strata_chain_worker_new::start_chain_worker_service_from_ctx;
-#[cfg(feature = "sequencer")]
-use strata_config::EpochSealingConfig;
 use strata_consensus_logic::{
     FcmContext, start_fcm_service,
     sync_manager::{spawn_asm_worker_with_ctx, spawn_csm_listener_with_ctx},
 };
-#[cfg(feature = "sequencer")]
-use strata_db_types::traits::DatabaseBackend;
 use strata_identifiers::OLBlockCommitment;
 use strata_node_context::NodeContext;
-#[cfg(feature = "sequencer")]
-use strata_ol_block_assembly::{
-    BlockasmBuilder, BlockasmHandle, FixedSlotSealing, MempoolProviderImpl,
-};
 use strata_ol_checkpoint::OLCheckpointBuilder;
 use strata_ol_mempool::{MempoolBuilder, MempoolHandle, OLMempoolConfig};
-#[cfg(feature = "sequencer")]
-use strata_ol_sequencer::TemplateManager;
-#[cfg(feature = "sequencer")]
-use strata_storage::ops::l1tx_broadcast;
 
-#[cfg(feature = "sequencer")]
-use crate::helpers::generate_sequencer_address;
-#[cfg(feature = "sequencer")]
-use crate::run_context::SequencerServiceHandles;
 use crate::{
     context::check_and_init_genesis,
     helpers::rollup_to_btcio_params,
     run_context::{RunContext, ServiceHandles},
 };
 
-/// TTL for template manager.
-// TODO: make this configurable.
-const TEMPLATE_MANAGER_TTL: Duration = Duration::from_secs(60);
+#[cfg(feature = "sequencer")]
+mod sequencer_services {
+    use std::{sync::Arc, time::Duration};
+
+    use anyhow::{Result, anyhow};
+    use strata_btcio::{
+        broadcaster::{L1BroadcastHandle, spawn_broadcaster_task},
+        writer::{EnvelopeHandle, start_envelope_task},
+    };
+    use strata_config::EpochSealingConfig;
+    use strata_db_types::traits::DatabaseBackend;
+    use strata_node_context::NodeContext;
+    use strata_ol_block_assembly::{
+        BlockasmBuilder, BlockasmHandle, FixedSlotSealing, MempoolProviderImpl,
+    };
+    use strata_ol_mempool::MempoolHandle;
+    use strata_ol_sequencer::TemplateManager;
+    use strata_storage::ops::l1tx_broadcast;
+
+    use crate::{
+        helpers::generate_sequencer_address,
+        run_context::{SequencerServiceHandles, ServiceHandlesBuilder},
+    };
+
+    /// TTL for template manager.
+    // TODO: make this configurable.
+    const TEMPLATE_MANAGER_TTL: Duration = Duration::from_secs(60);
+
+    pub(super) fn start_if_enabled(
+        nodectx: &NodeContext,
+        mempool_handle: Arc<MempoolHandle>,
+    ) -> Result<Option<SequencerServiceHandles>> {
+        if !nodectx.config().client.is_sequencer {
+            return Ok(None);
+        }
+
+        let broadcast_handle = Arc::new(start_broadcaster(nodectx));
+        let envelope_handle = start_writer(nodectx, broadcast_handle.clone())?;
+        let blockasm_handle = Arc::new(start_block_assembly(nodectx, mempool_handle)?);
+        let template_manager = Arc::new(TemplateManager::new(
+            blockasm_handle,
+            nodectx.storage().clone(),
+            TEMPLATE_MANAGER_TTL,
+        ));
+
+        Ok(Some(SequencerServiceHandles::new(
+            broadcast_handle,
+            envelope_handle,
+            template_manager,
+        )))
+    }
+
+    pub(super) fn attach_service_handles(
+        builder: ServiceHandlesBuilder,
+        sequencer_handles: Option<SequencerServiceHandles>,
+    ) -> ServiceHandlesBuilder {
+        builder.with_sequencer_handles(sequencer_handles)
+    }
+
+    /// Starts the L1 broadcaster task.
+    ///
+    /// Manages L1 transaction broadcasting and tracks confirmation status.
+    fn start_broadcaster(nodectx: &NodeContext) -> L1BroadcastHandle {
+        let broadcast_db = nodectx.storage().db().broadcast_db();
+        let broadcast_ctx = l1tx_broadcast::Context::new(broadcast_db);
+        let broadcast_ops = Arc::new(broadcast_ctx.into_ops(nodectx.storage().pool().clone()));
+
+        spawn_broadcaster_task(
+            nodectx.executor(),
+            nodectx.bitcoin_client().clone(),
+            broadcast_ops,
+            super::rollup_to_btcio_params(nodectx.params().rollup()),
+            nodectx.config().btcio.broadcaster.poll_interval_ms,
+        )
+    }
+
+    /// Starts the L1 writer/envelope task.
+    ///
+    /// Bundles L1 intents, creates envelope transactions, and publishes to Bitcoin.
+    fn start_writer(
+        nodectx: &NodeContext,
+        broadcast_handle: Arc<L1BroadcastHandle>,
+    ) -> Result<Arc<EnvelopeHandle>> {
+        let sequencer_address = nodectx
+            .task_manager()
+            .handle()
+            .block_on(generate_sequencer_address(nodectx.bitcoin_client()))?;
+
+        let writer_db = nodectx.storage().db().writer_db();
+
+        start_envelope_task(
+            nodectx.executor(),
+            nodectx.bitcoin_client().clone(),
+            Arc::new(nodectx.config().btcio.writer.clone()),
+            super::rollup_to_btcio_params(nodectx.params().rollup()),
+            sequencer_address,
+            writer_db,
+            nodectx.status_channel().as_ref().clone(),
+            nodectx.storage().pool().clone(),
+            broadcast_handle,
+        )
+    }
+
+    /// Starts the OL block assembly service.
+    ///
+    /// Assembles OL blocks from mempool transactions.
+    fn start_block_assembly(
+        nodectx: &NodeContext,
+        mempool_handle: Arc<MempoolHandle>,
+    ) -> Result<BlockasmHandle> {
+        let sequencer_config = nodectx
+            .config()
+            .sequencer
+            .clone()
+            .ok_or_else(|| anyhow!("Sequencer config required for block assembly"))?;
+
+        let epoch_sealing_config = nodectx.config().epoch_sealing.clone().unwrap_or_default();
+        let slots_per_epoch = match epoch_sealing_config {
+            EpochSealingConfig::FixedSlot { slots_per_epoch } => slots_per_epoch,
+        };
+
+        let mempool_provider = MempoolProviderImpl::new(mempool_handle);
+        let epoch_sealing = FixedSlotSealing::new(slots_per_epoch);
+        let state_provider = nodectx.storage().ol_state().clone();
+
+        nodectx.task_manager().handle().block_on(async {
+            BlockasmBuilder::new(
+                nodectx.params().clone(),
+                nodectx.storage().clone(),
+                mempool_provider,
+                epoch_sealing,
+                state_provider,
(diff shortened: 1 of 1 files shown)


bin/strata/src/sequencer/signer.rs
```diff
@@ -0,0 +1,71 @@
...
+    // Spawn the duty executor worker.
+    runctx.executor().spawn_critical_async(
+        "sequencer-duty-executor",
+        duty_executor_worker(
+            handles.template_manager().clone(),
+            handles.envelope_handle().clone(),
+            runctx.storage().clone(),
+            runctx.fcm_handle().clone(),
+            duty_rx,
+            runctx.task_manager.handle().clone(),
```
krsnapaudel: Does it make sense to have an accessor for this also? Other fields seem to have it.

bin/strata/src/main.rs
```diff
@@ -44,6 +46,15 @@ fn main() -> Result<()> {
...
     init_logging(rt.handle(), &config);
 
+    // Validate sequencer flag is enabled if sequencer feature is enabled
+    if args.sequencer && !cfg!(feature = "sequencer") {
+        return Err(anyhow!(
+            "Sequencer flag enabled but binary built without `sequencer` feature"
+        ));
+    }
+
+    let signer_args = args.clone();
```
krsnapaudel: Do we need a clone?

bin/strata/src/sequencer/signer.rs
storopoli: Yes, thanks for pointing it out check 94a2b6d473ca453cb35c1888706612f5996a033a

bin/strata/src/main.rs
storopoli: I've refactored in 94a2b6d473ca453cb35c1888706612f5996a033a

bin/strata/src/sequencer/duty_fetcher.rs
```diff
@@ -0,0 +1,57 @@
...
+        let duties =
+            match extract_duties(template_manager.as_ref(), tip_blkid, storage.as_ref()).await {
+                Ok(duties) => duties,
+                Err(err) => {
+                    error!("duty_fetcher_worker: failed to extract duties: {err}");
+                    continue;
+                }
+            };
+
+        info!(count = %duties.len(), "got new duties");
```
krsnapaudel: May be log only when non-empty.

bin/strata/src/sequencer/duty_executor.rs
```diff
@@ -0,0 +1,180 @@
...
+    envelope_handle: Arc<EnvelopeHandle>,
+    storage: Arc<NodeStorage>,
+    fcm_handle: Arc<strata_consensus_logic::FcmServiceHandle>,
+    duty: Duty,
+    sequencer_key: Buf32,
+    failed_duties_tx: mpsc::Sender<Buf32>,
+) {
+    let duty_id = duty.generate_id();
+    debug!(?duty_id, ?duty, "handle_duty");
+    let duty_result = match duty.clone() {
```
krsnapaudel: Do we need this clone?

bin/strata/src/sequencer/duty_executor.rs
```diff
@@ -0,0 +1,180 @@
...
+use tracing::{debug, error, info, warn};
+
+use super::helpers::{sign_checkpoint, sign_header};
+
+/// Worker for executing duties for the sequencer.
+pub(crate) async fn duty_executor_worker(
+    template_manager: Arc<TemplateManager>,
+    envelope_handle: Arc<EnvelopeHandle>,
+    storage: Arc<NodeStorage>,
+    fcm_handle: Arc<strata_consensus_logic::FcmServiceHandle>,
```
krsnapaudel: Nit: Why do we need the full path?

bin/strata/src/sequencer/duty_executor.rs
```diff
@@ -0,0 +1,180 @@
...
+        select! {
+            duty = duty_rx.recv() => {
+                if let Some(duty) = duty {
+                    let duty_id = duty.generate_id();
+                    if seen_duties.contains(&duty_id) {
+                        debug!(?duty_id, "skipping already seen duty");
+                        continue;
+                    }
+                    seen_duties.insert(duty_id);
+                    handle.spawn(handle_duty(
```
krsnapaudel, line 45: Is there a benefit in doing `handle.spawn` here? Claude Code thinks you could just do `tokio::spawn` and skip the `handle` argument.

bin/strata/src/sequencer/signer.rs
```diff
@@ -0,0 +1,71 @@
...
+    runctx.executor().spawn_critical_async(
+        "sequencer-duty-executor",
+        duty_executor_worker(
+            handles.template_manager().clone(),
+            handles.envelope_handle().clone(),
+            runctx.storage().clone(),
+            runctx.fcm_handle().clone(),
+            duty_rx,
+            runctx.task_manager().handle().clone(),
+            sequencer_key.sk,
```
krsnapaudel, line 65: `sequencer_key.sk` is Copy right?
Does `ZeroizeOnDrop` help?

bin/strata/src/main.rs
```diff
@@ -44,8 +46,15 @@ fn main() -> Result<()> {
     // Initialize logging
     init_logging(rt.handle(), &config);
 
+    // Validate sequencer flag is enabled if sequencer feature is enabled
+    if args.sequencer && !cfg!(feature = "sequencer") {
```
krsnapaudel: Does `args.sequencer` exist if it's `!cfg!(feature = "sequencer")`

Is this better?
```
#[cfg(not(feature = "sequencer"))]
if args.sequencer {
    return Err(anyhow!("Binary built without `sequencer` feature"));
}
```

bin/strata/src/services.rs
```diff
@@ -63,15 +76,21 @@ pub(crate) fn start_strata_services(nodectx: NodeContext) -> Result<RunContext>
...
+    // Sequencer-specific tasks (only started when is_sequencer=true)
+    #[cfg(feature = "sequencer")]
     let sequencer_handles = if nodectx.config().client.is_sequencer {
         let broadcast_handle = Arc::new(start_broadcaster(&nodectx));
         let envelope_handle = start_writer(&nodectx, broadcast_handle.clone())?;
         let blockasm_handle = Arc::new(start_block_assembly(&nodectx, mempool_handle.clone())?);
+        let template_manager = Arc::new(TemplateManager::new(
+            blockasm_handle,
+            nodectx.storage().clone(),
+            Duration::from_millis(60_000), // 60 seconds
```
krsnapaudel: Should we define the constant somewhere?

bin/strata/src/sequencer/mod.rs
```diff
@@ -0,0 +1,10 @@
+//! OL RPC server implementation for sequencer node.
```
krsnapaudel: Wrong docstring?

bin/strata/src/sequencer/duty_fetcher.rs
```diff
@@ -0,0 +1,57 @@
...
+) -> anyhow::Result<()> {
+    let mut interval = interval(poll_interval);
+    'top: loop {
+        interval.tick().await;
+        let tip_blkid = match status_channel.get_ol_sync_status().map(|s| *s.tip_blkid()) {
+            Some(tip) => tip,
+            None => match storage.ol_block().get_canonical_block_at_async(0).await {
+                Ok(Some(commitment)) => *commitment.blkid(),
+                Ok(None) => {
+                    warn!("duty_fetcher_worker: genesis block not found yet");
```
krsnapaudel: Per codebase conventions, use structured tracing fields or `#[tracing::instrument]` instead of manually prefixing function names in log strings. I think other places have something similar.
`duty_fetcher.rs` (lines 27, 31, 41, 50)

bin/strata/src/run_context.rs
```diff
@@ -36,32 +37,47 @@ impl RunContext {
...
     pub(crate) fn status_channel(&self) -> &Arc<StatusChannel> {
         self.common.status_channel()
     }
 
+    /// Returns the mempool handle.
     pub(crate) fn mempool_handle(&self) -> &Arc<MempoolHandle> {
         &self.service_handles.mempool_handle
     }
 
+    #[cfg(feature = "sequencer")]
```
krsnapaudel: Lines 60, 71, 77 in `run_context.rs`

Should docstring come first?

bin/strata/src/rpc/mod.rs
```diff
@@ -83,30 +102,36 @@ async fn spawn_rpc(deps: RpcDeps) -> Result<()> {
...
-    // Create sequencer rpc handler if sequencer
-    if let Some(seq_deps) = deps.seq_deps {
-        let block_template_cache_ttl = Duration::from_millis(60000); // One minute
-        let tmp_mgr = TemplateManager::new(
-            seq_deps.blockasm_handle().clone(),
-            deps.storage.clone(),
-            block_template_cache_ttl,
-        );
+    // Create and register OL fullnode RPC server
+    let ol_fullnode_server = OLRpcServer::new(
```
krsnapaudel: Do we need two of the rpc servers? Can we do something like this?
```
let ol_rpc_server = OLRpcServer::new(
    deps.storage.clone(),
    deps.status_channel.clone(),
    deps.mempool_handle.clone(),
);
let ol_fullnode_server = ol_rpc_server.clone();  // if OLRpcServer: Clone

let ol_module = OLClientRpcServer::into_rpc(ol_rpc_server);
let ol_fullnode_module = OLFullNodeRpcServer::into_rpc(ol_fullnode_server);
```

delbonis: I think it's probably actually fine.  The term "server" is a little misleading here since they're all the request handlers under the same actual *listener*.

bin/strata/src/sequencer/rpc.rs
```diff
@@ -23,13 +24,21 @@ use crate::rpc::errors::{db_error, internal_error, not_found_error};
...
+
+    /// Status channel.
     status_channel: Arc<StatusChannel>,
+
+    /// Template manager.
     template_manager: Arc<TemplateManager>,
+
+    /// Envelope handle.
     envelope_handle: Arc<EnvelopeHandle>,
 }
```
delbonis, line 38: Can we hide these behind a provider trait to make the components a little more independent?  I realized it's pretty annoying to have the RPC server impl be forced all the way down into the binary crate.

See how I made [the skeleton](https://github.com/alpenlabs/mosaic/blob/main/crates/rpc/server/src/server.rs#L10) in the Mosaic repo.

bin/strata/src/run_context.rs
```diff
@@ -70,56 +86,81 @@ impl RunContext {
...
     pub(crate) fn new(
         asm_handle: Arc<AsmWorkerHandle>,
         csm_monitor: Arc<ServiceMonitor<CsmWorkerStatus>>,
         mempool_handle: Arc<MempoolHandle>,
         chain_worker_handle: Arc<ChainWorkerHandle>,
         checkpoint_handle: Arc<OLCheckpointWorkerHandle>,
         fcm_handle: Arc<FcmServiceHandle>,
-        sequencer_handles: Option<SequencerServiceHandles>,
+        #[cfg(feature = "sequencer")] sequencer_handles: Option<SequencerServiceHandles>,
     ) -> Self {
```
delbonis: Should probably use a builder for this type instead of making the function's number of arguments dependent on the feature flags, since then we have to be careful in more places.

bin/strata/src/run_context.rs
```diff
@@ -70,56 +86,81 @@ impl RunContext {
...
+
+    /// Handle for the checkpoint worker.
     checkpoint_handle: Arc<OLCheckpointWorkerHandle>,
+
+    /// Handle for the FCM service.
     fcm_handle: Arc<FcmServiceHandle>,
-    /// Sequencer-specific handles (None for fullnodes)
+
+    #[cfg(feature = "sequencer")]
+    /// Handles for sequencer-specific services ([`None`] when not running as sequencer).
```
delbonis: comments before attributes!

bin/strata/src/services.rs
delbonis, line 1: Making all this conditional on the sequencer feature flag is ugly, is there a better way to structure this?

Worst case we can do something like:

```rs
#[cfg(...)]
mod seq_imports {
   // pub use ...;
}

#[cfg(...)]
use seq_imports::*;
```

bin/strata/src/sequencer/duty_executor.rs
storopoli: using `handle.spawn` specifically (vs `tokio::spawn`) mainly guarantees it runs on that runtime `handle`; behavior is otherwise similar if you are already inside the same runtime. I think it is more future proof than using tokio::spawn.

bin/strata/src/sequencer/duty_fetcher.rs
storopoli: Yes nice suggestion see da84b5147dfa0678b3f6d6fd043293e95c64e73d

bin/strata/src/sequencer/duty_executor.rs
storopoli: No, nice catch. Removed in da84b5147dfa0678b3f6d6fd043293e95c64e73d

bin/strata/src/sequencer/duty_executor.rs
storopoli: Removed here and somewhere else that used the same full path in da84b5147dfa0678b3f6d6fd043293e95c64e73d

bin/strata/src/sequencer/signer.rs
storopoli: Yes and `ZeroizeOnDrop` does not help
https://github.com/alpenlabs/alpen/blob/a9af22464f95f0762f87c696fb820d1727cfd7c7/crates/identifiers/src/buf.rs#L194-L200

I did some workaround here in da84b5147dfa0678b3f6d6fd043293e95c64e73d

bin/strata/src/main.rs
storopoli: Yes, thanks! Much better. Applied in da84b5147dfa0678b3f6d6fd043293e95c64e73d

bin/strata/src/services.rs
storopoli: Yes done in da84b5147dfa0678b3f6d6fd043293e95c64e73d with a `TODO` to make it configurable in the future if we ever need to change that.

bin/strata/src/sequencer/mod.rs
storopoli: Yes, changed in da84b5147dfa0678b3f6d6fd043293e95c64e73d

bin/strata/src/sequencer/duty_fetcher.rs
storopoli: Nice catch. Been enforcing this to my reviews as well. Done in da84b5147dfa0678b3f6d6fd043293e95c64e73d

bin/strata/src/run_context.rs
storopoli: Yes, changed in da84b5147dfa0678b3f6d6fd043293e95c64e73d

Sometimes I confuse them...

bin/strata/src/rpc/mod.rs
storopoli: Changed the comment and the binding to `listener` in da84b5147dfa0678b3f6d6fd043293e95c64e73d

bin/strata/src/run_context.rs
storopoli: Changed in da84b5147dfa0678b3f6d6fd043293e95c64e73d

bin/strata/src/sequencer/rpc.rs
storopoli: Yes, you pointed this out to me already in the past. I will defer this to a tech-debt ticket: https://alpenlabs.atlassian.net/browse/STR-2211

bin/strata/src/run_context.rs
storopoli: Check 0f086e327495f544b32691bcea0679436afe65ab

bin/strata/src/services.rs
storopoli: That's neat! Thanks for the suggestion. Done in 9540740bd4626c0aedc1a91658b58b4ef0327e48

PR #1345, opened 2026-02-04
feat(strata): insert epoch summary for epoch 0/genesis in the database

storopoli:
## Description

Insert epoch summary for epoch 0 in the database while we do the OL block genesis so that when we refer to “previous epoch” we will always have something present and don’t have to do checks if it is a null epoch or not. Currently in fcm we have to do a special case check to see if the epoch is null epoch or not.

It also fixes a bug in `chain-worker-new` which was using the "old" checkpoint DB (see d74ccf1efcf3fc8afc979b8eee7c58ba0a157484)

This also moves the OL genesis logic out of the binary and into an `ol-genesis` crate.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Tried to keep the changes very compact and localized.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-2156

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -616,6 +616,9 @@ uv run python entry.py
 - `factories/` - Service factories (Bitcoin, Strata)
 - `tests/` - Test files
 
+If the functional tests fail, you can find the logs in the `_dd` directory inside the functional tests directory.
+The datadir will be the outputted by the test framework and will be named after the test run.
+
 ### Functional Tests (Legacy)
 
 Located in `functional-tests/`. Uses `uv` for dependency management.
@@ -632,6 +635,9 @@ cd functional-tests
 # Start test environment
 ./run_test.sh -e basic
 ```
+
+If the functional tests fail, you can find the logs in the `_dd` directory inside the functional tests directory.
+The datadir will be the outputted by the test framework and will be named after the test run.
 
 ### Prover Tests
 


crates/ol/genesis/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+/// Construct genesis state + block artifacts using a supplied manifest.
+#[instrument(skip_all, fields(component = "ol_genesis"))]
+pub fn build_genesis_artifacts_with_manifest(
+    params: &Params,
+    genesis_manifest: AsmManifest,
+) -> Result<GenesisArtifacts> {
+    info!("building OL genesis block and state");
+
+    // Create initial OL state (uses genesis defaults).
+    // TODO: initialize with a Snark EE account for Alpen. Possibly with rollup params.
```
bewakes: While we are here, can we just initialize an account with vk that's `PredicateKey::always_accept()` and some state root? Because for the next deployment we don't care about the validity of EE stf transition. This is based on [this conversation](https://alpen-labs.slack.com/archives/C09C6JLF76H/p1770230782127279) in slack.

cc: @barakshani

crates/ol/genesis/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+pub type Result<T> = StdResult<T, GenesisError>;
+
+/// Build the default genesis manifest from rollup params.
+pub fn default_genesis_manifest(params: &Params) -> AsmManifest {
+    let genesis_l1 = &params.rollup().genesis_l1_view;
+
+    AsmManifest::new(
+        genesis_l1.height_u64(),
+        genesis_l1.blkid(),
+        // TODO: Properly fetch manifest from db and populate this, btc reader should read L1 and
```
bewakes: This should be doable now because we have btcio reader is up along with asm and csm. We can use the waiting logic in the previous genesis code. But tbh, we could get away without this for M2, because we don't expect anything to be in the genesis L1 block.

crates/ol/genesis/src/lib.rs
storopoli: See

- ad1757c6d689e15a09439866247653e1c554a396
- 2d3587ade725c27e9859fd366f393a97af647c59
- ad34688581472fdfb58acd00912603b770834205

crates/ol/genesis/src/lib.rs
storopoli: See bfa49fc8fcdf25583e8afb812a46b7b3943c415b

PR #1346, opened 2026-02-06
feat(ol-da): OL DA Blob fetching library

storopoli:
## Description

Implement OL DA fetching blobs so nodes can fetch the OL payload for checkpoint transaction.

I had to do some refactor in `asm` to reuse the test-utilities and also corrected an invalid Taproot `OP_CHECKMULTSIG` opcode (Cc @prajwolrg).

I also realized that we don't need to `Clone` `BitcoinTxid` which is a tuple wrapper struct around `Txid` which is `Copy`, so I derived it for `BitcoinTxid` as well.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [x] Dependency Update

## Notes to Reviewers

- Not used AI, due to having bloated stuff in previous attempts at this PR.

- [x]  Yes
- [ ]  No

- [SPS-ol-da-structure](https://www.notion.so/SPS-ol-da-structure-Orchestration-Layer-Data-Availability-Structure-2ea901ba000f800f80e5d249b08a82c4)
- [SPS-da-primitives](https://www.notion.so/SPS-da-primitives-Data-Availability-Primitives-2e9901ba000f804f94c4e0d26282fe64)
- [SPS-62](https://www.notion.so/SPS-62-OL-checkpoint-structure-and-verification-225901ba000f80cfb3cff69d0c916054) / [SPS-63](https://www.notion.so/SPS-63-OL-checkpointing-ASM-subprotocol-v1-201901ba000f80fba6bce24f77dbac4e)

## Checklist

## Related Issues

STR-1883

### DIFF
--- a/crates/ol/da/src/errors.rs
+++ b/crates/ol/da/src/errors.rs
@@ -1,5 +1,10 @@
+use bitcoin::consensus::encode;
 use strata_acct_types::AccountSerial;
+use strata_asm_proto_checkpoint_txs::CheckpointTxError;
+use strata_btc_types::BitcoinTxid;
+use strata_codec::CodecError;
 use strata_da_framework::DaError as FrameworkDaError;
+use strata_l1_txfmt::TxFmtError;
 use thiserror::Error;
 
 pub type DaResult<T> = Result<T, DaError>;
@@ -20,4 +25,24 @@ pub enum DaError {
 
     #[error("{0}")]
     Other(&'static str),
+}
+
+pub type DaExtractorResult<T> = Result<T, DaExtractorError>;
+
+#[derive(Debug, Error)]
+pub enum DaExtractorError {
+    #[error("failed to decode raw bitcoin transaction: {0}")]
+    BitcoinTxDecodeError(#[from] encode::Error),
+
+    #[error("failed to fetch checkpoint transaction: {0:?}")]
+    BitcoinTxNotFound(BitcoinTxid),
+
+    #[error("checkpoint transaction failed: {0}")]
+    CheckpointTxError(#[from] CheckpointTxError),
+
+    #[error("SPS-50 tag parsing failed: {0}")]
+    TagParse(#[from] TxFmtError),
+
+    #[error("OL DA payload decode failed: {0}")]
+    DaPayloadDecode(#[from] CodecError),
 }
--- a/crates/ol/da/src/extractor.rs
+++ b/crates/ol/da/src/extractor.rs
@@ -1,0 +1,110 @@
+//! # Checkpoint transaction extraction helpers for OL DA payload consumption.
+//!
+//! This is a very simple module that given any bitcoin client that implements the [`Reader`] trait,
+//! can fetch a checkpoint transaction by [`BitcoinTxid`] and decode its OL DA payload, namely
+//! [`OLDaPayloadV1`].
+
+use bitcoin::Transaction;
+use bitcoind_async_client::traits::Reader;
+use strata_asm_common::TxInputRef;
+use strata_asm_proto_checkpoint_txs::parser::extract_signed_checkpoint_from_envelope;
+use strata_btc_types::{BitcoinTxid, RawBitcoinTx};
+use strata_codec::decode_buf_exact;
+use strata_l1_txfmt::{MagicBytes, ParseConfig};
+
+use crate::{DaExtractorError, DaExtractorResult, OLDaPayloadV1};
+
+/// Decodes the OL DA payload from a checkpoint transaction.
+pub async fn decode_ol_da_payload<R>(
+    reader: &R,
+    checkpoint_txid: &BitcoinTxid,
+    magic_bytes: MagicBytes,
+) -> DaExtractorResult<OLDaPayloadV1>
+where
+    R: Reader,
+{
+    let raw_tx = fetch_raw_tx_from_reader(reader, checkpoint_txid).await?;
+    let tx: Transaction = raw_tx.try_into()?;
+    let tag = ParseConfig::new(magic_bytes).try_parse_tx(&tx)?;
+    let signed_checkpoint = extract_signed_checkpoint_from_envelope(&TxInputRef::new(&tx, tag))?;
+    let da_payload = decode_buf_exact(signed_checkpoint.inner().sidecar().ol_state_diff())?;
+    Ok(da_payload)
+}
+
+/// Fetches a raw Bitcoin transaction from a [`Reader`] by [`BitcoinTxid`].
+async fn fetch_raw_tx_from_reader<R>(
+    reader: &R,
+    checkpoint_txid: &BitcoinTxid,
+) -> DaExtractorResult<RawBitcoinTx>
+where
+    R: Reader,
+{
+    let txid = checkpoint_txid.inner();
+    let raw_tx_response = reader
+        .get_raw_transaction_verbosity_zero(&txid)
+        .await
+        .map_err(|_| DaExtractorError::BitcoinTxNotFound(*checkpoint_txid))?;
+
+    Ok(RawBitcoinTx::from(raw_tx_response.0))
+}
+
+#[cfg(test)]
+mod tests {
+
+    use bitcoin::ScriptBuf;
+    use strata_asm_common::test_utils::create_reveal_transaction_stub;
+    use strata_asm_proto_checkpoint_txs::{
+        CHECKPOINT_V0_SUBPROTOCOL_ID, CheckpointTxError, OL_STF_CHECKPOINT_TX_TYPE,
+    };
+    use strata_l1_envelope_fmt::parser::parse_envelope_payload;
+    use strata_l1_txfmt::TagDataRef;
+
+    use super::*;
+    use crate::DaExtractorError;
+
+    /// Magic bytes for testing purposes.
+    const TEST_MAGIC_BYTES: MagicBytes = MagicBytes::new(*b"ALPN");
+
+    /// Creates a checkpoint transaction with the given payload, subprotocol, tx type, and secret
+    /// key.
+    fn make_checkpoint_tx(payload: &[u8], subprotocol: u8, tx_type: u8) -> Transaction {
+        let tag_data = TagDataRef::new(subprotocol, tx_type, &[]).expect("build tag");
+        let tag_script = ParseConfig::new(TEST_MAGIC_BYTES)
+            .encode_script_buf(&tag_data)
+            .expect("encode tag script");
+        create_reveal_transaction_stub(payload.to_vec(), tag_script.into_bytes())
+    }
+
+    /// Extracts the leaf script from a transaction.
+    fn extract_leaf_script(tx: &Transaction) -> DaExtractorResult<ScriptBuf> {
+        if tx.input.is_empty() {
+            return Err(DaExtractorError::CheckpointTxError(
+                CheckpointTxError::MissingInputs,
+            ));
+        }
+
+        tx.input[0]
+            .witness
+            .taproot_leaf_script()
+            .map(|leaf| leaf.script.into())
+            .ok_or(DaExtractorError::CheckpointTxError(
+                CheckpointTxError::MissingLeafScript,
+            ))
+    }
+
+    #[test]
+    fn test_make_checkpoint_tx_envelope_roundtrip_large_payload() {
+        let payload = vec![0xAB; 1_300];
+        assert!(payload.len() > 520, "payload must exceed single push limit");
+
+        let tx = make_checkpoint_tx(
+            &payload,
+            CHECKPOINT_V0_SUBPROTOCOL_ID,
+            OL_STF_CHECKPOINT_TX_TYPE,
+        );
+
+        let script = extract_leaf_script(&tx).expect("extract envelope-bearing leaf script");
+        let parsed_payload = parse_envelope_payload(&script).expect("parse envelope payload");
+        assert_eq!(parsed_payload, payload);
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15010,10 +15010,17 @@ dependencies = [
 name = "strata-ol-da"
 version = "0.1.0"
 dependencies = [
+ "bitcoin",
+ "bitcoind-async-client",
  "strata-acct-types",
+ "strata-asm-common",
+ "strata-asm-proto-checkpoint-txs",
+ "strata-btc-types",
  "strata-codec",
  "strata-da-framework",
  "strata-identifiers",
+ "strata-l1-envelope-fmt",
+ "strata-l1-txfmt",
  "strata-ledger-types",
  "strata-ol-params",
  "strata-ol-state-types",
--- a/crates/ol/da/Cargo.toml
+++ b/crates/ol/da/Cargo.toml
@@ -5,18 +5,26 @@ edition = "2024"
 
 [dependencies]
 strata-acct-types.workspace = true
+strata-asm-common.workspace = true
+strata-asm-proto-checkpoint-txs.workspace = true
+strata-btc-types.workspace = true
 strata-codec.workspace = true
 strata-da-framework.workspace = true
 strata-identifiers.workspace = true
+strata-l1-txfmt.workspace = true
 strata-ledger-types.workspace = true
 strata-predicate.workspace = true
 strata-snark-acct-types.workspace = true
 
+bitcoin.workspace = true
+bitcoind-async-client.workspace = true
 thiserror.workspace = true
 
 [lints]
 workspace = true
 
 [dev-dependencies]
+strata-asm-common = { workspace = true, features = ["test-utils"] }
+strata-l1-envelope-fmt.workspace = true
 strata-ol-params.workspace = true
 strata-ol-state-types.workspace = true
--- a/crates/ol/da/src/lib.rs
+++ b/crates/ol/da/src/lib.rs
@@ -1,12 +1,11 @@
 //! DA scheme implementation types.
 
-// TODO will probably use later
-use strata_identifiers as _;
-
 mod errors;
+mod extractor;
 mod traits;
 mod types;
 
 pub use errors::*;
+pub use extractor::*;
 pub use traits::*;
 pub use types::*;


bin/strata/src/services.rs
```diff
@@ -34,19 +34,20 @@ pub(crate) fn start_strata_services(nodectx: NodeContext) -> Result<RunContext>
...
-
     // btcio reader task must start before genesis init because genesis requires ASM to
     // have the genesis manifest which will be available only after btcio reader provides
     // the L1 block to ASM.
     start_btcio_reader(&nodectx, asm_handle.clone());
 
-    // Check and do genesis if not yet. This should be done after asm/csm/btcio and before mempool
-    // because genesis requires asm to be working and mempool and other services expect genesis to
+    // Check and do genesis if not yet. This should be done after asm/btcio and before csm/mempool
+    // because genesis requires asm to be working and other services expect genesis to
```
barakshani:
```suggestion
    // Check for genesis and initialize if it hasn't. This should be done after starting asm/btcio and before starting 
    // csm/mempool, because genesis requires asm to be working and other services expect genesis to
```
sorry, the language here bothered me too much.

bin/strata/src/services.rs
```diff
@@ -34,19 +34,20 @@ pub(crate) fn start_strata_services(nodectx: NodeContext) -> Result<RunContext>
...
     // have happened.
     check_and_init_genesis(nodectx.storage().as_ref(), nodectx.params().as_ref())?;
 
+    // Start Csm worker after genesis to avoid checkpoint DA processing before genesis OL state is
```
barakshani:
```suggestion
    // Start csm worker after genesis to avoid checkpoint DA processing before genesis OL state is

File: bin/strata/src/services.rs [hunk 3]
```diff
@@ -34,19 +34,20 @@ pub(crate) fn start_strata_services(nodectx: NodeContext) -> Result<RunContext>
...
+    // initialized.
+    let csm_monitor = Arc::new(spawn_csm_listener_with_ctx(&nodectx, asm_handle.monitor())?);
```
barakshani: Can you explain the comment? What exactly do you refer to by "checkpoint DA processing" and why is this a particular thing we want to avoid (and not any kind of CSM processing)?

crates/csm-worker/src/processor.rs
```diff
@@ -87,6 +110,327 @@ fn process_checkpoint_log(
     Ok(())
 }
 
+/// Fetches, decodes, validates, applies, and persists OL DA for one checkpoint update.
```
barakshani:
```suggestion
/// Fetches, decodes, validates, applies, and persists OL DA for one checkpoint.
```
For good hygiene, let's not mix "checkpoints" with "updates".

crates/csm-worker/src/processor.rs
```diff
@@ -87,6 +110,327 @@ fn process_checkpoint_log(
...
+///
+/// This routine reconstructs the checkpoint preseal transition from L1 data and validates it
+/// against ASM-provided commitments before any checkpoint inclusion metadata is written:
+///
+/// 1. Fetches the checkpoint transaction from the configured Bitcoin client tx provider.
+/// 2. Decodes and validates the checkpoint DA payload from the sidecar.
+/// 3. Resolves checkpoint pre-state and verifies the pre-state root.
+/// 4. Applies DA payload, then applies L1-derived epoch sealing updates.
+/// 5. Verifies the post-state root and persists the resulting OL state snapshot.
+fn apply_checkpoint_da_update(
```
barakshani:
```suggestion
fn apply_checkpoint_da(

File: crates/csm-worker/src/processor.rs [hunk 6]
```diff
@@ -87,6 +110,327 @@ fn process_checkpoint_log(
...
+    state: &mut CsmWorkerState,
+    checkpoint_update: &CheckpointUpdate,
```
barakshani: ok, so looks like checkpoint update is thing : (

crates/csm-worker/src/processor.rs
barakshani: I have a different issue with this function name. While called _apply da_, it does much more then this -- apply da is only in point 4 out of the 5 different things this function does.

crates/csm-worker/src/processor.rs
```diff
@@ -87,6 +110,327 @@ fn process_checkpoint_log(
...
+                "failed to persist OL state for checkpoint terminal commitment {}",
+                terminal_commitment
+            )
+        })?;
+
+    Ok(())
+}
+
+/// Validates that the decoded checkpoint payload tip matches ASM checkpoint update metadata.
+fn validate_checkpoint_payload_tip(
```
barakshani: I am trying to understand the big picture here.
We get here from `apply_checkpoint_da_update` which is called by `process_checkpoint_log` where the ASM provides as a log with the `CheckpointUpdate` and we pull out the tx id and decode the checkpoint from it.
Here we verify that the values in the ASM log are compatible with what we just decoded. Shouldn't  we panic on _any_ mismatch?

crates/csm-worker/src/processor.rs
```diff
@@ -87,6 +110,327 @@ fn process_checkpoint_log(
...
+        *header.parent_blkid(),
+    ))
+}
+
+/// Resolves the pre-state slot from the checkpoint range start slot.
+fn resolve_pre_state_slot(l2_start: OLBlockCommitment) -> anyhow::Result<u64> {
+    l2_start
+        .slot()
+        .checked_sub(1)
+        .ok_or_else(|| anyhow!("invalid checkpoint L2 start slot 0 for pre-state resolution"))
```
barakshani: What exactly is the error here? that we are at slot 0?

crates/csm-worker/src/processor.rs
```diff
@@ -87,6 +110,327 @@ fn process_checkpoint_log(
     Ok(())
 }
 
+/// Fetches, decodes, validates, applies, and persists OL DA for one checkpoint update.
+///
+/// This routine reconstructs the checkpoint preseal transition from L1 data and validates it
+/// against ASM-provided commitments before any checkpoint inclusion metadata is written:
+///
+/// 1. Fetches the checkpoint transaction from the configured Bitcoin client tx provider.
+/// 2. Decodes and validates the checkpoint DA payload from the sidecar.
```
barakshani: What exactly do you mean by "validates the checkpoint DA payload"?

crates/csm-worker/src/processor.rs
```diff
@@ -87,6 +110,327 @@ fn process_checkpoint_log(
...
+/// 3. Resolves checkpoint pre-state and verifies the pre-state root.
```
barakshani: Out of curiosity, what do you mean when you say resolve here (I already went over the code, and I see where you use it)?

crates/csm-worker/src/processor.rs
```diff
@@ -87,6 +110,327 @@ fn process_checkpoint_log(
...
+            "pre-state root mismatch for epoch {}: expected {}, got {} (pre-state commitment: {})",
+            checkpoint_update.batch_info().epoch(),
+            checkpoint_update.chainstate_transition().pre_state_root,
+            pre_root,
+            pre_state_commitment
+        );
+    }
+
+    apply_da_payload(&mut pre_state, decoded.da_payload)
+        .context("failed to apply OL DA payload")?;
```
barakshani: is there a way to match the _preseal_ state before applying the next function, or we don't post it as part of the checkpoint?

crates/csm-worker/src/processor.rs
```diff
@@ -87,6 +110,327 @@ fn process_checkpoint_log(
...
+    state
+        .storage
+        .ol_state()
+        .put_toplevel_ol_state_blocking(terminal_commitment, pre_state)
+        .with_context(|| {
+            format!(
+                "failed to persist OL state for checkpoint terminal commitment {}",
+                terminal_commitment
+            )
+        })?;
```
barakshani: The sequencer, and in the future any node that _also_ syncs via L2, will already have this persisted, is that correct? How do we imagine the interplay between L2-given data and checkpoint-derived data?

crates/csm-worker/src/processor.rs
delbonis: I think it's because we have to special case if we're at the genesis block since the genesis block kinda doesn't have a pre-state since it's not executed normally.

crates/csm-worker/src/processor.rs
```diff
@@ -87,6 +110,327 @@ fn process_checkpoint_log(
...
+    ol_state: &mut OLState,
+) -> anyhow::Result<()> {
+    let start_height = batch_info.l1_range.0.height_u64();
+    let end_height = batch_info.l1_range.1.height_u64();
+    if end_height < start_height {
+        bail!(
+            "invalid L1 range in checkpoint batch info: start={} end={}",
+            start_height,
+            end_height
+        );
```
delbonis: Should convert this to a real error type/variant.

crates/csm-worker/src/processor.rs
```diff
@@ -87,6 +110,327 @@ fn process_checkpoint_log(
...
+                l2_start
+            )
+        })?;
+    let header = block.header();
+    if header.slot() != l2_start.slot() {
+        bail!(
+            "checkpoint L2 start slot mismatch (ol_block store): batch_info={} block_header={}",
+            l2_start.slot(),
+            header.slot()
+        );
```
delbonis: Should convert these.

crates/csm-worker/src/processor.rs
```diff
@@ -87,6 +110,327 @@ fn process_checkpoint_log(
...
+            )
+        })?;
+    let header = bundle.block().header().header();
+    if header.slot() != l2_start.slot() {
+        bail!(
+            "checkpoint L2 start slot mismatch (legacy l2 store): batch_info={} block_header={}",
+            l2_start.slot(),
+            header.slot()
+        );
+    }
```
delbonis: And these.

crates/csm-worker/src/processor.rs
```diff
@@ -87,6 +110,327 @@ fn process_checkpoint_log(
...
+
+    if payload_tip.l2_commitment() != expected_l2_commitment {
+        bail!(
+            "checkpoint L2 commitment mismatch: payload={} asm_log={}",
+            payload_tip.l2_commitment(),
+            expected_l2_commitment
+        );
+    }
```
delbonis: And these.

crates/csm-worker/src/processor.rs
```diff
@@ -87,6 +110,327 @@ fn process_checkpoint_log(
...
+    let post_root = pre_state
+        .compute_state_root()
+        .context("failed to compute post-state root")?;
+    if post_root != checkpoint_update.chainstate_transition().post_state_root {
+        bail!(
+            "post-state root mismatch for epoch {}: expected {}, got {}",
+            checkpoint_update.batch_info().epoch(),
+            checkpoint_update.chainstate_transition().post_state_root,
+            post_root
+        );
```
delbonis: And finally these.

crates/csm-worker/src/processor.rs
delbonis, line 1: To make this easier to test, we should abstract over the `NodeStorage` with a context trait that just exposes the database operations we need to perform.  This would also probably improve the ergonomic of a lot of the DB operations happening here.

Some of these functions *might* make sense to be converted to being member functions on `CsmWorkerState`, especially if they access the database through the context, but I'm not sure to what degree.

I also agree with Barak that this module is pretty large and does a lot of things, can we break out the DA logic into a separate module?  I also want to make sure that applying DA is something that we only do conditionally, if we're doing checkpoint sync, so that we don't make a bunch of extra database operations and compute stuff if we're also doing OL sync.

Another note here is that it seems like there's some things being implemented here that are being duplicated by the checkpoint subprotocol impl.  I had argued in that PR that we should split out the "checkpoint validation state machine" logic into its own crate that is independent of the context it's being used in so that we can use it in both places.  Do you think this makes sense to do here?

crates/ol/da/src/consumer.rs
```diff
@@ -0,0 +1,234 @@
...
+
+    /// OL DA payload decoded from the checkpoint sidecar.
+    pub da_payload: OLDaPayloadV1,
+}
+
+/// Decodes checkpoint sidecar DA payload from a raw Bitcoin checkpoint transaction.
+pub fn decode_checkpoint_da_blob(
+    raw_tx: &RawBitcoinTx,
+    magic_bytes: MagicBytes,
+) -> DaConsumerResult<DecodedCheckpointDa> {
```
delbonis: This naming is very misleading.  The checkpoint data is not part of DA, so this type/function contains/decodes a lot more distinct things than just the DA blob.

This type also makes me a little bit uncomfortable because it's sorta containing two different levels of abstraction.  The `SignedCheckpointPayload` contains the serialized `OLDaPayloadV1` as part of it.  Shouldn't we have a layer that performs the preliminary envelope validation and then exposes the semantically distinct parts as an orthogonal type?

crates/ol/da/src/consumer.rs
```diff
@@ -0,0 +1,234 @@
...
+    let payload_script = checkpoint_payload_script(tx)?;
+    let envelope_payload = parse_envelope_payload(&payload_script)
+        .map_err(|e| DaConsumerError::EnvelopeParse(e.to_string()))?;
+
+    SignedCheckpointPayload::from_ssz_bytes(&envelope_payload)
+        .map_err(|e| DaConsumerError::SignedCheckpointDecode(format!("{e:?}")))
+}
+
+/// Extracts the taproot leaf script that contains the checkpoint envelope payload.
+fn checkpoint_payload_script(tx: &Transaction) -> DaConsumerResult<ScriptBuf> {
```
delbonis: Is "checkpoint" being used as a verb in this function name?  It looks like this is an extraction function.

crates/ol/da/src/errors.rs
```diff
@@ -21,3 +21,39 @@ pub enum DaError {
...
+pub type DaConsumerResult<T> = Result<T, DaConsumerError>;
+
+#[derive(Debug, Error)]
+pub enum DaConsumerError {
+    #[error("failed to parse SPS-50 tag from checkpoint transaction: {0}")]
+    TagParse(String),
+
+    #[error(
+        "unsupported checkpoint transaction tag: expected subprotocol {expected_subprotocol} tx_type {expected_tx_type}, got subprotocol {actual_subprotocol} tx_type {actual_tx_type}"
+    )]
```
delbonis: `something happened (expected {expected}, got {got})`

Claude defaults to this colon based style and I think it's annoying and ugly.

crates/csm-worker/src/processor.rs
delbonis: It's not part of the checkpoint data, only the final state root, and I believe we decided to remove this check within the STFs anyways because it avoids having to do a bunch of hashing in the proof.

delbonis (changes requested):

bin/strata/src/services.rs
```diff
@@ -35,34 +34,26 @@ pub(crate) fn start_strata_services(nodectx: NodeContext) -> Result<RunContext>
...
```
bewakes: I don't think this will work. The way genesis is done right now, csm should be started before genesis. Because, in genesis we add the manifest from the genesis block. If we make a hard assumption that there won't be anything important in the genesis L1 block, we could get away with just adding empty manifests(and wtx root I believe). But otherwise, we need csm before genesis.

Also, genesis should have nothing to do with DA or any sort of syncing(da/block based) right?

crates/ol/da/src/consumer.rs
```diff
@@ -0,0 +1,234 @@
...
+    pub da_payload: OLDaPayloadV1,
+}
+
+/// Decodes checkpoint sidecar DA payload from a raw Bitcoin checkpoint transaction.
+pub fn decode_checkpoint_da_blob(
+    raw_tx: &RawBitcoinTx,
+    magic_bytes: MagicBytes,
+) -> DaConsumerResult<DecodedCheckpointDa> {
+    let tx = decode_bitcoin_tx(raw_tx)?;
+    verify_checkpoint_tag(&tx, magic_bytes)?;
```
bewakes: Can `strata_asm_proto_checkpoint_txs::parser::extract_signed_checkpoint_from_envelope` be used here? This seems to be a duplicated logic.

crates/csm-worker/src/processor.rs
```diff
@@ -87,6 +110,327 @@ fn process_checkpoint_log(
...
+        })?;
+    Ok((pre_state_commitment, pre_state.as_ref().clone()))
+}
+
+/// Resolves checkpoint pre-state in legacy mode with compatibility fallbacks.
+///
+/// This tries strict `l2`-derived parent resolution first. If the derived commitment cannot be
+/// mapped to an OL snapshot, it falls back to previous-checkpoint terminal state, then latest OL
+/// snapshot.
+// TODO: remove this once we delete the "old" code and functional tests
```
bewakes: If it is legacy then can't we just have the same logic as it is now(i.e. update the checkpoint info in client state, as done in the current `process_checkpoint_log`)? In the legacy code, we can just ignore the DA because we don't apply it. And for the new code, we can do the DA application on top of what's already being done.

bewakes: If it makes sense to do that then it would save us from adding redundant code that we will remove shortly.

bewakes (commented):
I talked with @irnb and understood that we don't want to put DA data in checkpoint log  so we do need to fetch checkpoint tx and extract da in the OL itself.

This PR does that and somehow updates the OLState, but that doesn't seem to be reflected in the chain worker(new) or in the `OLSyncStatusUpdate` in StatusChannel.

crates/ol/da/src/consumer.rs
```diff
@@ -0,0 +1,221 @@
...
+use strata_l1_txfmt::{MagicBytes, ParseConfig};
+use strata_ledger_types::{
+    IAccountState, IAccountStateConstructible, ISnarkAccountStateConstructible, IStateAccessor,
+};
+
+use crate::{DaConsumerError, DaConsumerResult, DaResult, OLDaPayloadV1, OLStateDiff, StateDiff};
+
+/// Decoded checkpoint DA data extracted from an L1 checkpoint transaction.
+#[derive(Debug)]
+pub struct DecodedCheckpointDa {
```
irnb: Could we update the SSZ type in `crates/checkpoint-types-ssz/ssz/payload.ssz`  so `ol_state_diff` uses `StateDiff` directly as its type, instead of `List[uint8, OL_DA_DIFF_MAX_SIZE]`?

We already have:
```
> crates/ol/da/src/types/payload.rs
pub struct StateDiff {
    /// Global state diff.
    pub global: GlobalStateDiff,

    /// Ledger state diff.
    pub ledger: LedgerDiff,
}
```

If we encode `StateDiff` as an SSZ container here, we can drop the newly introduced "DecodedCheckpointDa" ?

crates/ol/da/src/consumer.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// Signed checkpoint payload included in the transaction envelope.
+    pub signed_checkpoint: SignedCheckpointPayload,
+
+    /// OL DA payload decoded from the checkpoint sidecar.
+    pub da_payload: OLDaPayloadV1,
+}
+
+/// Decodes checkpoint sidecar DA payload from a raw Bitcoin checkpoint transaction.
+pub fn decode_checkpoint_da_blob(
```
irnb: I think we can reuse `extract_signed_checkpoint_from_envelope` from `crates/asm/txs/checkpoint/src/parser.rs` here to avoid duplicating the existing parsing logic. Would that fit this flow?

crates/ol/da/src/consumer.rs
```diff
@@ -0,0 +1,221 @@
...
+    da_diff.apply(state, &())
+}
+
+/// Converts a raw Bitcoin transaction wrapper into a decoded [`Transaction`].
+fn decode_bitcoin_tx(raw_tx: &RawBitcoinTx) -> DaConsumerResult<Transaction> {
+    Transaction::try_from(raw_tx).map_err(|e| DaConsumerError::TxDecode(e.to_string()))
+}
+
+/// Validates that the transaction carries the expected checkpoint subprotocol and tx type tags.
+fn verify_checkpoint_tag(tx: &Transaction, magic_bytes: MagicBytes) -> DaConsumerResult<()> {
```
irnb: Do we actually need this here? My understanding is the ASM layer should already handle this.

https://alpen-labs.slack.com/archives/C0A8X0K6LTF/p1770705435311119?thread_ts=1770670376.031569&cid=C0A8X0K6LTF

crates/ol/da/src/consumer.rs
storopoli: I don't know. We should probably defer this to a new ticket.
These should be `Codec` encoded to reduce the on-chain footprint to a minimum.

crates/ol/da/src/consumer.rs
```diff
@@ -0,0 +1,234 @@
...
+pub fn apply_state_diff<S>(state: &mut S, state_diff: StateDiff) -> DaResult<()>
+where
+    S: IStateAccessor,
+    S::AccountState: IAccountStateConstructible,
+    <S::AccountState as IAccountState>::SnarkAccountState: ISnarkAccountStateConstructible,
+{
+    let da_diff = OLStateDiff::<S>::from(state_diff);
+    da_diff.poll_context(state, &())?;
+    da_diff.apply(state, &())
+}
```
barakshani: these aren't used I think

barakshani (commented):
So this is truly a library now? Collection of functions for someone else to use? include fetching/decoding

crates/ol/da/src/consumer.rs
storopoli: I've simplified a LOT the code. Now the `consumer.rs` module in `ol/da/` only exports the `apply_state_diff` which does all the context polling and application:
https://github.com/alpenlabs/alpen/blob/f11e54c797d216ad199ddd39f0d3a68a39923048/crates/ol/da/src/consumer.rs#L36-L46

crates/asm/txs/admin/src/test_utils/mod.rs
```diff
@@ -111,11 +111,29 @@ pub fn create_test_admin_tx(
     create_reveal_transaction_stub(envelope_payload, tagged_payload)
 }
 
+/// Creates a test reveal transaction with the provided envelope payload and output tag script.
+///
+/// This helper is intended for cross-crate tests that need to construct a synthetic
+/// SPS-50-tagged reveal transaction without duplicating witness/script boilerplate.
+pub fn create_test_reveal_tx(envelope_payload: Vec<u8>, tag_script: ScriptBuf) -> Transaction {
```
irnb: @storopoli Minor Q: what’s the intended difference vs the existing helper here ([link](https://github.com/alpenlabs/alpen/pull/1346/changes#diff-9ea179b5340863c7abdc0c0586bd92c935197199bdc7b959b08034710f8e92bdR43))?
If the logic overlaps a lot, I’m wondering if it’s worth factoring into `crates/asm/txs/checkpoint/src/test_utils` for reuse across crates.

crates/ol/da/src/consumer.rs
```diff
@@ -0,0 +1,184 @@
...
+
+/// Applies a state diff to OL state using DA primitive semantics.
+pub fn apply_state_diff<S>(state: &mut S, state_diff: StateDiff) -> DaResult<()>
+where
+    S: IStateAccessor,
+    S::AccountState: IAccountStateConstructible,
+    <S::AccountState as IAccountState>::SnarkAccountState: ISnarkAccountStateConstructible,
+{
+    let da_diff = OLStateDiff::<S>::from(state_diff);
+    da_diff.poll_context(state, &())?;
```
bewakes: I am confused by the name of this method. It's being passed something that's not used. And when I look into the implementation, it seems to be doing some sort of validation. Not suggesting anything to be changed, but just curious if this is just meant to be a validation or something more.

bewakes (commented):
Looking at the PR, this seems to be exposing two functions basically, one to extract state diff from raw bitcoin tx and next to apply the diff to state. To me, they could just have been a module, or even be written in the same module where they will be used. 

The PR is named to be a library but if it's really required to be a library, I would expect something more from it, for example, the consumer could be a generic struct/trait that would allow fetching diffs from bitcoin client or database or somewhere else.

Maybe I am missing something but this doesn't feel like a library, instead just the minor extension of the existing two functions(namely `diff.apply(state)` and `extract_signed_checkpoint_from_envelope`). I think we could think more on this regarding how/where it will be used and design accordingly.

crates/ol/da/src/consumer.rs
```diff
@@ -0,0 +1,184 @@
...
+pub fn decode_ol_da_payload_from_checkpoint_tx(
+    raw_tx: &RawBitcoinTx,
+    magic_bytes: MagicBytes,
+) -> DaConsumerResult<DecodedCheckpointPayload> {
+    let tx = decode_bitcoin_tx(raw_tx)?;
+    let tag = parse_tx_tag(&tx, magic_bytes)?;
+    let signed_checkpoint = extract_signed_checkpoint_from_envelope(&TxInputRef::new(&tx, tag))?;
+    let da_payload = decode_buf_exact(signed_checkpoint.inner().sidecar().ol_state_diff())?;
+    Ok((signed_checkpoint, da_payload))
+}
```
delbonis: This is still kinda mixing reasoning about two levels of abstraction here, which is a red flag to me.  DA shouldn't be explicitly talked about about in a function that extracts the checkpoint payload from a checkpoint tx like this.

bewakes (approved):

crates/ol/da/src/extractor.rs
```diff
@@ -0,0 +1,110 @@
+//! # Checkpoint transaction extraction helpers for OL DA payload consumption.
+//!
+//! This is a very simple module that given any bitcoin client that implements the [`Reader`] trait,
+//! can fetch a checkpoint transaction by [`BitcoinTxid`] and decode its OL DA payload, namely
+//! [`OLDaPayloadV1`].
```
irnb, line 5: Not sure this is the best place to call it out, but it might be helpful to note that we don’t need to fetch OL logs from the checkpoint transaction for sync purpose. We only need to fetch `ol_statediff` and parse it as `OLDaPayloadV1`.

PR #1347, opened 2026-02-07
refactor(prover-perf): switch from clap to argh

krsnapaudel:
## Description

Clap to argh switch. This is what is possible now. Without spending a lot of time on work that may need to be redone.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Based on discussions with @prajwolrg, @delbonis, @bewakes and @sapinb, it makes sense to leave clap in `alpen-reth` and `alpen-ee-client`. They will end up in separate repos at some point.

- [ ]  Yes
- [x]  No

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15264,7 +15264,7 @@ name = "strata-provers-perf"
 version = "0.3.0-alpha.1"
 dependencies = [
  "anyhow",
- "clap",
+ "argh",
  "num-format",
  "reqwest 0.12.28",
  "serde_json",
--- a/bin/prover-perf/Cargo.toml
+++ b/bin/prover-perf/Cargo.toml
@@ -24,7 +24,7 @@ strata-sp1-guest-builder = { path = "../../provers/sp1", optional = true }
 zkaleido-sp1-host = { git = "https://github.com/alpenlabs/zkaleido", tag = "v0.1.0-alpha-rc18", optional = true }
 
 anyhow.workspace = true
-clap = { workspace = true, features = ["derive"] }
+argh.workspace = true
 num-format = "0.4.4"
 reqwest.workspace = true
 serde_json.workspace = true
--- a/bin/prover-perf/src/args.rs
+++ b/bin/prover-perf/src/args.rs
@@ -1,29 +1,119 @@
-use clap::Parser;
+use argh::FromArgs;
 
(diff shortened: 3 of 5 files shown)


storopoli:
> NOTE it's also not nice to take this as an order not followed.

This is not by any means "an order not followed". I just want to know what are the technical arguments on splitting the crates. Do we have a clear view that `argh` build times will not conflate with `clap` build times? And finally, why should the main binary of the "Alpen" rollup move outside of the `alpenlabs/alpen` repo? Approving this PR seems the first move in this task which I have not yet seem any formal reasoning or argumentation.

krsnapaudel:
> > NOTE it's also not nice to take this as an order not followed.
> 
> This is not by any means "an order not followed". I just want to know what are the technical arguments on splitting the > crates. Do we have a clear view that `argh` build times will not conflate with `clap` build times? And finally, why
> should the main binary of the "Alpen" rollup move outside of the `alpenlabs/alpen` repo? Approving this PR seems 
> the first move in this task which I have not yet seem any formal reasoning or argumentation.

@storopoli I have tagged them here. They should be able share their thoughts.
cc: @delbonis, @prajwolrg, @bewakes, @sapinb

krsnapaudel:
> This is not by any means "an order not followed". I just want to know what are the technical arguments on splitting the crates. Do we have a clear view that `argh` build times will not conflate with `clap` build times? And finally, why
should the main binary of the "Alpen" rollup move outside of the `alpenlabs/alpen` repo? Approving this PR seems  the first move in this task which I have not yet seem any formal reasoning or argumentation.

@storopoli I have tagged them here. They should be able to share their thoughts.
cc: @delbonis, @prajwolrg, @bewakes, @sapinb

bewakes:
@krsnapaudel @storopoli , my line of thought is that yes, if we have clap for even only one of the crates/binaries, the build time is still going to be the same. But I believe the binary sizes can still be thinner in the binaries that use argh instead of clap. So, following this line of thought, my recommendation was/is is to use argh wherever possible and retain clap where this is much complicated like alpen-client and alpen-reth.

I do not have strong opinions on whether we go all-argh, all-clap or mixed-argh-clap but I am inclined to mixed(minimal clap) for the thinner binary reasons.

PR #1333, opened 2026-02-02
STR-2121: Bug Fix Transaction mempool propagation

voidash:
## Description
For EE side when you propagate blocks there is no problem but if the there is transaction propagation then there is a bug. The bug is that bincode cannot serialize `SealedBlock` for some reason when it contains transactions. Empty blocks worked fine, but blocks with transactions caused serialization failures.
The fix converts SealedBlock to RLP bytes (Ethereum's native encoding) before bincode serialization, then decodes and re-seals on deserialization.
Also  Fullnodes can now forward transactions to the sequencer. Without this, transactions sent to fullnodes would stay in their local mempool and never get mined

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Is this PR addressing any specification, design doc or external reference document?

- [ ]  Yes
- [x]  No

If yes, please add relevant links:

## Checklist

- [ ] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [x] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [x] New and existing tests pass with my changes.
- [x] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

PR #1349, opened 2026-02-09
feat(alpen-ee): wire end-to-end DA pipeline into alpen sequencer

irnb:
## Description

Implements the complete EE data availability (DA) pipeline: sealed batches produce state-diff DA blobs, which are chunked and inscribed to Bitcoin L1 via the chunked envelope writer. Builds on top of #<STR-1368-PR-number> (chunked envelope writer).

### Key changes

* **State diff blobs**: `StateDiffBlobProvider` aggregates per-block state diffs into a `DaBlob` with `EvmHeaderDigest` metadata for chain reconstruction.
* **Cross-batch bytecode dedup**: `EeDaContext` / `EeDaContextDb` tracks previously published bytecodes. `StateDiffBlobProvider` filters them before posting, and `da_complete` marks them after DA confirmation.
* **Batch lifecycle integration**: `BatchDaProvider` returns envelope index on post, persisted in `BatchStatus::DaPending { envelope_idx }` for restart recovery. `DaBlobProvider::are_state_diffs_ready()` gates DA posting until Reth exex has written diffs.
* **Chunked envelope improvements**:

  * `CommitPublished` state prevents reveal broadcast before commit confirmation.
  * Multi-reveal waits for commit to enter mempool (101 KB descendant limit).
  * `prev_tail_wtxid` resolved at signing time instead of submission time.
* **HeaderDigestProvider trait**: Decouples DA blob construction from Reth's concrete provider types.
* **Payload serialization**: `bincode` → `serde_json` + `alloy_rlp` for EE payload encoding.

### Architecture

```text
Sealed Batch
  1. DaBlobProvider::get_blob()         (aggregate state diffs, filter published bytecodes)
  2. BatchDaProvider::post_batch_da()   (chunk blob, submit ChunkedEnvelopeEntry)
  3. Watcher task signs & broadcasts    (commit tx → reveal tx(s))
  4. BatchDaProvider::check_da_status() → DaStatus::Ready(L1DaBlockRef[])
  5. da_complete: mark bytecodes published via EeDaContext::update_da_filter()
```

---

## Notes to Reviewers

Due to the size of this PR, I've tried to keep commits atomic and well-described — each one is a self-contained logical change with a summary of what and why. Reviewing commit-by-commit should be easier than looking at the full diff.

The PR also includes functional tests (`functional-tests-new/tests/alpen_client/ee_da/`) covering four end-to-end scenarios against a Bitcoin regtest node:

1. **Normal State Diff**: ETH transfers produce account balance changes in the DA blob; validates blob decoding and wtxid chain continuity across envelopes.
2. **Empty Batch**: Blocks with no user transactions still produce a DA blob with an empty state diff, ensuring L1 chain continuity.
3. **Large State Diff (Multi-Chunk)**: 60 contract deployments exceed the single-chunk limit; validates chunked splitting, reassembly, and cross-reveal wtxid chaining.
4. **Cross-Batch Bytecode Dedup**: The same contract deployed in two consecutive batches appears only in the first batch's DA blob; the second batch omits it via EeDaContext filtering.

## Checklist

* [x] I have performed a self-review of my code.
* [x] I have commented my code where necessary.
* [ ] I have updated the documentation if needed.
* [ ] My changes do not introduce new warnings.
* [x] I have added (where necessary) tests that prove my changes are effective or that my feature works.
* [x] New and existing tests pass with my changes.
* [x] I have [[disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice)](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.


Cargo.toml
```diff
@@ -276,7 +276,8 @@ strata-codec = { git = "https://github.com/alpenlabs/strata-common", features =
   "derive",
 ], tag = "v0.1.0-alpha-rc4" }
 strata-codec-tests = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.1.0-alpha-rc4" }
-strata-l1-envelope-fmt = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.1.0-alpha-rc4" }
+# TODO: fix the following to use tag instead of rev once new release tag is created
```
irnb: Waiting for the next release tag from [alpenlabs/strata-common](https://github.com/alpenlabs/strata-common/).

prajwolrg: I updated this in https://github.com/alpenlabs/alpen/pull/1351.

bin/alpen-client/src/main.rs
delbonis, line 1: Can we structure all the things that are feature flagged with sequencer to be a little cleaner?

Similar to Jose's PR, worst case we can do

```rs
#[cfg(...)]
mod seq_imports {
  // pub use ...;
}

#[cfg(...)]
use seq_imports::*;
```

functional-tests-new/tests/alpen_client/ee_da/codec.py
delbonis, line 1: I hate to say it but should we write tests for these test utils?

bin/alpen-client/src/da_provider.rs
```diff
@@ -0,0 +1,331 @@
...
+    H: HeaderDigestProvider,
+{
+    fn get_blob(&self, batch_id: BatchId) -> eyre::Result<DaBlob> {
+        // 1. Look up the batch to get its block range. `BatchStorage` is async but
+        //    `DaBlobProvider::get_blob` is sync, so we bridge via `block_in_place` + `block_on` to
+        //    avoid deadlocking the Tokio runtime when all worker threads are occupied.
+        let (batch, _status) = block_in_place(|| {
+            Handle::current().block_on(self.batch_storage.get_batch_by_id(batch_id))
+        })?
+        .ok_or_else(|| eyre::eyre!("batch {batch_id:?} not found in storage"))?;
```
delbonis: This is kinda crazy, is there no better solution to this?  What happens if this gets called from not inside an async runtime?  Should we pass a runtime handle as part of the struct instead of using `Handle::current()`?

If it's not too much, can we have this heavy compute stuff could be moved to a normal thread not part of the async runtime?

bin/alpen-client/src/da_provider.rs
```diff
@@ -0,0 +1,331 @@
...
+            state_diff,
+        })
+    }
+
+    fn are_state_diffs_ready(&self, batch_id: BatchId) -> eyre::Result<bool> {
+        // Look up the batch to get its block range.
+        let (batch, _status) = block_in_place(|| {
+            Handle::current().block_on(self.batch_storage.get_batch_by_id(batch_id))
+        })?
+        .ok_or_else(|| eyre::eyre!("batch {batch_id:?} not found in storage"))?;
```
delbonis: Since this is duplicated, maybe we should have a helper fn that takes an `impl Future` and does this weird indirection if we can't restructure to avoid needing it.

delbonis (changes requested):
The actual impl here looks pretty good, but I'm worried about this weird juggling around IO in the async runtime.  I realize there's some constraints placed on us by Reth types, but there's *gotta* be a better solution than this wonkiness.

Also, should we attempt to use the service framework anywhere?

bin/alpen-client/src/da_provider.rs
storopoli: What if we just make the original trait `async`?
https://github.com/alpenlabs/alpen/blob/6a2e14319d078200a2b6ca51a095d789785f922a/crates/alpen-ee/common/src/traits/da.rs#L57-L76

Anyways if this is an alternative, move to a tech-debt ticket and carry on

bin/alpen-client/src/da_provider.rs
storopoli: Same as above see https://github.com/alpenlabs/alpen/pull/1349/changes#r2786760492

bin/alpen-client/src/da_provider.rs
```diff
@@ -0,0 +1,331 @@
...
+        Ok(true)
+    }
+}
+
+/// [`BatchDaProvider`] that posts DA via chunked envelope inscription.
+pub(crate) struct ChunkedEnvelopeDaProvider {
+    blob_provider: Arc<dyn DaBlobProvider>,
+    envelope_handle: Arc<ChunkedEnvelopeHandle>,
+    broadcast_ops: Arc<BroadcastDbOps>,
+    magic_bytes: [u8; 4],
```
storopoli: We have a type for that, it's `MagicBytes` and it leaves in `strata-common` under `strata-l1-txfmt`.

```suggestion
    magic_bytes: MagicBytes,
```

bin/alpen-client/src/da_provider.rs
```diff
@@ -0,0 +1,331 @@
...
+}
+
+impl ChunkedEnvelopeDaProvider {
+    pub(crate) fn new(
+        blob_provider: Arc<dyn DaBlobProvider>,
+        envelope_handle: Arc<ChunkedEnvelopeHandle>,
+        broadcast_ops: Arc<BroadcastDbOps>,
+        magic_bytes: [u8; 4],
```
storopoli: We have a type for that, it's `MagicBytes` and it leaves in `strata-common` under `strata-l1-txfmt`.

```suggestion
        magic_bytes: MagicBytes,
```

bin/alpen-client/src/da_provider.rs
```diff
@@ -0,0 +1,331 @@
...
+    ) -> eyre::Result<Vec<L1DaBlockRef>> {
+        // Only collect reveal txs — the commit tx is just a P2TR output and
+        // carries no DA witness data. The EE prover needs reveal witnesses only.
+        let mut tx_pairs: Vec<(Buf32, Buf32)> = Vec::with_capacity(entry.reveals.len());
+        for reveal in &entry.reveals {
+            tx_pairs.push((reveal.txid, reveal.wtxid));
+        }
+
+        // Group by (block_hash, block_height) -> Vec<(Txid, Wtxid)>.
+        let mut block_map: HashMap<(Buf32, u64), Vec<(Txid, Wtxid)>> = HashMap::new();
```
storopoli: We could have a type declared for this towards the top of this file:

```rs
type BlockMap = HashMap<(Buf32, u64), Vec<(Txid, Wtxid)>>
```

bin/alpen-client/src/main.rs
```diff
@@ -461,14 +570,51 @@ pub struct AdditionalConfig {
...
+    )]
     pub sequencer: bool,
 
     /// Sequencer's public key (hex-encoded, 32 bytes) for signature validation.
     #[arg(long, required = true, value_parser = parse_buf32)]
     pub sequencer_pubkey: Buf32,
+
+    // --- DA Configuration ---
+    /// Magic bytes (hex-encoded, 4 bytes) for tagging EE DA envelope transactions.
+    /// Example: `deadbeef`.
```
storopoli: Bad example is not 4 bytes

```suggestion
    /// Magic bytes (hex-encoded, 4 bytes) for tagging EE DA envelope transactions.
    /// Example: `ALPN`.
```

bin/alpen-client/src/main.rs
```diff
@@ -517,6 +663,15 @@ fn parse_buf32(s: &str) -> eyre::Result<Buf32> {
...
 
+/// Parse a hex-encoded string into 4 magic bytes.
+fn parse_magic_bytes(s: &str) -> eyre::Result<[u8; 4]> {
+    let s = s.strip_prefix("0x").unwrap_or(s);
+    let bytes = hex::decode(s).map_err(|e| eyre::eyre!("Failed to decode hex: {e}"))?;
+    bytes
+        .try_into()
+        .map_err(|_| eyre::eyre!("Magic bytes must be exactly 4 bytes"))
+}
+
```
storopoli: `MagicBytes` from `strata-l1-txfmt` (in `strata-common`) already has parsing encapsulated. Use that.

crates/alpen-ee/common/src/traits/engine/payload.rs
```diff
@@ -57,22 +59,38 @@ impl EnginePayload for AlpenBuiltPayload {
...
+            );
+            AlpenEnginePayloadError::Serialization(e)
+        })
     }
 
     fn from_bytes(bytes: &[u8]) -> Result<Self, Self::Error> {
-        let serializable = deserialize::<SerializablePayload>(bytes)?;
-        Ok(serializable.into())
+        let serializable = serde_json::from_slice::<SerializablePayload>(bytes)?;
+        serializable.try_into()
```
storopoli, line 68: Are these things touching the prover? Or just RPC stuff?
Should we really use `JSON` here?

crates/alpen-ee/common/src/types/da.rs
```diff
@@ -66,12 +66,17 @@ pub struct DaBlob {
...
 
-/// Maximum chunk payload size in bytes.
-const MAX_CHUNK_PAYLOAD: usize = 395_000;
-
 /// Serialized size of [`DaChunkHeader`] in bytes.
 const DA_CHUNK_HEADER_SIZE: usize = 37;
 
+/// Maximum size of the encoded chunk (header + payload) that the envelope
+/// builder accepts. Matches [`strata_l1_envelope_fmt::builder::MAX_ENVELOPE_PAYLOAD_SIZE`].
+const MAX_ENVELOPE_PAYLOAD: usize = 395_000;
```
storopoli, line 76: This matches `strata_l1_envelope_fmt::builder::MAX_ENVELOPE_PAYLOAD_SIZE` should have a test. I don't trust magical words in docstrings, I prefer type-checking of compilers.

crates/alpen-ee/sequencer/src/batch_lifecycle/task.rs
```diff
@@ -100,18 +100,39 @@ where
...
     };
+    use alpen_reth_db::{DbResult, EeDaContext};
     use eyre::eyre;
     use tokio::sync::watch;
 
     use super::*;
     use crate::batch_lifecycle::{state::init_lifecycle_state, test_utils::*};
 
+    /// Noop DA context for tests — reports nothing as published.
+    struct NoopDaContext;
```
storopoli, line 116: should we test-gate this?

```suggestion
    /// Noop DA context for tests — reports nothing as published.
    #[cfg(tests)]
    struct NoopDaContext;
```

crates/btcio/src/writer/chunked_envelope/handle.rs
```diff
@@ -209,60 +243,66 @@ async fn check_commit_and_broadcast_reveals(
...
-                tx_entry.status = L1TxStatus::Published;
-                bcast
-                    .put_tx_entry(txid, tx_entry)
-                    .await
-                    .map_err(|e| anyhow::anyhow!("failed to store reveal tx: {}", e))?;
-            }
+    // Deserialize all reveal transactions.
+    let mut reveal_txs = Vec::with_capacity(entry.reveals.len());
+    for reveal in &entry.reveals {
+        let tx: bitcoin::Transaction = btc_deserialize(&reveal.tx_bytes)
```
storopoli: import `Transaction` above.

crates/db/types/src/types.rs
```diff
@@ -411,16 +411,14 @@ pub struct ChunkedEnvelopeEntry {
...
-    /// Transaction IDs and reveal metadata are populated after signing.
-    pub fn new_unsigned(
-        chunk_data: Vec<Vec<u8>>,
-        magic_bytes: [u8; 4],
-        prev_tail_wtxid: Buf32,
-    ) -> Self {
+    /// Transaction IDs, reveal metadata, and `prev_tail_wtxid` are populated
+    /// at signing time by the watcher (which guarantees the predecessor entry
+    /// is already signed).
+    pub fn new_unsigned(chunk_data: Vec<Vec<u8>>, magic_bytes: [u8; 4]) -> Self {
```
storopoli: Again use the `MagicBytes`type.

bin/alpen-client/src/main.rs
irnb: [Fixed](https://github.com/alpenlabs/alpen/pull/1349/changes/8ea75e47bbe0f8979437e53b517a59a08a543d9a)

crates/alpen-ee/common/src/types/da.rs
irnb: [Fixed](https://github.com/alpenlabs/alpen/pull/1349/changes/8ea75e47bbe0f8979437e53b517a59a08a543d9a#diff-9549d2ad2d5641c41d6eb0ae7c1f5bf22ea7d07101563c664911c40145763169R370)

Thanks for calling this out.

bin/alpen-client/src/da_provider.rs
irnb: Good call , made DaBlobProvider async, which removes the runtime bridging entirely.

[Fixed](https://github.com/alpenlabs/alpen/pull/1349/changes/8ea75e47bbe0f8979437e53b517a59a08a543d9a#diff-0a0ea6bb7cc4d3527f235a4aabe4d3037c648b93c82cb40d26f06ea4fbd9b8bfR62)

bin/alpen-client/src/da_provider.rs
irnb: [Fixed](https://github.com/alpenlabs/alpen/pull/1349/changes/8ea75e47bbe0f8979437e53b517a59a08a543d9a#diff-7125ee72a009bdbe6299cd1deb78d1a718a2099675a990a78ec1199a31345445R123)

bin/alpen-client/src/da_provider.rs
irnb: [Fixed](https://github.com/alpenlabs/alpen/pull/1349/changes/8ea75e47bbe0f8979437e53b517a59a08a543d9a)

crates/alpen-ee/common/src/traits/engine/payload.rs
irnb: This is internal-only (chainstate storage for persisting built payloads between restarts),  not in the prover guest, not on-chain. 

The block is RLP encoded; the JSON wrapper just holds metadata (payload_id, fees, withdrawal_intents).

crates/alpen-ee/sequencer/src/batch_lifecycle/task.rs
irnb: Already inside #[cfg(test)] mod tests

crates/alpen-ee/common/src/traits/engine/payload.rs
storopoli: If is internal what's wrong with still using `bincode` or even migrating to `serde` (not `serde_json`) which has support for calling either machine or human-readable serialization schemes on the fly (see https://docs.rs/serde/latest/serde/trait.Serializer.html#method.is_human_readable and https://docs.rs/serde/latest/serde/trait.Deserializer.html#method.is_human_readable).

Does not make sense to force-feed JSON if we don't need that (RPC or human-readable only context).

crates/alpen-ee/sequencer/src/batch_lifecycle/task.rs
storopoli: Oh crap. my bad

crates/alpen-ee/common/src/traits/engine/payload.rs
storopoli: This is my only blocker for this PR

.github/workflows/functional-new.yml
```diff
@@ -149,9 +149,9 @@ jobs:
         # NOTE: Entire _dd ends up hundreds of MBs, so keep only logs (still huge...).
         run: |
           # service logs
-          find functional-tests-new/_dd -type f -name "service.log" | zip -@ logs.zip
+          find functional-tests-new/_dd -type f -name "service.log" 2>/dev/null | zip -@ logs.zip || true
           # functional test logs
-          zip -r logs.zip functional-tests-new/_dd -i '*/logs/*.log'
+          zip -r logs.zip functional-tests-new/_dd -i '*/logs/*.log' 2>/dev/null || true
```
storopoli, line 154: Oh that's why it was always failing in the zip thing for the functional tests every time a fn-test-new failed?

crates/alpen-ee/common/src/traits/engine/payload.rs
irnb: [Fixed](https://github.com/alpenlabs/alpen/pull/1349/commits/285a43adf514f383354fd16d5381f43c052686a6)

.github/workflows/functional-new.yml
irnb: yep.

bin/alpen-client/src/da_provider.rs
delbonis, line 1: Implementation-wise this seems okay, but I'm a bit concerned about it being in the binary crate.  Is there no way to make this more generic?

bin/alpen-client/src/main.rs
```diff
@@ -50,19 +39,41 @@ use reth_node_core::args::LogArgs;
...
 use strata_identifiers::{CredRule, OLBlockId};
+use strata_l1_txfmt::MagicBytes;
 use strata_ol_genesis::ALPEN_EE_ACCOUNT_ID_BYTES;
 use strata_primitives::buf::Buf32;
 use tokio::sync::{mpsc, watch};
 use tracing::{error, info};
-
 #[cfg(feature = "sequencer")]
-use crate::payload_builder::AlpenRethPayloadEngine;
+use {
```
delbonis, line 69: oh wait this is wicked smart

crates/alpen-ee/common/src/traits/engine/payload.rs
```diff
@@ -57,22 +59,39 @@ impl EnginePayload for AlpenBuiltPayload {
...
 #[derive(Debug, Serialize, Deserialize)]
 struct SerializablePayload {
     payload_id: PayloadId,
-    block: SealedBlock<<EthPrimitives as NodePrimitives>::Block>,
+    /// RLP-encoded sealed block
+    block_rlp: Vec<u8>,
     fees: U256,
-    requests: Option<Requests>,
+    /// Requests stored as Vec of raw bytes
+    requests: Option<Vec<Vec<u8>>>,
```
delbonis: It feels like this could benefit from a little more structure?

Also wuh, iterators?

crates/reth/db/src/lib.rs
```diff
@@ -44,16 +44,23 @@ pub trait StateDiffProvider {
...
     /// Marks the given code hashes as published. Idempotent.
     fn mark_code_hashes_published(&self, code_hashes: &[B256]) -> DbResult<()>;
+
+    /// Updates the DA filter with data from the given blocks.
+    ///
+    /// Reads state diffs for each block and records which data items have been
+    /// published to DA. Currently tracks deployed bytecodes; extensible for
+    /// address dedup and other filtering logic.
+    fn update_da_filter(&self, block_hashes: &[B256]) -> DbResult<()>;
 }
```
delbonis, line 66: With the way this works, can the filtering state be rolled back if we need to?  This seems like something we'd want to support for safety reasons.

crates/alpen-ee/common/src/traits/da.rs
```diff
@@ -44,18 +61,19 @@ pub trait BatchDaProvider: Send + Sync {
...
+    async fn get_blob(&self, batch_id: BatchId) -> eyre::Result<DaBlob>;
 
     /// Returns `true` if state diffs are ready for all blocks in the given batch.
     ///
     /// Used by the batch lifecycle to ensure state diffs have been written
     /// by the Reth exex before attempting to post DA. This prevents race
     /// conditions where DA posting is attempted before state diffs are ready.
-    fn are_state_diffs_ready(&self, batch_id: BatchId) -> eyre::Result<bool>;
+    async fn are_state_diffs_ready(&self, batch_id: BatchId) -> eyre::Result<bool>;
 }
```
delbonis, line 79: What was the reason for converting this to async?

bin/alpen-client/src/da_provider.rs
```diff
@@ -0,0 +1,358 @@
...
+
+        // 2. Aggregate per-block diffs via BatchBuilder.
+        let mut builder = BatchBuilder::new();
+        let mut block_count = 0u64;
+
+        for block_hash in batch.blocks_iter() {
+            // Convert Hash (Buf32) -> B256 for the StateDiffProvider interface.
+            let b256 = B256::from(block_hash.0);
+
+            match self.state_diff_provider.get_state_diff_by_hash(b256) {
```
delbonis: should this be converted to async?

crates/alpen-ee/common/src/traits/da.rs
storopoli: Your objection [here](https://github.com/alpenlabs/alpen/pull/1349#discussion_r2786255972) where I proposed converting to `async` in a reply [here](https://github.com/alpenlabs/alpen/pull/1349#discussion_r2786760492)

bin/alpen-client/src/da_provider.rs
storopoli: I don' think we need to convert this to `async`. You could move `get_state_diff_by_hash` to be `get_state_diff_by_hash_blocking` and make the default `get_state_diff_by_hash` async.
But probably not in the scope of this PR and I'd encourage deferring to a ticket.
We did make some traits async here mostly because we had `block_on` being called in `async` context in the `alpen-client` binary, which was very bad.

crates/alpen-ee/common/src/traits/engine/payload.rs
storopoli: This can be fixed now that https://github.com/alpenlabs/alpen/pull/1333 has been merged.

crates/reth/db/src/lib.rs
irnb: The filter is append-only right now  `mark_code_hashes_published` only adds entries, no rollback path exists. This is safe under normal conditions because duplicates on DA are harmless and `bytecode_needs_publish` conservatively returns true on any DB error, so we never silently drop data.

That said, you're right that a deeper reorg rolling back a `DaComplete` batch could leave stale entries in the filter. I've created a follow-up ticket to add unmark support and hook it into reorg handling. Keeping it out of this PR to avoid expanding scope.

https://alpenlabs.atlassian.net/browse/STR-2321

bin/alpen-client/src/da_provider.rs
irnb: only `RethHeaderSummaryProvider` is truly binary-coupled (it depends on                      `reth_provider::HeaderProvider`). 
Extracted `StateDiffBlobProvider` and `ChunkedEnvelopeDaProvider` into a new `alpen-ee-da`  crate and renamed the remaining file to header_summary.rs. Thanks for flagging this.

crates/alpen-ee/da/src/blob_provider.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        // 5. Construct the DaBlob with metadata.
+        Ok(DaBlob {
+            batch_id,
+            evm_header,
+            state_diff,
+        })
+    }
+
+    async fn are_state_diffs_ready(&self, batch_id: BatchId) -> eyre::Result<bool> {
```
evgenyzdanovich: nit.

the signature of the function (and the trait) can be improved - I don't think it requires (or shouldn't require) `eyre::result`. 
for things like is_ready?/exists?/etc, it's usually cleaner just to return bool and handle everything inside ( also handling errors and treating them as not ready).

crates/alpen-ee/da/src/blob_provider.rs
```diff
@@ -0,0 +1,173 @@
...
+    async fn get_blob(&self, batch_id: BatchId) -> eyre::Result<DaBlob> {
+        // 1. Look up the batch to get its block range.
+        let (batch, _status) = self
+            .batch_storage
+            .get_batch_by_id(batch_id)
+            .await?
+            .ok_or_else(|| eyre::eyre!("batch {batch_id:?} not found in storage"))?;
+
+        // 2. Aggregate per-block diffs via BatchBuilder.
+        let mut builder = BatchBuilder::new();
```
evgenyzdanovich, line 84: nit but important.

With how things are moving here and since we are integrating `BatchBuilder` here.. 

- PTAL at the first TODO/limitation ("No block ordering validation"). I deferred that until the integration, so If it makes sense to define such a validation - now is a good opportunity. 

- The second `TODO` from "Limitations" section at `BatchBuilder` docstrings **must** be removed. It's super misleading and directly contradicts with how it's used.

P.S. It's used correctly, the docstrings at BatchBuilder are stale out-of-date.

functional-tests-new/tests/alpen_client/ee_da/codec.py
```diff
@@ -0,0 +1,519 @@
...
+import logging
+from dataclasses import dataclass
+
+logger = logging.getLogger(__name__)
+
+# DA chunk header: version(1) + blob_hash(32) + chunk_index(2) + total_chunks(2)
+DA_CHUNK_HEADER_SIZE = 37
+
+# Minimum state_diff size for empty batch (3 varint counts for empty maps/vecs)
+EMPTY_STATE_DIFF_MAX_SIZE = 12
```
evgenyzdanovich, line 19: nit.

it's not varint, count in containers is encoded as fixed width (u32), not as a variable length integers.
https://github.com/alpenlabs/strata-common/blob/main/crates/codec/src/types.rs#L60

functional-tests-new/tests/alpen_client/ee_da/evm.py
```diff
@@ -0,0 +1,116 @@
...
+multi-slot contract deployments.
+"""
+
+# Dev account from alpen-dev-chain.json (standard Hardhat dev account)
+DEV_ACCOUNT_ADDRESS = "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
+DEV_ACCOUNT_PRIVATE_KEY = "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
+DEV_CHAIN_ID = 2892
+
+
+def send_eth_transfer(rpc, nonce: int, to_addr: str, value_wei: int) -> str:
```
evgenyzdanovich: everything in this file was present in the old functional-tests (see `functional-tests/utils`) - transfers / SC deploys / SC calls etc. I'm not sure what's the plan for the new fntests.

Ideally, I'd like to see old stuff adapted if possible, rather than new ad-hoc file being scratched and hardcoded for a single purpose that will be hard(er) to maintain.

functional-tests-new/tests/alpen_client/ee_da/test_da_pipeline.py
```diff
@@ -0,0 +1,562 @@
...
+        self._scenario_bytecode_dedup(state, l1_cursor)
+
+        self._final_validation(state)
+        return True
+
+    # -----------------------------------------------------------------
+    # Scenario 1: Normal State Diff
+    # -----------------------------------------------------------------
+
+    def _scenario_normal_state_diff(self, state: TestState, l1_cursor: int) -> int:
```
evgenyzdanovich: i'd consider it splitting into separate test cases (under broader "DA test suite" or similar) - it's quite hard to follow now and the amount if tests is only gonna grow.

with test cases being split, you won't need `println("=" * 80)` in the beginning of each tests case.

storopoli (approved):
ACK 2dba71f90db89a0195035b82b294c8d1c81681d3

crates/alpen-ee/da/src/blob_provider.rs
evgenyzdanovich: second is fixed here:
https://github.com/alpenlabs/alpen/pull/1366

feel free to consider first point later

crates/alpen-ee/da/src/blob_provider.rs
irnb: [Fixed](https://github.com/alpenlabs/alpen/pull/1349/changes/9e608ce652aa9e51cabc27c0f375d96e0c8e3325)

functional-tests-new/tests/alpen_client/ee_da/codec.py
irnb: Fixed

functional-tests-new/tests/alpen_client/ee_da/evm.py
irnb: Moved all EVM transaction helpers to `common/evm.py`. 
I wasn’t sure what the plan is for `functional-test-new`, but this should give @voidash enough visibility.

functional-tests-new/tests/alpen_client/ee_da/test_da_pipeline.py
irnb: Fixed. Split into 4 independent test files:
  - test_empty_batch.py 
  - test_normal_state_diff.py 
  - test_multi_chunk.py 
  - test_bytecode_dedup.py 
  
Thanks for calling this out.

crates/alpen-ee/da/src/blob_provider.rs
irnb: ordering is currently guaranteed structurally by all callers `StateDiffBlobProvider` iterates
  `batch.blocks_iter()` (ordered by construction) and RPC endpoints iterate by sequential block number. 
  
  Adding independent validation in the builder would require adding block identity fields to BlockStateChanges, which touches the DB schema and serialization. Deferring to a follow-up. (I’ll create a Jira ticket for it.)

storopoli (approved):
ACK 9e608ce652aa9e51cabc27c0f375d96e0c8e3325

delbonis (approved):

PR #1350, opened 2026-02-09
func-tests: sequencer producing empty blocks

storopoli:
## Description

Adds a functional test for sequencer empty-block behavior and updates `chain-worker-new` block execution to follow the OL STF layering used by block assembly.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Updated `chain-worker-new`:
- Avoids using verification-only flow directly in block assembly/execution path.
- Executes STF phases in layered order (`execute_epoch_initial_if_needed`, `execute_block_start`, tx execution, optional manifests), then validates commitments.
- Removes reliance on the previous state-root workaround wrapper approach.

`verify_block` is verification-oriented and expects a state accessor that supports `compute_state_root()`.  
`WriteTrackingState` intentionally does not provide that.  

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-2202

### DIFF
--- a/functional-tests-new/tests/strata/test_ol_sequencer_block.py
+++ b/functional-tests-new/tests/strata/test_ol_sequencer_block.py
@@ -1,0 +1,165 @@
+"""
+Tests that the sequencer produces empty blocks and checks parity with a fullnode.
+"""
+
+import contextlib
+import logging
+import tempfile
+from typing import Any
+
+import flexitest
+
+from common.base_test import StrataNodeTest
+from common.config import BitcoindConfig, ServiceType
+from common.config.params import GenesisL1View
+from common.rpc import RpcError
+from common.wait import wait_until_with_value
+from factories.strata import StrataFactory
+
+logger = logging.getLogger(__name__)
+
+BLOCKS_TO_VERIFY = 3
+FULLNODE_SYNC_SLOT_TIMEOUT = 10
+SLOT_LOOKAHEAD = 256
+
+
+@flexitest.register
+class OLSequencerBlockTest(StrataNodeTest):
+    """Sequencer produces empty blocks; fullnode parity is checked when synced."""
+
+    def __init__(self, ctx: flexitest.InitContext):
+        ctx.set_env("basic")
+
+    def main(self, ctx):
+        sequencer = self.get_service(ServiceType.Strata)
+        bitcoin = self.get_service(ServiceType.Bitcoin)
+
+        sequencer.wait_for_rpc_ready(timeout=20)
+        seq_rpc = sequencer.create_rpc()
+        btc_rpc = bitcoin.create_rpc()
+
+        bitcoind_config = BitcoindConfig(
+            rpc_url=f"http://localhost:{bitcoin.get_prop('rpc_port')}",
+            rpc_user=bitcoin.get_prop("rpc_user"),
+            rpc_password=bitcoin.get_prop("rpc_password"),
+        )
+        genesis_l1 = GenesisL1View.at_latest_block(btc_rpc)
+
+        # Use an ad-hoc fullnode with an explicit sync endpoint to the test sequencer.
+        fullnode_factory = StrataFactory(range(21543, 21643))
+        fullnode_datadir = tempfile.mkdtemp(prefix="strata_fullnode_")
+        fullnode = None
+
+        try:
+            fullnode = fullnode_factory.create_node(
+                bitcoind_config,
+                genesis_l1,
+                is_sequencer=False,
+                config_overrides={"client.sync_endpoint": sequencer.get_prop("rpc_url")},
+                datadir_override=fullnode_datadir,
+            )
+            fullnode.wait_for_ready(timeout=30)
+            fullnode_rpc = fullnode.create_rpc()
+
+            # We intentionally do not submit user txs in this test.
+            # If slots still advance, the produced blocks are empty from user-traffic perspective.
+            def get_block_at_slot(rpc, slot: int) -> dict[str, Any] | None:
+                try:
+                    blocks = rpc.strata_getRawBlocksRange(slot, slot)
+                except RpcError:
+                    return None
+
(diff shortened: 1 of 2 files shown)


functional-tests-new/tests/strata/test_ol_sequencer_block.py
barakshani, line 1: It would be great if you can add a short description what this test is doing

functional-tests-new/tests/strata/test_ol_sequencer_block.py
```diff
@@ -0,0 +1,165 @@
...
+            # Probe from genesis slot to avoid depending on chain-status RPC shape.
+            start_slot = 0
+            last_slot = start_slot + SLOT_LOOKAHEAD
+
+            def find_first_block() -> tuple[int, dict[str, Any]] | None:
+                for slot in range(start_slot, last_slot + 1):
+                    block = get_block_at_slot(seq_rpc, slot)
+                    if block is not None:
+                        return slot, block
+                return None
```
barakshani, line 86: why shouldn't this not always return 0? when has the sequencer actually started producing blocks?

functional-tests-new/tests/strata/test_ol_sequencer_block.py
```diff
@@ -0,0 +1,165 @@
...
+                lambda value: value is not None,
+                error_with="Sequencer did not produce an initial block",
+                timeout=45,
+                step=0.5,
+            )
+
+            first_slot, _ = found
+            logger.info("Found initial sequencer block at slot %s", first_slot)
+
+            # Fullnode must at least agree on the canonical genesis slot.
```
barakshani, line 99: isn't canonical genesis slot 0?

functional-tests-new/tests/strata/test_ol_sequencer_block.py
```diff
@@ -0,0 +1,165 @@
...
+            )
+            assert fullnode_genesis_block["blkid"] == seq_genesis_block["blkid"], (
+                f"Genesis block ID mismatch at slot {first_slot}"
+            )
+            assert fullnode_genesis_block["raw_block"] == seq_genesis_block["raw_block"], (
+                f"Genesis raw block mismatch at slot {first_slot}"
+            )
+
+            compared_slots = 0
+            for slot in range(first_slot + 1, first_slot + 1 + BLOCKS_TO_VERIFY):
```
barakshani, line 122: as above, I'm not sure when we started producing blocks.

functional-tests-new/tests/strata/test_ol_sequencer_block.py
```diff
@@ -0,0 +1,165 @@
...
+                        lambda blk: blk is not None,
+                        error_with=f"Waiting for fullnode to serve slot {slot}",
+                        timeout=FULLNODE_SYNC_SLOT_TIMEOUT,
+                        step=0.5,
+                    )
+                except AssertionError:
+                    logger.info(
+                        "Fullnode has not synced slot %s yet; skipping parity assertion", slot
+                    )
+                    continue
```
barakshani, line 143: I am not sure this is acceptable

functional-tests-new/tests/strata/test_ol_sequencer_block.py
```diff
@@ -0,0 +1,165 @@
...
+        btc_rpc = bitcoin.create_rpc()
+
+        bitcoind_config = BitcoindConfig(
+            rpc_url=f"http://localhost:{bitcoin.get_prop('rpc_port')}",
+            rpc_user=bitcoin.get_prop("rpc_user"),
+            rpc_password=bitcoin.get_prop("rpc_password"),
+        )
+        genesis_l1 = GenesisL1View.at_latest_block(btc_rpc)
+
+        # Use an ad-hoc fullnode with an explicit sync endpoint to the test sequencer.
```
bewakes, line 48: I don't think we need a fullnode to test the sequencer. We should be able to call the sequencer rpc endpoints to see if blocks are being produced.

We do need fullnode syncing test, but that comes later when we actually have it.

crates/chain-worker-new/src/state.rs
```diff
@@ -246,25 +252,97 @@ impl ChainWorkerServiceState {
...
+        if header.is_terminal() {
+            let expected_preseal = expectations
+                .preseal_state_root()
+                .ok_or(ExecError::ChainIntegrity)?;
+            if &pre_manifest_state_root != expected_preseal {
+                return Err(ExecError::ChainIntegrity.into());
+            }
+        } else if &pre_manifest_state_root != expectations.header_state_root() {
+            return Err(ExecError::ChainIntegrity.into());
+        }
```
bewakes, line 306: This is a lot of duplication of the stf code. As I see it, the simplest fix could be to make `WriteTrackingLayer` actually wrap base which is not only `IStateAccessor` but also `IStateBatchApplicable` so that we can  clone the base and apply the batches inside the `compute_state_root` method of the write tracking layer instead of returning unsupported error.

cc: @delbonis

functional-tests-new/tests/strata/test_ol_sequencer_block.py
```diff
@@ -0,0 +1,165 @@
...
+
+@flexitest.register
+class OLSequencerBlockTest(StrataNodeTest):
+    """Sequencer produces empty blocks; fullnode parity is checked when synced."""
+
+    def __init__(self, ctx: flexitest.InitContext):
+        ctx.set_env("basic")
+
+    def main(self, ctx):
+        sequencer = self.get_service(ServiceType.Strata)
```
bewakes, line 34: This test seems much more complicated and longer than it needs to be. To test if sequencer is producing blocks, we should just be able to query if blocks from 0 to N(~5) are being produced or not. No full nodes are required.

bewakes (changes requested):

PR #1348, opened 2026-02-08
feat(ol): parameterize OL genesis state with OLParams

prajwolrg:
## Description

- Replace hardcoded `OLState::new_genesis()` / `OLState::new_at()` with `OLState::from_genesis_params()`, enabling externally configurable OL genesis state including initial accounts, L1 block commitment, and header parameters
- Thread OLParams through the strata binary (new -`-ol-params` CLI flag pointing to a JSON file), NodeContext, genesis initialization, and the new functional test runner
- Allow path-based test selectors in the new functional test runner (e.g.,` ./run_tests.sh tests/test_foo.py`)
- Move AlpenEE `AccountId` outside of `strata-ol-genesis`. 

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [X] Refactor
- [ ] New or updated tests
- [X] Dependency Update

## Notes to Reviewers

Is this PR addressing any specification, design doc or external reference document?

- [ ]  Yes
- [X]  No

If yes, please add relevant links:

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [ ] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [x] New and existing tests pass with my changes.
- [ ] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

PR #1353, opened 2026-02-10
Remove OL epoch fields from AlpenEeParams

purusang:
## Description
Add per-account creation epoch storage and RPC in OL and refactor alpen-ee to fetch EpochCommitment in startup through the rpc.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## AI Usage
I have used AI to plan and implement. 

## Notes to Reviewers
The test failing in CI passes locally. 

The log in CI suggests that the test is trying to revert block from finalized epoch which cannot happen and fail. The following error is from [this code](https://github.com/alpenlabs/strata/blob/main/bin/strata-dbtool/src/cmd/chainstate.rs#L193) so I think the test needs to be patched.
```
Stderr: User error: Target block is inside finalized epoch: 3426f90515e97e678b62f3c6ff4759b9dc547aafcb6c504f8fd2e5cb81515956
```

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues
STR-2149

### DIFF
--- a/bin/strata/src/rpc/node.rs
+++ b/bin/strata/src/rpc/node.rs
@@ -323,11 +323,9 @@ impl OLClientRpcServer for OLRpcServer {
 
         self.storage
             .ol_checkpoint()
-            .get_epoch_commitments_at_async(epoch as u64)
+            .get_canonical_epoch_commitment_at_async(epoch as u64)
             .await
             .map_err(db_error)?
-            .first()
-            .copied()
             .ok_or_else(|| not_found_error(format!("No epoch commitment found for epoch {epoch}")))
     }
 


bin/alpen-client/src/main.rs
```diff
@@ -185,7 +179,13 @@ fn main() {
...
             let batch_da_provider = Arc::new(NoopDaProvider);
 
-            ensure_genesis(config.as_ref(), storage.as_ref())
+            // Fetch the genesis epoch commitment from the OL client once at startup.
+            let genesis_epoch = ol_client
+                .account_genesis_epoch()
+                .await
+                .context("failed to fetch account genesis epoch from OL")?;
+
+            ensure_genesis(config.as_ref(), &genesis_epoch, storage.as_ref())
```
storopoli, line 188: Can you please sync with @prajwolrg because you both have open PRs that do somewhat the same changes (and with very converging purposes). See #1348

bin/strata/src/rpc/node.rs
```diff
@@ -308,6 +308,29 @@ impl OLClientRpcServer for OLRpcServer {
...
+        self.storage
+            .checkpoint()
+            .get_epoch_commitments_at(epoch as u64)
+            .await
+            .map_err(db_error)?
+            .first()
+            .copied()
+            .ok_or_else(|| not_found_error(format!("No epoch commitment found for epoch {epoch}")))
+    }
+
```
storopoli, line 331: Are you sure that `first()` will get you the precise `EpochCommitment` you want?
I'm not convinced.

bewakes: @purusang I suggest you use `get_canonical_epoch_commmitment_at` method in the checkpoint manager.
And also, it should be `storage.ol_checkpoint()` instead of `storage.checkpoint()`.

bewakes: The suggested method uses `.first()` internally because we don't yet have a mechanism to determine what's the expected commitment.

bewakes (changes requested):
Looks good overall except the wrong reference to `storage.checkpoint()` which should be `storage.ol_checkpoint()`.

bin/alpen-client/src/main.rs
purusang: @storopoli here is an AI generated summary of what overlaps logically
This pr:                                                                                                       
- Strata binary (OL side): Stores genesis account creation epochs in a new account_genesis DB during init_ol_genesis, and exposes get_account_genesis_epoch_commitment via RPC                                                                                           
- Alpen-client (EE side): Fetches the genesis epoch from OL via ol_client.account_genesis_epoch() at startup and threads it through ensure_genesis → ensure_genesis_ee_account_state → store_ee_account_state, replacing the previously hardcoded build_ee_genesis_ol_epoch

 PR #1348 does something converging:
- Makes genesis accounts configurable via OLParams (no more hardcoded ALPEN_EE_ACCOUNT_ID_BYTES)
- Changes init_ol_genesis to take &OLParams instead of &Params
- The genesis account ID (which your PR needs to track creation epochs for) is now determined by OLParams, not hardcoded

Since @prajwol is on leave this week so I do not know if its possible to discuss this. What I think is either pr can go first and resolve conflicts while also keeping the logical changes made in both prs. Main thing is the change that we both made can be easily adjusted in both sides. Specially regarding Params changed to OLParams in his pr and also using account from OLParams vs using static.

purusang: But one thing we should decide is how we want to register new accounts. Params don't seem scalable.

bin/strata/src/rpc/node.rs
bewakes: @purusang please also use `get_canonical_epoch_commitment_at`. Using that, later when we have a mechanism to find the canonical commitment, we only have to change in one place.

bin/alpen-client/src/main.rs
storopoli: > Since @Prajwol is on leave this week so I do not know if its possible to discuss this. What I think is either pr can go first and resolve conflicts while also keeping the logical changes made in both prs. Main thing is the change that we both made can be easily adjusted in both sides. Specially regarding Params changed to OLParams in his pr and also using account from OLParams vs using static.

Then let's merge this first. I was more worried about nasty rebases. But since @prajwolrg is on leave, unfortunately he'll have to deal with the rebase.

storopoli (approved):
ACK 4dbb22cbfa909c02e3db57752bc5a88e4a4f21fc

again "old" code that should not be there almost flew under the radar on a code review 😓

PR #1354, opened 2026-02-10
M2 bug fixes

storopoli:
## Description

This PR fixes 2 sequencer bugs in the new OL path (`bin/strata` + `crates/ol/sequencer`):

1. Fixes an epoch-0 underflow risk in unsigned checkpoint scanning (edff937cb66d733df827c88ee26d358cc4412c2e)
2. Persists OL checkpoint status to `Signed(intent_idx)` after successful intent enqueue (both sequencer RPC and internal duty executor paths) (d7591cd3d16b198a7fd2c4b445704eb7593bd9ad)

Additionally, this introduces `EnvelopeHandle::submit_intent_async_with_idx(...)` so callers can persist the actual intent index returned by writer storage.

AI Assistance Notice: This PR was prepared with AI assistance (OpenAI Codex CLI), and reviewed/validated by the author.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- Signed status persistence now happens immediately after successful enqueue to envelope writer, using returned `intent_idx`.
- Existing `submit_intent_async(...)` behavior is preserved; new indexed variant is additive.

- [ ]  Yes
- [x]  No

## Checklist

- [x] I have disclosed my use of AI (https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

## Related Issues

N/A

### DIFF
--- a/bin/strata/src/sequencer/duty_fetcher.rs
+++ b/bin/strata/src/sequencer/duty_fetcher.rs
@@ -6,7 +6,7 @@ use strata_ol_sequencer::{Duty, TemplateManager, extract_duties};
 use strata_status::StatusChannel;
 use strata_storage::NodeStorage;
 use tokio::{sync::mpsc, time::interval};
-use tracing::{error, warn};
+use tracing::{debug, error, warn};
 
 /// Worker for fetching duties from the sequencer.
 #[tracing::instrument(skip_all, fields(component = "sequencer_duty_fetcher"))]
@@ -44,11 +44,14 @@ pub(crate) async fn duty_fetcher_worker(
                 }
             };
 
-        // Log non-empty duties
         if duties.is_empty() {
-            warn!(count = %duties.len(), "got no new duties, skipping");
+            debug!(count = %duties.len(), "got no new duties, skipping");
             continue;
         }
+
+        // Log non-empty duties
+        let duties_display: Vec<String> = duties.iter().map(ToString::to_string).collect();
+        debug!(duties = ?duties_display, "got some sequencer duties");
 
         for duty in duties {
             if duty_tx.send(duty).await.is_err() {
--- a/crates/ol/sequencer/src/duty.rs
+++ b/crates/ol/sequencer/src/duty.rs
@@ -3,6 +3,8 @@
 //! Key improvement: Templates are generated and embedded directly in duties,
 //! eliminating the need for separate template fetch requests.
 
+use std::fmt;
+
 use ssz::Encode;
 use strata_checkpoint_types_ssz::CheckpointPayload;
 use strata_crypto::hash;
@@ -56,6 +58,32 @@ impl Duty {
             Self::SignCheckpoint(c) => {
                 let encoded = c.checkpoint.as_ssz_bytes();
                 hash::raw(&encoded)
+            }
+        }
+    }
+}
+
+impl fmt::Display for Duty {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::SignBlock(duty) => {
+                write!(
+                    f,
+                    "SignBlock(slot: {}, epoch: {}, ts: {}, ready: {})",
+                    duty.slot(),
+                    duty.template.epoch(),
+                    duty.target_timestamp(),
+                    if duty.is_ready() { "yes" } else { "no" }
+                )
+            }
+            Self::SignCheckpoint(duty) => {
+                write!(
+                    f,
+                    "SignCheckpoint(epoch: {}, l1_height: {}, l2_slot: {})",
+                    duty.epoch(),
+                    duty.checkpoint.new_tip().l1_height,
+                    duty.checkpoint.new_tip().l2_commitment.slot
+                )
             }
         }
     }


krsnapaudel (commented):
Does it make sense to do this:
Modify `put_intent_entry` in the `L1WriterDatabase` trait to return the assigned index?

With Claude Code help:
The "snapshot `next_idx` before insert, snapshot after, scan the range" pattern to discover the assigned index is fragile and indirect. The underlying `put_intent_entry` in sled already knows the exact index it assigned (via `find_next_available_id` inside a transaction).

crates/btcio/src/writer/task.rs
```diff
@@ -78,32 +79,61 @@ impl EnvelopeHandle {
...
         }
 
         debug!(commitment = %id, "Received intent for processing");
 
         // Check if it is duplicate
         if self.ops.get_intent_by_id_async(id).await?.is_some() {
             warn!(commitment = %id, "Received duplicate intent");
-            return Ok(());
+            let next_idx = self.ops.get_next_intent_idx_async().await?;
+            return self.find_intent_idx_in_range(id, 0, next_idx).await;
```
krsnapaudel, line 105: Can we do this? And get rid of `find_intent_idx_in_range`?
```
if self.ops.get_intent_by_id_async(id).await?.is_some() {
    warn!(commitment = %id, "Received duplicate intent");
    return Ok(None);
}
```

crates/btcio/src/writer/task.rs
krsnapaudel: This is ignoring the index returned.

bin/strata/src/sequencer/rpc.rs
```diff
@@ -112,10 +113,17 @@ impl OLSequencerRpcServer for OLSeqRpcServer {
...
-                .map_err(|e| internal_error(e.to_string()))?;
+                .map_err(|e| internal_error(e.to_string()))?
+                .ok_or_else(|| internal_error("failed to resolve checkpoint intent index"))?;
+
+            entry.status = OLCheckpointStatus::Signed(intent_idx);
+            db.put_checkpoint_async(epoch, entry)
+                .await
+                .map_err(db_error)?;
         }
         // If already signed, then fine, return
```
krsnapaudel: Should we log this?

krsnapaudel (commented):
Some more comments:
1. `find_intent_idx_in_range` Can we get rid of this?
2. In `bundler.rs` `ops.put_intent_entry_async(*entry.intent.commitment(), entry)
    .await?` is ignoring the returned index.
3. `l1_writer_tests.rs` could be updated to assert the returned index.

crates/btcio/src/writer/task.rs
bewakes: It would be nice, but we can't quite do that easily because in the duty executor, it is expected an idx from the function call and if it returns None, it will inadvertently error out. Let's keep this as is.

bin/strata/src/sequencer/rpc.rs
bewakes: Done in https://github.com/alpenlabs/alpen/pull/1354/commits/547277c147ea7a6105f53daa55144377c064c205

crates/btcio/src/writer/task.rs
bewakes: Done in https://github.com/alpenlabs/alpen/pull/1354/commits/547277c147ea7a6105f53daa55144377c064c205

bin/strata/src/sequencer/duty_executor.rs
```diff
@@ -165,14 +183,22 @@ async fn handle_sign_checkpoint_duty(
...
+        .await
+        .map_err(|e| anyhow!("failed to submit checkpoint intent: {e}"))?
+        .ok_or_else(|| anyhow!("failed to resolve checkpoint intent index for epoch {epoch}"))?;
+
+    entry.status = OLCheckpointStatus::Signed(intent_idx);
+    checkpoint_db
+        .put_checkpoint_async(epoch, entry)
         .await
-        .map_err(|e| anyhow!("failed to submit checkpoint intent: {e}"))?;
+        .map_err(|e| anyhow!("failed persisting signed checkpoint status: {e}"))?;
```
krsnapaudel, line 196: "failed to persist"

bin/strata/src/sequencer/duty_fetcher.rs
```diff
@@ -45,7 +45,7 @@ pub(crate) async fn duty_fetcher_worker(
             };
 
         // Log non-empty duties
-        if !duties.is_empty() {
+        if duties.is_empty() {
```
krsnapaudel, line 47: The idea is to log only when duties are there. And continue when they are not there.

bin/strata/src/sequencer/duty_executor.rs
bewakes: I wouldn't make assumption on what the envelope handle does here, from the perspective of executor, all we are doing is submitting to the envelope handle and leave persisting and anything else to the handler. This seems to be fine as is.

bin/strata/src/sequencer/duty_fetcher.rs
bewakes: Check https://github.com/alpenlabs/alpen/pull/1354/commits/93b5314da2710235407dd7aa3434a3d148399f2d

bin/strata/src/sequencer/duty_executor.rs
krsnapaudel: This is about language.

bin/strata/src/sequencer/duty_fetcher.rs
krsnapaudel: We still have a debug log when duties are empty. My suggestion was to log only when we have duties.

bewakes: Yeah, I think its fine to log when there are no duties as well.

PR #1355, opened 2026-02-11
feat(asm-admin): relax admin seqno to monotonically increasing

prajwolrg:
## Description

- **Relax admin sequence number validation** from strictly sequential (`seqno == expected`) to monotonically increasing (`seqno > last_seqno`). This allows admin transactions to be submitted with any sequence number greater than the last executed one, removing the requirement that they arrive in exact +1 order.
- **Add `seqno` to the on-chain `SignedPayload`**, so the sequence number is committed in the envelope rather than being implicitly derived from authority state. The parser now returns the full `SignedPayload` instead of a destructured `(MultisigAction, SignatureSet)` tuple.
- **Move `confirmation_depth` from `AdministrationSubprotoParams` into `AdministrationSubprotoState`**. 

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/asm/spec/src/lib.rs
+++ b/crates/asm/spec/src/lib.rs
@@ -105,8 +105,8 @@ impl StrataAsmSpec {
         let admin_params = AdministrationSubprotoParams::new(
             admin_config.clone(),
             admin_config,
-            2,  // confirmation_depth: updates activate 2 blocks after submission
-            10, // max_seqno_gap: maximum allowed gap between consecutive sequence numbers
+            2, // confirmation_depth: updates activate 2 blocks after submission
+            NonZero::new(10).unwrap(),
         );
 
         Self {
--- a/crates/asm/subprotocols/admin/src/authority.rs
+++ b/crates/asm/subprotocols/admin/src/authority.rs
@@ -1,3 +1,5 @@
+use std::num::NonZero;
+
 use borsh::{BorshDeserialize, BorshSerialize};
 use strata_asm_params::Role;
 use strata_asm_txs_admin::parser::SignedPayload;
@@ -64,7 +66,7 @@ impl MultisigAuthority {
     pub fn verify_action_signature(
         &self,
         payload: &SignedPayload,
-        max_seqno_gap: u8,
+        max_seqno_gap: NonZero<u8>,
     ) -> Result<SeqNoToken, AdministrationError> {
         if payload.seqno <= self.last_seqno {
             return Err(AdministrationError::InvalidSeqno {
@@ -74,7 +76,7 @@ impl MultisigAuthority {
             });
         }
 
-        if payload.seqno > self.last_seqno + max_seqno_gap as u64 {
+        if payload.seqno > self.last_seqno + max_seqno_gap.get() as u64 {
             return Err(AdministrationError::SeqnoGapTooLarge {
                 role: self.role,
                 payload_seqno: payload.seqno,
--- a/crates/asm/subprotocols/admin/src/error.rs
+++ b/crates/asm/subprotocols/admin/src/error.rs
@@ -1,3 +1,5 @@
+use std::num::NonZero;
+
 use strata_asm_params::Role;
 use strata_asm_txs_admin::actions::UpdateId;
 use strata_crypto::threshold_signature::ThresholdSignatureError;
@@ -34,7 +36,7 @@ pub enum AdministrationError {
         role: Role,
         payload_seqno: u64,
         last_seqno: u64,
-        max_gap: u8,
+        max_gap: NonZero<u8>,
     },
 
     /// Indicates a threshold signature error (configuration or signature validation).
--- a/crates/asm/subprotocols/admin/src/handler.rs
+++ b/crates/asm/subprotocols/admin/src/handler.rs
@@ -1,3 +1,5 @@
+use std::num::NonZero;
+
 use strata_asm_checkpoint_msgs::CheckpointIncomingMsg;
 use strata_asm_common::{
     MsgRelayer,
@@ -102,7 +104,7 @@ pub(crate) fn handle_action(
     payload: SignedPayload,
     current_height: u64,
     relayer: &mut impl MsgRelayer,
-    max_seqno_gap: u8,
+    max_seqno_gap: NonZero<u8>,
 ) -> Result<(), AdministrationError> {
     // Determine the required role based on the action type
     let role = match &payload.action {
@@ -266,7 +268,7 @@ mod tests {
             strata_administrator,
             strata_sequencer_manager,
             confirmation_depth: 2016,
-            max_seqno_gap: 10,
+            max_seqno_gap: 10.try_into().unwrap(),
         };
 
         (config, strata_admin_sks, strata_seq_manager_sks)
@@ -742,7 +744,7 @@ mod tests {
         .unwrap();
 
         // Second action at seqno 11 (last_seqno is 1, gap = 10 = max_seqno_gap)
-        let gap_seqno = 1 + params.max_seqno_gap as u64;
+        let gap_seqno = 1 + params.max_seqno_gap.get() as u64;
         let action = MultisigAction::Update(updates[1].clone());
         let sighash = action.compute_sighash(gap_seqno);
         let sig_set = create_signature_set(&admin_sks, &signer_indices, sighash);
@@ -773,7 +775,7 @@ mod tests {
         let update = get_strata_administrator_update_actions(1)[0].clone();
 
         // Try action at seqno 11 (last_seqno is 0, gap = 11 > max_seqno_gap of 10)
-        let too_far_seqno = params.max_seqno_gap as u64 + 1;
+        let too_far_seqno = params.max_seqno_gap.get() as u64 + 1;
         let action = MultisigAction::Update(update);
         let sighash = action.compute_sighash(too_far_seqno);
         let sig_set = create_signature_set(&admin_sks, &signer_indices, sighash);
@@ -793,8 +795,8 @@ mod tests {
                 role: Role::StrataAdministrator,
                 payload_seqno: 11,
                 last_seqno: 0,
-                max_gap: 10,
-            })
+                max_gap,
+            }) if max_gap.get() == 10
         ));
     }
 }
--- a/crates/asm/subprotocols/admin/src/state.rs
+++ b/crates/asm/subprotocols/admin/src/state.rs
@@ -161,7 +161,7 @@ mod tests {
             strata_administrator,
             strata_sequencer_manager,
             confirmation_depth: 2016,
-            max_seqno_gap: 10,
+            max_seqno_gap: NonZero::new(10).unwrap(),
         }
     }
 
--- a/crates/asm/params/src/subprotocols/admin.rs
+++ b/crates/asm/params/src/subprotocols/admin.rs
@@ -1,3 +1,5 @@
+use std::num::NonZero;
+
 #[cfg(feature = "arbitrary")]
 use arbitrary::Arbitrary;
 use borsh::{BorshDeserialize, BorshSerialize};
@@ -29,7 +31,7 @@ pub struct AdministrationSubprotoParams {
     ///
     /// A payload with `seqno > last_seqno + max_seqno_gap` is rejected. This prevents
     /// excessively large jumps in sequence numbers while still allowing non-sequential usage.
-    pub max_seqno_gap: u8,
+    pub max_seqno_gap: NonZero<u8>,
 }
 
 /// Roles with authority in the administration subprotocol.
@@ -70,7 +72,7 @@ impl AdministrationSubprotoParams {
         strata_administrator: ThresholdConfig,
         strata_sequencer_manager: ThresholdConfig,
         confirmation_depth: u16,
-        max_seqno_gap: u8,
+        max_seqno_gap: NonZero<u8>,
     ) -> Self {
         Self {
             strata_administrator,


crates/asm/subprotocols/admin/src/authority.rs
```diff
@@ -49,23 +53,33 @@ impl MultisigAuthority {
     // could be added in the future if multiple signature schemes are needed.
     pub fn verify_action_signature(
         &self,
-        action: &MultisigAction,
-        signatures: &SignatureSet,
-    ) -> Result<(), ThresholdSignatureError> {
+        payload: &SignedPayload,
+    ) -> Result<(), AdministrationError> {
+        if payload.seqno <= self.last_seqno {
```
barakshani, line 71: Didn't we have any check here (payload.seqno = self.last_seqno + 1) before?

I will add some ceiling as well, say 10 or 100, so we won't get into a situation that the sequence number jumps too high.

prajwolrg: No check was needed previously because we didn't pass the `seqno` as part of the payload. We used the `seqno` that was stored in the state and incremented it after each use.

barakshani: Does it mean that earlier we did not defend against replays?

crates/asm/subprotocols/admin/src/handler.rs
```diff
@@ -352,67 +352,47 @@ mod tests {
...
-        );
+        let payload = SignedPayload::new(valid_seqno, action, sig_set);
+        let res = handle_action(&mut state, payload, current_height, &mut relayer);
         assert!(res.is_ok());
 
-        // Try queuing it again with same seq no
+        // Authority seqno is now 1. Try replaying with seqno 1 (<= current).
         let action = MultisigAction::Update(update.clone());
-        let sighash = action.compute_sighash(initial_seq_no);
+        let sighash = action.compute_sighash(1);
```
barakshani, line 393:
```suggestion
        let sighash = action.compute_sighash(valid_seqno);

File: crates/asm/subprotocols/admin/src/handler.rs [hunk 3]
```diff
@@ -644,44 +609,32 @@ mod tests {
...
-        .unwrap();
-
-        // Cancel the update action
+        let payload = SignedPayload::new(update_seqno, update_action, sig_set);
+        handle_action(&mut state, payload, current_height, &mut relayer).unwrap();
+
+        // Cancel the update action (authority seqno is now 1, use seqno 2)
         let cancel_action = MultisigAction::Cancel(CancelAction::new(update_id));
-        let sighash = cancel_action.compute_sighash(initial_seq_no + 1);
+        let cancel_seqno = last_seqno + 2;
```
barakshani, line 688: does cancel update need seqno? if the cancel refers to a particular sequence/update number, there's no risk of reply because you can't cancel twice the same update.

barakshani: I have no problem with all updates (including cancel) having seqno, but comparing to the previous code we had, i'm a little bit confused about how it worked earlier.

crates/asm/subprotocols/admin/src/error.rs
```diff
@@ -13,6 +13,12 @@ pub enum AdministrationError {
     #[error("no pending update found for action_id = {0:?}")]
     UnknownAction(UpdateId),
 
+    /// The payload's sequence number is not greater than the last executed sequence number.
+    #[error(
+        "invalid seqno: payload seqno {payload_seqno} must be greater than last seqno {last_seqno}"
+    )]
+    InvalidSeqno { payload_seqno: u64, last_seqno: u64 },
```
irnb:
```suggestion
    InvalidSeqno {
        role: Role,
        payload_seqno: u64,
        last_seqno: u64,
    }
```

For debugging multi-authority scenarios, it might help to include which authority/role rejected.

crates/asm/subprotocols/admin/src/authority.rs
```diff
@@ -49,23 +53,33 @@ impl MultisigAuthority {
...
 
-        verify_threshold_signatures(&self.config, signatures.signatures(), &sig_hash.into())
+        Ok(())
     }
 
-    /// Increments the seqno.
-    pub fn increment_seqno(&mut self) {
-        self.seqno += 1;
+    /// Updates the last executed seqno.
+    pub(crate) fn update_last_seqno(&mut self, seqno: u64) {
```
irnb: Nit (non-blocking): Today we rely on the caller to (1) call `verify_action_signature` before `update_last_seqno`, and (2) pass the matching `seqno`. Since this is replay-critical state, we could make that ordering/coupling compiler enforced by returning a “verified” token from verification and requiring it for the update:

```rust
pub(crate) struct VerifiedPayload { seqno: u64 }

pub fn verify_action_signature(
    &self,
    payload: &SignedPayload,
) -> Result<VerifiedPayload, AdministrationError> {
    // ... checks ...
    Ok(VerifiedPayload { seqno: payload.seqno })
}

pub(crate) fn update_last_seqno(&mut self, verified: VerifiedPayload) {
    self.last_seqno = verified.seqno;
}
```

Not urgent (both are `pub(crate)` and there’s a single caller today), but it reduces footguns and documents the intended flow in the API.

irnb (commented):
LGTM, just a few nits.

crates/asm/subprotocols/admin/src/authority.rs
delbonis: I think it's fine, it's just not possible to tell which seqno a message was for in isolation without exhaustively checking and concluding that it was invalid if none match.

crates/asm/subprotocols/admin/src/state.rs
```diff
@@ -26,6 +26,11 @@ pub struct AdministrationSubprotoState {
 
     /// UpdateId for the next update.
     next_update_id: UpdateId,
+
+    /// The confirmation depth (CD) setting, in Bitcoin blocks: after an update transaction
+    /// receives this many confirmations, the update is enacted automatically. During this
+    /// confirmation period, the update can still be cancelled by submitting a cancel transaction.
+    confirmation_depth: u16,
```
delbonis, line 32: I think "confirmation" is a misleading name here, since it makes it seem like bitcoin confirmations.  Really, this should be like "enactment grace period" or something.

crates/asm/txs/admin/src/actions/mod.rs
```diff
@@ -66,4 +66,13 @@ impl MultisigAction {
...
         }
     }
+
+    /// Constructs the SPS-50 [`TagData`] for this action.
+    ///
+    /// The tag is built from the administration subprotocol ID and the
+    /// action's [`TxType`], with no auxiliary data.
+    pub fn tag(&self) -> TagData {
+        TagData::new(ADMINISTRATION_SUBPROTOCOL_ID, self.tx_type(), vec![])
+            .expect("empty aux data always fits")
```
delbonis, line 79: Maybe we should have a `new_noaux` constructor to simplify this?

crates/asm/subprotocols/admin/src/authority.rs
prajwolrg: > Does it mean that earlier we did not defend against replays?

We did. We **had** to use the `seqno` that was in the state. If we used any other `seqno`, it would fail. The `seqno` was incremented after each admin operation.

crates/asm/subprotocols/admin/src/handler.rs
prajwolrg: > does cancel update need seqno? if the cancel refers to a particular sequence/update number, there's no risk of reply because you can't cancel twice the same update.

Right now we require a seqno for each `AdminTx` that goes onchain. But as you pointed out, we don't necessarily need it. 

> i'm a little bit confused about how it worked earlier.

In the earlier version as well, we incremented the seqno after each `AdminTx`.

crates/asm/subprotocols/admin/src/state.rs
prajwolrg: `confirmation_depth` is taken based from [here](https://www.notion.so/Time-related-settings-200901ba000f80159150c55035207ff7). I understand the name might be misleading, but I don't want to deviate from the doc yet. Maybe we can start a discussion there.

crates/asm/txs/admin/src/actions/mod.rs
prajwolrg: Created https://github.com/alpenlabs/strata-common/pull/75.

crates/asm/subprotocols/admin/src/authority.rs
prajwolrg: Thanks. This is a neat idea.

prajwolrg: Done in [bd2032a](https://github.com/alpenlabs/alpen/pull/1355/commits/bd2032a8eae7b322b2dad7b19875795683522870).

crates/asm/subprotocols/admin/src/error.rs
prajwolrg: Done in [d1d17fb](https://github.com/alpenlabs/alpen/pull/1355/commits/d1d17fb1a052a22aa4eabf1594c4d8cc0624dd83).

crates/asm/spec/src/lib.rs
```diff
@@ -104,7 +104,8 @@ impl StrataAsmSpec {
         let admin_params = AdministrationSubprotoParams::new(
             admin_config.clone(),
             admin_config,
-            2, // confirmation_depth: updates activate 2 blocks after submission
+            2,  // confirmation_depth: updates activate 2 blocks after submission
```
barakshani: you don't want to export it some general params file?

crates/asm/subprotocols/admin/src/config.rs
```diff
@@ -21,18 +21,26 @@ pub struct AdministrationSubprotoParams {
     /// receives this many confirmations, the update is enacted automatically. During this
     /// confirmation period, the update can still be cancelled by submitting a cancel transaction.
     pub confirmation_depth: u16,
+
+    /// Maximum allowed gap between consecutive sequence numbers for a given authority.
+    ///
+    /// A payload with `seqno > last_seqno + max_seqno_gap` is rejected. This prevents
+    /// excessively large jumps in sequence numbers while still allowing non-sequential usage.
+    pub max_seqno_gap: u64,
```
barakshani: why u64?

crates/asm/subprotocols/admin/src/state.rs
```diff
@@ -165,6 +165,7 @@ mod tests {
             strata_administrator,
             strata_sequencer_manager,
             confirmation_depth: 2016,
+            max_seqno_gap: 10,
```
barakshani: read from param file?

crates/asm/spec/src/lib.rs
prajwolrg: It is part of the config. We are using **legacy** `RollupParams` to construct the `AdminConfig`, so it's hardcoded, but this will be removed.

crates/asm/subprotocols/admin/src/config.rs
prajwolrg: Nice catch. This feel redundant.  Going with `u8` in [022f168](https://github.com/alpenlabs/alpen/pull/1355/commits/022f1680b3be59b700cea1d006a37927d2b66549).

crates/asm/subprotocols/admin/src/state.rs
prajwolrg: This line is modifying the Params file by adding `max_seqno_gap`.

crates/asm/subprotocols/admin/src/authority.rs
```diff
@@ -49,23 +63,47 @@ impl MultisigAuthority {
...
+    ) -> Result<SeqNoToken, AdministrationError> {
+        if payload.seqno <= self.last_seqno {
+            return Err(AdministrationError::InvalidSeqno {
+                role: self.role,
+                payload_seqno: payload.seqno,
+                last_seqno: self.last_seqno,
+            });
+        }
+
+        if payload.seqno > self.last_seqno + max_seqno_gap as u64 {
```
storopoli: The logic requires `payload.seqno > last_seqno` and also `payload.seqno <= last_seqno + max_seqno_gap`. If `max_seqno_gap` is `0`, no value can satisfy both constraints, so governance actions become permanently unprocessable.

Make `max_seqno_gap` a [`NonZero` using `std::num`](https://doc.rust-lang.org/beta/std/num/struct.NonZero.html)

storopoli: This is REALLY REALLY improbable but possible:
`self.last_seqno + max_seqno_gap` as `u64` can overflow near `u64::MAX` (panic in checked builds / wrap in unchecked builds).
Suggested fix: use `saturating_add` or `checked_add` with explicit handling.

barakshani: this is practically impossible, that I'm happy to say that it's just impossible: we are going to set max_seqno_gap = 10, but even with 1 million it would still be impossible. You can probably count how many bitcoin blocks it would take, and you'll end up with a good number of _centuries_.

storopoli: Yeah we can solve this.
Might come up in an upcoming audit.

prajwolrg: Wow! nice catch! Fixed in [30248bb](https://github.com/alpenlabs/alpen/pull/1355/commits/30248bb141ebf1c05749275661564479970da4d3).

storopoli (approved):
ACK c16e9a3b8b0dc6703391587d5a82e7127d0ebff3

PR #1356, opened 2026-02-11
STR-2218: WriteTrackingState returns Err(AcctError::Unsupported) for compute_state_root()

krsnapaudel:
## Description

`chain-worker-new` verifies blocks through an `IndexerState<WriteTrackingState<&OLState>>` stack, but `WriteTrackingState::compute_state_root()` returns `Err(AcctError::Unsupported)` because it holds an immutable reference plus a write batch — the batch hasn't been applied to the base state.

### Commit 1: `verify_block_with_root_fn` + `compute_state_root_materialized`
- Adds `verify_block_with_root_fn` that accepts a caller-provided closure for root computation; `verify_block` becomes a thin wrapper
- Adds `WriteTrackingState::compute_state_root_materialized()` helper that clones the base, applies the batch, and computes the root
- `chain-worker-new` uses `verify_block_with_root_fn` with the materialized root closure

### Commit 2: Tight trait bounds on `WriteTrackingState`
- Tightens `IStateAccessor` impl bounds on `WriteTrackingState` to require `S: Clone + IStateBatchApplicable`
- Implements `compute_state_root()` directly in the trait impl by materializing state (clone base, apply batch, compute root)
- Removes `verify_block_with_root_fn` and `compute_state_root_materialized` from commit 1
- `verify_block` calls `state.compute_state_root()` directly — single API path, no closure
- `BatchDiffState` gets `Clone` + `IStateBatchApplicable` (returning `Unsupported`) to satisfy tighter bounds
- `block_assembly.rs` functions use `BlockAssemblyStateAccess` trait instead of manually spelled-out bounds

### Approach 1 vs Approach 2

| | Approach 1 (closure) | Approach 2 (tight bounds) |
|---|---|---|
| **API surface** | Adds parallel `verify_block_with_root_fn` | Single `verify_block` — no new API |
| **Design clarity** | Caller decides how to compute root | State type owns its own root computation |
| **Caller complexity** | chain-worker-new must pass a closure | chain-worker-new just calls `verify_block` |
| **Bound changes** | None — existing consumers unaffected | Tightened bounds on `WriteTrackingState` affect all `IStateAccessor` methods |
| **Blast radius** | Minimal — only adds new functions | Wider — touches `BatchDiffState` and `block_assembly.rs` bounds |
| **Composability** | `WriteTrackingState` wraps any `IStateAccessor` | Base must be `Clone + IStateBatchApplicable` |

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Bibek and I are leaning towards approach 2. Squashing two commits will also give us that.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-2218](https://alpenlabs.atlassian.net/browse/STR-2218)

[STR-2218]: https://alpenlabs.atlassian.net/browse/STR-2218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/chain-worker-new/src/state.rs
+++ b/crates/chain-worker-new/src/state.rs
@@ -14,7 +14,7 @@ use strata_identifiers::OLBlockCommitment;
 use strata_ol_chain_types_new::{OLBlock, OLBlockHeader};
 use strata_ol_state_support_types::{IndexerState, IndexerWrites, WriteTrackingState};
 use strata_ol_state_types::{OLAccountState, OLState, WriteBatch};
-use strata_ol_stf::verify_block_with_root_fn;
+use strata_ol_stf::verify_block;
 use strata_primitives::{epoch::EpochCommitment, l1::L1BlockCommitment};
 use strata_service::ServiceState;
 use tracing::*;
@@ -250,18 +250,11 @@ impl ChainWorkerServiceState {
         let tracking_state = WriteTrackingState::new_from_state(parent_state);
         let mut indexer_state = IndexerState::new(tracking_state);
 
-        // Execute using new OL STF, providing materialized state root computation
-        // since WriteTrackingState::compute_state_root() is unsupported.
-        verify_block_with_root_fn(
+        verify_block(
             &mut indexer_state,
             block.header(),
             parent_header.cloned(),
             block.body(),
-            |s| {
-                s.inner()
-                    .compute_state_root_materialized()
-                    .map_err(Into::into)
-            },
         )?;
 
         // Extract outputs
--- a/crates/ol/state-support-types/src/batch_diff_layer.rs
+++ b/crates/ol/state-support-types/src/batch_diff_layer.rs
@@ -10,7 +10,7 @@ use strata_acct_types::{AccountId, AccountSerial, AcctError, AcctResult, Bitcoin
 use strata_asm_manifest_types::AsmManifest;
 use strata_identifiers::{Buf32, EpochCommitment, L1BlockId, L1Height};
 use strata_ledger_types::{IStateAccessor, NewAccountData};
-use strata_ol_state_types::WriteBatch;
+use strata_ol_state_types::{IStateBatchApplicable, WriteBatch};
 
 /// A read-only state accessor that overlays a stack of WriteBatch diffs.
 ///
@@ -21,6 +21,7 @@ use strata_ol_state_types::WriteBatch;
 /// The batch slice can be empty, making this a read-only wrapper for the base.
 /// This is useful for scenarios where you want to view state with pending
 /// changes applied without modifying anything.
+#[derive(Clone)]
 pub struct BatchDiffState<'batches, 'base, S: IStateAccessor> {
     base: &'base S,
     batches: &'batches [WriteBatch<S::AccountState>],
@@ -203,7 +204,14 @@ impl<'batches, 'base, S: IStateAccessor> IStateAccessor for BatchDiffState<'batc
     }
 
     fn compute_state_root(&self) -> AcctResult<Buf32> {
-        // TODO implement with new SSZ state summary type
+        Err(AcctError::Unsupported)
+    }
+}
+
+impl<'batches, 'base, S: IStateAccessor> IStateBatchApplicable
+    for BatchDiffState<'batches, 'base, S>
+{
+    fn apply_write_batch(&mut self, _batch: WriteBatch<Self::AccountState>) -> AcctResult<()> {
         Err(AcctError::Unsupported)
     }
 }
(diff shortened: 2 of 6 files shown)


crates/chain-worker-new/src/state.rs
```diff
@@ -250,7 +250,6 @@ impl ChainWorkerServiceState {
         let tracking_state = WriteTrackingState::new_from_state(parent_state);
         let mut indexer_state = IndexerState::new(tracking_state);
 
-        // Execute using new OL STF
```
storopoli, line 253: Why are we removing this comment now?

crates/ol/state-support-types/src/batch_diff_layer.rs
```diff
@@ -203,7 +204,14 @@ impl<'batches, 'base, S: IStateAccessor> IStateAccessor for BatchDiffState<'batc
     }
 
     fn compute_state_root(&self) -> AcctResult<Buf32> {
-        // TODO implement with new SSZ state summary type
```
storopoli, line 206: Should we keep this TODO?

crates/ol/state-support-types/src/write_tracking_layer.rs
```diff
@@ -64,7 +64,8 @@ impl<'base, S: IStateAccessor> WriteTrackingState<'base, S> {
     }
 }
 
-impl<'base, S: IStateAccessor> IStateAccessor for WriteTrackingState<'base, S>
+impl<'base, S: IStateAccessor + Clone + IStateBatchApplicable> IStateAccessor
+    for WriteTrackingState<'base, S>
 where
     S::AccountState: Clone + IAccountStateConstructible + IAccountStateMut,
```
storopoli, line 70: This can be much cleaner with

```suggestion
impl<'base, S> IStateAccessor for WriteTrackingState<'base, S>
where
S: IStateAccessor + Clone + IStateBatchApplicable,
S::AccountState: Clone + IAccountStateConstructible + IAccountStateMut,
```

crates/ol/state-support-types/src/batch_diff_layer.rs
```diff
@@ -21,6 +21,7 @@ use strata_ol_state_types::WriteBatch;
 /// The batch slice can be empty, making this a read-only wrapper for the base.
 /// This is useful for scenarios where you want to view state with pending
 /// changes applied without modifying anything.
+#[derive(Clone)]
```
storopoli, line 24: Do we really need clone in this "read-only" accessor?

`compute_state_root(&self)` needs to evaluate root after applying pending writes.
`self.base` is `&S` (immutable), so we can’t apply a batch directly to it.
Without a dedicated “compute root from base + batch” API, the generic fallback is:
1. clone base state (`S: Clone`)
2. apply batch
3. compute root

For me it seems a sign the trait model is a bit off. A cleaner direction is to split capabilities (read-only accessor vs batch-applicable vs root-computable), so read-only wrappers like `BatchDiffState` don’t need to implement capabilities they can’t truly support.

krsnapaudel: Yes there is some weirdness both ways.

krsnapaudel: @storopoli What are your thoughts on commit 1 approach?

storopoli: It leaks an internal concern into callers. `verify_block` callers now need to know how to compute roots, which is easy to misuse and harder to keep consistent.

krsnapaudel: Agreed. I will spend some more time tomorrow.

crates/ol/state-support-types/src/batch_diff_layer.rs
```diff
@@ -203,7 +204,14 @@ impl<'batches, 'base, S: IStateAccessor> IStateAccessor for BatchDiffState<'batc
...
+    }
+}
+
+impl<'batches, 'base, S: IStateAccessor> IStateBatchApplicable
+    for BatchDiffState<'batches, 'base, S>
+{
+    fn apply_write_batch(&mut self, _batch: WriteBatch<Self::AccountState>) -> AcctResult<()> {
         Err(AcctError::Unsupported)
     }
 }
```
delbonis, line 217: This trait should not be implemented it we can only return `AcctError::Unsupported` like this.  We don't have the excuse here like with `compute_state_root` since it's a separate trait of its own.

crates/ol/state-support-types/src/write_tracking_layer.rs
```diff
@@ -194,12 +195,14 @@ where
     }
 
     fn compute_state_root(&self) -> AcctResult<Buf32> {
-        // TODO implement with new SSZ state summary type
-        Err(AcctError::Unsupported)
+        let mut materialized = (*self.base).clone();
+        materialized.apply_write_batch(self.batch.clone())?;
+        materialized.compute_state_root()
```
delbonis, line 200: `materialized` is not really a good term for this.  Just call it `state_copy` or something.

PR #1358, opened 2026-02-12
User assigned withdrawal

purusang:
## Description

In this pr we add a feature to allow users to specify operator they want to process their withdrawal with. 

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
- Wire formats (Solidity ABI event, strata-codec structs, OL log payloads) don't support `Option<T>`, so `NO_PREFERRED_OPERATOR = u32::MAX` is used as a sentinel in the EE and OL layers
- The sentinel is defined in two places (`alpen_reth_primitives` and `strata_ol_msg_types`) because these are separate domains that shouldn't depend on each other
- `OperatorTable::add_operators()` guards against ever assigning index `u32::MAX`

## AI usage
Claude has been used for scope study, planning and partially in implementation. I have rephrased the `notes to reviewers` with the help of AI to make it clearer.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues
1763

### DIFF
--- a/functional-tests/tests/el_bridge_precompile.py
+++ b/functional-tests/tests/el_bridge_precompile.py
@@ -12,13 +12,13 @@ withdrawal_intent_event_abi = {
     "anonymous": False,
     "inputs": [
         {"indexed": False, "internalType": "uint64", "name": "amount", "type": "uint64"},
-        {"indexed": False, "internalType": "bytes", "name": "destination", "type": "bytes"},
         {"indexed": False, "internalType": "uint32", "name": "selectedOperator", "type": "uint32"},
+        {"indexed": False, "internalType": "bytes", "name": "destination", "type": "bytes"},
     ],
     "name": "WithdrawalIntentEvent",
     "type": "event",
 }
-event_signature_text = "WithdrawalIntentEvent(uint64,bytes,uint32)"
+event_signature_text = "WithdrawalIntentEvent(uint64,uint32,bytes)"
 
 
 @flexitest.register


bin/alpen-cli/src/cmd/withdraw.rs
```diff
@@ -101,3 +106,22 @@ pub async fn withdraw(
...
+fn encode_bridge_out_calldata(operator: Option<u32>, bosd: &Descriptor) -> Vec<u8> {
+    let bosd_bytes = bosd.to_bytes();
+    let mut buf = Vec::with_capacity(1 + 4 + bosd_bytes.len());
+    match operator {
+        None => {
+            // B=0: no operator preference
+            buf.push(0);
+        }
+        Some(idx) => {
+            // B=4: operator index encoded as 4 big-endian bytes
```
barakshani: why `B=4`?

crates/reth/primitives/src/lib.rs
```diff
@@ -19,6 +22,10 @@ pub struct WithdrawalIntent {
 
     /// withdrawal request transaction id
     pub withdrawal_txid: Buf32,
+
+    /// User's preferred operator index for withdrawal assignment.
+    /// [`NO_PREFERRED_OPERATOR`] means no preference (random assignment).
+    pub preferred_operator: u32,
```
barakshani:
```suggestion
    pub selected_operator: u32,
```
there's no notion of preference here, (as long as the selection is correct) only that operator can be assigned.

bin/alpen-cli/src/cmd/withdraw.rs
```diff
@@ -26,6 +26,10 @@ pub struct WithdrawArgs {
     /// the signet address to send funds to. defaults to a new internal wallet address
     #[argh(positional)]
     address: Option<String>,
+
+    /// preferred operator index for withdrawal assignment
```
barakshani:
```suggestion
    /// selected-operator index for withdrawal assignment

File: crates/alpen-ee/block-assembly/src/package.rs [hunk 4]
```diff
@@ -75,10 +76,11 @@ pub(crate) fn build_block_package<TPayload: EnginePayload>(
 fn create_withdrawal_init_message_payload(
     dest_desc: Descriptor,
     value: BitcoinAmount,
+    preferred_operator: u32,
 ) -> MsgPayload {
     // Encode the deposit message data
-    let withdrawal_data =
-        WithdrawalMsgData::new(0, dest_desc.to_bytes()).expect("valid descriptor");
+    let withdrawal_data = WithdrawalMsgData::new(0, dest_desc.to_bytes(), preferred_operator)
+        .expect("valid descriptor");
```
barakshani: is this still applicable?

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -183,12 +180,20 @@ impl AssignmentEntry {
...
-        let current_assignee = eligible_operators
-            .active_indices()
-            .nth(random_index)
-            .expect("random_index is within bounds of active_count");
+        // Honor preferred operator if eligible, otherwise fall back to random selection
+        let current_assignee = if let Some(idx) =
+            preferred_operator.filter(|&idx| eligible_operators.is_active(idx))
+        {
+            idx
+        } else {
```
barakshani, line 191:
```suggestion
        // Select a random operator from eligible ones
        } else {
        // Use ChaChaRng with L1 block ID as seed for deterministic random selection

File: crates/asm/subprotocols/bridge-v1/src/state/assignment.rs [hunk 6]
```diff
@@ -519,6 +527,74 @@ mod tests {
...
+            withdrawal_cmd,
+            fulfillment_deadline,
+            &current_active_operators,
+            seed,
+            Some(bogus_idx),
+        )
+        .unwrap();
+
+        // Should still get assigned to a valid active operator via random fallback
+        assert!(current_active_operators.is_active(assignment.current_assignee()));
```
barakshani, line 599: I'd also check that the assigned index _isn't_ bogus_idx.

barakshani: better, that it's within the range of current_active_operators.len()

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
```diff
@@ -329,6 +329,10 @@ impl OperatorTable {
                 continue;
             }
 
+            if self.next_idx == u32::MAX {
+                panic!("Operator index space exhausted: u32::MAX is reserved as a sentinel");
```
barakshani, line 333: I'd just set it to `0` in stead of panicking.

crates/asm/subprotocols/checkpoint/src/verification.rs
```diff
@@ -174,7 +175,14 @@ fn extract_and_validate_withdrawal_intents(
             return Err(InvalidCheckpointPayload::MalformedWithdrawalDestDesc.into());
         };
 
-        let withdraw_output = WithdrawOutput::new(destination, withdrawal_data.amt().into());
+        let preferred_operator = (withdrawal_data.preferred_operator != NO_PREFERRED_OPERATOR)
+            .then_some(withdrawal_data.preferred_operator);
```
barakshani: isn't this an error? we cannot add an assignee during validation (if I understand this func correctly). cc @prajwolrg

crates/asm/subprotocols/debug-v1/src/txs.rs
```diff
@@ -105,7 +105,7 @@ fn parse_withdrawal_from_aux_data(aux_data: &[u8]) -> Result<WithdrawOutput, Deb
     let dest = Descriptor::from_bytes(desc_bytes)
         .map_err(|e| DebugTxParseError::InvalidDescriptorFormat(e.to_string()))?;
 
-    let withdraw_output = WithdrawOutput::new(dest, amt);
+    let withdraw_output = WithdrawOutput::new(dest, amt, None);
```
barakshani: why `None`? shouldn't we take this value from somewhere?

crates/evm-ee/src/execution.rs
```diff
@@ -53,8 +53,12 @@ fn convert_withdrawal_intents_to_messages(
...
     for intent in withdrawal_intents {
         // Create withdrawal message data with fees=0 (currently ignored) and destination descriptor
-        let withdrawal_msg = WithdrawalMsgData::new(0, intent.destination.to_bytes().to_vec())
-            .expect("destination descriptor too long");
+        let withdrawal_msg = WithdrawalMsgData::new(
+            0,
+            intent.destination.to_bytes().to_vec(),
+            intent.preferred_operator,
+        )
+        .expect("destination descriptor too long");
```
barakshani: is this still applicable?

crates/ol/chain-types/src/log_payloads.rs
```diff
@@ -4,22 +4,29 @@ use strata_codec::{Codec, VarVec};
 
 /// Payload for a simple withdrawal intent log.
 ///
-/// This log is emitted when a withdrawal intent is created through the bridge.
-/// It contains the withdrawal amount and destination information.
+/// Emitted by the OL STF when a withdrawal message is processed at the bridge
+/// gateway account.
```
barakshani, line 8: are you sure this correction is in place? (I don't know)

crates/ol/stf/src/account_processing.rs
```diff
@@ -127,9 +127,11 @@ fn handle_bridge_gateway_message<S: IStateAccessor>(
     }
 
     // If it is, then we can emit a OL log with the amount and destination.
+    let preferred_operator = withdrawal_data.preferred_operator();
     let log_data = SimpleWithdrawalIntentLogData {
         amt: withdrawal_amt.into(),
         dest: withdrawal_data.into_dest_desc(),
+        preferred_operator,
```
barakshani:
```suggestion
    let log_data = SimpleWithdrawalIntentLogData {
        amt: withdrawal_amt.into(),
        dest: withdrawal_data.into_dest_desc(),
        preferred_operator: withdrawal_data.preferred_operator(),
```
for consistency.

crates/reth/evm/src/precompiles/bridge.rs
```diff
@@ -9,14 +9,23 @@ use crate::{
...
+/// - B=0: no operator preference
+/// - B=1..4: operator index encoded as B big-endian bytes
+/// - B>4: invalid
 pub(crate) fn bridge_context_call(mut input: PrecompileInput<'_>) -> PrecompileResult {
-    let destination = input.data;
+    let (preferred_operator, bosd_data) = parse_calldata(input.data)?;
 
     // Validate that this is a valid BOSD
-    let _ = try_into_bosd(destination)?;
+    let _ = try_into_bosd(bosd_data)?;
```
barakshani: is `bosd_data` (previously `destination`) already encoded as bosd? if so, is try_into (naming) correct, or should it be _validate_?

crates/reth/evm/src/precompiles/bridge.rs
```diff
@@ -62,7 +72,45 @@ pub(crate) fn bridge_context_call(mut input: PrecompileInput<'_>) -> PrecompileR
...
+        .split_first()
+        .ok_or_else(|| PrecompileError::other("Empty calldata"))?;
+
+    let b = b as usize;
+
+    if b == 0 {
+        return Ok((NO_PREFERRED_OPERATOR, rest));
+    }
+
+    if b > MAX_OPERATOR_INDEX_LEN {
```
barakshani: else if?

crates/reth/evm/src/precompiles/bridge.rs
```diff
@@ -62,7 +72,45 @@ pub(crate) fn bridge_context_call(mut input: PrecompileInput<'_>) -> PrecompileR
...
+        ));
+    }
+
+    if rest.len() < b {
+        return Err(PrecompileError::other(
+            "Calldata too short for operator index",
+        ));
+    }
+
+    let (operator_bytes, bosd_data) = rest.split_at(b);
```
barakshani: I'd also check that bosd_data has sufficient size (at least greater than zero)

crates/reth/evm/src/precompiles/bridge.rs
```diff
@@ -62,7 +72,45 @@ pub(crate) fn bridge_context_call(mut input: PrecompileInput<'_>) -> PrecompileR
...
+
+    let operator_idx = operator_bytes
+        .iter()
+        .fold(0u32, |acc, &byte| (acc << 8) | byte as u32);
+
+    Ok((operator_idx, bosd_data))
+}
+
+/// Ensures that input is a valid BOSD [`Descriptor`].
 fn try_into_bosd(maybe_bosd: &[u8]) -> Result<Descriptor, PrecompileError> {
```
barakshani: _check_ rather than _try_? I'm not sure about our naming conventions.

crates/reth/evm/src/precompiles/bridge.rs
```diff
@@ -72,3 +120,73 @@ fn try_into_bosd(maybe_bosd: &[u8]) -> Result<Descriptor, PrecompileError> {
...
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn test_parse_calldata_truncated_operator_bytes() {
+        // B=4 but only 2 operator bytes follow
+        let data = vec![4, 0x01, 0x02];
+        let result = parse_calldata(&data);
+        assert!(result.is_err());
+    }
```
barakshani, line 156: we will eventually need test that fail on mixing idx/bosd (for example we set `b` 2 bytes, but encode only 1 byte (say, idx = 3) then some of bosd are considered to be part of the idx.

barakshani (commented):
nice work so far

bin/alpen-cli/src/cmd/withdraw.rs
purusang: It could be any length in [1,4] but to focus on the main work of this pr I just used a constant value here (alpen-cli). But our precompile supports the dynamic range so should be easily modifiable here.

crates/evm-ee/src/execution.rs
barakshani: my question is about the error message: previously the problem must have been with the descriptor, now it can be with the operator as well.

crates/asm/subprotocols/checkpoint/src/verification.rs
barakshani: why don't you do `let preferred_operator = withdrawal_data.preferred_operator`?

crates/evm-ee/src/execution.rs
purusang: In this step operator cannot cause failure.

crates/asm/subprotocols/checkpoint/src/verification.rs
purusang: Here we are trying to avoid u32::max as the assigned operator and also converting the preferred operator to Option<u32>. The bridge subprotocol and onwards will deal with this optional field.

crates/ol/chain-types/src/log_payloads.rs
purusang: Yes, previous docstring was a bit ambiguous to me, specially the mention of bridge and a bit less informative.

crates/alpen-ee/block-assembly/src/package.rs
purusang: Yes

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
purusang: In this section we are adding an operator so wrapping around to 0 may indicate that we can overwrite 0th operator which may have been safe guarded at the respective part. So I think panicking would make more sense to me.

crates/asm/subprotocols/checkpoint/src/verification.rs
purusang: > isn't this an error? we cannot add an assignee during validation (if I understand this func correctly).

Here, in checkpoint subprotocol, we are not assigning. It is just passing up of user's selection. The actual assignment is happening in [bridge-v1/bridge.rs](https://github.com/alpenlabs/strata/blob/STR-1763-user-assigned-withdrawal/crates/asm/subprotocols/bridge-v1/src/state/bridge.rs#L148)

crates/reth/evm/src/precompiles/bridge.rs
purusang: The following code can handle that case. We want to error if b>4 since that would mean we support more than u32::MAX.

crates/reth/evm/src/precompiles/bridge.rs
purusang: Will update it to something with validate*

crates/asm/subprotocols/checkpoint/src/verification.rs
barakshani: I see, I misunderstood. However, what happens if `preferred_operator = NO_PREFERRED_OPERATOR`?

crates/reth/evm/src/precompiles/bridge.rs
purusang: That will fail in the next check `validate_bosd()/try_into_bosd()` if that happens, anyway have added a test for it.

crates/asm/subprotocols/debug-v1/src/txs.rs
purusang: 🤔

crates/asm/subprotocols/bridge-v1/src/state/operator.rs
barakshani: Note that this is a change to the previous implementation, cc @prajwolrg.

crates/asm/subprotocols/checkpoint/src/verification.rs
purusang: That will be `None`.

crates/ol/msg-types/src/withdrawal.rs
```diff
@@ -30,18 +33,25 @@ pub struct WithdrawalMsgData {
     /// Bitcoin Output Script Descriptor describing the withdrawal output.
     // TODO idk why, but I can't make the MAX_WITHDRAWAL_DESC_LEN const generic work
     dest_desc: VarVec<u8>,
+
+    /// User's preferred operator index for withdrawal assignment.
+    preferred_operator: u32,
```
barakshani: this isn't always u32, right? it is variable as well.

barakshani: IIUC right now we explicitly decode the user's message and later encode it again in the ASM log. I think we can keep it encoded.

crates/alpen-ee/block-assembly/src/package.rs
barakshani: see my comment here https://github.com/alpenlabs/alpen/pull/1358/changes#r2803616612.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -161,11 +161,8 @@ impl AssignmentEntry {
         fulfillment_deadline: BitcoinBlockHeight,
         current_active_operators: &OperatorBitmap,
         seed: L1BlockId,
+        preferred_operator: Option<OperatorIdx>,
```
barakshani: need to be `selected_operator` as you already started making this naming transition.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -183,12 +180,22 @@ impl AssignmentEntry {
...
-            .active_indices()
-            .nth(random_index)
-            .expect("random_index is within bounds of active_count");
+        // Honor preferred operator if eligible, otherwise fall back to random selection
+        let current_assignee = if let Some(idx) =
+            preferred_operator.filter(|&idx| eligible_operators.is_active(idx))
+        {
+            idx
+        } else {
+            // Select a random operator from eligible ones.
```
```suggestion
```
I know i added this, but given you write about "otherwise fall back to random selection" this is redundant.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
```diff
@@ -519,6 +529,76 @@ mod tests {
         assert_eq!(assignment.previous_assignees.active_count(), 0);
     }
 
+    #[test]
+    fn test_create_with_preferred_operator() {
```
```suggestion
    fn test_create_with_selected_operator() {

File: crates/asm/subprotocols/bridge-v1/src/state/assignment.rs [hunk 22]
```diff
@@ -519,6 +529,76 @@ mod tests {
...
+            }
+        };
+
+        let withdrawal_cmd: WithdrawalCommand = arb.generate();
+        let fulfillment_deadline: BitcoinBlockHeight = 100;
+        let seed: L1BlockId = arb.generate();
+        let current_active_operators = deposit_entry.notary_operators().clone();
+
+        // Pick the second active operator as preferred
+        let preferred_idx = current_active_operators
```
```suggestion
        // Pick the second active operator
        let selected_idx = current_active_operators

File: crates/asm/subprotocols/debug-v1/src/constants.rs [hunk 23]
```diff
@@ -17,8 +17,12 @@ pub(crate) const AMOUNT_SIZE: usize = 8;
 /// Offset of amount field in auxiliary data.
 pub(crate) const AMOUNT_OFFSET: usize = 0;
 
-/// Offset of descriptor field in auxiliary data.
-pub(crate) const DESCRIPTOR_OFFSET: usize = AMOUNT_OFFSET + AMOUNT_SIZE;
+/// Offset of the operator-length byte (`B`) in auxiliary data.
+pub(crate) const OPERATOR_LEN_OFFSET: usize = AMOUNT_OFFSET + AMOUNT_SIZE;
```
barakshani: don't we need DESCRIPTOR_OFFSET as well?

crates/asm/subprotocols/debug-v1/src/constants.rs
```diff
@@ -27,5 +31,7 @@ pub(crate) const MIN_DESCRIPTOR_SIZE: usize = 20;
 
 /// Minimum auxiliary data length for mock withdrawal intent.
 ///
-/// Format: `[amount: 8 bytes][descriptor: variable]`
-pub(crate) const MIN_MOCK_WITHDRAW_INTENT_AUX_DATA_LEN: usize = AMOUNT_SIZE + MIN_DESCRIPTOR_SIZE;
+/// Format: `[amount: 8 bytes][1 byte B][B bytes: operator index][descriptor: variable]`
+/// Minimum case is B=0 (no operator preference).
+pub(crate) const MIN_MOCK_WITHDRAW_INTENT_AUX_DATA_LEN: usize =
+    AMOUNT_SIZE + 1 + MIN_DESCRIPTOR_SIZE;
```
barakshani: I think we need AMOUNT_OFFSET. Do you know where this const is being used?

crates/asm/subprotocols/debug-v1/src/txs.rs
```diff
@@ -94,25 +103,52 @@ fn parse_withdrawal_from_aux_data(aux_data: &[u8]) -> Result<WithdrawOutput, Deb
...
-    // Extract amount using constants
+    // Extract amount (8 bytes, big-endian)
     let amount_end = AMOUNT_OFFSET + AMOUNT_SIZE;
     let amount_bytes: [u8; AMOUNT_SIZE] = aux_data[AMOUNT_OFFSET..amount_end].try_into().unwrap();
     let amount = u64::from_be_bytes(amount_bytes);
     let amt = BitcoinAmount::from_sat(amount);
 
-    // Extract descriptor (self-describing, no length field needed)
-    let desc_bytes = &aux_data[DESCRIPTOR_OFFSET..];
+    // Extract preferred operator using variable-length encoding
```
barakshani: selected

crates/asm/subprotocols/debug-v1/src/txs.rs
```diff
@@ -86,6 +90,11 @@ fn parse_mock_asm_log_tx(tx: &TxInputRef<'_>) -> Result<ParsedDebugTx, DebugTxPa
 }
 
 /// Parses withdrawal data from auxiliary data bytes.
+///
+/// Format: `[amount: 8 bytes][1 byte B][B bytes: operator index (big-endian)][descriptor: variable]`
+/// - B=0: no operator preference
+/// - B=1..4: operator index encoded as B big-endian bytes
```
barakshani: This deviates a bit from the proposed solution (I'm not saying it's not good, but I want to discuss this): in the proposed solution B ranges between 0 and 32, because it encodes how many bits we need for the operator index. In here we encode how many bytes are needed. The latter is probably more correct because on Bitcoin (and probably on Ethereum as well, where the withdrawal request tx is sent) we work on byte level. This means that we use full byte for B, even though it uses at most 3 bits.
We could have a better compression here (as indicated in the ticket), but for one byte it probably doesn't worth it (we had similar discussion, and conclusion, for how to encode the full deposit path (snark account id and subject id) for the deposit tx).

crates/asm/subprotocols/debug-v1/src/txs.rs
```diff
@@ -94,25 +103,52 @@ fn parse_withdrawal_from_aux_data(aux_data: &[u8]) -> Result<WithdrawOutput, Deb
...
+            expected: operator_data_offset + b,
+            actual: aux_data.len(),
+        });
+    } else {
+        // B=1..4: decode B big-endian bytes into a u32 operator index
+        let operator_idx = aux_data[operator_data_offset..operator_data_offset + b]
+            .iter()
+            .fold(0u32, |acc, &byte| (acc << 8) | byte as u32);
+        (Some(operator_idx), operator_data_offset + b)
+    };
```
barakshani: For the descriptor below we have 
```
let dest = Descriptor::from_bytes(desc_bytes).map_err
```
Shouldn't we have similar logic for operator index such that it can be used across different places (if needed)?

crates/asm/subprotocols/debug-v1/src/txs.rs
```diff
@@ -215,9 +256,10 @@ mod tests {
         ];
         let p2tr_descriptor = Descriptor::new_p2tr(&x_only_pubkey).unwrap();
 
-        // Create auxiliary data: [amount: 8 bytes][descriptor: 33 bytes]
+        // Create auxiliary data: [amount: 8][B=0: 1][descriptor: 33]
```
barakshani: better also test B>1, either here (since we have a test for B=0 above) or new tests.

crates/evm-ee/src/execution.rs
barakshani: keeping this open, similar to [above](https://github.com/alpenlabs/alpen/pull/1358/changes#r2812274613).

crates/reth/evm/src/precompiles/bridge.rs
```diff
@@ -62,13 +75,146 @@ pub(crate) fn bridge_context_call(mut input: PrecompileInput<'_>) -> PrecompileR
...
+    Descriptor::from_bytes(data)
+        .map_err(|_| PrecompileError::other("Invalid BOSD: expected a valid BOSD descriptor"))?;
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const DUMMY_BOSD: &[u8] = &[0x00; 20]; // placeholder BOSD bytes
```
barakshani: is this a valid descriptor?

crates/reth/evm/src/precompiles/bridge.rs
```diff
@@ -62,13 +75,146 @@ pub(crate) fn bridge_context_call(mut input: PrecompileInput<'_>) -> PrecompileR
...
+        // as part of the operator index, producing a wrong operator and shifted BOSD.
+        let mut data = vec![2, 0x03]; // B=2, first operator byte
+        data.extend_from_slice(DUMMY_BOSD); // 0x00 of BOSD will be eaten as 2nd operator byte
+
+        let (operator, bosd) = parse_calldata(&data).unwrap();
+        // Operator becomes (0x03 << 8) | 0x00 = 768 instead of intended 3
+        assert_eq!(operator, 0x0300);
+        // BOSD is truncated by 1 byte
+        assert_eq!(bosd.len(), DUMMY_BOSD.len() - 1);
     }
```
barakshani, line 156: Shouldn't this give an error on decoding the descriptor? I want this to fail on the descriptor logic (size mismatch).

crates/alpen-ee/block-assembly/src/package.rs
prajwolrg: nit: The value `0` here seems to be the operator fee. Let's create a constant here for the operator fee and use that instead of using it directly. Helps avoid confusion.

prajwolrg: I also wonder if we would ever need `operator_fee` be information passed through EE. Right now the `operator_fee` is defined as part of the ASM bridge subprotocol. I imagine it'll be helpful for scenarios where we want to do some sort of fee bumping in the future, but I still think it's worth documenting why this is needed (if it's needed at all). CC: @delbonis @sapinb.

crates/ol/msg-types/src/withdrawal.rs
```diff
@@ -15,6 +15,9 @@ pub const WITHDRAWAL_REJECTION_MSG_TYPE_ID: u16 = 0x05;
 /// Maximum length for withdrawal destination descriptor.
 pub const MAX_WITHDRAWAL_DESC_LEN: usize = 255;
 
+/// Sentinel value indicating no preferred operator for withdrawal assignment.
+pub const NO_PREFERRED_OPERATOR: u32 = u32::MAX;
```
prajwolrg: I wonder if we should turn this into preferred_operator value as `Varint` instead of u32. Since this information would be part of the information sent to L1, it might be worth it making it `Varint`. We already have a ticket [here](https://alpenlabs.atlassian.net/browse/STR-1861). In that case it might make sense making it `Varint::ZERO` instead, as that would reduce DA cost if user don't have preferred operator.

crates/reth/evm/src/precompiles/bridge.rs
```diff
@@ -9,14 +9,23 @@ use crate::{
...
+const MAX_OPERATOR_INDEX_LEN: usize = 4;
+
 /// Custom precompile to burn rollup native token and add bridge out intent of equal amount.
 /// Bridge out intent is created during block payload generation.
 /// This precompile validates transaction and burns the bridge out amount.
+///
+/// Calldata format: `[1 byte B][B bytes: operator index (big-endian)][BOSD bytes]`
+/// - B=0: no operator preference
+/// - B=1..4: operator index encoded as B big-endian bytes
+/// - B>4: invalid
```
prajwolrg: nit: `B>4: invalid` is a bit confusing here. Did you mean `B<4 is invalid` instead? B cannot be greater than 4.

crates/reth/evm/src/precompiles/bridge.rs
```diff
@@ -9,14 +9,23 @@ use crate::{
     utils::wei_to_sats,
 };
 
+/// Maximum number of bytes used to encode the operator index in calldata.
+/// Operator index is a u32, so at most 4 bytes.
+const MAX_OPERATOR_INDEX_LEN: usize = 4;
```
prajwolrg: nit: We can use `std::mem::size_of::<u32>`. We can also use define `OperatorIdx` type and do `std::mem::size_of::<OperatorIdx>`.

crates/asm/subprotocols/debug-v1/src/constants.rs
purusang: Descriptor offset become variable based on the length of Operator in runtime so handled there.

crates/reth/evm/src/precompiles/bridge.rs
prajwolrg: I don't think so. Since we have the first byte be a tag identifier, I think we need more than 20 bytes. Maybe we can use `Descriptor::arbitrary()` to get a descriptor. But for the tests of `parse_calldata`, this should be fine.

crates/reth/evm/src/precompiles/bridge.rs
```diff
@@ -62,7 +72,45 @@ pub(crate) fn bridge_context_call(mut input: PrecompileInput<'_>) -> PrecompileR
...
 }
 
-/// Ensure that input is a valid BOSD [`Descriptor`].
+/// Parses bridge out calldata into a preferred operator index and BOSD bytes.
+///
+/// Format: `[1 byte B][B bytes: operator index (big-endian)][BOSD bytes]`
+/// - B=0: no preference, returns [`NO_PREFERRED_OPERATOR`]
+/// - B=1..4: decodes B bytes as a big-endian u32 operator index
+/// - B>4: error
+fn parse_calldata(data: &[u8]) -> Result<(u32, &[u8]), PrecompileError> {
```
prajwolrg: blocking: We shouldn't parse the calldata in this way. This means that if any contract wants to call this precompile, it would need to write the logic to encode it solidity. Since we post the statediff, having a larger calldata doesn't affect the DA size, so I don't see a reason to introduce this complexity in the code. 

It makes more sense to do this optimization as part of encoding the `OLLog` instead as OLLogs are part of the checkpoint payload. We can implement a custom `Codec` logic for WithdrawalMsgData that does this optimization.

crates/alpen-ee/block-assembly/src/package.rs
barakshani: in the future we will want the user to be able to specify the fee

crates/reth/evm/src/precompiles/bridge.rs
barakshani: you propose we send the full 32 bits for operator index?

prajwolrg: I'm not saying we send the full 32 bits for operator index. I'm saying that this is the wrong place to do this optimization/encoding. We shouldn't try to parse calldata in precompile in any other way that supported by solidity ABI.

crates/reth/evm/src/precompiles/bridge.rs
```diff
@@ -77,46 +74,20 @@ pub(crate) fn bridge_context_call(mut input: PrecompileInput<'_>) -> PrecompileR
...
-    if rest.len() < b {
-        return Err(PrecompileError::other(
-            "Calldata too short for operator index",
-        ));
-    }
-
-    let (operator_bytes, bosd_data) = rest.split_at(b);
-
-    if bosd_data.is_empty() {
+    if data.len() < OPERATOR_INDEX_SIZE + 1 {
```
barakshani: make <= instead? (seems more elegant to me than adding 1)

crates/ol/chain-types/src/log_payloads.rs
```diff
@@ -15,6 +15,7 @@ pub struct SimpleWithdrawalIntentLogData {
     pub dest: VarVec<u8>,
 
     /// User's selected operator index for withdrawal assignment.
+    // TODO(STR-1861): encode as varint to reduce DA cost in checkpoint payloads.
```
barakshani, line 18: str-1861 is complete, what the relation to it?

prajwolrg: STR-1861 is not complete. It's assigned to me for now.

barakshani: oops

prajwolrg (dismissed):
The changes looks good.

I'm slightly in favor of using `u32::ZERO` as the sentinel value instead of `u32::MAX`, but that means operator index must start at 1 instead of 0, which is the current case. But revisiting this can be done as a separate ticket as well. I created a separate ticket to track this [here](https://alpenlabs.atlassian.net/browse/STR-2332).

crates/evmexec/src/engine.rs
```diff
@@ -423,6 +423,7 @@ fn to_bridge_withdrawal_intent(
         amt,
         destination,
         withdrawal_txid,
+        ..
     } = rpc_withdrawal_intent;
     WithdrawalIntent::new(BitcoinAmount::from_sat(amt), destination, withdrawal_txid)
```
storopoli: `selected_operator` from `alpen_reth_node::WithdrawalIntent` is dropped here via `..` and not propagated into `strata_bridge_types::WithdrawalIntent`. This silently loses user operator selection on the `strata-evmexec` path. Can we plumb this field through (or explicitly reject non-sentinel selections on this path)?

storopoli (changes requested):

crates/evmexec/src/engine.rs
barakshani: wait wait, is this tested??

storopoli: Not for this exact `evmexec` conversion path yet.

In evmexec, the path flagged in `to_bridge_withdrawal_intent` is not covered by a test asserting operator propagation.

Current `evmexec` test uses empty intents: 

https://github.com/alpenlabs/alpen/blob/f223385482d4f19fec4a9378bb5a7c0dd82e404b/crates/evmexec/src/engine.rs#L486-L494

And we only check if the result `is_ok()`:

https://github.com/alpenlabs/alpen/blob/f223385482d4f19fec4a9378bb5a7c0dd82e404b/crates/evmexec/src/engine.rs#L610-L624

BUT! I don't know if `evmexec` will be used/is relevant for M2 new EE <> OL code since I don't see any trace of being used by the `strata` binary OL client.

So this issue is related to the `strata-client` OLD bundled OL/EE client.

I think we can solve this.

barakshani: Let's add a test.

crates/ol/msg-types/src/withdrawal.rs
storopoli: I still think that this is relevant and a very low-hanging fruit optimization.

storopoli (approved):
ACK 7839e7ecb54030e9f2b8801832112c554ab80b3a

I still think we should go with the `Varint` and `Varint::ZERO` for "no use preferred operator just use a random one" as @prajwolrg described here:

https://github.com/alpenlabs/alpen/pull/1358/changes/ee27a9255c3eb1f50d63232a3e122bb3a3ba3b7e#r2815199054

EDIT: Or `u32::ZERO`

crates/ol/msg-types/src/withdrawal.rs
barakshani: why is this an optimization?

storopoli: Because we assume that most deposits won't have a preferred operator, hence they will be coded as `0` using a `Varint` which would use less bytes to represent this in DA costs.

But now we are hardcoding this to be a `u32` and the maximum value.

barakshani: Not good. I will comment on the ticket. But this should not be a blocker for getting this in. (anyway there will be part 2)

barakshani: This [ticket](https://alpenlabs.atlassian.net/browse/STR-2332)!

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
delbonis, line 1: We should rework this type, maybe using a builder.  The "constructor", `create_with_random_assignment`, does a *lot* of context-dependent logic that should not be in a constructor.

This is also evidenced by having to add `None` in all those test cases, where using a builder could help us avoid having to do that.

crates/asm/msgs/bridge/src/lib.rs
```diff
@@ -32,12 +32,23 @@ pub struct WithdrawOutput {
 
     /// Amount to withdraw (in satoshis).
     pub amt: BitcoinAmount,
+
+    /// User's selected operator index for withdrawal assignment.
+    pub selected_operator: Option<u32>,
 }
```
delbonis: This type is supposed to describe the output on the fulfillment transaction, as indicated by the doc.  This isn't the right place for this information since it's not part of the transaction output.

bin/alpen-cli/src/cmd/withdraw.rs
```diff
@@ -101,3 +106,15 @@ pub async fn withdraw(
...
+///
+/// When no operator is selected, `u32::MAX` is written as the sentinel value.
+fn encode_bridge_out_calldata(operator: Option<u32>, bosd: &Descriptor) -> Vec<u8> {
+    let bosd_bytes = bosd.to_bytes();
+    let mut buf = Vec::with_capacity(4 + bosd_bytes.len());
+    let selected = operator.unwrap_or(u32::MAX);
+    buf.extend_from_slice(&selected.to_be_bytes());
+    buf.extend_from_slice(&bosd_bytes);
+    buf
+}
```
delbonis: Should not construct this data on the fly here.  Use/define a type that describes the structured calldata fields and have it serialize to bytes.

crates/alpen-ee/block-assembly/src/package.rs
```diff
@@ -75,10 +76,15 @@ pub(crate) fn build_block_package<TPayload: EnginePayload>(
...
 ) -> MsgPayload {
     // Encode the deposit message data
-    let withdrawal_data =
-        WithdrawalMsgData::new(0, dest_desc.to_bytes()).expect("valid descriptor");
+    let withdrawal_data = WithdrawalMsgData::new(
+        DEFAULT_OPERATOR_FEE,
+        dest_desc.to_bytes(),
+        selected_operator,
+    )
+    .expect("valid descriptor");
```
delbonis: Should not panic if it's not a valid descriptor, we can't guarantee that here.

crates/asm/subprotocols/debug-v1/src/txs.rs
```diff
@@ -86,6 +87,10 @@ fn parse_mock_asm_log_tx(tx: &TxInputRef<'_>) -> Result<ParsedDebugTx, DebugTxPa
 }
 
 /// Parses withdrawal data from auxiliary data bytes.
+///
+/// Format: `[amount: 8 bytes][selected_operator: 4 bytes (big-endian u32)][descriptor: variable]`
+/// - `u32::MAX`: no operator selection (`selected_operator = None`)
+/// - Any other value: `selected_operator = Some(value)`
 fn parse_withdrawal_from_aux_data(aux_data: &[u8]) -> Result<WithdrawOutput, DebugTxParseError> {
```
delbonis: Define a `DebugWithdrawalAuxData` type that captures these fields and have a `from_aux_buf` function that decodes it instead of this clunky procedure.

crates/bridge-types/src/bridge_ops.rs
```diff
@@ -16,14 +16,24 @@ pub struct WithdrawalIntent {
 
     /// withdrawal request transaction id
     withdrawal_txid: Buf32,
+
+    /// User's selected operator index for withdrawal assignment.
+    /// `u32::MAX` means no selection (random assignment).
+    selected_operator: u32,
```
delbonis: This is kinda ad-hoc behavior we have to replicate in many places, maybe we should have the `selected_operator` constructor return `Option<u32>` if it's `u32::MAX`?  And then the constructor could be the same and panic if it's `Some(u32::MAX)`.

crates/bridge-types/src/bridge_ops.rs
```diff
@@ -42,6 +52,10 @@ impl WithdrawalIntent {
     pub const fn withdrawal_txid(&self) -> &Buf32 {
         &self.withdrawal_txid
     }
+
+    pub const fn selected_operator(&self) -> u32 {
+        self.selected_operator
+    }
 }
```
delbonis: I don't think any of these make sense to be `const`.

crates/evm-ee/src/execution.rs
```diff
@@ -52,9 +52,12 @@ fn convert_withdrawal_intents_to_messages(
...
-        let withdrawal_msg = WithdrawalMsgData::new(0, intent.destination.to_bytes().to_vec())
-            .expect("destination descriptor too long");
+        let withdrawal_msg = WithdrawalMsgData::new(
+            DEFAULT_OPERATOR_FEE,
+            intent.destination.to_bytes().to_vec(),
+            intent.selected_operator,
+        )
+        .expect("invalid withdrawal destination descriptor");
 
         let msg_data = encode_to_vec(&withdrawal_msg).expect("encoding failed");
```
delbonis, line 62: As a note, this is the kind of struct that we should be defining for all of these flat byte descriptors, and how it'd be used.

crates/ol/msg-types/src/withdrawal.rs
```diff
@@ -15,6 +15,18 @@ pub const WITHDRAWAL_REJECTION_MSG_TYPE_ID: u16 = 0x05;
...
 pub const MAX_WITHDRAWAL_DESC_LEN: usize = 255;
 
+/// Sentinel value indicating no selected operator for withdrawal assignment.
+pub const NO_SELECTED_OPERATOR: u32 = u32::MAX;
+
+/// Converts a raw operator index to `Option<u32>`, treating
+/// [`NO_SELECTED_OPERATOR`] as `None`.
+pub fn resolve_selected_operator(raw: u32) -> Option<u32> {
+    (raw != NO_SELECTED_OPERATOR).then_some(raw)
+}
```
delbonis: Maybe it'd be even better than what I suggested in the other comment to define a `struct OperatorSelection(u32);` that captures this whole concept, providing `specific(idx: u32)` and `any()` constructors.

crates/ol/msg-types/src/withdrawal.rs
```diff
@@ -58,12 +77,18 @@ impl WithdrawalMsgData {
...
+        self.selected_operator
+    }
 }
 
 impl Codec for WithdrawalMsgData {
     fn encode(&self, enc: &mut impl Encoder) -> Result<(), CodecError> {
         self.fees.encode(enc)?;
         self.dest_desc.encode(enc)?;
+        self.selected_operator.encode(enc)?;
         Ok(())
```
delbonis: This encoding is unaesthetic since it puts a fixed-size field *after* a variable size field.  Generally you put simpler fields (especially fixed-size ones) before more complex ones.

Also, I think we can just use the `Codec` macros for this!

crates/reth/primitives/src/lib.rs
```diff
@@ -27,5 +31,7 @@ sol! {
         uint64 amount,
         /// BOSD descriptor for withdrawal destinations in L1.
         bytes destination,
+        /// Selected operator index. `u32::MAX` means no selection.
+        uint32 selectedOperator,
     );
```
delbonis: Similar case, should reorder these.

crates/ol/msg-types/src/withdrawal.rs
storopoli: Ok solving this.

crates/asm/subprotocols/bridge-v1/src/state/assignment.rs
purusang: Have created a [ticket](https://alpenlabs.atlassian.net/browse/STR-2356) for this.

PR #1363, opened 2026-02-16
fix(asm): make AuxRequestCollector bounds-aware to prevent DoS 

irnb:
## Description

      A malicious L1 transaction tagged as a checkpoint can claim an arbitrary
      L1 height (e.g. 999999). The old AuxRequestCollector blindly forwarded
      this to the aux resolver, which failed on non-existent manifest hashes
      blocking processing of the entire L1 block and stalling the ASM.

      Make the collector bounds-aware by passing the accumulator's last
      inserted height at construction. Requests for ranges beyond this bound
      or with inverted start/end are silently dropped. Invalid payloads are
      still rejected during process_txs validation.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/asm/common/src/aux/collector.rs
+++ b/crates/asm/common/src/aux/collector.rs
@@ -4,7 +4,10 @@
 
 use bitcoin::Txid;
 
-use crate::aux::data::{AuxRequests, ManifestHashRange};
+use crate::{
+    aux::data::{AuxRequests, ManifestHashRange},
+    logging,
+};
 
 /// Collects auxiliary data requests from subprotocols.
 ///
@@ -19,14 +22,16 @@ use crate::aux::data::{AuxRequests, ManifestHashRange};
 #[derive(Debug)]
 pub struct AuxRequestCollector {
     requests: AuxRequests,
+    min_manifest_height: u64,
     max_manifest_height: u64,
 }
 
 impl AuxRequestCollector {
-    /// Creates a new empty collector bounded by the given maximum manifest height.
-    pub fn new(max_manifest_height: u64) -> Self {
+    /// Creates a new empty collector bounded by the given manifest height range.
+    pub fn new(min_manifest_height: u64, max_manifest_height: u64) -> Self {
         Self {
             requests: AuxRequests::default(),
+            min_manifest_height,
             max_manifest_height,
         }
     }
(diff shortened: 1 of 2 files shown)


crates/asm/common/src/aux/collector.rs
```diff
@@ -27,6 +37,10 @@ impl AuxRequestCollector {
     /// * `start_height` - Starting L1 block height (inclusive)
     /// * `end_height` - Ending L1 block height (inclusive)
     pub fn request_manifest_hashes(&mut self, start_height: u64, end_height: u64) {
+        if start_height > end_height || end_height > self.max_manifest_height {
+            return;
```
prajwolrg: I think we should add a log here that we are not resolving some heights.

crates/asm/common/src/aux/collector.rs
```diff
@@ -10,15 +10,25 @@ use crate::aux::data::{AuxRequests, ManifestHashRange};
...
+/// The collector is bounds-aware: it silently drops manifest hash requests for
+/// heights beyond what is available in the accumulator. This prevents malicious
+/// L1 transactions (e.g. a checkpoint claiming a beyond-tip L1 height) from
+/// causing the aux resolver to fail, which would block processing of the entire
+/// L1 block.
+#[derive(Debug)]
 pub struct AuxRequestCollector {
     requests: AuxRequests,
+    max_manifest_height: u64,
 }
```
prajwolrg, line 27: I wonder if we should also have `min_manifest_height`. It can be the `genesis_height`.

prajwolrg (changes requested):
Nice catch!

crates/asm/common/src/aux/collector.rs
irnb: In practice, this shouldn’t be a concern because the subprotocols should select the start range based on its internal state. That said, it’s still good to add an extra check for future subprotocols use.

Also, using `offset()` method from `AsmHistoryAccumulatorState`, we have access to `genesis_height`.

I’ll address this.

crates/asm/common/src/aux/collector.rs
irnb: [Fixed](https://github.com/alpenlabs/alpen/pull/1363/changes/4a78491ff50ef31d6aea2ddde22f8afc272786da)

crates/asm/common/src/aux/collector.rs
irnb: [Fixed](https://github.com/alpenlabs/alpen/pull/1363/changes/4a78491ff50ef31d6aea2ddde22f8afc272786da)

PR #1365, opened 2026-02-16
Str 2142 correctly fetch prev epoch commitment

bewakes:
## Description
Fixes current issue with previous epoch always being null. The following needed to be done for the fix:
- Block assembly used `None` manifest container when manifest entries were empty. But this did not respect the sealing policy when L1 was slower because of which blocks were not marked as terminal even when every N slots sealing policy was used.
- Chain worker did not send epoch commitment info to listeners when the epoch ended. Due to this checkpoint builder was not able to save epoch summary to the database.
- Fetch previous epoch commitment from database instead of using Null epoch commitment.
- Also add functional test to check for epoch progression.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/bin/strata/src/rpc/node.rs
+++ b/bin/strata/src/rpc/node.rs
@@ -141,7 +141,7 @@ impl OLClientRpcServer for OLRpcServer {
                 .await
                 .map_err(db_error)?
                 .ok_or_else(|| {
-                    not_found_error(format!("No epoch commitment found for epoch {epoch}"))
+                    not_found_error(format!("No epoch commitment found for epoch {}", epoch - 1))
                 })?
         } else {
             EpochCommitment::null()


functional-tests-new/tests/strata/test_epoch_progression.py
```diff
@@ -0,0 +1,48 @@
...
+        cur_epoch = initial_status["confirmed"]
+
+        epochs_to_check = 3
+
+        for _ in range(1, epochs_to_check + 1):
+            epoch = wait_until_with_value(
+                lambda: strata.get_sync_status(rpc)["confirmed"],
+                lambda v, cur_epoch=cur_epoch: v is not None and v["epoch"] > cur_epoch["epoch"],
+                timeout=10,
+                error_with="Epoch not progressing",
```
storopoli, line 45: This test only checks that `confirmed.epoch` increases, which can still pass when `last_slot` stays `0` and `last_blkid` stays zero due to fallback behavior. Add assertions that `confirmed.last_slot > 0` and `confirmed.last_blkid` is non-zero after `epoch` progression so the regression is actually covered.

This can occur because in

https://github.com/alpenlabs/alpen/blob/ee7c2ccd000b97c3e2e527f36003f74cab75a826/crates/consensus-logic/src/fcm/service.rs#L159-L166

can still fall back to:
- `last_slot = 0`
- zero `last_blkid`

The test would still pass as long as `epoch` increments.

crates/chain-worker-new/src/state.rs
```diff
@@ -165,8 +165,20 @@ impl ChainWorkerServiceState {
...
+            let epoch = EpochCommitment::new(
+                block.header().epoch(),
+                block.header().slot(),
+                block.header().compute_blkid(),
+            );
+
+            debug!(%epoch, "Sending epoch commitment to listeners");
+            if let Err(e) = self.ctx.epoch_summary_sender().send(Some(epoch)) {
+                warn!(%e, "Could not send epoch summary to receivers");
+            }
```
storopoli: `handle_complete_epoch` calls `store_summary` in

https://github.com/alpenlabs/alpen/blob/ee7c2ccd000b97c3e2e527f36003f74cab75a826/crates/chain-worker-new/src/state.rs#L346

which already sends the `epoch_summary` update in

https://github.com/alpenlabs/alpen/blob/ee7c2ccd000b97c3e2e527f36003f74cab75a826/crates/chain-worker-new/src/context.rs#L203-L209

So listeners get two updates for one epoch completion.

storopoli: This is not urgent.

purusang (dismissed):
LGTM

barakshani (changes requested):
new tests fail, please check why

bin/strata/src/rpc/node.rs
```diff
@@ -141,21 +135,20 @@ impl OLClientRpcServer for OLRpcServer {
...
+            self.storage
+                .ol_checkpoint()
+                .get_canonical_epoch_commitment_at_async((epoch - 1) as u64)
                 .await
-                .ok()
-                .and_then(|c| c.first().copied());
-            prev_commitments.unwrap_or_else(EpochCommitment::null)
+                .map_err(db_error)?
+                .ok_or_else(|| {
+                    not_found_error(format!("No epoch commitment found for epoch {epoch}"))
```
storopoli:
```suggestion
                    not_found_error(format!("No epoch commitment found for epoch {epoch-1}"))
```

PR #1375, opened 2026-02-18
feat(asm)!: use AsmParams to configure ASM subprotocols end-to-end

prajwolrg:
## Description

Replace the old `RollupParams`-based ASM configuration with a dedicated `AsmParams` structure that is threaded from CLI through NodeContext to the ASM worker and spec.

- Add `gen-asm-params` subcommand to datatool for generating `AsmParams` from operator keys, deposit amounts, and other inputs
- Thread `AsmParams` from CLI args through `NodeContext` to the ASM worker
- Build AsmSpec directly from AsmParams, removing the old `RollupParams` dependency in the ASM layer
- Load both v0 and new version of the checkpoint subprotocol
- Update the test harness to use a builder pattern with optional subprotocol param overrides
- Generate ASM params in functional tests
- Fix bridge param Arbitrary impl to generate at least one operator
- Generate random keypair for admin in tests

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/functional-tests-new/common/config/params.py
+++ b/functional-tests-new/common/config/params.py
@@ -157,6 +157,7 @@ class AdminSubprotocolConfig:
     strata_administrator: ThresholdConfig
     strata_sequencer_manager: ThresholdConfig
     confirmation_depth: int = 144
+    max_seqno_gap: int = 100
 
 
 @dataclass


storopoli (commented):
`AsmParams` is randomized in harness setup, creating nondeterministic bridge/admin/checkpoint config.

tests/harness/test_harness.rs
```diff
@@ -127,20 +131,26 @@ impl AsmTestHarness {
         let mut params = strata_test_utils_l2::gen_params().rollup;
         params.network = Network::Regtest;
         let genesis_view = get_genesis_l1_view(&client, &genesis_hash).await?;
-        params.genesis_l1_view = genesis_view;
+        params.genesis_l1_view = genesis_view.clone();
         let params = Arc::new(params);
 
-        // 4. Create worker context
+        // 4. Build AsmParams via arbitrary, then override magic and l1_view
+        let mut asm_params: AsmParams = ArbitraryGenerator::new().generate();
```
storopoli: Using `ArbitraryGenerator` here means only `magic`/`l1_view` are deterministic, while bridge/admin/checkpoint config remains random. This can create invalid runtime state (for example empty bridge operators) and cause panics/timeouts in integration tests. Please construct deterministic `AsmParams` from fixture values instead.

crates/asm/spec/src/lib.rs
```diff
@@ -4,115 +4,64 @@
...
-            })
-            .collect();
-        let threshold = NonZero::new(1).unwrap();
-        let admin_config = ThresholdConfig::try_new(admin_pubkeys.clone(), threshold).unwrap();
-
-        let admin_params = AdministrationSubprotoParams::new(
-            admin_config.clone(),
-            admin_config,
-            2, // confirmation_depth: updates activate 2 blocks after submission
+        assert!(
```
storopoli: These `assert!` checks panic on malformed `asm_params` input. Since ASM params are now externally provided, this should return a structured error (for example a dedicated `AsmSpecError`/`InitError::InvalidAsmParams`) so startup can fail gracefully with actionable diagnostics instead of aborting the process.

tests/harness/test_harness.rs
prajwolrg: I think it's better to use `Arbitrary` instead of using some fixed values. You were right to point about the operators needing to be non-empty. I've updated the implementation to support for that. 

I'm also curious to know your thought on whether we should use something like [vec1](https://crates.io/crates/vec1) to enforce that there is at least an operator in the params, or manually create a new type that enforces the invariant.

crates/asm/spec/src/lib.rs
prajwolrg: I'll handle this in a separate PR. For now, this is irrelevant.

bewakes (dismissed):
LGTM

storopoli (dismissed):
ACK 7ee718e6fc3e52abdcf6752ec70902868ea777cd

PR #1377, opened 2026-02-18
feat(functional-tests): add OL mock deposit+withdrawal test (STR-2020)

voidash:
## Description

Add an end-to-end functional test that verifies deposit and withdrawal flows through the OL without a real EE, using the debug subprotocol to inject mock deposits via Bitcoin transactions.

### What this PR does

**Rust changes:**
- Add `debug-asm` feature flag that conditionally switches ASM worker from `StrataAsmSpec` to `DebugAsmSpec`, enabling the debug subprotocol (ID 255)
- Add `create-mock-deposit` CLI command to `strata-test-cli` — builds a Bitcoin tx with SPS-50 tagged OP_RETURN that injects `DepositIntentLogData` via the debug subprotocol's MockAsmLog mechanism
- Add `build-snark-withdrawal` CLI command — constructs `RpcOLTransaction` JSON with SSZ-encoded `SnarkAccountUpdate` containing a withdrawal `OutputMessage`

**Python test framework changes:**
- Add `OLIsolatedEnvConfig` — starts Bitcoin + Strata (OL only, no EE) with configurable genesis snark accounts
- Add `common/test_cli.py` — subprocess wrapper for `strata-test-cli` commands
- Add `EpochSealingConfig` support (slots_per_epoch=5 for fast terminal blocks in tests)
- Fetch real `GenesisL1View` from bitcoind (target, timestamps) instead of hardcoded values

**Test flow (`test_mock_withdrawal`):**
1. Start bitcoind (regtest) + strata sequencer with a genesis snark account
2. Inject 20 BTC deposit via `create-mock-deposit` (debug subprotocol)
3. Mine Bitcoin blocks, wait for OL terminal block to process L1 manifest
4. Assert balance == 20 BTC
5. Submit 1 BTC withdrawal via `build-snark-withdrawal` + `strata_submitTransaction`
6. Assert balance == 19 BTC

### Type of Change

- [x] New or updated tests

## Notes to Reviewers

The `debug-asm` feature is only enabled during functional tests (`run_tests.sh` passes `-F debug-asm`). Production builds are unaffected.

The two new `strata-test-cli` commands replace previous Python encoding utilities (SSZ, strata-codec) with authoritative Rust implementations, eliminating encoding parity bugs.

- [x] Yes

STR-2020

## Checklist

## AI Assistance Notice

AI was used to assist with code review, test hardening fixes, and writing this PR description.

## Related Issues

STR-2020

### DIFF
--- a/bin/strata-test-cli/src/mock_ee/withdrawal.rs
+++ b/bin/strata-test-cli/src/mock_ee/withdrawal.rs
@@ -69,7 +69,7 @@ pub(crate) fn build_snark_withdrawal_json(
             "update_operation_encoded": ssz_hex,
             "update_proof": ""
         },
-        "attachments": {
+        "constraints": {
             "min_slot": null,
             "max_slot": null
         }


delbonis (dismissed):

functional-tests-new/envconfigs/ol_isolated.py
```diff
@@ -0,0 +1,151 @@
+"""
```
bewakes: There's already an env called `StrataEnvConfig` that only has strata and bitcoin. This new one most likely won't be necessary.

functional-tests-new/tests/ol_isolated/test_mock_withdrawal.py
```diff
@@ -0,0 +1,175 @@
...
+        addr = btc_rpc.proxy.getnewaddress()
+        btc_rpc.proxy.generatetoaddress(8, addr)
+
+        # Step 3: Wait for OL to reach a terminal block (epoch boundary).
+        # L1 manifests are only processed during terminal blocks. With
+        # slots_per_epoch=5 the first post-genesis terminal block is at slot 5.
+        # Wait for enough blocks to guarantee we cross at least one terminal
+        # boundary after the ASM has processed the new Bitcoin blocks.
+        logger.info("Waiting for OL to reach terminal block and process deposit...")
+        strata.wait_for_additional_blocks(10, rpc, timeout_per_block=15)
```
bewakes: Let's not use hardcoded `10`. This should be calculated from actual `slots_per_epoch` derived from rollup params.

functional-tests-new/tests/ol_isolated/test_mock_withdrawal.py
```diff
@@ -0,0 +1,175 @@
...
+        # Step 6: Submit withdrawal
+        logger.info("Submitting withdrawal transaction...")
+        tx_id = rpc.strata_submitTransaction(tx_json)
+        logger.info(f"Withdrawal submitted, ID: {tx_id}")
+
+        # Step 7: Wait for OL blocks to process withdrawal
+        strata.wait_for_additional_blocks(2, rpc, timeout_per_block=15)
+
+        # Step 8: Verify final balance
+        final_balance = get_account_balance(rpc, account_id_hex)
```
irnb, line 163: I think this test should be failing.

Ref: 
https://github.com/alpenlabs/alpen/pull/1383#discussion_r2871001316

voidash: the test passes because the mock intentionally bypasses bridge-v1 because of Debug subprotocol but i agree the type ID mismatch between `DepositLog`  and `DepositIntentLogData` needs to be resolved. and  I think that's a separate issue from the test itself.

If OL should match on `DEPOSIT_LOG_TYPE_ID` and process DepositLog, then we can rewrite this test to match the canonical type we end up choosing.

crates/asm/worker/src/state.rs
```diff
@@ -32,13 +35,19 @@ pub struct AsmWorkerServiceState<W> {
...
+    asm_spec: DebugAsmSpec,
 }
 
 impl<W: WorkerContext + Send + Sync + 'static> AsmWorkerServiceState<W> {
     /// A new (uninitialized) instance of the service state.
     pub fn new(context: W, asm_params: Arc<AsmParams>) -> Self {
+        #[cfg(not(feature = "debug-asm"))]
         let asm_spec = StrataAsmSpec::from_asm_params(&asm_params);
+        #[cfg(feature = "debug-asm")]
+        let asm_spec = DebugAsmSpec::from_asm_params(&asm_params);
```
delbonis: I don't think this should be baked in at this level.  I would have made it so that building with the feature flag just enables the debug spec, but it still has to be specified to be used in the network params.  It would be valid for the ASM worker to use `Box<dyn _>` here.

functional-tests-new/tests/ol_isolated/test_mock_withdrawal.py
delbonis: oh no!  yeah we should resolve this, maybe in this PR?  both sides should use the same basic type instead of having two separate defs of it

bewakes (commented):
The func tests are failing. The rest LGTM.

storopoli (dismissed):
ACK a527fdaa3555167423a4a95552211e7c2546f110

PR #1380, opened 2026-02-19
STR-1885-wire-da-refs-to-ledger-refs

irnb:
## Description

Connects `Vec<L1DaBlockRef>` (collected during the batch lifecycle) to `LedgerRefs` in snark account updates, providing the public parameters required for EE DA verification in the EE proof statement. 
Previously, the update submitter discarded DA block references and sent `LedgerRefs::new_empty()` to the OL.

- Commit 1 (EE side): feat(ee): wire DA block refs into LedgerRefs
- Commit 2 (OL side): feat(ol): convert L1 height to MMR index for ledger ref proofs
- Commit 3 and 4: fixes some things found in the review process
- Commit 5: authenticates claim hash against ASM manifest commitment and wires everything into EE and OL sequencer/nodes

#### Follow-up fixes in this PR

- Deduplicate `AccumulatorClaim`s by L1 height before proof generation (multiple DA txs can land in one L1 block).
- Authenticate `AccumulatorClaim.entry_hash` against ASM manifest commitments end-to-end:
  - EE now fetches canonical ASM manifest hash by L1 height from OL RPC and uses it in `LedgerRefs`.
  - OL block assembly verifies claim hash matches the canonical ASM MMR leaf.
  - Snark account verification enforces claim hash/proof hash consistency against the ASM MMR.

#### **Challenge: height vs MMR index**

The EE only knows L1 block heights, but `AccumulatorClaim.idx`, which becomes part of the `SnarkAccountUpdate` proven claim is verified against the manifests MMR, which is 0 indexed. The EE should not be responsible for knowing MMR internals.

#### Current solution

Following the ASM aux accumulator pattern (`AsmHistoryAccumulatorState::offset` in `crates/asm/common/src/mmr.rs`):

- Claims carry L1 block heights,  the external semantic the EE naturally knows,
- The OL stores `manifests_mmr_offset = genesis_height + 1` alongside the MMR in `EpochalState`
- Height-to-MMR-index conversion (`height - offset`) happens at the point of MMR access, in both proof generation (block assembly) and verification (STF)

This keeps the EE free of MMR indexing internals while still binding claim hashes to canonical ASM manifest commitments.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [x]  Yes
- [ ]  No

[Data Availability: Data Flow & Plan](https://www.notion.so/Data-Availability-Data-Flow-Plan-2ac901ba000f803eb42bfaf996a0616b)

## Checklist

## Related Issues

STR-1885

### DIFF
--- a/crates/ledger-types/src/state_accessor.rs
+++ b/crates/ledger-types/src/state_accessor.rs
@@ -64,16 +64,6 @@ pub trait IStateAccessor {
     /// Gets the ASM manifests MMR for ledger reference verification.
     fn asm_manifests_mmr(&self) -> &Mmr64;
 
-    /// Gets the first L1 block height represented in the manifests MMR.
-    fn asm_manifests_mmr_start_height(&self) -> L1Height;
-
-    /// Resolves an L1 block height into the corresponding ASM manifests MMR leaf index.
-    ///
-    /// Returns `None` when the height is before the MMR start height.
-    fn asm_manifest_mmr_index_for_height(&self, height: L1Height) -> Option<u64> {
-        (height as u64).checked_sub(self.asm_manifests_mmr_start_height() as u64)
-    }
-
     // ===== Account methods =====
 
     /// Checks if an account exists.
@@ -112,4 +102,31 @@ pub trait IStateAccessor {
     /// Computes the full state root, using whatever things we've updated.
     // TODO maybe don't use `AcctResult`, actually convert all/most of these to use a new error type
     fn compute_state_root(&self) -> AcctResult<Buf32>;
+}
+
+/// Resolves the first L1 block height represented by the ASM manifests MMR.
+///
+/// This is derived from canonical state as:
+/// `mmr_start_height = last_l1_height + 1 - (manifests_mmr_entries - 1)`.
+///
+/// For an empty MMR, this reduces to `last_l1_height + 1`.
+pub fn asm_manifests_mmr_start_height(state: &impl IStateAccessor) -> Option<L1Height> {
+    let last_l1_height_u64 = state.last_l1_height() as u64;
+    let num_entries = state.asm_manifests_mmr().num_entries();
+    let last_entry_idx = num_entries.saturating_sub(1);
+    let start_height_u64 = last_l1_height_u64
+        .checked_add(1)?
+        .checked_sub(last_entry_idx)?;
+    start_height_u64.try_into().ok()
+}
+
+/// Resolves an L1 block height into the corresponding ASM manifests MMR leaf index.
+///
+/// Returns `None` when the height is before the MMR start height.
+pub fn asm_manifest_mmr_index_for_height(
+    state: &impl IStateAccessor,
+    height: L1Height,
+) -> Option<u64> {
+    let start_height_u64 = asm_manifests_mmr_start_height(state)? as u64;
+    (height as u64).checked_sub(start_height_u64)
 }
--- a/crates/snark-acct-sys/src/verification.rs
+++ b/crates/snark-acct-sys/src/verification.rs
@@ -3,7 +3,7 @@ use strata_acct_types::{
     AccountId, AcctError, AcctResult, BitcoinAmount, Mmr64, StrataHasher, tree_hash::TreeHash,
 };
 use strata_identifiers::L1Height;
-use strata_ledger_types::{ISnarkAccountState, IStateAccessor};
+use strata_ledger_types::{ISnarkAccountState, IStateAccessor, asm_manifest_mmr_index_for_height};
 use strata_merkle::MerkleProof;
 use strata_snark_acct_types::{
     LedgerRefProofs, LedgerRefs, MessageEntry, MessageEntryProof, ProofState, SnarkAccountUpdate,
@@ -97,8 +97,8 @@ pub fn verify_message_index(
 /// Verifies the ledger ref proofs against the provided asm mmr for an account.
 ///
 /// The operation carries manifest commitment references keyed by L1 height
-/// (`AccumulatorClaim.idx`). The state accessor resolves those heights into
-/// ASM manifest MMR indices for proof verification.
+/// (`AccumulatorClaim.idx`). The verifier resolves those heights into ASM
+/// manifest MMR indices from canonical state view for proof verification.
 fn verify_ledger_refs(
     target: AccountId,
     state_accessor: &impl IStateAccessor,
@@ -124,11 +124,12 @@ fn verify_ledger_refs(
                     account_id: target,
                     ref_idx: manifest_ref.idx(),
                 })?;
-        let mmr_idx = state_accessor
-            .asm_manifest_mmr_index_for_height(l1_height)
-            .ok_or_else(|| AcctError::InvalidLedgerReference {
-                account_id: target,
-                ref_idx: manifest_ref.idx(),
+        let mmr_idx =
+            asm_manifest_mmr_index_for_height(state_accessor, l1_height).ok_or_else(|| {
+                AcctError::InvalidLedgerReference {
+                    account_id: target,
+                    ref_idx: manifest_ref.idx(),
+                }
             })?;
         if manifest_ref_proof.entry_idx() != mmr_idx {
             return Err(AcctError::InvalidLedgerReference {
--- a/crates/ledger-types/src/lib.rs
+++ b/crates/ledger-types/src/lib.rs
@@ -38,7 +38,9 @@ pub use account::{
     ISnarkAccountStateConstructible, ISnarkAccountStateMut, NewAccountData,
 };
 pub use coin::Coin;
-pub use state_accessor::IStateAccessor;
+pub use state_accessor::{
+    IStateAccessor, asm_manifest_mmr_index_for_height, asm_manifests_mmr_start_height,
+};
 // transitional crap
 pub use strata_asm_manifest_types::AsmManifest;
 pub use strata_identifiers::{EpochCommitment, L1BlockId, L1Height};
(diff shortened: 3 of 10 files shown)


crates/alpen-ee/sequencer/src/update_submitter/update_builder.rs
```diff
@@ -96,12 +100,23 @@ fn build_update_operation(
     let extra_data = UpdateExtraData::new(new_tip_blkid, processed_inputs as u32, 0);
     let extra_data_buf = encode_to_vec(&extra_data)?;
 
-    // 4. Build update operation
+    // 4. Build ledger refs from DA block references (idx = L1 block height)
+    let mut l1_header_refs: Vec<AccumulatorClaim> = da_refs
+        .iter()
+        .map(|da_ref| {
+            AccumulatorClaim::new(da_ref.block.height_u64(), *da_ref.block.blkid().as_ref())
```
storopoli: `AccumulatorClaim.entry_hash` here is set to the L1 block ID, but OL verifies ledger refs against the ASM manifests MMR where leaves are manifest hashes. That mismatch will cause `L1HeaderHashMismatch` once non-empty ledger refs are wired through. This is a correctness blocker; the claim hash needs to match the exact leaf hash committed in `MmrId::Asm`.

crates/alpen-ee/sequencer/src/update_submitter/update_builder.rs
```diff
@@ -96,12 +100,23 @@ fn build_update_operation(
...
+        })
+        .collect();
+    // Dedup by height — multiple DA txns may land in the same L1 block
+    l1_header_refs.dedup_by_key(|c| c.idx());
```
storopoli, line 148: `dedup_by_key` only removes adjacent duplicates. If `da_refs` contains the same height in non-consecutive positions, duplicates remain. Consider sorting by `idx` before dedup or collecting via a map/set keyed by height so deduplication is deterministic and complete.

storopoli (changes requested):
Nice work!

storopoli (approved):
ACK 647c069d11ee9e792e56dd4b926f5886ef82ed7d

crates/alpen-ee/sequencer/src/update_submitter/update_builder.rs
prajwolrg: I don't think OL should have ASM manifests MMR. I know this is the status quo, but we should likely change that. It makes sense for OL to maintain MMR of `L1BlockId` and/or `WtxidsRoot`. The logs from ASM manifest is meant to be consumed by OL. EE should not be aware of ASM Manifest at all.

crates/ledger-types/src/state_accessor.rs
```diff
@@ -64,6 +64,9 @@ pub trait IStateAccessor {
     /// Gets the ASM manifests MMR for ledger reference verification.
     fn asm_manifests_mmr(&self) -> &Mmr64;
 
+    /// Gets the offset for mapping L1 block heights to manifests MMR leaf indices.
+    fn asm_manifests_mmr_offset(&self) -> u64;
```
prajwolrg: I'd like all the API should be based on the `L1Height` instead of MMR indices. The conversion should happen internally.

crates/alpen-ee/sequencer/src/update_submitter/update_builder.rs
storopoli: Agree on the layering concern. For this PR we are keeping the current correctness fix (claim hash must be authenticated against the canonical ASM commitment) to close STR-1885 safely. We can follow up with a separate refactor to expose a neutral height-based commitment API so EE does not depend on ASM manifest semantics directly.

storopoli:
> I don't think OL should have ASM manifests MMR. I know this is the status quo, but we should likely change that. It makes sense for OL to maintain MMR of L1BlockId and/or WtxidsRoot. The logs from ASM manifest is meant to be consumed by OL. EE should not be aware of ASM Manifest at all.


Check f4b8a2fb0eb5ce1add2ab9506ba57e6bfc6564b2

crates/ledger-types/src/state_accessor.rs
storopoli: Yes, I did a workaround in 1520c2b9a0827ebc1063fc47532331abc88941ae

with 

```diff
-    /// Gets the offset for mapping L1 block heights to manifests MMR leaf indices.
-    fn asm_manifests_mmr_offset(&self) -> u64;
+    /// Gets the first L1 block height represented in the manifests MMR.
+    fn asm_manifests_mmr_start_height(&self) -> L1Height;
+
+    /// Resolves an L1 block height into the corresponding ASM manifests MMR leaf index.
+    ///
+    /// Returns `None` when the height is before the MMR start height.
+    fn asm_manifest_mmr_index_for_height(&self, height: L1Height) -> Option<u64> {
+        (height as u64).checked_sub(self.asm_manifests_mmr_start_height() as u64)
+    }
```

I still need somehow MMR indices but these are more accessory and primarily we're using `L1Height`.

crates/ol/block-assembly/src/test_utils.rs
```diff
@@ -644,7 +648,7 @@ pub(crate) const DEFAULT_ACCOUNT_BALANCE: u64 = 100_000_000_000;
 
 /// Info about a manifest in the MMR - links index and hash together.
 pub(crate) struct ManifestInfo {
-    pub index: u64,
+    pub height: u64,
     pub hash: Hash,
 }
```
delbonis: Maybe `ManifestRef` or even `ManifestCommitment` might be better here.

This could actually be more generally useful so maybe we should move this to the identifiers crate?  Just don't mess up the dependency hierarchy.

crates/snark-acct-sys/src/verification.rs
```diff
@@ -112,13 +117,37 @@ fn verify_ledger_refs(
...
-        let generic_proof = MerkleProof::from_cohashes(cohashes, lref.idx());
-        if !generic_mmr.verify::<StrataHasher>(&generic_proof, hash.as_ref()) {
+        let generic_proof = MerkleProof::from_cohashes(cohashes, mmr_idx);
+        if !generic_mmr.verify::<StrataHasher>(&generic_proof, lref.entry_hash().as_ref()) {
             return Err(AcctError::InvalidLedgerReference {
                 account_id: target,
-                ref_idx: proof.entry_idx(),
+                ref_idx: lref.idx(),
             });
         }
```
delbonis: Maybe we can move some of this more generic merkle proof checking logic out into its own function so we can use it with other ledger refs in the future?  Could merge this with the inbox mmr proof.

crates/snark-acct-sys/src/verification.rs
delbonis: This seems like it's conflating "manifest hashes" with "L1 headers" in this use of the terms, not sure if this was my mistake when defining the types.  This should be made more consistent to focus on the manifests specifically, since the L1 blkids are part of those.

crates/ol/block-assembly/src/test_utils.rs
storopoli: Check 871726c8cc126d9f0c7e9ad48d127931925c1ba4 I've refactored and renamed to `ManifestCommitment`

crates/snark-acct-sys/src/verification.rs
storopoli: Corrected in 871726c8cc126d9f0c7e9ad48d127931925c1ba4

crates/snark-acct-sys/src/verification.rs
storopoli: I've refactored in 871726c8cc126d9f0c7e9ad48d127931925c1ba4 and added a test.

crates/ledger-types/src/state_accessor.rs
bewakes: I don't think the state accessor needs this exposed. The block assembly can compute mmr index using the genesis L1 height info(I presume the formula is `mmr_index = l1_ref_height - genesis_l1_height`). Once block assembly correctly computes this and inserts proofs in  an snark update, it should suffice OL stf to verify and execute the update.

storopoli: For this PR, though, block assembly computing the index is necessary but not sufficient. STF is the trust boundary, so it must independently re-derive the expected MMR index from the claimed `L1Height` and reject mismatches (`proof.entry_idx != resolved_idx`).

That’s why the conversion currently lives behind `IStateAccessor`: it centralizes height->index resolution from canonical state (`manifests_mmr_start_height`) instead of duplicating offset arithmetic at call sites.

So the accessor method is not to expose MMR details; it exists so STF verification can enforce claim/proof consistency against canonical state.

bewakes: I understand that this does centralize the computation. However, this has nothing to do with the state, this is something derivable from OL params. We can thus expose this centralized conversion function outside the state accessor.
Having this in state accessor without touching any aspects of state means that we need to just have this method implmented in every IStateAccessor implementor.

The only reason I am averse to this interface is that it has nothing to do with the OL State and the logic can be centralized elsewhere. If there is some IStateAccessor implementation that would be different than this "static" computation logic, then it might make sense to have it in state accessor but otherwise I don't think so.

storopoli: Ok.
Moved height→MMR conversion out of `IStateAccessor` into shared helpers in `ledger-types` (`crates/ledger-types/src/state_accessor.rs`):
* `asm_manifests_mmr_start_height(...)`
* `asm_manifest_mmr_index_for_height(...)`

Check 3121635c81f6414e9cedc45f760770576dc8a866

PR #1383, opened 2026-02-20
STR 2203 Functional tests for OL checkpointing

krsnapaudel:
## Description

We have two tests now.
1. Checkpoint created (checked by signing duties) for epoch >= 1.
2. Checkpoint finalized for epoch 1.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

NOTES:
1. OL checkpoint builder does not create a checkpoint for epoch 0. Related to [STR-2365](https://alpenlabs.atlassian.net/browse/STR-2365).
2. Relying on OL STF handling checkpoint logs and updating ASM recorded epoch as confirmed epoch.
4. Do we need a checkpoint finalized test? Then who should set epoch finalized? Are we skipping CSM?

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-2203](https://alpenlabs.atlassian.net/browse/STR-2203)

[STR-2365]: https://alpenlabs.atlassian.net/browse/STR-2365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[STR-2203]: https://alpenlabs.atlassian.net/browse/STR-2203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/functional-tests-new/tests/strata/test_epoch_progression.py
+++ b/functional-tests-new/tests/strata/test_epoch_progression.py
@@ -13,38 +13,32 @@ logger = logging.getLogger(__name__)
 
 @flexitest.register
 class TestSequencerEpochProgression(StrataNodeTest):
-    """Test that sequencer is correctly progressing epoch."""
-
-    # NOTE: This test should be covered when we actually do more integrated
-    # test with alpen-client because alpen-client also needs epoch progression to work properly
+    """Test that sequencer is correctly progressing epochs."""
 
     def __init__(self, ctx: flexitest.InitContext):
         ctx.set_env("checkpoint")
 
     def main(self, ctx):
-        # Get sequencer service
         strata = self.get_service(ServiceType.Strata)
 
-        # Wait for RPC to be ready
         logger.info("Waiting for Strata RPC to be ready...")
         rpc = strata.wait_for_rpc_ready(timeout=10)
 
-        # Get initial sync status
         initial_status = strata.get_sync_status(rpc)
         prev_epoch = initial_status["parent"]
-        logger.info(f"initial prev epoch {prev_epoch}")
+        logger.info("initial parent epoch %s", prev_epoch["epoch"])
         assert prev_epoch["last_blkid"] != "00" * 32
 
         epochs_to_check = 3
 
-        for _ in range(1, epochs_to_check + 1):
+        for _ in range(epochs_to_check):
             epoch = wait_until_with_value(
                 lambda: strata.get_sync_status(rpc)["parent"],
                 lambda v, cur_epoch=prev_epoch: v is not None and v["epoch"] > cur_epoch["epoch"],
                 timeout=10,
-                error_with="Epoch not progressing",
+                error_with="Parent epoch not progressing",
             )
-            logger.info(f"cur prev epoch {epoch}")
+            logger.info("parent epoch advanced to %s", epoch["epoch"])
             assert epoch["last_blkid"] != "00" * 32
             prev_epoch = epoch
 
--- a/functional-tests-new/tests/checkpoint/test_checkpoint_finalized.py
+++ b/functional-tests-new/tests/checkpoint/test_checkpoint_finalized.py
@@ -1,0 +1,55 @@
+"""
+Verify that signed checkpoints are posted to L1 and the ASM finalizes them.
+"""
+
+import logging
+
+import flexitest
+
+from common.base_test import StrataNodeTest
+from common.config import ServiceType
+from tests.checkpoint.helpers import mine_until_finalized_epoch
+
+logger = logging.getLogger(__name__)
+
+
+@flexitest.register
+class TestCheckpointFinalized(StrataNodeTest):
+    """Signed checkpoint posted to L1, ASM finalizes it, finalized epoch advances."""
+
+    def __init__(self, ctx: flexitest.InitContext):
+        ctx.set_env("checkpoint")
+
+    def main(self, ctx):
+        bitcoin = self.get_service(ServiceType.Bitcoin)
+        strata = self.get_service(ServiceType.Strata)
+
+        # Wait for RPC to be ready
+        logger.info("Waiting for Strata RPC to be ready...")
+        strata_rpc = strata.wait_for_rpc_ready(timeout=20)
+
+        btc_rpc = bitcoin.create_rpc()
+        mine_addr = btc_rpc.proxy.getnewaddress()
+
+        # Get initial sync status
+        initial_status = strata.get_sync_status(strata_rpc)
+        logger.info(
+            "initial finalized cursor epoch %s (genesis baseline)",
+            initial_status["finalized"]["epoch"],
+        )
+
+        epochs_to_check = 3
+
+        for target_epoch in range(1, epochs_to_check + 1):
+            epoch = mine_until_finalized_epoch(
+                btc_rpc=btc_rpc,
+                strata=strata,
+                strata_rpc=strata_rpc,
+                mine_addr=mine_addr,
+                target_epoch=target_epoch,
+                timeout=120,
+                step=1.0,
+            )
+            logger.info("finalized epoch advanced to %s", epoch["epoch"])
+
+        return True


krsnapaudel:
> NACK
> 
> The issue is that modelling it that way (default state being `verified_tip.epoch = 0`) is just wrong. If we haven't seen the genesis epoch checkpoint yet then it's incorrect to say that we've verified epoch 0, as others recognized. This is a distinction between modeling it as `u64` vs as a `Option<u64>`. We should start up in some pre-genesis state.
> 
> It'd also be expanding the scope of this ticket substantially to propagate a significant epoch bookkeeping change up something that would impact specs.

Yes same. While the fix to create OL checkpoint starting from epoch 1 is easy, seeing `verified_tip.epoch` set to 0 without verifying the checkpoint seems wrong.

delbonis (dismissed):
so this is good then?  I haven't done a deep review

crates/status/src/chain.rs
```diff
@@ -19,6 +19,9 @@ pub struct ChainSyncStatus {
     /// The previous epoch (ie. epoch most recently completed).
     pub prev_epoch: EpochCommitment,
 
+    /// The last epoch confirmed by the ASM on L1 (ie. checkpoint posted and accepted).
+    pub confirmed_epoch: EpochCommitment,
+
     /// The finalized epoch, ie what's witnessed on L1.
     pub finalized_epoch: EpochCommitment,
```
barakshani: how do these two differ?

krsnapaudel: My understanding is mined in L1 block and processed by ASM is confirmed. Finalization is not clear to me. So I don't have a test for it.

delbonis: Finalization is when we locally see the checkpoint as having been buried on L1 by n blocks.

barakshani: what do you mean by locally? ASM is "locally" as well.
So you're saying that "confirmed" is with n=0? Do we really need both?

functional-tests-new/common/wait.py
```diff
@@ -54,9 +59,42 @@ def wait_until_with_value(
...
-            ety = type(e)
-            logger.warning(f"caught exception {ety}, will still wait for timeout: {e}")
+        except _RETRYABLE as e:
+            logger.warning(f"caught {type(e).__name__}, will still wait for timeout: {e}")
 
         time.sleep(step)
     raise AssertionError(error_with)
+
+
+def wait_for_confirmed_epoch(
```
bewakes: Can these functions be more localized? Maybe in `StrataService` or in `checkpoint/helpers`. This module mostly contains the basic waiting primitives.

functional-tests-new/factories/strata.py
```diff
@@ -70,16 +77,26 @@ def create_node(
...
 
-        # Generate rollup params via datatool.
+        # Generate rollup params via datatool (also produces keys used below).
         params_data = generate_rollup_params(datadir, bconfig, genesis_l1_height)
 
+        # Override with caller-provided rollup params if given, always applying genesis L1.
+        if rollup_params is not None:
+            with open(params_data.params_path, "w") as f:
+                f.write(rollup_params.with_genesis_l1(genesis_l1).as_json_string())
+
```
bewakes: The `generate_rollup_params` also generates sequencer keys and has key path as one of the return values. There might be a gap here that the sequencer keys present in the passed rollup params may not align with what's returned from the `generate_rollup_params`.

functional-tests-new/tests/checkpoint/test_checkpoint_created.py
```diff
@@ -0,0 +1,52 @@
...
+        # Drive L1 forward so OL can produce blocks and complete epoch 1.
+        btc_rpc.proxy.generatetoaddress(5, addr)
+
+        # Wait for OL to reach the terminal slot of epoch 1.
+        epoch_sealing = strata.props["epoch_sealing"]
+        assert epoch_sealing is not None, "checkpoint env must set epoch_sealing"
+        assert epoch_sealing["policy"] == "FixedSlot", "test assumes FixedSlot policy"
+        slots_per_epoch = epoch_sealing["slots_per_epoch"]
+        assert slots_per_epoch is not None and slots_per_epoch > 0
+        epoch1_terminal_slot = (2 * slots_per_epoch) - 1
```
bewakes: Shouldn't this be just `slots_per_epoch`? given that we start with slot 1 and epoch 1. If `slots_per_epoch` is 4, then the first epoch ends at slot 4 right and not 2 * 4 -1 = 7. Am I missing something?

functional-tests-new/tests/checkpoint/test_checkpoint_confirmed.py
```diff
@@ -0,0 +1,74 @@
...
+        assert slots_per_epoch is not None and slots_per_epoch > 0
+        epoch1_terminal_slot = (2 * slots_per_epoch) - 1
+        strata.wait_for_block_height(
+            epoch1_terminal_slot, strata_rpc, timeout=120, poll_interval=0.5
+        )
+
+        # 1. Wait for checkpoint builder to create at least one entry, then
+        #    wait for the sequencer to sign all pending checkpoints.
+        wait_for_checkpoint_duty(strata_rpc, timeout=30, step=0.5, min_epoch=1)
+        wait_for_no_unsigned_checkpoints(strata_rpc, timeout=30, step=0.5)
```
bewakes: Is there a possibility that checkpoints are signed faster than we query for duties? Maybe not.

functional-tests-new/tests/checkpoint/test_checkpoint_created.py
krsnapaudel: You are right.

functional-tests-new/tests/checkpoint/test_checkpoint_confirmed.py
krsnapaudel: Might happen but the check will still pass.

functional-tests-new/tests/strata/test_epoch_progression.py
```diff
@@ -19,33 +19,38 @@ class TestSequencerEpochProgression(StrataNodeTest):
...
 
-        for _ in range(1, epochs_to_check + 1):
+        for target_epoch in range(1, epochs_to_check + 1):
             epoch = wait_until_with_value(
-                lambda: strata.get_sync_status(rpc)["confirmed"],
-                lambda v, cur_epoch=prev_epoch: v is not None and v["epoch"] > cur_epoch["epoch"],
-                timeout=10,
-                error_with="Epoch not progressing",
+                lambda: (
+                    btc_rpc.proxy.generatetoaddress(1, mine_addr),
```
bewakes: Clever piggybacking of mining blocks! This avoids having to do frequent mining in a background thread.

crates/ol/stf/src/manifest_processing.rs
```diff
@@ -2,7 +2,8 @@
 
 use strata_acct_types::{AccountId, MsgPayload};
 use strata_asm_common::{AsmLogEntry, AsmManifest};
-use strata_asm_manifest_types::{CheckpointAckLogData, DepositIntentLogData};
+use strata_asm_logs::{CheckpointTipUpdate, constants::CHECKPOINT_TIP_UPDATE_LOG_TYPE};
+use strata_asm_manifest_types::DepositIntentLogData;
```
irnb: We shouldn't be importing any logs from `strata_asm_manifest_types`, ASM subprotocol logs are defined in `crates/asm/logs`. Not sure `manifest-types` should even have log definitions tbh.

cc: @prajwolrg

krsnapaudel: @irnb @prajwolrg How does the OL process the checkpoint update log then? NOTE I only changed the type of log processed.

delbonis: The OL could import `strata-asm-log-types`, or whatever the crate is.

crates/status/src/chain.rs
delbonis: We only need *really* need finalized in the client, in reported RPCs, not ASM/OL.

functional-tests-new/common/services/strata.py
```diff
@@ -202,3 +203,79 @@ def check_block_generation_in_range(self, rpc: JsonRpcClient, start: int, end: i
...
             self.wait_for_block_height(target_height, rpc)
         return self.get_cur_block_height(rpc)
+
+    def wait_for_confirmed_epoch(
+        self,
+        target_epoch: int,
+        rpc: JsonRpcClient | None = None,
+        timeout: int = 60,
+        poll_interval: float = 1.0,
+    ) -> int:
```
delbonis: Maybe this could accept `target_epoch = None` to just wait for "the next epoch" to be confirmed.

functional-tests-new/envconfigs/strata.py
```diff
@@ -15,15 +15,34 @@ class StrataEnvConfig(flexitest.EnvConfig):
...
+        )
         return flexitest.LiveEnv(services)
 
     @staticmethod
-    def get_services(ectx: flexitest.EnvContext, pre_generate_blocks: int = 0):
+    def get_services(
+        ectx: flexitest.EnvContext,
+        pre_generate_blocks: int = 0,
+        epoch_sealing_config: EpochSealingConfig | None = None,
+    ):
```
delbonis, line 44: Wait, why is this `@staticmethod`, does that even make sense?  This is doing a lot of stuff that a mere "getter" fn shouldn't be doing.  What is this actually for?

functional-tests-new/tests/checkpoint/helpers.py
```diff
@@ -0,0 +1,73 @@
...
+        step=step,
+    )
+
+
+def wait_for_no_unsigned_checkpoints(
+    rpc,
+    timeout: int = 60,
+    step: float = 1.0,
+    method: str = "strata_strataadmin_getSequencerDuties",
+):
```
delbonis: It feels like this fn should be defined around waiting for all of the current checkpoint signing duties to have been completed.  But I don't see this being called anywhere, am I missing something?

Also, the method probably doesn't need to be overridable.

functional-tests-new/tests/checkpoint/test_checkpoint_confirmed.py
```diff
@@ -0,0 +1,55 @@
...
+                    btc_rpc.proxy.generatetoaddress(1, mine_addr),
+                    strata.get_sync_status(strata_rpc)["confirmed"],
+                )[1],
+                lambda v, target_epoch=target_epoch: (
+                    v is not None and v["epoch"] >= target_epoch and v["last_blkid"] != "00" * 32
+                ),
+                timeout=120,
+                step=1.0,
+                error_with=f"Confirmed epoch did not reach {target_epoch}",
+            )
```
delbonis: Encapsulate this better!  We should generally avoid any calls to the low-level wait fns directly in tests, since they're often reused across tests and this makes refactoring easier in the future when we change the RPC interfaces.

delbonis (changes requested):

crates/ol/stf/src/manifest_processing.rs
krsnapaudel: @delbonis @irnb @prajwolrg So will there be another ticket/PR to clean up ASM crates stuff?

The import is needed — the question is which crate it should come from. Apparently a similar update is needed in csm-worker as well. If you want it in `strata-asm-logs` or `strata-asm-logs-types` instead of `strata-asm-manifest-types`, where do we do this refactor? Afaiu, this PR is not the place for it. Can one of you with knowledge of how these should be organized create a ticket?

crates/status/src/chain.rs
krsnapaudel: @delbonis Then what do we do with `confirmed` in `GetChainStatus` response? Set it to prev epoch like it's done now?

barakshani: >Set it to prev epoch like it's done now?

no, `confirmed` means on-chain, while in our codebase, that's no longer necessarily the case. I suggest we add an explicit field for `parent` checkpoint.
There may also be value in recording the actual latest confirmed checkpoint, or alternatively record only confirmed (and remove finalized). Writing it as a suggestion for discussion, I haven't been involved in RPC work much to have a strong opinion.

functional-tests-new/envconfigs/strata.py
krsnapaudel: @delbonis Apparently
It's `@staticmethod` so that `AlpenClientEnv` (in `el_ol.py:48`) can call `strata_services = StrataEnvConfig.get_services(ectx, self.pre_generate_blocks)` to reuse the Bitcoin+Strata setup logic without subclassing `StrataEnvConfig`.  Are you fine with just renaming it to say `create_services`?

functional-tests-new/tests/checkpoint/helpers.py
krsnapaudel: Removing this since it's not used.

crates/status/src/chain.rs
krsnapaudel: @barakshani Yes I am going to add `parent` to the RPC response. Apparently, we cannot remove `confirmed` because `alpen-client` depends on it. Although it is incorrectly set to prev epoch.

crates/csm-worker/src/processor.rs
```diff
@@ -158,14 +206,38 @@ fn update_client_state_with_checkpoint(
     Ok(())
 }
 
+/// Build a synthetic [`L1Checkpoint`] from a v1 checkpoint tip update.
+fn checkpoint_from_tip_update(
+    checkpoint_tip_update: &CheckpointTipUpdate,
+    asm_block: &L1BlockCommitment,
+) -> L1Checkpoint {
+    let tip = checkpoint_tip_update.tip();
+    let l1_reference = CheckpointL1Ref::new(*asm_block, Buf32::zero(), Buf32::zero());
```
prajwolrg, line 220: Should probably add a FIXME/TODO/Ticket so we don't forget to fix this later. I wonder if this is causing some of the tests to fail.

crates/csm-worker/src/processor.rs
```diff
@@ -158,14 +206,38 @@ fn update_client_state_with_checkpoint(
...
+
+    let batch_info = BatchInfo::new(
+        tip.epoch,
+        (*asm_block, *asm_block),
+        (*tip.l2_commitment(), *tip.l2_commitment()),
+    );
```
prajwolrg, line 230: The batch info here is wrong. The checkpoint most likely process L1 blocks earlier than `asm_block` and the starting point of the l2 is incorrect.

crates/csm-worker/src/processor.rs
```diff
@@ -158,14 +206,38 @@ fn update_client_state_with_checkpoint(
...
+
+    let batch_transition = BatchTransition {
+        epoch: tip.epoch,
+        chainstate_transition: ChainstateRootTransition {
+            pre_state_root: Buf32::zero(),
+            post_state_root: Buf32::zero(),
+        },
+    };
```
prajwolrg, line 242: This should not be required anymore. If possible we should remove batch_transition information from L1Checkpoint altogether.

crates/ol/stf/src/manifest_processing.rs
prajwolrg: Yes I do feel a need to have a ticket/PR that actually makes sure that we have bunch of things in place before making sure that the scope of this ticket is validated. I realized some of the places that can be cleaned up as part of the review. I'm not sure the extent of it. Let me take another look at this.

functional-tests-new/common/config/params.py
```diff
@@ -43,10 +43,32 @@ class GenesisL1View:
...
+            h = blkheight - i
+            if h < 0:
+                last_11_timestamps.append(0)
+            else:
+                bh = btc_rpc.proxy.getblockhash(h)
+                last_11_timestamps.append(btc_rpc.proxy.getblock(bh, 1)["time"])
+
+        return GenesisL1View(
+            l1blk_commitment, next_target, epoch_start_timestamp, last_11_timestamps
+        )
```
prajwolrg: I don't think we should have this. After https://github.com/alpenlabs/alpen/pull/1396, we should use datatool to create the OLParams instead.

prajwolrg (changes requested):

crates/ol/stf/src/manifest_processing.rs
evgenyzdanovich: Dilli you are right - we (protocol) should clean things up. The intention was to expose all the necessary (for the users) stuff through `asm/logs`, but I guess that intention was lost in the messiness of the days :) and we somehow ended up with logs defined elsewhere, so Hamid is also right in this regard. 

Mentioned that, IMO we (protocol) should clean it up. I'm fine with the changes (especially given _it was already defined and used before the current PR_).

crates/csm-worker/src/processor.rs
prajwolrg: `BatchTransition` is the public parameter of the old checkpoint proof. So removing this properly mean we need to remove it to update the old public parameter. This would be outside the scope of this PR.

crates/csm-worker/src/processor.rs
prajwolrg: Maybe we can add a TODO: and continue here. The information is not available directly.

storopoli: Let's try to have a ticket for every relevant `TODO` that we add in PRs.

crates/csm-worker/src/processor.rs
krsnapaudel: @prajwolrg Created [this ticket](https://alpenlabs.atlassian.net/browse/STR-2438).

crates/csm-worker/src/processor.rs
krsnapaudel: @prajwolrg Created [this ticket](https://alpenlabs.atlassian.net/browse/STR-2438). Please update with what the issue is or what the definition of Done is.

crates/ol/stf/src/manifest_processing.rs
krsnapaudel: I have created this [ticket for CSM worker side](https://alpenlabs.atlassian.net/browse/STR-2438) and [this for refactoring AMS logs](https://alpenlabs.atlassian.net/browse/STR-2439).

crates/csm-worker/src/processor.rs
krsnapaudel: @prajwolrg, @storopoli Created [this ticket](https://alpenlabs.atlassian.net/browse/STR-2438).

krsnapaudel:
> I have some doubts whether we need to consume logs also on the CSM side, but we certainly must handle it in the ol-stf. And this PR goes extra few miles to just define fn tests..
> 
> Big kudos, it's a tedious work. LGTM from subprotocol.

Thanks @evgenyzdanovich. If we don't handle checkpoint logs in CSM side now, epoch/checkpoint finalization does not work.

crates/csm-worker/src/processor.rs
prajwolrg: I updated the public parameters in https://github.com/alpenlabs/alpen/pull/1464. It removes `BatchTransition` fully.

PR #1395, opened 2026-02-22
feat!(admin-tx): add tagged sighash for multisig actions

prajwolrg:
## Description

The previous sighash was `SHA256(borsh_hash(action) ‖ seqno)`, which relied on Borsh's internal tagging (enum discriminants, field ordering) for domain separation. This is fragile: reordering enum variants, adding/removing fields, or restructuring types silently changes the sighash. 

The explicit payload layout also makes it straightforward to compute the sighash in other languages (TypeScript, Python, Go, etc.) without needing a Borsh implementation — any language with SHA-256 and byte concatenation can reproduce the hash.

* Replace bare `TxType` constants with a `repr(u8)` `AdminTxType` enum that is the single source of truth for admin transaction type discriminants.
* Introduce a `Sighash` trait that each action type implements, providing domain-separated tagged hashes `(SHA256(SHA256(tag) ‖ seqno ‖ payload))`.
* Each `AdminTxType` variant has a unique BIP-340–style sighash tag (strata/admin/<variant>) with lazy-computed tag hashes
* Simplify test harness by having `build_envelope_tx` accept a `TagData` directly and having `AdminContext::sign` return only the payload

### Type of Change

- [x] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [x]  Yes: https://www.notion.so/SPS-65-Strata-Administration-Subprotocol-265901ba000f80e583d7ff093da6b369
- [ ]  No

## Checklist

### DIFF
--- a/crates/asm/txs/admin/src/parser.rs
+++ b/crates/asm/txs/admin/src/parser.rs
@@ -44,6 +44,7 @@ impl SignedPayload {
 /// - The transaction lacks a taproot leaf script in its witness
 /// - The envelope payload cannot be parsed
 /// - The signed payload cannot be deserialized
+// TODO: https://alpenlabs.atlassian.net/browse/STR-2366
 pub fn parse_tx(tx: &TxInputRef<'_>) -> Result<SignedPayload, AdministrationTxParseError> {
     let tx_type = tx.tag().tx_type();
 


crates/asm/txs/admin/src/actions/updates/multisig.rs
```diff
@@ -34,3 +36,33 @@ impl MultisigUpdate {
...
     }
 }
+
+impl Sighash for MultisigUpdate {
+    fn tx_type(&self) -> AdminTxType {
+        match self.role {
+            Role::StrataAdministrator => AdminTxType::StrataAdminMultisigUpdate,
+            Role::StrataSequencerManager => AdminTxType::StrataSeqManagerMultisigUpdate,
+        }
+    }
```
storopoli, line 46: The parser reads `tx_type` but only uses it for error mapping; it never enforces that tag `tx_type` matches `signed_payload.action.tx_type()`. That leaves a protocol-consistency gap (tag can claim one action while payload/signature authorizes another).

https://github.com/alpenlabs/alpen/blob/d3458b3f6966095e6fe88c0cc4a5c2e447953dbf/crates/asm/txs/admin/src/parser.rs#L47-L64

crates/asm/txs/admin/src/constants.rs
```diff
@@ -1,28 +1,198 @@
...
+            }
+            // SHA256("strata/admin/asm_stf_vk_update")
+            Self::AsmStfVkUpdate => {
+                &hex!("3b997494e33fb473fd81869e0128ff44489e8ee2b07a791ab157514672d36ced")
+            }
+        }
+    }
+}
+
+impl fmt::Display for AdminTxType {
```
irnb, line 92: nit: Is there a workaround to avoid the manual Display impl here?

crates/asm/txs/admin/src/constants.rs
```diff
@@ -1,28 +1,198 @@
...
+
+impl AdminTxType {
+    /// Returns the precomputed `SHA256(tag)` for this transaction type's
+    /// sighash tag, where `tag = "strata/admin/<tx_type_name>"`.
+    pub fn sighash_tag_hash(&self) -> &'static [u8; 32] {
+        match self {
+            // SHA256("strata/admin/cancel")
+            Self::Cancel => {
+                &hex!("35d7714c91591bcdd57783d64211bff38b471d09e2d3c941c3631b3a4083d64e")
+            }
```
irnb, line 63: Are the hardcoded hex literals necessary for performance, or could we compute these from the tag strings using `const fn SHA256`  or `LazyLock`  (compile time)? 

That would make the tag  strings the single source of truth and eliminate the risk of hash/string mismatch.

irnb (changes requested):
Nice work 
The BIP-340-style tagged sighash approach is neat. 🔥 

Left a couple of minor comments.

crates/asm/txs/admin/src/constants.rs
prajwolrg: I couldn't find constant function for generating the SHA256 hash. That'd have been the best. 

Directly using hex literal is just for performance, and I added it later in [d1ce181](https://github.com/alpenlabs/alpen/pull/1395/commits/d1ce18181cf2716ae0193e24141da579fd62e3eb). I was using LazyLock earlier.

crates/asm/txs/admin/src/constants.rs
prajwolrg: For this particular use case, I could use something like the following, since `Display` output is just the variant name.

```rust
impl fmt::Display for AdminTxType {
        fmt::Debug::fmt(self, f)
    }
}
```

Alternatively I could use `strum::Display` which is meant for this purpose.

I'm not sure if it's better to couple `Display` to `Debug` semantics.

crates/asm/txs/admin/src/actions/updates/multisig.rs
prajwolrg: This is a known issue. I added a TODO for now in [bad8a69](https://github.com/alpenlabs/alpen/pull/1395/commits/bad8a69ef8bca26b61b799cb56e8037cbbc34c4a). This will be handled properly as part of that ticket.

crates/asm/txs/admin/src/constants.rs
irnb: https://github.com/alpenlabs/alpen/pull/1395/changes#diff-84578b80a127cb458451b73990648dd41cbc049a8e2ee3e4180fe85917928be6R147

irnb (approved):

PR #1396, opened 2026-02-23
feat!: make genesis OL block deterministic on OL params

prajwolrg:
## Description

* Make the genesis OL block fully deterministic by deriving it solely from OLParams, removing the dependency on waiting for an ASM manifest from the database at startup
* Add a `gen-ol-params` subcommand to `datatool` for generating OL params, and update `gen-asm-params` to accept OL params so it can compute the correct `genesis_ol_blkid`.
* Update functional test new to use datatool for generating OL and ASM params instead of Python-only dataclasses

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist


crates/ol/genesis/src/lib.rs
```diff
@@ -76,7 +61,7 @@ pub fn build_genesis_artifacts_with_manifest(
     let genesis_info = BlockInfo::new_genesis(genesis_ts);
 
     // Build genesis block components.
-    let genesis_components = BlockComponents::new_manifests(vec![genesis_manifest]);
+    let genesis_components = BlockComponents::new_empty();
```
storopoli: This makes the genesis block non-terminal which hits https://github.com/alpenlabs/alpen/blob/3e47bb55267b45567f1114f7642ee7288a643f67/crates/ol/stf/src/verification.rs#L271-L275

This is because in OL STF, terminality is derived from whether the block body has an `l1_update`. No `manifest` container => no `l1_update` => `terminal` flag becomes `false`:

https://github.com/alpenlabs/alpen/blob/3e47bb55267b45567f1114f7642ee7288a643f67/crates/ol/stf/src/assembly.rs#L303-L317

This is probably why CI is failing.

bin/datatool/src/cmd/asm_params.rs
storopoli, line 1: Here `genesis_l1_height` comes from `genesis_l1_view` (CLI input path), while `genesis_ol_blkid` is derived from `--ol-params`. There is no check that both represent the same genesis L1 anchor.

This can be a source of very nasty to debug errors.

We should validate `ol_params.last_l1_block` against the resolved genesis L1 view and hard-fail on mismatch.

bin/strata/src/genesis.rs
storopoli: We removed this logic in the PR. Can you update this comment?

functional-tests-new/factories/strata.py
```diff
@@ -33,27 +63,84 @@ def _ensure_sequencer_key(self, path: Path) -> None:
...
+                    "-o",
+                    str(asm_params_path),
+                ],
+            )
+        except RuntimeError:
+            logger.warning("strata-datatool unavailable, falling back to Python AsmParams")
+            asm_params = AsmParams().with_genesis_l1(genesis_l1)
+            with open(asm_params_path, "w") as f:
+                f.write(asm_params.as_json_string())
+        return asm_params_path
```
storopoli: This fallbacks to `AsmParams().with_genesis_l1(genesis_l1)` whenever datatool fails, but that fallback does not set `checkpoint.genesis_ol_blkid`. It defaults to `0` in

https://github.com/alpenlabs/alpen/blob/6fdd3c19703174f96ff4115e44955adf81c6a395/functional-tests-new/common/config/params.py#L163-L169

delbonis (dismissed):

crates/ol/genesis/src/lib.rs
prajwolrg: I've updated this to use `new_manifests` function with empty manifest in [4bfc803](https://github.com/alpenlabs/alpen/pull/1396/commits/4bfc80354d76d1b9997e69225d5d49239e5393b5).

functional-tests-new/factories/strata.py
prajwolrg: Removed the python classes entirely as suggested by Trey.

bin/datatool/src/cmd/asm_params.rs
prajwolrg: Oh yes. Done as part of [66b7c28](https://github.com/alpenlabs/alpen/pull/1396/commits/66b7c288b6147dbbc489a9cdbc92f65d74c77448).

PR #1397, opened 2026-02-23
feat(checkpoint-proof): integrate DA witness verification and terminal header binding 

irnb:
## Description

This PR implements the core Checkpoint Proof Statement logic, covering two major bindings:

1. L1 Reconstruction Gap (`TerminalHeaderSupplement`): Introduces a minimal `TerminalHeaderSupplement` struct and commits its hash to the `CheckpointClaim`, enabling fresh nodes to reconstruct the chain tip from L1 data alone.
2. `OL StateDiff` Binding: Integrates DA witness verification into the checkpoint proof statement, ensuring the `state_diff_hash` in the claim is computed from the actual DA bytes and verified against the `preseal state root` inside the zkVM.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/asm/subprotocols/checkpoint/src/verification.rs
+++ b/crates/asm/subprotocols/checkpoint/src/verification.rs
@@ -132,8 +132,7 @@ fn construct_full_claim(
     // Reconstruct terminal_header_complement_hash from the sidecar data posted on L1.
     // The ZK proof committed to this same hash derived from the executed terminal header,
     // so matching it here cryptographically binds the sidecar fields to proven execution.
-    let terminal_header_complement_hash =
-        sidecar.terminal_header_complement().compute_hash();
+    let terminal_header_complement_hash = sidecar.terminal_header_complement().compute_hash();
 
     Ok(CheckpointClaim::new(
         new_tip.epoch,


crates/asm/subprotocols/checkpoint/src/verification.rs
```diff
@@ -120,13 +129,19 @@ fn construct_full_claim(
     // Hash SSZ-encoded OL logs (convert to Vec for SSZ encoding)
     let ol_logs_vec = sidecar.ol_logs().to_vec();
     let ol_logs_hash = hash::raw(&ol_logs_vec.as_ssz_bytes()).into();
+    // Reconstruct terminal_header_supplement_hash from the sidecar data posted on L1.
+    // The ZK proof committed to this same hash derived from the executed terminal header,
+    // so matching it here cryptographically binds the sidecar fields to proven execution.
+    let terminal_header_supplement_hash =
+        hash::raw(&sidecar.terminal_header_supplement().as_ssz_bytes()).into();
```
delbonis: I would think we should use the SSZ tree hash here, right?

crates/checkpoint-types-ssz/src/payload.rs
```diff
@@ -32,10 +32,63 @@ impl CheckpointTip {
...
+///
+/// A fresh sequencer can reconstruct OL state from L1 but cannot recover the
+/// terminal header needed to continue block production. Most header fields are
+/// derivable (`slot`/`blkid` from `new_tip.l2_commitment`, `epoch` from
+/// `new_tip.epoch`, `state_root` from DA + manifest reconstruction, `is_terminal`
+/// by checkpoint semantics), but these four are not available from L1 data.
+///
+/// The proof commits to `hash(SSZ(TerminalHeaderSupplement))` in [`crate::CheckpointClaim`],
+/// so the L1 verifier can enforce sidecar integrity without a full header preimage.
+impl TerminalHeaderSupplement {
```
delbonis: I don't love this type name.  I'm not sure what a better one would be.

crates/ol/checkpoint/src/errors.rs
```diff
@@ -8,9 +9,13 @@ use thiserror::Error;
...
 
     /// Missing epoch summary for the given commitment.
     #[error("missing summary for epoch commitment {0:?}")]
-    MissingEpochSummary(EpochCommitment),
+    EpochSummary(EpochCommitment),
+
+    /// Missing terminal block header for the expected terminal commitment.
+    #[error("missing terminal block header for commitment {0:?}")]
+    TerminalBlock(OLBlockCommitment),
 }
```
delbonis, line 21: `CheckpointInputError`?

this type feels a little ad-hoc, not sure what to do here

crates/ol/stf/src/verification.rs
```diff
@@ -324,9 +324,27 @@ fn compute_logs_root(logs: &[OLLog]) -> Buf32 {
 /// This is derived from data in the checkpoint. (right?)
 #[derive(Clone, Debug)]
 pub struct EpochExecExpectations {
+    preseal_state_root: Buf32,
```
delbonis: I'm pretty sure we don't need the preseal state root in the checkpoint.

crates/ol/checkpoint/src/errors.rs
irnb: The current name `CheckpointNotReady` intentionally captures the transient "try again later" semantics  these aren't  permanent input validation errors, they mean the checkpoint builder can't assemble the payload yet because upstream data hasn't arrived (epoch summary not computed, terminal block not yet available, etc.). 
 `CheckpointInputError` would suggest the inputs are invalid, which isn't the case. Also, the `Missing` prefix was dropped from the variants since the enum name already conveys that context. Happy to rename if you feel strongly though.

crates/ol/stf/src/verification.rs
irnb: Good call, you're right, the preseal state root doesn't need to be here. The `EpochExecExpectations` struct and  `verify_epoch_with_diff` have no production callers; the checkpoint proof statment uses `verify_epoch_preseal_with_diff` which takes the preseal root directly as a `&Buf32` from `terminal_l1_update.preseal_state_root()`.

crates/checkpoint-types-ssz/src/payload.rs
irnb: Renamed to `TerminalHeaderComplement`. I also considered `TerminalHeaderAux`, but "complement" felt more precise here. Let me know if you have a better alternative in mind.

irnb:
The main goal of this PR is integrating DA witness verification into the checkpoint proof statement, making the checkpoint mechanism (subprotocol + proof) complete and functional for L1 sync.

Regarding the terminal header discussion: we're going forward with publishing `TerminalHeaderComplement` in the checkpoint sidecar for now, since we're confident this doesn't create any soundness gap and it solves the problem of allowing a fresh sequencer to continue the chain (creating new OL blocks, generating new checkpoint proofs) after syncing purely via L1 data. That said, we agreed to create a follow-up ticket ([STR-2378](https://alpenlabs.atlassian.net/browse/STR-2378)) to explore other alternatives for filling this gap.
 
 cc: @prajwolrg @storopoli 


[STR-2378]: https://alpenlabs.atlassian.net/browse/STR-2378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1405, opened 2026-02-26
STR 2042 mempool generalised priority ordering

krsnapaudel:
## Description

Generalizes OL mempool transaction ordering behind a pluggable `MempoolPriorityPolicy` trait, while preserving current FIFO behavior unchanged.

1. `MempoolPriorityPolicy` trait (`ordering.rs`)
- Associated type `Priority: Ord + Copy + Debug + Send + Sync + 'static`, used as `BTreeMap` keys in the priority index.
- `compute_priority(tx, timestamp_micros) -> Self::Priority` computes priority from the transaction and insertion timestamp. `txid` is derived internally to prevent mismatch.

2. `FifoPriority` default implementation
Reproduces current timestamp-FIFO semantics via `FifoPriorityKey`.

Intrinsic vs extrinsic ordering (`ordering.rs`)
- `PackageKey::SnarkAcctUpdate(AccountId)` groups same-account snark transactions that share an intrinsic ordering constraint.
- `IntrinsicOrder` with `PartialOrd`, comparable only within the same package (same-account `seq_no`).
- `FifoPriorityKey::Ord` applies intrinsic order first, falls back to FIFO timestamp cross-package.

3. `txid` tiebreaker (correctness fix)
All `FifoPriorityKey` variants include a `txid` field. Without it, two transactions with the same timestamp would silently overwrite each other in the `BTreeMap`.

4. Transaction ID
`compute_txid()` on canonical `OLTransaction` uses SSZ tree-hash over transaction data. Mempool reuses this directly, no bespoke txid computation or manual `TreeHash` impls.

5. Generic propagation
`MempoolEntry<P>`, `MempoolServiceState<SP, Prio>`, `MempoolService<SP, Prio>`, `MempoolBuilder<Prio>`. Default `= FifoPriority` only on `MempoolBuilder` (public entry point).

6. `load_from_db()`
Retains timestamp-based pre-sort for `seq_no` reconstruction correctness; explicitly documented as independent of the active priority policy.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-2042](https://alpenlabs.atlassian.net/browse/STR-2042)

[STR-2042]: https://alpenlabs.atlassian.net/browse/STR-2042?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/ol/mempool/src/builder.rs
+++ b/crates/ol/mempool/src/builder.rs
@@ -3,6 +3,7 @@
 use std::{
     fmt::{Debug, Formatter},
     future::Future,
+    marker::PhantomData,
     sync::Arc,
 };
 
@@ -15,6 +16,7 @@ use tokio::sync::{mpsc, watch};
 
 use crate::{
     MempoolCommand, MempoolHandle,
+    ordering::MempoolPriorityPolicy,
     service::MempoolService,
     state::{MempoolContext, MempoolServiceState},
     types::OLMempoolConfig,
@@ -23,14 +25,15 @@ use crate::{
 /// Builder for creating and launching mempool service.
 ///
 /// Separates service initialization logic from the handle interface.
-pub struct MempoolBuilder {
+pub struct MempoolBuilder<Prio: MempoolPriorityPolicy> {
     config: OLMempoolConfig,
     storage: Arc<NodeStorage>,
     status_channel: StatusChannel,
     current_tip: OLBlockCommitment,
+    _phantom: PhantomData<Prio>,
 }
 
-impl Debug for MempoolBuilder {
+impl<Prio: MempoolPriorityPolicy> Debug for MempoolBuilder<Prio> {
     #[expect(clippy::absolute_paths, reason = "qualified Result avoids ambiguity")]
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("MempoolBuilder")
@@ -40,7 +43,7 @@ impl Debug for MempoolBuilder {
     }
 }
 
-impl MempoolBuilder {
+impl<Prio: MempoolPriorityPolicy> MempoolBuilder<Prio> {
     /// Create a new mempool builder.
     pub fn new(
         config: OLMempoolConfig,
@@ -53,6 +56,7 @@ impl MempoolBuilder {
             storage,
             status_channel,
             current_tip,
+            _phantom: PhantomData,
         }
     }
 
@@ -70,7 +74,7 @@ impl MempoolBuilder {
 
         // Create mempool state with context and current tip
         let mut state =
-            MempoolServiceState::new_with_context(ctx.clone(), self.current_tip).await?;
+            MempoolServiceState::<_, Prio>::new_with_context(ctx.clone(), self.current_tip).await?;
 
         // Load existing transactions from database
         state.load_from_db().await?;
@@ -82,7 +86,7 @@ impl MempoolBuilder {
         let mempool_input = MempoolInput::new(command_rx, ol_sync_rx);
 
         // Launch service with mempool input
-        let monitor = ServiceBuilder::<MempoolService<_>, _>::new()
+        let monitor = ServiceBuilder::<MempoolService<_, Prio>, _>::new()
             .with_state(state)
             .with_input(mempool_input)
             .launch_async("mempool", texec)
--- a/crates/ol/mempool/src/ordering.rs
+++ b/crates/ol/mempool/src/ordering.rs
@@ -1,0 +1,124 @@
+//! Mempool ordering policies and ordering keys.
+
+use std::fmt::Debug;
+
+use strata_identifiers::OLTxId;
+use strata_ol_chain_types_new::OLTransaction;
+
+/// Policy trait for computing mempool transaction priorities.
+///
+/// Implementations define how priorities are computed from transaction data and insertion
+/// metadata. Priority ordering is interpreted with the invariant that iterating in ascending order
+/// yields highest-priority transactions first.
+pub trait MempoolPriorityPolicy: Clone + Copy + Debug + Send + Sync + 'static {
+    /// Ordering priority used by the policy.
+    type Priority: Ord + Copy + Debug + Send + Sync + 'static;
+
+    /// Compute an ordering priority for a transaction.
+    fn compute_priority(tx: &OLTransaction, timestamp_micros: u64) -> Self::Priority;
+}
+
+/// FIFO priority policy.
+///
+/// This is the current default behavior and will continue to be used unless another policy is
+/// explicitly selected.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Hash)]
+pub struct FifoPriority;
+
+impl MempoolPriorityPolicy for FifoPriority {
+    type Priority = FifoPriorityKey;
+
+    fn compute_priority(tx: &OLTransaction, timestamp_micros: u64) -> Self::Priority {
+        FifoPriorityKey::for_tx(timestamp_micros, tx)
+    }
+}
+
+/// FIFO package-ordering key.
+///
+/// This key is only for global package ordering and does not encode intra-package dependencies.
+/// The `timestamp_micros` is in microseconds since UNIX epoch.
+///
+/// Derived [`Ord`] is load-bearing here: field declaration order defines sort order.
+/// `timestamp_micros` is first and `txid` is second (tie-breaker); do not reorder fields.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct FifoPriorityKey {
+    timestamp_micros: u64,
+    txid: OLTxId,
+}
+
+impl FifoPriorityKey {
+    /// Creates a FIFO priority key from `timestamp_micros` and `txid`.
+    pub(crate) fn new(timestamp_micros: u64, txid: OLTxId) -> Self {
+        Self {
+            timestamp_micros,
+            txid,
+        }
+    }
+
+    /// Create ordering key for a transaction.
+    pub(crate) fn for_tx(timestamp_micros: u64, tx: &OLTransaction) -> Self {
+        Self::new(timestamp_micros, tx.compute_txid())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::cmp::Ordering;
+
+    use super::*;
+    use crate::{
+        test_utils::{
+            create_test_generic_tx, create_test_snark_tx_with_seq_no, create_test_txid_with,
+        },
+        types::MempoolEntry,
+    };
+
+    #[test]
+    fn test_priority_orders_by_timestamp_micros() {
+        let tx = create_test_snark_tx_with_seq_no(1, 0);
+        let entries: Vec<_> = (1..=3)
+            .map(|ts| {
+                MempoolEntry::<FifoPriority>::new(
+                    tx.clone(),
+                    FifoPriorityKey {
+                        timestamp_micros: ts,
+                        txid: create_test_txid_with(ts as u8),
+                    },
+                    100,
+                )
+            })
+            .collect();
+
+        for i in 0..entries.len() - 1 {
+            assert_eq!(
+                entries[i].priority.cmp(&entries[i + 1].priority),
+                Ordering::Less
+            );
+        }
+    }
+
+    #[test]
+    fn test_priority_tie_breaks_with_txid() {
+        let tx = create_test_generic_tx();
+        let key_a = FifoPriorityKey::for_tx(1_000_000, &tx);
+        let key_b = FifoPriorityKey {
+            timestamp_micros: key_a.timestamp_micros,
+            txid: create_test_txid_with(255),
+        };
+        assert!(key_a != key_b);
+        assert_ne!(key_a.cmp(&key_b), Ordering::Equal);
+    }
+
+    #[test]
+    fn test_priority_tie_break_uses_txid() {
+        let key_a = FifoPriorityKey {
+            timestamp_micros: 99,
+            txid: create_test_txid_with(1),
+        };
+        let key_b = FifoPriorityKey {
+            timestamp_micros: 99,
+            txid: create_test_txid_with(2),
+        };
+        assert_eq!(key_a.cmp(&key_b), Ordering::Less);
+    }
+}
--- a/crates/ol/mempool/src/package.rs
+++ b/crates/ol/mempool/src/package.rs
@@ -1,0 +1,570 @@
+//! Package-level mempool bookkeeping.
+//!
+//! A package is a group of transactions that share an intrinsic ordering rule.
+
+use std::collections::{BTreeMap, HashMap, VecDeque};
+
+use strata_acct_types::AccountId;
+use strata_identifiers::OLTxId;
+use strata_ol_chain_types_new::{OLTransaction, TransactionPayload};
+use tracing::error;
+
+use crate::ordering::MempoolPriorityPolicy;
+
+/// Identifies a package by its grouping key.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub(crate) enum PackageKey {
+    /// Package for snark account updates targeting a specific account.
+    SnarkAccountUpdate(AccountId),
+
+    /// Standalone package for a transaction with no intrinsic dependencies.
+    Standalone(OLTxId),
+}
+
+/// A transaction's package identity and intra-package position.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub(crate) enum PackageMember {
+    /// Per-account ordered snark update.
+    SnarkAccountUpdate {
+        /// Target account ID.
+        account_id: AccountId,
+
+        /// Per-account sequence number.
+        seq_no: u64,
+    },
+
+    /// Standalone transaction with no intrinsic dependencies.
+    Standalone {
+        /// Transaction ID.
+        txid: OLTxId,
+    },
+}
+
+impl PackageMember {
+    /// Construct package membership metadata for a transaction.
+    pub(crate) fn from_tx(tx: &OLTransaction, txid: OLTxId) -> Self {
+        match tx.payload() {
+            TransactionPayload::SnarkAccountUpdate(payload) => Self::SnarkAccountUpdate {
+                account_id: *payload.target(),
+                seq_no: payload.operation().update().seq_no(),
+            },
+            TransactionPayload::GenericAccountMessage(_) => Self::Standalone { txid },
+        }
+    }
+
+    /// Return this member's package key.
+    pub(crate) fn package_key(&self) -> PackageKey {
+        match self {
+            Self::SnarkAccountUpdate { account_id, .. } => {
+                PackageKey::SnarkAccountUpdate(*account_id)
+            }
+            Self::Standalone { txid } => PackageKey::Standalone(*txid),
+        }
+    }
+}
+
+/// Internal package-index invariant violations used for diagnostics.
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum PackageInvariantError {
+    /// Attempted to insert a txid that already exists in the package index.
+    #[error("duplicate txid in package index: {0:?}")]
+    DuplicateTxId(OLTxId),
+
+    /// Transaction was mapped to a package key but package entry was missing.
+    #[error("missing package entry for key: {0:?}")]
+    MissingPackage(PackageKey),
+
+    /// Package entry shape and member shape did not match.
+    #[error("package/member mismatch (key: {key:?}, member: {member:?})")]
+    PackageMemberMismatch {
+        key: PackageKey,
+        member: PackageMember,
+    },
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct PackageTx<Prio: MempoolPriorityPolicy> {
+    txid: OLTxId,
+    priority: Prio::Priority,
+}
+
+/// Package contents with explicit intrinsic ordering semantics.
+#[derive(Debug, Clone)]
+enum PackageContents<Prio: MempoolPriorityPolicy> {
+    /// Ordered by snark seq_no.
+    SnarkAccountUpdate(BTreeMap<u64, PackageTx<Prio>>),
+
+    /// Single standalone transaction.
+    Standalone(Option<PackageTx<Prio>>),
+}
+
+impl<Prio: MempoolPriorityPolicy> PackageContents<Prio> {
+    fn new_for_member(member: PackageMember) -> Self {
+        match member {
+            PackageMember::SnarkAccountUpdate { .. } => Self::SnarkAccountUpdate(BTreeMap::new()),
+            PackageMember::Standalone { .. } => Self::Standalone(None),
+        }
+    }
+
+    fn front(&self) -> Option<PackageTx<Prio>> {
+        match self {
+            Self::SnarkAccountUpdate(txs) => txs.first_key_value().map(|(_, tx)| *tx),
+            Self::Standalone(tx) => *tx,
+        }
+    }
+
+    fn insert(
+        &mut self,
+        member: PackageMember,
+        txid: OLTxId,
+        priority: Prio::Priority,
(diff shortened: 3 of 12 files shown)


crates/ol/mempool/src/types.rs
```diff
@@ -11,6 +11,37 @@ use strata_snark_acct_types::SnarkAccountUpdate;
...
+/// yields highest-priority transactions first.
+pub trait MempoolPriorityPolicy: Clone + Copy + Debug + Send + Sync + 'static {
+    /// Ordering key used by the policy.
+    type Key: Ord + Copy + Debug + Send + Sync + 'static;
+
+    /// Compute an ordering key for a transaction.
+    ///
+    /// `txid` is provided for deterministic tie-breaking when two transactions otherwise share the
+    /// same priority.
+    fn compute_key(tx: &OLMempoolTransaction, timestamp_micros: u64, txid: OLTxId) -> Self::Key;
```
bewakes: I think it would make this definition simpler and more ergonomic if `OLMempoolTransaction` cointained `timestamp_micros` and possibly `txid` as well. Because now the implementor has to think about bringing in timestamp and txid. Plus, key should be computable from the mempool transaction only since that's the thing we want to have an priority ordering.

bewakes: Also the term `Key` and `compute_key` tend to convey different meaning like the transaction is going to be hashed. Possibly we can rename them to `Score`/`Priority` and `compute_score`/`compute_priority`?

krsnapaudel: @bewakes I will add `timestamp_micros`. Not sure about `txid`, since it's a derived id.

crates/ol/mempool/src/ordering.rs
```diff
@@ -0,0 +1,396 @@
...
+    /// Generic account message transaction.
+    ///
+    /// Ordered by timestamp only (pure FIFO).
+    Gam {
+        /// Timestamp (microseconds since UNIX epoch) for FIFO ordering.
+        timestamp_micros: u64,
+        /// Transaction ID for deterministic tie-breaking.
+        txid: OLTxId,
+    },
+}
```
delbonis: This should be restructured.  There's common fields (`timestamp_micros`, `txid`) that could be promoted to a container type, and then type-specific ordering fields can be kept as the individual types.  But I don't really like how it doesn't naturally generalize across tx types.

The issue is actually deeper though, really part of the modelling issue is that we're trying to define an intrinsic *total order* to txs.  But really the only *intrinsic* ordering is just a *partial order* based on the seqno **iff** they're both snark acct update txs for the same account.  This is why we have to have so much information in this and the `Ord` impl is so complex.  We'd really want to implement `PartialOrd` (and not `Ord`) on some type that describes just that information.  *Then* we can define a total order *extrinsically*, based on the timestamp (or maybe feerate in the future).

A more holistic way to model it would be to borrow the concept of "transaction packages" from bitcoin, where transactions that have intrinsic linkages on validity (in bitcoin's case, based on immediately-spent outputs, in our case, based on seqno) are grouped into the same package.  Then, we can construct blocks with packages in ~arbitrary orders.  We'd probably implement this by having some `PackageKey` concept and table that maps package keys to existing packages with known keys, and a table of unpackaged txs (since not every tx would necessarily have a need to be grouped into a package).

```
enum PackageKey {
    SnarkAcctUpdate(AccountId),
}
```

For more information, see [package relay](https://bitcoinops.org/en/topics/package-relay/) which is a different but related concept where this concept of clustering txs is promoted up to the tx relay logic.

crates/ol/mempool/src/types.rs
```diff
@@ -157,6 +160,15 @@ pub struct OLMempoolTransaction {
...
+
+    /// Insertion timestamp in microseconds since UNIX epoch.
+    pub timestamp_micros: u64,
+}
+
+#[derive(Encode)]
+struct MempoolTxIdPreimage<'a> {
+    payload: &'a OLMempoolTxPayload,
+    attachment: &'a TransactionAttachment,
 }
```
delbonis: What is this type?  Why is it `Encode` but there's no `Decode`?

crates/ol/mempool/src/service.rs
```diff
@@ -21,14 +25,15 @@ pub struct MempoolServiceStatus {
 ///
 /// # Type Parameters
 ///
-/// - `P`: The state provider type that implements [`StateProvider`].
+/// - `SP`: The state provider type that implements [`StateProvider`].
+/// - `Prio`: The mempool priority policy.
 #[derive(Debug)]
-pub(crate) struct MempoolService<P: StateProvider> {
-    _phantom: PhantomData<P>,
+pub(crate) struct MempoolService<SP: StateProvider, Prio: MempoolPriorityPolicy = FifoPriority> {
```
delbonis: Shouldn't make this a default type parameter.

crates/ol/mempool/src/ordering.rs
```diff
@@ -0,0 +1,396 @@
...
+/// Ordering key for mempool transactions.
+///
+/// Provides collision-free ordering with different strategies for different transaction types:
+/// - **Snark transactions**: Strict per-account seq_no ordering with FIFO tiebreaking across
+///   accounts
+/// - **GAM transactions**: Pure FIFO ordering by `timestamp_micros`
+///
+/// The `timestamp_micros` is in microseconds since UNIX epoch.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum MempoolOrderingKey {
```
delbonis: This ordering key looks like it's trying to be universal, but it's really just local to the fifo policy.

crates/ol/mempool/src/ordering.rs
```diff
@@ -0,0 +1,396 @@
...
+        /// Timestamp (microseconds since UNIX epoch) for FIFO ordering.
+        timestamp_micros: u64,
+        /// Transaction ID for deterministic tie-breaking.
+        txid: OLTxId,
+    },
+}
+
+impl MempoolOrderingKey {
+    /// Create ordering key for a transaction.
+    pub(crate) fn for_transaction(tx: &OLMempoolTransaction, txid: OLTxId) -> Self {
```
delbonis: `from_tx`

crates/ol/mempool/src/types.rs
```diff
@@ -210,153 +229,44 @@ impl OLMempoolTransaction {
...
     pub fn compute_txid(&self) -> OLTxId {
-        let encoded = ssz::Encode::as_ssz_bytes(self);
+        let preimage = MempoolTxIdPreimage {
+            payload: &self.payload,
+            attachment: &self.attachment,
+        };
+        let encoded = ssz::Encode::as_ssz_bytes(&preimage);
         let hash_bytes = hash::raw(&encoded);
         OLTxId::from(hash_bytes)
     }
```
delbonis: This is not a txid, don't call it a txid.

Also:

> This follows the established pattern: SSZ encode → SHA256 hash.

This is not really an established pattern, we should generally be using the SSZ tree hash for things like this.

crates/ol/mempool/src/types.rs
krsnapaudel: @delbonis This is for computing txid of mempool tx. No longer relevant.

crates/ol/mempool/src/types.rs
krsnapaudel: @delbonis I have left `compute_txid` in place returning `OLTxId`. The ID is now computed via SSZ tree-hash over canonical transaction content (payload + attachment), same hashing approach used for all other protocol types. It's a content-addressed identifier for the mempool-form transaction. Is your objection to name of method and return type or how id is computed?

delbonis: My objection was that you defined a new type and then used the flat hash of that encoded message as a txid, which was completely arbitrary.

~~*I think* this new approach is fine for now.~~ I reviewed it and retract this point.

crates/ol/mempool/src/types.rs
```diff
@@ -115,6 +143,33 @@ pub enum OLMempoolTxPayload {
...
+                let root = <_ as TreeHash<H>>::tree_hash_root(inner);
+                mix_in_selector_with_hasher::<H>(&root, 0u8).expect("valid selector")
+            }
+            Self::SnarkAccountUpdate(inner) => {
+                let root = <_ as TreeHash<H>>::tree_hash_root(inner);
+                mix_in_selector_with_hasher::<H>(&root, 1u8).expect("valid selector")
+            }
+        }
+    }
+}
```
delbonis: Why are you defining tree hash on this type manually?  If this is naturally SSZ-able we should be able to do `#[derive(TreeHash)]`.  If not, then we shouldn't be implementing treehash on it in the first place and should be more careful about how the types are being modeled.

crates/ol/mempool/src/types.rs
delbonis, line 1: There's something really weird going on here with these types.  It's a pretty big red flag to me that for the mempool which treats transactions mostly as black boxes (aside from target, seqno, etc.) is defining special mempool representations of transaction types with their own serialization and treehashing.  Can you see if it's possible to use the canonical transaction types, sure maybe with containers with extra metadata, or if not explain why this is necessary?

crates/ol/mempool/src/types.rs
```diff
@@ -146,17 +201,18 @@ impl OLMempoolTxPayload {
...
 pub struct OLMempoolTransaction {
     /// Transaction payload (generic message or snark account update).
     pub payload: OLMempoolTxPayload,
 
     /// Transaction attachment (min_slot, max_slot constraints).
     pub attachment: TransactionAttachment,
+
+    /// Insertion timestamp in microseconds since UNIX epoch.
+    pub timestamp_micros: u64,
 }
```
delbonis: This type's purpose is confused / at odds with itself because it contains two *intrinsic* components of the transaction itself alongside *extrinsic* information (timestamp).  These are different levels of abstraction and they should be organized in a way that reflects that.

crates/ol/mempool/src/types.rs
```diff
@@ -207,161 +270,42 @@ impl OLMempoolTransaction {
...
-                    timestamp_micros: t2,
-                    ..
-                },
-            ) => timestamp_micros.cmp(t2),
-        }
+        let mut leaves = [0u8; 64];
+        leaves[..32].copy_from_slice(payload_root.as_ref());
+        leaves[32..].copy_from_slice(attachment_root.as_ref());
+        let root = merkle_root_with_hasher::<Sha256Hasher>(&leaves, 2);
+        OLTxId::from(Buf32::from(root.0))
```
delbonis: Actually no, I didn't quite see that this is how this was implemented.  Even if this is mathematically correct, you should just be directly calling out to the transaction types instead of reimplementing low-level hashing operations!  This is how we get incompatibilities across different areas of the codebase, since different components use bespoke versions of operations like "compute txid".  We should be defining "computing a txid" in *one* singular way, in one place, and everything else just calls that.

crates/ol/mempool/src/types.rs
krsnapaudel: As discussed in Slack, I will wait for a separate PR to remove mempool-specific tx types. Since they should have ssz support, we will not need manual TreeHash impls as well.

crates/ol/mempool/src/ordering.rs
```diff
@@ -0,0 +1,478 @@
...
+/// yields highest-priority transactions first.
+pub trait MempoolPriorityPolicy: Clone + Copy + Debug + Send + Sync + 'static {
+    /// Ordering priority used by the policy.
+    type Priority: Ord + Copy + Debug + Send + Sync + 'static;
+
+    /// Compute an ordering priority for a transaction.
+    ///
+    /// `txid` is provided for deterministic tie-breaking when two transactions otherwise share the
+    /// same priority.
+    fn compute_priority(tx: &OLMempoolTransaction, txid: OLTxId) -> Self::Priority;
```
bewakes: Is there a reason `txid` is being passed separately with `tx`? Since `txid` can be derived from tx, the definition of this method should only have `tx` as a parameter imo.

bewakes: The way it is now leaves room for passing one tx and a different txid not corresponding to that tx.

crates/ol/mempool/src/ordering.rs
```diff
@@ -0,0 +1,478 @@
...
+        FifoPriorityKey::from_tx(tx, txid)
+    }
+}
+
+/// Package key for intrinsic transaction dependencies.
+///
+/// This captures only intrinsic validity linkages between transactions. Today the only intrinsic
+/// linkage is per-account sequencing for snark account updates.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub(crate) enum PackageKey {
```
bewakes: What is a package?

crates/ol/mempool/src/ordering.rs
```diff
@@ -0,0 +1,478 @@
...
+            (
+                Self::Snark {
+                    timestamp_micros: t1,
+                    txid: tx1,
+                    ..
+                },
+                Self::Snark {
+                    timestamp_micros: t2,
+                    txid: tx2,
+                    ..
```
bewakes: Can we define the `timestamp_micros()` and `txid()` methods in `FifoPriorityKey` so that this matching becomes cleaner? The destructuring is quite verbose and repititive here.

crates/ol/mempool/src/ordering.rs
krsnapaudel: @bewakes This is a group of txs from the same account. So their ordering is linked.

crates/ol/mempool/src/ordering.rs
krsnapaudel: @bewakes Now computing txid again. The idea was to pass precomputed txid.

crates/ol/mempool/src/types.rs
krsnapaudel: Not relevant anymore.

crates/ol/mempool/src/types.rs
krsnapaudel: Not relevant now.

crates/ol/mempool/src/types.rs
krsnapaudel: Not relevant now.

crates/ol/mempool/src/types.rs
krsnapaudel: Not relevant anymore.

crates/ol/mempool/src/builder.rs
```diff
@@ -23,14 +25,15 @@ use crate::{
...
 ///
 /// Separates service initialization logic from the handle interface.
-pub struct MempoolBuilder {
+pub struct MempoolBuilder<Prio: MempoolPriorityPolicy = FifoPriority> {
     config: OLMempoolConfig,
     storage: Arc<NodeStorage>,
     status_channel: StatusChannel,
     current_tip: OLBlockCommitment,
+    _phantom: PhantomData<Prio>,
 }
```
delbonis: Don't use a default type param like this.  It's meant to be used for things like allowing providing an allocator in a collection type, but this is actually a meaningful behavioral param.

crates/ol/mempool/src/ordering.rs
```diff
@@ -0,0 +1,338 @@
...
+}
+
+/// Intrinsic ordering relation between transactions.
+///
+/// This represents only protocol-intrinsic ordering. It is intentionally a partial order:
+/// transactions are comparable only when they belong to the same package.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub(crate) enum IntrinsicOrder {
+    /// Snark update with per-account sequence ordering.
+    SnarkSeq { package: PackageKey, seq_no: u64 },
```
delbonis: What if we added sequencer accounts that are used to pay fees to submit snark account updates / collect exec sequencer fees.  Then the types would like this:

```
enum PackageKey {
  SnarkAcctUpdate(AccountId),
  ExternalAccount(AccountId),
}

enum IntrinsicOrder {
  SnarkSeq { package: PackageKey, seq_no: u64 }
  External { package: PackageKey, seq_no: u64 }
  Unordered
}
```

This would allow us to construct `IntrinsicOrder::SnarkSeq { package: PackageKey::ExternalAccount(id), seq_no: ... }` and vice-versa with the other two variants.  This would be nonsensical, which suggests this type hierarchy should be thought through some more to extract the actual essence.

crates/ol/mempool/src/ordering.rs
```diff
@@ -0,0 +1,240 @@
...
+}
+
+impl FifoPriorityKey {
+    /// Create ordering key for a transaction.
+    pub(crate) fn for_transaction(tx: &OLTransaction, timestamp_micros: u64) -> Self {
+        Self {
+            timestamp_micros,
+            txid: tx.compute_txid(),
+        }
+    }
```
delbonis: This being the sole constructor is weird.

```rs
pub fn new(ts_micros: u64, txid: OLTxId) -> Self { Self { ... } }
pub fn from_tx(ts_micros: u64, tx: &OLTransaction) -> Self { Self::new(ts_micros, tx.compute_txid()) }
```

crates/ol/mempool/src/state.rs
```diff
@@ -19,12 +19,74 @@ use tracing::warn;
...
+    types::{MempoolEntry, OLMempoolConfig, OLMempoolRejectReason, OLMempoolStats},
     validation::validate_transaction,
 };
 
+/// Materialized per-package transaction index.
+#[derive(Debug, Clone)]
+enum PackageIndex {
+    SnarkBySeq(BTreeMap<u64, OLTxId>),
+    Standalone(Option<OLTxId>),
+}
```
delbonis: This type doesn't make any sense.

You need a type that's like:

```rs
pub struct PackageContents {
  txids: Vec<OLTxId>,
  // maybe other stuff
}
```

crates/ol/mempool/src/state.rs
```diff
@@ -19,12 +19,74 @@ use tracing::warn;
...
+            (Self::SnarkBySeq(txs), PackageMember::SnarkAccountUpdate { seq_no, .. }) => {
+                txs.remove(&seq_no);
+            }
+            (Self::Standalone(stored_txid), PackageMember::Standalone { .. }) => {
+                if *stored_txid == Some(txid) {
+                    *stored_txid = None;
+                }
+            }
+            _ => {
+                unreachable!("package index variant and package member variant must match")
```
delbonis: The `unreachable!`s are *very* reachable from the outside:

 ```rs
Self::SnarkBySeq(BTreeMap::new()).insert(PackageMember::Standalone { ... }, ...)
```

File: crates/ol/mempool/src/state.rs [hunk 19]
```diff
@@ -105,36 +167,46 @@ impl<P: StateProvider> Debug for MempoolContext<P> {
...
+    /// Package ordering index: package priority → package key.
+    package_priority_index: BTreeMap<Prio::Priority, PackageKey>,
+
+    /// Materialized package index: package key → package transactions.
+    packages: HashMap<PackageKey, PackageIndex>,
 
-    /// Ordering index: MempoolOrderingKey → transaction ID.
-    ordering_index: BTreeMap<MempoolOrderingKey, OLTxId>,
+    /// Tracks current priority for each package.
+    package_priorities: HashMap<PackageKey, Prio::Priority>,
```
delbonis: This could be part of the package type I suggested above.

crates/ol/mempool/src/state.rs
```diff
@@ -105,36 +167,46 @@ impl<P: StateProvider> Debug for MempoolContext<P> {
...
+
+    /// Reverse index: txid → package key.
+    tx_to_package: HashMap<OLTxId, PackageKey>,
 
     /// Per-account mempool state.
     /// Tracks all txids and sequence numbers for each account.
     account_state: HashMap<AccountId, AccountMempoolState>,
```
delbonis, line 123: There's a *lot* of *very careful* bookkeeping to keep all of these indexes consistent that needs to be perfectly right every time.  It's a big ball of mutable state, and also seems more complex than it needs to be.

For example, why do we need to maintain an index of prio->key?  If we store the prio as part of the package list entries, and we keep around the packages to construct blocks from, we would just binary search on the list instead of keeping a separate index.

This internal bookkeeping should also be split out into a separate type (and module, really) so that it can be modelled, reasoned about, and tested independently from the high-level/service-level mempool bookkeeping (context, state, stats, etc.).

delbonis (changes requested):

crates/ol/mempool/src/state.rs
krsnapaudel: @delbonis Agreed we should replace PackageIndex with a clearer package-store abstraction.
Only caveat: plain `Vec<OLTxId>` is unsafe for snark packages unless seq ordering/replacement invariants are explicitly enforced; we need seq-indexed contents for correctness.

crates/ol/mempool/src/state.rs
krsnapaudel: @delbonis Moved package stuff to `package.rs`.

crates/ol/mempool/src/package.rs
```diff
@@ -0,0 +1,570 @@
...
+                        key: member.package_key(),
+                        member,
+                    });
+                }
+                *slot = Some(PackageTx { txid, priority });
+                Ok(())
+            }
+            _ => {
+                debug_assert!(
+                    false,
```
delbonis, line 140: Would be more idiomatic to just do:

```rs
#[cfg(debug_assertions)]
panic!(...);
```

Same exists in several other places.

However the change made here (from `unreachable!`) doesn't really address the issue I was pointing at with the earlier comment.  When there's large classes of inputs where it seems like the "right thing to do" is to just panic that's a signal that the abstractions being built aren't appropriate for the problem we're attempting to solve.

crates/ol/mempool/src/package.rs
```diff
@@ -0,0 +1,570 @@
...
+}
+
+/// Package contents with explicit intrinsic ordering semantics.
+#[derive(Debug, Clone)]
+enum PackageContents<Prio: MempoolPriorityPolicy> {
+    /// Ordered by snark seq_no.
+    SnarkAccountUpdate(BTreeMap<u64, PackageTx<Prio>>),
+
+    /// Single standalone transaction.
+    Standalone(Option<PackageTx<Prio>>),
```
delbonis, line 98: What does `PackageContents::Standalone(None)` actually represent?

krsnapaudel:
@delbonis Is this the right direction?

Inner layer (PackageStore or similar):
```
struct PackageStore {
    packages: HashMap<PackageKey, Vec<OLTxId>>,
    tx_to_package: HashMap<OLTxId, PackageKey>,
}
```
Methods: `insert(key, txid, position)`, `remove(txid)`, `get_package(key)`, `packages()`. No knowledge of snark, `seq_no`, priority, or tx types.

Outer layer:
Decides `PackageKey` for a tx (`snark → SnarkAccountUpdate(account_id)`, `GAM → Standalone(txid)`)
Decides insertion position (snark → ordered by `seq_no`, GAM → append)
Decides package priority for block assembly (front tx's FIFO timestamp, feerate, etc.)

There are some issues with this though.

`Vec<OLTxId>` loses intrinsic ordering semantics
Snark ordering is by `seq_no`, not append order.
You’d need extra outer-side logic/state to keep vec sorted and gap-safe.

`insert(key, txid, position)` is fragile
Caller now owns correctness of position.
Invalid positions / mismatched package type become easy bugs.

crates/ol/mempool/src/package.rs
krsnapaudel: Empty Standalone  package.

storopoli:
Note that this is not in scope for testnet4 or even mainnet. We can tackle this post-mainnet by first having high-level discussions and then condensing these into an implementation spec/doc/design. Then moving to code. I'm fine in putting this on hold, but I'm also fine in keep this alive if you both make sure that this will not jeopardize or steal focus for other tasks that are needed for testnet4/mainnet.

krsnapaudel:
Due to issues raised above, we will close this now and revisit later.

PR #1407, opened 2026-02-26
STR 1202 Generic MMR database interface

krsnapaudel:
## Description

Replaces the legacy `GlobalMmrDatabase` trait with a storage-only `MmrIndexDatabase` interface, separating MMR algorithm logic from the database layer. Closes the TOCTOU race condition across all MMR read/write paths.

### Before
- `GlobalMmrDatabase` bundled MMR algorithms (append, pop, proof) into the DB trait
- Callers managed `MmrMetadata` manually
- `BitManipulatedMmrAlgorithm` scattered across `db/types` and test harnesses
- Borsh-typed `TypedMmrHandle<T>` coupled serialization to the handle
- Flat-position encoding (`u64`) leaked into public API
- Leaf count derived via binary search — O(log N) point reads per mutation
- Proof generation split across separate validate + count-read + proof-gen DB calls (race-prone)
- Block assembly used per-leaf `get_leaf_blocking` validation loops with 10 MMR-specific error variants

### After
**`MmrIndexDatabase` trait** — storage-only, 5 methods:
- `get_node(mmr_id, NodePos) -> Option<Hash>`
- `get_preimage(mmr_id, LeafPos) -> Option<Vec<u8>>`
- `get_leaf_count(mmr_id) -> u64` — O(1) from dedicated `leaf_count_tree`
- `fetch_node_paths(Vec<MmrNodePos>, preimages) -> MmrNodeTable` — transactional batch read
- `apply_update(MmrBatchWrite) -> ()` — atomic multi-tree transaction with CAS preconditions on leaf count and node hashes

**`NodePos` / `LeafPos`** — structured `(height, index)` types with private fields. No flat-position encoding in the public API.

**`MmrBatchWrite` / `BatchWrite`** — `BTreeMap`-backed atomic batch writes spanning multiple MMR namespaces. `expected_leaf_count` enables compare-and-set; preconditions fail atomically on mismatch.

**`MmrIndexManager` / `MmrIndexHandle`** — manager layer owns all orchestration:
- `append_leaf`, `pop_leaf` with `NodePos`-native algorithm
- Leaf count persisted per MMR ID; CAS (`expected_leaf_count`) prevents concurrent overwrite
- Bounded retry (max 3) on `MmrPreconditionFailed` for append/pop races
- `generate_proof_at`, `generate_proofs_at` (historical, caller-provided `at_leaf_count`)
- `generate_proofs_for(start, expected_hashes, at_leaf_count)` — contiguous validation + proof in one `fetch_node_paths` snapshot
- `generate_proofs_for_indices(indices_and_hashes, at_leaf_count)` — arbitrary-index variant
- `get_state_at(leaf_count) -> MmrStateView` (historical peaks + root)
- Async methods use `spawn_blocking`

**`MmrPayloadHandle`** — bytes-only pre-image storage. No borsh dependency; callers own serialization.

**Ops pattern** — `MmrIndexOps` via `inst_ops_simple!` macro, matching other managers.

### Race-condition closure
- **Write path**: leaf count read + node fetch + batch write are atomic via CAS preconditions in `apply_update`. Binary-search count derivation replaced with O(1) persisted count.
- **Read/proof path**: caller provides `at_leaf_count` from in-memory OL state; single `fetch_node_paths` covers validation + proof nodes atomically.
- **ASM worker**: `generate_mmr_proof_at(index, at_leaf_count)` — `at_leaf_count` derived from authoritative anchor state (`history_accumulator.num_entries()`), no split reads.
- **Block assembly**: per-leaf `get_leaf_blocking` validation loops removed; inbox proofs use `generate_proofs_for`, L1 header proofs use `generate_proofs_for_indices`. `at_leaf_count` sourced from OL in-memory state (snark account inbox MMR / `asm_manifests_mmr`).
- **Error simplification**: 10 MMR-specific `BlockAssemblyError` variants collapsed into `Db(DbError::Mmr*)`. Mempool reason mapping refined — MMR errors map to `Invalid`, `MmrPreconditionFailed` maps to `Failed` (transient).

### Removed
- `GlobalMmrDatabase` trait, `GlobalMmrManager`, `MmrHandle`, `TypedMmrHandle`
- `GlobalMmrDb` sled implementation (`global_mmr/` module)
- `BitManipulatedMmrAlgorithm`, `MmrAlgorithm` trait, `MmrMetadata`, `AppendLeafResult`, `PopLeafResult`
- `flat_pos()` / `from_flat()` on `NodePos`
- `borsh` dependency from storage crate
- `get_leaf_position` reverse hash lookup
- `derive_leaf_count_blocking` / binary-search count derivation
- `validate_inbox_entries` / `validate_l1_header_claims` (per-leaf validation loops)
- `map_l1_header_mmr_error` / `map_inbox_mmr_error` (error mapping helpers)
- 10 MMR-specific `BlockAssemblyError` variants (`L1HeaderLeafNotFound`, `InboxLeafNotFound`, `InboxMmrNodeNotFound`, `InvalidMmrRange`, etc.)

### Test harness
`TestAsmWorkerContext` migrated from in-memory `HashMap<u64, [u8; 32]> + MmrMetadata + BitManipulatedMmrAlgorithm` to a simple `Vec<[u8; 32]>` leaf list with `CompactMmr64` for proof generation.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-1202](https://alpenlabs.atlassian.net/browse/STR-1202)

[STR-1202]: https://alpenlabs.atlassian.net/browse/STR-1202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/ol/block-assembly/src/context.rs
+++ b/crates/ol/block-assembly/src/context.rs
@@ -78,12 +78,13 @@ pub trait BlockAssemblyAnchorContext: Send + Sync + 'static {
 
 /// Generates MMR proofs needed during block assembly.
 pub trait AccumulatorProofGenerator: Send + Sync + 'static {
-    /// Validates inbox message indices and generates message entry proofs.
-    fn generate_inbox_proofs(
+    /// Generates inbox message entry proofs at `at_leaf_count`.
+    fn generate_inbox_proofs_at(
         &self,
         target: AccountId,
         messages: &[MessageEntry],
         start_idx: u64,
+        at_leaf_count: u64,
     ) -> BlockAssemblyResult<Vec<MessageEntryProof>>;
 
     /// Validates claims and generates L1 header reference proofs.
@@ -141,76 +142,6 @@ impl<M, S> BlockAssemblyContext<M, S> {
             ))
         })
     }
-
-    fn validate_l1_header_claims<T: IStateAccessor>(
-        &self,
-        l1_header_refs: &[AccumulatorClaim],
-        state: &T,
-    ) -> BlockAssemblyResult<Vec<(u64, Hash)>> {
-        let mmr_num_leaves = state.asm_manifests_mmr().num_entries();
-        let mmr_start_height = asm_manifests_mmr_start_height(state)
-            .ok_or_else(|| BlockAssemblyError::Other("invalid manifests MMR start height".into()))?
-            as u64;
-
-        let mmr_handle = self.storage.mmr_index().as_ref().get_handle(MmrId::Asm);
-        let mut resolved_refs = Vec::with_capacity(l1_header_refs.len());
-        for claim in l1_header_refs {
-            let mmr_idx = self.height_to_mmr_index(claim.idx(), mmr_start_height)?;
-            if mmr_idx >= mmr_num_leaves {
-                return Err(BlockAssemblyError::L1HeaderLeafNotFound(mmr_idx));
-            }
-            let actual_hash = mmr_handle
-                .get_leaf_blocking(mmr_idx)
-                .map_err(map_l1_header_mmr_error)?
-                .ok_or(BlockAssemblyError::L1HeaderLeafNotFound(mmr_idx))?;
-            let claimed_hash = claim.entry_hash();
-            if actual_hash != claimed_hash {
-                return Err(BlockAssemblyError::L1HeaderHashMismatch {
-                    idx: mmr_idx,
-                    expected: actual_hash,
-                    actual: claimed_hash,
-                });
-            }
-            resolved_refs.push((mmr_idx, actual_hash));
-        }
-
-        Ok(resolved_refs)
-    }
-
-    fn validate_inbox_entries(
-        &self,
-        target: AccountId,
-        messages: &[MessageEntry],
-        start_idx: u64,
-    ) -> BlockAssemblyResult<()> {
-        let mmr_handle = self
-            .storage
-            .mmr_index()
-            .as_ref()
-            .get_handle(MmrId::SnarkMsgInbox(target));
-        for (offset, message) in messages.iter().enumerate() {
-            let idx = start_idx + offset as u64;
-            let expected_hash: Hash = <MessageEntry as TreeHash>::tree_hash_root(message).into();
-            let actual_hash = mmr_handle
-                .get_leaf_blocking(idx)
-                .map_err(|e| map_inbox_mmr_error(e, target))?
-                .ok_or(BlockAssemblyError::InboxLeafNotFound {
-                    idx,
-                    account_id: target,
-                })?;
-
-            if actual_hash.as_ref() != expected_hash.as_ref() {
-                return Err(BlockAssemblyError::InboxEntryHashMismatch {
-                    idx,
-                    account_id: target,
-                    expected: expected_hash,
-                    actual: actual_hash,
-                });
-            }
-        }
-
-        Ok(())
-    }
 }
 
 #[async_trait]
@@ -228,7 +159,7 @@ where
             .ol_block()
             .get_block_data_async(id)
             .await
-            .map_err(BlockAssemblyError::Database)
+            .map_err(BlockAssemblyError::Db)
     }
 
     async fn fetch_state_for_tip(
@@ -250,7 +181,7 @@ where
             .storage
             .asm()
             .fetch_most_recent_state()
-            .map_err(BlockAssemblyError::Database)?
+            .map_err(BlockAssemblyError::Db)?
         {
             Some((commitment, _)) => commitment.height(),
             None => return Ok(Vec::new()),
@@ -267,9 +198,9 @@ where
                 .l1()
                 .get_block_manifest_at_height_async(height)
                 .await
-                .map_err(BlockAssemblyError::Database)?
+                .map_err(BlockAssemblyError::Db)?
                 .ok_or_else(|| {
-                    BlockAssemblyError::Database(DbError::Other(format!(
+                    BlockAssemblyError::Db(DbError::Other(format!(
                         "L1 block manifest not found at height {height}"
                     )))
                 })?;
@@ -301,88 +232,44 @@ where
     }
 }
 
-/// Convert MMR-related database errors to appropriate block assembly errors for L1 header proofs.
-fn map_l1_header_mmr_error(e: DbError) -> BlockAssemblyError {
-    match e {
-        DbError::MmrLeafNotFound(idx) => BlockAssemblyError::L1HeaderLeafNotFound(idx),
-        DbError::MmrInvalidRange { start, end } => {
-            BlockAssemblyError::InvalidMmrRange { start, end }
-        }
-        DbError::MmrIndexOutOfRange { requested, cur } => {
-            BlockAssemblyError::L1HeaderMmrIndexOutOfRange {
-                requested,
-                current: cur,
-            }
-        }
-        DbError::MmrPayloadNotFound(pos) => {
-            BlockAssemblyError::L1HeaderMmrPayloadNotFound(pos.index())
-        }
-        other => BlockAssemblyError::Database(other),
-    }
-}
-
-/// Convert MMR-related database errors to appropriate block assembly errors for inbox proofs.
-fn map_inbox_mmr_error(e: DbError, account_id: AccountId) -> BlockAssemblyError {
-    match e {
-        DbError::MmrLeafNotFound(idx) => BlockAssemblyError::InboxLeafNotFound { idx, account_id },
-        DbError::MmrLeafNotFoundForAccount(idx, account_id) => {
-            BlockAssemblyError::InboxLeafNotFound { idx, account_id }
-        }
-        DbError::MmrNodeNotFound(pos) => {
-            BlockAssemblyError::InboxMmrNodeNotFound { pos, account_id }
-        }
-        DbError::MmrInvalidRange { start, end } => BlockAssemblyError::InboxMmrInvalidRange {
-            start,
-            end,
-            account_id,
-        },
-        DbError::MmrIndexOutOfRange { requested, cur } => {
-            BlockAssemblyError::InboxMmrIndexOutOfRange {
-                requested,
-                current: cur,
-                account_id,
-            }
-        }
-        DbError::MmrPayloadNotFound(pos) => BlockAssemblyError::InboxMmrPayloadNotFound {
-            idx: pos.index(),
-            account_id,
-        },
-        other => BlockAssemblyError::Database(other),
-    }
-}
-
 impl<M, S> AccumulatorProofGenerator for BlockAssemblyContext<M, S>
 where
     M: Send + Sync + 'static,
     S: Send + Sync + 'static,
 {
-    fn generate_inbox_proofs(
+    fn generate_inbox_proofs_at(
         &self,
         target: AccountId,
         messages: &[MessageEntry],
         start_idx: u64,
+        at_leaf_count: u64,
     ) -> BlockAssemblyResult<Vec<MessageEntryProof>> {
         if messages.is_empty() {
             return Ok(Vec::new());
         }
 
-        // Get MMR handle for this account's inbox
         let mmr_handle = self
             .storage
             .mmr_index()
             .as_ref()
             .get_handle(MmrId::SnarkMsgInbox(target));
-
-        self.validate_inbox_entries(target, messages, start_idx)?;
-
-        // Generate proofs for the range of messages (end index is inclusive).
-        let end_idx = start_idx + messages.len() as u64 - 1;
-        let inbox_leaf_count = mmr_handle
-            .get_num_leaves_blocking()
-            .map_err(|e| map_inbox_mmr_error(e, target))?;
+        let expected_hashes: Vec<Hash> = messages
+            .iter()
+            .map(|message| <MessageEntry as TreeHash>::tree_hash_root(message).into())
+            .collect();
         let merkle_proofs = mmr_handle
-            .generate_proofs_at(start_idx, end_idx, inbox_leaf_count)
-            .map_err(|e| map_inbox_mmr_error(e, target))?;
+            .generate_proofs_for(start_idx, &expected_hashes, at_leaf_count)
+            .map_err(|err| match err {
+                DbError::MmrLeafHashMismatch { idx, expected, got } => {
+                    BlockAssemblyError::InboxEntryHashMismatch {
+                        idx,
+                        account_id: target,
+                        expected,
+                        actual: got,
+                    }
+                }
+                other => BlockAssemblyError::Db(other),
+            })?;
 
         // Verify we got the expected number of proofs
         if merkle_proofs.len() != messages.len() {
@@ -410,22 +297,41 @@ where
         l1_header_refs: &[AccumulatorClaim],
         state: &T,
     ) -> BlockAssemblyResult<LedgerRefProofs> {
-        let mmr_handle = self.storage.mmr_index().as_ref().get_handle(MmrId::Asm);
-        let mmr_num_leaves = state.asm_manifests_mmr().num_entries();
-
-        let resolved_refs = self.validate_l1_header_claims(l1_header_refs, state)?;
-
-        // Generate proofs using MMR leaf indices
-        let mut l1_header_proofs = Vec::new();
-        for (mmr_idx, entry_hash) in resolved_refs {
-            let merkle_proof = mmr_handle
-                .generate_proof_at(mmr_idx, mmr_num_leaves)
-                .map_err(map_l1_header_mmr_error)?;
-
-            let mmr_proof = MmrEntryProof::new(entry_hash, merkle_proof);
-            l1_header_proofs.push(mmr_proof);
+        if l1_header_refs.is_empty() {
+            return Ok(LedgerRefProofs::new(Vec::new()));
         }
 
+        let mmr_handle = self.storage.mmr_index().as_ref().get_handle(MmrId::Asm);
+        let at_leaf_count = state.asm_manifests_mmr().num_entries();
+        let mmr_start_height = asm_manifests_mmr_start_height(state)
+            .ok_or_else(|| BlockAssemblyError::Other("invalid manifests MMR start height".into()))?
+            as u64;
+        let indices_and_hashes = l1_header_refs
+            .iter()
+            .map(|claim| {
+                let mmr_idx = self.height_to_mmr_index(claim.idx(), mmr_start_height)?;
+                Ok((mmr_idx, claim.entry_hash()))
+            })
+            .collect::<BlockAssemblyResult<Vec<_>>>()?;
+
+        let merkle_proofs = mmr_handle
+            .generate_proofs_for_indices(&indices_and_hashes, at_leaf_count)
+            .map_err(|err| match err {
+                DbError::MmrLeafHashMismatch { idx, expected, got } => {
+                    BlockAssemblyError::L1HeaderHashMismatch {
+                        idx,
+                        expected,
+                        actual: got,
+                    }
+                }
+                other => BlockAssemblyError::Db(other),
+            })?;
+
+        let l1_header_proofs = indices_and_hashes
+            .into_iter()
+            .zip(merkle_proofs)
+            .map(|((_, entry_hash), merkle_proof)| MmrEntryProof::new(entry_hash, merkle_proof))
+            .collect();
         Ok(LedgerRefProofs::new(l1_header_proofs))
     }
 }
@@ -550,7 +456,7 @@ mod tests {
                     idx: 0,
                     expected,
                     actual
-                } if expected == expected_hash && actual == wrong_hash
+                } if expected == wrong_hash && actual == expected_hash
             ),
             "Expected L1HeaderHashMismatch, got: {:?}",
             err
@@ -581,8 +487,12 @@ mod tests {
         let err = result.unwrap_err();
         let expected_mmr_idx = nonexistent_height - 1; // offset = genesis_height(0) + 1
         assert!(
-            matches!(&err, BlockAssemblyError::L1HeaderLeafNotFound(idx) if *idx == expected_mmr_idx),
-            "Expected L1HeaderLeafNotFound error, got: {:?}",
+            matches!(
+                &err,
+                BlockAssemblyError::Db(DbError::MmrIndexOutOfRange { requested, cur })
+                    if *requested == expected_mmr_idx && *cur == 1
+            ),
+            "Expected Db(MmrIndexOutOfRange) error, got: {:?}",
             err
         );
     }
@@ -600,8 +510,14 @@ mod tests {
         assert!(result.is_err(), "Should fail when MMR is empty");
         let err = result.unwrap_err();
         assert!(
-            matches!(err, BlockAssemblyError::L1HeaderLeafNotFound(0)),
-            "Expected L1HeaderLeafNotFound, got: {:?}",
+            matches!(
+                err,
+                BlockAssemblyError::Db(DbError::MmrIndexOutOfRange {
+                    requested: 0,
+                    cur: 0
+                })
+            ),
+            "Expected Db(MmrIndexOutOfRange {{ requested: 0, cur: 0 }}), got: {:?}",
             err
         );
     }
@@ -640,7 +556,7 @@ mod tests {
 
         let ctx = create_test_context(storage);
 
-        let result = ctx.generate_inbox_proofs(account_id, &entries, 0);
+        let result = ctx.generate_inbox_proofs_at(account_id, &entries, 0, entries.len() as u64);
 
         assert!(
             result.is_ok(),
@@ -660,7 +576,7 @@ mod tests {
 
         let ctx = create_test_context(storage);
 
-        let result = ctx.generate_inbox_proofs(account_id, &[], 0);
+        let result = ctx.generate_inbox_proofs_at(account_id, &[], 0, 0);
 
         assert!(result.is_ok(), "Should succeed with empty messages");
         let proofs = result.unwrap();
@@ -686,7 +602,8 @@ mod tests {
 
         // Request proofs starting at index 2 for last 2 messages
         let messages_to_prove = &entries[2..];
-        let result = ctx.generate_inbox_proofs(account_id, messages_to_prove, 2);
+        let result =
+            ctx.generate_inbox_proofs_at(account_id, messages_to_prove, 2, entries.len() as u64);
 
         assert!(
             result.is_ok(),
@@ -708,7 +625,7 @@ mod tests {
         let ctx = create_test_context(storage);
 
         let messages = vec![create_test_message(1, 1, 1000)];
-        let result = ctx.generate_inbox_proofs(account_id, &messages, 0);
+        let result = ctx.generate_inbox_proofs_at(account_id, &messages, 0, 0);
 
         assert!(result.is_err(), "Should fail when MMR has no messages");
     }
@@ -727,13 +644,17 @@ mod tests {
         let claimed_messages = vec![create_test_message(2, 2, 2000)];
         let ctx = create_test_context(storage);
 
-        let result = ctx.generate_inbox_proofs(account_id, &claimed_messages, 5);
+        let result = ctx.generate_inbox_proofs_at(account_id, &claimed_messages, 5, 1);
 
         assert!(result.is_err(), "Should fail for missing inbox index");
         let err = result.unwrap_err();
         assert!(
-            matches!(err, BlockAssemblyError::InboxLeafNotFound { .. }),
-            "Expected InboxLeafNotFound, got: {:?}",
+            matches!(
+                err,
+                BlockAssemblyError::Db(DbError::MmrIndexOutOfRange { .. })
+                    | BlockAssemblyError::Db(DbError::MmrLeafNotFound(_))
+            ),
+            "Expected Db(MmrIndexOutOfRange|MmrLeafNotFound), got: {:?}",
             err
         );
     }
@@ -752,7 +673,7 @@ mod tests {
         let claimed_messages = vec![create_test_message(2, 2, 2000)];
         let ctx = create_test_context(storage);
 
-        let result = ctx.generate_inbox_proofs(account_id, &claimed_messages, 0);
+        let result = ctx.generate_inbox_proofs_at(account_id, &claimed_messages, 0, 1);
 
         assert!(
             result.is_err(),
--- a/crates/ol/block-assembly/src/block_assembly.rs
+++ b/crates/ol/block-assembly/src/block_assembly.rs
@@ -8,7 +8,7 @@ use std::{
 use strata_config::SequencerConfig;
 use strata_db_types::errors::DbError;
 use strata_identifiers::{Epoch, OLBlockCommitment, OLTxId, Slot};
-use strata_ledger_types::IStateAccessor;
+use strata_ledger_types::{IAccountState, ISnarkAccountState, IStateAccessor};
 use strata_ol_chain_types_new::{
     BlockFlags, OLBlockBody, OLBlockHeader, OLL1ManifestContainer, OLL1Update, OLTransaction,
     OLTxSegment, SnarkAccountUpdateTxPayload, TransactionPayload,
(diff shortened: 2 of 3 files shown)


crates/storage/src/managers/global_mmr.rs
```diff
@@ -83,35 +129,100 @@ impl GlobalMmrManager {
...
+
+        let run = |tx: &dyn GenericMmrNodeDatabaseTx| -> DbResult<()> {
+            let metadata = tx
+                .get_mmr_metadata(mmr_id.clone())?
+                .unwrap_or_else(MmrMetadata::empty);
+
+            let result = BitManipulatedMmrAlgorithm::append_leaf(hash.0, &metadata, |pos| {
+                tx.get_mmr_node(mmr_id.clone(), pos)?
+                    .map(|h| h.0)
+                    .ok_or_else(|| DbError::Other(format!("MMR node not found at pos {}", pos)))
```
delbonis: Make an error variant for this.

crates/storage/src/managers/global_mmr.rs
```diff
@@ -83,35 +129,100 @@ impl GlobalMmrManager {
...
+            .unwrap_or_else(MmrMetadata::empty))
+    }
+
+    fn append_leaf_internal_blocking(
+        &self,
+        hash: Hash,
+        preimage: Option<Vec<u8>>,
+    ) -> DbResult<u64> {
+        let mmr_id = self.mmr_id_bytes();
+        let leaf_index = Cell::new(0u64);
```
delbonis: Could have just made the tx closure return this value instead of needing to do this roundabout thing.

krsnapaudel: No longer relevant.

crates/storage/src/managers/global_mmr.rs
krsnapaudel: Now using `MmrIndexManagerError::MissingNodeAtPos`.

crates/db/types/src/types.rs
```diff
@@ -391,6 +393,188 @@ impl MempoolTxData {
...
+impl MmrIndexDbBatchWrite {
+    pub fn with_write(mut self, op: MmrIndexWriteOp) -> Self {
+        self.writes.push(op);
+        self
+    }
+
+    pub fn with_precondition(mut self, precondition: MmrIndexPrecondition) -> Self {
+        self.preconditions.push(precondition);
+        self
+    }
```
delbonis: This isn't a builder, `with_`-style fns don't improve ergonomics here with how it's like to be used.

crates/db/types/src/types.rs
```diff
@@ -391,6 +393,188 @@ impl MempoolTxData {
...
+        leaf_index: u64,
+    },
+}
+
+/// Atomic write unit for MMR indexing updates.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct MmrIndexDbBatchWrite {
+    pub writes: Vec<MmrIndexWriteOp>,
+    pub preconditions: Vec<MmrIndexPrecondition>,
+}
```
delbonis: This design is too complex, we don't need to care about the order, just the values being written.  This should be centered around `BTreeMap`s.

Also, I'm thinking that maybe we should split it up in two parts for ergonomic reasons so that we can have a single-MMR write batch that we can use in the single-MMR handle, but then the multiple-MMR write batch is composed out of just a table of these.

crates/db/types/src/types.rs
```diff
@@ -391,6 +393,188 @@ impl MempoolTxData {
...
+    },
+}
+
+/// Write operation for an atomic MMR indexing batch.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum MmrIndexWriteOp {
+    SetLeafCount {
+        mmr_id: RawMmrId,
+        leaf_count: u64,
+    },
```
delbonis: We don't need to store the leaf count anywhere, it's implicit by the highest index at height 0.

crates/db/types/src/types.rs
delbonis, line 1: The things added here should be in their own module since it's complex and MMR-specific.

crates/db/types/src/types.rs
```diff
@@ -391,6 +393,188 @@ impl MempoolTxData {
     }
 }
 
+/// Position of a node within a mountain, represented by `(height, index)`.
+///
+/// The index is the zero-based offset within the level at `height`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct NodePos {
+    pub height: u8,
+    pub index: u64,
```
delbonis: Don't make these pub in case we decide to change how they're encoded.

crates/db/types/src/types.rs
```diff
@@ -391,6 +393,188 @@ impl MempoolTxData {
...
+impl NodePos {
+    pub fn new(height: u8, index: u64) -> Self {
+        Self { height, index }
+    }
+
+    /// Returns this node's parent position.
+    pub fn parent(self) -> Self {
+        Self {
+            height: self.height + 1,
+            index: self.index / 2,
```
delbonis: Idiomatically, this should be a `>> 1`.

crates/db/types/src/types.rs
```diff
@@ -391,6 +393,188 @@ impl MempoolTxData {
...
+    }
+
+    /// Converts a flat post-order MMR position to a structured position.
+    pub fn from_flat(flat_pos: u64) -> Self {
+        let height = pos_height_in_tree(flat_pos);
+        let leftmost_leaf_pos = leftmost_leaf_flat_pos(flat_pos, height);
+        let leftmost_leaf_index = leaf_pos_to_index(leftmost_leaf_pos);
+        let index = leftmost_leaf_index >> height;
+        Self { height, index }
+    }
```
delbonis: What is a "flat pos"?  Does this need to be exposed in the public interface?

crates/db/tests/src/mmr_index_tests.rs
```diff
@@ -0,0 +1,235 @@
...
+        .with_write(MmrIndexWriteOp::PutNode {
+            mmr_id: mmr_id.clone(),
+            pos,
+            hash,
+        })
+        .with_write(MmrIndexWriteOp::PutPayload {
+            mmr_id: mmr_id.clone(),
+            leaf_index: 0,
+            payload: payload.clone(),
+        });
```
delbonis: Should have helpers to improve ergonomics of this.

crates/storage/src/managers/mmr_index.rs
delbonis, line 1: GitHub is making the product worse so I can't review this normally, but comments summary:

* This should use the ops interface pattern.
* I think `MmrStateView` probably should be in the db types crate.  Also, what is this `root` field?
* `bag_peaks` doesn't take `&self` and isn't a constructor so should be a module-level fn.  Also, what does this actually do / is used for?  Is this defined anywhere?  It seems like it's just doing a flat hash of the peaks, but do we actually use that anywhere?  That feels like it could be an issue because it could lead to confusion about peaks.
* `node_getter` should be a `make_node_getter`
* Good example of the `with_`-style helpers actually worsening ergonomics in `pop_leaf_blocking` and `append_leaf_internal_blocking`, maybe others.
* `append_blocking` returns a `CodecError` but is actually doing a borsh decode, which is misleading.
* `get_blocking` returns an `Other` error, which should be its own variant
* `get` blocks the async executor by calling `get_blocking`, which is incorrect, make sure I'm not missing any other cases of this.
* Many functions use opaque/unclear `index` arguments instead of `NodePos`, which is what should be in the public interface.
* `MmrIndexManager` doesn't have a way to apply batches across multiple MMRs.

crates/db/types/src/types.rs
krsnapaudel: Using direct methods now.

crates/db/types/src/types.rs
krsnapaudel: Not needed anymore. It was added so the sled `MmrIndexDatabase` implementation could convert flat positions (used as the key in the existing `(mmr_id, u64)` node schema) back into `NodePos` when reading from the database.

crates/db/tests/src/mmr_index_tests.rs
krsnapaudel: `MmrIndexWriteOp` builder pattern is gone. The current API:

```
// Preconditions go on MultiMmrBatch directly
batch.add_precondition(MmrIndexPrecondition::Node { ... });

// Writes go through entry() which returns &mut MmrBatch
let mmr_batch = batch.entry(mmr_id);
mmr_batch.put_node(pos, hash);
mmr_batch.put_payload(leaf_pos, payload);
mmr_batch.del_node(pos);
mmr_batch.del_payload(leaf_pos);
```

crates/storage/src/managers/mmr_index.rs
krsnapaudel: Thanks @delbonis. I have tried to address all of your comments. PTAL.

crates/db/types/src/types.rs
krsnapaudel: Split into single-MMR + multi-MMR composition
Single-MMR unit: `MmrBatch`
Multi-MMR unit: `MultiMmrBatch { batches: BTreeMap<RawMmrId, MmrBatch>, ... }`
So multi-batch is literally a table of per-MMR batches, as requested.

crates/db/types/src/lib.rs
```diff
@@ -2,7 +2,7 @@
 
 pub mod chainstate;
 pub mod errors;
-pub mod mmr_helpers;
+pub mod mmr_index;
```
delbonis: We have a `pub use` so this is not necessary.  The other `pub use`s should be removed too but not in this PR.

crates/storage/src/managers/mmr_index.rs
```diff
@@ -0,0 +1,373 @@
...
+}
+
+#[expect(
+    missing_debug_implementations,
+    reason = "Inner MmrIndexHandle doesn't have Debug implementation"
+)]
+pub struct MmrPayloadHandle<H = Sha256Hasher> {
+    handle: MmrIndexHandle,
+    _phantom: PhantomData<H>,
+}
```
delbonis: Why is this a separate type if it is just a wrapper around `MmrIndexHandle`?

crates/storage/src/managers/mmr_index.rs
```diff
@@ -0,0 +1,373 @@
...
+        };
+
+        let mut batch = MultiMmrBatch::default();
+        let mmr_id = self.mmr_id_bytes();
+
+        batch.add_precondition(MmrIndexPrecondition::Node {
+            mmr_id: mmr_id.clone(),
+            pos: result.leaf_pos.node_pos(),
+            expected: Some(result.leaf_hash),
+        });
```
delbonis: We need to add preconditions for each of its parents up the tree.

We should have a constructor for this type like `from_preconds_table(MmrNodeTable)` that initializes it with all of the nodes we just fetched.

crates/db/types/src/mmr_index.rs
```diff
@@ -0,0 +1,656 @@
...
+    }
+
+    /// Returns this node's parent position.
+    pub fn parent(self) -> Self {
+        debug_assert!(self.height < u8::MAX, "NodePos::parent height overflow");
+        Self {
+            height: self.height + 1,
+            index: self.index >> 1,
+        }
+    }
```
delbonis: Maybe have `parent` return `Option<Self>` and have a `parent_unchecked` which just does this panic if it can't be constructed.

crates/db/types/src/mmr_index.rs
```diff
@@ -0,0 +1,656 @@
...
+        mmr_id: RawMmrId,
+        pos: NodePos,
+        expected: Option<Hash>,
+    },
+    /// Expected current payload at `pos`. `None` means key absent.
+    Payload {
+        mmr_id: RawMmrId,
+        pos: LeafPos,
+        expected: Option<Vec<u8>>,
+    },
```
delbonis, line 210: Can just split this up into two tables in the batch type.

crates/db/types/src/mmr_index.rs
```diff
@@ -0,0 +1,656 @@
...
+    /// Returns a mutable reference to the [`MmrBatch`] for `mmr_id`,
+    /// creating an empty one if absent.
+    pub fn entry(&mut self, mmr_id: RawMmrId) -> &mut MmrBatch {
+        self.batches.entry(mmr_id).or_default()
+    }
+
+    /// Adds a precondition that must hold before any writes are applied.
+    pub fn add_precondition(&mut self, precondition: MmrIndexPrecondition) {
+        self.preconditions.push(precondition);
+    }
```
delbonis: Have helpers like this that call this function:

* `add_node_precond(&mut self, mmr: MmrId, pos: NodePos, val: Buf32)`
* `add_preimg_precond(&mut self, mmr: MmrId, idx: LeafPos, val: Vec<u8>)`

crates/db/types/src/mmr_index.rs
```diff
@@ -0,0 +1,656 @@
...
+
+/// Multi-MMR atomic write batch.
+///
+/// Accumulates per-namespace [`MmrBatch`]es and global preconditions for
+/// atomic cross-MMR writes. Fields are private to allow the encoding to
+/// change without breaking callers.
+#[derive(Debug, Default, Clone)]
+pub struct MultiMmrBatch {
+    batches: BTreeMap<RawMmrId, MmrBatch>,
+    preconditions: Vec<MmrIndexPrecondition>,
```
delbonis: Should be a part of `MmrBatch`.

crates/db/types/src/mmr_index.rs
```diff
@@ -0,0 +1,656 @@
...
+        expected: Option<Vec<u8>>,
+    },
+}
+
+/// Single-MMR atomic write batch.
+///
+/// Used by `MmrIndexHandle` to build writes for one MMR namespace. Fields are
+/// private to allow the encoding to change without breaking callers.
+#[derive(Debug, Default, Clone)]
+pub struct MmrBatch {
```
delbonis: "batch write" is more clear

crates/ol/block-assembly/src/context.rs
delbonis, line 1: There should just be a `Db(#[from] DbError)` variant on `BlockAssemblyError` instead of manually needing to describe all the conversions, unless there really is some special casing somewhere.

crates/storage/src/managers/mmr_algorithm.rs
delbonis, line 1: The functions in this module accept a closure to make the low-level database query, but we don't actually have to do this.  In all cases, the nodes being queried are not actually data-dependent, so we can precompute the nodes we need to fetch without actually touching the database.  Then we can use the batch read call (see ticket) to fetch *all* of the nodes upfront in a single operation (avoids a race condition, see slack) and compute whatever we have to, then produce the write batch.  So there only needs to be 2 batched calls to the database.  This is especially helpful when we are appending to multiple MMRs, since then we can aggregate the writes across all the MMRs at the end.

crates/storage/src/managers/mmr_index.rs
```diff
@@ -0,0 +1,373 @@
...
+    }
+
+    acc.into()
+}
+
+/// Read-only view of MMR state at a specific leaf count.
+#[derive(Debug, Clone)]
+pub struct MmrStateView {
+    pub leaf_count: u64,
+    pub root: Hash,
```
delbonis: What is `root`?  If this is needed for something then it must be added to the MMR spec.

crates/ol/block-assembly/src/context.rs
krsnapaudel: We do want to distinguish invalid accumulator claim on block assembly side. Actually doing inner error matching in one place to remove duplicate types.

crates/storage/src/managers/mmr_index.rs
krsnapaudel: `MmrPayloadHandle` is now deleted.

crates/storage/src/managers/mmr_index.rs
krsnapaudel: Apparently used for some internal convenience. It's confusing. So removed now.

crates/storage/src/managers/mmr_algorithm.rs
krsnapaudel: @delbonis It seems this requires leaf count to be available. Did you mean caller will provide leaf count?

Addressed this in the manager path.

1. removed DB-query closures from algorithmic loops.
2. precompute required positions first (pure arithmetic), then do one batched `fetch_node_paths(...)` read.
3.  compute in memory and emit one MmrBatchWrite, then do one atomic `apply_update(...)`.
4. `append_many` aggregates positions across MMR IDs and uses one batched read + one batched write for the whole set.

Race behavior:
If state changes between planning and write, `expected_leaf_count` + node/preimage preconditions fail with `MmrPreconditionFailed`.
Manager does bounded retry, so this is safe (retry/error), not silent corruption.

Caveat:
Write paths still do a leaf-count read before planning, so total DB calls are not always strictly two unless count is caller-provided or fetched in the same snapshot as nodes. The core node materialization + write path is already single-read/single-write.

crates/db/types/src/mmr_index.rs
```diff
@@ -0,0 +1,920 @@
...
+    pub fn node_dels(&self) -> impl Iterator<Item = NodePos> + '_ {
+        self.node_writes
+            .iter()
+            .filter_map(|(&pos, opt)| opt.is_none().then_some(pos))
+    }
+
+    /// Iterates over pending preimage writes in position order.
+    // NOTE: Returns `&Vec<u8>` (instead of `&[u8]`) so typed-sled insert callers
+    // can pass preimages through without allocating a temporary `Vec<u8>`.
+    pub fn preimage_puts(&self) -> impl Iterator<Item = (LeafPos, &Vec<u8>)> + '_ {
```
delbonis, line 362: Is this true?  The vecs returned aren't owned so they'd still have to clone them, right?

crates/storage/src/managers/mmr_algorithm.rs
```diff
@@ -0,0 +1,237 @@
...
+    positions
+}
+
+pub(crate) fn pop_fetch_positions(leaf_count: u64) -> Vec<NodePos> {
+    if leaf_count == 0 {
+        return Vec::new();
+    }
+
+    vec![LeafPos::new(leaf_count - 1).node_pos()]
+}
```
delbonis: This function names are misleading, it makes them seem like they're somehow *doing* an append or pop operation.  Add `make_` or `compute_` prefixes.

crates/storage/src/managers/mmr_algorithm.rs
```diff
@@ -0,0 +1,237 @@
...
+    pub(crate) leaf_hash: Hash,
+    pub(crate) nodes_to_remove: Vec<NodePos>,
+}
+
+fn highest_mountain_size(leaves: u64) -> u64 {
+    debug_assert!(leaves > 0, "highest_mountain_size: leaves must be > 0");
+    1u64 << (63 - leaves.leading_zeros())
+}
+
+pub(crate) fn peak_positions_for_leaf_count(leaf_count: u64) -> Vec<NodePos> {
```
delbonis: Same here, what does it mean to do "peak" or "highest" to something?

crates/storage/src/managers/mmr_algorithm.rs
```diff
@@ -0,0 +1,237 @@
...
+            "peak_positions_for_leaf_count: start_leaf + size overflow"
+        );
+        start_leaf += size;
+        remaining -= size;
+    }
+
+    peaks
+}
+
+fn mountain_peak_for_leaf(leaf_index: u64, leaf_count: u64) -> Result<NodePos, MmrError> {
```
delbonis: and here

crates/storage/src/managers/mmr_algorithm.rs
```diff
@@ -0,0 +1,237 @@
...
+
+    while current_pos != peak_pos {
+        positions.push(current_pos.neighbor());
+        current_pos = current_pos.parent_unchecked();
+    }
+
+    Ok(positions)
+}
+
+pub(crate) fn proofs_fetch_positions(
```
delbonis: here

crates/storage/src/managers/mmr_algorithm.rs
delbonis: I guess the caller could provide the leaf count since we should have that in most cases (or be able to get it from the indexer state layer), although it would be good to support both paths obviously.

crates/storage/src/managers/mmr_index.rs
```diff
@@ -0,0 +1,567 @@
...
+        // One batched write for all MMR updates.
+        self.ops.apply_update_blocking(batch)?;
+        Ok(appended_indexes)
+    }
+
+    /// Appends one leaf per distinct MMR namespace in a single read+write cycle.
+    ///
+    /// Retries boundedly on MMR precondition failures to handle concurrent writers.
+    pub fn append_many_blocking(&self, requests: Vec<MmrAppendRequest>) -> DbResult<Vec<u64>> {
+        const MAX_PRECONDITION_RETRIES: usize = 3;
```
delbonis: Must be configurable on construction.

crates/storage/src/managers/mmr_index.rs
```diff
@@ -0,0 +1,567 @@
...
+                        && attempt + 1 < MAX_PRECONDITION_RETRIES =>
+                {
+                    last_precondition_err = Some(err);
+                    continue;
+                }
+                Err(err) => return Err(err),
+            }
+        }
+        Err(last_precondition_err.unwrap_or_else(|| {
+            DbError::Other("append_many_blocking exhausted retries unexpectedly".to_string())
```
delbonis: Maybe add an error variant for exceeding the number of retries.  Doesn't need to be specific to MMRs.

crates/db/types/src/mmr_index.rs
```diff
@@ -0,0 +1,920 @@
...
+
+/// Read-side table grouped by MMR namespace.
+#[derive(Debug, Default, Clone)]
+pub struct MmrNodeTable {
+    mmrs: BTreeMap<RawMmrId, NodeTable>,
+}
+
+impl MmrNodeTable {
+    /// Returns a mutable table entry for the requested MMR namespace.
+    pub fn entry(&mut self, mmr_id: RawMmrId) -> &mut NodeTable {
```
delbonis: Should maybe be something like `get_or_create_table_mut`, to be consistent.  Just `entry` is vague.  Also add a simple non-creating `get_table_mut` to match `get_table`.

crates/db/types/src/mmr_index.rs
krsnapaudel: @delbonis Good catch on the misleading comment. The `&Vec<u8>` return type is actually intentional though — `MmrIndexPreimageSchema` has Value = `Vec<u8>`, and typed-sled's transactional `insert(&S::Key, &S::Value)` requires exactly `&Vec<u8>`. Returning `&[u8]` would force a `.to_vec()` at every insert call site. Will fix the comment to explain the real reason.

crates/storage/src/managers/mmr_algorithm.rs
krsnapaudel: Yes both paths are supported.

crates/storage/src/managers/mmr_index.rs
krsnapaudel: Did not find a reusable config, so added a retry config inside `MmrIndexManagerConfig`.

crates/db/types/src/mmr_index.rs
delbonis: Ah well that's a little annoying, maybe should do something about that in upstream in a future PR.

crates/db/store-sled/src/mmr_index/db.rs
```diff
@@ -0,0 +1,286 @@
...
+
+impl MmrIndexDb {
+    fn abort_precondition(
+        mmr_id: RawMmrId,
+        detail: String,
+    ) -> error::ConflictableTransactionError<error::Error> {
+        error::ConflictableTransactionError::Abort(error::Error::abort(
+            DbError::MmrPreconditionFailed { mmr_id, detail },
+        ))
+    }
```
delbonis: Not a member fn or constructor so should be module-level fn, named something like `make_precond_fail_abort`.

crates/db/types/src/errors.rs
```diff
@@ -129,10 +148,47 @@ pub enum DbError {
...
+    #[error("MMR position out of bounds: pos {pos}, max size {max_size}")]
+    MmrPositionOutOfBounds { pos: u64, max_size: u64 },
+
     /// Invalid MMR index range
     #[error("Invalid MMR index range: {start}..{end}")]
     MmrInvalidRange { start: u64, end: u64 },
 
+    /// MMR node not found at the given tree position.
+    #[error("MMR node not found at position ({}, {})", .0.height(), .0.index())]
+    MmrNodeNotFound(NodePos),
```
delbonis: Should define a custom `Display`/`Debug` impl for `NodePos`.

crates/db/types/src/errors.rs
```diff
@@ -103,18 +116,24 @@ pub enum DbError {
...
     CodecError(String),
 
+    #[error("serialize error {0}")]
+    SerializeError(String),
+
+    #[error("deserialize error {0}")]
+    DeserializeError(String),
+
     #[error("transaction error {0}")]
     TransactionError(String),
```
delbonis: `error` is redundant in all of these error strings and variant names, should aim to be of the form `#[error("foo: {0}")] Foo(#[from] FooError)` when wrapping subcomponent errors, if possible.

crates/db/types/src/errors.rs
```diff
@@ -129,10 +148,47 @@ pub enum DbError {
...
+    #[error("MMR index out of range: requested {requested}, current {current}")]
+    MmrIndexOutOfRange { requested: u64, current: u64 },
+
+    /// MMR preimage payload not found at leaf position.
+    #[error("MMR preimage payload not found at leaf position {0:?}")]
+    MmrPayloadNotFound(LeafPos),
+
+    /// Tree position is out of bounds for current MMR size.
+    #[error("MMR position out of bounds: pos {pos}, max size {max_size}")]
+    MmrPositionOutOfBounds { pos: u64, max_size: u64 },
```
delbonis: * `MMR leaf hash mismatch at index {idx} (expected {expected:?}, got {got:?}`
* `MMR index out of range (requested {requested}, cur {cur})`
* `LeafPos` could *also* have a better `Display`/`Debug` impl if it makes it easier
* `MMR pos out of bounds (pos {pos}, max {max})` (should this be `index`?)

crates/db/types/src/errors.rs
krsnapaudel: I think these should not be needed. Removing.

crates/db/types/src/errors.rs
krsnapaudel: @delbonis Do we need a `Debug`? I understand `Display` is needed.

crates/db/types/src/errors.rs
krsnapaudel: @delbonis Last one is about position, not index.
/// Tree position is out of bounds for current MMR size.
MmrPositionOutOfBounds { pos: u64, max_size: u64 },

There's a separate one for index.
/// Requested leaf index is out of range for current leaf count.
    #[error("MMR index out of range: requested {requested}, current {current}")]
    MmrIndexOutOfRange { requested: u64, current: u64 }

delbonis (approved):
ok looks good I think

crates/asm/worker/src/aux_resolver.rs
```diff
@@ -213,7 +220,9 @@ impl<'a> AuxDataResolver<'a> {
                     .ok_or(WorkerError::ManifestHashNotFound { index: mmr_index })?;
 
                 // Generate MMR proof for this index
-                let proof_b32 = self.context.generate_mmr_proof(mmr_index)?;
+                let proof_b32 = self
+                    .context
+                    .generate_mmr_proof_at(mmr_index, self.at_leaf_count)?;
```
prajwolrg, line 226: Why do we need `at_leaf_count` for the generation of the proof? I think having the leaf index should be enough for generating the proof.

crates/db/store-sled/src/mmr_index/db.rs
```diff
@@ -0,0 +1,284 @@
...
+    error::ConflictableTransactionError::Abort(error::Error::abort(
+        DbError::MmrPreconditionFailed { mmr_id, detail },
+    ))
+}
+
+define_sled_database!(
+    pub struct MmrIndexDb {
+        node_tree: MmrIndexNodeSchema,
+        preimage_tree: MmrIndexPreimageSchema,
+        leaf_count_tree: MmrIndexLeafCountSchema,
```
prajwolrg, line 24: Why is `leaf_count_tree`? Is it so that we can check if the proof for the index can be generated?

crates/asm/worker/src/aux_resolver.rs
krsnapaudel: `at_leaf_count` lets callers say "prove this leaf's inclusion in the MMR as it was when it had N leaves" — matching the snapshot the verifier will check against.

crates/db/store-sled/src/mmr_index/db.rs
krsnapaudel: That's one use, but the primary reason is the CAS (compare-and-set) precondition. When appending to an MMR concurrently, `expected_leaf_count` on `BatchWrite` says "only apply this if the MMR still has N leaves" — if another writer appended between your read and write, the count won't match and the transaction aborts.

crates/asm/worker/src/aux_resolver.rs
prajwolrg: Oh nice! That's really neat. The docstring does say:
```
    /// Generates an MMR proof for the given leaf index at a specific leaf count.
````
But I couldn't infer it from this. Maybe we could expand on the docstring to be more clear on this.

File: crates/db/store-sled/src/mmr_index/db.rs (hunk 32 above)
Comment (prajwolrg) [reply crates/db/store-sled/src/mmr_index/db.rs:24]: Interesting! Where do we have the need to append an MMR concurrently? I see that it's mentioned for the TOCTOU race condition. I'm curious where does it occur in our case?

File: crates/asm/worker/src/traits.rs [hunk 33]
```diff
@@ -41,10 +41,14 @@ pub trait WorkerContext {
     /// This should be called after each STF execution with the manifest hash.
     fn append_manifest_to_mmr(&self, manifest_hash: Hash) -> WorkerResult<u64>;
 
-    /// Generates an MMR proof for the given leaf index.
+    /// Generates an MMR proof for the given leaf index at a specific leaf count.
     ///
     /// Returns a Merkle proof that can be used to verify the inclusion of the leaf.
```
prajwolrg:
```suggestion
  /// Generates an MMR inclusion proof for a leaf at a specific MMR size.
  ///
  /// The `at_leaf_count` parameter specifies the number of leaves that existed
  /// in the MMR when the proof should be constructed. This allows callers to
  /// generate proofs against a historical snapshot of the MMR rather than the
  /// current state.
  ///
  /// Returns a Merkle proof that can be used by a verifier to check the leaf's
  /// inclusion against the corresponding MMR root for that snapshot.
  
```

prajwolrg (changes requested):

crates/asm/worker/src/aux_resolver.rs
prajwolrg: Maybe we could go with something like https://github.com/alpenlabs/alpen/pull/1407#discussion_r2902818680.

prajwolrg: I do wonder why we needed this? Is it got to do something with mempool design? I can't think of a reason why this is needed on ASM, because the proof generation happens exactly for the same block on which the checkpoint transaction is there.

crates/db/types/src/mmr_index.rs
```diff
@@ -0,0 +1,954 @@
...
+
+    /// Writes or overwrites preimage bytes for a leaf.
+    pub fn put_preimage(&mut self, pos: LeafPos, preimage: Vec<u8>) {
+        self.preimage_writes.insert(pos, Some(preimage));
+    }
+
+    /// Deletes preimage bytes for a leaf.
+    pub fn del_preimage(&mut self, pos: LeafPos) {
+        self.preimage_writes.insert(pos, None);
+    }
```
bewakes, line 331: I am slightly concerned about the ordering of del and put here and in other node operations. I can see that in the db implementation there's a specific order: put first and then del is done. I wonder what happens if `put(K, v) -> del(K) -> put(K, v)` is seen. Since the underlying structure is a map, I believe this only will have a single entry for put and a single entry for del. But since put is executed first and then del, the operation eventually deletes the entry, but looking at the sequence, the entry should not be deleted.

bewakes (changes requested):
Slight concern over the batch operations ordering.

crates/db/types/src/mmr_index.rs
delbonis: Since the 3rd op in that sequence overwrites the `None` in the map, this should be fine right?

crates/db/store-sled/src/mmr_index/db.rs
delbonis: I'm still pretty confident we can remove the "leaf count" tree since we can infer it from the keys that exist in the node tree, but I didn't want to keep fighting it.

krsnapaudel: @delbonis Yes we can infer leaf count. The decision to add a tree was to avoid a range scan vs a key lookup. If there's a strong preference, we can remove it. Storing it explicitly keeps the CAS check trivial: read one key, compare, done.
@prajwolrg For the current use cases, no concurrent writes are expected. Since this is a generic shared state API, we are being conservative.

crates/db/types/src/mmr_index.rs
krsnapaudel: @bewakes As @delbonis said, there should be no issues.
The storage is a single `BTreeMap<NodePos, Option<Hash>>` — not separate put and del collections.

For `put(K, v) → del(K) → put(K, v)`:
```
put(K, v) → map[K] = Some(v)
del(K) → map[K] = None
put(K, v) → map[K] = Some(v)
```

bewakes: Oh yes, totally!

crates/db/store-sled/src/mmr_index/db.rs
delbonis: Up to you guys, ig.

crates/asm/worker/src/aux_resolver.rs
krsnapaudel: @prajwolrg You are probably right for ASM case. The API is needed for block assembly when generating proofs for L1 header refs. Block assembly is working with a staging state.

delbonis (approved):
ok looks good I think

krsnapaudel:
@prajwolrg @delbonis Had to rebase again.

PR #1448, opened 2026-03-01
refactor(asm): narrow Subprotocol trait interface to L1BlockCommitment

prajwolrg:
## Description

* Replaces `AnchorState` with `L1BlockCommitment` in `Subprotocol::process_txs`, since subprotocols cannot parse the full `AnchorState` — it contains opaque state of other subprotocols. Also since the `AsmHistoryAccumulatorState` is used upfront before subprotocol-processing to construct `VerifiedAuxData`, it is not required either. The only piece subprotocols actually need is the L1 block reference being processed.
* Passes `L1BlockCommitment` (as l1ref) into `Subprotocol::process_msgs` in place of `Params`, enabling the bridge subprotocol to use the actual L1 block ID when creating withdrawal assignments instead of a zero placeholder.
* Updates all subprotocol implementations and the ASM STF pipeline to use the narrowed interface.
* Remove redundant docstring and update it for clarity.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/asm/common/src/subprotocol.rs
+++ b/crates/asm/common/src/subprotocol.rs
@@ -23,10 +23,10 @@ use crate::{
 /// and state management.
 ///
 /// The ASM filters SPS-50 transactions by subprotocol ID and passes only the relevant
-/// transactions to each subprotocol. Each subprotocol then:
+/// transactions to each subprotocol. Each subprotocol does the following, in order:
 ///
-/// 1. processes the SPS-50 transactions matching its ID to update its own state, create
-///    inter-protocol messages to be consumed by other subprotocols, and emit logs, and then
+/// 1. processes its transactions to update its own state, create inter-protocol messages to be
+///    consumed by other subprotocols, and emit logs,
 /// 2. receives incoming inter-protocol messages from other subprotocols to finalize and serialize
 ///    its state for inclusion in the global AnchorState.
 ///


crates/asm/common/src/subprotocol.rs
```diff
@@ -21,11 +22,16 @@ use crate::{
...
 ///
-/// 1. processes each new L1 block to update its own state and emit outgoing inter-protocol
-///    messages, and then
-/// 2. receives incoming messages to finalize and serialize its state for inclusion in the global
-///    AnchorState.
+/// 1. processes the SPS-50 transactions matching its ID to update its own state, create
+///    inter-protocol messages to be consumed by other subprotocols, and emit logs, and then
+/// 2. receives incoming inter-protocol messages from other subprotocols to finalize and serialize
+///    its state for inclusion in the global AnchorState.
+///
```
barakshani: I don't like how this is structured:
1. "the SPS-50 transactions matching its ID" this is redundant and explained just above.
2. "and then": I would write above "Each subprotocol does the following, in order" and remove this.

crates/asm/stf/src/stf.rs
```diff
@@ -17,29 +19,6 @@ use crate::{
...
-/// # Errors
-///
-/// This function will return an error if:
-/// - The block header fails PoW continuity validation
-/// - Subprotocol loading, processing, or finishing fails
-///
-/// # Type Parameters
-///
-/// * `S` - The ASM specification type that defines magic bytes, subprotocol behavior, and genesis
-///   configs
```
barakshani, line 42: Why have you decided to delete this?

crates/asm/subprotocols/bridge-v1/src/subprotocol.rs
```diff
@@ -74,38 +70,11 @@ impl Subprotocol for BridgeV1Subproto {
         }
     }
 
-    /// Processes transactions for the Bridge V1 subprotocol and handles expired assignment
-    /// reassignment.
-    ///
-    /// This function is the main transaction processing entry point that:
-    /// 1. Processes each transaction based on its type:
-    ///    - **Deposit transactions** (`DEPOSIT_TX_TYPE`): Validates and records Bitcoin deposits in
```
barakshani, line 82: why removing this?

crates/asm/common/src/subprotocol.rs
```diff
@@ -55,15 +61,15 @@ use crate::{
 ///     fn process_txs(
 ///         state: &mut Self::State,
 ///         txs: &[TxInputRef],
-///         anchor_pre: &AnchorState,
+///         l1ref: &L1BlockCommitment,
```
irnb, line 64: nit: l1ref naming is ambiguous, protocol txs inside the block can also carry their own L1 refs.
Maybe anchor_l1ref, stf_l1ref or etc to tie it explicitly to the block being processed?

crates/asm/common/src/errors.rs
```diff
@@ -47,4 +49,8 @@ pub enum AsmError {
     /// Wrapped error from manifest-types crate
     #[error(transparent)]
     ManifestError(#[from] AsmManifestError),
+
+    /// Failed to verify auxiliary data.
+    #[error("invalid auxiliary data")]
+    InvalidAuxData(#[from] AuxError),
```
irnb, line 55: nit: This error variant is introduced here but doesn't seem to be used anywhere in this PR. Is it intended for a follow-up?

irnb (commented):
Looks good overall, just a few nits.

crates/asm/common/src/errors.rs
prajwolrg: This is used in `compute_asm_transition`. Previously this function was panicking instead of throwing an error. Check [5bcbb97](https://github.com/alpenlabs/alpen/pull/1448/commits/5bcbb97ae443b1b1f29075757ec8a1d79caaed4b).

crates/asm/common/src/subprotocol.rs
prajwolrg: > l1ref naming is ambiguous, protocol txs inside the block can also carry their own L1 refs.

I don't follow this. `l1ref` here is the L1Block that is being processed.

crates/asm/subprotocols/bridge-v1/src/subprotocol.rs
prajwolrg: I felt this was redundant. We have abstracted out the function for processing transactions into a separate function so mentioning that here felt redundant.

crates/asm/stf/src/stf.rs
prajwolrg: I felt this was redundant. The function signature and type already encapsulates all the information.

crates/asm/common/src/subprotocol.rs
prajwolrg: Updated in [d3c96f8](https://github.com/alpenlabs/alpen/pull/1448/commits/d3c96f864498259162d708df94301762bebdd260).

barakshani (approved):

PR #1450, opened 2026-03-02
feat(asm-checkpoint): deposit bookkeeping to validate the checkpoint honorability

evgenyzdanovich:
## Description

Track successful deposits on the ASM Checkpoint Subprotocol to additionally validate the checkpoint honorability.

Bridge subproto now defines a new subproto msg (that targets checkpoint-v1) and emits those when deposits are handled. Checkpoint (v1) handles those messages and tracks total available funds in the state. That field is used for validating "honorability" of any checkpoint (against withdrawals in that checkpoint).

Also includes some bridge <-> checkpoint ASM integration tests covering deposit tracking and checkpoint validation.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

I admit asm test harness is monstrous and must be refactored. 

- [ ]  Yes
- [x]  No

## Checklist


crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -18,6 +20,13 @@ pub struct CheckpointState {
     /// Last verified checkpoint tip position.
     /// Tracks the OL state that has been proven and verified by ASM.
     pub verified_tip: CheckpointTip,
+
+    /// Cumulative available deposit value in satoshis.
+    ///
+    /// Tracks the total deposit value that has been processed by the bridge but not yet
+    /// consumed by withdrawal dispatches. Used to reject checkpoints whose withdrawal
```
barakshani:
```suggestion
    /// consumed by withdrawal dispatches. Used for rejecting checkpoints whose withdrawal

File: crates/asm/subprotocols/checkpoint/src/handler.rs [hunk 2]
```diff
@@ -44,6 +44,22 @@ pub(crate) fn handle_checkpoint_tx(
...
+                logging::warn!(
+                    epoch,
+                    available_sat = state.available_deposit_sum(),
+                    required_sat = required,
+                    "checkpoint rejected: withdrawal intents exceed available deposit backing"
+                );
+                return;
+            }
+
+            state.deduct_withdrawals(&withdrawal_intents);
```
barakshani: given the possible error a few lines below (`expect`), this deduction might be premature, no? Wouldn't it be more suitable after the `relayer.relay_msg(&bridge_msg)` for loop below?

crates/asm/subprotocols/checkpoint/src/errors.rs
```diff
@@ -77,4 +77,15 @@ pub enum InvalidCheckpointPayload {
...
     #[error("L1 height overflow: verified tip L1 height is at maximum value")]
     L1HeightOverflow,
+
+    /// Withdrawal intents exceed available deposit backing.
+    ///
+    /// The checkpoint contains withdrawal intents totaling more satoshis than the
+    /// deposits currently tracked by the checkpoint subprotocol. The checkpoint is
+    /// rejected to prevent the bridge from panicking on unassignable withdrawals.
+    #[error("insufficient deposits: available {available_sat} sat, required {required_sat} sat")]
+    InsufficientDeposits {
```
barakshani: nit
```suggestion
    InsufficientFunds {
```
or `Sats`, since you count sats and not deposits.

crates/asm/subprotocols/checkpoint/src/subprotocol.rs
```diff
@@ -94,6 +94,10 @@ impl Subprotocol for CheckpointSubprotocol {
         // so no additional validation is performed on incoming messages.
         for msg in msgs {
             match msg {
+                CheckpointIncomingMsg::DepositProcessed(amount) => {
+                    logging::info!(amount_sat = amount.to_sat(), "Recording processed deposit");
+                    state.record_deposit(*amount);
```
barakshani, line 88: why don't we use the config amount for deposits?

crates/asm/msgs/checkpoint/src/lib.rs
```diff
@@ -1,30 +1,50 @@
...
+use strata_btc_types::BitcoinAmount;
 use strata_predicate::PredicateKey;
 use strata_primitives::buf::Buf32;
 
-/// Incoming messages that the checkpoint v0 subprotocol can receive from other subprotocols.
+/// Incoming messages for checkpoint subprotocols.
+///
+/// Messages are routed to either checkpoint V0 (ID=10) or the new checkpoint (ID=1)
+/// based on their variant. Admin configuration updates target V0, while deposit
+/// notifications target the new checkpoint subprotocol.
```
prajwolrg: nit: I don't think we should hardcode the docstring with Ids. Makes the docstring out of sync if we change the constants at some later point. Let's maybe doclink this to the constant if possible, or leave it completely.

crates/asm/msgs/checkpoint/src/lib.rs
```diff
@@ -1,30 +1,50 @@
...
     // TODO: (@PG) make this directly take PredicateKey
     UpdateSequencerKey(Buf32),
+
     /// Update the rollup proving system verifying key used for Groth16 proof verification.
     UpdateCheckpointPredicate(PredicateKey),
+
+    /// Notification that a deposit has been processed by the bridge subprotocol.
+    ///
+    /// Routed to the new checkpoint subprotocol (ID=1) so it can track available deposit
+    /// value and reject checkpoints whose withdrawal intents exceed available backing.
```
prajwolrg: nit: same thing about hardcoding constants.

crates/asm/subprotocols/checkpoint/src/handler.rs
```diff
@@ -44,6 +44,22 @@ pub(crate) fn handle_checkpoint_tx(
         Ok(withdrawal_intents) => {
             logging::info!(epoch, "checkpoint validated successfully");
 
+            // Gate withdrawal dispatch on deposit availability.
+            // If the checkpoint's withdrawal intents exceed the deposits tracked so far,
+            // reject this checkpoint to prevent the bridge from panicking.
+            if !state.can_honor_withdrawals(&withdrawal_intents) {
+                let required: u64 = withdrawal_intents.iter().map(|w| w.amt().to_sat()).sum();
```
prajwolrg: nit: we seem to be computing the withdrawal sum twice. Once here for logging and once inside of the `can_honor_withdrawals` function.

prajwolrg: I think it'd be better to do this check as part of `validate_checkpoint_and_extract_withdrawal_intents` function. That way we can return an error type for this and log the error directly.

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -18,6 +20,13 @@ pub struct CheckpointState {
...
+    /// intents exceed the available deposit backing.
+    available_deposit_sum: u64,
```
prajwolrg: I think we should save this information as something like `BTreeMap<UTXOValue, count>`, and then derive the total deposit sum from there. The withdrawal request should be exactly the same value as that of the UTXOValue. 

Consider the case that somehow OL sends two withdrawal requests with denomination 5BTC, but bridge denomination is 10BTC. If there is a single 10BTC available on bridge, this will still panic.

This can also be really useful if we decide to change the bridge denomination down the line.

crates/asm/subprotocols/checkpoint/src/subprotocol.rs
prajwolrg: Related https://github.com/alpenlabs/alpen/pull/1450/changes#r2876376263.

crates/test-utils/btcio/src/client.rs
```diff
@@ -28,3 +28,21 @@ pub fn get_bitcoind_and_client() -> (Node, Client) {
...
+    }
+    let mut conf = Conf::default();
+    conf.args.push("-txindex");
+    let bitcoind = Node::with_conf("bitcoind", &conf).unwrap();
+    let url = bitcoind.rpc_url();
+    let (user, password) = get_auth(&bitcoind);
+    let auth = Auth::UserPass(user, password);
+    let client = Client::new(url, auth, None, None, None).unwrap();
+    (bitcoind, client)
+}
```
prajwolrg, line 48: nit: This repeats what's already in `get_bitcoind_and_client`. Maybe we can create a function `get_bitcoind_and_client_with_config`. that takes the config directly. Then `get_bitcoind_and_client` can use this newly created function, so can `get_bitcoind_and_client_with_txindex`.

crates/asm/msgs/checkpoint/src/lib.rs
```diff
@@ -1,30 +1,50 @@
...
+use strata_asm_txs_checkpoint::CHECKPOINT_SUBPROTOCOL_ID;
 use strata_asm_txs_checkpoint_v0::CHECKPOINT_V0_SUBPROTOCOL_ID;
+use strata_btc_types::BitcoinAmount;
 use strata_predicate::PredicateKey;
 use strata_primitives::buf::Buf32;
 
-/// Incoming messages that the checkpoint v0 subprotocol can receive from other subprotocols.
+/// Incoming messages for checkpoint subprotocols.
+///
+/// Messages are routed to either checkpoint V0 (ID=10) or the new checkpoint (ID=1)
```
irnb: I think we can remove the checkpoint v0–related logic from this file.

tests/harness/bridge.rs
prajwolrg, line 1: neat!

crates/asm/msgs/checkpoint/src/lib.rs
```diff
@@ -1,30 +1,50 @@
...
+/// based on their variant. Admin configuration updates target V0, while deposit
```
irnb: The new checkpoint subprotocol supports these variants as well.

prajwolrg (changes requested):
Nice work! Loved the tests. Most of the comments are nits. I think we need to handle the storage of `available_deposit_sum` in a more granular level.

crates/asm/msgs/checkpoint/Cargo.toml
```diff
@@ -10,6 +10,8 @@ workspace = true
 borsh.workspace = true
 
 strata-asm-common.workspace = true
+strata-asm-txs-checkpoint.workspace = true
 strata-asm-txs-checkpoint-v0.workspace = true
+strata-btc-types.workspace = true
 strata-predicate.workspace = true
 strata-primitives.workspace = true
```
irnb, line 16: In `crates/asm/msgs/bridge`, we’re using `strata-primitives` for `BitcoinAmount`. 
It’s basically a re-export from `strata-btc-types` (where you added it). Maybe we should unify the import source for `BitcoinAmount` in both places.

crates/asm/subprotocols/checkpoint/src/handler.rs
```diff
@@ -44,6 +44,22 @@ pub(crate) fn handle_checkpoint_tx(
         Ok(withdrawal_intents) => {
             logging::info!(epoch, "checkpoint validated successfully");
 
+            // Gate withdrawal dispatch on deposit availability.
+            // If the checkpoint's withdrawal intents exceed the deposits tracked so far,
+            // reject this checkpoint to prevent the bridge from panicking.
+            if !state.can_honor_withdrawals(&withdrawal_intents) {
```
irnb: Since this is one of the checkpoint validity criteria, maybe we could move this check into the `validate_checkpoint_and_extract_withdrawal_intents` method.

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -81,4 +96,29 @@ impl CheckpointState {
...
+    }
+
+    /// Deducts the total withdrawal amount from the available deposit sum.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the total withdrawal exceeds the available deposit sum. Callers must verify
+    /// with [`can_honor_withdrawals`](Self::can_honor_withdrawals) first.
+    pub(crate) fn deduct_withdrawals(&mut self, withdrawal_intents: &[WithdrawOutput]) {
+        let total_withdrawal: u64 = withdrawal_intents.iter().map(|w| w.amt().to_sat()).sum();
```
irnb: I think we can avoid recalculating `total_withdrawal` here. we already have that result from `can_honor_withdrawals`.

crates/asm/subprotocols/checkpoint/src/handler.rs
irnb: Oh, I just saw Prajwol left the same comments, I’ll close similar one.

delbonis (dismissed):

crates/asm/subprotocols/checkpoint/src/handler.rs
evgenyzdanovich: the error is practically impossible (we have to make interfaces better on the ASM side, logs construction should be really infallible).

Given doesn't matter on the correctness, I like it better as it is now - first we handle state, then we send messages & emit logs. Intervening logs emission / message mending in between state changes _would make_ the code reader they are somehow related/dependent, while they are, in fact, "orthogonal".

crates/asm/subprotocols/checkpoint/src/subprotocol.rs
evgenyzdanovich: >why don't we use the config amount for deposits?

I'm not sure what are you referring to @barakshani 

are you talking about denomination from _bridge_ subprotocol config? 
if so - checkpoint isn't aware about it (and likely shouldn't be), the whole point is to separation of concerns and expose only minimal necessary info from one subproto to another.

tests/harness/bridge.rs
evgenyzdanovich: I admit it currently not ideal, but thanks! 
P.S. I have strong plans to reconcile all the ASM harnesses and asm testing utils into something nicer, with better ergonomics (as part of ASM testing epic).

crates/asm/subprotocols/checkpoint/src/handler.rs
evgenyzdanovich: strongly agree on the latter!

I didn't realize kinda a greater plan until today's sync - in light the task you mentioned today - generic pure verification for checkpoints (STR-2220). So for now all the checkpoint validations are kept there in `validate_checkpoint_and_extract_withdrawal_intents`. 

That function can be potentially used as is (in various places), or generalized somehow (to accomodate different data sources / contexts / etc).

Addressed in the last commit.

crates/asm/subprotocols/checkpoint/src/state.rs
evgenyzdanovich: thanks, nice.

crates/test-utils/btcio/src/client.rs
evgenyzdanovich: agree. I plan to reconcile all the harness / test-utils (and infra) as part of ASM testing epic.
ACK, for now it does the needed thing (and doesnt index nonwallet tx when it's not necessary).

crates/asm/msgs/checkpoint/src/lib.rs
evgenyzdanovich: The deposit tracking / honorability of checkpoint (basically current changes in this PR) are NOT implemented for v0 (intentionally, since we are going to get rid of it eventually).

But the problem is msg struct (`type Msg = CheckpointIncomingMsg;`) is reused across checkpoint v0 and v1, so we can't remove it now and it must be removed together with removal of v0.

crates/asm/msgs/checkpoint/src/lib.rs
evgenyzdanovich: removed

crates/asm/msgs/checkpoint/src/lib.rs
evgenyzdanovich: adjusted the comment

crates/asm/msgs/checkpoint/Cargo.toml
evgenyzdanovich: oh yeah! thanks for catching it, appreciate. with that we are avoiding adding a new dep. the btc amount is provided through `strata-primitives`

crates/asm/subprotocols/checkpoint/src/errors.rs
```diff
@@ -83,8 +83,8 @@ pub enum InvalidCheckpointPayload {
     /// The checkpoint contains withdrawal intents totaling more satoshis than the
     /// deposits currently tracked by the checkpoint subprotocol. The checkpoint is
     /// rejected to prevent the bridge from panicking on unassignable withdrawals.
-    #[error("insufficient deposits: available {available_sat} sat, required {required_sat} sat")]
-    InsufficientDeposits {
+    #[error("insufficient funds: available {available_sat} sat, required {required_sat} sat")]
+    InsufficientFunds {
         available_sat: u64,
         required_sat: u64,
     },
```
prajwolrg: I think we should also now have better error message for this to better reflect what exactly is happening.

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -97,28 +105,157 @@ impl CheckpointState {
...
-        let total_withdrawal: u64 = withdrawal_intents.iter().map(|w| w.amt().to_sat()).sum();
-        self.available_deposit_sum = self
-            .available_deposit_sum
-            .checked_sub(total_withdrawal)
-            .expect("deduct_withdrawals called without sufficient deposit backing");
+        let funds = self
+            .compute_deducted_funds(withdrawal_intents)
+            .expect("deduct_withdrawals called without prior verification");
+        self.available_funds = funds;
+    }
```
prajwolrg: I think we should be able to enforce this at a type level, instead of assuming that the caller does it. We created `SeqNoToken` type in https://github.com/alpenlabs/alpen/pull/1355/changes/1cbe313267bc686f79b7463da9fd02b086e1612c the admin subprotocol to solve this.

Maybe we can have something similar for this. Since we are using `compute_deducted_funds` for both the `verify_can_honor_withdrawals` and again in the `deduct_withdrawals`, we can maybe create a newtype to handle this.

crates/asm/subprotocols/checkpoint/src/errors.rs
evgenyzdanovich: done

crates/asm/subprotocols/checkpoint/src/state.rs
evgenyzdanovich: sure, that kinda infers we have to (additionally) return some kind of "token" **from the validate function** that can be consumed on mutation, so I went ahead and extracted return type into a separate holder struct as well. PTAL.

crates/asm/subprotocols/checkpoint/src/state.rs
```diff
@@ -81,4 +115,147 @@ impl CheckpointState {
...
+        let mut funds = self.available_funds.clone();
+
+        let insufficient = || InvalidCheckpointPayload::InsufficientFunds {
+            available_sat: self.available_deposit_sum(),
+            required_sat: withdrawal_intents.iter().map(|w| w.amt().to_sat()).sum(),
+        };
+
+        for intent in withdrawal_intents {
+            let denom = intent.amt();
+            let count = funds.get_mut(&denom).ok_or_else(&insufficient)?;
```
barakshani, line 151: This is tricky. IIUC, we start with a map (log_amount --> num_utxo_with_amount), where right now we support a single log_amount. Here, again IIUC, we first lock in the amount, then we check how many utxo's we have, for _that_ amount.

However, it's possible that we have other utxo denominations that sums up to this amount. For example, suppose we have `d= {5,10}` and user withdraws 10 btc. We may only have 5btc UTXOs, but we would fail on this check.

Sense we don't have a real design for multi-denomination deposits, hard to say how well the logic here will play in such world, however I think it's a fair assumption that when logs are being emitted by the OL, the OL may not have this mapping, hence it will have a single log of 10btc, instead of 2 logs of 5 btc.

barakshani: You even test for this (almost) : )

evgenyzdanovich: yeah, ACK.

if we end up with multi denom (or even worse, _arbitrary_ denoms), it simply does not work in general case.
The code still _intends_ to cover single denom case and hopefully can potentially cover some "unintended cases" when we can have several denominations and those can be matched exactly one-to-one. 

In fact, the problem as you described, may not be solvable with poly complexity generally; to me it closely resembles knapsack or its variation and likely may be reduced to it. So I'm pretty sure we don't wanna go that deep and support general case (on the ASM side), at least for now.

barakshani: We _definitely_ don't want to solve the very general (arbitrary) denomination case. This would be thinking too much into the future, but some small set of deposit denominations (1,2,5,10 BTC) may happen.

Anyway, I'm a bit curious now on ending up with this design if the intention is a single denomination. First, a small note is that in this case we can just count UTXOs, instead of utxo*denom. But more interesting to me is that I think the previous impl was already good enough: we track total amount, which is guaranteed to be a multiple of the fixed denom. Sounds _correct_ and _simpler_. So why reverting to this one?

prajwolrg: I suggested this approach here https://github.com/alpenlabs/alpen/pull/1450#discussion_r2876376263. We could've gotten away with just storing the count and the supported denomination for now, but this feels like a better approach longer term.

barakshani: Thanks, yeah I've seen that comment, but I trust Evgeniy to not blindly follow you : )

> but this feels like a better approach longer term

Do you still think so?

This isn't really blocking as long as we're on that same page. I would, however, add a note to the code warning the reader that it is likely not to be compatible with multi denom. Hopefully this note will be noticed when this time come.

evgenyzdanovich: ok.

evgenyzdanovich: done last commit, I've adjusted docstring mentioning limitations  - although It's been done before IMO, in the previous revision

```
 /// Verifies that the available funds can cover all withdrawal intents using exact
    /// denomination matching.
```

but maybe it wasn't clear enough.

barakshani (dismissed):

crates/asm/msgs/checkpoint/src/lib.rs
```diff
@@ -1,30 +1,46 @@
...
-        CHECKPOINT_V0_SUBPROTOCOL_ID
+        match self {
+            // Admin config updates target checkpoint V0.
+            Self::UpdateSequencerKey(_) | Self::UpdateCheckpointPredicate(_) => {
+                CHECKPOINT_V0_SUBPROTOCOL_ID
+            }
+            // Deposit notifications target the new checkpoint subprotocol.
+            Self::DepositProcessed(_) => CHECKPOINT_SUBPROTOCOL_ID,
+        }
     }
```
delbonis, line 44: This kinda makes me scared.  I made this not be just an associated const so the trait could be dyn safe but I didn't expect to actually have it be able to return different values depending on the variant.

But I guess it's fine?

delbonis (dismissed):

crates/asm/msgs/checkpoint/src/lib.rs
evgenyzdanovich: its fine with 2-phase schema we have (first we run process txs per each subproto where we accumulate messages - this is where actually we call `.id()`; and then we dispatch messages and process), meaning it should work correctly&deterministically according to protocol and how code is written.

It's certainly an loophole / "side-effect" of it not being const :) Here (and elsewhere), it shouldn't be overused too much, and I see it as a dirty hack.

The problem (and thus the hack) stems from the fact that we have two active checkpoint subprotocols (v0-aka-legacy and v1) and **those are using the same struct for incoming messages.** hence the hack, since I wanted to avoid putting extra effort into adjusting legacy code and not touch anything else.

It'll be **naturally** removed soon when we complete the migration (remove v0 entirely, forward everything to v1)  - we'll have to get rid of ID for v0 and we'll end up here. 
For now, this is a "controlled dispatch switch" for the time while we have ongoing migration.

delbonis: Right, makes sense.

evgenyzdanovich:
@irnb could you please approve - my guess if you previously requested changes for the PR, it will not automerge until you explicitly approve, even if approval coverage is achieved :)
Or somebody else with "push-bypassing-check rights" :)

irnb:
Re-running the failed test, will enable merge once it's green.

PR #1452, opened 2026-03-03
(btcio/writer): Remove 2x multiplier in fee estimation

bewakes:
## Description

It seems like the issue with excessive fees in signet is because of the existing 2x multiplier in the fee estimation. In signet, the issue most likely unfolds like this.
1. Let the existing fee rate be x.
2. When queried by the fee policy, it would be multiplied by 2 and the fee rate is interpreted as 2x.
3. Transactions are created and submitted to bitcoin with the fee rate of 2x.
4. Once the transactions are accepted, the network's fee rate becomes roughly 2x because signet mostly has empty blocks.
5. When queried by the fee policy it would again be multiplied by 2 and the new fee rate would be interpreted to be 4x, and so on.

This PR removes the 2x multiplier. The query made to `estimatesmartfee` RPC should already be enough because it returns the "conservative" estimation by default(which is slightly higher) than the other option("economical"). See https://chainquery.com/bitcoin-cli/estimatesmartfee.
The PR also adds some logging for observability of the fee rate and actual fee being paid in a transaction.
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/btcio/src/writer/builder.rs
+++ b/crates/btcio/src/writer/builder.rs
@@ -162,15 +162,15 @@ pub fn create_envelope_transactions(
     let reveal_fee = commit_value - env_config.reveal_amount;
 
     debug!(
-        commit_output_sats = commit_value,
-        reveal_amount_sats = env_config.reveal_amount,
-        reveal_fee_sats = reveal_fee,
+        commit_output_sats = %commit_value,
+        reveal_amount_sats = %env_config.reveal_amount,
+        reveal_fee_sats = %reveal_fee,
         reveal_fee_btc = format!(
             "{:.8}",
             reveal_fee as f64 / BitcoinAmount::SATS_FACTOR as f64
         ),
-        fee_rate_sats_per_byte = env_config.fee_rate,
-        reveal_script_size_bytes = reveal_script.len(),
+        fee_rate_sats_per_byte = %env_config.fee_rate,
+        reveal_script_size_bytes = %reveal_script.len(),
         "Calculated reveal transaction fees"
     );
 


crates/btcio/src/writer/builder.rs
```diff
@@ -144,6 +158,21 @@ pub fn create_envelope_transactions(
...
+        commit_output_sats = commit_value,
+        reveal_amount_sats = env_config.reveal_amount,
+        reveal_fee_sats = reveal_fee,
+        reveal_fee_btc = format!(
+            "{:.8}",
+            reveal_fee as f64 / BitcoinAmount::SATS_FACTOR as f64
+        ),
+        fee_rate_sats_per_byte = env_config.fee_rate,
+        reveal_script_size_bytes = reveal_script.len(),
+        "Calculated reveal transaction fees"
```
storopoli: One minor nit.
Can we be explicit that all these metadata will be `Display`-rendered with:

```suggestion
        commit_output_sats = %commit_value,
        reveal_amount_sats = %env_config.reveal_amount,
        reveal_fee_sats = %reveal_fee,
        reveal_fee_btc = %format!(
            "{:.8}",
            reveal_fee as f64 / BitcoinAmount::SATS_FACTOR as f64
        ),
        fee_rate_sats_per_byte = %env_config.fee_rate,
        reveal_script_size_bytes = %reveal_script.len(),
        "Calculated reveal transaction fees"
```

PS: I never know if the `%` goes on LHS or RHS. But the compiler complains if you put on the wrong side. It is like a damn USB-A plug. I always do on the wrong side all the time.

bewakes: Done in https://github.com/alpenlabs/alpen/pull/1452/commits/92e2fb2ef4f84f85eb90746ed0bcbc80a3e7ee9a

PR #1453, opened 2026-03-03
fix(btcio): prevent broadcaster crashes on `bitcoind` HTTP 500 tx rejections.

storopoli:
## Description

Now that `bitcoind-async-client` `0.10.1` is in this branch (includes [alpenlabs/bitcoind-async-client#90](https://github.com/alpenlabs/bitcoind-async-client/pull/90)), this PR is refactored to rely on typed RPC errors instead of masked HTTP 500 responses.

### What changed

- keep explicit invalid-input classification on typed RPC errors (`is_missing_or_invalid_input` and `ClientError::Server(-22, ..)`) so these are marked/resigned as before
- keep `ClientError::Status(500, ..)` only as a defensive fallback in broadcaster (`L1TxStatus::Unpublished` retry path)
- remove redundant reveal-flow `Status(500)` branch in chunked envelope handling (generic non-fatal path already tracks status)
- preserve empty-input pre-check behavior (`InvalidInputs` before broadcast)

### Tests updated

- keep broadcaster HTTP-500 fallback regression test
- add broadcaster regression test for typed `Server(-22, ..)` -> `InvalidInputs`
- repurpose chunked-envelope HTTP-500 test to typed `Server(-22, ..)` -> `NeedsResign`

### Why this refactor

`0.10.1` now surfaces Bitcoin RPC failures with their proper error codes/messages, so the old broad HTTP-500 handling is no longer the primary mechanism for classifying invalid submissions.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [x] Dependency Update

## Notes to Reviewers

- This keeps the broadcaster resilient to ambiguous transport/proxy failures while restoring typed RPC errors as the primary decision path.
- It intentionally does not change root-cause tx construction behavior (for example, genuinely invalid/empty inputs).

Done with the help of GPT-5.3 codex.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

See http://grafana.development.stratabtc.org/goto/dfewgxucwwkjke?orgId=1

### DIFF
--- a/crates/btcio/src/broadcaster/task.rs
+++ b/crates/btcio/src/broadcaster/task.rs
@@ -116,7 +116,11 @@ async fn process_unfinalized_entries(
 
 /// Takes in `[L1TxEntry]`, checks status and then either publishes or checks for confirmations and
 /// returns its new status. Returns [`None`] if status is not changed.
-#[instrument(skip_all, fields(%txid), name = "process_txentry")]
+#[instrument(
+    skip_all,
+    fields(component = "btcio_broadcaster", %txid),
+    name = "process_txentry"
+)]
 async fn process_entry(
     rpc_client: &(impl Broadcaster + Wallet),
     txentry: &L1TxEntry,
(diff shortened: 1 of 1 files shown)


irnb:
> I also added debug logs so you can confirm whether it was empty-input txs or other rejected submissions.

I’ve seen this before. In a functional test, I put a lot of pressure on the bitcoin node by rapidly sending Bitcoin transactions and reading from the node at the same time. In a weird case, the node accepted the tx in mempool, but `bitcoind-async-client` returned an HTTP 500.

Thanks for catching and fixing this!

crates/btcio/src/broadcaster/task.rs
```diff
@@ -216,6 +245,14 @@ async fn publish_tx(
     }
 }
 
+fn is_invalid_submission_error(err: &ClientError) -> bool {
+    err.is_missing_or_invalid_input()
+        || matches!(
+            err,
+            ClientError::Server(-22, _) | ClientError::Status(500, _)
```
irnb: There's a risk with mapping `ClientError::Status(500, _)` to `InvalidInputs/NeedsResign`. `bitcoind-async-client` could return `Status(500, _)` even when the tx was actually accepted into the mempool, e.g. due to a garbled HTTP response or etc which could lead to double-posting the same data on L1.

storopoli: Yeah, I don't like this. I'll refactor the PR. This was a quick logging fix for M2 deployment issues to have more logging.

storopoli: Check now in f2dc61e7d3b984c857d23e651f5b46ba6d63b334

crates/btcio/src/broadcaster/task.rs
```diff
@@ -198,17 +201,57 @@ async fn publish_tx(
     txentry: &L1TxEntry,
 ) -> BroadcasterResult<L1TxStatus> {
```
delbonis, line 216: Should probably wrap the logic of this function in a debug span that contains the txid so that we don't have to include `%txid` as a param in each log.  And sure the inputs and outputs too.

crates/btcio/src/broadcaster/task.rs
```diff
@@ -198,17 +201,57 @@ async fn publish_tx(
...
     let tx = txentry.try_to_tx().expect("could not deserialize tx");
+    let txid = tx.compute_txid();
+    let input_count = tx.input.len();
+    let output_count = tx.output.len();
+
+    if tx.input.is_empty() {
+        debug!(
+            %txid,
+            %input_count,
```
delbonis: Redundant?

storopoli: See 87609bb92a970049ae320c42646f1d8d58bd8c98

crates/btcio/src/broadcaster/task.rs
storopoli: See 87609bb92a970049ae320c42646f1d8d58bd8c98

PR #1382, opened 2026-02-20
fix(exex): use actual block number in FinishedHeight for WAL pruning

purusang:
- Fix ProverWitnessGenerator emitting FinishedHeight with block number 0 (from outcome.first_block()) instead of the real block number, which prevented WAL files from ever being pruned
- Add functional test verifying WAL files are pruned after epoch finalization

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
Here is the [slack discussion](https://alpen-labs.slack.com/archives/C08D9E3J8N5/p1771425459139759) on this issue.
Atm none of the wal files in our clusters has been deleted.
 

Is this PR addressing any specification, design doc or external reference document?

- [ ]  Yes
- [ ]  No

If yes, please add relevant links:
## AI usage
To generate func test.
## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [x] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [x] New and existing tests pass with my changes.
- [x] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

PR #1454, opened 2026-03-03
Backport: fix auto deletion of wal files

purusang:
## Description
This is a [back ported fix](https://github.com/alpenlabs/alpen/pull/1382) which has already been tested in staging.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist


storopoli:
> new `test_exex_wal_pruning` constantly fails. Does it pass locally?

Yes it does. That's why I'm stalled in this PR :/

functional-tests-new/tests/alpen_client/test_exex_wal_pruning.py
```diff
@@ -0,0 +1,166 @@
...
+                value=1_000_000_000,
+                gas_price=gas_price,
+                gas=21_000,
+            )
+            tx_hash = send_raw_transaction(ee_rpc, raw_tx)
+            wait_for_receipt(ee_rpc, tx_hash, timeout=30)
+
+        # Mine L1 blocks while polling Strata status until epoch 1 is confirmed.
+        #
+        # In functional-tests-new `el_ol`, epoch finalization may not always be available
```
barakshani: This, ideally, would be fixed in https://alpenlabs.atlassian.net/browse/STR-2424.

functional-tests-new/tests/alpen_client/test_exex_wal_pruning.py
```diff
@@ -0,0 +1,166 @@
...
+                "Skipping WAL-pruning assertion because pruning is finalization-driven."
+            )
+            return True
+
+        def wal_files_pruned() -> bool:
+            # Keep L1 moving while checking pruning so EE keeps receiving
+            # finalization-related forkchoice updates.
+            btc_rpc.proxy.generatetoaddress(2, mine_address)
+            current = _list_wal_files(wal_dir_root)
+            return len(wal_files_before - current) > 0
```
barakshani: Why do we expect this inequality to hold? we have not stopped block production, it's possible that we produced more blocks, hence more wal files than the number `wal_files_before`.

functional-tests-new/tests/alpen_client/test_exex_wal_pruning.py
```diff
@@ -0,0 +1,166 @@
...
+                gas=21_000,
+            )
+            tx_hash = send_raw_transaction(ee_rpc, raw_tx)
+            wait_for_receipt(ee_rpc, tx_hash, timeout=30)
+
+        # Mine L1 blocks while polling Strata status until epoch 1 is confirmed.
+        #
+        # In functional-tests-new `el_ol`, epoch finalization may not always be available
+        # (for example when no proving/finalization pipeline is wired in the environment),
+        # but confirmation still advances with L1 progress and is sufficient to exercise
```
barakshani: if so, we should not check for finalization, because the test may fail unjustifiably.

functional-tests-new/tests/alpen_client/test_exex_wal_pruning.py
```diff
@@ -0,0 +1,166 @@
...
+        # Best effort: observe finalization if this env supports it.
+        try:
+            status_after_finalization = wait_until_with_value(
+                lambda: _mine_and_get_sync_status(strata_seq, btc_rpc, mine_address),
+                lambda status: status["finalized"] is not None
+                and status["finalized"]["epoch"] >= 1,
+                error_with="Epoch 1 was not finalized in time",
+                timeout=120,
+                step=2,
+            )
```
purusang: @storopoli  To prune wal files, the ideal requirement should be to have the proofs to be generated and stored in db (or might only be witness stored in db 🤔 ). In releases/0.2.0 I waited for OL checkpoint finalization because that was the safest signal to cleanup wal files. And in fast mode epoch could be finalized in seconds there. So we just waited for a finalized checkpoint. And we checked if reth cleaned up wal files.

In main branch, I found out that OL checkpoints never get finalized (which @bewakes  confirmed as well) due to which this condition never passes. So I think this whole functional test can be simplified by just having two waits: 

- first we wait until we have some `n` number of wal files
- after that we wait until some of those preexisting wal files gets deleted, don't need to wait for all of them to be cleaned up.

In this test we do not care if OL checkpoint finalizes or not. We just want to make sure that the wal files are cleared by reth after sometime.

functional-tests-new/tests/alpen_client/test_exex_wal_pruning.py
purusang: this is a set difference.

barakshani: what does it mean, you take len?
if we produced one more block after the `wal_files_before` checkmark, then `current` set would have wal files that `wal_files_before` doesn't have, regardless of pruning.

purusang: I meant [set operation](https://www.geeksforgeeks.org/maths/difference-of-sets/).
If nothing gets pruned then len(A-B) will be 0 because B will have everything from A and some additional ones. 
But once some of it gets pruned, no matter how much is added, we will get len(A-B)>0 since there will be some A-only files.

> then current set would have wal files that wal_files_before doesn't have, regardless of pruning.

This is right but will be handled by Set subtraction operation.

functional-tests-new/tests/alpen_client/test_exex_wal_pruning.py
storopoli: See c92fffe5c702f0a34f2fef7e0d225aa3a8babfc4

functional-tests/tests/el_exex_wal_pruning.py
```diff
@@ -0,0 +1,114 @@
...
+        # WAL directory is under reth's datadir.
+        reth_datadir = reth.get_prop("datadir")
+        wal_dir = os.path.join(reth_datadir, "exex", "wal")
+
+        # Wait for some blocks to be produced so WAL files accumulate
+        wait_until(
+            lambda: int(rethrpc.eth_blockNumber(), base=16) > 10,
+            error_with="blocks not advancing",
+            timeout=30,
+        )
```
delbonis: This should use a waiter.

storopoli: Done, amended in 977333a71604a7f9fbdc0d8e16e5801567cdb108

PR #1456, opened 2026-03-04
fix(asm): add checkpoint_txid to CheckpointTipUpdate log

irnb:
The `CheckpointTipUpdate` log was missing the L1 transaction ID that
carried the checkpoint transaction. The `txid` is needed on the consumer                                                                         side to reference the originating L1 transaction and fetch checkpoint tx witness.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/asm/logs/src/checkpoint.rs
+++ b/crates/asm/logs/src/checkpoint.rs
@@ -83,21 +83,21 @@ impl AsmLog for CheckpointUpdate {
 
 /// Records a verified [`CheckpointTip`] update from the v1 checkpoint subprotocol.
 ///
-/// Carries the tip (epoch, L1 height, L2 commitment) along with the hash of
-/// the L1 transaction that delivered the checkpoint proof. The inner
+/// Carries the tip (epoch, L1 height, L2 commitment) along with the txid of
+/// the L1 transaction that delivered the checkpoint tx. The inner
 /// [`CheckpointTip`] is encoded via [`CodecSsz`] per its SSZ schema.
 #[derive(Debug, Clone, Codec)]
 pub struct CheckpointTipUpdate {
     /// The new verified checkpoint tip.
     tip: CodecSsz<CheckpointTip>,
 
-    /// Hash of the L1 transaction that carried the checkpoint proof.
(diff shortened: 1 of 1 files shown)


crates/asm/logs/src/checkpoint.rs
```diff
@@ -121,7 +132,8 @@ mod tests {
     fn checkpoint_tip_update_roundtrip() {
         let l2_commitment = OLBlockCommitment::new(42, OLBlockId::from(Buf32::from([0xAB; 32])));
         let tip = CheckpointTip::new(7, 100, l2_commitment);
-        let update = CheckpointTipUpdate::new(tip);
+        let txid = Buf32::from([0xCD; 32]);
```
krsnapaudel, line 135: I see tests use hardcoded values. I have tried my best to replace tests like these with proptest based tests. Can we do that?

crates/asm/logs/src/checkpoint.rs
```diff
@@ -82,27 +83,37 @@ impl AsmLog for CheckpointUpdate {
...
 
 impl CheckpointTipUpdate {
-    /// Creates a new [`CheckpointTipUpdate`] from a [`CheckpointTip`].
-    pub fn new(tip: CheckpointTip) -> Self {
+    /// Creates a new [`CheckpointTipUpdate`] from a [`CheckpointTip`] and the
+    /// raw txid bytes of the L1 transaction that carried the checkpoint.
+    pub fn new(tip: CheckpointTip, checkpoint_txid: Buf32) -> Self {
         Self {
             tip: CodecSsz::new(tip),
+            checkpoint_txid,
```
delbonis, line 104: Should this be txid or wtxid?

prajwolrg: Having `txid` should be fine. Computing `wtxid` means we would have to recompute the hash of all the witness which includes the state diff. Since the computation of wtxid is already done as part of validating the block body, it is not required.

PR #1457, opened 2026-03-05
refactor(asm): simplify subprotocol trait and params

prajwolrg:
## Description

The Subprotocol trait had a `Params` associated type that was designed to support per-block-height parameter changes, but this capability was never actually used. In practice, `Params` was only ever read during init to construct the initial state, and then threaded through every method signature (`pre_process_txs`, `process_txs`) unused — implementations just ignored them with `_params`.

So this PR removes it. 

- Remove unused `params` and `prestate` arguments from the `Subprotocol::preprocess` method
- Remove unused `params` from `process_txs`, simplifying the subprotocol trait surface
- Make `Subprotocol::init` infallible (returns `Self::State` instead of Result)
- Move `max_seq_no` from admin init config into admin subprotocol state
- Rename `Params` to `InitConfig` across all subprotocols to better reflect their purpose as one-time initialization data rather than runtime parameters
- Rename leftover _params / _genesis field names to _config for consistency

### Type of Change

- [ ] Documentation update
- [X] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/asm/common/src/subprotocol.rs
+++ b/crates/asm/common/src/subprotocol.rs
@@ -109,7 +109,6 @@ pub trait Subprotocol: 'static {
     /// * `state` - Current state of the subprotocol
     /// * `txs` - Slice of L1 transactions relevant to this subprotocol
     /// * `collector` - Interface for registering auxiliary input requirements
-    /// * `anchor_pre` - The previous anchor state for context
     fn pre_process_txs(
         _state: &Self::State,
         _txs: &[TxInputRef<'_>],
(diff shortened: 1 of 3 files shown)


delbonis (approved):

delbonis:
Little confused about removing the prestate arg from preprocess though.  I get that it's unused but it's something we would pretty plausibly need in the future.

crates/asm/common/src/subprotocol.rs
```diff
@@ -116,13 +110,10 @@ pub trait Subprotocol: 'static {
     /// * `txs` - Slice of L1 transactions relevant to this subprotocol
     /// * `collector` - Interface for registering auxiliary input requirements
     /// * `anchor_pre` - The previous anchor state for context
-    /// * `params` - Subprotocol's current params
```
storopoli: You also need to delete `anchor_pre` from the docstrings.

prajwolrg: Nice catch. Fixed in [871e608](https://github.com/alpenlabs/alpen/pull/1457/commits/871e608384dfab355a78ad5fec5e6094fdf056ca).

PR #1273, opened 2026-01-08
test fix: tx receipt timeout

barakshani:
## Description

Increase the timeout : (

I want to have it in the code to see if the probability of this error appearing is getting lower, when functional tests are being run on other PRs

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Is this PR addressing any specification, design doc or external reference document?

- [ ]  Yes
- [x]  No

If yes, please add relevant links:

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [x] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [x] New and existing tests pass with my changes.
- [x] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

## Related Issues

[STR-2050](https://alpenlabs.atlassian.net/browse/STR-2050)

[STR-2050]: https://alpenlabs.atlassian.net/browse/STR-2050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1458, opened 2026-03-05
fix(alpen-ee): set default sequencer blocktime to 5s

storopoli:
## Description

Set the EE sequencer default blocktime from 1s to 5s by updating `DEFAULT_BLOCKTIME_MS` in `crates/alpen-ee/sequencer/src/block_builder/config.rs`.

Also updates functional-tests-new to handle the new block time correctly.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This PR now includes the functional-test follow-up needed for the 5-second EE block time change.

- EE waits now derive timeout budgets from expected block counts instead of fixed 10-30s assumptions.
- DA sealing helpers and the large multi-chunk DA test use workload-aware confirmation windows.
- Targeted validation passed for `test_block_generation`, `test_fullnode_sync`, `test_block_queries`, `test_tx_forwarding`, `test_mesh_discovery`, `test_multi_chunk`, and `test_normal_state_diff`, plus `just fmt-ws` and `just lint-check-ws`.

- [ ]  Yes
- [x]  No

AI assistance notice: This PR body description was prepared with AI assistance (Cursor/Codex).

## Checklist

## Related Issues

N/A

### DIFF
--- a/functional-tests-new/tests/alpen_client/ee_da/test_multi_chunk.py
+++ b/functional-tests-new/tests/alpen_client/ee_da/test_multi_chunk.py
@@ -155,11 +155,18 @@ class TestDaMultiChunkTest(BaseTest):
             current_l2_block = sequencer.get_block_number()
             blocks_needed = expected_batch_last_block + batch_sealing_block_count
             if current_l2_block < blocks_needed:
+                # Large DA payloads slow post-batch block production on CI, so the
+                # generic 1s-per-block wait budget is too tight for this step.
+                block_wait_timeout = timeout_for_expected_blocks(
+                    blocks_needed - current_l2_block,
+                    seconds_per_block=2.0,
+                    slack_seconds=30,
+                )
                 logger.debug(
                     f"Attempt {attempt + 1}: Waiting for L2 block"
                     f" {blocks_needed} (current: {current_l2_block})"
                 )
-                sequencer.wait_for_block(blocks_needed)
+                sequencer.wait_for_block(blocks_needed, timeout=block_wait_timeout)
 
             logger.debug(f"Attempt {attempt + 1}: Waiting for DA transactions to reach mempool...")
             time.sleep(10)


functional-tests-new/common/config/constants.py
```diff
@@ -35,6 +35,12 @@
 SATS_TO_WEI = 10_000_000_000
 GWEI_TO_WEI = 1_000_000_000
 
+# =============================================================================
+# EE Timing
+# =============================================================================
+DEFAULT_EE_BLOCK_TIME_SECONDS = 5.0
```
barakshani: why? we want block time in test much faster.

functional-tests-new/common/evm_utils.py
```diff
@@ -7,21 +7,31 @@
 
 from .accounts import ManagedAccount
 from .rpc import RpcError
-from .wait import wait_until
+from .wait import timeout_for_expected_blocks, wait_until
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_RECEIPT_WAIT_BLOCKS = 3
```
barakshani: We can wait and see, but I think this is too low. In https://github.com/alpenlabs/alpen/pull/1273 I had to make it `10`, because flakiness was still observable.

functional-tests-new/common/config/constants.py
storopoli: Yes, that is exactly what I've said in https://github.com/alpenlabs/alpen/pull/1458#issuecomment-4013139408

The problem is that we don't have a "runtime" knob for the EE sequencer to change the blocktime. It is hardcoded in that constant in the Rust code.
I can repurpose this PR to expose a runtime flag either through CLI flags, ENV vars, or JSON/TOML config files for that and drop all the changes that I have done in the functional-tests-new and just change that the EE sequencer in CI should run with 1s block time.

barakshani: This should be part of the new functional tests efforts, so should definitely involve @voidash as well as @bewakes. I would leave this PR lean, and make sure this is done through the "proper" channels.

functional-tests-new/common/evm_utils.py
barakshani: oh wait, I think you introduced here a new method.

storopoli: Done in 8cbb04d418c8c0c342aefea2b1f7b5173ff67ee3
Changed to `10`.

barakshani: I was saying I misunderstood this param, probably different than what I was comparing to.

functional-tests-new/common/config/constants.py
storopoli: Solved in 2ccc6d2662d5f73f77298962fbf9323cfb02ddb1

delbonis (approved):

functional-tests-new/common/wait.py
```diff
@@ -21,14 +42,29 @@ def wait_until(
     Wait until a function call returns truth value, given time step, and timeout.
     This function waits until function call returns truth value at the interval of step seconds.
     """
-    for _ in range(math.ceil(timeout / step)):
+    deadline = time.monotonic() + timeout
```
bewakes, line 51: As far as I understand, the change being made here is to account for function calls that take longer than the step so that if it takes longer then we are not doing extra calls.

bewakes (approved):

PR #1162, opened 2025-11-24
STR 1900: SSZ support for new OL chain types

krsnapaudel:
## Description

Add ssz support for `strata-ol-chain-types-new`. Requires
* ssz support for `strata-identifiers`. Transparent wrappers there have manual implementation using macros. `*Commitment` types are using a schema.
* ssz support for `strata-asm-common`.
* Some fixes to `ssz-gen` related to Unions. See [this PR](https://github.com/alpenlabs/ssz-gen/pull/39).
* [Another PR](https://github.com/alpenlabs/ssz-gen/pull/40) to fix `ssz-gen` related to Unions and Options. 
* [Another PR](https://github.com/alpenlabs/ssz-gen/pull/43) to fix `serde` and `StableContainer` issues.

This PR was created with help from Claude Code.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [x] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [x] New and existing tests pass with my changes.
- [x] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

## Related Issues

[STR-1900](https://alpenlabs.atlassian.net/browse/STR-1900)

[STR-1900]: https://alpenlabs.atlassian.net/browse/STR-1900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1460, opened 2026-03-05
STR 2370, 2371 migrate strata-dbtool to OL databases and port functional tests

krsnapaudel:
## Description

* Migrate strata-dbtool commands (`get-ol-state`, `get-ol-block`, `get-syncinfo`,`revert-ol-state`, checkpoint/epoch summary commands)
* Remove `get-client-state-update` command
* Port all dbtool functional tests to functional-tests-new using the updated binary and OL-only commands
* Fullnode tests are written but not yet runnable — full node blocks sync is not implemented

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This PR depends on some commits from #1383 . Without some fixes and test env from there, we cannot run `strata-dbtool` tests.

NOTE: Needs a rebase after #1383 goes in.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-2270](https://alpenlabs.atlassian.net/browse/STR-2270)
[STR-2271](https://alpenlabs.atlassian.net/browse/STR-2271)

[STR-2270]: https://alpenlabs.atlassian.net/browse/STR-2270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bin/strata-dbtool/src/output/checkpoint.rs
```diff
@@ -1,28 +1,31 @@
...
+use strata_checkpoint_types::EpochSummary;
+use strata_identifiers::Epoch;
 
 use super::{helpers::porcelain_field, traits::Formattable};
 
 /// Epoch information displayed to the user
 #[derive(serde::Serialize)]
 pub(crate) struct EpochInfo<'a> {
-    pub epoch_index: u64,
+    pub epoch: u64,
```
barakshani: I wonder about this change. I believe the convention has been that epoch is the abstract notion of time of a full batch, and we indeed use "index" or "epoch number" to specify its index in the sequence of epochs.

I do see that this was already [introduced to the code](https://github.com/alpenlabs/alpen/pull/1162/changes#diff-cae9fe3061981be091cae831bb26f63c939b57302fc960569ebb707d78f0f56aR13) a while ago.

bin/strata-dbtool/src/cmd/checkpoint.rs
```diff
@@ -232,28 +202,29 @@ pub(crate) fn get_checkpoints_summary(
...
+        let summary_info = CheckpointsSummaryInfo {
+            expected_checkpoints_count: 0,
+            checkpoints_found_in_db: 0,
+            checkpoints_in_l1_blocks: count_checkpoints_in_asm_logs(db, args.height_from)?,
+            unexpected_checkpoints: Vec::new(),
+        };
+        return output(&summary_info, args.output_format);
+    };
+
+    // Checkpoint for epoch 0 is not expected: checkpoint building starts from epoch 1.
```
barakshani: what do you mean? here and in `get_checkpoint_epoch_range()` above we do start from `0`.

bin/strata-dbtool/src/cmd/ol.rs
```diff
@@ -0,0 +1,195 @@
...
+        header_state_root: format!("{:?}", header.state_root()),
+        manifests: manifest_data,
+    };
+
+    // Use the output utility
+    output(&block_info, args.output_format)
+}
+
+/// Get OL block summary - check all OL blocks exist in database.
+pub(crate) fn get_ol_summary(
```
barakshani, line 96: what do we need this function for? it may be very expensive to run

bin/strata-dbtool/src/cmd/ol_state.rs
```diff
@@ -0,0 +1,317 @@
...
+pub(crate) struct RevertOLStateArgs {
+    /// target OL block id
+    #[argh(positional)]
+    pub(crate) block_id: String,
+
+    /// delete blocks after target block
+    #[argh(switch, short = 'd')]
+    pub(crate) delete_blocks: bool,
+
+    /// allow reverting blocks inside checkpointed epoch
```
barakshani, line 41:
```suggestion
    /// allow reverting blocks inside checkpointed epoch
    /// this should only be executed on checkpoints that haven't been posted on chain

File: bin/strata-dbtool/src/cmd/ol_state.rs [hunk 5]
```diff
@@ -0,0 +1,317 @@
...
+    #[argh(switch, short = 'c')]
+    pub(crate) revert_checkpointed_blocks: bool,
```
barakshani, line 43:
```suggestion
    pub(crate) revert_nonposted_checkpointed_blocks: bool,
```
for clarity and being explicit.

bin/strata-dbtool/src/cmd/olstate.rs
```diff
@@ -0,0 +1,317 @@
...
+        .get_latest_toplevel_ol_state()
+        .internal_error("Failed to get latest OL state")?
+        .ok_or_else(|| {
+            DisplayedError::InternalError(
+                "OL state not found in database".to_string(),
+                Box::new(()),
+            )
+        })?;
+
+    let finalized_slot = tip_state.asm_recorded_epoch().last_slot();
```
barakshani: Is `tip_state.asm_recorded_epoch()` really final (sufficiently buried) epoch?

bin/strata-dbtool/src/cmd/ol_state.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    let finalized_slot = tip_state.asm_recorded_epoch().last_slot();
+    if target_slot < finalized_slot {
+        return Err(DisplayedError::UserError(
+            "Target block is inside finalized epoch".to_string(),
+            Box::new(target_block_id),
+        ));
+    }
+
+    let checkpoint_last_slot = get_latest_checkpoint_last_slot(db)?;
```
barakshani, line 153:
```suggestion
    let nonposted_checkpoint_last_slot = get_latest_checkpoint_last_slot(db)?;
```
Ideally we are able to check that the checkpoint isn't posted (because of the previous check it means that it's either in mempool or "confirmed" (but not "finalized")). In practice, because it is being run manually, this MUST be checked be whoever run the tool, but adding safeguards here is desired.

bin/strata-dbtool/src/cmd/ol_state.rs
```diff
@@ -0,0 +1,317 @@
...
+    if !args.revert_checkpointed_blocks && target_slot < checkpoint_last_slot {
```
barakshani, line 154: I think inequality should be <=

bin/strata-dbtool/src/cmd/olstate.rs
```diff
@@ -0,0 +1,317 @@
...
+    println!();
+
+    if target_slot >= chain_tip_slot {
+        println!("No changes would be made.");
+        println!(
+            "Target slot ({}) is at or after the chain tip slot ({}).",
+            target_slot, chain_tip_slot
+        );
+        return Ok(());
+    }
```
barakshani: might make more sense to have it on top and prevent the extra computation, making it an early exit.

crates/csm-worker/src/processor.rs
barakshani, line 1: why this pr needs to touch csm-worker code?

functional-tests-new/tests/dbtool/revert_checkpointed_block_fn.py
barakshani, line 1: Let's add some comment on top stating why this test is currently disabled

barakshani, line 1: I recommend changing the name of this test to `revert_nonposted_checkpointed_block_fn` to make it clear to any future reader -- onchain checkpoints cannot be reverted.

Even if in practice you are not changing the code above to guarantee that, I think it's better to make the name explicit, and then add a comment that in the test it may be that we revert posted/confirmed-but-non-final checkpoints -- of course, this isn't desired.

crates/csm-worker/src/processor.rs
barakshani: Oh, I see that use you commit from a different PR. In that case, I think it's best to make this PR into that branch. When the other PR gets merged, this PR while automatically target main branch.

functional-tests-new/tests/dbtool/revert_checkpointed_block_fn.py
```diff
@@ -0,0 +1,107 @@
...
+        btc_service = self.get_service(ServiceType.Bitcoin)
+        fn_service = self.get_service(STRATA_FULLNODE_SERVICE_NAME)
+        setup = setup_revert_ol_state_test_fullnode(seq_service, fn_service, btc_service)
+        old_seq_tip, old_fn_tip = wait_for_seq_fn_progress(
+            seq_service,
+            fn_service,
+            setup["seq_rpc"],
+            setup["fn_rpc"],
+        )
+        time.sleep(2)
```
barakshani: why is this needed?

functional-tests-new/tests/dbtool/revert_checkpointed_block_fn.py
```diff
@@ -0,0 +1,107 @@
...
+
+        seq_service.stop()
+        fn_service.stop()
+        datadir = setup["fn_datadir"]
+        latest_checkpoint = get_latest_checkpoint(datadir)
```
barakshani, line 58: as stated, we need to make sure that the checkpoint isn't posted, otherwise revert should fail.

functional-tests-new/tests/dbtool/revert_checkpointed_block_should_fail.py
barakshani, line 1: revert_checkpointed_block_should_fail --> revert_checkpointed_block_disabled_should_fail

or anything better you can come up with.

functional-tests-new/tests/dbtool/revert_ol_state_fn.py
barakshani, line 1: is this dry run as well?

barakshani:
I'm not sure, but I think that you should include `test_epoch_progression.py` from your other PR to make CI pass, or wait until that PR gets merged.

functional-tests-new/tests/dbtool/revert_ol_state_fn.py
krsnapaudel: @barakshani No.

functional-tests-new/tests/dbtool/revert_checkpointed_block_should_fail.py
krsnapaudel: @barakshani Why disabled? Also this is a direct migration. Any changes to test should be in another ticker.

functional-tests-new/tests/dbtool/revert_checkpointed_block_fn.py
krsnapaudel: @barakshani Not really the test does not guarantee non-posted. It just finds the latest checkpoint/epoch in db. Reverts to the start of that epoch. There is no check for posted to L1 or not.

bin/strata-dbtool/src/cmd/olstate.rs
krsnapaudel: @barakshani This is actually incorrect. Fetching it from `client_state`. I was under the impression that `client_state` db can be removed. Apparently not.

bin/strata-dbtool/src/output/checkpoint.rs
krsnapaudel: @barakshani The idea is to avoid confusion and type conversions. Since we have an `Epoch` type, we are using it consistently.

bin/strata-dbtool/src/cmd/checkpoint.rs
krsnapaudel: You are right. `get_checkpoint_epoch_range` needs an update.

bin/strata-dbtool/src/cmd/ol.rs
krsnapaudel: Yes, we could add a `from_slot` parameter to limit.

bin/strata-dbtool/src/cmd/ol_state.rs
krsnapaudel: @barakshani Note this has to be a separate ticket. Mainly because I need to investigate what API we have to determine "not posted" to l1. Created [this ticket](https://alpenlabs.atlassian.net/browse/STR-2541) for all of your comments related to "nonposted" checkpoint.

bin/strata-dbtool/src/cmd/ol_state.rs
krsnapaudel: @barakshani See above for the link to a new ticket.

bin/strata-dbtool/src/cmd/ol_state.rs
krsnapaudel: Last slot is fine. It's the target, not reverted.

functional-tests-new/tests/dbtool/revert_checkpointed_block_fn.py
krsnapaudel: Good catch. Will remove or add appropriate wait.

bin/strata-dbtool/src/cmd/ol_state.rs
barakshani: My added line is a "Note for user" than a code comment. The user should understand they should not try this on a posted checkpoint. Feel free to revise the language.

Thanks for the ticket, I meant to do so, but haven't gotten to it.

bin/strata-dbtool/src/output/checkpoint.rs
barakshani: Honestly it confused me, otherwise I wouldn't have commented. So there are place where we now use `Epoch` type and epoch name, but here we use epoch name for `u64`. A point for thought for you, you can decide if it's worth the effort.

functional-tests-new/tests/dbtool/revert_checkpointed_block_should_fail.py
barakshani: "Disabled" meant to say the flag (`-c` I think) it disabled. Sorry if it wasn't clear, and as I wrote, perhaps you have better suggestion as I understand it's not immediate.

I will comment on the ticket, but this it very confusing. We have tests called "revert_checkpointed_block_fn" and "revert_checkpointed_block_should_fail". I fell for that several times, and I'm sure other team members in the future as well.

krsnapaudel: -c is a flag to override and revert checkpointed blocks. Should fail because -c was not used in this test. Explicit override is needed.

bin/strata-dbtool/src/output/checkpoint.rs
krsnapaudel: `Epoch` type is now the type for epoch number. Not sure what confused you.

bin/strata-dbtool/src/cmd/ol_state.rs
```diff
@@ -0,0 +1,308 @@
...
+        l1_next_expected_height: l1_safe_block_height.saturating_add(1),
+        l1_safe_block_height,
+        l1_safe_block_blkid: top_level_state.last_l1_blkid(),
+    };
+
+    output(&ol_state_info, args.output_format)
+}
+
+/// Revert OL state to specified block.
+pub(crate) fn revert_ol_state(
```
bewakes: This function is too big and hard to read., please split this into smaller functions.

bin/strata-dbtool/src/output/checkpoint.rs
barakshani: sorry, was mistaken with something else. Was trying to say that we have been using "epoch" in a different way, like "epoch summaries", so using "epoch" for epoch number is confusing. As noted, this notion has already noted out codebase, so I understand the attempt to make it consistent here. At least something to be aware of.

bewakes:
> @bewakes, @storopoli I think we need the last summarized epoch exposed through RPC. Is this not considered useful in general?

@krsnapaudel it seems to be useful and we can add it to the `get_chainStatus` method in which you already seem to have added `parent` attribute. Since epoch summary is created and stored in database after a terminal block is processed in chain worker. The rpc method should just fetch it from the db and add it to the ChainStatus response.

bin/strata-dbtool/src/cmd/syncinfo.rs
```diff
@@ -27,59 +28,61 @@ pub(crate) fn get_syncinfo(
...
+    let (latest_commitment, top_level_state) = db
+        .ol_state_db()
+        .get_latest_toplevel_ol_state()
+        .internal_error("Failed to get latest OL state")?
+        .ok_or_else(|| {
+            DisplayedError::InternalError(
+                "OL state not found in database".to_string(),
+                Box::new(()),
+            )
+        })?;
```
storopoli, line 69: You're picking the "canonical" block by using the `get_tip_slot()` and saving into `ol_tip_height`. Then you read the top-level state from `get_latest_toplevel_ol_state()` instead of the same "chosen" block (`ol_tip_height`).

That means the reported tip block can come from one branch while `current_slot`, `previous_block`, `safe_block`, etc. come from another.

bin/strata-dbtool/src/cmd/l1.rs
```diff
@@ -111,43 +110,42 @@ pub(crate) fn get_l1_summary(
     let (l1_tip_height, l1_tip_block_id) = get_l1_chain_tip(db)?;
 
     let start_height = args.height_from;
+    if start_height > l1_tip_height {
+        return Err(DisplayedError::UserError(
+            format!("Provided height is above canonical L1 tip {l1_tip_height}"),
+            Box::new(start_height),
+        ));
+    }
```
storopoli, line 121: `get-l1-summary` no longer verifies that ASM manifests exist, even though the command contract and docstring still say it does. It now only checks whether a canonical L1 block id exists at each height, so missing manifest rows can still produce `all_manifests_present = true`.

PR #1466, opened 2026-03-06
fix(func-test): prevent revert target from falling inside finalized epoch

purusang:
This functional test (`revert_checkpointed_block_fn`) has been flaky because epochs can be finalized for the targeted reversal block when bitcoin is running and sequencer taking some time to shut down. As a result chainstate could not be reverted since the targeted block may fall in a finalized epoch.
I have also moved the wallet creation logic to `bitcoind -wallet` since its needed to every env.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/functional-tests/tests/dbtool/revert_checkpointed_block_fn.py
+++ b/functional-tests/tests/dbtool/revert_checkpointed_block_fn.py
@@ -1,9 +1,8 @@
-import time
-
 import flexitest
 
 from envs import net_settings, testenv
 from mixins.dbtool_mixin import FullnodeDbtoolMixin
+from utils.constants import BLOCK_GENERATION_INTERVAL_SECS
 from utils.dbtool import (
     get_latest_checkpoint,
     restart_fullnode_after_revert,
@@ -12,7 +11,7 @@ from utils.dbtool import (
     verify_checkpoint_deleted,
     verify_revert_success,
 )
-from utils.utils import wait_until
+from utils.utils import generate_blocks, wait_until
 
 
 @flexitest.register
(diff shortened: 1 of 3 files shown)


storopoli (approved):
ACK 92876e46a0f10e68455028ab4e05cd0c0a7efe6f
Can this also be applied to `functional-tests-new`?
Also @voidash can you review this please?

functional-tests/tests/dbtool/revert_checkpointed_block_fn.py
```diff
@@ -31,16 +30,21 @@ def main(self, ctx: flexitest.RunContext):
         # Setup: generate blocks, finalize epoch 1, and wait for checkpoint 2
         setup_revert_chainstate_test(self, web3_attr="web3")
 
+        # Stop bitcoin immediately to prevent further checkpoint finalization,
+        # which would finalize more epochs and cause the revert target to
+        # fall inside a finalized epoch.
+        self.btc.stop()
```
barakshani, line 36: is there a way to stop btcio?

stopping bitcoin won't help -- we have a race: we freeze everything (stopping prover, sequencer etc.) and get the _last_ checkpoint. This last checkpoint can be either on bitcoin already (non/finalized) or not sent yet. Despite the test name, we need it to be not posted yet, otherwise we cannot revert (we cannot revert txes on bitcoin, even if they are "not final").

So the race is on the state the last checkpoint is in. We want it to be not posted it. If we can stop btcio, that would do.

purusang: Yes I just realised that after current failure in ci. And won't be spending time anymore.

purusang: Also our btc blocktime is 0.5s and when we wait for 3 l2 blocks it will make prev checkpoint finalized....
Like you said we should be able to revert epoch not posted to btc. Hope this will be informative to future migration that @krsnapaudel is doing.

PR #1468, opened 2026-03-06
feat!(asm-checkpoint): strip checkpoint payload from excessive signature and adjust verification according to SPS-51

evgenyzdanovich:
## Description

Changes:
- Encode checkpoint payload using Codec (over SSZ) to reduce actual payload bytes onchain
- Strip checkpoint payload from signature to reduce excessive bytes onchain
- Change the validation of checkpoint from validating the checkpoint signature (in the payload) to **validating the sequencer key as set in the chechkpoint state against pubkey in the envelope.** This is a compliant change with the existing [SPS-51 envelope container spec](https://www.notion.so/SPS-51-Generic-simple-envelope-format-1d1901ba000f807f9b36e491d765e786#211901ba000f8066b8d0c824feb355c9).

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/asm/subprotocols/checkpoint/src/errors.rs
+++ b/crates/asm/subprotocols/checkpoint/src/errors.rs
@@ -1,6 +1,6 @@
 use strata_asm_common::AuxError;
 use strata_identifiers::Epoch;
-use strata_predicate::PredicateError;
+use strata_predicate::{PredicateError, PredicateTypeId};
 use thiserror::Error;
 
 /// Result type for checkpoint subprotocol operations.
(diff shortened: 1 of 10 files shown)


crates/asm/subprotocols/checkpoint/src/errors.rs
```diff
@@ -19,9 +19,15 @@ pub enum CheckpointValidationError {
...
 #[derive(Debug, Error)]
 pub enum InvalidCheckpointPayload {
-    /// Predicate verification failed.
-    #[error("sequencer predicate verification failed: {0}")]
-    SequencerPredicateVerification(PredicateError),
+    /// Envelope pubkey does not match the sequencer predicate pubkey.
+    #[error(
+        "envelope pubkey does not match sequencer predicate: expected {expected:?}, got {actual:?}"
+    )]
+    SequencerPubkeyMismatch { expected: Vec<u8>, actual: Vec<u8> },
```
prajwolrg: nit: I think it might make sense to display this as hex in the error, but that just adds an additional dep.

bin/strata/src/main.rs
```diff
@@ -61,16 +61,35 @@ fn main() -> Result<()> {
     // Check for db consistency, external rpc clients reachable, etc.
     do_startup_checks(&nodectx)?;
 
+    // Load sequencer key early so it can be shared with both the envelope writer
+    // (for SPS-51 taproot authentication) and the duty executor (for block signing).
```
prajwolrg, line 65: I wonder if it's good/bad practice to include SPS-* in the docstrings, given that the actual documents are not public. CC: @storopoli @barakshani @delbonis.

delbonis: I think it's fine since we fully intend to make them all public post-mainnet.

evgenyzdanovich: I also additionally mentioned SPS-51 in our AGENTS.md (for whole who use it), so the agent is more aware of the context while it scans the code (and eventually encounters comments mentioning sps-51).

crates/asm/subprotocols/checkpoint/src/errors.rs
evgenyzdanovich: That's a nice catch and quite important for the "usability" of the error - so, thank you! 
The last thing I wanna see while receiving such an error is **a key represented as a sequence of bytes in logs** :D

I also think, for things like this (esp if it's used one time), it's fine to have private inline helpers. Here we can rely upon the "native hexing/formatting" through the `std::fmt` to not introduce the dep. 

Please see how it looks - what's your take?

delbonis: Debug printing a `[42, 13, 37, ...]` is pretty ugly, especially when we usually represent pubkeys as hex.

prajwolrg: Yes makes sense.

evgenyzdanovich: @delbonis ye, we agree, that was fixed in  here https://github.com/alpenlabs/alpen/pull/1468/changes/f55063431f082cb7133f7f7d1d94648867ab4f6f

crates/btcio/src/writer/builder.rs
```diff
@@ -51,6 +51,14 @@ pub struct EnvelopeConfig {
...
     /// Bitcoin fee rate, sats/vByte
     pub fee_rate: u64,
+    /// Optional keypair for the taproot envelope script.
+    ///
+    /// When set, this keypair is used as the `<pubkey>` in `<pubkey> CHECKSIG` of the
+    /// envelope script. Per SPS-51, consumers can recognize this pubkey and treat the
+    /// taproot script-spend signature as transitively signing the envelope contents.
+    ///
+    /// When `None`, a random keypair is generated per transaction (legacy behavior).
+    pub envelope_keypair: Option<UntweakedKeypair>,
```
prajwolrg, line 61: This is neat. Nice!

tests/harness/test_harness.rs
```diff
@@ -468,38 +525,19 @@ impl AsmTestHarness {
...
 
-/// Build envelope script for embedding data in Bitcoin transaction.
+/// Build a simple envelope script for subprotocols that don't need SPS-51 auth.
 ///
-/// Creates an OP_FALSE OP_IF ... OP_ENDIF envelope containing the payload,
-/// followed by OP_TRUE to satisfy tapscript's requirement that exactly one
-/// value remains on the stack after execution.
-fn build_envelope_script(payload: &[u8]) -> ScriptBuf {
+/// Creates an `OP_FALSE OP_IF <data> OP_ENDIF OP_TRUE` tapscript.
+fn build_simple_envelope_script(payload: &[u8]) -> ScriptBuf {
```
prajwolrg, line 528: I think this can be replaced with [`build_envelope_script`](https://github.com/alpenlabs/strata-common/blob/main/crates/l1proto/envelope-fmt/src/builder.rs#L174) in `strata-l1-txfmt`.

evgenyzdanovich: yup, thx

PR #1469, opened 2026-03-06
STR 2210 Move the OL sequencer worker to use the "service worker"

krsnapaudel:
## Description

Replaces the manual worker-spawn pattern (`duty_fetcher_worker` + `duty_executor_worker`) with the strata-service AsyncService framework, matching how other services (blockasm, mempool, FCM) are structured.

### Changes
Add `SequencerService` implementing `Service + AsyncService` traits
Add `SequencerBuilder` with `launch()` returning `ServiceMonitor`
Add `SequencerTimerInput` implementing `AsyncServiceInput` for tick-based polling
Rewrite `start_sequencer_signer` to use builder instead of raw spawns
Store returned `ServiceMonitor` in `main.rs` (follows codebase convention)
Delete `duty_fetcher.rs` and `duty_executor.rs`

No behavior changes: duty dedup, per-duty spawn, failed-duty retry, block sleep semantics, and key zeroization are all preserved.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-2210](https://alpenlabs.atlassian.net/browse/STR-2210)

[STR-2210]: https://alpenlabs.atlassian.net/browse/STR-2210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bin/strata/src/sequencer/service.rs
```diff
@@ -0,0 +1,341 @@
...
+    pub(crate) fcm_handle: Arc<FcmServiceHandle>,
+    pub(crate) status_channel: Arc<StatusChannel>,
+    pub(crate) sequencer_key: Buf32,
+    pub(crate) seen_duties: HashSet<Buf32>,
+    pub(crate) active_duties: Arc<AtomicU32>,
+    pub(crate) failed_duty_count: Arc<AtomicU32>,
+    pub(crate) failed_duties_tx: mpsc::Sender<Buf32>,
+    pub(crate) failed_duties_rx: mpsc::Receiver<Buf32>,
+    pub(crate) duties_dispatched: u64,
+}
```
delbonis: Surely not all of these need to be `pub(crate)`?

While we're taking the time to do this refactor, we should probably try to hide the external dependencies behind a context trait.

bin/strata/src/sequencer/service.rs
```diff
@@ -0,0 +1,341 @@
...
+        match input {
+            SequencerEvent::Tick => process_tick(state).await,
+        }
+
+        Ok(Response::Continue)
+    }
+}
+
+#[tracing::instrument(skip_all, fields(component = "sequencer_service"))]
+async fn process_tick(state: &mut SequencerServiceState) {
```
delbonis: This instrumentation is kinda redundant since `process_input` gets instrumented by the service framework.

bin/strata/src/sequencer/service.rs
```diff
@@ -0,0 +1,341 @@
...
+struct DutyContext {
+    blockasm_handle: Arc<BlockasmHandle>,
+    envelope_handle: Arc<EnvelopeHandle>,
+    storage: Arc<NodeStorage>,
+    fcm_handle: Arc<FcmServiceHandle>,
+    sequencer_key: Buf32,
+    active_duties: Arc<AtomicU32>,
+    failed_duty_count: Arc<AtomicU32>,
+    failed_duties_tx: mpsc::Sender<Buf32>,
+}
```
delbonis: Even without a context trait, couldn't we make the service state type contain an instance of this type pre-constructed so we can just clone *that*?

The `duty_context` fn not having a verby name as a member fn implies it's an accessor, but it's actually constructing a new value, so it should be someting like `create_duty_context`.

Also, item ordering here is weird, why is the `impl ServiceState for SequencerServiceState` mixed in with `DutyContext` items?

bin/strata/src/sequencer/service.rs
```diff
@@ -0,0 +1,341 @@
...
+    async fn before_shutdown(
+        state: &mut Self::State,
+        err: Option<&anyhow::Error>,
+    ) -> anyhow::Result<()> {
+        let active = state.active_duties.load(Ordering::Relaxed);
+        if let Some(err) = err {
+            warn!(%active, %err, "sequencer service shutting down with error");
+        } else {
+            info!(%active, "sequencer service shutting down");
+        }
```
delbonis: I think this is kinda redundant, the service worker itself lifecycle logs for this doesn't it?

bin/strata/src/sequencer/service.rs
```diff
@@ -0,0 +1,341 @@
...
+        ctx.active_duties.fetch_add(1, Ordering::Relaxed);
+
+        tokio::spawn(async move {
+            if let Err(err) = handle_duty(&ctx, duty).await {
+                error!(?duty_id, %err, "duty failed");
+                ctx.failed_duty_count.fetch_add(1, Ordering::Relaxed);
+                let _ = ctx.failed_duties_tx.send(duty_id).await;
+            }
+
+            ctx.active_duties.fetch_sub(1, Ordering::Relaxed);
```
delbonis: This approach using an atomic is problematic.  If a task panics it'll never do the decrement, even when unwinding, which could cause weird behavior elsewhere.  You should use a type that will do the decrement when `Drop` is called, which *should* still work even when unwinding.

However, is this just for status monitoring?  Maybe it makes sense to derive the value being tracked here from [`sender_strong_count`](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.sender_strong_count).

bin/strata/src/sequencer/service.rs
```diff
@@ -0,0 +1,341 @@
...
+        .put_block_data_async(block.clone())
+        .await
+        .map_err(|e| anyhow!("failed storing block: {e}"))?;
+
+    let blkid = block.header().compute_blkid();
+    if !fcm_handle
+        .submit_chain_tip_msg_async(ForkChoiceMessage::NewBlock(blkid))
+        .await
+    {
+        return Err(anyhow!("failed submitting block to fork choice manager"));
```
delbonis: Should we convert these errors to an enum?

And the other cases.

bin/strata/src/sequencer/service.rs
krsnapaudel: The context trait is now in place — `NodeSequencerContext` keeps all handles private and only exposes behavior.

bin/strata/src/sequencer/service.rs
krsnapaudel: Removed.

bin/strata/src/sequencer/service.rs
krsnapaudel: `DutyContext` is now pre-constructed, so the accessor name is accurate. Item ordering is fixed: `DutyContext`, `ActiveDutyGuard`, then `SequencerServiceState` with its own impl blocks including impl `ServiceState` grouped together, then `SequencerService`.

bin/strata/src/sequencer/service.rs
krsnapaudel: Now `ActiveDutyGuard::new(...)` increments. `Drop` decrements.

bin/strata/src/sequencer/service.rs
krsnapaudel: Yes added error enums.

bin/strata/src/sequencer/builder.rs
```diff
@@ -0,0 +1,88 @@
...
+        let failed_duty_count = Arc::new(AtomicU32::new(0));
+        let (failed_duties_tx, failed_duties_rx) = mpsc::channel(8);
+
+        let context = Arc::new(NodeSequencerContext::new(
+            self.blockasm_handle,
+            self.envelope_handle,
+            self.storage,
+            self.fcm_handle,
+            self.status_channel,
+        ));
```
delbonis: The context should be constructed separately and passed in.  This builder isn't really providing much to help construct the service, it's still very manual, just another step removed.

PR #1474, opened 2026-03-09
1877 persisted da in blockasm

bewakes:
## Description

This is a draft implementation of one of the DA data generation designs proposed in https://www.notion.so/DA-data-generation-Design-319901ba000f80b5b2e4fdad48416bd1.

The following things are done in this PR:
1. Update BlockAssembly context trait to include `fetch_accumulated_da` and `store_accumulated_da` methods.
2. Wrap `WriteTrackingState` in block assembly by `DaAccumulatingState`.
3. Derive `Codec` for `EpochDaAccumulator` which has been cascaded into subsequent internal types. This Codec derivation is for storing the accumulator in the database and use it for subsequent block processing.

### Todo
To make the intended functionality complete, the following need to be done:
- [ ] Storage definition for accumulator.
- [ ] Codec/serialization tests.
- [ ] Update block assembly tests.
- [ ] Implement the new methods in the block assembly context.
- [ ] Spawn a thin cleanup service to prune accumulated DA data after an epoch is finalized.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist


crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -186,16 +220,79 @@ impl AccountDelta {
...
+        self.account_id.encode(enc)?;
+        Ok(())
+    }
+
+    fn decode(dec: &mut impl Decoder) -> Result<Self, CodecError> {
+        let serial = AccountSerial::decode(dec)?;
+        let account_id = AccountId::decode(dec)?;
+        Ok(Self { serial, account_id })
+    }
+}
```
delbonis, line 291: "serial" is redundant.

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -450,6 +547,107 @@ impl EpochDaAccumulator {
...
+impl Clone for EpochDaAccumulator {
+    fn clone(&self) -> Self {
+        // For DaCounterBuilder, recreate it preserving both original and new values
+        let slot_builder = self.slot_builder.as_ref().map(|builder| {
+            let mut new_builder =
+                DaCounterBuilder::<CtrU64ByU16>::from_source(*builder.original_value());
+            // Set the new value to match the current state
+            let _ = new_builder.set(*builder.new_value());
+            new_builder
+        });
```
delbonis, line 559: Surely we can clone it directly?

crates/ol/block-assembly/src/da_accumulation.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+impl Codec for AccumulatedDaData {
+    fn encode(&self, enc: &mut impl Encoder) -> Result<(), CodecError> {
+        self.epoch.encode(enc)?;
+        self.accumulator.encode(enc)?;
+        // For now, we'll skip encoding logs since OLLog might not have Codec
+        // In a real implementation, you'd need to ensure OLLog implements Codec
+        // or serialize them differently
```
delbonis, line 84: Wrap it in the `CodecSsz` shim in the type so this can just use the codec derive macro.  It's `#[repr(transparent)]` so it can be safely `mem::transmute`d.

crates/ol/block-assembly/src/da_accumulation.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// The epoch this accumulation belongs to.
+    epoch: Epoch,
+
+    /// The epoch DA accumulator that tracks state changes.
+    /// This is serialized and persisted between blocks.
+    accumulator: EpochDaAccumulator,
+
+    /// All logs emitted in the epoch up to and including this block.
+    logs: Vec<OLLog>,
+}
```
delbonis, line 26: This naming is misleading since it includes logs, which are not part of the DA diff data that `EpochDaAccumulator` contains.

crates/ol/state-support-types/src/lib.rs
```diff
@@ -8,7 +8,7 @@ pub use strata_ol_state_types::{LedgerWriteBatch, SerialMap, WriteBatch};
 pub use write_tracking_layer::WriteTrackingState;
 
 mod batch_diff_layer;
-mod da_accumulating_layer;
+pub mod da_accumulating_layer;
```
delbonis, line 11: This doesn't need to be pub.

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -485,6 +683,35 @@ impl<S: IStateAccessor> DaAccumulatingState<S> {
...
+    /// This can be used to serialize the accumulator state between blocks.
+    pub fn accumulator(&self) -> &EpochDaAccumulator {
+        &self.epoch_acc
+    }
+
+    /// Takes ownership of the accumulator, replacing it with a default one.
+    /// This is useful for extracting the accumulator for serialization.
+    pub fn take_accumulator(&mut self) -> EpochDaAccumulator {
+        mem::take(&mut self.epoch_acc)
+    }
```
delbonis, line 708: This is a C++-ish pattern.  This should be `into_accumulator` or have the type defined around a `&mut EpochDaAccumulator` in the first place so that the consumer can manage it without keeping the layer around across STF invocations or needing to juggle around the `EpochDaAccumulator` around.

crates/ol/state-support-types/src/da_accumulating_layer.rs
bewakes: No, `DaCounterBuilder` cannot be cloned as it doesn't implement `Clone` ( requires `Clone` for `type Base`).

delbonis: I don't think I had a reason specifically for excluding that, I would think it's safe to add it.

crates/ol/block-assembly/src/block_assembly.rs
```diff
@@ -253,21 +260,35 @@ where
     // Create output buffer to collect logs from all transaction executions.
     let output_buffer = ExecOutputBuffer::new_empty();
 
+    // Fetch accumulated DA data from parent block if non terminal else start fresh.
+    let accumulated_da = if parent_block.header().is_terminal() {
+        AccumulatedDaData::new(block_epoch, EpochDaAccumulator::default(), vec![])
+    } else {
+        ctx.fetch_accumulated_da(parent_commitment)
+            .await?
+            .ok_or(BlockAssemblyError::Other(
```
storopoli, line 269: This turns the TODO stub in `BlockAssemblyContext` into a hard runtime failure for every non-terminal child block. `fetch_accumulated_da()` still returns `Ok(None)` and `store_accumulated_da()` is still a no-op in the default context 

https://github.com/alpenlabs/alpen/blob/c552ce91086d1c2af67a5a29ebcd3414ca2e7257/crates/ol/block-assembly/src/context.rs#L297-L312

so the second block in an epoch now errors with `Other(Cannot find expected da data)` instead of continuing to build. This path should stay non-fatal or be guarded behind a real persistence implementation.

crates/ol/block-assembly/src/da_accumulation.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub fn into_parts(self) -> (EpochDaAccumulator, Vec<OLLog>) {
+        (self.accumulator, self.logs)
+    }
+}
+
+impl Codec for AccumulatedDaData {
+    fn encode(&self, enc: &mut impl Encoder) -> Result<(), CodecError> {
+        self.epoch.encode(enc)?;
+        self.accumulator.encode(enc)?;
+        // For now, we'll skip encoding logs since OLLog might not have Codec
```
storopoli, line 82: This persisted format is lossy: `encode()` drops `logs`, and `decode()` reconstructs them as empty. But block assembly explicitly carries the prior epoch logs forward via `into_parts()` and appends the current block logs before re-storing

https://github.com/alpenlabs/alpen/blob/c552ce91086d1c2af67a5a29ebcd3414ca2e7257/crates/ol/block-assembly/src/block_assembly.rs#L277-L317

Once `fetch/store_accumulated_da()` is backed by real storage, reloading any accumulated value will silently discard the earlier withdrawal and log history for the epoch.

PR #1476, opened 2026-03-09
refactor(docker): use datatool for params generation, add fullnode mode

voidash:
## Summary
- Replaces hardcoded JSON heredocs in `init-alpen-client-keys.sh` with `datatool` subcommands (`genparams`, `gen-ol-params`, `gen-asm-params`, `genxpriv`, `genseqpubkey`)
- Adds `--fullnode` mode that copies params from a sequencer's output dir and extracts the sequencer pubkey from `rollup-params.json`
- Adds signet support via `BITCOIN_NETWORK` env var

## Test plan
- [ ] Run `./init-alpen-client-keys.sh --sequencer <datatool>` on regtest and verify all params/keys are generated
- [ ] Run `./init-alpen-client-keys.sh --fullnode <datatool> --params-dir <seq-configs>` and verify params are copied and `.env.alpen-client-fullnode` is correct
- [ ] Run `BITCOIN_NETWORK=signet ./init-alpen-client-keys.sh --sequencer <datatool>` and verify signet genesis values
- [ ] Verify `docker compose --env-file .env.alpen-client -f docker-compose-alpen-client.yml up` starts correctly with generated configs

https://alpenlabs.atlassian.net/browse/STR-2538

### DIFF
--- a/docker/.env.alpen.example
+++ b/docker/.env.alpen.example
@@ -1,4 +1,9 @@
 # Run ./init-network.sh to generate .env.alpen with real values.
+#
+# For signet, generate with:
+#   BITCOIN_NETWORK=signet GENESIS_L1_HEIGHT=<height> \
+#   BITCOIN_RPC_URL=... BITCOIN_RPC_USER=... BITCOIN_RPC_PASSWORD=... \
+#   ./init-network.sh <datatool_path>
 
 SEQUENCER_PRIVATE_KEY=
 SEQUENCER_PUBKEY=
@@ -10,6 +15,8 @@ CHAIN_SPEC=testnet
 
 EE_DA_MAGIC_BYTES=ALPN
 L1_REORG_SAFE_DEPTH=1
+# Must match the genesis height used when generating params.
+# Set via GENESIS_L1_HEIGHT when running init-network.sh.
 GENESIS_L1_HEIGHT=0
 BATCH_SEALING_BLOCK_COUNT=5
 
--- a/docker/init-network.sh
+++ b/docker/init-network.sh
@@ -8,10 +8,19 @@ set -euo pipefail
 #   ./init-network.sh <datatool_path>
 #   ./init-network.sh --sequencer <datatool_path>
 #   ./init-network.sh --fullnode <datatool_path> --params-dir <path>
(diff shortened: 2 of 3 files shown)


docker/init-alpen-client-keys.sh
```diff
@@ -76,171 +145,100 @@ generate_key_file() {
...
-# --- asm-params.json ---
-# NOTE: genesis_ol_blkid is a placeholder. With "AlwaysAccept" checkpoint
-# predicate this value is not validated. For production deployments, compute
-# the correct value using: strata-datatool gen-asm-params --ol-params <ol-params>
+    ROLLUP_PARAMS="${OUTPUT_DIR}/rollup-params.json"
+    if [ ! -f "${ROLLUP_PARAMS}" ]; then
+        "${DATATOOL_PATH}" -b "${BITCOIN_NETWORK}" \
+            genparams \
+            -o "${ROLLUP_PARAMS}" \
+            -n alpn \
```
prajwolrg: Should use "ALPN" instead of "alpn". "ALPN" is used most of the places. 

This shouldn't matter as long as everything is consistent. We started using `ALPN` in bridge as well in https://github.com/alpenlabs/strata-bridge/pull/390.

prajwolrg: It seems we use "alpn" in some other places as well. Let's make it consistent everywhere in the codebase.

delbonis: agree with this

bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -383,7 +383,7 @@ mod tests {
             &mut wallet,
             &header_aux,
             &deposit_output,
-            "alpn".parse().expect("valid magic bytes"),
+            "ALPN".parse().expect("valid magic bytes"),
```
prajwolrg: nit: we can also use `MagicByte::new(*b"ALPN")` and that'd mean the check is in compile time.

prajwolrg (dismissed):
LGTM! Good work with comments.

storopoli (dismissed):
ACK 20a6b52edde1b7af0b973511976810fc5224971c

PR #1478, opened 2026-03-09
feat(strata): make OL block time configurable

storopoli:
## Description

Add a Strata-local OL block time setting that can be overridden with `STRATA_OL_BLOCK_TIME_MS`, with a default fallback of `5_000` ms.

This wires the configured block time through the sequencer duty extraction and RPC paths so new OL templates use `parent_timestamp + block_time_ms` instead of relying on rollup params. It also normalizes OL block template timestamps to milliseconds so duty scheduling and fallback timestamps use the same unit (a bug that I've found).

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Done with the help of Cursor using GPT-5.4 High.

- Adds `sequencer.ol_block_time_ms` to `SequencerConfig` with a default of `5000`.
- Supports env override via `STRATA_OL_BLOCK_TIME_MS`.
- Rejects `0` for the configured OL block time.
- Updates OL block template fallback timestamps to use milliseconds.

- [ ]  Yes
- [x]  No

## Checklist

AI assistance notice: This PR was prepared with AI assistance for implementation and validation.

## Related Issues
None.


bin/strata/src/rpc/mod.rs
```diff
@@ -43,15 +43,23 @@ struct SeqRpcDeps {
 
     /// Block assembly handle.
     blockasm_handle: Arc<BlockasmHandle>,
+
+    /// Target OL block time in milliseconds for local sequencing.
+    ol_block_time_ms: u64,
```
krsnapaudel: Should this be in sequencer config?

delbonis: I assume that this is extracted directly from the chain params so we wouldn't exactly want it to be configurable by the sequencer client.

prajwolrg: nit: should use `std::time::Duration` instead of `u64`.

bin/strata/src/sequencer/duty_fetcher.rs
```diff
@@ -15,6 +15,7 @@ pub(crate) async fn duty_fetcher_worker(
     blockasm_handle: Arc<BlockasmHandle>,
     storage: Arc<NodeStorage>,
     status_channel: Arc<StatusChannel>,
+    ol_block_time_ms: u64,
```
prajwolrg: same: should use `Duration` instead of `u64`.

bin/strata/src/sequencer/rpc.rs
```diff
@@ -45,12 +48,14 @@ impl OLSeqRpcServer {
         status_channel: Arc<StatusChannel>,
         blockasm_handle: Arc<BlockasmHandle>,
         envelope_handle: Arc<EnvelopeHandle>,
+        ol_block_time_ms: u64,
```
prajwolrg: This feels really weird. I think there should be a better way than passing this information this way.

prajwolrg: The `BlockasmServiceState` should have a `config` file alongside `params` file. The block time should be part of that `config` instead.

bin/strata/src/sequencer/duty_fetcher.rs
storopoli: Done in f07dabd89da7d94dd33d1f87104899642f9ad0b5

bin/strata/src/rpc/mod.rs
storopoli: Done in f07dabd89da7d94dd33d1f87104899642f9ad0b5

bin/strata/src/sequencer/rpc.rs
storopoli: Ok I did this:

- moves OL block time to a config file: 323bc7e4c5bf60e89394b238a2f38148147f1b82
- makes sure `datatool` generates this config in the new functional tests: cf450146d6d9283f1e4f6b7659f16421aa630acf
- adds the generation of the new config file in the docker stuff: 48ef34f6507e89948c1ccb0cc8b40ce78f55b80d

Please check.

bin/strata/src/rpc/mod.rs
storopoli: Right now with the changes requested this is done by a runtime configuration JSON file.

bin/strata/src/errors.rs
```diff
@@ -43,6 +46,9 @@ pub(crate) enum InitError {
     #[error("missing sync endpoint (required for non-sequencer nodes)")]
     MissingSyncEndpoint,
 
+    #[error("invalid OL block time ms: {0}")]
+    InvalidOlBlockTimeMs(String),
```
prajwolrg: why is this string?

bin/strata/src/context.rs
```diff
@@ -152,6 +163,54 @@ fn load_rollup_params(path: &Path) -> Result<RollupParams, InitError> {
...
+    let stem = params_path
+        .file_stem()
+        .and_then(|stem| stem.to_str())
+        .unwrap_or("params");
+    let file_name = match params_path.extension().and_then(|ext| ext.to_str()) {
+        Some(ext) => format!("{stem}.blockasm.{ext}"),
+        None => format!("{stem}.blockasm"),
+    };
+    params_path.with_file_name(file_name)
+}
```
prajwolrg, line 189: I think it makes sense to have some form of configuration just for the sequencer `sequencer.toml`, which won't be needed by other full nodes. Right now it can include information about OL block time, but this can be extended to have more information about checkpoint configuration. 

It makes sense to decouple params and config, but this seems like we are conflating it.

prajwolrg (changes requested):

bin/strata/src/errors.rs
storopoli: We are now using TOML and this is properly parsed as `u64`.
See a9785b3c3377f3b9daf4b58fac48a35fe8a46948

bin/strata/src/context.rs
storopoli: Yeah fair point. As we've discussed this has been changed to TOML and I've reverted all the JSON stuff into TOML (`sequencer.toml`) in a9785b3c3377f3b9daf4b58fac48a35fe8a46948

PR #1487, opened 2026-03-12
feat(sled): add rkyv serialization codec

storopoli:
## Description

Adds `rkyv` macro to `strata-db-store-sled`.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Subtask of STR-1912.

Done with the help of GPT-5.3 high.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-1914

---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #1506 
- <kbd>&nbsp;1&nbsp;</kbd> #1487 👈

### DIFF
--- a/crates/db/store-sled/src/mempool/schemas.rs
+++ b/crates/db/store-sled/src/mempool/schemas.rs
@@ -1,9 +1,22 @@
 use sled::IVec;
 use ssz_derive::{Decode, Encode};
 use strata_identifiers::OLTxId;
+use thiserror::Error;
 use typed_sled::codec::{CodecError, KeyCodec, ValueCodec};
 
 use crate::define_table_without_codec;
+
+/// Errors raised while decoding mempool SSZ payloads from sled.
+#[derive(Debug, Error)]
+enum MempoolDecodeError {
+    /// Decoding the SSZ-encoded [`OLTxId`] key failed.
+    #[error("failed to decode mempool transaction key from SSZ")]
+    Key(#[source] ssz::DecodeError),
+
+    /// Decoding the SSZ-encoded [`MempoolTxEntry`] value failed.
+    #[error("failed to decode mempool transaction entry from SSZ")]
+    Value(#[source] ssz::DecodeError),
+}
 
 /// Wrapper type for storing transaction bytes and ordering metadata.
 #[derive(Clone, Debug, PartialEq, Eq, Encode, Decode)]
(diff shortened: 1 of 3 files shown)


crates/db/store-sled/src/macros.rs
```diff
@@ -216,6 +216,35 @@ macro_rules! impl_borsh_value_codec {
...
+                    ::typed_sled::codec::CodecError::SerializationFailed {
+                        schema: $table_name::tree_name(),
+                        source: err.into(),
+                    }
+                })
+            }
+        }
+    };
+}
+
```
delbonis, line 257: So this is a bit of a weird case that breaks our model.  This *works*, but it doesn't actually give us the zero-copy advantages that rkyv has.

We'd want to use `access` from the [high-level API](https://docs.rs/rkyv/latest/rkyv/api/high/index.html).  But this means that we can't directly use `ValueCodec`, since that assumes that the decodec form is the same thing as the thing that encoded it.  I'm not sure what to do here since. this might require us changing the typed-sled API to work with this properly. 😬

This is especially annoying since `access` is defined as

```rs
pub fn access<T, E>(bytes: &[u8]) -> Result<&T, E>
```

---

Maybe we can add an associated type called `Decoded` with a default value of `Self` and have `decode_value` return that?

I'm thinking it might be helpful for us to have a type like this, going along with the above thing:

```rs
struct Rk<B, P> { // pronounced "arc", as in "rkyv"
  buf: B,
  _pd: P,
}

impl<B: AsRef<[u8]>, P: Portable> Rk<B, P> {
  pub fn try_new<E: ...>(buf: B) -> Result<Self, E> {
    let _ = rkyv::access(buf.as_ref())?;
    Self { buf, _pd: PhantomData }
  }
}

impl<B: AsRef<[u8]>, P: Portable> AsRef<P> for Rk<B, P> {
  pub fn as_ref(&self) -> &P {
    rkyv::access_unchecked(self.buf.as_ref())
  }
}
```

I'm surprised that rkyv doesn't provide a type that lets you do this.

storopoli: Ok now that https://github.com/alpenlabs/typed-sled/pull/18 is merged we can solve this

crates/db/store-sled/src/mempool/schemas.rs
```diff
@@ -50,14 +51,18 @@ impl KeyCodec<MempoolTxSchema> for OLTxId {
...
 
-    fn decode_value(data: &[u8]) -> Result<Self, CodecError> {
-        ssz::Decode::from_ssz_bytes(data).map_err(|err| CodecError::SerializationFailed {
-            schema: MempoolTxSchema::tree_name(),
-            source: anyhow::anyhow!("SSZ decode error: {:?}", err).into(),
+    fn decode_value(data: IVec) -> Result<Self::Decoded, CodecError> {
+        ssz::Decode::from_ssz_bytes(data.as_ref()).map_err(|err| {
+            CodecError::DeserializationFailed {
+                schema: MempoolTxSchema::tree_name(),
+                source: anyhow::anyhow!("SSZ decode error: {:?}", err).into(),
```
delbonis: Can we add an error variant for this instead of making an anyhow error that we immediately put back into a thiserror error?

storopoli: See 3843849b05c2531b5c528216cd1d062a3b3070c0

PR #1516, opened 2026-03-20
feat(observability): Prometheus /metrics endpoint with tracing bridge (STR-2557)

voidash:
Adds Prometheus metrics support using the `metrics` crate (same facade Reth uses).

- **Chain progress gauges**: `alpen_l1_tip_height`, `alpen_l2_sync_tip_slot`, `alpen_l2_finalized_slot`, `alpen_ol_tip_slot`
- **Tracing-to-metrics bridge**: `MetricsLayer` captures span busy/idle timing as `alpen_span_busy_seconds` / `alpen_span_idle_seconds` histograms — flows to Prometheus instead of only Loki
- **DB transaction histogram**: `alpen_db_transaction_duration_seconds` for sled transactions (independent of log level)
- **`strata` binary**: `metrics-exporter-prometheus` serves `/metrics` on configurable `logging.metrics_port`
- **`alpen-client`**: our gauges automatically appear on Reth's built-in `--metrics` endpoint

All calls are no-ops when no recorder is installed.

[STR-2557](https://alpenlabs.atlassian.net/browse/STR-2557)

[STR-2557]: https://alpenlabs.atlassian.net/browse/STR-2557?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -255,7 +255,7 @@ strata-crypto = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.
 strata-identifiers = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.1.0-alpha-rc16" }
 strata-l1-envelope-fmt = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.1.0-alpha-rc16" }
 strata-l1-txfmt = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.1.0-alpha-rc16" }
-strata-logging = { git = "https://github.com/alpenlabs/strata-common", rev = "8aaa27a6b251a6f232e02765a43577cf2020f423" }
+strata-logging = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.1.0-alpha-rc17" }
 strata-merkle = { git = "https://github.com/alpenlabs/strata-common", features = [
   "legacy_compact",
 ], tag = "v0.1.0-alpha-rc16" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -255,7 +255,7 @@ strata-crypto = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.
 strata-identifiers = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.1.0-alpha-rc16" }
 strata-l1-envelope-fmt = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.1.0-alpha-rc16" }
 strata-l1-txfmt = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.1.0-alpha-rc16" }
-strata-logging = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.1.0-alpha-rc16" }
+strata-logging = { git = "https://github.com/alpenlabs/strata-common", rev = "8aaa27a6b251a6f232e02765a43577cf2020f423" }
 strata-merkle = { git = "https://github.com/alpenlabs/strata-common", features = [
   "legacy_compact",
 ], tag = "v0.1.0-alpha-rc16" }
@@ -449,6 +449,8 @@ jsonrpsee-http-client = { version = "0.26.0" }
 jsonrpsee-types = { version = "0.26.0" }
 k256 = { version = "0.13.4", features = ["schnorr"] }
 lru = "0.17"
+metrics = "0.24"
+metrics-exporter-prometheus = "0.16"
 mockall = "0.14"
 musig2 = { version = "0.1.0", features = ["serde"] }
 num_enum = "0.7"
--- a/bin/strata/src/main.rs
+++ b/bin/strata/src/main.rs
@@ -49,8 +49,22 @@ fn main() -> Result<()> {
         .build()
         .map_err(InitError::RuntimeBuild)?;
 
-    // Initialize logging
+    // Install Prometheus metrics recorder before logging so the MetricsLayer
+    // can record from the very first spans. Bind to loopback only. Operators
+    // who need remote scraping can front it with a reverse proxy.
+    if let Some(port) = config.logging.metrics_port {
+        metrics_exporter_prometheus::PrometheusBuilder::new()
+            .with_http_listener(([127, 0, 0, 1], port))
+            .install()
+            .map_err(|e| anyhow!("failed to install Prometheus metrics exporter: {e}"))?;
+    }
+
+    // Initialize logging (MetricsLayer is only added when a recorder is installed).
     init_logging(rt.handle(), &config);
+
+    if config.logging.metrics_port.is_some() {
+        info!("Prometheus metrics endpoint started");
+    }
 
     // Validate sequencer flag isn't used when sequencer feature is disabled.
     #[cfg(not(feature = "sequencer"))]
@@ -116,5 +130,6 @@ fn init_logging(rt: &Handle, config: &strata_config::Config) {
         log_file_prefix: config.logging.log_file_prefix.as_deref(),
         json_format: config.logging.json_format,
         default_log_prefix: "alpen",
+        enable_metrics_layer: config.logging.metrics_port.is_some(),
     });
 }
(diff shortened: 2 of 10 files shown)


crates/common/src/logging/manager.rs
```diff
@@ -105,11 +105,14 @@ pub fn init(config: LoggerConfig) {
         tracing_opentelemetry::layer().with_tracer(tt)
     });
 
+    let metrics_layer = super::metrics_layer::MetricsLayer;
```
delbonis: Import it so we can avoid the `super::metrics_layer::` here?

crates/asm/worker/Cargo.toml
```diff
@@ -20,6 +20,8 @@ strata-service.workspace = true
 strata-state.workspace = true
 strata-tasks.workspace = true
 
+metrics.workspace = true
+
 anyhow.workspace = true
 async-trait.workspace = true
```
delbonis: import grouping

crates/consensus-logic/Cargo.toml
```diff
@@ -39,6 +39,8 @@ strata-storage.workspace = true
 strata-tasks.workspace = true
 zkaleido.workspace = true
 
+metrics.workspace = true
+
 anyhow.workspace = true
 bitcoin.workspace = true
 bitcoind-async-client.workspace = true
```
delbonis: grouping

crates/common/src/logging/metrics_layer.rs
```diff
@@ -0,0 +1,71 @@
...
+                let busy = timing.busy;
+                let idle = total.saturating_sub(busy);
+
+                metrics::histogram!("alpen_span_busy_seconds", "span" => name.to_string())
+                    .record(busy.as_secs_f64());
+                metrics::histogram!("alpen_span_idle_seconds", "span" => name.to_string())
+                    .record(idle.as_secs_f64());
+            }
+        }
+    }
```
delbonis: This is *super* clever and super useful for finding cases where we block the executor.

This code is fairly important since it'll get hit many times per second, so it maybe makes sense to try to micro-optimize it.  I would think that *if it's safe to assume* that we don't enter/exit a span multiple times (which we shouldn't in correct code), we could avoid storing `last_entered` as an `Option` and always store the value directly.  We should never call `on_exit` without a recent call to `on_enter`, so I don't think this could ever mess up right?

Also use millis/micros here instead of float seconds, especially in this case because *ideally* spans are only *busy* for a minuscule amount of time.

Is the metrics collection only turned on if specifically queried?  I'm worried about the perf impact of there being quite a few heap allocs (extension entry, two strings) for each span that gets created.

crates/db/store-sled/src/config.rs
```diff
@@ -62,8 +62,12 @@ impl SledDbConfig {
...
     {
-        trees
+        let start = Instant::now();
+        let result = trees
             .transaction_with_retry(self.backoff.as_ref(), self.retry_count.into(), f)
-            .map_err(to_db_error)
+            .map_err(to_db_error);
+        metrics::histogram!("alpen_db_transaction_duration_seconds")
+            .record(start.elapsed().as_secs_f64());
+        result
```
delbonis: Uhh this is a bit weird since the transaction duration will vary a *lot* between database types, and even within them depending on what kinds of operations being done.  To do this I could imagine we would want to *somehow* pass through a string from the referencing the *operation* being done (not just the database type), but that seems kinda hard to do here without impacting ergonomics substantially.  Maybe it's fine to just remove it since we'll be getting information like this in a span.

Also again I would think we'd want to record it as millis/micros instead of float seconds.

crates/sync/src/state.rs
delbonis, line 1: Ohhhh wait don't tell me it *has* to be a f64 does it???

delbonis (changes requested):
This is pretty cool to get visibility into.  Some questions about specific execution choices though.  I looked at the docs and yeah it's weird they require it be an `f64`.

Also, the `alpen_` prefix doesn't quite make sense here, it should be `strata_` in most cases.

I don't know if it makes sense to do this in this PR but I could see this being useful metrics to collect as well:

* OL block execution duration
* OL block construction duration

crates/common/src/logging/metrics_layer.rs
voidash: `MetricsLayer` is now conditional or gated behind `enable_metrics_layer` on `LoggingInitConfig`, so zero per-span overhead when no recorder is installed. Also i moved Prometheus recorder
  installation before logging init so metrics flow from the first spans.

bin/strata/src/main.rs
```diff
@@ -47,9 +47,22 @@ fn main() -> Result<()> {
...
 
-    // Initialize logging
+    // Install Prometheus metrics recorder before logging so the MetricsLayer
+    // can record from the very first spans.
+    if let Some(port) = config.logging.metrics_port {
+        metrics_exporter_prometheus::PrometheusBuilder::new()
+            .with_http_listener(([0, 0, 0, 0], port))
+            .install()
+            .expect("failed to install Prometheus metrics exporter");
+    }
```
storopoli: This exposes `/metrics` on all interfaces with no auth. Unless that is explicitly intended, this is an avoidable security/ops footgun; metrics endpoints usually default to loopback or at least make the bind address configurable.

bin/strata/src/main.rs
```diff
@@ -47,9 +47,22 @@ fn main() -> Result<()> {
...
         .map_err(InitError::RuntimeBuild)?;
 
-    // Initialize logging
+    // Install Prometheus metrics recorder before logging so the MetricsLayer
+    // can record from the very first spans.
+    if let Some(port) = config.logging.metrics_port {
+        metrics_exporter_prometheus::PrometheusBuilder::new()
+            .with_http_listener(([0, 0, 0, 0], port))
+            .install()
+            .expect("failed to install Prometheus metrics exporter");
```
storopoli: This `expect` turns routine startup problems like “port already in use” or “recorder already installed” into an immediate panic. This should be surfaced as an initialization error instead of crashing the process.

storopoli (changes requested):

bin/strata/src/main.rs
voidash: It is not common pattern to expose the `/metrics` . I have made modification such that it's on loopback by default.

storopoli (dismissed):
ACK c201fc36dc19c467743df625d368128c5cd7cb94

delbonis (dismissed):

storopoli:
@alpenlabs/chain-protocol please take a look

prajwolrg (changes requested):
The changes itself looks good. Please upstream the changes to https://github.com/alpenlabs/strata-common/tree/main/crates/logging as well.

I created a separate PR https://github.com/alpenlabs/alpen/pull/1657 where we rely on `strata-common`'s `strata-logging` crate instead.

Cargo.toml
```diff
@@ -257,7 +257,7 @@ strata-crypto = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.
 strata-identifiers = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.1.0-alpha-rc16" }
 strata-l1-envelope-fmt = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.1.0-alpha-rc16" }
 strata-l1-txfmt = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.1.0-alpha-rc16" }
-strata-logging = { git = "https://github.com/alpenlabs/strata-common", tag = "v0.1.0-alpha-rc16" }
+strata-logging = { git = "https://github.com/alpenlabs/strata-common", rev = "8aaa27a6b251a6f232e02765a43577cf2020f423" }
```
storopoli: We should use a tag here

storopoli (changes requested):

Cargo.toml
voidash: is it ok if i bump the tag?

storopoli: Yes, definitely

storopoli (approved):
ACK 91da32d8b208bbdf46d5b42a687e5ba9cc3b213d

delbonis (approved):

storopoli (approved):
ACK 8351f04a989ab70a4c082820edb26e42bb5b77c2

PR #1528, opened 2026-03-24
lint: add lint for anyhow errors in thiserror errors

delbonis:
## Description

This PR just adds a style lint for anyhow errors in thiserror errors.  This pattern caused an issue, so we're resolving the root cause, at least where the thiserror macro is used to generate the conversions.

Claude was used to write the script.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

### DIFF
--- a/contrib/find_anyhow_in_thiserror.py
+++ b/contrib/find_anyhow_in_thiserror.py
@@ -1,0 +1,135 @@
+#!/usr/bin/env python3
+"""Detect anyhow::Error wrapped with #[from] or #[source] inside thiserror error types.
+
+Wrapping anyhow::Error this way obscures the original error's type information,
+making debugging and tracing harder. Use a specific error type instead.
+"""
+
+import argparse
+import re
+import sys
+from pathlib import Path
+
+
+def check_file(path: Path) -> list[tuple[int, list[str]]]:
+    """Return list of (line_number, all_lines) for violations in the file."""
+    try:
+        text = path.read_text()
+    except (OSError, UnicodeDecodeError):
+        return []
+
+    # Quick bail: file must use both thiserror and anyhow::Error
+    if "thiserror" not in text or "anyhow::Error" not in text:
+        return []
+
+    lines = text.splitlines()
+    violations: list[tuple[int, list[str]]] = []
+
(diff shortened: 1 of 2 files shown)


contrib/find_anyhow_in_thiserror.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python3
```
storopoli, line 1: Should we should use the `uv` shebang instead of this?

See: https://docs.astral.sh/uv/guides/scripts/#using-a-shebang-to-create-an-executable-file

contrib/find_anyhow_in_thiserror.py
```diff
@@ -0,0 +1,135 @@
...
+
+def main() -> int:
+    parser = argparse.ArgumentParser(
+        description="Detect anyhow::Error wrapped with #[from]/#[source] in thiserror types."
+    )
+    parser.add_argument(
+        "-C",
+        "--context",
+        type=int,
+        default=0,
```
storopoli: Move this to default to `2` and remove the `-C 2` in `./contrib/find_anyhow_in_thiserror.py -C 2` in `.justfile` to be just

`./contrib/find_anyhow_in_thiserror.py`

storopoli (changes requested):

contrib/find_anyhow_in_thiserror.py
delbonis: I mean, it's only using things from the Python stdlib and leaving it as-is makes it more portable (don't need to get uv).

storopoli: Ok fair point!

storopoli (approved):
ACK 203a35e4853e9e3f157235abb948b027a393ef45

storopoli:
@delbonis I think we need to enable this lint only for changes lines (`+` lines in the git diff)

barakshani:
@delbonis what is the status of this PR? It sits here for almost 4 months now.

delbonis:
@storopoli @barakshani Going to revive this now, actually.

PR #1585, opened 2026-04-09
feat: support predicate key update of snark account

prajwolrg:
## Description

Adds a new ASM log type, `EePredicateKeyUpdate`, that carries `(account: AccountSerial, new_predicate: PredicateKey)`, and wires up the OL STF to consume it during manifest processing — applying the new predicate key to the targeted snark account's `update_vk` field.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This is the **consumer-only** half of the EE verifying-key rotation flow. The ASM-side producer (emitting the log from the relevant subprotocol) lives in the separate ASM repository and will be implemented there in a follow-up. Until that lands, the new log type is dead code at runtime — unknown log types are already silently skipped by `process_asm_log`, so there is no functional impact on existing flows.

- [ ]  Yes
- [x]  No

## Checklist


crates/ol/state-support-types/src/indexer_layer.rs
```diff
@@ -160,6 +160,15 @@ impl<S: ISnarkAccountStateMut> ISnarkAccountStateMut for IndexerSnarkAccountStat
...
     }
+
+    fn set_update_vk(&mut self, new_vk: PredicateKey) {
+        // NOTE: predicate-key updates are intentionally NOT pushed to
+        // `self.writes`. They are observable via the ASM log that drove the
+        // mutation, so DA consumers can replay logs to reconstruct the new vk.
+        // Revisit if a future indexer needs DA-level tracking of vk rotations.
+        self.inner.set_update_vk(new_vk);
+        self.modified = true;
+    }
```
delbonis, line 171: This assumes that predkey changes can only happen from ASM logs, which is definitely not something that will always be true.  The indexer layer should always index every write, regardless of the source we think it'll be triggered from.

crates/asm/logs/src/ee_predicate_update.rs
```diff
@@ -0,0 +1,85 @@
...
+/// The target account is identified by its [`AccountSerial`]; the OL STF
+/// resolves the serial and applies the new predicate key during manifest
+/// processing.
+#[derive(Debug, Clone, Codec)]
+pub struct EePredicateKeyUpdate {
+    /// Serial of the snark account whose predicate key is being updated.
+    account: AccountSerial,
+
+    /// New predicate key to install on the target account.
+    new_predicate: CodecSsz<PredicateKey>,
```
delbonis, line 22: Kindaaa don't like having ssz in this encoding, but I guess it's fine since it's just an asm log.

delbonis (changes requested):
Very clean, just some minor comments.

crates/ol/stf/src/manifest_processing.rs
```diff
@@ -164,3 +174,25 @@ fn process_checkpoint_tip_update<S: IStateAccessor>(
 
     Ok(())
 }
+
+fn process_ee_predicate_key_update<S: IStateAccessor>(
+    state: &mut S,
+    data: &EePredicateKeyUpdate,
+) -> ExecResult<()> {
+    // Resolve the account serial. Silently skip if not found, matching the
```
storopoli: I don't think we should silently drop this log on unknown serial or non-snark targets. This is a control-plane key rotation, not a best-effort user deposit: if ASM emits a bad target, OL keeps the old `update_vk`, and the failure only shows up later when the next account update is verified against the stale key here: https://github.com/alpenlabs/alpen/blob/b04e7451269b4752e9d0669ebcbdcd26e500546b/crates/ol/stf/src/proof_verification.rs#L183-L198 . I'd rather return an execution error here so the bad manifest is rejected immediately.

storopoli (changes requested):

crates/asm/logs/src/ee_predicate_update.rs
```diff
@@ -0,0 +1,85 @@
...
+#[cfg(test)]
+mod tests {
+    use strata_codec::{decode_buf_exact, encode_to_vec};
+    use strata_identifiers::AccountSerial;
+    use strata_predicate::PredicateKey;
+
+    use super::*;
+
+    #[test]
+    fn ee_predicate_key_update_roundtrip() {
```
bewakes, line 64: Would this test be more complete with more cases(or ideally proptest) that checks for other variants of `PredicateKey` as well?

bewakes (dismissed):
This is great!

crates/ol/stf/src/manifest_processing.rs
bewakes: I think this is tricky because someone possibly can spam OL with invalid serials. Logs would be better instead.

prajwolrg: The OL STF can't return an error from manifest processing without halting the chain. `ExecError` from `process_block_manifests` is block-fatal — it propagates through `execute_block_manifests` and fails both block construction and verification. If the terminal block of an epoch can't seal, the checkpoint can't advance.

ASM manifests are produced ASM STF over L1 data; they are not user input the OL gets to reject. Every existing log handler in this file follows the same convention — silent skip on parse failure, unknown type, and unknown referent. 

If we want to surface staleness sooner, the right tool is observability — a `warn!` or `error!` log here.

crates/asm/logs/src/ee_predicate_update.rs
prajwolrg: I don't think that's really needed, since we only need to verify that the PredicateKey was updated, adding other variants doesn't add any value.

crates/ol/state-support-types/src/indexer_layer.rs
prajwolrg: Done in [190aa4a](https://github.com/alpenlabs/alpen/pull/1585/commits/190aa4a1fd1f1bf381c456c0a8e4c3b1fd172e2f).

crates/ol/stf/src/manifest_processing.rs
prajwolrg: I've emitted the `warn!` log in [2a12b8e](https://github.com/alpenlabs/alpen/pull/1585/commits/2a12b8e94b5dba73f5114be625fc71133f045f99).

crates/asm/logs/src/ee_predicate_update.rs
delbonis: This isn't necessary because the predkey type is already exhaustively tested, so we only need n+m tests instead of n*m.

delbonis (approved):

crates/asm/logs/src/ee_predicate_update.rs
storopoli, line 1: Local ASM crates were removed in #1642 and `strata-asm-logs` is now pulled from the external `alpenlabs/asm` repo, which already contains this file byte-for-byte

storopoli (changes requested):

crates/asm/logs/src/ee_predicate_update.rs
prajwolrg: This isn't even used. Must have been leftover after rebase.

storopoli (approved):
ACK eeb66954ff99c24a418198455fe67fb56ff1471e

PR #1587, opened 2026-04-09
STR 3018: `get_account_epoch_summary`: Fetch and serve account's inbox messages in an epoch

krsnapaudel:
## Description

- Add `get_account_inbox_messages` to `OLRpcProvider` trait for fetching inbox messages by index range from the account inbox MMR
- Wire `get_acct_epoch_summary` to compute `[prev_next_inbox_msg_idx, cur_next_inbox_msg_idx)` and populate `RpcUpdateInputData.messages` instead of returning `vec![]`
- Refactor `get_canonical_epoch_commitment_at` to accept `Epoch` instead of `u64`

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-3018](https://alpenlabs.atlassian.net/browse/STR-3018)

[STR-3018]: https://alpenlabs.atlassian.net/browse/STR-3018?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/strata/src/rpc/provider.rs
+++ b/bin/strata/src/rpc/provider.rs
@@ -3,10 +3,12 @@
 use std::sync::Arc;
 
 use async_trait::async_trait;
+use ssz::Decode;
+use strata_acct_types::MessageEntry;
 use strata_asm_common::AsmManifest;
 use strata_checkpoint_types::EpochSummary;
 use strata_csm_types::CheckpointL1Ref;
-use strata_db_types::DbResult;
+use strata_db_types::{DbError, DbResult, MmrId};
 use strata_identifiers::{AccountId, Epoch, L1Height, OLBlockId, OLTxId};
 use strata_ol_chain_types_new::{OLBlock, OLTransaction};
 use strata_ol_mempool::{MempoolHandle, OLMempoolResult};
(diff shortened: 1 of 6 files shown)


bin/strata/src/rpc/provider.rs
```diff
@@ -104,6 +103,36 @@ impl OLRpcProvider for NodeRpcProvider {
             .await
     }
 
+    async fn get_account_inbox_messages(
+        &self,
+        account_id: AccountId,
+        start_idx: u64,
+        end_idx_exclusive: u64,
```
storopoli, line 110: We should think about adding a sanity cap like `MAX_MESSAGES_TO_RETRIEVE` or add pagination somehow in a follow-up change.
If agreed, this can be a tech-debt ticket

bin/strata/src/rpc/provider.rs
```diff
@@ -104,6 +103,36 @@ impl OLRpcProvider for NodeRpcProvider {
...
+        let mut messages = Vec::with_capacity((end_idx_exclusive - start_idx) as usize);
+        for idx in start_idx..end_idx_exclusive {
+            let preimage = mmr_handle.get(idx).await?;
+            let message = MessageEntry::from_ssz_bytes(&preimage).map_err(|e| {
+                DbError::Other(format!(
+                    "failed to decode account inbox message at index {idx} for account {account_id}: {e}"
+                ))
+            })?;
+            messages.push(message);
+        }
```
storopoli, line 131: Each message is fetched individually via `mmr_handle.get(idx).await?`. A batch API on `MmrIndexHandle` would be a good future optimization.

storopoli (dismissed):
ACK 8d6567d60d1d3f8c662e8e2228371fcb28bb1a9e

Couple non-blocking comments that can be tracked down in tickets...
Nice work.

bin/strata/src/rpc/provider.rs
krsnapaudel: Created [a ticket for it](https://alpenlabs.atlassian.net/browse/STR-3036).

bin/strata/src/rpc/provider.rs
krsnapaudel: @storopoli How will pagination work? Will RPC calls be paginated or internal handlers? I would think RPC calls right? Then `MAX_MESSAGES_TO_RETRIEVE` is probably a good idea. The RPC api needs to specify the limit.

storopoli: I thought that `jsonrpsee` would support "pagination" easily. Did some research and it is not the case. so the `MAX_MESSAGES_TO_RETRIEVE` would be ideal.

storopoli: Something like

```rs
const MAX_RAW_BLOCKS_RANGE: usize = 1_000;

if block_count > MAX_RAW_BLOCKS_RANGE {
    return Err(invalid_params_error("Invalid block range"));
}
```

krsnapaudel: @storopoli Actually this is an internal call. Not RPC api. Does it still make sense to put a limit? If we do, the `getAccountEpochSummary` will fail due to internal limit. The api itself has not way of specifying the limit.

storopoli: Let's leave as it is. I don't see this being a major source of issues, we could have high resource usage in the future if this is spammed in the future.

storopoli (approved):
ACK 6d7f73b61e5a42eb1135e1981b1332d78a22203e

PR #1637, opened 2026-04-10
feat(btcio): add mempool esplora fee policy for writer (STR-2997)

storopoli:
## Description

This PR adds a new btcio writer fee policy, `mempool`, alongside the now-renamed existing `smart` (`BitcoinD`) and `fixed` modes.

When `fee_policy = "mempool"`, btcio fetches recommended fees from `{mempool_base_url}/api/v1/fees/recommended`, parses the JSON response, and uses `fastestFee` directly as the fee rate in sat/vB.

`BitcoinD` now uses the raw estimate directly. If the mempool fee lookup fails at runtime due to transport, HTTP status, or JSON parsing issues, btcio logs a `WARN` and falls back to `estimatesmartfee(1)`.

In addition, this PR exposes `mempool_base_url` in btcio writer config and updates examples/test config scaffolding accordingly.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The new config shape is:

```toml
[btcio.writer]
fee_policy = "mempool"
mempool_base_url = "https://mempool.space"
```

Supported mempool Esplora-compatible backends include:

- Official: https://mempool.space/docs/api/rest#get-recommended-fees
- Testnet Alpen: https://bitcoin.testnet.alpenlabs.io/docs/api/rest#get-recommended-fees
- Development Deployment: https://bitcoin.development.stratabtc.org/docs/api/rest#get-recommended-fees

The implementation treats `mempool_base_url` as a base URL and appends `/api/v1/fees/recommended`, so e.g. `https://mempool.space` resolves to `https://mempool.space/api/v1/fees/recommended`.

Behavioral points to review:

- mempool uses fastestFee directly
- runtime mempool lookup failures emit WARN and fall back to `estimatesmartfee(1)`
- missing or invalid mempool_base_url is treated as a config error

- [x]  Yes
- [ ]  No

- [STR-2997](https://alpenlabs.atlassian.net/browse/STR-2997)
- https://mempool.space/docs/api/rest#get-recommended-fees
- https://bitcoin.testnet.alpenlabs.io/docs/api/rest#get-recommended-fees
- https://bitcoin.development.stratabtc.org/docs/api/rest#get-recommended-fees
- [btcio Fee Estimation](https://www.notion.so/btcio-Fee-Estimation-335901ba000f8052a9c0da39b037f952)

## Checklist

## Related Issues

STR-2997

[STR-2997]: https://alpenlabs.atlassian.net/browse/STR-2997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/btcio/src/writer/fees.rs
+++ b/crates/btcio/src/writer/fees.rs
@@ -5,7 +5,7 @@ use anyhow::{anyhow, Context};
 use bitcoind_async_client::traits::Reader;
 use reqwest::Url;
 use serde::Deserialize;
-use strata_config::btcio::{FeePolicy, WriterConfig};
+use strata_config::btcio::{FeePolicy, MempoolExplorerFeePolicy, WriterConfig};
 use tracing::warn;
 
 /// Represents the response from the mempool.space recommended fees endpoint.
@@ -35,7 +35,9 @@ pub(crate) async fn resolve_fee_rate<R: Reader>(
             .estimate_smart_fee(*conf_target)
             .await
             .context("failed to estimate smart fee"),
-        FeePolicy::MempoolExplorer => resolve_mempool_fee_rate(client, config).await,
+        FeePolicy::MempoolExplorer { policy } => {
+            resolve_mempool_fee_rate(client, config, *policy).await
+        }
         FeePolicy::Fixed { fee_rate } => Ok(*fee_rate),
     }
 }
@@ -45,6 +47,7 @@ pub(crate) async fn resolve_fee_rate<R: Reader>(
 async fn resolve_mempool_fee_rate<R: Reader>(
     client: &R,
     config: &WriterConfig,
+    mempool_fee_policy: MempoolExplorerFeePolicy,
 ) -> anyhow::Result<u64> {
     let base_url = config
         .mempool_base_url
@@ -53,7 +56,13 @@ async fn resolve_mempool_fee_rate<R: Reader>(
     let url = mempool_recommended_fees_url(base_url)?;
 
     match fetch_mempool_recommended_fees(url).await {
-        Ok(fees) => Ok(fees.fastest_fee),
+        Ok(fees) => match mempool_fee_policy {
+            MempoolExplorerFeePolicy::Fastest => Ok(fees.fastest_fee),
+            MempoolExplorerFeePolicy::HalfHour => Ok(fees.half_hour_fee),
+            MempoolExplorerFeePolicy::Hour => Ok(fees.hour_fee),
+            MempoolExplorerFeePolicy::Economy => Ok(fees.economy_fee),
+            MempoolExplorerFeePolicy::Minimum => Ok(fees.minimum_fee),
+        },
         Err(err) => {
             warn!(
                 %base_url,
@@ -96,7 +105,7 @@ async fn fetch_mempool_recommended_fees(url: Url) -> anyhow::Result<MempoolRecom
 mod tests {
     use std::sync::Arc;
 
-    use strata_config::btcio::{FeePolicy, WriterConfig};
+    use strata_config::btcio::{FeePolicy, MempoolExplorerFeePolicy, WriterConfig};
     use tokio::{
         io::{AsyncReadExt, AsyncWriteExt},
         net::TcpListener,
@@ -182,7 +191,9 @@ mod tests {
         .await;
         let client = TestBitcoinClient::new(1);
         let config = WriterConfig {
-            fee_policy: FeePolicy::MempoolExplorer,
+            fee_policy: FeePolicy::MempoolExplorer {
+                policy: MempoolExplorerFeePolicy::Fastest,
+            },
             mempool_base_url: Some(server),
             ..WriterConfig::default()
         };
@@ -195,11 +206,36 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_resolve_fee_rate_uses_selected_mempool_policy() {
+        let server = spawn_single_response_server(
+            "200 OK",
+            "{\"fastestFee\":7,\"halfHourFee\":6,\"hourFee\":5,\"economyFee\":4,\"minimumFee\":3}",
+        )
+        .await;
+        let client = TestBitcoinClient::new(1);
+        let config = WriterConfig {
+            fee_policy: FeePolicy::MempoolExplorer {
+                policy: MempoolExplorerFeePolicy::Economy,
+            },
+            mempool_base_url: Some(server),
+            ..WriterConfig::default()
+        };
+
+        let fee_rate = resolve_fee_rate(&client, &config)
+            .await
+            .expect("mempool fee lookup should succeed");
+
+        assert_eq!(fee_rate, 4);
+    }
+
+    #[tokio::test]
     async fn test_resolve_fee_rate_falls_back_to_smart_fee_on_invalid_json() {
         let server = spawn_single_response_server("200 OK", "not-json").await;
         let client = TestBitcoinClient::new(1);
         let config = WriterConfig {
-            fee_policy: FeePolicy::MempoolExplorer,
+            fee_policy: FeePolicy::MempoolExplorer {
+                policy: MempoolExplorerFeePolicy::Fastest,
+            },
             mempool_base_url: Some(server),
             ..WriterConfig::default()
         };
@@ -216,7 +252,9 @@ mod tests {
         let server = spawn_single_response_server("500 Internal Server Error", "").await;
         let client = TestBitcoinClient::new(1);
         let config = WriterConfig {
-            fee_policy: FeePolicy::MempoolExplorer,
+            fee_policy: FeePolicy::MempoolExplorer {
+                policy: MempoolExplorerFeePolicy::Fastest,
+            },
             mempool_base_url: Some(server),
             ..WriterConfig::default()
         };
@@ -232,7 +270,9 @@ mod tests {
     async fn test_resolve_fee_rate_errors_when_mempool_base_url_is_missing() {
         let client = TestBitcoinClient::new(1);
         let config = WriterConfig {
-            fee_policy: FeePolicy::MempoolExplorer,
+            fee_policy: FeePolicy::MempoolExplorer {
+                policy: MempoolExplorerFeePolicy::Fastest,
+            },
             mempool_base_url: None,
             ..WriterConfig::default()
         };
@@ -250,7 +290,9 @@ mod tests {
     async fn test_resolve_fee_rate_errors_when_mempool_base_url_is_invalid() {
         let client = TestBitcoinClient::new(1);
         let config = WriterConfig {
-            fee_policy: FeePolicy::MempoolExplorer,
+            fee_policy: FeePolicy::MempoolExplorer {
+                policy: MempoolExplorerFeePolicy::Fastest,
+            },
             mempool_base_url: Some("not a url".to_string()),
             ..WriterConfig::default()
         };
--- a/crates/btcio/src/writer/fees.rs
+++ b/crates/btcio/src/writer/fees.rs
@@ -30,7 +30,7 @@ pub(crate) async fn resolve_fee_rate<R: Reader>(
     client: &R,
     config: &WriterConfig,
 ) -> anyhow::Result<u64> {
-    match &config.fee_policy {
+    let fee_rate = match &config.fee_policy {
         FeePolicy::BitcoinD { conf_target } => client
             .estimate_smart_fee(*conf_target)
             .await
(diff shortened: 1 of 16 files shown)


barakshani:
> This change also removes the previous * 2 multiplier

why? I thought we wanted to do so only when we have fee bump in place

crates/config/src/btcio.rs
```diff
@@ -28,15 +28,21 @@ pub struct WriterConfig {
...
+    /// Base URL for mempool.space-compatible fee API.
+    pub mempool_base_url: Option<String>,
 }
 
 /// Definition of how fees are determined while creating l1 transactions.
 #[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq)]
 #[serde(rename_all = "lowercase")]
 pub enum FeePolicy {
-    /// Use estimatesmartfee.
+    /// Use mempool.space explorer recommended fees endpoint.
```
barakshani: are you sure you want to specify the mempool fee provider in the code? I think this is exported to the config file

crates/btcio/src/writer/fees.rs
```diff
@@ -0,0 +1,277 @@
...
+    config: &WriterConfig,
+) -> anyhow::Result<u64> {
+    let base_url = config
+        .mempool_base_url
+        .as_deref()
+        .ok_or_else(|| anyhow!("mempool_base_url must be set when fee_policy = \"mempool\""))?;
+    let url = mempool_recommended_fees_url(base_url)?;
+
+    match fetch_mempool_recommended_fees(url).await {
+        Ok(fees) => Ok(fees.fastest_fee),
```
barakshani: maybe we should determine this via config?

crates/btcio/src/writer/fees.rs
```diff
@@ -0,0 +1,277 @@
...
+fn mempool_recommended_fees_url(base_url: &str) -> anyhow::Result<Url> {
+    let mut url =
+        Url::parse(base_url).with_context(|| format!("invalid mempool_base_url: {base_url}"))?;
+
+    if !url.path().ends_with('/') {
+        let path = format!("{}/", url.path());
+        url.set_path(&path);
+    }
+
+    url.join("api/v1/fees/recommended")
```
barakshani: I assume this works specifically for mempool.space, why not have it with base_url for further flexibility?

barakshani (commented):
cool! excited to see this getting into our code

crates/btcio/src/writer/fees.rs
```diff
@@ -0,0 +1,277 @@
...
+    fastest_fee: u64,
+    #[serde(rename = "halfHourFee")]
+    half_hour_fee: u64,
+    #[serde(rename = "hourFee")]
+    hour_fee: u64,
+    #[serde(rename = "economyFee")]
+    economy_fee: u64,
+    #[serde(rename = "minimumFee")]
+    minimum_fee: u64,
+}
```
barakshani, line 31: After poking with mempool.space endpoint, looks like the minimum can only be as low as 1. 
However, it turns out there's another endpoint that gives decimal as well: https://mempool.space/api/v1/fees/precise

I am fine sticking to whole numbers, and have no problems with always paying at least 1, but wanted to bring this to your attention.

crates/config/src/btcio.rs
storopoli: Yeah this is confusing. I wanted to name the explorer from the mempool.space. Not the URL. What you’ve suggested is exactly what we’re doing. There’s a configuration that you need to specify the URL. I’ll check how the explorer is officially called and will update the docstrings

crates/btcio/src/writer/fees.rs
storopoli: No. It works for almost all mempool explorers including our testnet live (and future) and current staging and dev deployments.

storopoli: The idea is to not use mempool.space but our own for both testnet4 and mainnet.

crates/btcio/src/writer/fees.rs
storopoli: The problem with the decimal one is that I did not see that endpoint in your current testnet and dev deployment explorers (check the links in the pr body description above). Since I want this to work out of the box with our explorers, I am using the option that is available in all explorers that we have deployed. Maybe we need to update the docker images to have that. Also adding sub 1-sat here will change the fee to use floats and that will be a much bigger change as well.

storopoli: NOTE to myself: verify if these fee rates can be changed to u32 or if we need u64.

crates/btcio/src/writer/fees.rs
storopoli: What? The URL? Yes it is already config defined.

storopoli:
> > This change also removes the previous * 2 multiplier
> 
> why? I thought we wanted to do so only when we have fee bump in place

Yeah you are right. I’ve made that change as a standalone commit. Will delete that commit from the codebase

crates/config/src/btcio.rs
barakshani: I have seen mempool_base_url, but if we write here mempool.space, we are basically indicating that we using mempool.space, and this doesn't have to be the case.

crates/btcio/src/writer/fees.rs
barakshani: kinda echoing the main theme in my review: what if we decide not to use mempool.space for bitcoin mainnet fee?

crates/btcio/src/writer/fees.rs
barakshani: got it, I indeed did not consider our own maintained version. I will open a ticket to upgrade our backend (mempool.sapce), but that's low prio.
Consider adding some comment here on why we don't use the _precise_ endpoint -- you don't know who will get here in the future and what changes they may decide to make.

> Also adding sub 1-sat here will change the fee to use floats and that will be a much bigger change as well.

understood, but it's not that we're doing any complex arithmetic, and we should already have pretty similar logic with L2 fee/gas.

crates/btcio/src/writer/fees.rs
barakshani: not the url, using `fastest_fee`.

crates/btcio/src/writer/fees.rs
barakshani: ticket opened: https://alpenlabs.atlassian.net/browse/STR-3038

crates/btcio/src/writer/fees.rs
```diff
@@ -0,0 +1,277 @@
...
+pub(crate) async fn resolve_fee_rate<R: Reader>(
+    client: &R,
+    config: &WriterConfig,
+) -> anyhow::Result<u64> {
+    match &config.fee_policy {
+        FeePolicy::BitcoinD => client
+            .estimate_smart_fee(1)
+            .await
+            .context("failed to estimate smart fee"),
+        FeePolicy::Mempool => resolve_mempool_fee_rate(client, config).await,
```
bewakes: nit: Can we rename this to `MempoolSpace` to make it clearer? Naming it just `Mempool` gives an impression that we are directly accessing bitcoin mempool. We are indirectly, but then if there's another provider that also uses mempool for estimation, then this collides.

crates/btcio/src/writer/fees.rs
bewakes: Agree, while we are at it we might as well make this configurable.

crates/btcio/src/writer/fees.rs
bewakes: Since this is in the code path for `FeePolicy::Mempool`, this should be fine for a particular fee provider. If we decide to add another provider, then we would likely have another `FeePolicy::Provider` variant. However, it's better to have the url configured as well, just in case we want fastest/econoimcal/etc fee rate.

bewakes (commented):
Nice work! Two things I wanted to flag:

1. **`Mempool` renaming** — Left a few inline nits. Nothing blocking, but I think the naming could be clearer.
2. **Fee policy abstraction** — One thought: a trait-based abstraction here could make this more flexible down the line. Not sure the complexity is justified yet though — would love your take on whether it's worth the indirection.

crates/btcio/src/writer/fees.rs
barakshani: Are we tailoring the code the mempool.space? why?

My understanding was that we want to add a _mempool_ policy (getting fee from mempool activity), and the provider can be any provider we choose.

crates/btcio/src/writer/fees.rs
barakshani: See above, why do we need to add more fee providers? We can have the provider (+any following url) in the config. I understood this was mostly to support our mainnet/testnet/staging.
Someone else (not eng team) will maintain the system in production. We should think about their easiness of life (and ours), if they ever want to change to a different provider.

> just in case we want fastest/econoimcal/etc fee rate.

`recommended` is the endpoint we query, then `fastest/econoimcal` are the enum-like result we get.

crates/btcio/src/writer/fees.rs
storopoli: I don't want to use any `Space` related since the repo is `mempool/mempool` (see https://github.com/mempool/mempool). What if we changed the naming to `MempoolExplorer` instead?

crates/btcio/src/writer/fees.rs
storopoli: > Since this is in the code path for FeePolicy::Mempool, this should be fine for a particular fee provider.

I get bibek's intentions. Any "mempool"-based fee estimation can be plugged in here. This would create a high level abstraction, which would be nice but I am urging for us to punt this high-level abstraction once we have the need and the design to have more than 1 mempool-based fee estimation provider.

crates/config/src/btcio.rs
storopoli: I've changed the docstring in 84a9a4266e6debb1ebaf65d943b2caa635061337, please check.

crates/btcio/src/writer/fees.rs
storopoli: Changed to `MempoolExplorer` in 84a9a4266e6debb1ebaf65d943b2caa635061337. Take a look.

crates/btcio/src/writer/fees.rs
storopoli: Added a TODO as well in 9379bdca0828b6433c738659061ef46cbde28563 with the ticket number.

crates/btcio/src/writer/fees.rs
storopoli: Made configurable in 86eb5d38af846b844dad9b7e9aa545f23d3ecac6 please take a look

barakshani (dismissed):

storopoli:
@barakshani you forgot the doubling that was removed and I've incorporated back in 4123fc51d75e2c5c151e48da4247791861b7ae93. Now it is ready to be reviewed. (I almost forgot as well).

barakshani (dismissed):

crates/btcio/src/writer/fees.rs
```diff
@@ -0,0 +1,329 @@
...
+    if !url.path().ends_with('/') {
+        let path = format!("{}/", url.path());
+        url.set_path(&path);
+    }
+
+    url.join("api/v1/fees/recommended")
+        .with_context(|| format!("invalid recommended-fees URL for base: {base_url}"))
+}
+
+async fn fetch_mempool_recommended_fees(url: Url) -> anyhow::Result<MempoolRecommendedFees> {
```
delbonis: This function should compute the url itself instead of assuming that it's passed the recommended fees url correctly.  It should just take the base url as an arg.

EXCEPT really we probably want to make an `EsploraClient` that has the base url and maybe also the HTTP client as fields and then we can just compute the url internally.

crates/btcio/src/writer/fees.rs
```diff
@@ -0,0 +1,329 @@
...
+    let url = mempool_recommended_fees_url(base_url)?;
+
+    match fetch_mempool_recommended_fees(url).await {
+        Ok(fees) => match mempool_fee_policy {
+            MempoolExplorerFeePolicy::Fastest => Ok(fees.fastest_fee),
+            MempoolExplorerFeePolicy::HalfHour => Ok(fees.half_hour_fee),
+            MempoolExplorerFeePolicy::Hour => Ok(fees.hour_fee),
+            MempoolExplorerFeePolicy::Economy => Ok(fees.economy_fee),
+            MempoolExplorerFeePolicy::Minimum => Ok(fees.minimum_fee),
+        },
```
delbonis: super minor nit but maybe make this an accessor on `MempoolExplorerFeePolicy`?

delbonis (changes requested):
generally looks good, minor comments

crates/btcio/src/writer/fees.rs
storopoli: Made an accessor in 569463065b7caa1300d17961cde9f83f898cc6c6, please check

crates/btcio/src/writer/fees.rs
storopoli: Refactored in 569463065b7caa1300d17961cde9f83f898cc6c6

delbonis (approved):

bewakes (approved):

irnb (approved):

PR #1638, opened 2026-04-10
STR-3002 IStateAccessor impl for OLState refactor

delbonis:
## Description

The primary purpose of this PR is to split the impl of IStateAccessor around OLState to instead be on a separate type instead of OLState directly.  This makes it mimic how the types will be structured in the future and gives allows for more space in the design for various bookkeeping improvements.  Some changes to improve ergonomics to the ledger types and write batch types were also made in the process, which helps us by cleaning up the abstractions.

This is partially in service of epoch sealing framework, which will also leverage this decoupling.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist


delbonis:
This is not complete, need to properly fix state root calculation and some other cleanups.

crates/ol/state-support-types/src/memory_state_layer.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    fn find_account_id_by_serial(&self, serial: AccountSerial) -> AcctResult<Option<AccountId>> {
+        Ok(self.serials.get(&serial).copied())
+    }
+
+    fn next_account_serial(&self) -> AccountSerial {
+        self.state.global.get_next_avail_serial()
+    }
+
+    fn compute_state_root(&self) -> AcctResult<Buf32> {
```
storopoli, line 157: Returning `Buf32::zero()` here means every OL state now has the same commitment. That breaks the state-root checks in block assembly/verification and in the checkpoint guest, because `MemoryStateBaseLayer` is now on the production path and both `compute_state_root` and `compute_state_root_with_writes` ignore the actual state. Permalink: https://github.com/alpenlabs/alpen/blob/0f82d6149c019d2d333a8bfa2d438e2072c3f79f/crates/ol/state-support-types/src/memory_state_layer.rs#L157-L178

crates/ol/state-types/src/toplevel.rs
```diff
@@ -25,7 +20,7 @@ impl OLState {
         let ledger = TsnlLedgerAccountsTable::from_genesis_account_params(&params.accounts)?;
         let total_ledger_funds = ledger.calculate_total_funds();
 
-        let global = GlobalState::new(params.header.slot);
+        let global = GlobalState::new(params.header.slot, AccountSerial::first_nonreserved());
```
storopoli, line 23: `from_genesis_account_params` already assigns serials to each genesis account, but `next_avail_serial` is still initialized to `AccountSerial::first_nonreserved()` here. With any non-empty genesis, the first post-genesis account will reuse an existing serial instead of allocating the next free one. Permalink: https://github.com/alpenlabs/alpen/blob/0f82d6149c019d2d333a8bfa2d438e2072c3f79f/crates/ol/state-types/src/toplevel.rs#L20-L24

crates/ol/state-support-types/src/batch_diff_layer.rs
```diff
@@ -70,7 +72,7 @@ impl<'batches, 'base, S: IStateAccessor> IStateAccessor for BatchDiffState<'batc
     fn cur_slot(&self) -> u64 {
```
storopoli, line 72: This layer is documented as overlaying a stack of pending batches, but the global/epochal reads only consult `self.batches.last()`. If an earlier pending batch changed `cur_slot`/`cur_epoch`/MMR fields and the newest batch only touched ledger state, those earlier writes are silently dropped and the read falls back to base. `compute_state_root_with_writes` below has the same issue because it ignores `self.batches` entirely. Permalink: https://github.com/alpenlabs/alpen/blob/0f82d6149c019d2d333a8bfa2d438e2072c3f79f/crates/ol/state-support-types/src/batch_diff_layer.rs#L72-L145

storopoli (changes requested):
This is a very good PR.

Really like the way that OL-STF code can now operate against a composable stack of layers: base state -> write tracking -> indexer/DA accumulation.

`OLState` itself is less coupled to transient concerns like serial lookup caches, pending writes, and indexer bookkeeping.

I see how this gives more room for future work like epoch sealing and other layered bookkeeping without baking those concerns directly into the core state type.

Some issues I've found with AI-assisted review (GPT-5.4 high reasoning effort). I know that the state-root calculation is not working for now, but it is worth to flag it anyways.

crates/ol/state-support-types/src/batch_diff_layer.rs
delbonis: ah shit that's a claude screwup, I changed how the write batches record changes to toplevel values and it didn't reflect it properly everywhere

delbonis:
Rebase this PR, now getting an unrelated Alloy-related build error, not sure what to do here.

crates/ol/state-support-types/src/memory_state_layer.rs
delbonis: fixed

crates/ol/state-support-types/src/batch_diff_layer.rs
delbonis: fixed

crates/ol/state-types/src/toplevel.rs
delbonis: fixed

delbonis:
Created some more tickets out of TODOs/FIXMEs I encountered:

* https://alpenlabs.atlassian.net/browse/STR-3227
* https://alpenlabs.atlassian.net/browse/STR-3228
* https://alpenlabs.atlassian.net/browse/STR-3229

delbonis:
Claude did a good job troubleshooting that test failure that was far removed from the changes, very interesting:

```
  1. Found the test datadir. The user pointed me at 28-10-ouaqn. I listed it and saw per-test subdirectories plus results.json.
  2. Confirmed the failure pattern. results.json showed many tests failing with process 'strata_sequencer' crashed — same root cause across tests, so I only needed to dig into one.
  3. Found the sequencer log. Each test dir had a strata_sequencer/service.log. I picked checkpoint/test_checkpoint_created and grepped the log for panic|ERROR|FATAL.
  4. Located the panic site. First hit:
  panicked at crates/chain-worker-new/src/state.rs:330:14:
  terminal block must have L1 height in write batch
  4. That's a panic message + file:line — basically the answer.
  5. Read the panicking code. handle_complete_epoch did epochal.last_l1_height.expect(...) on write_batch().epochal_writes(). The fields are Option<L1Height> / Option<L1BlockId>.
  6. Asked: when are those fields set? Grepped for last_l1_height = across crates/ol. Only one writer: WriteTrackingState::append_manifest in write_tracking_layer.rs. So the fields
  are populated only when an ASM manifest is appended.
  7. Asked: when does a terminal block append a manifest? Looked at crates/ol/stf/src/manifest_processing.rs::process_block_manifests. It loops over mf_cont.manifests() — if the
  container is empty, no append_manifest call, so the diff stays None. Terminal blocks with no new L1 manifests are normal in regtest.
  8. Confirmed it was a recent regression. git log on the branch + git log -S 'last_l1_height' -- write_batch.rs pointed at commit b7904c65b ("moved IStateAccessor impl..."). git
  show on that commit showed the WriteBatch going from full state copies to diff-only Option fields — but the chain-worker consumer wasn't updated to match the new semantics.
  9. Picked the fix location. The post-execution OLState already has the merged L1 info via epoch_state().last_l1_height() / last_l1_blkid(). try_exec_block was consuming new_state
  into persist_execution_output before calling handle_complete_epoch, so I reordered the calls and threaded &new_state through.

  The whole chain was: panic message → file:line → field semantics → producer (only one writer) → producer's caller (conditional) → blame the recent refactor that introduced the
  diff semantics.
````

**Comment by krsnapaudel** [comment]
@delbonis Block assembly PR (#1673) went in. Hope the rebase is not too bad.

**Comment by delbonis** [comment]
@krsnapaudel it was rather extensive, lots of tests broke

**Comment by delbonis** [comment]
Ok everything important passes now.

File: crates/ol/state-support-types/src/memory_state_layer.rs [hunk 4]
```diff
@@ -0,0 +1,184 @@
...
+        self.state.create_new_account(id, serial, new_acct_data)?;
+        self.serials.insert(serial, id);
+        Ok(serial)
+    }
+}
+
+impl IStateBatchApplicable for MemoryStateBaseLayer {
+    fn apply_write_batch(&mut self, batch: WriteBatch<Self::AccountState>) -> StateResult<()> {
+        self.state.apply_write_batch(batch)
+    }
```
storopoli, line 194: `MemoryStateBaseLayer` builds `serials` once in `new`, but `apply_write_batch` only mutates the inner `OLState`. Any batch-created account is then visible by ID but invisible through `find_account_id_by_serial`, so later manifest/DA handling that resolves deposits or predicate-key updates by serial can drop valid updates against the same post-state. Update `serials` while applying new accounts, or rebuild it after delegating to `OLState::apply_write_batch`.

crates/chain-worker-new/src/state.rs
```diff
@@ -319,11 +324,11 @@ impl ChainWorkerServiceState {
...
-        // Get L1 info from the write batch (epochal state has latest L1 after manifest sealing)
-        let epochal = last_block_output.write_batch().epochal();
-        let new_tip_height = epochal.last_l1_height();
-        let new_tip_blkid = epochal.last_l1_blkid();
-        let new_l1_block = L1BlockCommitment::new(new_tip_height, *new_tip_blkid);
+        // Read L1 info from the post-state. The write batch only stores diffs, so it
+        // may not contain a last_l1_* update if the terminal block had no new manifests.
+        let epoch_state = new_state.epoch_state();
+        let new_l1_block =
+            L1BlockCommitment::new(epoch_state.last_l1_height(), *epoch_state.last_l1_blkid());
```
storopoli: Add a unit test covering this case..

This is exactly the path that produced the runtime panic during this PR's testing (`terminal block must have L1 height in write batch`). The fix here — reading L1 info from the post-state `epoch_state` rather than from the diff-only `WriteBatch` — is correct, but the regression is currently only caught by functional tests.

A targeted unit test for `handle_complete_epoch` with a terminal block whose `WriteBatch` carries no `last_l1_*` update would lock this in cheaply and document the invariant. Right now `crates/chain-worker-new/src/state.rs` shows 0% patch coverage in codecov, and the `try_exec_block` → `handle_complete_epoch` chain is only exercised end-to-end.

Suggested shape:

- Build an `OLState` whose `epoch_state` has known `last_l1_height`/`last_l1_blkid`.
- Apply a terminal block whose `WriteBatch.epochal_writes` has `last_l1_height = None` and `last_l1_blkid = None` (i.e. no new manifest in this block).
- Assert that the resulting `EpochSummary.new_l1_block` matches the pre-existing `epoch_state` values rather than panicking or producing zeros.

Not blocking, but worth landing alongside this PR since the panic was real and the test is small.

crates/acct-types/src/errors.rs
```diff
@@ -89,7 +89,4 @@ pub enum AcctError {
 
     #[error("bitcoin amount overflow")]
     BitcoinAmountOverflow,
-
-    #[error("operation not supported in this context")]
-    Unsupported,
```
storopoli, line 94: YES!

crates/db/tests/src/ol_state_tests.rs
```diff
@@ -74,27 +79,27 @@ pub fn proptest_delete_toplevel_ol_state(
 pub fn proptest_put_and_get_write_batch(
     db: &impl OLStateDatabase,
     commitment: OLBlockCommitment,
-    state: OLState,
+    _state: OLState,
```
storopoli: do we need to keep the `_state`?

crates/db/tests/src/ol_state_tests.rs
```diff
@@ -74,27 +79,27 @@ pub fn proptest_delete_toplevel_ol_state(
...
+        retrieved_wb.global_writes().cur_slot,
+        wb.global_writes().cur_slot,
     );
 }
 
 pub fn proptest_delete_write_batch(
     db: &impl OLStateDatabase,
     commitment: OLBlockCommitment,
-    state: OLState,
+    _state: OLState,
```
storopoli: same here: do we need to keep the `_state` in this function arguments?

crates/ledger-types/src/errors.rs
```diff
@@ -0,0 +1,190 @@
...
+    InsufficientBalance(BitcoinAmount, BitcoinAmount),
+
+    #[error("out-of-order seqno change (cur {0:?}, new {1:?})")]
+    OooSeqnoChange(Seqno, Seqno),
+
+    #[error("inconsistent acct serial (id {0}, acct {1}, table {2})")]
+    AcctSerialInconsistent(AccountId, AccountSerial, AccountSerial),
+
+    #[error("inconsistent next serial ordering (cur {0}, new {1})")]
+    NextSerialSequence(AccountSerial, AccountSerial),
```
storopoli: Non-blocking but these would be much ergonomic to use as something like:

```suggestion
    #[error("mismatched account ID (got {got}, expected {expected}")]
    MismatchedAcctType {
       got: AccountTypeId,
       expected: AccountTypeId,
    }

    // and so on.
```

especially the one with 3 items in the tuple.

Also docstrings.

crates/ledger-types/src/errors.rs
```diff
@@ -0,0 +1,190 @@
...
+    TxConditionCheckFailed,
+
+    #[error("structural requirement check failed ({0})")]
+    TxStructureCheckFailed(&'static str),
+
+    #[error("transaction has expired (max slot {0}, cur slot {1})")]
+    TransactionExpired(Slot, Slot),
+
+    #[error("transaction is not mature (min slot {0}, cur slot {1})")]
+    TransactionNotMature(Slot, Slot),
```
storopoli, line 98: same as above.

crates/ledger-types/src/errors.rs
```diff
@@ -0,0 +1,190 @@
...
+    GenesisCoordsNonzero,
+
+    #[error("tried to skip epoch (parent {0}, current {1})")]
+    SkipEpochs(Epoch, Epoch),
+
+    #[error("tried to skip too many slots (parent {0}, current {1})")]
+    SkipTooManySlots(Slot, Slot),
+
+    #[error("incorrect epoch sequencing (parent {0}, parent terminal {2}, self {1})")]
+    IncorrectEpoch(Epoch, Epoch, bool),
```
storopoli, line 134: same as above

crates/ledger-types/src/errors.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    #[error("incorrect slot (expected {expected}, got {got})")]
+    IncorrectSlot { expected: u64, got: u64 },
```
storopoli, line 137: This one is nice!

crates/ledger-types/src/errors.rs
```diff
@@ -0,0 +1,190 @@
...
+    InconsistentBodyTerminality,
+
+    #[error("genesis block was not a terminal")]
+    GenesisNonterminal,
+
+    #[error("insufficient account balance (acct {0}, need {1})")]
+    InsufficientAccountBalance(AccountId, BitcoinAmount),
+
+    #[error("invalid sequence number for account {0} (expected {1}, actual {2})")]
+    InvalidSequenceNumber(AccountId, u64, u64),
```
storopoli: same as above

crates/ol/block-assembly/src/context.rs
```diff
@@ -166,6 +170,8 @@ where
             .get_state_for_tip_async(tip)
             .await
             .map_err(|e| BlockAssemblyError::StateProvider(Box::new(e)))
+            // keep current logic: stringified provider error
+            .map(|opt| opt.map(Arc::new))
```
storopoli, line 174: Should this be like this?
Should we have a tech-debt ticket for this?
Just curios

crates/ol/state-support-types/src/memory_state_layer.rs
```diff
@@ -0,0 +1,184 @@
...
+use strata_ol_state_types::{IStateBatchApplicable, OLAccountState, OLState, WriteBatch};
+
+use crate::write_tracking_layer::IComputeStateRootWithWrites;
+
+/// Base layer wrapping [`OLState`].
+#[derive(Clone, Debug)]
+pub struct MemoryStateBaseLayer {
+    state: OLState,
+    serials: BTreeMap<AccountSerial, AccountId>,
+}
```
storopoli, line 27: Non-blocking.
This struct is very important now. I'd like to add docstrings to the fields.

storopoli (changes requested):
Mostly non-blocking except for any batch-created account being invisible through `find_account_id_by_serial`.

crates/ol/state-support-types/src/memory_state_layer.rs
delbonis: Fixed, added test.

crates/ol/state-support-types/src/memory_state_layer.rs
delbonis: Added docstrings.

crates/ledger-types/src/errors.rs
delbonis: converted

crates/ol/block-assembly/src/context.rs
delbonis: I'm not really sure the best approach here, I'd like to rewrite a lot more of this later on.

crates/chain-worker-new/src/state.rs
delbonis: Added test, slightly more limited.

delbonis:
Ok everything should work now!

storopoli (approved):
ACK 4f96e64aba27b7fd82ae99ffd9a2b1ef31580e5c

storopoli:
@alpenlabs/chain-core take a look please, this is ready to merge.

PR #1650, opened 2026-04-14
feat(config): add v1 L2 fee-model config and reusable fee-quote types (STR-2977)

storopoli:
## Description

Introduces the sequencer fee-model configuration and shared internal types needed by the Alpen v1 fee model, prior to wiring up RPC behavior or execution-time charging.

### What changed

- **`L1FeePolicyConfig`** — extracted as a reusable struct wrapping `FeePolicy` + `mempool_base_url` + `mempool_fallback_conf_target`, shared between btcio writer config
  and EE config.
- **`SequencerFeeModelConfig`** — new TOML-deserializable config with `prover_fee_per_gas_wei`, `da_overhead_multiplier_bps`, `ol_overhead_wei`, and `l1_fee_rate_source`.
 All three fee params are mandatory; `l1_fee_rate_source` defaults to `btcio_writer`.
- **`FeeModelConfig` / `FeeQuoteInputs` / `FeeBreakdown` / `GasEquivalentQuote`** — runtime types in `alpen-ee-common` implementing the v1 formula with checked U256
arithmetic.
- **`AlpenEeConfig`** — gains `with_fee_model_config()` to carry the resolved fee model + L1 fee policy into the EE layer.
- Docker `sequencer.toml` and `init-keys.sh` updated with default `[fee_model]` section.

### v1 formula implemented

```py
execution_fee    = (base_fee + priority_fee) × gas_used
prover_fee       = prover_fee_per_gas × gas_used
da_fee           = ⌈ l1_fee_rate × diff_size × da_overhead_multiplier_bps / 10_000 ⌉

non_execution_fee = prover_fee + da_fee + ol_overhead

total_fee        = execution_fee + non_execution_fee
```

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- No RPC behavior changes — this is config + types only.
- `da_overhead_multiplier_bps` uses basis points (10 000 = 1×) to allow both undercharging and overcharging, as the design doc requires.
- `FeeQuoteInputs::l1_fee_rate_wei_per_byte` will need a sat/vB → wei/byte conversion when the fee-quote RPC is wired up (not in scope here).
- `SequencerFeeModelConfig` stores values as `u64`, converted to `U256` at resolution time in `AlpenEeConfig::fee_model()`. This caps configurable values at ~18.4 ETH in wei, which is practically fine for v1.
- `da_fee` uses ceiling division when applying `da_overhead_multiplier_bps` so the
  operator never silently undercharges from integer truncation. `GasEquivalentQuote`
  uses the same rounding direction when converting non-execution wei → gas units.
- Public types (`FeeModelConfig`, `FeeQuoteInputs`, `FeeBreakdown`, `GasEquivalentQuote`,
  `L1FeeRateSource`) derive `Serialize`/`Deserialize` so STR-2979 can reuse
  `FeeBreakdown` directly as the `eth_estimateDiffSize` response body without a
  parallel wire type.

- [x]  Yes
- [ ]  No

- [Alpen L2 Fee Model (Notion)](https://www.notion.so/Alpen-L2-Fee-Model-333901ba000f80d899b7c628cf0469d9)
- [STR-2977 (Jira)](https://alpenlabs.atlassian.net/browse/STR-2977)

## Checklist

**AI disclosure:** Initial implementation was done mostly manually with GPT-5.4 assistance. Then, Opus 4.6 was used to review this branch against the Jira ticket and design doc, implement review fixes, and assist with rebasing.

## Related Issues

STR-2977

### DIFF
--- a/crates/alpen-ee/common/src/types/fees.rs
+++ b/crates/alpen-ee/common/src/types/fees.rs
@@ -1,0 +1,383 @@
+//! Shared fee-model types and helpers for Alpen v1 fee quoting.
+//!
+//! This module defines the internal representation of fee-model inputs
+//! ([`FeeQuoteInputs`]), the configurable parameters that drive the formula
+//! ([`FeeModelConfig`]), the computed fee breakdown ([`FeeBreakdown`]), and
+//! a budgeting-only gas-equivalent view ([`GasEquivalentQuote`]). It
+//! intentionally does not perform any RPC handling or execution-time
+//! charging; those layers consume these types so the v1 formula lives in a
+//! single place.
+
+use alloy_primitives::U256;
+use serde::{Deserialize, Serialize};
+
+/// Basis-points scaling factor for DA fee multipliers.
+pub const DA_OVERHEAD_MULTIPLIER_SCALE_BPS: u32 = 10_000;
+
+/// Source for the L1 fee rate used by the fee model.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum L1FeeRateSource {
+    /// Reuse the btcio writer policy used for publication.
+    BtcioWriter,
+}
+
+/// Runtime configuration for the v1 fee model.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct FeeModelConfig {
+    /// Static proving fee charged per unit of raw EVM gas.
+    pub prover_fee_per_gas_wei: U256,
+
+    /// Basis-points multiplier applied to estimated DA cost.
+    pub da_overhead_multiplier_bps: u32,
+
+    /// Small additive fee charged for OL and infrastructure overhead.
+    pub ol_overhead_wei: U256,
+
+    /// Source used to resolve the L1 fee rate.
+    pub l1_fee_rate_source: L1FeeRateSource,
+}
+
+/// Inputs required to quote fees for a transaction.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct FeeQuoteInputs {
+    /// Estimated raw EVM gas for the transaction.
+    pub raw_evm_gas: u64,
+
+    /// Current base fee per gas in wei.
+    pub base_fee_per_gas: U256,
+
+    /// Current priority fee per gas in wei.
+    pub priority_fee_per_gas: U256,
+
+    /// Resolved L1 fee rate in wei per byte of DA payload.
+    pub l1_fee_rate_wei_per_byte: U256,
+
+    /// Estimated diff size in bytes for the transaction.
+    pub diff_size_bytes: u64,
+}
+
+impl FeeQuoteInputs {
+    /// Returns the total execution gas price in wei.
+    pub fn execution_gas_price(&self) -> Result<U256, FeeModelError> {
+        fee_add(
+            self.base_fee_per_gas,
+            self.priority_fee_per_gas,
+            "execution_gas_price",
+        )
+    }
+}
+
+/// Quoted fee breakdown for a single transaction.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct FeeBreakdown {
+    /// Raw EVM gas used for execution and block accounting.
+    pub raw_evm_gas: u64,
+
+    /// Total execution gas price in wei.
+    pub execution_gas_price: U256,
+
+    /// EVM execution fee in wei.
+    pub execution_fee: U256,
+
+    /// Proving fee in wei.
+    pub prover_fee: U256,
+
+    /// DA fee in wei.
+    pub da_fee: U256,
+
+    /// OL overhead fee in wei.
+    pub ol_overhead_fee: U256,
+
+    /// Sum of all non-execution fees in wei.
+    pub non_execution_fee: U256,
+
+    /// Total quoted fee in wei.
+    pub total_fee: U256,
+}
+
+impl FeeBreakdown {
+    /// Converts fee amounts into a [`GasEquivalentQuote`] for budgeting.
+    ///
+    /// Uses ceiling division when converting non-execution wei into gas
+    /// units so the budgeting view never undercharges relative to the
+    /// quoted [`FeeBreakdown::total_fee`].
+    pub fn gas_equivalent_quote(&self) -> Result<GasEquivalentQuote, FeeModelError> {
+        let non_execution_gas = if self.non_execution_fee.is_zero() {
+            0
+        } else if self.execution_gas_price.is_zero() {
+            return Err(FeeModelError::ZeroExecutionGasPrice);
+        } else {
+            let gas = self.non_execution_fee.div_ceil(self.execution_gas_price);
+            u64::try_from(gas)
+                .map_err(|_| FeeModelError::GasEquivalentOverflow("non_execution_gas"))?
+        };
+
+        let total_gas = self
+            .raw_evm_gas
+            .checked_add(non_execution_gas)
+            .ok_or(FeeModelError::GasEquivalentOverflow("total_gas"))?;
+
+        Ok(GasEquivalentQuote {
+            execution_gas: self.raw_evm_gas,
+            non_execution_gas,
+            total_gas,
+        })
+    }
+}
+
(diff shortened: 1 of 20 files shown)


crates/btcio/src/writer/fees.rs
delbonis, line 1: In the tests here where we construct these fee-related type instances it's quite a bit more verbose than it needs to be.  Maybe we could add more constructors / a builder to these types?

delbonis (changes requested):
Generally looks good.  Not a fan of making all these fields pub but maybe it's fine.

storopoli:
> Generally looks good. Not a fan of making all these fields pub but maybe it's fine.

Crap, forgot about that. I'll make them `pub(crate)` with accessors.

Ok check out 8143deb3ea7236aae32c5ccf553aaeb92ddc96ab for the accessors.

crates/btcio/src/writer/fees.rs
storopoli: Check 46e7582a86603b15fde6d0f03c5c0afbaff11ec5 done

purusang (approved):
Ack docker

delbonis (approved):

crates/alpen-ee/common/src/types/fees.rs
```diff
@@ -0,0 +1,477 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct GasEquivalentQuote {
+    /// Raw EVM execution gas.
+    pub(crate) execution_gas: u64,
+
+    /// Additional budgeting gas equivalent for non-execution fees.
+    pub(crate) non_execution_gas: u64,
+
+    /// Total budgeting gas.
+    pub(crate) total_gas: u64,
```
bewakes: Is`total_gas` not equal to `execution_gas + non_execution_gas`? Because that would make this field redundant - we can just have a function `total_gas()`.

crates/config/src/btcio.rs
```diff
@@ -20,21 +20,87 @@ pub struct ReaderConfig {
...
+impl WriterConfig {
+    /// Returns the configured L1 fee-policy configuration.
+    pub fn l1_fee_policy(&self) -> &L1FeePolicyConfig {
+        &self.l1_fee_policy
+    }
+
+    /// Returns the configured L1 fee policy.
+    pub fn fee_policy(&self) -> &FeePolicy {
+        self.l1_fee_policy.fee_policy()
+    }
```
bewakes: nit: the name is weird. The outer container and the inner value have technically the same name: "fee policy". You can rename `l1_fee_policy -> (l1_)fee_policy_config` and the inner can remain the same.

crates/config/src/btcio.rs
```diff
@@ -20,21 +20,87 @@ pub struct ReaderConfig {
...
+
+/// Reusable configuration for resolving Bitcoin fee rates.
+#[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Eq)]
+pub struct L1FeePolicyConfig {
+    /// How fees are determined while creating L1 transactions.
+    #[serde(flatten)]
+    pub(crate) fee_policy: FeePolicy,
     /// Base URL for mempool.space-compatible fee API.
-    pub mempool_base_url: Option<String>,
+    pub(crate) mempool_base_url: Option<String>,
```
bewakes: `mempool_base_url` should be inside `MempoolExplorer` variant.      
                                                                                                                            
Same goes for `mempool_fallback_conf_target` which should be inside `BitcoinD` variant. But, this is being used as a fallback for `MempoolExplorer` so that can't be done.                                                                                                                                                   
                                                                                                                                                                                                                                               It doesn't make sense to have some value for `mempool_base_url` when the fee policy is say `Fixed` or `BitcoinD`. The values will be ignored, but we don't want to be able to create a config whose values don't make sense. Let's solve this from the type level.

Also, it looks like `BitcoinD` config is always required because it is a fallback. So we have two configurations representing `conf_target` which is again weird - maybe keep `conf_target` inside `L1FeePolicyConfig` and remove `conf_target` from bitcoind?

bewakes (changes requested):
Apologies for requesting changes late but there's a concern over the `L1FeePolicyConfig` structure - I think some of the attributes need to be inside the `FeePolicy` variants.

crates/alpen-ee/common/src/types/fees.rs
storopoli: Ok done in 08d5c9ce0e53fb22df3686514f0c4a21c452d0bb please check

crates/config/src/btcio.rs
storopoli: renamed in 08d5c9ce0e53fb22df3686514f0c4a21c452d0bb

crates/config/src/btcio.rs
storopoli: fixed in 08d5c9ce0e53fb22df3686514f0c4a21c452d0bb

bewakes (approved):

crates/alpen-ee/common/src/types/fees.rs
```diff
@@ -0,0 +1,477 @@
...
+
+    /// Source used to resolve the L1 fee rate.
+    pub(crate) l1_fee_rate_source: L1FeeRateSource,
+}
+
+/// Inputs required to quote fees for a transaction.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct FeeQuoteInputs {
+    /// Estimated raw EVM gas for the transaction.
+    pub(crate) raw_evm_gas: u64,
```
prajwolrg, line 47: Why is this non-U56? like other values?

prajwolrg (changes requested):
Mostly looks good. I'm curious on the usage of `u64` vs `U256` in a few places. For revm-place it makes sense to use U256 because that's what EVM uses. I think it might be worth it to add rationale as a comment in a palce, maybe where the config is first defined.

crates/alpen-ee/common/src/types/fees.rs
storopoli: Good callout. The split is intentional: `u64` is used for bounded quantities/counts, while `U256` is used for wei-denominated amounts and arithmetic that can overflow 64 bits.

In `FeeQuoteInputs`, `raw_evm_gas` and `diff_size_bytes` are quantities, while `base_fee_per_gas`, `priority_fee_per_gas`, and `l1_fee_rate_wei_per_byte` are monetary values:
https://github.com/alpenlabs/alpen/blob/08d5c9ce0e53fb22df3686514f0c4a21c452d0bb/crates/alpen-ee/common/src/types/fees.rs#L40-L56

That matches the Reth/EVM side where gas limits/usage are `u64`, then converted to `U256` only when calculating fees:
https://github.com/alpenlabs/alpen/blob/08d5c9ce0e53fb22df3686514f0c4a21c452d0bb/crates/reth/node/src/payload_builder.rs#L211-L217
https://github.com/alpenlabs/alpen/blob/08d5c9ce0e53fb22df3686514f0c4a21c452d0bb/crates/reth/node/src/payload_builder.rs#L280-L285

The fee model follows the same boundary: convert gas/byte counts into `U256` at the point of fee arithmetic:
https://github.com/alpenlabs/alpen/blob/08d5c9ce0e53fb22df3686514f0c4a21c452d0bb/crates/alpen-ee/common/src/types/fees.rs#L273-L291

For config, the TOML-facing values stay as small operator-provided scalars, then get widened into the runtime `FeeModelConfig`:
https://github.com/alpenlabs/alpen/blob/08d5c9ce0e53fb22df3686514f0c4a21c452d0bb/crates/config/src/config.rs#L172-L190
https://github.com/alpenlabs/alpen/blob/08d5c9ce0e53fb22df3686514f0c4a21c452d0bb/crates/alpen-ee/config/src/config.rs#L91-L112

I agree this deserves a short rationale comment near the fee-model types.

storopoli: Check d0975aefa4f8328634092ac5e40a8fe96950e916

storopoli: @prajwolrg gently ping

prajwolrg (approved):

PR #1656, opened 2026-04-16
(datatool): Generate alpen predicate key in ol params; accept balance…

bewakes:
## Description
Resolves https://alpenlabs.atlassian.net/browse/STR-2367

- Replace hardcoded `AlwaysAccept` predicate for the Alpen EE snark account with configurable SP1 Groth16 predicate                                                 
- New CLI flags for gen-ol-params:                    
    -  `--alpen-chain-spec`: Used to generate alpen account state root to be inserted to ol params. defaults to dev chain spec                                                               
    - `--alpen-predicate`: always-accept or sp1-groth16 (defaults to feature-gated, same pattern as
  `--checkpoint-predicate`)                                                                                              
    - `--alpen-balance`: initial balance in sats (default 0)
    - `--alpen-inner-state`: inner state root as 64-char hex (default zero)                                              
- Requires `-F sp1-builder` during build to resolve the real VK from `GUEST_ALPEN_ACCT_VK_HASH_STR`

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This PR was created with the help of Claude Code.

- [ ]  Yes
- [x]  No

## Checklist


bin/datatool/src/args.rs
```diff
@@ -306,6 +309,24 @@ pub(crate) struct SubcOlParams {
         description = "path to JSON-serialized genesis L1 view (required when btc-client feature is disabled)"
     )]
     pub(crate) genesis_l1_view_file: Option<String>,
+
+    #[argh(
+        option,
+        description = "alpen snark account predicate type: 'always-accept' or 'sp1-groth16' (default: feature-gated)"
+    )]
+    pub(crate) alpen_predicate: Option<AcctPredicateOverride>,
```
voidash, line 317: would it be better if it's `--ee-account-predicate`? it would match how we have been naming like `gen-ol-params`

bin/datatool/src/cmd/ol_params.rs
```diff
@@ -25,13 +27,20 @@ pub(super) fn exec(cmd: SubcOlParams, ctx: &mut CmdContext) -> anyhow::Result<()
...
-    // TODO: handle EE accounts properly https://alpenlabs.atlassian.net/browse/STR-2367
     let mut ol_params = OLParams::new_empty(genesis_l1_view.blk);
 
+    let acct_predicate = resolve_acct_predicate(cmd.alpen_predicate)?;
+    let balance = BitcoinAmount::from_sat(cmd.alpen_balance.unwrap_or(0));
+    let inner_state = match cmd.alpen_inner_state {
+        Some(hex) => hex
+            .parse::<Buf32>()
+            .map_err(|e| anyhow::anyhow!("invalid alpen-inner-state hex: {e}"))?,
+        None => Buf32::zero(),
```
voidash: `Buf32::zero()` if we don't pass `alpen_inner_state` would still have the bug. 
Even for fresh network this is wrong. We should show error if they don't provide inner state. But what is better is we should have another flag `--chain-spec` . you provide that file and it computes you the inner state.

voidash: that way if we need to migrate we can use `alpen-inner-state` and if fresh `chainspec`

voidash (changes requested):

bin/datatool/src/cmd/ol_params.rs
bewakes: True. It is included in https://github.com/alpenlabs/alpen/pull/1656/commits/f0b597f06804a3b753b21a1d55a9efcfc000a3e9

bin/datatool/src/args.rs
bewakes: I see your concern. However, I am refraining from using the term `ee` because it is generic. Here, we are providing config for Alpen EE so wanted to be more specific.
In case of Strata `ol` makes sense because there's only one OL. At least the way I see it.

prajwolrg (dismissed):
LGTM!

One thing we should do as a followup, is also replace `PredicateKey::always_accept()` with `PredicateKey::Bip340Schnorr`, once we have fully integrated the proof generation. The native proof generation is compatible with `Bip340Schnorr` and that'd make testing more meaningful.

voidash (dismissed):

bin/datatool/src/cmd/ol_params.rs
```diff
@@ -18,20 +24,33 @@ const ALPEN_EE_ACCOUNT_ID: AccountId = AccountId::new([1u8; 32]);
...
         cmd.genesis_l1_view_file.as_deref(),
         cmd.genesis_l1_height,
         ctx,
     )?;
 
-    // TODO: handle EE accounts properly https://alpenlabs.atlassian.net/browse/STR-2367
     let mut ol_params = OLParams::new_empty(genesis_l1_view.blk);
 
+    let acct_predicate = resolve_acct_predicate(cmd.alpen_predicate)?;
+    let balance = BitcoinAmount::from_sat(cmd.alpen_balance.unwrap_or(0));
```
storopoli, line 43: Allowing a nonzero `--alpen-balance` here makes the generated OL genesis params internally inconsistent with the EE genesis state. `EeAccountState::new` commits `tracked_balance` into the inner-state root:
https://github.com/alpenlabs/alpen/blob/0e2de5e634551df83bce9bef080adf3cfdc03f2a/crates/ee-acct-types/src/state.rs#L11-L26

https://github.com/alpenlabs/alpen/blob/0e2de5e634551df83bce9bef080adf3cfdc03f2a/crates/ee-acct-types/src/state.rs#L116-L120

but EE genesis is still always created with zero tracked balance 

https://github.com/alpenlabs/alpen/blob/0e2de5e634551df83bce9bef080adf3cfdc03f2a/crates/alpen-ee/genesis/src/utils.rs#L8-L14

Update verification also checks that tracked balance matches the verified balance bookkeeping

https://github.com/alpenlabs/alpen/blob/0e2de5e634551df83bce9bef080adf3cfdc03f2a/crates/ee-acct-runtime/src/ee_program.rs#L147-L150

So any nonzero value here will produce a bad genesis unless the EE-side genesis path is updated too.

bin/datatool/src/args.rs
```diff
@@ -306,6 +309,30 @@ pub(crate) struct SubcOlParams {
...
+
+    #[argh(
+        option,
+        description = "alpen EE account initial balance in sats (default 0)"
+    )]
+    pub(crate) alpen_balance: Option<u64>,
+
+    #[argh(
+        option,
+        description = "alpen EE account inner state root as 64-char hex; overrides --alpen-chain-config if both are provided (default zero)"
```
storopoli: This help text is stale now. It says the default is zero, but the implementation falls back to deriving the root from the dev chain spec when neither flag is provided 

https://github.com/alpenlabs/alpen/blob/f0b597f06804a3b753b21a1d55a9efcfc000a3e9/bin/datatool/src/cmd/ol_params.rs#L73-L80

bin/datatool/src/acct_predicate.rs
```diff
@@ -0,0 +1,91 @@
...
+fn resolve_sp1_groth16() -> anyhow::Result<PredicateKey> {
+    #[cfg(feature = "sp1-builder")]
+    {
+        Ok(build_sp1_predicate())
+    }
+
+    #[cfg(not(feature = "sp1-builder"))]
+    {
+        anyhow::bail!(
+            "--acct-predicate sp1-groth16 requires the binary to be built with -F sp1-builder"
```
storopoli: The error message points users to `--acct-predicate`, but the flag introduced by this PR is `--alpen-predicate`

https://github.com/alpenlabs/alpen/blob/f0b597f06804a3b753b21a1d55a9efcfc000a3e9/bin/datatool/src/args.rs#L313-L317

storopoli (changes requested):

bin/datatool/src/cmd/ol_params.rs
bewakes: This should be fine because this is just a tool whose goal is to make things configurable and it doesn't have to take into consideration what's there in the other system. It's the responsibility of the one who generates the params to make sure proper values are passed.

bin/datatool/src/args.rs
bewakes: Oh yes. Done in https://github.com/alpenlabs/alpen/pull/1656/commits/56cd3fabb20d00d92c69c23e6c3a84960c13ac16

bin/datatool/src/acct_predicate.rs
bewakes: Fixed in https://github.com/alpenlabs/alpen/pull/1656/commits/56cd3fabb20d00d92c69c23e6c3a84960c13ac16

storopoli (approved):
ACK 56cd3fabb20d00d92c69c23e6c3a84960c13ac16

bin/datatool/src/args.rs
storopoli: @voidash ping on this. We need to resolve this conversation for the PR to be in a mergeable state. Let me know how you want to proceed.

prajwolrg (approved):

PR #1658, opened 2026-04-16
feat(alpen-client): serve alpen_getBlockStatus on fullnodes (STR-3076)

storopoli:
## Description

On Alpen fullnodes, `alpen_getBlockStatus` was unavailable because the handler depended on sequencer-only state (`EeNodeStorage`/`BatchStorage`) and its registration was gated behind`#[cfg(feature = "sequencer")]` + `if ext.sequencer`.

The handler is rewritten to use two pieces of state that exist on every
Alpen node:

- Reth's canonical-chain lookup (`BlockchainProvider::block_number`) to resolve the target hash, the confirmed head, and the finalized head to canonical block numbers.
- The `OLTracker`-derived `ConsensusHeads` watcher (already injected into `EeRpcServer`) for the current confirmed / finalized EE block hashes.

Block status is derived by height comparison: target <= finalized -> `Finalized`; target <= confirmed -> `Confirmed`; else `Pending`.
Block `0` is special-cased to `Finalized` to preserve existing semantics. Non-canonical or unknown hashes return `-32602 block not found`, matching the previous error shape.

Registration is lifted to apply unconditionally, so the method is now served on both sequencer and fullnode.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Three independent commits for easier review:

1. `refactor(alpen-ee/rpc-server)` — rewrite the handler using `BlockchainProvider<N>` + `ConsensusHeads`; delete `BlockBatchLookup`, `FrontierResolveError`, and the `find_batch_*` / `resolve_frontier_batch_idx` helpers; swap `storage` for `ctx.provider().clone()` at the call site. Registration gate is preserved so sequencer behavior is unchanged.
2. `feat(alpen-client)` — lift the `#[cfg(feature = "sequencer")]` +  `if ext.sequencer` guard so the RPC is registered on every Alpen node.
3. `test(functional-tests)` — replace the "expect -32601 method-not-found on fullnode" assertion with positive parity checks; assert the fullnode converges to the sequencer's status for every block at each progression milestone, with a short poll-with-timeout to absorb `OLTracker` lag.

Reorg semantics are tightened: non-canonical known hashes now uniformly return `-32602` rather than relying on whether sequencer storage had seen the block.

This PR was written primarily by Claude Code (Opus 4.7) with human review and guidance on the design approach (the solution came from the ticket comments by myself and @sapinb).

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-3076

### DIFF
--- a/crates/alpen-ee/rpc/server/src/block_status.rs
+++ b/crates/alpen-ee/rpc/server/src/block_status.rs
@@ -1,232 +1,116 @@
 //! Alpen EE RPC handler implementation for block-status methods.
 
-use std::{error::Error, fmt, num::NonZeroU64, sync::Arc};
-
 use alloy_primitives::B256;
-use alpen_ee_common::{BatchStorage, ConsensusHeads, ExecBlockStorage, StorageError};
+use alpen_ee_common::ConsensusHeads;
 use alpen_ee_rpc_api::{AlpenEeRpcServer, BlockStatus, BlockStatusResponse};
 use async_trait::async_trait;
 use jsonrpsee::core::RpcResult;
-use strata_acct_types::Hash;
+use reth_node_builder::NodeTypesWithDB;
+use reth_provider::{
+    providers::{BlockchainProvider, ProviderNodeTypes},
+    BlockHashReader, BlockNumReader, ProviderResult,
+};
 use tokio::sync::watch;
 
-use crate::errors::{block_not_found_error, frontier_unavailable_error, internal_error};
+use crate::errors::{block_not_found_error, internal_error};
 
-/// RPC handler for [`AlpenEeRpcServer`].
-#[derive(Debug)]
-pub struct EeRpcServer<S> {
-    storage: Arc<S>,
-    consensus_rx: watch::Receiver<ConsensusHeads>,
-}
-
-/// Batch lookup result for an execution block hash.
-enum BlockBatchLookup {
-    /// Block hash does not exist in execution storage.
-    BlockNotFound,
-
-    /// Block exists but no batch covers it yet.
-    NoBatchCoverage,
-
-    /// Block is covered by the genesis marker batch (idx 0).
-    CoveredGenesis,
-
-    /// Block is covered by a non-genesis batch index.
-    Covered(NonZeroU64),
-}
-
-/// Errors while resolving a consensus frontier hash to batch coverage.
-#[derive(Debug)]
-enum FrontierResolveError {
-    /// Frontier block exists but no batch covers it yet.
-    NoBatchCoverage(Hash),
-
-    /// Frontier block hash does not exist in execution storage.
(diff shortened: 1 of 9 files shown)


krsnapaudel:
@storopoli So you are also doing [this](https://alpenlabs.atlassian.net/browse/STR-2937) here, right?

storopoli:
Yes but unintentional

crates/alpen-ee/rpc/server/src/block_status.rs
```diff
@@ -1,232 +1,116 @@
...
-            .map_err(frontier_unavailable_error)?;
+        let heads = self.consensus_rx.borrow().clone();
+
+        // Finalized check: skip when the head is unset or not canonical on
+        // this node (transient during sync / reorg — OLTracker may still be
+        // tracking a fork that Reth hasn't reorged to).
+        let finalized_b256 = B256::from_slice(heads.finalized().as_slice());
+        if !finalized_b256.is_zero() {
+            match canonical_block_number(&self.provider, finalized_b256) {
+                Ok(Some(fin_num)) if target_num <= fin_num => {
```
krsnapaudel, line 92: @storopoli Is this wise, just checking block number range? It's probably better for you to just address the ticket. Not the RPC part of it.

delbonis (approved):

crates/alpen-ee/rpc/server/src/block_status.rs
storopoli: The implementation here follows the solution proposed in the STR-3076 comments: derive block status from `OLTracker` confirmed/finalized EE hashes plus canonical Reth height lookups, and expose the RPC on fullnodes. It does not implement the broader ticket DoD of repopulating `ExecBlockRecords` on fullnodes via the EE verification pipeline. If we want to keep the original ticket scope, this should probably be split or tracked as a follow-up.

On a side note I've dived into that and the PR is becoming huge to fully fulfill that original DoD.

krsnapaudel: @storopoli My comment is about not relying only on block numbers. Seems your approach simpler than what I had. Populating `ExecBlockRecords` on fullnodes can be skipped if we have a way to serve RPC.

krsnapaudel (commented):
@storopoli The solution looks good. The namespace needs to be updated `alpen_getBlockStatus`.

bewakes (approved):

storopoli:
> @storopoli The solution looks good. The namespace needs to be updated `alpen_getBlockStatus`.

Check 78ad6f1486f4ce7fc49b29961f4285fbf18c17ea. Nice catch, totally forgot about that.

krsnapaudel (approved):

PR #1668, opened 2026-04-21
docker(strata): bake strata-datatool + SP1 guest ELFs into image

alexhui01:
## Summary

Ship \`strata-datatool\` and the SP1 guest ELF that the runtime prover reads
inside the strata image. One atomic artifact per alpen commit — no separate
ELF distribution, no init container, no VK drift between whoever runs the
build.

## Approach

Mirrors the existing prover-client CI pattern:

1. GHA runner installs the SP1 toolchain (\`cargo prove\`).
2. \`cargo build --bin strata-datatool --features sp1-docker-builder\` runs
   **on the runner itself**, not inside Docker. \`sp1-docker-builder\` still
   performs the guest compilation inside a pinned Succinct container, so
   ELFs are byte-reproducible.
3. The datatool binary and \`guest-checkpoint-new.elf\` are copied into
   \`docker/strata/artifacts/\` (inside the docker build context).
4. \`docker buildx build\` just \`COPY\`s them into the runtime stage.

I originally tried to do step 2 *inside* the Dockerfile with a
\`--mount=type=bind,source=/var/run/docker.sock\` RUN step, but that
requires the BuildKit builder to have access to the host's Docker socket
— not how \`docker/setup-buildx-action\` is configured by default. Local
testing reproduced the \`\"/var/run/docker.sock\": not found\` error. Option B
avoids the BuildKit gymnastics entirely.

### \`docker/strata/Dockerfile\`
- COPY \`docker/strata/artifacts/strata-datatool\` → \`/usr/local/bin/strata-datatool\`
- COPY \`docker/strata/artifacts/elfs\` → \`/app/elfs\`
- \`ENV ELF_BASE_PATH=/app/elfs/sp1\` so the prover finds ELFs with no extra config

### \`.github/workflows/ci-build.yml\`
- \`if matrix.program == 'strata'\`: install rust toolchain + SP1, cache cargo, run the datatool build, stage artifacts into the build context.
- Existing \`docker buildx build\` step unchanged.

### \`.gitignore\`
- Add \`docker/strata/artifacts/\` so nobody accidentally commits build output.

## Why only \`guest-checkpoint-new.elf\`?

strata reads only that ELF at runtime. The other four guest programs are
embedded inside the datatool binary (via \`strata-sp1-guest-builder\`'s
build script) and re-emitted by \`datatool genparams --elf-dir\` during
params generation. Saves ~11 MB on the runtime image.

## Downstream (deployments repo)

- \`generate-params.yml\` drops its Rust/SP1/cargo build steps, pulls
  \`strata:\$SHA\`, \`docker cp\`s the datatool out, runs it. Workflow
  runtime: ~40 min → ~1 min.
- strata-ol Helm chart drops the bespoke \`ELF_BASE_PATH\` override — the
  image default is correct.

## Costs

- Strata CI job grows by one cargo build (~10-15 min warm,
  ~30-40 min cold). Swatinem rust cache handles incremental. The Docker
  builds for strata and alpen-client still parallelize via the matrix.
- Strata runtime image +6 MB (5 MB datatool + 1 MB ELF).

## Testing

- [ ] Local \`docker buildx build\` against a manually-staged
      \`docker/strata/artifacts/\` directory (smoke test the Dockerfile).
- [ ] CI on this branch — verifies the workflow changes produce a valid
      image end-to-end.
- [ ] Pull the resulting image, run \`strata-datatool genparams
      --elf-dir /tmp/x\`, verify the VK in the output matches what the
      runtime's embedded ELF hashes to.

## Related

- Supersedes the in-Dockerfile DinD approach discussed in
  \`runbooks/proposal-bake-datatool-elfs-into-strata-image-2026-04-20.md\`
  (will update to reflect this PR).
- Closes the "where do ELFs come from" gap that has been blocking dev OL
  since the prover-enabled strata image was rolled out.


.github/workflows/ci-build.yml
```diff
@@ -120,6 +120,58 @@ jobs:
...
+
+      - name: Build strata-datatool + SP1 guest ELFs
+        if: matrix.program == 'strata'
+        run: |
+          # sp1-builder uses the host rustc; matches the existing
+          # prover-client CI pattern in cd.yml. We'd prefer sp1-docker-builder
+          # for byte-reproducible ELFs, but the pinned Succinct toolchain
+          # container currently ships rustc 1.88 while alpen workspace deps
+          # require rustc >=1.91. Revisit once Succinct updates.
+          cargo build --bin strata-datatool --features sp1-builder --release
```
MdTeach: This could be addressed in a follow-up PR, but sp1-docker-builder is essential for production deployments. It should be fine for internal deployments as-is. 

Additionally, we’re not blocked by the SP1 toolchain, since it already supports `1.95.0-nightly (ce69df6f7, 2026-02-12)`.

.github/workflows/ci-build.yml
```diff
@@ -120,6 +120,58 @@ jobs:
...
+
+          mkdir -p docker/strata/artifacts/elfs/sp1
+          install -m 0755 target/release/strata-datatool docker/strata/artifacts/strata-datatool
+
+          # strata only reads guest-checkpoint-new.elf at runtime. The other
+          # guest ELFs are embedded inside the datatool binary (via
+          # strata-sp1-guest-builder's build script) and re-emitted by
```
purusang, line 155: I have not looked into it, but is this correct @MdTeach ?

purusang (changes requested):
Its good for ci-cd. But I feel like it will complicate local docker builds and runs for developers since we have added some steps in Dockerfiles that depends on work done in ci. For eg. for developer to run strata they will have to manually replicate the CI steps:                                                        
1. Install SP1 toolchain
2. cargo build --bin strata-datatool --features sp1-builder --release                                                                  
3. mkdir -p docker/strata/artifacts/elfs/sp1                         
4. install -m 0755 target/release/strata-datatool docker/strata/artifacts/strata-datatool                                              
5. cp provers/sp1/guest-checkpoint-new/cache/guest-checkpoint-new.elf docker/strata/artifacts/elfs/sp1/
6. Then docker build

.github/workflows/ci-build.yml
alexhui01: Got it. Let me remove this comment for now.

.github/workflows/ci-build.yml
alexhui01: @purusang Should we use sp1-docker-build if this is essential for production deployment?

alexhui01: And speaking of your concerns, how developers cover this before to have elf files with prover client?

purusang: I was only asking if that comment was correct.
> embedded inside the datatool binary

docker/strata/Dockerfile
```diff
@@ -73,9 +74,15 @@ RUN mkdir -p /app/data /app/.cache /app/.local/share && \
     chown -R alpen:alpen /app
 
 COPY --from=builder /usr/local/bin/strata /usr/local/bin/strata
+# strata-datatool + the SP1 guest ELF are built on the CI runner with
+# --features sp1-docker-builder (matches the prover-client pattern in
+# cd.yml). At docker build time they live in the build context under
+# docker/strata/artifacts/. Regeneration steps live in ci-build.yml.
+COPY docker/strata/artifacts/strata-datatool /usr/local/bin/strata-datatool
+COPY docker/strata/artifacts/elfs /app/elfs
```
purusang, line 82: My concern regarding being dependent on CI was about these steps.

.github/workflows/ci-build.yml
purusang: >Should we use sp1-docker-build 

I guess that's what @MdTeach had been suggesting. He can confirm on this.

delbonis (approved):
> `/usr/local/bin`

Much applause for using unix paths correctly even in the container image.

storopoli (approved):
ACK ca4e7c8e3fcad8c5d7ead410f307465c93b8d07a

Once the comments raised here are addressed we can merge.

.github/workflows/ci-build.yml
```diff
@@ -120,6 +120,48 @@ jobs:
...
+        if: matrix.program == 'strata'
+        run: |
+          cargo build --bin strata-datatool --features sp1-builder --release
+
+          mkdir -p docker/strata/artifacts/elfs/sp1
+          install -m 0755 target/release/strata-datatool docker/strata/artifacts/strata-datatool
+
+          # strata only reads guest-checkpoint-new.elf at runtime. The other
+          # guest ELFs are embedded inside the datatool binary (via
+          # strata-sp1-guest-builder's build script) and re-emitted by
```
MdTeach, line 155: Alex, don't follow this:

> guest ELFs are embedded inside the datatool binary

You are saying ELF is embedded to datatool image ??

.github/workflows/ci-build.yml
purusang: >  Should we use sp1-docker-build if this is essential for production deployment?

@alexhui01 I had discussed this with @MdTeach , this is an entirely different concern. With this flag our sp1 key will be consistent across every builds. So it should be used as a feature flag while doing cargo build.

.github/workflows/ci-build.yml
alexhui01: Yes, that's what I get from the AI analysis, the elfs is embedded to the datatool binary. So the datatool binary matters here to guarantee we have produced the same elfs results which we can use for the later param generation process as well.

.github/workflows/ci-build.yml
evgenyzdanovich: Alex - answering your question: before, we used to have a whole different code and pipeline, with a separate binary - prover client -that handles everything (and most, if not all, parts were already in place in docker for that specific binary, with a separate step(s)).

right now the prover is integrated (quite recently) - every binary uses / would use the prover library code (also new), hence we have a drift with elfs.

P.S. that was part of the reason I was so eager to trace the git history (slack thread).

docker/strata/Dockerfile
purusang: Now the local docker run will use Dockerfile.local

purusang (approved):

MdTeach (approved):

purusang (approved):

PR #1670, opened 2026-04-21
STR 2751 Investigate flaky OL block timings

krsnapaudel:
## Description

Investigated flaky block time issue from [STR-2751](https://alpenlabs.atlassian.net/browse/STR-2751). Found that his is related to OL canonical chain tip resolution.

- Align sequencer tip-resolution across `GenerationTick` (template generation) and the duty-extraction RPC so both read from the same source. Before this change, GenerationTick preferred `status_channel` while `get_sequencer_duties` read directly from storage — so during the FCM window between publishing a new tip to the status channel and writing the block's `Valid` status to storage, the two paths disagreed: GenerationTick cached a template for the new tip while the signer's RPC was still asking for a template keyed on the old tip (whose template had just been consumed). The signer saw empty duties for a slot that was otherwise ready.
- Stamp the block's `timestamp` with `SystemTime::now()` instead of the deterministic `parent_ts + ol_block_time_ms`. Chain time now reflects real production cadence, which surfaces cadence anomalies in the block headers themselves.
- Warn when the wall-clock interval from the parent block exceeds `ol_block_time_ms * (1 + BLOCK_TS_DRIFT_TOLERANCE_PCT / 100)` (default 20%). Silent in a healthy chain; fires on the production race symptom.

Reproduced the issue in a functional test (without the fix) and tested the fix using two approaches. Sampled 80 block times from block headers and block duties (by adding explicit timestamp to duty struct).
1. Removed the status channel based tip resolution so both template generation and rpc get tip from db.
2. Adding status channel based tip resolution to rpc path so both template generation and rpc first check status channeld and fall back to db.

In both cases, block times are consistent after making both parties use shared tip resolution.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This fix is related to [STR-3105](https://alpenlabs.atlassian.net/browse/STR-3105). Just addressing the block time issue here. Let me know if STR-3105 needs to go in first.

We could add a functional test that was used to test the fix. But that requires changing the defaults for block time and duty poll interval if we want to sample ample number of block times. Also need explicit timestamp in duty struct. Hence run locally, but not committed.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-2751](https://alpenlabs.atlassian.net/browse/STR-2751)

[STR-2751]: https://alpenlabs.atlassian.net/browse/STR-2751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[STR-3105]: https://alpenlabs.atlassian.net/browse/STR-3105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/strata/src/sequencer/node_context.rs
+++ b/bin/strata/src/sequencer/node_context.rs
@@ -1,14 +1,23 @@
 //! Concrete [`SequencerContext`] implementation for the Strata node.
 
-use std::sync::Arc;
+use std::{
+    sync::Arc,
+    time::{SystemTime, UNIX_EPOCH},
+};
 
 use async_trait::async_trait;
+use strata_identifiers::OLBlockId;
 use strata_ol_block_assembly::{BlockAssemblyError, BlockasmHandle};
 use strata_ol_sequencer::{BlockGenerationConfig, SequencerContext, SequencerContextError};
-use strata_primitives::{OLBlockCommitment, OLBlockId};
 use strata_status::StatusChannel;
 use strata_storage::NodeStorage;
 use tracing::{debug, warn};
+
+use crate::sequencer::tip::resolve_canonical_tip;
+
+/// Percentage drift of block wall-clock spacing above the configured
+/// `ol_block_time_ms` that triggers a cadence warning.
+const BLOCK_TS_DRIFT_TOLERANCE_PCT: u64 = 20;
 
 /// Node-level context providing concrete infrastructure for the sequencer service.
 pub(crate) struct NodeSequencerContext {
@@ -32,36 +41,19 @@ impl NodeSequencerContext {
             ol_block_time_ms,
         }
     }
-
-    /// Resolve the current chain tip block ID.
-    async fn resolve_tip(&self) -> Result<OLBlockId, SequencerContextError> {
-        if let Some(tip) = self
-            .status_channel
-            .get_ol_sync_status()
-            .map(|s| *s.tip_blkid())
-        {
-            return Ok(tip);
-        }
-
-        match self.storage.ol_block().get_canonical_tip_async().await {
-            Ok(Some(commitment)) => Ok(*commitment.blkid()),
-            Ok(None) => {
-                warn!("canonical tip not found yet");
-                Ok(OLBlockId::default())
-            }
-            Err(e) => Err(SequencerContextError::Db(e)),
-        }
-    }
 }
 
 #[async_trait]
 impl SequencerContext for NodeSequencerContext {
     async fn generate_template_for_tip(&self) -> Result<Option<OLBlockId>, SequencerContextError> {
-        let tip_blkid = self.resolve_tip().await?;
-        if tip_blkid == OLBlockId::default() {
+        let Some(parent_commitment) = resolve_canonical_tip(&self.status_channel, &self.storage)
+            .await
+            .map_err(SequencerContextError::Db)?
+        else {
             debug!("template generation skipped: canonical tip unavailable");
             return Ok(None);
-        }
+        };
+        let tip_blkid = *parent_commitment.blkid();
 
         debug!(tip_blkid = ?tip_blkid, "template generation attempt");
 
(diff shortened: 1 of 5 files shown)


bewakes (approved):

bin/strata/src/sequencer/node_context.rs
```diff
@@ -77,16 +69,31 @@ impl SequencerContext for NodeSequencerContext {
...
+            warn!(
+                time_since_parent,
+                block_time_ms = self.ol_block_time_ms,
+                threshold_ms,
+                parent_ts,
+                target_ts,
+                "block wall-clock interval exceeds block_time by more than {BLOCK_TS_DRIFT_TOLERANCE_PCT}%",
+            );
+        }
+
```
purusang, line 90: I am not sure how this helps in resolving 2*block_time issue we had been seeing. It is also changing policy of timestamp that we want to embed in our headers. If there is some external document to this, please lmk.

purusang: Having current time seems like the correct approach, should there be a guard to prevent faster block productions than configured?

krsnapaudel: @purusang The tick only fires after configured block time. I don't think calculating current block timestamp from `parent_ts` + block time was correct. This hides the drift in block production. This change here is for correctness and visibility only, not part of the fix for actual drift observed.

bin/strata/src/sequencer/tip.rs
```diff
@@ -0,0 +1,22 @@
...
+/// `storage.ol_block().get_canonical_tip_async()` when the channel has not
+/// been initialized yet.
+pub(crate) async fn resolve_canonical_tip(
+    status_channel: &StatusChannel,
+    storage: &NodeStorage,
+) -> DbResult<Option<OLBlockCommitment>> {
+    if let Some(tip) = status_channel.get_ol_sync_status().map(|s| s.tip) {
+        return Ok(Some(tip));
+    }
+    storage.ol_block().get_canonical_tip_async().await
```
purusang, line 21: Do you think this fallback may not be necessary after https://github.com/alpenlabs/alpen/pull/1672?

krsnapaudel: @purusang I think it's good to have the fallback right? Not necessarily used, but db is the source of truth.

purusang: I don't mean to block merging this since https://github.com/alpenlabs/alpen/pull/1672 may take some days and you may resolve and move on.
Its a watch channel so only path to falling back is when status channel is not set and in current codebase it happens only in genesis, (or may be even in restarts) so it is dead code after that one instance. And after 1672, I do not see that happening.

bin/strata/src/sequencer/node_context.rs
```diff
@@ -77,16 +69,31 @@ impl SequencerContext for NodeSequencerContext {
             })?;
         let parent_header = parent_block.header();
 
-        let parent_commitment = OLBlockCommitment::new(parent_header.slot(), tip_blkid);
-        let target_ts = parent_header
-            .timestamp()
-            .saturating_add(self.ol_block_time_ms);
+        let parent_ts = parent_header.timestamp();
+        let target_ts = SystemTime::now()
```
storopoli, line 73: `target_ts` now comes from the wall clock at template-generation time, but template generation still runs on a fixed interval from service startup. After any late block, the next generation tick can land almost immediately after the new parent arrives, and the signer will submit right away because `wait_duration()` is already zero. That compresses block spacing below `ol_block_time_ms`, whereas the old `parent_ts + ol_block_time_ms` logic preserved the configured cadence.

storopoli (changes requested):
Very good PR. There's an issue with having `target_ts` with `SystemTime::now()`.

bin/strata/src/sequencer/node_context.rs
purusang: Hmm, then this [check](https://github.com/alpenlabs/alpen/pull/1670#discussion_r3122225539) is needed?

krsnapaudel: @storopoli "whereas the old `parent_ts + ol_block_time_ms` logic preserved the configured cadence.
This is not accurate. This hides the fact that block was likely not produced according to the cadence. Do you want a false indication of same duration?

storopoli: Agreed that `parent_ts + ol_block_time_ms` hid late production in the header timestamp.

My concern is different: in the current flow the template timestamp is also the signer’s readiness gate (`wait_duration()`), so changing it to `SystemTime::now()` changes behavior, not just reporting. After a late block, the next generation tick can happen almost immediately, and the signer can submit right away, which allows parent/child spacing below `ol_block_time_ms`.

I think this needs a clamp like `max(now, parent_ts + ol_block_time_ms)` rather than raw `now()`.

krsnapaudel: @storopoli But how does the clamp change anything? It's just manipulating time incorrectly. The time should be when block template generation is called.

storopoli: Revisiting this, I don't think this is a blocker like I did before.
Solving

storopoli (approved):
ACK 44998c666340b3413c2e2496c579a6dcf8820c6c

PR #1671, opened 2026-04-22
feat(ee-da): add `block_hash` to `EvmHeaderSummary` (STR-2320)

storopoli:
## Description

Extends `EvmHeaderSummary` — the compact EVM header we pin inside each
`DaBlob` posted to L1 — with one new fields: `block_hash` (the last
block's own hash, which becomes `parent_hash` of the next block).
`state_root` can be reconstructible, if the recovery path indexes DA from the trusted base and applies each `BatchStateDiff` in order, and does not need to be carried in EvmHeaderSummary. The original shape shipped in #1340 was designed around
block-assembly needs only (EIP-1559 base-fee calc, gas-limit adjust,
timestamp monotonicity). With the EE proof statement now finalized, a
sequencer recovering purely from L1 DA also needs this field to
verify chain continuity.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- **DA wire format:** adding fields to a `#[derive(Codec)]` struct
  changes the on-wire layout of `DaBlob`. Blobs written with the
  previous layout will not decode into the new struct. This is
  considered safe because the chunked-envelope DA format is still
  pre-production — if that assumption is wrong, we should land a
  `DaBlob` version byte first.
- **Test strategy:** the trait impl is thin, so the reth `Header →
  EvmHeaderSummary` mapping was extracted into a free function
  `summarize_header` and unit-tested directly. The test asserts every
  field maps from the correct source and includes a `block_hash !=
  state_root` sanity check specifically designed to catch a
  field-swap regression (verified locally by introducing and then
  reverting a swap).
- **AI assistance:** this PR was drafted with AI assistance per
  [CONTRIBUTING.md](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice).

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-2320

### DIFF
--- a/crates/alpen-ee/common/src/types/da.rs
+++ b/crates/alpen-ee/common/src/types/da.rs
@@ -4,6 +4,7 @@
 //! into Bitcoin envelope chunks for inscription.
 
 use alpen_reth_statediff::BatchStateDiff;
+use strata_acct_types::Hash;
 use strata_codec::{decode_buf_exact, encode_to_vec, BufDecoder, Codec, CodecError};
 use strata_crypto::hash;
 use strata_identifiers::Buf32;
@@ -12,15 +13,19 @@ use crate::BatchId;
 
 /// Compact summary of the last EVM block header in a batch.
 ///
-/// Captures the subset of the EVM block header needed to build the next
-/// block during DA-only chain reconstruction. A new sequencer recovering
-/// from L1 DA has the [`BatchStateDiff`] (account/storage changes) but
-/// **not** the block headers themselves — these fields fill that gap.
+/// Captures the subset of the EVM block header a sequencer recovering
+/// purely from L1 DA needs to (a) build the next EVM block and (b) feed
+/// the EE proof statement. A fresh sequencer has the [`BatchStateDiff`]
+/// (account/storage changes) but **not** the block headers themselves —
+/// these fields fill that gap.
 ///
 /// - `base_fee`, `gas_used`, `gas_limit` feed the EIP-1559 base-fee calculation and gas-limit
 ///   adjustment for the next block.
 /// - `timestamp` enforces monotonicity (`next > parent`).
 /// - `block_num` identifies where the chain continues.
+/// - `block_hash` becomes the `parent_hash` of the next block and anchors chain-continuity checks
+///   in the EE proof.
+/// - `state_root` pins the pre-state root the next block's EE proof opens against.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Codec)]
 pub struct EvmHeaderSummary {
     /// Block number of the last EVM block in this batch.
@@ -33,6 +38,16 @@ pub struct EvmHeaderSummary {
     pub gas_used: u64,
     /// Gas limit of the last EVM block.
     pub gas_limit: u64,
+    /// Block hash of the last EVM block in this batch.
+    ///
+    /// Becomes the `parent_hash` of the next block; required by the EE
+    /// proof for chain-continuity verification.
+    pub block_hash: Hash,
+    /// Post-state root of the last EVM block.
+    ///
+    /// Required by the EE proof so the next block's pre-state root can
+    /// be pinned to DA.
+    pub state_root: Hash,
 }
 
 /// DA blob containing batch metadata and state diff.
@@ -308,6 +323,8 @@ mod tests {
                 base_fee: 1_000_000_000,
                 gas_used: 15_000_000,
                 gas_limit: 30_000_000,
+                block_hash: Hash::from([0x33; 32]),
+                state_root: Hash::from([0x44; 32]),
             },
             state_diff: BatchStateDiff::default(),
         }
--- a/crates/alpen-ee/common/src/types/da.rs
+++ b/crates/alpen-ee/common/src/types/da.rs
@@ -4,7 +4,6 @@
 //! into Bitcoin envelope chunks for inscription.
 
 use alpen_reth_statediff::BatchStateDiff;
-use strata_acct_types::Hash;
 use strata_codec::{decode_buf_exact, encode_to_vec, BufDecoder, Codec, CodecError};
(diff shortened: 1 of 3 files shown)


delbonis (approved):
Looks good.  I'm a little confused why summary fields all *only* refer to the terminal though.

storopoli:
> I'm a little confused why summary fields all _only_ refer to the terminal though.

Yeah, that’s intentional. `EvmHeaderSummary` is really “the subset of the last block in the batch that must survive into DA so the next block can be built/proven from L1 only.”

So all of its fields are terminal-by-design: `block_num`, `timestamp` etc... are the parent inputs the next EE proof needs.

Earlier blocks in the batch are already collapsed into the aggregated `state_diff`, so we only need the terminal header boundary here. The main case where non-terminal headers would matter is `BLOCKHASH` OPCODE/ancestor-header support, which I kept out of scope for this PR.

I can add a clarifying sentence to the doc comment if that would make the intent more obvious.

delbonis (approved):

storopoli:
Waiting on @irnb's review to add to the queue

crates/alpen-ee/common/src/types/da.rs
```diff
@@ -33,6 +38,16 @@ pub struct EvmHeaderSummary {
     pub gas_used: u64,
     /// Gas limit of the last EVM block.
     pub gas_limit: u64,
+    /// Block hash of the last EVM block in this batch.
+    ///
+    /// Becomes the `parent_hash` of the next block; required by the EE
+    /// proof for chain-continuity verification.
+    pub block_hash: Hash,
```
irnb: @storopoli  I don't understand why we need this?

Our EE proof should be responsible for verifying the transition from
`prev_batch.blocks[-1].hash`  to  `curr_batch.blocks[-1].hash`
But here we're also including `curr_batch.blocks[-2].hash`.

crates/alpen-ee/common/src/types/da.rs
```diff
@@ -33,6 +38,16 @@ pub struct EvmHeaderSummary {
...
+    /// Post-state root of the last EVM block.
+    ///
+    /// Required by the EE proof so the next block's pre-state root can
+    /// be pinned to DA.
+    pub state_root: Hash,
```
irnb: Same question for `curr_batch.blocks[-2].state_root`.

irnb (changes requested):

crates/alpen-ee/common/src/types/da.rs
storopoli: Good point. I dropped `block_hash` from `EvmHeaderSummary` because it was duplicating `DaBlob.batch_id.last_block`, which already carries `curr_batch.blocks[-1].hash`.

I kept `state_root` because that value is not in the batch id. It is the terminal post-state root for `curr_batch.blocks[-1]`, and it is the value a fresh L1-only sequencer needs as the next block/proof pre-state root. I also tightened the docs to call out that distinction explicitly.

bewakes (approved):
LGTM

crates/alpen-ee/common/src/types/da.rs
```diff
@@ -33,6 +38,11 @@ pub struct EvmHeaderSummary {
     pub gas_used: u64,
     /// Gas limit of the last EVM block.
     pub gas_limit: u64,
+    /// Post-state root of the last EVM block.
+    ///
+    /// Required by the EE proof so the next block's pre-state root can
+    /// be pinned to DA.
+    pub state_root: Hash,
```
irnb: @storopoli Won't the state root be reconstructible by indexing all DA txs and applying state diffs up to the tip?
`curr_batch.blocks[-1].state` should be reconstructible from state diff data on L1. 

I think we can drop this, let me know if I'm missing something. 🙏

irnb (changes requested):

crates/alpen-ee/common/src/types/da.rs
storopoli: I think you’re right. If the recovery path indexes DA from the trusted base and applies each `BatchStateDiff` in order, the terminal post-state root is reconstructible and does not need to be carried in `EvmHeaderSummary`.

I’ll drop `state_root` too and keep `EvmHeaderSummary` scoped to the header fields that are not derivable from the DA state diff but are still needed for next-block assembly (`block_num`, `timestamp`, `base_fee`, `gas_used`, `gas_limit`).

Please heads-up because that will need more work on the DA part of things. Especially when syncing from L1 in EE, so you should update/create tickets if needed.

See 1299aeb38f437d7e6825a621611fae48a0c50b3d

irnb: cc: @krsnapaudel

irnb:
```
#[derive(Debug, Clone, Copy, PartialEq, Eq, Codec)]
pub struct EvmHeaderSummary {
    /// Block number of the last EVM block in this batch.
    pub block_num: u64,
    /// Unix timestamp (seconds) of the last EVM block.
    pub timestamp: u64,
    /// Base fee per gas (EIP-1559) of the last EVM block.
    pub base_fee: u64,
    /// Total gas consumed by the last EVM block.
    pub gas_used: u64,
    /// Gas limit of the last EVM block.
    pub gas_limit: u64,
}
```

@storopoli, just for final confirmation these are the fields from the last block's header in a batch, and they're the only non-reconstructible fields from DA that a fresh sequencer would need to build the next block on top of it and generate EE proofs.

storopoli: Yes.

storopoli: Solving this.

storopoli:
@irnb ping.

irnb (approved):

PR #1677, opened 2026-04-23
STR-3150: wire in prover requests deadline

evgenyzdanovich:
## Description

Wire in prover requests deadline

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/bin/strata/src/prover/mod.rs
+++ b/bin/strata/src/prover/mod.rs
@@ -83,7 +83,11 @@ pub(crate) fn start_prover_service(
             // re-wraps it in its own Arc inside RemoteStrategy. SP1Host
             // is Clone (only holds a SP1ProvingKey), so cloning from the
             // shared static is fine.
-            let host: zkaleido_sp1_host::SP1Host = (**CHECKPOINT_NEW_HOST).clone();
+            let mut host: zkaleido_sp1_host::SP1Host = (**CHECKPOINT_NEW_HOST).clone();
+            if let Some(secs) = prover_config.sp1_proof_deadline_secs {
+                host = host.with_deadline(Duration::from_secs(secs));
+                info!(deadline_secs = secs, "sp1 prover deadline configured");
+            }
             ProverBuilder::new(spec)
                 .task_store(task_store)
                 .receipt_hook(hook)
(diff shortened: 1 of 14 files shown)


irnb (approved):

evgenyzdanovich:
@irnb a slight (yet important) change I just pushed after you approved -- setting default (IF we forget to specify it in the config) to 4h, instead of relying on sp1 built-ins

delbonis (approved):

bewakes (approved):

evgenyzdanovich:
rebase changes -- had to set the deadline for the ee proofs (they would also have the same issue otherwise). 
please re-approve

bewakes (approved):

bin/strata/src/prover/mod.rs
```diff
@@ -33,6 +33,11 @@ use crate::run_context::RunContext;
 // TODO(STR-3064): make this configurable via ProverConfig.retry_interval.
 const PROVER_RETRY_INTERVAL: Duration = Duration::from_secs(5);
 
+/// Default end-to-end deadline applied to the SP1 prover network when
+/// `ProverConfig::sp1_proof_deadline_secs` is not set. Chosen to comfortably
+/// cover checkpoint proofs while still failing fast on stuck requests.
+const DEFAULT_SP1_DEADLINE_SECS: u64 = 4 * 60 * 60;
+
```
storopoli, line 40: This along with the same constant in `bin/alpen-client/src/main.rs` should be moved to `crates/primitives/src/constants.rs` since they are the same and should be kept in sync (same default)

storopoli (approved):
ACK 87da9284c4561eca62c9da8089e382b66e26231e

Non-blocking comment on duplicate constants.

bin/strata/src/prover/mod.rs
evgenyzdanovich: Im not so sure about shared defaults across binaries, and here specifically in this case - proofs naturally take different time, I dont see any reason why would we want to have the same default.

For now it's rather a safety mechanism (_primarily for the case when we have human mistake and didnt specify deadline in some configs_) than part of the logic that survives any meaningful refactoring.

Also, it doesnt feels like a _primitive constant_, so I'm hesitant of putting it there: to be clear, `primitives/constants` is already a total mess -- a bunch of constants both unused/should be unused and used for various purposes -- but it's not a reason to add another one and make it even more convoluted).

PR #1679, opened 2026-04-23
feat: functional test for Alpen predicate key rotation via admin

prajwolrg:
## Description

Adds an end-to-end functional test that exercises the new `AlpenAdministrator` admin flow: submitting an admin `PredicateUpdate` tx on L1, watching it flow through the ASM → OL pipeline, and asserting the Alpen snark account's `update_vk` rotates in OL state.

The test rotates the predicate through `NeverAccept` → `AlwaysAccept` → `Sp1Groth16:<hex>` → `Bip340Schnorr:<hex>`, waiting for each transition to land in OL state before submitting the next.

**Depends on [alpenlabs/asm#78](https://github.com/alpenlabs/asm/pull/78)**, which adds the `AlpenAdministrator` role — a new admin authority (alongside the existing roles) that has the authority to update the EE verification key (i.e. the snark account's `PredicateKey`). The ASM dep in this repo is bumped to the commit from that PR.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist


bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
bewakes, line 1: Not a blocker for this PR, but the commit-reveal transaction building feels it is being duplicated to what we have in `btcio::writer::builder`. Ideally we would only need to create an envelope payload and the rest of the things like size/fee estimation, generating key pairs, assembling commit and reveal txs should be generic. I suggest to see if that can be quickly done to prevent logic duplication. If can't be done, that's fine too.

delbonis, line 1: Should be called `create_snark_acct_predicate_update`, see [ticket](https://alpenlabs.atlassian.net/browse/STR-3184).

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
```diff
@@ -0,0 +1,374 @@
...
+    pub btc_url: String,
+
+    /// bitcoin RPC username
+    #[argh(option)]
+    pub btc_user: String,
+
+    /// bitcoin RPC password
+    #[argh(option)]
+    pub btc_password: String,
+
```
delbonis, line 71: I think these across the different subcommands can all be `#[argh(flatten)]`ed from a single def.

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
```diff
@@ -0,0 +1,374 @@
...
+        &reveal_script,
+        &taproot_spend_info,
+        &envelope_keypair,
+    )?;
+
+    Ok((commit_tx, reveal_tx))
+}
+
+fn build_ee_update_action(key: PredicateKey) -> MultisigAction {
+    let update = PredicateUpdate::new(key, ProofType::EeStf);
```
delbonis, line 210: How does this work if we're not changing the "blessed" Alpen snark account but changing some other one?  I can't find this enum variant in the asm repo.

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
```diff
@@ -0,0 +1,374 @@
...
+        .add_envelopes(&envelope_chunks)
+        .map_err(|e| Error::TxBuilder(format!("failed to add envelope bytes: {e}")))?
+        .build_without_min_check()
+        .map_err(|e| Error::TxBuilder(format!("failed to finalize envelope script: {e}")))?;
+
+    let taproot_spend_info = TaprootBuilder::new()
+        .add_leaf(0, reveal_script.clone())
+        .map_err(|e| Error::TxBuilder(format!("failed to add taproot leaf: {e}")))?
+        .finalize(SECP256K1, xonly_pubkey)
+        .map_err(|e| Error::TxBuilder(format!("failed to finalize taproot tree: {e:?}")))?;
```
delbonis: This error handling is a little awkward, maybe we could just anyhow this to improve cleanliness?  I don't expect this to get triggered normally in practice.

delbonis (changes requested):
I'm also minorly concerned about the envelope building code seemingly being duplicated.  Do we not have this written out as a resuable library?  If not then we can merge it as-is but should create a ticket to create one in the strata-common repo.

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
prajwolrg: This only supports "blessed" Alpen snark account. https://github.com/alpenlabs/asm/pull/78 explicitly only supports Alpen update. If we want to change some other one, we need to come up with another mechanism.

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
prajwolrg: The duplication was needed because of `btcio::writer::builder::build_and_sign_envelope_txs` isn't a drop in replacement because:
1. it signs the commit via bitcoind-loaded descriptors while the CLI uses a local bdk wallet from a hardcoded test XPRIV, 
2. `create_envelope_transactions` does its own naive UTXO selection that doesn't plug into bdk's coin selection, and
3. the high-level entry points require a full async `WriterContext`. The fix likely involves extracting the `envelope/taproot/reveal` primitives so the sequencer and CLI paths can share them while each owns its own commit funding/signing.

prajwolrg: Should we call this `alpen_acct_predicate_update`, since it is meant to support Alpen account (or the account with `AccountSerial(128)`)?

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
prajwolrg: Sure. Since this is for `test-cli`, I think using anyhow should be good.

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
prajwolrg: I don't think flatten is supported in any released `argh` version. Look at [this](https://github.com/google/argh/issues/126) issue.

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
prajwolrg: `anyhow` is not a dependency on `strata-test-cli`. I'm not sure if it's a good idea to introduce this in this PR now.

bewakes (approved):
Looks fine.

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
delbonis: That seems fine?

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
delbonis: Maybe we should loop product/@john-light into this?

john-light: A question I have first is: do we intend to generalize this mechanism to be usable by all snark accounts, or will this be unique to Alpen?

delbonis: I'm in favor of the generality, but I don't know if that's undesirable for product reasons.

prajwolrg: I'm not in favor of the generality. I think each snark account would want to have it's own governance mechanism and timing requirements for the upgrades.

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
prajwolrg: Not really. See [5845301](https://github.com/alpenlabs/alpen/pull/1679/commits/5845301cbbf10a9649505fab8b0432df8edbdf38). I was hesitant to make changes to the codebase broadly. The same pattern was used in other places as well.

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
john-light: In that case I think it's fine to specify `alpen` in the name, for clarity's sake, since we will probably remove or significantly change this later when we implement support for multiple snark accounts, and they will each have their own update mechanism. Right?

delbonis: Sure but then I'm not sure of the wisdom of enshrining a particular account ID/serial as having this dedicated mechanism to change its key baked all the way down to the ASM layer.  Shouldn't the entire mechanism exist in the orchestration layer through some more general mechanism?  Yes this system is "just another governance mechanism", but it privileges a particular account at the protocol level in a way that isn't directly extensible/generalizable for future use cases.

For example, I had mentally toyed with the idea that accounts could have a "governor" account tied to them that has the ability to unilaterally issue commands, like in this case "change the update key".

prajwolrg: I agree that we want to eventually go with a more generic approach. But for mainnet, I thought we were in the agreement that this is how we wanted to have upgrades for the Alpen EE.

delbonis: Alright I'm going to approve this and we can punt the scoping discussion.

delbonis (approved):

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
```diff
@@ -0,0 +1,372 @@
...
+        json!({
+            "commit_txid": commit_txid,
+            "reveal_txid": reveal_txid,
+        })
+    );
+
+    Ok(())
+}
+
+// TODO(STR-3191): deduplicate ennvelope commit/reveal transaction
```
storopoli:
```suggestion
// TODO(STR-3191): deduplicate envelope commit/reveal transaction
```

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
```diff
@@ -0,0 +1,372 @@
...
+    #[argh(option, default = "2")]
+    pub fee_rate: u64,
+
+    /// commit output value in sats (default 20000)
+    #[argh(option, default = "20_000")]
+    pub commit_output_sats: u64,
+}
+
+fn parse_predicate_key(value: &str) -> Result<PredicateKey, String> {
+    serde_json::from_str(&format!("\"{value}\"")).map_err(|e| e.to_string())
```
storopoli: This is bad. Inputs containing `"` or `\` will produce malformed JSON. Use `serde_json::Value::String(value.to_owned())` instead.

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
```diff
@@ -0,0 +1,372 @@
...
+                let msg = err.to_string().to_lowercase();
+                let should_retry = msg.contains("missing") || msg.contains("invalid input");
+                if should_retry && attempt + 1 < DEFAULT_RETRY_COUNT {
+                    thread::sleep(Duration::from_millis(RETRY_SLEEP_MS));
+                    continue;
+                }
+                return Err(err);
+            }
+        }
+    }
```
storopoli, line 369: This is bad. The same as https://github.com/alpenlabs/alpen/pull/1746#discussion_r3180154737
We should do something in `bitcoind-async-client`. Will do a PR since this is the 2nd time I see this thing arising today in PRs here.

storopoli (changes requested):

bin/strata-test-cli/src/cmd/create_ee_predicate_update.rs
storopoli: Ok please bump to `0.10.4`. Tagged and published in https://crates.io/crates/bitcoind-async-client

Fixed in https://github.com/alpenlabs/bitcoind-async-client/pull/110

bewakes: Thanks Jose for the prompt fix in `bitcoind-async-client` library. One thing though, since we heavily use the library(and the community also uses it), do you think it's best to have at least a single person from Alpen review the PRs in the library? I see that it was merged without any reviews.

storopoli: Ideally if it were a complex bug fix or a new feature, then yes.
But a trivial unglobbing of error codes with a deprecation notice I don't see much benefit.

EDIT: As we're in dire need of the unglobbing :)

prajwolrg: `strata-test-cli` doesn't depend on `bitcoind-async-client`. I'd defer for this fix as part of https://github.com/alpenlabs/alpen/pull/1746 instead.

storopoli: Ok done, please rebase.

storopoli:
@codex review

storopoli (approved):
ACK e5155b603d225146275e19b12adb16898c8eb35b

PR #1686, opened 2026-04-27
Str 3095 chain indexing rework

bewakes:
## Description

**Schema (`crates/db/types/src/ol_state_index.rs`)**
- New types per [the schema doc](https://www.notion.so/OL-State-Indexing-Schema-34b901ba000f80a5bb52ede704b6e318).
- Values are CBOR (so we can evolve fields without resync), keys are Borsh.

**Trait (`OLStateIndexingDatabase`)**
- Two write paths because the producers are different: one-shot per epoch for checkpoint-sync, per-block RMW append for block-sync. Plus a finalization stamp.
- Reads cover the epoch row, per-(account, epoch) updates and inbox, and creation-epoch lookup.

**Sled (`crates/db/store-sled/src/ol_state_index/`)**
- Four trees: epoch row, per-(account, epoch) updates and inbox writes, reverse creation index.
- Multi-tree writes go through `with_retry` so a single call is atomic — no half-applied state.

**Storage layer (`crates/storage/`)**
- New manager with the usual blocking/async/chan trio, exposed on `NodeStorage`.
- Slot-range filter helpers at the manager level for upcoming `get_blocks_summaries` work — no DB-side change needed since each record carries its block commitment.

**Chain worker/genesis changes**
- `store_block_output` builds the per-block writes from `IndexerWrites` + surrounding context. `store_summary` stamps the epoch commitment after the canonical summary insert.
- Genesis seeds epoch 0 with the genesis account list.

**RPC (`bin/strata/src/rpc/`, `crates/ol/rpc/types/`)**
- `update_inputs` on the epoch summary is now a `Vec` (was `Option`). One entry per record, each with its own proof state, extra_data, and inbox messages fetched for its own `[prev_next_idx, this_next_idx)` range.
- Added one helper for "epoch -> terminal-block OL state" since current and previous epoch lookups were duplicated.

**Removed**
- Old account genesis database and manager. Moved the tree into the indexing DB.
- `store_auxiliary_data` in chain-worker-new (was a stubbed warning).
<hr/>
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The code changes in the PR are assisted by Claude code and design is by author.

- [ ]  Yes
- [ ]  No

## Checklist


crates/db/types/src/ol_state_index.rs
```diff
@@ -0,0 +1,285 @@
...
+/// One call to `apply_block_indexing` consumes one of these; the DB appends
+/// to existing per-(account, epoch) rows and updates the common epoch row
+/// with any newly created accounts.
+#[derive(Clone, Debug, Default)]
+pub struct BlockIndexingWrites {
+    pub epoch: Epoch,
+    pub block: OLBlockCommitment,
+    pub created_accounts: Vec<AccountId>,
+    pub account_updates: BTreeMap<AccountId, Vec<AccountUpdateRecord>>,
+    pub account_inbox_writes: BTreeMap<AccountId, Vec<InboxMessageRecord>>,
```
delbonis: It would be good to unify the "data" fields with the epoch-level indexing container so that we can talk about the indexed writes without the contextual metadata.

Also make these non-pub.

crates/db/types/src/ol_state_index.rs
```diff
@@ -0,0 +1,285 @@
...
+/// Single-call payload for checkpoint-sync indexing writes.
+///
+/// Applied atomically by
+/// [`OLStateIndexingDatabase::apply_epoch_indexing`](crate::traits::OLStateIndexingDatabase::apply_epoch_indexing).
+#[derive(Clone, Debug)]
+pub struct EpochIndexingWrites {
+    pub epoch: Epoch,
+    pub common: EpochIndexingData,
+    pub account_updates: BTreeMap<AccountId, AccountUpdateEntry>,
+    pub account_inbox: BTreeMap<AccountId, AccountInboxEntry>,
```
delbonis: Make these non-pub.

crates/db/types/src/ol_state_index.rs
```diff
@@ -0,0 +1,285 @@
...
+}
+
+/// Inbox message append, optionally tagged with the block in which it was inserted.
+///
+/// `block_commitment` is `Some` for block-sync writes, `None` for checkpoint-sync.
+/// `entry_bytes` is the SSZ encoding of the underlying `MessageEntry`.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct InboxMessageRecord {
+    entry_bytes: Vec<u8>,
+    block_commitment: Option<OLBlockCommitment>,
```
delbonis, line 244: Can't this be inferred from context?

crates/db/types/src/ol_state_index.rs
```diff
@@ -0,0 +1,285 @@
...
+
+/// Block-sync metadata for an account update.
+///
+/// Present only when produced by a block-syncing node; checkpoint-sync
+/// records leave this `None`.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct AccountUpdateMeta {
+    block_commitment: OLBlockCommitment,
+    final_state_root: Hash,
+}
```
delbonis, line 147: Probably would be better to make these two fields `Option`s instead of making this whole container an `Option` in `AccountUpdateRecord`.

crates/db/types/src/ol_state_index.rs
delbonis, line 1: Something that I see here that's somewhat worrying me is conflating the types used in the database interface with the types used on-disk.  This doesn't have to be wrong, sometimes it's perfectly fine, but we're trying to be flexible and future-proof here.  We must not try to express API constraints (ie. what's available in what contexts) in the types used in the database.  The database doesn't care that block-level data can only exist in full sync mode, if we launch the client and the database has data there, it should be able to express whatever it has.  We can make sense of it above the database layer.

crates/db/types/src/traits.rs
```diff
@@ -661,33 +665,59 @@ pub trait OLBlockDatabase: Send + Sync + 'static {
...
 
-    /// Gets the creation epoch for an account, if recorded.
-    fn get_account_creation_epoch(&self, account_id: AccountId) -> DbResult<Option<Epoch>>;
+    /// Atomically applies a single block's incremental indexing writes.
+    ///
+    /// Appends to existing per-(account, epoch) entries, updates the common
+    /// row's `created_accounts`, and inserts creation-epoch index entries
+    /// for any newly created accounts. Caller is responsible for not invoking
+    /// this twice for the same block.
+    fn apply_block_indexing(&self, writes: BlockIndexingWrites) -> DbResult<()>;
```
delbonis: These should take the epoch (commitment?) / block (commitment?) as args instead of embedding it in the writes types.

crates/db/types/src/traits.rs
```diff
@@ -661,33 +665,59 @@ pub trait OLBlockDatabase: Send + Sync + 'static {
...
         &self,
-        key: (AccountId, Epoch),
-    ) -> DbResult<Option<NonEmptyVec<AccountExtraDataEntry>>>;
+        key: AccountEpochKey,
+    ) -> DbResult<Option<AccountUpdateEntry>>;
+
+    /// Returns the per-(account, epoch) inbox entry.
+    ///
+    /// Returns `None` when no inbox writes were recorded for the account in the epoch.
+    fn get_account_inbox_entry(&self, key: AccountEpochKey) -> DbResult<Option<AccountInboxEntry>>;
```
delbonis: These should just take the account and epoch as args instead of requiring this new key type.

crates/db/types/src/traits.rs
```diff
@@ -661,33 +665,59 @@ pub trait OLBlockDatabase: Send + Sync + 'static {
...
-        extra_data: AccountExtraDataEntry,
-    ) -> DbResult<()>;
+    /// Sets the epoch commitment on the existing common row.
+    ///
+    /// Called once by block-sync producers at epoch finalization. Errors if
+    /// no common row exists for the epoch.
+    fn set_epoch_commitment(&self, epoch: Epoch, commitment: EpochCommitment) -> DbResult<()>;
+
+    /// Returns the common indexing data for the given epoch.
+    fn get_epoch_indexing_data(&self, epoch: Epoch) -> DbResult<Option<EpochIndexingData>>;
```
delbonis, line 726: This isn't a problem to present but don't feel required to.

crates/db/types/src/traits.rs
```diff
@@ -661,33 +665,59 @@ pub trait OLBlockDatabase: Send + Sync + 'static {
...
+///   Used by checkpoint-sync producers.
+/// - [`apply_block_indexing`](Self::apply_block_indexing): incremental per-block write. Used by
+///   block-sync producers.
+///
+/// Block-sync also calls [`set_epoch_commitment`](Self::set_epoch_commitment)
+/// once at epoch finalization to stamp the commitment onto the existing common
+/// row; checkpoint-sync includes the commitment in its single write.
+///
+/// Both paths target the same tables; atomicity granularity differs.
+pub trait OLStateIndexingDatabase: Send + Sync + 'static {
```
delbonis, line 678: Needs fn to rollback epoch/block writes.

crates/db/types/src/ol_state_index.rs
```diff
@@ -0,0 +1,285 @@
...
+    PartialOrd,
+    Serialize,
+    Deserialize,
+    BorshSerialize,
+    BorshDeserialize,
+)]
+pub struct AccountEpochKey {
+    pub epoch: Epoch,
+    pub account_id: AccountId,
+}
```
delbonis: I see this type in a lot of places it shouldn't be.  As a database concept it's fine but it shouldn't appear in interfaces unless it *really makes sense to be*.

crates/storage/src/managers/ol_state_indexing.rs
```diff
@@ -0,0 +1,192 @@
...
+        &self,
+        key: AccountEpochKey,
+        start_slot: u64,
+        end_slot: u64,
+    ) -> DbResult<Vec<InboxMessageRecord>> {
+        let Some(entry) = self.ops.get_account_inbox_entry_async(key).await? else {
+            return Ok(Vec::new());
+        };
+        Ok(filter_inbox_by_slot(entry.records(), start_slot, end_slot))
+    }
```
delbonis: These functions are sorta overly-specific and feel misscoped.  They're hard to call correctly because it requires knowing the bounds of the epoch already.  We also would only know the slots in full sync mode so the filtering here can't work correctly in checkpoint sync mode.  I believe the filtering fns below would just discard all messages if we don't have the slot field stored alongside the rest of it.

delbonis (changes requested):
This is a good start but has some modelling issues with the database/interface types.

Also, **remove all uses of borsh**.  Borsh is not suitable for keys because it uses little endian integers and that breaks sorting, plus we already want to get rid of it across the board in the first place.  Use strata-codec or just write direct conversions from the type to a `[u8; N]`-wrapping key type.

bewakes:
Thanks for the prompt feedback @delbonis. I'll make according changes tomorrow.

crates/db/types/src/ol_state_index.rs
bewakes: I had started out that way, but then realized that both are either `None` or `Some` together. So I went with grouping them.

bewakes:
@delbonis PTAL at the latest changes. Here are the major changes commit:
- restructuring some data: https://github.com/alpenlabs/alpen/pull/1686/commits/3689212b7c64d2e6710e78cb45aa70cd0add2b52
- borsh removal: https://github.com/alpenlabs/alpen/pull/1686/commits/1298bd1a306898d0579d0b1b97d64d59f2fdcf54
- rollback: https://github.com/alpenlabs/alpen/pull/1686/commits/f05a47b441b1d91565507bcef0be581f37938df9

bin/strata/src/rpc/node.rs
```diff
@@ -431,122 +577,65 @@ impl<P: OLRpcProvider> OLClientRpcServer for OLRpcServer<P> {
...
-        // This guarantees all returned blocks are from the same chain, not different forks.
-        // Once we reach a finalized block, we can fetch remaining blocks directly by slot.
-        let mut chain_blocks: Vec<(u64, OLBlockId)> = Vec::new();
-
-        // Get the block at end_slot (we'll walk backwards from here)
-        let end_block_id = self.get_canonical_block_at_height(end_slot).await?;
-
-        let Some(current_block_id) = end_block_id else {
-            // No block at end_slot, return empty
+        let chain_blocks = self.collect_canonical_chain(start_slot, end_slot).await?;
```
storopoli, line 587: `get_blocks_summaries` accepts any inclusive slot range and then walks every block, loads each block/state, and prefetches records across the touched epoch span. Other range RPCs in this file reject oversized requests, but this new endpoint only checks ordering. A public caller can request a huge historical range and force expensive DB/state work; add a max range, likely reusing the configured headers/raw-block limit or introducing a specific config.

crates/db/store-sled/src/ol_state_index/db.rs
```diff
@@ -0,0 +1,342 @@
...
+                }
+
+                for (acct, records) in writes.account_inbox() {
+                    if records.is_empty() {
+                        continue;
+                    }
+                    let key = AccountEpochKey::new(epoch, *acct);
+                    let mut existing = inbox_t.get(&key)?.unwrap_or_default();
+                    existing.extend(records.iter().cloned());
+                    inbox_t.insert(&key, &existing)?;
```
storopoli, line 160: `apply_block_indexing` only checks for duplicate blocks while appending account update records. Re-applying the same block with only created accounts or inbox messages will append duplicate `created_accounts` / inbox rows instead of returning `DuplicateBlockIndexing`. Also, because `store_block_output` writes the batch before indexing and stores top-level state after indexing, a retry after a partial write can either duplicate update-free indexing data or fail before the state snapshot is written. Consider tracking block application at the epoch/block level, or checking all write families before mutating any rows.

storopoli (changes requested):
Looks good, but needs to address two issues.

crates/chain-worker-new/src/state.rs
```diff
@@ -282,16 +282,11 @@ impl ChainWorkerServiceState {
         output: &OLBlockExecutionOutput,
         new_state: OLState,
     ) -> WorkerResult<()> {
-        // Store the write batch
         self.ctx
             .store_block_output(block, block_commitment, output)?;
 
-        // Store the full toplevel state
         self.ctx.store_toplevel_state(block_commitment, new_state)?;
```
krsnapaudel: We should have index writing also be an idempotent insert in order for crash and restart issues here. Probably need this logic for upsert.
* If no index records for this block exist: append normally.
* If index records for this block already exist and match the incoming deterministic output: no-op.
* If index records for this block already exist but differ: return an error.

crates/chain-worker-new/src/context.rs
```diff
@@ -222,6 +211,12 @@ impl ChainWorkerContext for ChainWorkerContextImpl {
         let commitment = summary.get_epoch_commitment();
         self.ol_checkpoint_mgr
             .insert_epoch_summary_blocking(summary)?;
+        // Stamp the commitment onto the in-progress epoch indexing row.
+        // NOTE: not in the same transaction as the epoch summary insert; a
+        // crash between these two calls leaves the indexing row without a
+        // commitment, recoverable by re-stamping on restart.
+        self.ol_state_indexing_mgr
+            .set_epoch_commitment_blocking(commitment.epoch(), commitment)?;
```
krsnapaudel: Is this really true? What happens if we crash between
1. `insert` and `set_`? 
2. after the end of this function and before block status is set to valid?
`insert_epoch_summary_blocking` will error out on restart. So we probably need that to handle insert with similar cases as suggested for index writes (non-existent case: success, identical content: no-op, prior summary with different content: error).

krsnapaudel (changes requested):

crates/db/types/src/ol_state_index.rs
delbonis: This ties in with what I was saying elsewhere about how we don't want the database to make assumptions about the way we populate the data.

bin/strata/src/rpc/node.rs
```diff
@@ -133,86 +144,194 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
...
-        current_next_inbox_idx: Option<u64>,
-        prev_epoch_commitment: EpochCommitment,
-    ) -> RpcResult<Vec<MessageEntry>> {
-        // Epoch 0 is genesis-only in current flow: there is no prior terminal state and no
-        // collectable inbox updates, so messages are always empty.
-        if epoch == 0 {
-            return Ok(Vec::new());
+        first_epoch: Epoch,
+        last_epoch: Epoch,
+    ) -> RpcResult<(Vec<AccountUpdateRecord>, Vec<InboxMessageRecord>)> {
```
delbonis: Complex return type, should define a type that captures this.

crates/db/types/src/ol_state_index.rs
```diff
@@ -0,0 +1,237 @@
...
+/// `epoch_commitment` is set once at epoch finalization; `created_accounts`
+/// grows incrementally as blocks in the epoch execute (block-sync) or is
+/// populated in one shot (checkpoint-sync). Each entry pairs the account
+/// with the block that created it (block-sync), or `None` when block
+/// attribution is unavailable (checkpoint-sync). `None`-attributed entries
+/// are immune to per-block rollback and only drop on full-epoch rollback.
+#[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
+pub struct EpochIndexingData {
+    epoch_commitment: Option<EpochCommitment>,
+    created_accounts: Vec<(AccountId, Option<OLBlockCommitment>)>,
```
delbonis: Probably makes sense to create a type for this record so we can naturally associate more data with it in the future if we want to without breaking compat.

crates/chain-worker-new/src/context.rs
```diff
@@ -154,57 +155,45 @@ impl ChainWorkerContext for ChainWorkerContextImpl {
...
+            let record = InboxMessageRecord::new(entry_bytes, Some(commitment));
+            account_inbox_writes
+                .entry(write.account_id())
+                .or_default()
+                .push(record);
+        }
+
+        let writes = IndexingWrites::new(created_accounts, account_updates, account_inbox_writes);
+        self.ol_state_indexing_mgr
+            .apply_block_indexing_blocking(epoch, commitment, writes)?;
```
delbonis: This should be split out into a separate function for organizational purposes and to make it independently testable.

I see the list of created accounts is actually being extracted from the write tracking layer.  This isn't necessarily wrong, but it would be more correct to track it as part of the indexing layer so that the function to convert from the indexing layer's types to the indexing database interface types is only operating on indexing-related types.  It would be okay to make whatever narrow changes are necessary there to expose that.

delbonis (changes requested):
Still some modelling comments on data structures, but better.

Also concur with Jose's comments.

bewakes:
Thanks @krsnapaudel @delbonis and @storopoli for your valuable feedback!

@krsnapaudel, I've addressed your concerns in https://github.com/alpenlabs/alpen/pull/1686/commits/d8f6a5d578ae728dc68ad5260cdefae38334b617

@delbonis, I've addressed your concerns in https://github.com/alpenlabs/alpen/pull/1686/commits/2b49e009a8a823baa98472902ec4d7e40c29871e and https://github.com/alpenlabs/alpen/pull/1686/commits/9e31de2a533d080f549f010a686f4362db76e1ce

@storopoli, I've addressed your concerns in https://github.com/alpenlabs/alpen/pull/1686/commits/fee11391719af23c18ceb649a64d1692345fde76

storopoli (approved):
ACK 930b4eafb5e9036c98a1d488a18b42058e82acf2

crates/db/store-sled/src/ol_state_index/db.rs
```diff
@@ -0,0 +1,334 @@
...
+                // out-of-order applies regardless of which write families
+                // are populated this call.
+                let mut common = epoch_t.get(&epoch)?.unwrap_or_default();
+                if common
+                    .last_applied_block()
+                    .is_some_and(|prev| block.slot() <= prev.slot())
+                {
+                    return Err(ConflictableTransactionError::Abort(TSledError::abort(
+                        DbError::DuplicateBlockIndexing { epoch, block },
+                    )));
```
krsnapaudel, line 133: Does this hide a bunch of issues (retry, fork, out-of-order, and content mismatch) into one error? Should we check 3 things in each write or append?
1. Entry does not exist. Append.
2. Entry exists and expected. No op.
3. Entry exists and does not match. Error. Then the caller should not absorb this error.

delbonis (approved):
Approved, still have to resolve other comments though.

crates/db/store-sled/src/ol_state_index/db.rs
bewakes: From database's perspective, it is an attempt to index something that's already indexed. This is intentional and the callers can distinguish why this happened(because of restart, fork, etc) based on the context.

crates/db/types/src/ol_state_index.rs
bewakes: Since the key is `(AccountId, Epoch)` this needs to be stored.

irnb (approved):

krsnapaudel (approved):
Not exactly the way I had in mind, but works. LGTM.

PR #1687, opened 2026-04-28
feat(observability): consolidate metrics onto OpenTelemetry pipeline

voidash:
## Summary

This PR consolidates Alpen metrics and tracing onto `strata-logging`'s OpenTelemetry pipeline, consuming strata-common's OTLP meter + `metrics`-crate recorder bridge so all metric sources flow through one OTLP push to the collector.

- Drop the direct `metrics-exporter-prometheus` workspace dependency and `bin/strata`'s `PrometheusBuilder::install()` block.
- Remove `LoggingConfig.metrics_port`; per-binary scrape ports are gone.
- Tie `enable_metrics_layer` to `otlp_url.is_some()` in `bin/strata` and `bin/strata-signer`.
- Bump `opentelemetry` 0.26 -> 0.31 and `tracing-opentelemetry` 0.27 -> 0.32 to align with strata-common.
- Fold in #1689: migrate `bin/alpen-client` from Reth `LogArgs` to `strata-logging`, add `--otlp-url` / `--service-label`, initialize logging inside the Reth tokio runtime context, and flush OTLP buffers on exit.
- Fix `tests/keepalive_stub_test.py`, which overrode `run()` even though flexitest calls `main()`.

## Notes

`metrics-exporter-prometheus` still appears transitively through Reth's `reth-node-metrics`, but Alpen binaries no longer install a separate Prometheus recorder directly.

Deployment chart updates remain out of scope: dropping `--color never`, setting `OTEL_EXPORTER_OTLP_ENDPOINT`, and exposing collector-side Prometheus/ServiceMonitor config live in the deployments follow-up (STR-3193). `bin/alpen-bridge` migration remains in the strata-bridge repo.

AI Assistance Notice: AI used to draft the OTel API migration, the consolidation in `bin/strata/main.rs`, the #1689 fold-in, and this PR description.

PR #1689, opened 2026-04-28
feat(alpen-client): migrate logging from reth CLI to strata-logging

voidash:
## Summary

Drops `LogArgs` from `bin/alpen-client/AdditionalConfig`. Adds `--otlp-url` (env: `OTEL_EXPORTER_OTLP_ENDPOINT`) and `--service-label` flags that map to `strata_logging::LoggingInitConfig`. Builds the tokio runtime via `reth_cli_runner::tokio_runtime()` first, calls `init_logging_from_config(...)` inside `rt.handle().enter()`, then hands the runtime to `CliRunner::from_runtime(rt)`. The OTLP exporters' periodic export tasks need an active tokio handle to spawn against. Wires `strata_logging::finalize()` on the way out so OTLP buffers flush before exit on both success and error paths.

After this lands and the parent [`feat/consolidate-metrics-on-otel`](https://github.com/alpenlabs/vertex-core/tree/feat/consolidate-metrics-on-otel) merges, every `metrics::*!` call inside `bin/alpen-client` flows through the OTel pipeline that [alpenlabs/strata-common#96](https://github.com/alpenlabs/strata-common/pull/96) wires up. That includes reth's ~347 free metrics, every `crates/alpen-ee/*` callsite, `crates/btcio` writer side, and `crates/evmexec`. Spans from this binary also start exporting via OTLP for the first time.

Behavior change worth noting: the `--color` CLI flag is gone (was only used by the fullnode chart with `--color never`; cosmetic). `RUST_LOG` still respected via strata-logging's `EnvFilter`. File logging defaults off.

## Out of scope

- Deployment chart updates (separate PR in `alpenlabs/deployments`). Drops `--color never` from fullnode args, sets `OTEL_EXPORTER_OTLP_ENDPOINT` per cluster pointing at the existing OTel collector DaemonSet on 4317.
- `bin/alpen-bridge` migration. Same shape but in `alpenlabs/strata-bridge` and tracked separately.
- `bin/strata` / `bin/strata-signer` (covered by parent consolidation PR).


voidash:
Folded into #1687; closing this stacked draft so review stays on the combined PR.

PR #1695, opened 2026-04-29
STR 3163 Reject unsigned OL blocks when credential checks are enabled

krsnapaudel:
Reject unsigned non-genesis OL blocks when sequencer credential checks are enabled.

### Problem
`check_ol_block_proposal_valid` previously returned `Ok(())` when a non-genesis `OLBlock` had `signature() == None`, regardless of the configured `CredRule`. Since the caller treats `Ok` as valid, this allowed unsigned blocks to bypass signature verification entirely under `CredRule::SchnorrKey(_)`.

### Changes
- reject missing block signatures in `crates/consensus-logic/src/fcm/service.rs` when `CredRule` is not `Unchecked`
- preserve current behavior for:
  - genesis blocks
  - `CredRule::Unchecked`
- keep invalid provided signatures rejected as before
- add regression tests covering:
  - unsigned block + `Unchecked` -> accepted
  - garbage signature + `Unchecked` -> accepted
  - unsigned block + checked credentials -> rejected
  - invalid signature + checked credentials -> rejected
  - valid signature + checked credentials -> accepted
  - unsigned genesis block -> accepted

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/consensus-logic/src/fcm/service.rs
+++ b/crates/consensus-logic/src/fcm/service.rs
@@ -240,9 +240,10 @@ async fn handle_new_block(fcm_state: &mut FcmState, bundle: &OLBlock) -> anyhow:
 
     // First, decide if the block seems correctly signed and we haven't
     // already marked it as invalid.
-    let check_res = check_ol_block_proposal_valid(blkid, bundle, fcm_state.ctx().params().rollup());
-    if check_res.is_err() {
-        // It's invalid, write that and return.
+    if let Err(err) =
+        check_ol_block_proposal_valid(blkid, bundle, fcm_state.ctx().params().rollup())
+    {
+        warn!(%err, "rejecting block");
         return Ok(false);
     }
 
@@ -365,29 +366,31 @@ async fn handle_epoch_finalization(
 }
 
 /// Checks OL block's credential to ensure that it was authentically proposed.
+///
+/// Slot-0 (genesis) blocks are not expected as proposals — genesis is fixed at node init.
 pub fn check_ol_block_proposal_valid(
     blkid: &OLBlockId,
     block: &OLBlock,
     params: &RollupParams,
-) -> anyhow::Result<()> {
-    // If it's not the genesis block, check that the block is correctly signed.
-    if block.header().slot() > 0 {
-        let Some(sig) = block.signed_header().signature() else {
-            // Just ignore blocks without signature
-            warn!(%blkid, "Received block without signature. ignoring");
-            return Ok(());
-        };
-        let msg: Buf32 = block.header().compute_blkid().into();
-        let is_valid = match params.cred_rule {
-            CredRule::Unchecked => true,
-            CredRule::SchnorrKey(pubkey) => verify_schnorr_sig(sig, &msg, &pubkey),
-        };
-        if !is_valid {
-            warn!(%blkid, "Received block with invalid signature.");
-            return Err(anyhow!("block creds check failed"));
-        }
+) -> Result<(), Error> {
+    if block.header().slot() == 0 {
+        return Err(Error::UnexpectedGenesisBlock(*blkid));
     }
-
+    let sig = match block.signed_header().signature() {
+        Some(sig) => sig,
+        None => match &params.cred_rule {
+            CredRule::Unchecked => return Ok(()),
+            CredRule::SchnorrKey(_) => return Err(Error::MissingBlockSignature(*blkid)),
+        },
+    };
+    let msg: Buf32 = block.header().compute_blkid().into();
+    let is_valid = match &params.cred_rule {
+        CredRule::Unchecked => true,
+        CredRule::SchnorrKey(pubkey) => verify_schnorr_sig(sig, &msg, pubkey),
+    };
+    if !is_valid {
+        return Err(Error::InvalidBlockSignature(*blkid));
+    }
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


crates/consensus-logic/src/fcm/service.rs
```diff
@@ -372,15 +372,20 @@ pub fn check_ol_block_proposal_valid(
...
-            // Just ignore blocks without signature
-            warn!(%blkid, "Received block without signature. ignoring");
-            return Ok(());
+        let sig = match block.signed_header().signature() {
+            Some(sig) => sig,
+            None => match &params.cred_rule {
+                CredRule::Unchecked => return Ok(()),
+                CredRule::SchnorrKey(_) => {
+                    warn!(%blkid, "Received unsigned block while credential checks are enabled.");
+                    return Err(anyhow!("missing block signature"));
```
delbonis: Add typed error variant, maybe with blkid too.

crates/consensus-logic/src/fcm/service.rs
```diff
@@ -372,15 +372,20 @@ pub fn check_ol_block_proposal_valid(
...
+                }
+            },
         };
         let msg: Buf32 = block.header().compute_blkid().into();
-        let is_valid = match params.cred_rule {
+        let is_valid = match &params.cred_rule {
             CredRule::Unchecked => true,
-            CredRule::SchnorrKey(pubkey) => verify_schnorr_sig(sig, &msg, &pubkey),
+            CredRule::SchnorrKey(pubkey) => verify_schnorr_sig(sig, &msg, pubkey),
         };
```
delbonis: Isn't there a `verify_block_credential` fn somewhere we can use that already captures this?

delbonis (changes requested):

prajwolrg (changes requested):
Can we use the value from ASM params' Checkpoint subprotocol's `sequencer_predicate` for this check, instead of using the old `CredRule`. Relevant: https://alpenlabs.atlassian.net/browse/STR-3050

crates/consensus-logic/src/fcm/service.rs
krsnapaudel: @delbonis Apparently there is for the legacy L2 block.

krsnapaudel:
@prajwolrg Are you aware of this?

`verify_claim_witness` rejects every signature produced by today's sequencer.

Why
Both sides claim BIP-340 conformance, but they disagree on what the API-level msg argument is:
`strata_crypto::sign_schnorr_sig / verify_schnorr_sig (secp256k1 backend)`: Message is the 32-byte BIP-340 `m`, used directly in the challenge.
`PredicateKey::verify_claim_witness → k256 VerifyingKey::verify`: SHA-256-hashes the input bytes first, then uses the hash as BIP-340's `m`.
Same 32 input bytes, different `m`s, different challenges, verification fails. Confirmed with a focused probe (same sk, derived x-only pk, same msg, same sig → secp256k1 self-verify true, k256 cross-verify error "signature error").

storopoli:
> @prajwolrg Are you aware of this?
> 
> `verify_claim_witness` rejects every signature produced by today's sequencer.
> 
> Why Both sides claim BIP-340 conformance, but they disagree on what the API-level msg argument is: `strata_crypto::sign_schnorr_sig / verify_schnorr_sig (secp256k1 backend)`: Message is the 32-byte BIP-340 `m`, used directly in the challenge. `PredicateKey::verify_claim_witness → k256 VerifyingKey::verify`: SHA-256-hashes the input bytes first, then uses the hash as BIP-340's `m`. Same 32 input bytes, different `m`s, different challenges, verification fails. Confirmed with a focused probe (same sk, derived x-only pk, same msg, same sig → secp256k1 self-verify true, k256 cross-verify error "signature error").

I'd suggest we move this to a new ticket. In this PR, OL block assembly still derives signing behavior from `params.rollup.cred_rule` in

https://github.com/alpenlabs/alpen/blob/5ef472cc4f13bf5d6363e61c0f70f5e2f5064d8d/crates/ol/block-assembly/src/service.rs#L267

and `RollupParams` still defines `cred_rule` as the rule for block signatures in 

https://github.com/alpenlabs/alpen/blob/5ef472cc4f13bf5d6363e61c0f70f5e2f5064d8d/crates/params/src/lib.rs#L27

storopoli (approved):
ACK 5ef472cc4f13bf5d6363e61c0f70f5e2f5064d8d

krsnapaudel:
@prajwolrg Do you agree with @storopoli? It seems trying to make block assembly use `strata-predicate` will also not fix the issue raised above.

prajwolrg:
> @prajwolrg Are you aware of this?
> 
> `verify_claim_witness` rejects every signature produced by today's sequencer.
> 
> Why Both sides claim BIP-340 conformance, but they disagree on what the API-level msg argument is: `strata_crypto::sign_schnorr_sig / verify_schnorr_sig (secp256k1 backend)`: Message is the 32-byte BIP-340 `m`, used directly in the challenge. `PredicateKey::verify_claim_witness → k256 VerifyingKey::verify`: SHA-256-hashes the input bytes first, then uses the hash as BIP-340's `m`. Same 32 input bytes, different `m`s, different challenges, verification fails. Confirmed with a focused probe (same sk, derived x-only pk, same msg, same sig → secp256k1 self-verify true, k256 cross-verify error "signature error").

I wasn't aware of this. Thanks for flagging this. Since we are using [SPS-51 trick](https://www.notion.so/SPS-51-Generic-simple-envelope-format-1d1901ba000f807f9b36e491d765e786), we never ran into this problem. But we should use the consistent signature format everywhere. 



> @prajwolrg Do you agree with @storopoli? It seems trying to make block assembly use `strata-predicate` will also not fix the issue raised above.

Using a consistent format should fix the issue. But I understand it might be a larger refactor that what the original ticket intended. So I'm fine with it being a separate task as well.

krsnapaudel:
@storopoli @prajwolrg Create [this ticket](https://alpenlabs.atlassian.net/browse/STR-3335) to track the transition to `sequencer_predicate`. Apparently, block assembly is using code from old L2 chain types.

prajwolrg (approved):

crates/consensus-logic/src/fcm/service.rs
```diff
@@ -364,29 +366,31 @@ async fn handle_epoch_finalization(
...
-        let is_valid = match params.cred_rule {
-            CredRule::Unchecked => true,
-            CredRule::SchnorrKey(pubkey) => verify_schnorr_sig(sig, &msg, &pubkey),
-        };
-        if !is_valid {
-            warn!(%blkid, "Received block with invalid signature.");
-            return Err(anyhow!("block creds check failed"));
-        }
+) -> Result<(), Error> {
+    if block.header().slot() == 0 {
```
bewakes, line 376: This is still returning Ok when slot is 0 without checking anything.

Should OL even be proposing block for slot 0? It is pre-defined at the genesis and there's no reason to propose it later or propagate. But just in case it is propagated, we could possibly check if this matches the genesis. Idk if this makes sense to do here though.

crates/consensus-logic/src/fcm/service.rs
```diff
@@ -512,3 +516,132 @@ async fn revert_ol_state_to_block(
...
+        get_test_schnorr_keys()[0].clone()
+    }
+
+    fn test_rollup_params(cred_rule: CredRule) -> RollupParams {
+        let mut params = gen_params();
+        params.rollup.cred_rule = cred_rule;
+        params.rollup
+    }
+
+    fn test_block(slot: u64, signature: Option<Buf64>) -> OLBlock {
```
bewakes: nit: `test_block` indicates it is doing the test. This is actually constructing so better to rename this as `construct_test_block`.

crates/consensus-logic/src/fcm/service.rs
krsnapaudel: @bewakes Adding an error variant. Genesis block should not show up here.

delbonis (approved):
ok looks good

PR #1698, opened 2026-04-30
fix(alpen-cli): encode correct EE serial in deposit descriptor (+ OL warn on silent drops)

voidash:
Real bridge deposits on staging-v2 never reached the Alpen EE snark account.

**Root cause.** `bin/alpen-cli/src/cmd/deposit.rs` encoded `AccountSerial::zero()` in the DRT destination descriptor (legacy from before genesis reserved serials 0..128 for system accounts). Serial 0 resolves to placeholder `AccountId::zero()`, and `account_processing::process_message` silently discards the funds when the target does not exist. Decoded staging DT/DRT in txid `97bcf8b9...` at L1 5262.

## AI Assistance

Diagnosis, fix selection, test scaffolding, and this description.

### DIFF
--- a/bin/alpen-cli/src/cmd/deposit.rs
+++ b/bin/alpen-cli/src/cmd/deposit.rs
@@ -20,7 +20,7 @@ use rand_core::OsRng;
 use shrex::encode;
 use strata_asm_proto_bridge_v1_txs::deposit_request::DrtHeaderAux;
 use strata_cli_common::errors::{DisplayableError, DisplayedError};
-use strata_identifiers::{AccountSerial, SubjectIdBytes};
+use strata_identifiers::{AccountSerial, SubjectIdBytes, SYSTEM_RESERVED_ACCTS};
 use strata_l1_txfmt::{MagicBytes, ParseConfig};
 use strata_ol_bridge_types::DepositDescriptor;
 use strata_primitives::crypto::even_kp;
@@ -115,9 +115,16 @@ fn prepare_deposit_request(
 
     let alpen_subject_bytes =
         SubjectIdBytes::try_new(alpen_address.to_vec()).expect("must be valid subject bytes");
-    // Legacy: deposit intent supports a single execution environment (zero)
-    let deposit_descriptor = DepositDescriptor::new(AccountSerial::zero(), alpen_subject_bytes)
-        .expect("AccountSerial::zero() is always within valid range");
+    // The Alpen EE is the first user account registered at genesis. System
+    // serials occupy 0..SYSTEM_RESERVED_ACCTS, so it lands at serial 128.
+    // Encoding 0 here resolves to the placeholder system slot and the OL
+    // silently discards the deposit. TODO(STR-XXXX): query the EE serial
+    // from OL params or via RPC instead of hardcoding.
+    let deposit_descriptor = DepositDescriptor::new(
+        AccountSerial::new(SYSTEM_RESERVED_ACCTS),
+        alpen_subject_bytes,
+    )
+    .expect("EE serial is within valid range");
     let header_aux = DrtHeaderAux::new(
         recovery_public_key.serialize(),
         deposit_descriptor.encode_to_varvec(),
--- a/bin/alpen-cli/src/cmd/deposit.rs
+++ b/bin/alpen-cli/src/cmd/deposit.rs
@@ -20,9 +20,10 @@ use rand_core::OsRng;
 use shrex::encode;
 use strata_asm_proto_bridge_v1_txs::deposit_request::DrtHeaderAux;
 use strata_cli_common::errors::{DisplayableError, DisplayedError};
-use strata_identifiers::{AccountSerial, SubjectIdBytes, SYSTEM_RESERVED_ACCTS};
+use strata_identifiers::SubjectIdBytes;
 use strata_l1_txfmt::{MagicBytes, ParseConfig};
 use strata_ol_bridge_types::DepositDescriptor;
+use strata_ol_stf::ALPEN_EE_ACCT_SERIAL;
 use strata_primitives::crypto::even_kp;
 
 use crate::{
@@ -115,16 +116,11 @@ fn prepare_deposit_request(
 
     let alpen_subject_bytes =
         SubjectIdBytes::try_new(alpen_address.to_vec()).expect("must be valid subject bytes");
-    // The Alpen EE is the first user account registered at genesis. System
-    // serials occupy 0..SYSTEM_RESERVED_ACCTS, so it lands at serial 128.
-    // Encoding 0 here resolves to the placeholder system slot and the OL
-    // silently discards the deposit. TODO(STR-2021): query the EE serial
-    // from OL params or via RPC instead of hardcoding.
-    let deposit_descriptor = DepositDescriptor::new(
-        AccountSerial::new(SYSTEM_RESERVED_ACCTS),
-        alpen_subject_bytes,
-    )
-    .expect("EE serial is within valid range");
+    // Encoding the wrong serial here (e.g. 0) resolves to a placeholder
+    // system slot and the OL silently discards the deposit. The constant
+    // pins the OL-side identity of the Alpen EE account.
+    let deposit_descriptor = DepositDescriptor::new(ALPEN_EE_ACCT_SERIAL, alpen_subject_bytes)
+        .expect("EE serial is within valid range");
     let header_aux = DrtHeaderAux::new(
         recovery_public_key.serialize(),
         deposit_descriptor.encode_to_varvec(),
--- a/bin/alpen-cli/src/cmd/deposit.rs
+++ b/bin/alpen-cli/src/cmd/deposit.rs
@@ -118,7 +118,7 @@ fn prepare_deposit_request(
     // The Alpen EE is the first user account registered at genesis. System
     // serials occupy 0..SYSTEM_RESERVED_ACCTS, so it lands at serial 128.
     // Encoding 0 here resolves to the placeholder system slot and the OL
-    // silently discards the deposit. TODO(STR-XXXX): query the EE serial
+    // silently discards the deposit. TODO(STR-2021): query the EE serial
     // from OL params or via RPC instead of hardcoding.
     let deposit_descriptor = DepositDescriptor::new(
         AccountSerial::new(SYSTEM_RESERVED_ACCTS),
--- a/bin/alpen-cli/src/cmd/deposit.rs
+++ b/bin/alpen-cli/src/cmd/deposit.rs
@@ -116,9 +116,6 @@ fn prepare_deposit_request(
 
     let alpen_subject_bytes =
         SubjectIdBytes::try_new(alpen_address.to_vec()).expect("must be valid subject bytes");
-    // Encoding the wrong serial here (e.g. 0) resolves to a placeholder
-    // system slot and the OL silently discards the deposit. The constant
-    // pins the OL-side identity of the Alpen EE account.
     let deposit_descriptor = DepositDescriptor::new(ALPEN_EE_ACCT_SERIAL, alpen_subject_bytes)
         .expect("EE serial is within valid range");
     let header_aux = DrtHeaderAux::new(
--- a/bin/alpen-cli/src/cmd/deposit.rs
+++ b/bin/alpen-cli/src/cmd/deposit.rs
@@ -23,12 +23,11 @@ use strata_cli_common::errors::{DisplayableError, DisplayedError};
 use strata_identifiers::SubjectIdBytes;
 use strata_l1_txfmt::{MagicBytes, ParseConfig};
 use strata_ol_bridge_types::DepositDescriptor;
-use strata_ol_stf::ALPEN_EE_ACCT_SERIAL;
 use strata_primitives::crypto::even_kp;
 
 use crate::{
     alpen::AlpenWallet,
-    constants::SIGNET_BLOCK_TIME,
+    constants::{ALPEN_EE_ACCT_SERIAL, SIGNET_BLOCK_TIME},
     link::{OnchainObject, PrettyPrint},
     recovery::DescriptorRecovery,
     seed::Seed,
--- a/crates/ol/stf/src/account_processing.rs
+++ b/crates/ol/stf/src/account_processing.rs
@@ -36,8 +36,12 @@ pub(crate) fn process_message<S: IStateAccessorMut>(
 
             // Check if the account exists first.
             if !state.check_account_exists(target)? {
-                // If we don't find it then we can sweep it into limbo.
-                warn!(%target, "target account does not exist");
+                warn!(
+                    %target,
+                    %sender,
+                    value = %msg.value(),
+                    "limboing message to nonexistent target account",
+                );
                 handle_misplaced_funds(state, coin)?;
                 return Ok(());
             }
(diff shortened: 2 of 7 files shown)


bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -115,9 +115,16 @@ fn prepare_deposit_request(
 
     let alpen_subject_bytes =
         SubjectIdBytes::try_new(alpen_address.to_vec()).expect("must be valid subject bytes");
-    // Legacy: deposit intent supports a single execution environment (zero)
-    let deposit_descriptor = DepositDescriptor::new(AccountSerial::zero(), alpen_subject_bytes)
-        .expect("AccountSerial::zero() is always within valid range");
+    // The Alpen EE is the first user account registered at genesis. System
+    // serials occupy 0..SYSTEM_RESERVED_ACCTS, so it lands at serial 128.
+    // Encoding 0 here resolves to the placeholder system slot and the OL
+    // silently discards the deposit. TODO(STR-XXXX): query the EE serial
```
storopoli: blocker for the STR ticket number.

crates/ol/stf/src/account_processing.rs
```diff
@@ -34,8 +34,19 @@ pub(crate) fn process_message<S: IStateAccessor>(
...
             // Check if the account exists first.
             if !state.check_account_exists(target)? {
-                // If we don't find it then we can just ignore it.
-                // TODO do something with the funds we're throwing away by doing this
+                // Funds are silently dropped. This commonly trips when a deposit
+                // descriptor encodes an unregistered or system-reserved serial,
+                // which `find_account_id_by_serial` resolves to `AccountId::zero()`
+                // rather than returning None. Surface it so it does not go
+                // unnoticed in production.
+                // TODO(STR-XXXX): track these on a counter and decide whether to
```
storopoli: blocker for the STR ticket number.

storopoli: Actually use https://alpenlabs.atlassian.net/browse/STR-2021

storopoli:
```suggestion
                // TODO(STR-2021): track these on a counter and decide whether to
```

crates/ol/stf/src/account_processing.rs
```diff
@@ -78,8 +89,13 @@ pub(crate) fn process_transfer<S: IStateAccessor>(
         _ => {
             // Check if the account exists first.
             if !state.check_account_exists(target)? {
-                // If we don't find it then we can just ignore it.
-                // TODO: do something with the funds we're throwing away by doing this
+                // Funds silently dropped. See `process_message` for context.
+                // TODO(STR-XXXX): counter + refund/burn policy.
```
storopoli:
```suggestion
                // TODO(STR-2021): counter + refund/burn policy.
```

functional-tests/tests/ol_isolated/test_deposit_unregistered_serial.py
```diff
@@ -0,0 +1,101 @@
...
+        )
+        logger.info("mock deposit broadcast txid=%s", txid)
+
+        addr = btc_rpc.proxy.getnewaddress()
+        btc_rpc.proxy.generatetoaddress(8, addr)
+
+        # Cross at least one terminal block so the deposit-bearing manifest is
+        # processed by the OL STF.
+        slots_per_epoch = strata.props["slots_per_epoch"]
+        strata.wait_for_additional_blocks(2 * slots_per_epoch, rpc, timeout_per_block=15)
```
storopoli: we should probably have a wait for checkpoint or wait for terminal something like this. Cc @delbonis

storopoli (changes requested):
looks good some placeholder blockers on the `TODO(STR-XXXX)` stuff

functional-tests/tests/ol_isolated/test_deposit_unregistered_serial.py
```diff
@@ -0,0 +1,121 @@
...
+    return summaries[0]["balance"]
+
+
+@flexitest.register
+class TestDepositUnregisteredSerial(StrataNodeTest):
+    def __init__(self, ctx: flexitest.InitContext):
+        ctx.set_env("ol_isolated")
+
+    def main(self, ctx):
+        from common.test_cli import create_mock_deposit
```
bewakes: Any reason to have the import statement here?

functional-tests/tests/ol_isolated/test_deposit_unregistered_serial.py
```diff
@@ -0,0 +1,121 @@
...
+            current_terminal = int(rpc.strata_getChainStatus()["latest_terminal"]["epoch"])
+            if current_terminal > start_terminal:
+                logger.info(
+                    "terminal epoch advanced %d -> %d",
+                    start_terminal,
+                    current_terminal,
+                )
+                break
+        else:
+            raise AssertionError(f"terminal epoch did not advance past {start_terminal} within 60s")
```
bewakes: We should use some sort of `wait_until*` here. The passed lambda can do the mining.

functional-tests/tests/ol_isolated/test_deposit_unregistered_serial.py
```diff
@@ -0,0 +1,121 @@
...
+
+        balance = get_account_balance(rpc, TEST_ACCOUNT_ID_HEX)
+        delta = balance - initial
+        if delta != 0:
+            raise AssertionError(
+                f"deposit with unregistered serial must not credit any account, "
```
bewakes: nit: on "does not credit any account", this only checks that `TX_ACCOUNT_ID_HEX` is not created, not none of the accounts in OL.

bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -115,9 +115,16 @@ fn prepare_deposit_request(
...
         SubjectIdBytes::try_new(alpen_address.to_vec()).expect("must be valid subject bytes");
-    // Legacy: deposit intent supports a single execution environment (zero)
-    let deposit_descriptor = DepositDescriptor::new(AccountSerial::zero(), alpen_subject_bytes)
-        .expect("AccountSerial::zero() is always within valid range");
+    // The Alpen EE is the first user account registered at genesis. System
+    // serials occupy 0..SYSTEM_RESERVED_ACCTS, so it lands at serial 128.
+    // Encoding 0 here resolves to the placeholder system slot and the OL
+    // silently discards the deposit. TODO(STR-2021): query the EE serial
+    // from OL params or via RPC instead of hardcoding.
+    let deposit_descriptor = DepositDescriptor::new(
```
bewakes: I am curious if this also solves the inconsistently created descriptor or aux_data for deposit. Another reason the deposits did not go into OL was that the descriptor could not be decoded in OL.
cc: @Rajil1213

bewakes (changes requested):
Some nits and suggestions.

functional-tests/tests/ol_isolated/test_deposit_unregistered_serial.py
storopoli: Yes, I think I've added that comment somewhere in this PR in a previous review.

bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -115,9 +115,16 @@ fn prepare_deposit_request(
...
+        AccountSerial::new(SYSTEM_RESERVED_ACCTS),
```
storopoli: I'd really prefer if we add a `ALPEN_EE_REF` in https://github.com/alpenlabs/alpen/blob/9afcf3bb42d9a9a70f02fcec4cb11a1b70605984/crates/ol/stf/src/constants.rs#L1-L17 and then import it here.
We can reuse that constant somewhere and it is a much better eng best practice to not hard-code constants in binaries.

storopoli (changes requested):
Agree with @bewakes very good review comments.

bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -115,9 +115,16 @@ fn prepare_deposit_request(
 
     let alpen_subject_bytes =
         SubjectIdBytes::try_new(alpen_address.to_vec()).expect("must be valid subject bytes");
-    // Legacy: deposit intent supports a single execution environment (zero)
-    let deposit_descriptor = DepositDescriptor::new(AccountSerial::zero(), alpen_subject_bytes)
-        .expect("AccountSerial::zero() is always within valid range");
+    // The Alpen EE is the first user account registered at genesis. System
+    // serials occupy 0..SYSTEM_RESERVED_ACCTS, so it lands at serial 128.
+    // Encoding 0 here resolves to the placeholder system slot and the OL
```
prajwolrg: This isn't really necessary.

bin/alpen-cli/src/cmd/deposit.rs
prajwolrg: I was thinking of adding `AccountSerial::alpen()`, but it didn't make much sense to add it to the `strata-common` repo.

prajwolrg: We should definitely use the constant, but I'm not sure if this is the right place. I think the right place is somewhere in the `ee`* related crate.

prajwolrg (changes requested):

bin/alpen-cli/src/cmd/deposit.rs
storopoli: > We should definitely use the constant, but I'm not sure if this is the right place. I think the right place is somewhere in the `ee`* related crate.


I disagree.
It is an OL Snark Account constant.
AFAIK if we change the `AccountSerial` of Alpen mid-flight, EE would care less.
However, OL would totally break.

delbonis (approved):
Approving this now but actually the limbo funds PR #1701 will address the silent drops.

functional-tests/tests/ol_isolated/test_deposit_unregistered_serial.py
voidash: Addressed. The test now polls for the OL tip epoch to advance after the deposit lands on L1, rather than waiting a fixed slot count.

functional-tests/tests/ol_isolated/test_deposit_unregistered_serial.py
voidash: Addressed. The test now uses `wait_until_with_value`; the polling closure mines L1 blocks and returns the current OL tip epoch until it advances.

functional-tests/tests/ol_isolated/test_deposit_unregistered_serial.py
voidash: Addressed. The local import is gone; `create_mock_deposit` is imported at module scope now.

bin/alpen-cli/src/cmd/deposit.rs
voidash: Addressed. `ALPEN_EE_ACCT_SERIAL` now lives in `crates/ol/stf/src/constants.rs` and `alpen-cli` imports that constant instead of hard-coding the serial in the binary.

bin/alpen-cli/src/cmd/deposit.rs
voidash: This fixes the descriptor serial bug specifically: the CLI now builds the deposit descriptor with the OL-side Alpen EE serial constant, and it still uses `DepositDescriptor::new(...).encode_to_varvec()`, which is the same typed format the OL decodes. If there is a separate aux-data encoding mismatch, that would be a separate issue from the serial=0 bug.

bin/alpen-cli/src/cmd/deposit.rs
voidash: Addressed. I removed the redundant deposit serial explanation from `deposit.rs`; the imported `ALPEN_EE_ACCT_SERIAL` constant now carries the intent.

functional-tests/tests/ol_isolated/test_deposit_unregistered_serial.py
```diff
@@ -0,0 +1,119 @@
...
+            btc_rpc.proxy.generatetoaddress(2, addr)
+            return int(rpc.strata_getChainStatus()["tip"]["epoch"])
+
+        current_epoch = wait_until_with_value(
+            mine_and_get_tip_epoch,
+            lambda epoch: epoch > start_epoch,
+            error_with=f"tip epoch did not advance past {start_epoch} within 60s",
+            timeout=60,
+            step=0.5,
+        )
```
bewakes, line 123: Might be a good idea to sanity check the mock deposit tx is actually confirmed deep enough. Not a blocker though.

bewakes (approved):

bin/alpen-cli/src/cmd/deposit.rs
prajwolrg: > It is an OL Snark Account constant.

@storopoli I disagree.
A bridge account being treated as an OL Snark Account constant makes sense because it is conceptually tied to that role. But `ALPEN_EE_ACCT_SERIAL` is specifically tied to Alpen being the first account registered after the reserved system range. 
The current naming/commenting implies that this serial is inherently associated with the OL Snark Account itself, whereas in reality it is just a consequence of registration order at genesis. If tomorrow we deployed a different OL Snark Account first, this constant would either become incorrect or force us to preserve this serial assignment purely because the name semantically “reserves” it for Alpen.

> AFAIK if we change the AccountSerial of Alpen mid-flight, EE would care less.

That is correct. It is related with OL more in that sense, but I still think shouldn't be the constant of the OL system.

----

Note: This is not a blocking comment.

prajwolrg (approved):

bin/alpen-cli/src/cmd/deposit.rs
bewakes: I agree with @prajwolrg . OL doesn't really care about what ALPEN_EE_ACCT_SERIAL is. In fact, OL doesn't even know if Alpen exists or not. It only makes sense to have OL special account constants in OL.
So, it shouldn't be in `ol/stf/src/constants.rs`. This is not like BRIDGE_GATEWAY_ACCOUNT.

crates/ol/stf/src/constants.rs
```diff
@@ -15,3 +16,11 @@ pub const SEQUENCER_ACCT_ID: AccountId = BRIDGE_GATEWAY_ACCT_ID;
 /// Serial of the bridge gateway account.
 // TODO make this different, really, it should be the sequencer producing the block
 pub const SEQUENCER_ACCT_SERIAL: AccountSerial = BRIDGE_GATEWAY_ACCT_SERIAL;
+
+/// Serial of the Alpen EE snark account.
+///
+/// System serials occupy `0..SYSTEM_RESERVED_ACCTS`, so the Alpen EE
+/// account, registered first at genesis, lands at serial
+/// `SYSTEM_RESERVED_ACCTS` (currently 128).
+// TODO(STR-2021): query this from OL params or via RPC at runtime.
```
prajwolrg: STR-2021 is for a different issue. I don't think it's related with this.

prajwolrg (changes requested):

crates/ol/stf/src/constants.rs
bewakes: Let's just move this elsewhere. There's no reason to have ALPEN related constants here.

storopoli (approved):
ACK db3e7f890a389391b074124b278f4fccb6884b0c

storopoli:
@voidash please rebase

prajwolrg (approved):

PR #1699, opened 2026-04-30
test(bridge): real-bridge E2E deposit + bridgeout fn-test

voidash:
First bridge fn-test that drives the loop end to end through the production subprotocol path. Previous bridge fn-tests stop at the OL boundary (debug subprotocol or "SAU landed in mempool" assertions). This one runs real DRT + DT through `strata-test-cli`, snark account credit, EVM bridgeout via the precompile, WF tx confirmed on regtest paying 9.5 BTC out to the recipient address.

## Stacked on

[#1746](https://github.com/alpenlabs/alpen/pull/1746). Production-code bug fixes that unblocked the loop live there. The diff on this PR is the test, the `el_ol_bridge` envconfig, and harness helpers.

[STR-3095]: https://alpenlabs.atlassian.net/browse/STR-3095?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bin/strata-test-cli/src/bridge/dt.rs
```diff
@@ -27,29 +27,26 @@ use strata_primitives::{buf::Buf32, constants::RECOVER_DELAY};
...
-    parse::{generate_taproot_address, parse_operator_keys},
+    parse::generate_taproot_address,
 };
 
 /// Creates a deposit transaction (DT)
 ///
 /// # Arguments
 /// * `tx_bytes` - Raw DRT transaction bytes
-/// * `operator_keys` - Vector of operator secret keys as bytes (78 bytes each)
+/// * `signers` - Operator private keys (already parsed)
```
storopoli: I think that other matters here right @Rajil1213? If yes, we should add a warning in the docstring here.

bin/strata-test-cli/src/cmd/compute_drt_output.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[argh(option)]
+    pub recovery_pubkey: String,
+
+    /// alpen EE address (20-byte EVM address), hex
+    #[argh(option)]
+    pub alpen_address: String,
+
+    /// destination account serial (default: SYSTEM_RESERVED_ACCTS = 128)
+    #[argh(option, default = "SYSTEM_RESERVED_ACCTS")]
+    pub account_serial: u32,
```
storopoli: This is incorrect. 0-127 are system reserved accounts. `128` is the first non-reserved, AKA Alpen!

bin/strata-test-cli/src/cmd/compute_drt_output.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    println!("{result}");
+    Ok(())
+}
+
+fn parse_network(s: &str) -> Result<Network, DisplayedError> {
+    match s {
+        "regtest" => Ok(Network::Regtest),
+        "signet" => Ok(Network::Signet),
+        "testnet" => Ok(Network::Testnet),
```
storopoli: "testnet" should be `Network::Testnet` or `Network::Testnet4`?
I think that the old testnet is pretty much deprecated and will be sunsetted soon.
So, according to https://docs.rs/bitcoin/latest/bitcoin/enum.Network.html


```suggestion
        "testnet" => Ok(Network::Testnet4),
```

This kind of issue would not have happened if we had an inner "network" type and we impl a `From<bitcoin::Network>` for that type because the matching would have to be exhaustive.

bin/strata-test-cli/src/parse.rs
```diff
@@ -7,24 +9,26 @@ use strata_primitives::buf::Buf32;
 
 use crate::error::Error;
 
-/// Parses operator EvenSecretKey from extended private key bytes
-///
-/// Takes raw bytes representing extended private keys and derives
-/// operator keys using the standard key derivation path.
+/// Parses operator [`EvenSecretKey`]s from a list of strings, accepting either
+/// the BIP32 base58 form (`tprv8...` / `xprv...`) emitted by `datatool genxpriv`
```
storopoli:
```suggestion
/// the BIP32 base58 form (`tprv...` / `xprv...`) emitted by `datatool genxpriv`
```

why the `8`?

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -281,14 +281,20 @@ impl EpochDaAccumulator {
...
+            // break the OL checkpoint cycle on every real deposit. Skip the
+            // existence check for special sources; they are constants that
+            // the DA proof side recognises directly.
+            if !source_id.is_special() {
+                let exists = state
+                    .check_account_exists(source_id)
+                    .map_err(|_| DaAccumulationError::MessageSourceMissing(source_id))?;
+                if !exists {
+                    return Err(DaAccumulationError::MessageSourceMissing(source_id));
+                }
```
storopoli: This is exactly #1700 so probably need to rebase after that.

functional-tests/common/bridge.py
```diff
@@ -0,0 +1,204 @@
...
+logger = logging.getLogger(__name__)
+
+
+# Genesis registers the Alpen EE as the first user account, which lands at
+# serial 128 (system serials occupy 0..128).
+ALPEN_EE_ACCOUNT_SERIAL = 128
+
+# DT input value (DRT output 1) must exceed `denomination` so the DT has a
+# positive mining fee. The bridge subprotocol enforces DT output 1 ==
+# denomination (10 BTC); the difference is the miner fee. 1000 sats covers a
```
storopoli: Are we still using 10 BTC as denomination in tests?

functional-tests/common/bridge.py
```diff
@@ -0,0 +1,204 @@
...
+        For SPS-50 DRT payloads (~60 bytes), bitcoind's `createrawtransaction`
+        wants the data portion only and wraps it in OP_RETURN itself. The first
+        two bytes of `op_return_hex` are 0x6a (OP_RETURN) and a single
+        push-length byte; this stripping is correct for payloads <= 75 bytes.
+        """
+        if len(self.op_return_hex) < 4:
+            raise ValueError(f"op_return_hex too short: {self.op_return_hex!r}")
+        if self.op_return_hex[:2] != "6a":
+            raise ValueError(f"expected OP_RETURN (0x6a) prefix, got {self.op_return_hex[:2]!r}")
+        return self.op_return_hex[4:]
```
storopoli, line 77: This is extremely brittle and will only work for 75 bytes, which is fine for now. Anything over 75 bytes will have two or more bytes `OP_PUSHDATA1` followed by the varint size of the payload.

You should do an additional guard here if the hex payload is `<=75` bytes.

functional-tests/common/bridge.py
```diff
@@ -0,0 +1,204 @@
...
+
+def random_xonly_pubkey_hex() -> str:
+    """Generate a random 32-byte x-only pubkey for the DRT recovery path.
+
+    The recovery path is never exercised in happy-path tests, so any byte
+    string that parses as a valid x-only pubkey works. We pick a value
+    deterministically via `secrets` so test runs vary, which is fine because
+    the field is not consumed in the credit path.
+    """
+    return secrets.token_hex(32)
```
storopoli: How can you guarantee that a 32-byte random hex is a valid x-coord of a point in secp256k1, i.e. a valid X-only PK?

functional-tests/entry.py
```diff
@@ -310,6 +310,9 @@ def main(argv: list[str]) -> int:
         ),
         # Environments containing both ee and ol
         "el_ol": EeOLEnv(pre_generate_blocks=110),
+        # Same as `el_ol` but with a fast OL block time, so bridge tests can
+        # drive a deposit -> bridgeout -> WF cycle within reasonable runtime.
+        "el_ol_bridge": EeOLEnv(pre_generate_blocks=110, ol_block_time_ms=500),
```
storopoli: 0.5s OL block time will definitely be very flaky fn tests.
@purusang you've had your fingers burned with trying to reduce block times in fn tests right?

functional-tests/tests/alpen_client/test_real_bridge_deposit_withdraw.py
```diff
@@ -0,0 +1,479 @@
...
+        # `extract_and_validate_withdrawal_intents` produces no
+        # `BridgeIncomingMsg::DispatchWithdrawal` for the bridge subprotocol.
+        # Result: when we broadcast a WF tx the bridge rejects validation with
+        # `WithdrawalValidationError::NoAssignmentFound`.
+        #
+        # Both gaps are upstream of this PR. We DO broadcast a real WF tx so
+        # the bitcoin-layer half is exercised and visible on chain. The
+        # `Fulfilled withdrawal assignment` assertion is the right canonical
+        # signal once both gaps are fixed; for now we only assert WF
+        # confirmation on bitcoin.
```
storopoli: I appreciate the narrative (and hardcoded git blame) but I don't think we should leave this huge comment here.

functional-tests/tests/alpen_client/test_real_bridge_deposit_withdraw.py
```diff
@@ -0,0 +1,479 @@
...
+    The strata factory generates this file before booting the node and uses
+    its contents to populate the bridge subprotocol's genesis operator set.
+    Reading the same file means the keys we sign DTs with always match the
+    on-chain bridge state.
+    """
+    datadir = Path(strata_service.props["datadir"])
+    path = datadir / OPERATOR_KEYS_FILENAME
+    if not path.exists():
+        raise RuntimeError(f"operator key file not found: {path}")
+    return [path.read_text().strip()]
```
storopoli: This silently truncates. returns `[file.read_text().strip()]` regardless of file contents.

storopoli (changes requested):

functional-tests/common/bridge.py
storopoli: Solved thanks

crates/alpen-ee/ol-tracker/src/task.rs
```diff
@@ -81,7 +88,7 @@ pub(crate) struct OLEpochOperations {
 pub(crate) enum TrackOLAction {
     /// Extend local view of the OL chain with new epochs.
     /// TODO: stream
-    Extend(Vec<OLEpochOperations>, OLChainStatus),
+    Extend(Vec<OLEpochOperations>, Box<OLChainStatus>),
```
bewakes, line 91: Why is this `Box`ed now?

crates/alpen-ee/ol-tracker/src/task.rs
```diff
@@ -92,29 +99,51 @@ pub(crate) async fn track_ol_state(
...
 
-    let best_ol_confirmed = &ol_status.confirmed;
-    let best_ol_epoch = best_ol_confirmed.epoch();
+    // Pick the chain-tip signal to advance against. `confirmed` is the
+    // canonical (CSM-based, requires L1 checkpoint observation) signal.
+    //
+    // The dev/test path advances against the latest terminal OL epoch that
+    // Strata has processed locally. This lets the EE block builder consume OL
+    // inbox messages without waiting for the checkpoint observation cadence,
+    // which can lag badly when the native noop prover emits SAUs quickly.
```
bewakes: In dev/test, can't we tune L1 block time so that the lag becomes minimal without having to introduce `track_finalized_epoch`?

crates/alpen-ee/sequencer/src/block_builder/config.rs
```diff
@@ -2,12 +2,13 @@ use std::num::NonZeroU8;
 
 use strata_acct_types::AccountId;
 
+const BRIDGE_GATEWAY_REF: u8 = 0x10;
```
bewakes: Please reference [this ticket](https://alpenlabs.atlassian.net/browse/STR-3315) and add a TODO here if you think this constant should not be duplicated here.

crates/evm-ee/src/execution.rs
```diff
@@ -21,16 +21,17 @@ use strata_ee_acct_types::{
...
 
 use crate::{
     types::{EvmBlock, EvmHeader, EvmPartialState, EvmWriteBatch},
     utils::{build_and_recover_block, compute_hashed_post_state, validate_deposits_against_block},
 };
 
 /// Address where withdrawal intent msgs are forwarded.
-//FIXME: should be set with real bridge gateway account
-const BRIDGE_GATEWAY_ACCOUNT: [u8; 32] = [1u8; 32];
+const BRIDGE_GATEWAY_REF: u8 = 0x10;
```
bewakes, line 36: Duplication of the constant.

crates/ol/block-assembly/src/block_assembly.rs
```diff
@@ -440,15 +440,51 @@ where
...
+                TransactionPayload::SnarkAccountUpdate(payload) => {
+                    let operation = payload.operation();
+                    (
+                        Some(operation.update().seq_no()),
+                        Some(operation.update().proof_state().new_next_msg_idx()),
+                        Some(operation.messages_iter().count()),
+                    )
+                }
+                TransactionPayload::GenericAccountMessage(_) => (None, None, None),
+            };
```
bewakes: Looks like this block is getting values for emitting logs. I suggest to separate this block out into a function.

bewakes: The 3-tuple can also be replaced by a type - something like `SauSummary`.

crates/alpen-ee/ol-tracker/src/task.rs
delbonis: I assume because it's gotten to be a large type layout and this makes the enum smaller.

crates/alpen-ee/ol-tracker/src/task.rs
delbonis: Yeah we probably want to keep the behavior closer to how it would work in a real world environment.

crates/ol/rpc/types/src/chain_status.rs
```diff
@@ -62,6 +62,9 @@ pub struct RpcOLChainStatus {
 
     /// Finalized epoch commitment.
     pub finalized: EpochCommitment,
+
+    /// Latest epoch terminal block processed by the OL fork-choice manager.
+    pub latest_terminal: EpochCommitment,
```
delbonis: Would be best to call this just `latest`.  Strictly speaking this isn't the terminal, it's an epoch commitment which *contains* the terminal block info.  It's important to be careful about this in RPC interfaces.

delbonis (changes requested):
Concur with some of Bibek's comments.

Also, those tests are rather large, is there any way we can consolidate some of the common behavior?

delbonis (changes requested):
Concur with some of Bibek's comments.

Also, those tests are rather large, is there any way we can consolidate some of the common behavior?

crates/alpen-ee/ol-tracker/src/task.rs
voidash: Kept the dev/test flag and tightened the code comment. The issue this avoids is the CSM/checkpoint-observation path lagging behind rapid noop-prover SAU emission; shortening regtest L1 block time alone does not remove that pipeline lag. Production default behavior still tracks the canonical confirmed epoch.

functional-tests/common/bridge.py
voidash: Addressed. The helper no longer says the denomination is hard-coded to 10 BTC; it uses the DRT amount returned by `compute-drt-output`, which comes from the configured bridge params, and the comment now says “configured bridge denomination”.

crates/alpen-ee/ol-tracker/src/task.rs
voidash: Addressed. I removed the `Box`; `TrackOLAction::Extend` now carries `OLChainStatus` directly.

crates/evm-ee/src/execution.rs
voidash: Addressed as a tracked follow-up rather than adding a new dependency edge here. The constant now sends withdrawals to the real bridge gateway account, and the TODO now points at STR-3358 and the duplicated default bridge gateway account in the sequencer config without referencing a stale symbol.

crates/ol/rpc/types/src/chain_status.rs
voidash: Addressed. The RPC field is now `latest`, with docs describing it as the latest epoch terminal block processed by OL fork-choice.

crates/ol/block-assembly/src/block_assembly.rs
voidash: Addressed in the current diff. The log-value extraction is split into `sau_summary_for_logs`, and the tuple has been replaced with `SauSummary`.

storopoli:
@codex Please review this PR

functional-tests/tests/alpen_client/test_real_bridge_deposit_withdraw.py
```diff
@@ -0,0 +1,693 @@
...
+            recipient_bosd_hex=recipient_bosd_hex,
+            withdraw_sats=bridge_denom_sats,
+            miner_addr=miner_addr,
+            ee_output_log_offset=ee_output_log_offset,
+        )
+
+        saw_update_at_epoch = wait_for_account_update_seq(
+            strata_rpc,
+            ALPEN_ACCOUNT_ID,
+            min_next_seq_no=submitted_seq_no + 1,
```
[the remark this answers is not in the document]
storopoli: @voidash this seems a legit/good finding. Please take a look.

storopoli (dismissed):
ACK 1315a5bc5a0cf60804bbf5c71950c2b59670c635

My previous blockers were addressed. But Codex has found something that seems serious and CI is still not green.

bin/strata-test-cli/src/error.rs
```diff
@@ -23,6 +23,30 @@ pub(crate) enum Error {
 
     #[error("Invalid BitcoinD response")]
     BitcoinD,
+
+    #[error("Transaction builder error: {0}")]
+    TxBuilder(String),
+
+    #[error("Transaction parser error: {0}")]
+    TxParser(String),
```
delbonis: `error` is redundant in error strings

bin/strata-test-cli/src/parse.rs
```diff
@@ -38,20 +41,20 @@ pub(crate) fn parse_operator_keys(operator_keys: &[[u8; 78]]) -> Result<Vec<Even
...
-        aggregate_schnorr_keys(operator_wallet_pks.iter()).context("failed to aggregate keys")?;
+    let x_only_pub_key = aggregate_schnorr_keys(operator_wallet_pks.iter())
+        .map_err(|e| Error::TxBuilder(format!("Failed to aggregate keys: {}", e)))?;
 
     // Build the taproot spending tree (empty tree in this case)
     let taproot_builder = TaprootBuilder::new();
     let spend_info = taproot_builder
         .finalize(SECP256K1, x_only_pub_key)
-        .map_err(|_| anyhow::anyhow!("taproot finalization failed"))?;
+        .map_err(|_| Error::TxBuilder("Taproot finalization failed".to_string()))?;
```
delbonis: Can we make these into error variants?

crates/alpen-ee/common/src/traits/ol_client.rs
```diff
@@ -142,10 +144,14 @@ pub enum OLClientError {
     #[error("unexpected inbox message count: expected {expected} message lists, got {actual}")]
     UnexpectedInboxMessageCount { expected: usize, actual: usize },
 
-    /// Chain status slots are not in the correct order (tip >= confirmed >= finalized).
-    #[error("unexpected chain status slot order: {tip} >= {confirmed} >= {finalized}")]
+    /// Chain status slots are not in the correct order
+    /// (tip >= latest terminal >= confirmed >= finalized).
+    #[error(
+        "unexpected chain status slot order: tip={tip}, latest={latest}, confirmed={confirmed}, finalized={finalized}"
+    )]
```
delbonis: Colons imply a new level of hierarchy, should be formatted like `unexpected chain status slot order (tip {tip}, latest {latest}, confirmed {confirmed}, finalized {finalized})`

Should also change other error strings here.

crates/ol/block-assembly/src/block_assembly.rs
```diff
@@ -656,6 +698,32 @@ where
...
+            let operation = p.operation();
+            Some(SauSummary {
+                seq_no: operation.update().seq_no(),
+                new_next_msg_idx: operation.update().proof_state().new_next_msg_idx(),
+                processed_message_count: operation.messages_iter().count(),
+            })
+        }
+        TransactionPayload::GenericAccountMessage(_) => None,
+    }
+}
```
delbonis: Can we make this be a `from_tx_payload` fn on `SauSummary`?

bin/alpen-client/src/prover/spec_acct.rs
```diff
@@ -140,54 +146,80 @@ impl ProofSpec for AcctSpec {
...
+                            "no EE account state available yet (genesis not loaded?)".to_string(),
+                        )
+                    })?;
+                let fallback = acct_at_epoch.ee_state().clone();
+                if fallback.last_exec_blkid() != parent_blkid {
+                    return Err(PaasError::TransientFailure(format!(
+                        "EE pre-state for batch {batch_id} unavailable: \
+                         no exec record for parent={parent_blkid:?}, \
+                         and OL-accepted tip={:?} doesn't match",
+                        fallback.last_exec_blkid(),
```
delbonis: Can we make this into typed errors?

bin/alpen-client/src/prover/spec_acct.rs
```diff
@@ -321,6 +351,64 @@ impl ProofSpec for AcctSpec {
     }
 }
 
+fn da_refs_from_status(batch_id: BatchId, status: BatchStatus) -> ProverResult<Vec<L1DaBlockRef>> {
+    match status {
+        BatchStatus::DaComplete { da }
+        | BatchStatus::ProofPending { da }
+        | BatchStatus::ProofReady { da, .. } => Ok(da),
+        BatchStatus::Genesis | BatchStatus::Sealed | BatchStatus::DaPending { .. } => Err(
+            PaasError::TransientFailure(format!("batch {batch_id} does not have DA refs yet")),
```
delbonis: also here

bin/alpen-client/src/prover/spec_acct.rs
```diff
@@ -321,6 +351,64 @@ impl ProofSpec for AcctSpec {
...
+            let height = da_ref.block.height();
+            let hash = l1_header_commitments.get(&height).copied().ok_or_else(|| {
+                PaasError::TransientFailure(format!(
+                    "missing L1 header commitment for height {height}"
+                ))
+            })?;
+            let mmr_idx = (height as u64).checked_sub(mmr_offset).ok_or_else(|| {
+                PaasError::PermanentFailure(format!(
+                    "L1 height {height} is before MMR start offset {mmr_offset}"
+                ))
```
delbonis: And these.

delbonis (dismissed):
Generally very good, good docs.  Just some comments about error handling/style.

bin/strata-test-cli/src/error.rs
voidash: @delbonis Addressed in 0b0606a47. The transaction builder/parser variants now render as `transaction builder: ...` and `transaction parser: ...`, without redundant `error` wording.

storopoli (approved):
ACK 0b0606a47c9746e5c38ae8dedf0664ed5717581b

crates/proof-impl/alpen-chunk/src/lib.rs
```diff
@@ -33,3 +33,24 @@ pub fn process_ee_chunk(zkvm: &impl ZkVmEnvSerde) {
...
+/// Wired in `bin/alpen-client/src/main.rs` behind the
+/// `--dev-native-noop-prover` flag, which itself only takes effect when
+/// `--dev-native-prover` is also set.
+pub fn process_ee_chunk_noop(zkvm: &impl ZkVmEnvSerde) {
+    let _genesis: Genesis = zkvm.read_serde();
+    let buf = zkvm.read_buf();
+    let input: &ArchivedPrivateInput = rkyv::access::<ArchivedPrivateInput, RkyvError>(&buf)
+        .expect("failed to access rkyv archive");
+    zkvm.commit_buf(input.chunk_transition_ssz());
+}
```
prajwolrg: I don't fully understand the need for this. Why do we need to do this? Can't we run the STF in native mode?

functional-tests/entry.py
```diff
@@ -315,6 +315,36 @@ def main(argv: list[str]) -> int:
...
+        # bump to 2000ms and accept a longer runtime.
+        #
+        # `dev_native_noop_prover=True` skips the alpen-client's chunk +
+        # acct STF verification entirely (the OL predicate is always-accept
+        # in tests, so no proof correctness is being validated anyway).
+        # Without this, batch 2 stalls in `ProofPending` for 10+ minutes
+        # per native-prover run and the deposit -> EE crossing can't land
+        # within a CI budget. See `bin/alpen-client/src/main.rs` and
+        # `crates/proof-impl/{alpen-chunk,alpen-acct}/src/lib.rs` for the
+        # noop wiring.
```
prajwolrg: This is interesting! I wonder why this is. In the native mode the there shouldn't be any overhead with proof generation. Is it possible that the proof generation is failing instead?

prajwolrg (changes requested):
The test themselves looks good. I'm more concerned about the use of noop chunks.

crates/proof-impl/alpen-chunk/src/lib.rs
voidash: Yes, there were two failures `MismatchedPostState` and `InconsistentChunkIo`. They are now fixed

prajwolrg (approved):

storopoli:
@voidash given this is urgent. I've fixed the conflict using GitHub Web UI.

storopoli (approved):
ACK a8c1905f8f8f896d9f94aaa05f65d305a0b6c1fd

crates/alpen-ee/ol-tracker/src/task.rs
```diff
@@ -92,29 +99,52 @@ pub(crate) async fn track_ol_state(
...
+    // native noop prover emits SAUs quickly.
+    let mut effective_ol_status = ol_status;
+    let best_ol_commitment = if track_finalized_epoch {
+        let latest = ol_status.latest;
+        effective_ol_status.confirmed = latest;
+        effective_ol_status.finalized = latest;
+        &effective_ol_status.confirmed
+    } else {
+        &effective_ol_status.confirmed
+    };
```
sapinb, line 123: This doesnt make sense. `track_finalized_epoch` here means it works on `latest` ... the naming and the comments for this are all wrong.

storopoli: @voidash please take a look at this comment

voidash: This is misnamed . it should have been `latest` . This is dev hack specific to be used in  test only. 
As per why this was done, to make test fast so that deposit reaches EE without waiting for full checkpoint. Discussed and sapin is making changes in his PR

PR #1701, opened 2026-04-30
STR-3310: Add bookkeeping for funds in limbo

delbonis:
## Description

This PR adds bookkeeping to track funds that have been put into "limbo".  We can't access these funds, but the ledger will keep track of them so we can do something with them later when we decide on a plan.

Claude was used for some basic refactoring downstream of the human changes.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/ol/da/src/types/global.rs
+++ b/crates/ol/da/src/types/global.rs
@@ -2,7 +2,7 @@
 
 use strata_da_framework::{
     DaCounter,
-    counter_schemes::{self, CtrU64ByU16},
+    counter_schemes::{CtrU64BySignedVarInt, CtrU64ByU16},
     make_compound_impl,
 };
 
@@ -11,26 +11,37 @@ use strata_da_framework::{
 pub struct GlobalStateDiff {
     /// Slot counter diff.
     pub cur_slot: DaCounter<CtrU64ByU16>,
+
+    /// Limbo funds counter diff.
+    pub limbo_funds_sats: DaCounter<CtrU64BySignedVarInt>,
 }
 
 impl Default for GlobalStateDiff {
     fn default() -> Self {
         Self {
             cur_slot: DaCounter::new_unchanged(),
+            limbo_funds_sats: DaCounter::default(),
         }
     }
 }
 
 impl GlobalStateDiff {
     /// Creates a new [`GlobalStateDiff`] from a slot counter.
-    pub fn new(cur_slot: DaCounter<counter_schemes::CtrU64ByU16>) -> Self {
-        Self { cur_slot }
+    pub fn new(
+        cur_slot: DaCounter<CtrU64ByU16>,
+        limbo_funds_sats: DaCounter<CtrU64BySignedVarInt>,
+    ) -> Self {
+        Self {
+            cur_slot,
+            limbo_funds_sats,
+        }
     }
 }
 
 make_compound_impl! {
     GlobalStateDiff < (), crate::DaError > u8 => GlobalStateTarget {
-        cur_slot: counter (counter_schemes::CtrU64ByU16),
+        cur_slot: counter (CtrU64ByU16),
+        limbo_funds_sats: counter (CtrU64BySignedVarInt)
     }
 }
 
@@ -39,4 +50,7 @@ make_compound_impl! {
 pub struct GlobalStateTarget {
     /// Current slot value.
     pub cur_slot: u64,
+
+    /// Limbo funds value.
+    pub limbo_funds_sats: u64,
 }
--- a/crates/ol/state-support-types/src/da_accumulating_layer.rs
+++ b/crates/ol/state-support-types/src/da_accumulating_layer.rs
@@ -350,8 +350,11 @@ impl EpochDaAccumulator {
     /// adding more transactions would risk exceeding `OL_DA_DIFF_MAX_SIZE`.
     /// Overestimates are acceptable; underestimates are not.
     pub fn estimated_encoded_size(&self) -> usize {
-        // Global diff: 1 byte mask + up to 2 bytes slot counter.
-        let global_size = 3;
+        // Global diff:
+        // * 1 byte mask
+        // * up to 2 bytes for slot counter
+        // * up to 9 bytes for limbo funds counter
+        let global_size = 1 + 2 + 9;
 
         // New accounts list: 2-byte u16 length prefix + per-entry overhead.
         // Per new account: 32 (AccountId) + 8 (balance) + 1 (type disc)
--- a/crates/ol/state-support-types/src/da_accumulating_layer.rs
+++ b/crates/ol/state-support-types/src/da_accumulating_layer.rs
@@ -191,6 +191,9 @@ pub struct EpochDaAccumulator {
     /// Slot counter builder for the epoch.
     slot_builder: Option<DaCounterBuilder<CtrU64ByU16>>,
 
+    /// Limbo funds counter builder for the epoch.
+    limbo_funds_builder: Option<DaCounterBuilder<CtrU64BySignedVarInt>>,
+
     /// Expected first serial based on the pre-state next_account_serial.
     expected_first_serial: Option<AccountSerial>,
 
@@ -208,6 +211,10 @@ impl fmt::Debug for EpochDaAccumulator {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("EpochDaAccumulator")
             .field("slot_builder_set", &self.slot_builder.is_some())
+            .field(
+                "limbo_funds_builder_set",
+                &self.limbo_funds_builder.is_some(),
+            )
             .field("expected_first_serial", &self.expected_first_serial)
             .field("new_account_records", &self.new_account_records)
             .field("new_account_ids", &self.new_account_ids)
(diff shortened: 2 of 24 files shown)


crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -387,7 +407,13 @@ impl EpochDaAccumulator {
...
         };
 
-        Ok(GlobalStateDiff::new(cur_slot))
+        let limbo_funds_sats = if let Some(builder) = self.limbo_funds_builder.take() {
+            builder.into_write()?
+        } else {
+            DaCounter::new_unchanged()
+        };
+
+        Ok(GlobalStateDiff::new(cur_slot, limbo_funds_sats))
```
storopoli, line 419: `estimated_encoded_size()` is documented as a conservative upper bound, but it still hard-codes the global diff to 3 bytes even though this PR adds `limbo_funds_sats` to `GlobalStateDiff`. That means any tx that changes limbo funds can make the estimate too small by up to the signed-varint payload here, and block assembly uses this estimate to decide whether a tx still fits in the checkpoint budget.

Permalinks:
- estimator: https://github.com/alpenlabs/alpen/blob/49d6695a03ec9f212586659a6533608455cefb31/crates/ol/state-support-types/src/da_accumulating_layer.rs#L352-L354
- caller: https://github.com/alpenlabs/alpen/blob/49d6695a03ec9f212586659a6533608455cefb31/crates/ol/block-assembly/src/block_assembly.rs#L475-L476

Because underestimates are explicitly not allowed here, I think the new limbo-funds counter needs to be included in the upper bound before this lands.

storopoli (changes requested):
This was found by Codex GPT-5.5. I would have not found it because it is subtle and not included in the PR diff.

crates/ol/state-support-types/src/da_accumulating_layer.rs
delbonis: added 9 to it

irnb (approved):

crates/ol/stf/src/account_processing.rs
```diff
@@ -32,17 +31,18 @@ pub(crate) fn process_message<S: IStateAccessorMut>(
...
         // Any other address we assume is a ledger account, so we have to look it up.
         _ => {
+            let coin = Coin::new_unchecked(msg.value());
+
             // Check if the account exists first.
             if !state.check_account_exists(target)? {
-                // If we don't find it then we can just ignore it.
-                // TODO do something with the funds we're throwing away by doing this
+                // If we don't find it then we can sweep it into limbo.
+                state.add_limbo_funds_coin(coin)?;
```
bewakes: Should add a log here indicating why funds are being added to limbo.

crates/ol/stf/src/account_processing.rs
```diff
@@ -68,25 +68,27 @@ pub(crate) fn process_transfer<S: IStateAccessorMut>(
...
 ) -> ExecResult<()> {
+    let coin = Coin::new_unchecked(value);
+
     match target {
-        // Bridge gateway transfer.
+        // Bridge gateway transfer, not permitted.
         BRIDGE_GATEWAY_ACCT_ID => {
-            // TODO: what do we do with direct transfers to bridge accounts? Emit log?
+            // Just sweep to limbo unconditionally.
+            state.add_limbo_funds_coin(coin)?;
```
bewakes: Should add a log here indicating why funds are being added to limbo.

crates/ol/stf/src/account_processing.rs
```diff
@@ -68,25 +68,27 @@ pub(crate) fn process_transfer<S: IStateAccessorMut>(
...
         }
 
         // Any other address we assume is a ledger account, so we have to look it up.
         _ => {
             // Check if the account exists first.
             if !state.check_account_exists(target)? {
-                // If we don't find it then we can just ignore it.
-                // TODO: do something with the funds we're throwing away by doing this
+                // If we don't find it then we can sweep it into limbo.
+                state.add_limbo_funds_coin(coin)?;
```
bewakes: Should add a log here indicating why funds are being added to limbo.

crates/ol/stf/src/account_processing.rs
```diff
@@ -97,21 +99,23 @@ pub(crate) fn process_transfer<S: IStateAccessorMut>(
...
     payload: MsgPayload,
     context: &BasicExecContext<'_>,
 ) -> ExecResult<()> {
+    let coin = Coin::new_unchecked(payload.value());
+
     // 1. Parse the message from the payload data.
     let Ok(msg) = MsgRef::try_from(payload.data()) else {
-        // Invalid message format, just ignore.
+        // Invalid message format, sweep to limbo.
+        state.add_limbo_funds_coin(coin)?;
```
bewakes: Should add a log here indicating why funds are being added to limbo.

crates/ol/stf/src/account_processing.rs
```diff
@@ -97,21 +99,23 @@ pub(crate) fn process_transfer<S: IStateAccessorMut>(
...
         return Ok(());
     };
 
     let Some(withdrawal_data) = msg.try_as_withdrawal() else {
-        // Not a withdrawal message, or malformed, just ignore.
-        //
-        // TODO maybe reroute this to a different thing?
+        // Not a withdrawal message, or malformed, sweep to limbo.
+        state.add_limbo_funds_coin(coin)?;
```
bewakes: Should add a log here indicating why funds are being added to limbo.

crates/ol/stf/src/account_processing.rs
```diff
@@ -124,7 +128,8 @@ fn handle_bridge_gateway_message<S: IStateAccessorMut>(
     // 3. Verify the amount is a positive exact multiple of the denomination.
     let amt_raw: u64 = withdrawal_amt.into();
     if amt_raw == 0 || !amt_raw.is_multiple_of(withdrawal_denom) {
-        warn!(%amt_raw, "ignoring withdrawal with invalid amount");
+        // Sweep to limbo.
+        state.add_limbo_funds_coin(coin)?;
```
bewakes: Existing `warn!` is removed, please add it back.

bewakes (changes requested):
Nicely done.

I have two requests:
1. Please add logs to sites where we are adding funds to limbo along with the reason(decode error, nonexistent account, etc).
2. Also need to add explicit test cases for limbo funds, covering all the reasons/causes that can result in limbo.

delbonis:
@bewakes I was trying to move away from having logs in STF code because it's going to get reexecuted multiple times and it'd lead to spammy logging output.  I'd argue that now that we *do* have bookkeeping for it the logs are less necessary, hence why I removed that one.  I will add a log in chainexec code to warn if the limbo balance changes.

bewakes:
@delbonis, that's a fair point on multiple reexecution. I am mainly concerned over observability during our devnet/testnet deployments where we'd want to quickly debug and work on fixes. We'd want to see exactly what caused the limbo state.

It's fine if we can have the reason for limboness exposed in chain worker — that'd be my preference. If not, then I'd say let's add to the stf and control the spam-ness by having RUST_LOG flag set for ol_stf to mute DEBUG/TRACE.

One example from the latest deployment is that with the logs added, we were able to find the issue in cli that bridge uses which was constructing aux data incorrectly. Plus, in practice the limbo case would be very rare and in the case when it happens it might be fine to have a couple more logs.

storopoli (approved):
ACK 62dcae3ec98707c793923fa270f2f5c1a1476f0e

storopoli:
@delbonis is there a way to add WARN logs only when someone sends funds to limbo?
That is something that I think is important to have as @bewakes suggested.

delbonis:
Readded logs, also wrapped the call to the state fn in a `handle_misplaced_funds` fn so we just have to change that fn when we change this handling.

bewakes:
The logs are covered now, perfect!

One more thing. This changes OL STF behavior in five places:
 - `process_message` to a nonexistent account
 - `process_transfer` to the bridge gateway
 - `process_transfer` to a nonexistent account
 - `handle_bridge_gateway_message` malformed / non-withdrawal / bad-amount paths
 - `process_deposit_log` malformed descriptor / unknown serial paths

But I don't see any unit or functional tests covering the changes. This is pretty important as this directly touches OL state and stf. Please add comprehensive test cases for the changes. A DA round-trip test and a size-estimate assertion would also be good to have.

If any of this needs to land as a follow-up, please open the ticket before this can be merged.

storopoli:
@codex review

crates/ol/da/src/types/global.rs
```diff
@@ -11,26 +11,37 @@ use strata_da_framework::{
 pub struct GlobalStateDiff {
     /// Slot counter diff.
     pub cur_slot: DaCounter<CtrU64ByU16>,
+
+    /// Limbo funds counter diff.
+    pub limbo_funds_sats: DaCounter<CtrU64BySignedVarInt>,
```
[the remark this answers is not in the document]
delbonis: oh shit

bewakes (approved):

crates/ol/da/src/types/global.rs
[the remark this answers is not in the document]
storopoli: Ok let's solve this then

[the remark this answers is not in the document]
delbonis: solved

delbonis:
waiting for CI, will merge after

crates/ol/da/src/types/global.rs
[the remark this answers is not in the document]
storopoli: @codex is this finding solved?

> **<sub><sub>P1 Badge</sub></sub>  Apply limbo DA diff during reconstruction**
> 
> When an epoch contains any malformed deposit/message/transfer that moves funds into limbo, `DaAccumulatingState` now emits this new `limbo_funds_sats` counter, but `OLStateDiff::apply` still only applies `global.cur_slot` before checking the reconstructed preseal root. As a result `OLDaSchemeV1::apply_to_state`/checkpoint proof verification reconstructs a state with the old limbo balance, so the preseal root no longer matches for exactly the epochs this bookkeeping is meant to cover. > Apply `self.diff.global.limbo_funds_sats` to `target.limbo_funds()` via the accessor methods before ledger diffs are applied.

storopoli (approved):
ACK 67a2720a81c2500799bc7bd2c04e5b424be4cd08

PR #1740, opened 2026-05-01
(ol): Emit snark update logs

bewakes:
## Description

This simply emits a snark update log after processing the snark udpate. The `SnarkAccountUpdateLogData` already existed so this is smaller than expected.

The core changes were made by author and claude code was instructed to do refactoring and writing unit tests.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/ol/block-assembly/src/block_assembly.rs
+++ b/crates/ol/block-assembly/src/block_assembly.rs
@@ -2396,9 +2396,8 @@ mod tests {
             .build();
         let tx_overflow_id = tx_overflow.compute_txid();
 
-        // Pre-fill the buffer so that tx_fill's 10 logs exactly reach the cap,
-        // leaving no room for tx_overflow's 1 log.
-        let prefill = MAX_LOGS_PER_BLOCK as usize - 10;
+        // tx_fill emits 11 logs (10 withdrawals + 1 snark-update); tx_overflow emits 2.
+        let prefill = MAX_LOGS_PER_BLOCK as usize - 11;
         output_buffer
             .emit_logs((0..prefill).map(|i| OLLog::new(AccountSerial::from(i as u32), vec![])))
             .expect("pre-fill should succeed");
@@ -2559,10 +2558,10 @@ mod tests {
         let tx1_id = tx1.compute_txid();
         let tx2_id = tx2.compute_txid();
 
-        // Control case: one below hard-1 (i.e. hard-2) should still accept tx1.
+        // tx1 emits 2 logs (withdrawal + snark-update).
         let control = run_process_transactions_with_seeded_checkpoint_logs(
             account_id,
-            (MAX_OL_LOGS_PER_CHECKPOINT as usize) - 2,
+            (MAX_OL_LOGS_PER_CHECKPOINT as usize) - 3,
             vec![(tx1_id, tx1.clone())],
         )
         .await;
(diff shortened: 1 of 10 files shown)


crates/ol/stf/src/transaction_processing.rs
```diff
@@ -123,7 +124,7 @@ fn process_update_tx<S: IStateAccessorMut>(
 
     // 3. Actually take balance and write new account inner state.
     let upd = sau_payload.operation().update();
-    state.update_account(target, |astate| -> ExecResult<_> {
+    let serial = state.update_account(target, |astate| -> ExecResult<_> {
```
delbonis: We can get the serial just by doing `let serial = account_state.serial();`, we don't have to pull it out of the update closure like this.

crates/ol/stf/src/transaction_processing.rs
```diff
@@ -142,9 +143,15 @@ fn process_update_tx<S: IStateAccessorMut>(
...
-        Ok(())
+        Ok(astate.serial())
     })??;
 
+    // Emit log after successful update.
+    let new_msg_idx = upd.proof_state().new_next_msg_idx();
+    let extra_data = upd.extra_data().to_vec();
+    let log = SnarkAccountUpdateLogData::new(new_msg_idx, extra_data)
+        .expect("extra_data bounded by SSZ MAX_EXTRA_DATA_BYTES(1024) exceeds VarVec bound");
+    context.emit_typed_log(serial, &log)?;
```
delbonis, line 153: Should make this more general and less error-prone by adding a constructor for `SnarkAccountUpdateLogData` that accepts the snark account update data (whatever the minimal type is) as an arg and does this construction instead of leaving it up to the exec code.

This makes it so that we can construct the log produced by the tx without having to exec it (the only additional data we'd need is the serial, which is stable).

crates/ol/stf/src/test_utils.rs
```diff
@@ -663,9 +675,15 @@ impl SnarkUpdateBuilder {
...
             ledger_ref_proofs: vec![],
+            extra_data: vec![],
         }
     }
 
+    pub fn with_extra_data(mut self, extra_data: Vec<u8>) -> Self {
+        self.extra_data = extra_data;
+        self
+    }
+
```
delbonis: Should accept the SSZ `List<u8, ...>` matching the type and then have another `with_` fn that accepts a `impl TryInto<List<u8, ...>>` which would propagate the conversion error.

crates/ol/block-assembly/src/block_assembly.rs
delbonis, line 1: There's a lot of very manual math that needed to be updated by hand here, is there any way to make this more robust?  Maybe not actionable.

delbonis (changes requested):
Generally good, some minor cleanups possible.

crates/ol/block-assembly/src/block_assembly.rs
bewakes: Yeah, not liking this very much either. Maybe some actionable pattern would emerge once we have more cases like this coming up.

prajwolrg (approved):
LGTM!

delbonis (approved):

PR #1745, opened 2026-05-01
STR 3025 Properly populate updates and inbox messages in `RpcAccountBlockSummary`

krsnapaudel:
## Description

`strata_getBlocksSummaries` previously returned `updates` with empty `messages` and dropped `new_inbox_messages` entirely. This wires both fields against the new OL state-indexing tables (#1686).

- **`updates[].processed_messages`** — populated by walking the inbox MMR per record. Cursor seeds from the prior epoch's terminal `next_inbox_msg_idx` and advances through each record's `next_inbox_idx`. Genesis epoch short-circuits to cursor=0 with no fetch (no messages can have been consumed pre-genesis).
- **`new_inbox_messages`** — populated from `InboxMessageRecord` rows tagged with the block's commitment, decoded from SSZ.
- Per-block lookup is O(1) via a `HashMap<OLBlockCommitment, Vec<UpdateInputData>>` built at fetch time. Records keyed to blocks outside the queried chain range are dropped.
Also refactored the use of slots and epochs to use `Slot` and `Epoch` types.

This PR is created with help from Claude Code and Codex.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-3025](https://alpenlabs.atlassian.net/browse/STR-3025)

[STR-3025]: https://alpenlabs.atlassian.net/browse/STR-3025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/strata/src/rpc/node.rs
+++ b/bin/strata/src/rpc/node.rs
@@ -1,14 +1,17 @@
 //! OL RPC server implementation for a strata node.
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::{HashMap, HashSet},
+    sync::Arc,
+};
 
 use async_trait::async_trait;
 use jsonrpsee::core::RpcResult;
 use ssz::{Decode, Encode};
 use strata_acct_types::MessageEntry;
 use strata_checkpoint_types::EpochSummary;
-use strata_db_types::ol_state_index::{AccountUpdateRecord, InboxMessageRecord};
+use strata_db_types::ol_state_index::InboxMessageRecord;
 use strata_identifiers::{
-    AccountId, Epoch, EpochCommitment, L1BlockCommitment, L1Height, L2BlockCommitment,
+    AccountId, Epoch, EpochCommitment, Hash, L1BlockCommitment, L1Height, L2BlockCommitment,
     OLBlockCommitment, OLBlockId, OLTxId,
 };
 use strata_ledger_types::{IAccountState, ISnarkAccountState};
@@ -45,7 +48,7 @@ pub(crate) struct OLRpcServer<P: OLRpcProvider> {
 
 /// Convenient wrapper for account records.
 struct AccountRecords {
-    updates: Vec<AccountUpdateRecord>,
+    updates_by_block: HashMap<OLBlockCommitment, Vec<UpdateInputData>>,
     inbox: Vec<InboxMessageRecord>,
 }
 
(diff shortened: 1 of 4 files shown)


bin/strata/src/rpc/node.rs
```diff
@@ -235,7 +231,60 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
                 .await
                 .map_err(db_error)?
             {
-                all_updates.extend(records);
+                let skip_fetch = epoch == 0;
```
bewakes: A few things we could simplify and organize in this block:                                                                                                                                                                                                         
                                                                                                                                                                                                                                               
  1. The toplevel OL state only needs to be fetched once — for the epoch before `first_epoch`. The inbox is monotonic per account, so each record's `next_inbox_idx` is the next one's start, even across epoch boundaries.
  2. We can flatten records across the range and build `(prev, next)` index pairs grouped by block commitment in a single pass. Checkpoint-sync rows just advance the cursor without emitting anything.                                          
  3. Inbox messages could then be fetched once for the whole range and sliced locally, instead of one call per record.

bin/strata/src/rpc/node.rs
```diff
@@ -235,7 +231,60 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
...
+                        })?
+                        .to_vec();
+
+                    let messages = if skip_fetch {
+                        Vec::new()
+                    } else {
+                        self.provider
+                            .get_account_inbox_messages(account_id, cursor, next_cursor)
+                            .await
+                            .map_err(db_error)?
```
storopoli: Non-blocking: this helper now hydrates `extra_data` and `processed_messages` for every indexed record in each touched epoch before `get_blocks_summaries` filters down to the requested block commitments. That means a record for some other block in the same epoch can still trip the `DirectSet` path or an inbox-fetch error and cause the whole RPC to fail, even though the PR description says records outside the queried chain range are dropped later. It seems safer to pre-filter records to the queried block commitments before hydrating them, or defer the inbox/message reconstruction until the per-block summary step.

Permalink: https://github.com/alpenlabs/alpen/blob/9aa52a33bb07940bb5304c24361e59f600fe1872/bin/strata/src/rpc/node.rs#L256-L272

storopoli (approved):
ACK 9aa52a33bb07940bb5304c24361e59f600fe1872

bin/strata/src/rpc/node.rs
krsnapaudel: @bewakes Addressed. PTAL.

bewakes (approved):

krsnapaudel:
@storopoli Can we merge this now?

PR #1746, opened 2026-05-02
fix(btcio,alpen-ee): broadcaster -25 disambiguation + block-builder inbox-fetch off-by-one

voidash:
Three production-code fixes that surfaced building the real-bridge E2E test ([#1699](https://github.com/alpenlabs/alpen/pull/1699)), plus the bridge indexing fixes needed for the full deposit -> withdrawal path. Each is independently load-bearing: revert any one and the test stalls in a different place.

| File | Bug | Fix |
|---|---|---|
| `alpen-ee/sequencer/src/block_builder/task.rs` | `get_inbox_messages` was called with `from_slot = last_local_block.ol_block().slot()`, range inclusive on both ends. Successive EE batches re-included the OL inbox slot the previous batch already drained. SAU shipped with `processed_messages.len() = 1` and unchanged `next_inbox_msg_idx`; block-assembly rejected with `expected 2, got 1`. seq_no=0 applied, every later SAU starved. | Start at `last_consumed_slot + 1`. Pick a monotonic `effective_ol_block` for the exec record so the OL commitment never regresses during restart catch-up, finalization lag, or reorg. |
| `btcio/src/broadcaster/io.rs` | Every `Server(-25, _)` was folded to `AlreadyInMempool`. Bitcoind reuses -25 for several reject reasons; `bad-txns-inputs-missingorspent` (input UTXO gone) was being treated as benign and entries pinned at `Published`. ASM checkpoint validation stalled at epoch 4 of ~160. | `is_benign_minus25_message` matches the hyphenated tokens bitcoind emits (`already-in-mempool`, `already-known`, `already-in-block-chain`). Everything else under -25 routes to `InvalidInputs`. |
| `btcio/src/broadcaster/processor.rs` | `Published` entries had two failure modes: `Ok(None)` from `gettransaction` regressed them to `Unpublished` (publish/revert oscillation pinning the watcher), and the broadcaster never re-called `send_raw_transaction` on stuck-Published, so genuinely-dropped txs stayed `Published` forever. | Hold `Published` across a transient lookup miss. New `probe_published_entry` runs the confirmation lookup and re-attempts publication only on actual not-found; `Some(info)` at 0-conf holds without re-publishing. `InvalidInputs` from the probe routes the entry so the watcher rebuilds. |
| `chain-worker-new`, `storage/mmr_index`, `strata RPC block summaries`, `alpen-client/prover/spec_acct.rs` | Snark inbox messages were present in OL state, but their MMR preimages were not indexed for the prover path. The block-summary RPC also reported proof-state `next_inbox_msg_idx`, which can lag the inbox accumulator count after new inbox inserts. That lets the prover assemble an inconsistent deposit-processing input. | Index snark inbox MMR writes when storing block output, add MMR leaf preimage append helpers, report the inbox accumulator entry count from block summaries, and guard the client prover against inconsistent pre-inbox indices with a transient retryable failure. |
| `alpen-ee/sequencer/src/ol_chain_tracker/state.rs` | Pruning tracked OL blocks could drop the base block's `next_inbox_msg_idx`, causing later inbox fetch ranges to restart from the wrong base. | Preserve `base_block_next_inbox_msg_idx` across prune cases, including the empty-after-prune path. |
| `db/store-sled/src/asm/db.rs` | `get_latest_asm_state` asserted when a concurrent read saw latest state metadata before the matching state row. | Fall back through common-block state lookups instead of asserting. |

Covered by focused unit tests in `broadcaster::{io,processor}`, Strata RPC block summaries, storage, chain-worker, ASM sled DB, and EE sequencer OL tracker state.

## Local verification

- [x] `git diff --check HEAD`
- [x] `cargo test -p strata --bin strata blocks_summaries_reports_inbox_accumulator_count`
- [x] `cargo test -p strata-storage`
- [x] `cargo test -p strata-chain-worker-new`
- [x] `cargo test -p strata-db-store-sled asm::db::tests::test_get_asm`
- [x] `cargo test -p alpen-ee-sequencer ol_chain_tracker::state`

## AI Assistance

Bug-chain debugging, code review, implementation, and this description.

### DIFF
--- a/crates/alpen-ee/sequencer/src/block_builder/task.rs
+++ b/crates/alpen-ee/sequencer/src/block_builder/task.rs
@@ -56,6 +56,50 @@ fn should_fetch_inbox_messages(last_local_ol_blkid: &OLBlockId, best_ol_blkid: &
     last_local_ol_blkid != best_ol_blkid
 }
 
+/// Picks the OL block commitment to record on the next exec block.
+///
+/// During restart catch-up, finalization lag, or reorg handling, the OL
+/// tracker can report a block behind the local EE tip. Returning
+/// `last_local_ol_block` in that case keeps the OL slot recorded on the EE
+/// chain monotonic; using the older `best_ol_block` would cause inbox
+/// fetching to re-include slots already reflected in account state.
+fn effective_ol_block(
+    best_ol_block: OLBlockCommitment,
+    last_local_ol_block: OLBlockCommitment,
+) -> OLBlockCommitment {
+    if best_ol_block.slot() < last_local_ol_block.slot() {
+        last_local_ol_block
+    } else {
+        best_ol_block
+    }
+}
+
+/// Computes the inbox-fetch slot range for the next exec block.
+///
+/// Returns `None` when no fetch is needed: either the OL tip is unchanged
+/// from what's already in account state, or the slot delta is empty.
+/// Otherwise returns `Some((from_slot, to_slot))` with both ends inclusive.
+///
+/// `from_slot = last_local_ol_slot + 1` is the off-by-one fix: the previously
+/// recorded slot's messages are already drained, so re-fetching it produces a
+/// SAU whose `processed_messages` does not match `new_next_inbox_msg_idx` and
+/// OL rejects it with "invalid next msg index".
+fn compute_inbox_fetch_range(
+    last_local_ol_slot: u64,
+    last_local_ol_blkid: &OLBlockId,
+    effective_ol_slot: u64,
+    effective_ol_blkid: &OLBlockId,
+) -> Option<(u64, u64)> {
+    if !should_fetch_inbox_messages(last_local_ol_blkid, effective_ol_blkid) {
+        return None;
+    }
+    let from_slot = last_local_ol_slot.saturating_add(1);
+    if from_slot > effective_ol_slot {
+        return None;
+    }
+    Some((from_slot, effective_ol_slot))
+}
+
 /// Constructs BlockAssemblyInputs from the current state.
 fn create_block_assembly_inputs<'a>(
     last_local_block: &ExecBlockRecord,
@@ -241,37 +285,20 @@ async fn build_next_block(
         .await
         .context("build_next_block: failed to get finalized OL block")?;
 
-    // Record a monotonic OL commitment on the new exec block. During restart
-    // catch-up, finalization lag, or reorg handling, the OL tracker can report
-    // a block behind the local EE tip. Writing that lower slot makes inbox
-    // fetching include slots already reflected in account state.
-    let effective_ol_block: OLBlockCommitment =
-        if best_ol_block.slot() < last_local_block.ol_block().slot() {
-            *last_local_block.ol_block()
-        } else {
-            best_ol_block
-        };
+    let effective_ol_block = effective_ol_block(best_ol_block, *last_local_block.ol_block());
 
-    let (inbox_messages, next_inbox_msg_idx) = if should_fetch_inbox_messages(
+    let (inbox_messages, next_inbox_msg_idx) = match compute_inbox_fetch_range(
+        last_local_block.ol_block().slot(),
         last_local_block.ol_block().blkid(),
+        effective_ol_block.slot(),
         effective_ol_block.blkid(),
     ) {
-        // Fetch only slots newer than the last OL slot reflected in account
-        // state. Re-including a drained slot produces a SAU whose
-        // processed_messages do not match new_next_inbox_msg_idx, and OL
-        // rejects it with "invalid next msg index".
-        let from_slot = last_local_block.ol_block().slot().saturating_add(1);
-        if from_slot > effective_ol_block.slot() {
-            (vec![], last_local_block.next_inbox_msg_idx())
-        } else {
-            ol_chain_handle
-                .get_inbox_messages(from_slot, effective_ol_block.slot())
-                .await
-                .context("build_next_block: failed to get inbox messages")?
-                .into_parts()
-        }
-    } else {
-        (vec![], last_local_block.next_inbox_msg_idx())
+        Some((from_slot, to_slot)) => ol_chain_handle
+            .get_inbox_messages(from_slot, to_slot)
+            .await
+            .context("build_next_block: failed to get inbox messages")?
+            .into_parts(),
+        None => (vec![], last_local_block.next_inbox_msg_idx()),
     };
 
     // build next block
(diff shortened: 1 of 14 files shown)


crates/btcio/src/broadcaster/processor.rs
```diff
@@ -72,7 +73,73 @@ where
...
+            _ => Ok(L1TxStatus::Published),
+        },
+    }
+}
+
+/// Maps a confirmed `TxConfirmationInfo` (`info.confirmations >= 1`) to the
+/// corresponding `Confirmed` or `Finalized` status. Caller must ensure the
+/// `confirmations >= 1` precondition; the block_hash/block_height fields are
+/// only populated by bitcoind once a tx is mined.
+fn confirmation_status(info: &TxConfirmationInfo, reorg_safe_depth: i64) -> L1TxStatus {
```
bewakes, line 114: This function does not consider the possibility of confirmations = 0 - if confirmations is less than finality depth, it assumes it is confirmed.. This works fine for what's there now, but this should also pull in the case where confirmations is 0, handle that and return appropriate status. This might also make the existing `check_tx_confirmations` obsolete.

crates/alpen-ee/sequencer/src/block_builder/task.rs
```diff
@@ -241,15 +240,36 @@ async fn build_next_block(
         .get_finalized_block()
         .await
         .context("build_next_block: failed to get finalized OL block")?;
+
+    // Record a monotonic OL commitment on the new exec block. During restart
+    // catch-up, finalization lag, or reorg handling, the OL tracker can report
+    // a block behind the local EE tip. Writing that lower slot makes inbox
+    // fetching include slots already reflected in account state.
+    let effective_ol_block: OLBlockCommitment =
```
bewakes: I'd suggest to add unit tests for the bug the change fixes.

bewakes (commented):
Thanks for the fixes! Some comments and suggestions.

crates/btcio/src/broadcaster/io.rs
```diff
@@ -10,9 +10,22 @@ use strata_btc_types::BlockHashExt;
...
+/// `validation.cpp`/`policy/policy.cpp`): `txn-already-in-mempool`,
+/// `txn-already-known`, `txn-already-in-block-chain`. A space-separated check
+/// like `"already in mempool"` does not match any of these and would route
+/// every benign `-25` to `InvalidInputs`, causing spurious envelope rebuilds.
+fn is_benign_minus25_message(msg: &str) -> bool {
+    msg.contains("already-in-mempool")
+        || msg.contains("already-known")
+        || msg.contains("already-in-block-chain")
+}
+
```
storopoli, line 28: This is bad. We should try to fix this in [`alpenlabs/bitcoind-async-client` in `error.rs`](https://github.com/alpenlabs/bitcoind-async-client/blob/main/src/error.rs#L101-L103):

```rs
impl ClientError {
    // ..

    pub fn is_missing_or_invalid_input(&self) -> bool {
        matches!(self, Self::Server(-26, _)) || matches!(self, Self::Server(-25, _))
    }
}
```

As you can see we are globbing together -26 and -25 error codes. Unfortunately there error codes documentation is only in a [C++ header in Bitcoin core codebase](https://github.com/bitcoin/bitcoin/blob/8f4a3ba8972dae9412ba975a040cea22c227f983/src/rpc/protocol.h#L47-L48).

-25 is only used for mempool refusals: https://github.com/bitcoin/bitcoin/blob/8f4a3ba8972dae9412ba975a040cea22c227f983/src/rpc/util.cpp#L395

and -26 is used for bad inputs in the raw transaction RPC calls: https://github.com/bitcoin/bitcoin/blob/8f4a3ba8972dae9412ba975a040cea22c227f983/src/rpc/rawtransaction.cpp#L651

Please make a PR in `bitcoind-async-client` we can merge quickly and do a quick version bump in crates.io

storopoli (changes requested):

crates/btcio/src/broadcaster/io.rs
storopoli: This has arisen also in #1679. Will do a PR to `bitcoind-async-client` and will report back here.

storopoli: Ok please bump to `0.10.4`. Tagged and published in https://crates.io/crates/bitcoind-async-client

Fixed in https://github.com/alpenlabs/bitcoind-async-client/pull/110

crates/btcio/src/broadcaster/processor.rs
voidash: Addressed in the current head. `confirmation_status` now treats `confirmations <= 0` as `Published`, and the confirmed-entry path still overrides exact 0-conf to `Unpublished` for reorg recovery. Added coverage for zero/negative confirmations and the published-entry 0-conf no-republish case. Verified with `cargo test -p strata-btcio confirmation_status` and `cargo test -p strata-btcio test_published_entry_at_zero_conf_does_not_republish`.

crates/alpen-ee/sequencer/src/block_builder/task.rs
voidash: Addressed in the current head. The inbox-fetch logic is split into `effective_ol_block` and `compute_inbox_fetch_range`, with unit coverage for the behind-tracker case and the off-by-one regression (`slot 5 -> 7` fetches `[6, 7]`, not `[5, 7]`). Verified with `cargo test -p alpen-ee-sequencer block_builder::task::tests`.

storopoli:
Ashish please use version `0.10.5` of `bitcoind-async-client` it does not have the `bitreq` bump so should be compatible with `alpen`.

crates/db/store-sled/src/asm/db.rs
```diff
@@ -47,14 +47,21 @@ impl AsmDatabase for AsmDBSled {
...
 
-        // Assert that the block for the state and for the logs is the same.
-        // It should be because we are putting it within transaction.
-        Ok(state.and_then(|s| {
-            logs.map(|l| {
-                assert_eq!(s.0, l.0);
-                (s.0, AsmState::new(s.1, l.1))
-            })
-        }))
+        match (state, logs) {
```
bewakes: nit: Why not just get min of both at first and get data for that min? This should handle the equality case as well.

crates/chain-worker-new/src/context.rs
```diff
@@ -346,3 +374,60 @@ fn build_indexing_writes(
...
     IndexingWrites::new(created_accounts, account_updates, account_inbox_writes)
 }
+
+/// Applies snark inbox writes to the MMR proof index.
+///
+/// The OL state itself stores the compact MMR root/peaks. Block assembly needs
+/// historical nodes from [`MmrIndexManager`] to generate proofs for later snark
+/// account updates, so the chain worker mirrors each accepted inbox append into
+/// the proof index. The operation is idempotent for crash-restart retries.
+fn index_inbox_mmr_writes(
```
bewakes, line 379: Please add tests that verify this is idempotent as well as does correct writes.

sapinb (approved):
ACK f76b8765d991c322635d75a30268df7c277244ab for `alpen-ee/*` changes

crates/chain-worker-new/src/context.rs
```diff
@@ -165,8 +173,28 @@ impl ChainWorkerContext for ChainWorkerContextImpl {
...
+            Ok(()) => {
+                index_inbox_mmr_writes(&self.mmr_index_mgr, output)?;
+            }
+            Err(DbError::BlockIndexingConflict {
+                attempted,
+                last_applied,
+                ..
+            }) if attempted == commitment && last_applied == commitment => {
+                index_inbox_mmr_writes(&self.mmr_index_mgr, output)?;
+                return Err(DbError::BlockIndexingConflict {
```
bewakes: Shouldn't this be discarded instead or returning error here? Also, it should always be the case that `attempted == commitment` but fine to have the check

crates/btcio/src/broadcaster/processor.rs
```diff
@@ -85,39 +159,12 @@ where
...
-                            block_hash,
-                            block_height,
-                        })
-                    }
-                }
-            },
-            Ok(None) => Ok(L1TxStatus::Unpublished),
-            Err(e) => Err(e),
+        match txinfo_res? {
+            Some(info) if info.confirmations == 0 => Ok(L1TxStatus::Unpublished),
```
bewakes: This is not correct, right? if there's info then it is published. confirmations = 0 means it is in the mempool i.e. it is published.
Also, I don't think we need this function anymore.

crates/btcio/src/broadcaster/processor.rs
```diff
@@ -605,6 +652,112 @@ mod test {
...
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_published_entry_at_zero_conf_does_not_republish() {
+        // A `Published` entry that `gettransaction` reports as 0-conf in the
+        // wallet is alive in mempool and waiting to be mined. Re-publishing
+        // every poll for the entire pre-confirmation window would only spam
+        // bitcoind and the broadcaster logs. The probe must hold at Published
+        // without calling `send_raw_transaction`. We poison the broadcast path
+        // so any re-publish would surface as `InvalidInputs`; the assertion
+        // proves the probe never went down that road.
```
bewakes: This is unnecessary and too verbose. Makes it harder to review. The intent is already clear from the function name.

crates/btcio/src/broadcaster/processor.rs
```diff
@@ -605,6 +652,112 @@ mod test {
...
+        assert_eq!(
+            res,
+            Some(L1TxStatus::Published),
+            "Published entry must hold its status when get_transaction returns NotFound"
+        );
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_published_entry_dropped_from_mempool_advances_to_invalid_inputs() {
+        // A `Published` entry whose lookup says "not found" AND whose
```
bewakes: This is unnecessary and too verbose. Makes it harder to review. The intent is already clear from the function name.

bewakes (changes requested):
This is better. But there are two major blockers:
1. Setting status to be `Unpublished` when confirmations = 0 is not correct.
2. Missing tests for inbox mmr insertions and idempotency.

crates/chain-worker-new/src/context.rs
voidash: added tests

crates/db/store-sled/src/asm/db.rs
voidash: Applied the simplification `min(state_block, logs_block)`

bewakes (approved):

prajwolrg (approved):

storopoli (approved):
ACK 23fdceb29a7df01ccccf77447578b6488660abd0

PR #1747, opened 2026-05-03
feat(ol/rpc): add decoded RPC endpoints for blocks, txs, and account changes

voidash:
## Summary
- Adds 4 decoded JSON RPC endpoints to `OLFullNodeRpc`: `strata_getBlockBySlot`, `strata_getRecentBlocks`, `strata_getBlockTransactions`, `strata_getBlockAccountChanges`
- Backs `getBlockAccountChanges(slot)` with the persisted per-block OL write batch, so OL Explorer can fetch created/updated account post-state block by block without enumerating global account state
- Defines new RPC response types in `strata-ol-rpc-types` for decoded blocks, transactions/effects, account state, and block account changes
- Removes the broad `strata_listAccounts` surface and the ledger iterator helpers that were only needed for full-state account listing

Closes [STR-3097](https://alpenlabs.atlassian.net/browse/STR-3097).

## Out of scope
- `getBlockLogs` is deferred. Per-block logs are not persisted today (only `logs_root` is in the header), so the endpoint needs new storage infrastructure.
- Static HTML serving for the explorer UI ([STR-3098](https://alpenlabs.atlassian.net/browse/STR-3098)).
- E2E tests ([STR-3099](https://alpenlabs.atlassian.net/browse/STR-3099)).

## Test plan
- [x] `cargo fmt --all --check`
- [x] `cargo check -p strata-ol-rpc-api -p strata-ol-rpc-types --features strata-ol-rpc-types/jsonschema --locked`
- [x] `cargo test -p strata --bin strata rpc::node_tests --locked`
- [x] `cargo clippy -p strata-ol-rpc-types -p strata-ol-rpc-api -p strata --bin strata --tests --all-features --locked -- -D warnings`
- [ ] Manual smoke against a running strata node

[STR-3097]: https://alpenlabs.atlassian.net/browse/STR-3097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[STR-3098]: https://alpenlabs.atlassian.net/browse/STR-3098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[STR-3099]: https://alpenlabs.atlassian.net/browse/STR-3099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/ol/rpc/types/src/account_state.rs
```diff
@@ -53,3 +55,102 @@ impl RpcSnarkAccountState {
...
+///
+/// Distinct from [`RpcSnarkAccountState`] in that it omits the update verification
+/// key, which is not available from the runtime account state.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]
+pub struct RpcAccountSnarkSummary {
+    seq_no: u64,
+    inner_state_root: HexBytes32,
+    next_inbox_msg_idx: u64,
+}
```
storopoli, line 215: This should be included in the `RpcAccountType` like

```rs
pub enum RpcAccountType {
    Empty,
    Snark(RpcAccountSnarkSummary),
}
```

Then you don't need to have the optional field `snark` in `RpcAccountEntry`.
You can then play around with macros on how to flatten or nest this inside `JsonSchema`.

`schemars` will follow the `#[serde(...)]` macros (see https://graham.cool/schemars/deriving/attributes/ and https://serde.rs/)

crates/ol/rpc/types/src/tx.rs
```diff
@@ -156,6 +162,220 @@ pub enum RpcTxConversionError {
...
+    /// Human-readable type name.
+    type_name: String,
+    /// Target account.
+    target: HexBytes32,
+    /// Inclusion constraints.
+    constraints: RpcTxConstraints,
+    /// Effects produced when this transaction is applied.
+    effects: RpcTxEffectsView,
+    /// SAU-specific fields, present only when `type_id == 1`.
+    sau: Option<RpcSauTxSummary>,
```
storopoli: Same general principle as before, you can resolve this with a enum variant that carries data if is a SAU.

crates/ol/rpc/types/src/account_state.rs
```diff
@@ -85,48 +83,50 @@ impl RpcAccountEntry {
...
+/// summary data inline so the wire format is self-describing.
+#[derive(Clone, Debug, Serialize, Deserialize)]
 #[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]
-#[serde(rename_all = "snake_case")]
-pub enum RpcAccountType {
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum RpcAccountKind {
     Empty,
-    Snark,
+    Snark(RpcAccountSnarkSummary),
```
storopoli: Oh fuck you did here. Ok sorry solving this

storopoli (approved):
ACK 193c0e9941473bd6b16f81243c226256dc8cdfa4

Reviewed commit by commit. My suggestions were addressed in follow-up commits.

bin/strata/src/rpc/node.rs
```diff
@@ -1065,4 +1065,102 @@ impl<P: OLRpcProvider> OLFullNodeRpcServer for OLRpcServer<P> {
...
+        block_or_tag: OLBlockOrTag,
+        start: u64,
+        count: u64,
+    ) -> RpcResult<RpcAccountListPage> {
+        if count as usize > self.max_headers_range {
+            return Err(invalid_params_error(format!(
+                "count {} exceeds max_headers_range {}",
+                count, self.max_headers_range
+            )));
+        }
```
bewakes: I don't think this `max_headers_range` check makes sense here. The cap is for max headers/blocks. Since we are listing accounts in a block, I think we don't even need a pagination here. The reason being, the amount of data returned is always less than the size of the block which we already support. There's no point in pagination here imo. The signature can just be `list_accounts(block_or_tag)`.

bin/strata/src/rpc/node_tests.rs
```diff
@@ -2785,3 +2863,421 @@ async fn raw_blocks_range_exceeds_max_returns_invalid_params() {
...
+        .expect("account present in ledger");
+    assert!(matches!(our_entry.kind(), RpcAccountKind::Snark(_)));
+    let snark = our_entry.snark().expect("snark summary");
+    assert_eq!(snark.seq_no(), 7);
+    assert_eq!(page.total(), page.entries().len() as u64);
+    assert!(page.next_offset().is_none());
+}
+
+#[tokio::test]
+async fn list_accounts_count_exceeds_max_returns_invalid_params() {
```
bewakes: Relating to above comment, this test is likely irrelevant.

bin/strata/src/rpc/node_tests.rs
```diff
@@ -2785,3 +2863,421 @@ async fn raw_blocks_range_exceeds_max_returns_invalid_params() {
...
+            0,
+            (TEST_MAX_HEADERS_RANGE as u64) + 1,
+        )
+        .await;
+    assert!(result.is_err());
+    assert_eq!(result.unwrap_err().code(), INVALID_PARAMS_CODE);
+}
+
+#[tokio::test]
+async fn list_accounts_zero_count_returns_empty_page() {
```
bewakes: This too.

bin/strata/src/rpc/node_tests.rs
```diff
@@ -2785,3 +2863,421 @@ async fn raw_blocks_range_exceeds_max_returns_invalid_params() {
...
+        .list_accounts(OLBlockOrTag::Slot(4), 0, 0)
+        .await
+        .expect("list accounts");
+    assert!(page.entries().is_empty());
+    assert_eq!(page.total(), 1);
+    assert!(page.next_offset().is_none());
+}
+
+#[tokio::test]
+async fn list_accounts_zero_count_still_resolves_block() {
```
bewakes: The test name and body do not match. The name says "still resolves block" but we are asserting err.

bin/strata/src/rpc/node_tests.rs
```diff
@@ -2785,3 +2863,421 @@ async fn raw_blocks_range_exceeds_max_returns_invalid_params() {
...
+        .await
+        .expect("page 3");
+    assert!(page3.entries().is_empty());
+    assert_eq!(page3.total(), 3);
+    assert!(page3.next_offset().is_none());
+}
+
+#[tokio::test]
+async fn list_accounts_resolves_confirmed_tag_to_confirmed_epoch() {
+    // Regression cover for the codex P2 finding: `Confirmed` must resolve to
```
bewakes: The comment doesn't need to mention about "codex P2 finding", this is completely a meta-commentary.

bewakes (commented):
Thanks for putting this in place.

I have some minor concerns over the rpc for listing accounts in a block. I don't think it needs pagination. See comments.

storopoli:
@voidash you solved a bunch of comments from @bewakes' review but I see no code being pushed. Has this been done in an offline setting/conversation with @bewakes? Can you please cultivate the good habit or replying on review comments before solving them? This helps reviewers understand what is going.

crates/ol/rpc/api/src/lib.rs
```diff
@@ -103,6 +103,31 @@ pub trait OLFullNodeRpc {
...
+    #[method(name = "getBlockTransactions")]
+    async fn get_block_transactions(&self, slot: u64) -> RpcResult<Vec<RpcOLTxDetail>>;
+
+    /// List accounts on the ledger at the given block.
+    ///
+    /// `block_or_tag` accepts the same forms as `getSnarkAccountState`:
+    /// `"latest"`, `"confirmed"`, `"finalized"`, a slot number, or a block ID
+    /// (`0x...`).
+    #[method(name = "listAccounts")]
+    async fn list_accounts(&self, block_or_tag: OLBlockOrTag) -> RpcResult<Vec<RpcAccountEntry>>;
```
delbonis: This is tricky, since when we open up account creation this would eventually have to return *way too much* data.  Maybe we can leave it in a debug namespace to indicate that it's not stable and might/will be removed in the future.

crates/ol/rpc/types/src/account_state.rs
```diff
@@ -53,3 +55,102 @@ impl RpcSnarkAccountState {
...
+}
+
+/// Account-type discriminator for [`RpcAccountEntry`].
+///
+/// Mirrors the on-chain `OLAccountTypeState` and carries snark-specific
+/// summary data inline so the wire format is self-describing.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum RpcAccountKind {
```
delbonis: Calling this `Kind` is a little weird.  Should be `RpcAccountTypeData` to match the protocol name and also clarify it's not just indicating the type but also contains the data (if applicable).

crates/ol/rpc/types/src/account_state.rs
```diff
@@ -53,3 +55,102 @@ impl RpcSnarkAccountState {
         &self.update_vk
     }
 }
+
+/// Account list entry returned by `strata_listAccounts`.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]
+pub struct RpcAccountEntry {
+    /// Account ID.
+    id: HexBytes32,
```
delbonis, line 98: Should split out the ID from the rest of the fields.  Create an `RpcAccountState` type that contains the other fields and then the entry type can contain `id` and `state`.

crates/ol/rpc/types/src/account_state.rs
```diff
@@ -53,3 +55,102 @@ impl RpcSnarkAccountState {
...
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum RpcAccountKind {
+    Empty,
+    Snark(RpcAccountSnarkSummary),
+}
+
+/// Snark-account summary fields surfaced in account listings.
+///
+/// Distinct from [`RpcSnarkAccountState`] in that it omits the update verification
+/// key, which is not available from the runtime account state.
```
delbonis: wait why isn't it?

crates/ol/rpc/types/src/tx.rs
```diff
@@ -156,6 +157,212 @@ pub enum RpcTxConversionError {
...
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum RpcOLTxKind {
+    GenericAccountMessage,
+    SnarkAccountUpdate(RpcSauTxSummary),
+}
+
+/// Summary of transfers and messages produced by a transaction.
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]
+pub struct RpcTxEffectsView {
```
delbonis: What does `View` mean here?

crates/ol/rpc/types/src/tx.rs
```diff
@@ -156,6 +157,212 @@ pub enum RpcTxConversionError {
...
+}
+
+/// Payload-type discriminator for [`RpcOLTxDetail`].
+///
+/// Mirrors the on-chain [`TransactionPayload`] enum and carries any
+/// type-specific summary data inline, so the wire format is self-describing.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum RpcOLTxKind {
```
delbonis: Again just call this "type".

delbonis (changes requested):
Some comments about API stability and data structures.

voidash:
> you solved a bunch of comments from @bewakes' review but I see no code being pushed. 

I Made changes locally and marked it resolved. I should put habit to only resolve after code is pushed. 
Thanks for pointing that out

crates/ol/rpc/api/src/lib.rs
voidash: This endpoint is meant for production, putting it under debug namespace may make it unavailable in prod deployments. 
Maybe pagination is what we want here. wdyt

voidash: now that i think about this. This is the wrong shape for the OL explorer. blockscout has indexers. Ol explorer should also have something like that

voidash:
@codex review

crates/ol/rpc/api/src/lib.rs
voidash: removed `list_accounts`

bewakes (approved):
LGTM.

crates/ol/rpc/api/src/lib.rs
delbonis: awesome, even simpler

delbonis (approved):

PR #1749, opened 2026-05-03
feat(strata): permissive CORS on the RPC port

voidash:
## Summary
- Wraps the jsonrpsee server with `tower-http::cors::CorsLayer::permissive()` via `ServerBuilder::set_http_middleware`
- Browser-based tools (the OL explorer at [alpenlabs/ol-explorer](https://github.com/alpenlabs/ol-explorer), future ones) can now call `strata_*` methods cross-origin without being blocked by the browser's same-origin policy
- New functional test `test_rpc_cors` covering both the OPTIONS preflight and the actual cross-origin POST

## Why permissive
The JSON-RPC port has no authenticated state — no cookies, no session tokens, just stateless reads/writes against public chain data. There's nothing for a malicious page to escalate via CORS, so allowing any origin/method/header is appropriate. 

## Test plan
- [x] `cargo build -p strata` clean
- [x] `just lint-check-ws` clean
- [x] `cd functional-tests && uv run entry.py -t test_rpc_cors` passes against a live strata + bitcoind regtest

## Deps
Adds `tower-http = "0.6"` (cors feature only) to workspace deps; pulls `tower` + `tower-http` into `bin/strata`.

### DIFF
--- a/bin/strata/src/rpc/mod.rs
+++ b/bin/strata/src/rpc/mod.rs
@@ -10,8 +10,6 @@ use std::sync::Arc;
 
 use anyhow::{Result, anyhow};
 use jsonrpsee::{RpcModule, server::ServerBuilder, types::ErrorObjectOwned};
-use tower::ServiceBuilder;
-use tower_http::cors::CorsLayer;
 use node::*;
 use provider::NodeRpcProvider;
 #[cfg(feature = "sequencer")]
@@ -31,6 +29,8 @@ use strata_ol_rpc_api::{OLClientRpcServer, OLFullNodeRpcServer};
 use strata_primitives::buf::Buf32;
 use strata_status::StatusChannel;
 use strata_storage::NodeStorage;
+use tower::ServiceBuilder;
+use tower_http::cors::CorsLayer;
 
 use crate::run_context::RunContext;
 #[cfg(feature = "sequencer")]
--- a/bin/strata/src/rpc/mod.rs
+++ b/bin/strata/src/rpc/mod.rs
@@ -10,6 +10,8 @@ use std::sync::Arc;
 
 use anyhow::{Result, anyhow};
 use jsonrpsee::{RpcModule, server::ServerBuilder, types::ErrorObjectOwned};
+use tower::ServiceBuilder;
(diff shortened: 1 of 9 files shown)


bin/strata/src/rpc/mod.rs
```diff
@@ -197,7 +199,15 @@ async fn spawn_rpc(deps: RpcDeps) -> Result<()> {
     }
 
     let addr = format!("{}:{}", deps.rpc_host, deps.rpc_port);
+    // Permissive CORS so that browser-based tools (e.g. the OL explorer
+    // hosted at a different origin) can call the JSON-RPC. This is a
+    // developer-facing endpoint with no authenticated state, so allowing
+    // any origin/method/header is appropriate. Adjust if the RPC ever
+    // grows session/cookie state.
+    let cors = CorsLayer::permissive();
```
bewakes: Can we make it by default non permissive and then overridable by env var? That should give best of both worlds.

storopoli:
@codex review

bin/strata/src/rpc/mod.rs
```diff
@@ -89,6 +93,23 @@ impl SeqRpcDeps {
...
+            "1" | "true" | "yes" | "on" => Ok(true),
+            "0" | "false" | "no" | "off" => Ok(false),
+            _ => Err(anyhow!(
+                "{STRATA_RPC_PERMISSIVE_CORS_ENV_VAR} must be one of \
+                 1/true/yes/on or 0/false/no/off"
+            )),
+        },
+        Err(env::VarError::NotPresent) => Ok(false),
+        Err(env::VarError::NotUnicode(_)) => Err(anyhow!(
+            "{STRATA_RPC_PERMISSIVE_CORS_ENV_VAR} must contain valid unicode"
```
storopoli: I'd even say that not unicode but ASCII, since it only accepts 1/true/yes/on or 0/false/no/off

storopoli (approved):
ACK 3fb103545fc8be3f45ced8c5896f7f0eeb1ec900

@delbonis should look this because we are changing some factories and env configs in functional tests.

delbonis (approved):
Looks good!

PR #1754, opened 2026-05-05
Str 3241 (csm): extract checkpoint txid and payload

bewakes:
## Populate real txid/wtxid in checkpoint L1 ref

CSM wrote `Buf32::zero()` for the txid and wtxid of every `CheckpointL1Ref`, so `strata_getCheckpointInfo` returned all-zero hex strings — breaking any explorer/sync use case that needs to find the envelope on Bitcoin.

CSM now scans the L1 block for the SPS-50-tagged checkpoint envelope tx whose payload matches ASM's `CheckpointTipUpdate` (`epoch` + `l2_commitment`), verifies them and persists the real txid/wtxid.

While there, the L1-extracted `CheckpointPayload` is also persisted (in a new sled tree, separate from the sequencer's locally-built payload table) so future checkpoint sync has the data it needs. The payload + L1 ref are written in a single sled transaction.

## Notes
- The PR got a bit bigger than expected, mostly with the testcases and some restructuring in CSM.
- ASM's checkpoint validation logic is duplicated. This is inevitable since there's no indexing of txid from ASM.
- L1 block fetched by blockid, with retry+backoff via `strata_common::retry`.
- Errors propagate; no silent skip — the tip log fires once, so swallowing failures would lose the txid permanently.
- Old observations keep the zero placeholders; no backfill.
- `Buf32` -> `RBuf32` for some of the types also got in.
- AI usage: This PR was implemented with the assistance of Claude Code.

## Tests

- Unit: extractor (match/no-match/multi-tx), processor (happy path, fetch failure, no-match, idempotent), DB (atomic write, overwrite, commitment validation, table separation).
- Functional: `test_checkpoint_finalized` asserts non-zero txid/wtxid via `strata_getCheckpointInfo`. Locally verified: reverting the fix makes the fntest fail on the zero-hash assertion.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist


crates/csm-worker/src/checkpoint_extract.rs
```diff
@@ -0,0 +1,192 @@
...
+            wtxid,
+            payload: envelope.payload,
+        };
+
+        if found.is_some() {
+            error!(
+                ?txid,
+                epoch = expected.epoch,
+                "multiple checkpoint txs match the validated tip; keeping the first"
+            );
```
storopoli, line 97: This matches only `(epoch, l2_commitment)` and then keeps the first matching transaction when duplicates exist. If a block contains two checkpoint envelopes for the same tip, CSM can persist the txid/wtxid for a different envelope than the one ASM accepted. That gets riskier with the `asm/95` path, where verification can distinguish accepted vs rejected checkpoint payloads.

Permalink: https://github.com/alpenlabs/alpen/blob/90381c8108de313235bbe0277b89fd8192fd965d/crates/csm-worker/src/checkpoint_extract.rs#L56-L74

Please fail closed on multiple matches, since that is not expect.

crates/csm-worker/src/processor.rs
```diff
@@ -124,27 +128,41 @@ fn update_client_state_with_checkpoint(
...
+            anyhow::anyhow!(
+                "no checkpoint envelope tx in L1 block {asm_block} matched the validated tip for epoch {}",
+                tip.epoch
+            )
+        })?;
+
+    let observation = CheckpointL1Ref::new(*asm_block, extracted.txid, extracted.wtxid);
+    state
+        .ctx
+        .put_checkpoint_l1_observation(commitment, extracted.payload, observation.clone())?;
```
storopoli: This path now bubbles fetch/extraction errors, but `process_checkpoint_tip_log` has already persisted and published the synthetic client-state update before calling this function. The service then catches the returned error and only logs it, so a transient failure here can advance checkpoint state while permanently losing the txid/wtxid observation that this PR is trying to make durable.

Permalinks:
- https://github.com/alpenlabs/alpen/blob/90381c8108de313235bbe0277b89fd8192fd965d/crates/csm-worker/src/processor.rs#L146-L165
- https://github.com/alpenlabs/alpen/blob/90381c8108de313235bbe0277b89fd8192fd965d/crates/csm-worker/src/processor.rs#L73-L75
- https://github.com/alpenlabs/alpen/blob/90381c8108de313235bbe0277b89fd8192fd965d/crates/csm-worker/src/service.rs#L61-L64

Please either return the error from the service so the status update is retried/fails visibly, or reorder the flow so the L1 observation is fetched and written before the client-state update.

crates/csm-worker/src/checkpoint_extract.rs
```diff
@@ -0,0 +1,192 @@
...
+    CHECKPOINT_SUBPROTOCOL_ID, OL_STF_CHECKPOINT_TX_TYPE, extract_checkpoint_from_envelope,
+};
+use strata_asm_proto_checkpoint_types::{CheckpointPayload, CheckpointTip};
+use strata_l1_txfmt::{MagicBytes, ParseConfig};
+use strata_primitives::buf::Buf32;
+use tracing::{debug, error, warn};
+
+/// Identification of the L1 transaction that carries a validated checkpoint,
+/// along with its decoded payload.
+#[derive(Debug, Clone)]
```
storopoli, line 24: Should definitely derive `Eq` and `PartialEq` here

crates/csm-worker/src/checkpoint_extract.rs
```diff
@@ -0,0 +1,192 @@
...
+            error!(
+                ?txid,
+                epoch = expected.epoch,
+                "multiple checkpoint txs match the validated tip; keeping the first"
+            );
+            continue;
+        }
+        debug!(
+            ?txid,
+            ?wtxid,
```
storopoli, line 105: These can be `Display` logs no?

```suggestion
            %txid,
            %wtxid,
```

crates/csm-worker/src/checkpoint_extract.rs
```diff
@@ -0,0 +1,192 @@
...
+        let wtxid = Buf32::from(tx.compute_wtxid().to_byte_array());
+        let candidate = ExtractedCheckpoint {
+            txid,
+            wtxid,
+            payload: envelope.payload,
+        };
+
+        if found.is_some() {
+            error!(
+                ?txid,
```
storopoli:
```suggestion
                %txid,
```

crates/csm-worker/src/checkpoint_extract.rs
```diff
@@ -0,0 +1,192 @@
...
+                epoch = expected.epoch,
```
storopoli, line 94:
```suggestion
                epoch = %expected.epoch,
```

crates/csm-worker/src/checkpoint_extract.rs
```diff
@@ -0,0 +1,192 @@
...
+        {
+            continue;
+        }
+
+        let tx_input = TxInputRef::new(tx, tag);
+        let envelope = match extract_checkpoint_from_envelope(&tx_input) {
+            Ok(env) => env,
+            Err(e) => {
+                warn!(
+                    txid = ?tx.compute_txid(),
```
storopoli:
```suggestion
                    txid = %tx.compute_txid(),
```

crates/csm-worker/src/processor.rs
```diff
@@ -171,8 +189,8 @@ fn mark_ol_checkpoint_l1_observed(
     debug!(
         ?commitment,
         l1_height = asm_block.height(),
-        ?checkpoint_txid,
-        ?checkpoint_wtxid,
+        txid = ?extracted.txid,
+        wtxid = ?extracted.wtxid,
```
storopoli:
```suggestion
        txid = %extracted.txid,
        wtxid = %extracted.wtxid,
```

storopoli (changes requested):
Nice start. Some nits and some soundness issues.

bewakes:
@codex review

crates/csm-worker/src/processor.rs
```diff
@@ -124,27 +131,63 @@ fn update_client_state_with_checkpoint(
...
 
-    let observation = CheckpointL1Ref::new(*asm_block, checkpoint_txid, checkpoint_wtxid);
-    ol_checkpoint.put_checkpoint_l1_ref_blocking(commitment, observation.clone())?;
+    // ASM signed off on this block, so a fetch failure means the bitcoind RPC is
+    // unreachable beyond what `get_l1_block`'s internal retry can absorb. Bubble
+    // the error so the operator notices instead of losing the txid silently.
+    let block = state
+        .ctx
+        .get_l1_block(asm_block.blkid())
+        .with_context(|| format!("fetching L1 block {asm_block} for checkpoint observation"))?;
```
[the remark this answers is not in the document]
bewakes: This should be fixed now.
@codex reveiw

crates/csm-worker/src/checkpoint_extract.rs
bewakes: Good observation. This has been fixed in https://github.com/alpenlabs/alpen/pull/1754/commits/0b5c9eefbbb4b34918426d7e479e70d230db2399 where we do the actual checkpoint verfication.

crates/csm-worker/src/processor.rs
bewakes: This and the nits below have been fixed in https://github.com/alpenlabs/alpen/pull/1754/commits/b9552786f07de7b24eebed9374dd34c72ffb8d74

crates/csm-worker/src/checkpoint_extract.rs
bewakes: Haven't come across the requirement yet. So maybe not until we need it?

crates/csm-worker/src/processor.rs
bewakes: This was something codex on this PR also found out.

crates/csm-worker/src/service.rs
```diff
@@ -45,21 +54,36 @@ impl SyncService for CsmWorkerService {
...
             if let Err(e) = process_log(state, log, &asm_block) {
                 error!(%asm_block, err = %e, "Failed to process ASM log");
+                block_ok = false;
+                break;
+            }
+        }
+
+        if block_ok {
+            if let Err(e) = commit_block(state, asm_block) {
+                error!(%asm_block, err = %e, "Failed to commit CSM block");
```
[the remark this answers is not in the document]
storopoli: This is a valid finding

storopoli (changes requested):

crates/csm-worker/src/service.rs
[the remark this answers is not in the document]
bewakes: Fixed in https://github.com/alpenlabs/alpen/pull/1754/commits/6fd3a62fe1f5bb5cce4798e76c1220d3a063fa1b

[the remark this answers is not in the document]
bewakes: Yes, fixed in https://github.com/alpenlabs/alpen/pull/1754/commits/6fd3a62fe1f5bb5cce4798e76c1220d3a063fa1b

bewakes:
@codex review

crates/db/store-sled/src/ol_checkpoint/db.rs
```diff
@@ -389,6 +390,80 @@ impl OLCheckpointDatabase for OLCheckpointDBSled {
...
+        );
+        if commitment != expected_commitment {
+            return Err(DbError::InvalidArgument);
+        }
+
+        self.config.with_retry(
+            (&self.l1_observed_payload_tree, &self.l1_ref_tree),
+            |(opt, lot)| {
+                opt.insert(&commitment, &payload)?;
+                lot.insert(&commitment, &l1_ref)?;
```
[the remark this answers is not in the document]
bewakes: Fixed in https://github.com/alpenlabs/alpen/pull/1754/commits/868d5c2476210f328864040b87c8ed7ac6f33e81

crates/csm-worker/src/checkpoint_extract.rs
storopoli: Yes agreed

crates/csm-worker/src/checkpoint_extract.rs
```diff
@@ -0,0 +1,394 @@
...
+use strata_l1_txfmt::{MagicBytes, ParseConfig};
+use strata_primitives::buf::Buf32;
+use tracing::{debug, warn};
+
+/// Identification of the L1 transaction that carries a validated checkpoint,
+/// along with its decoded payload.
+#[derive(Debug, Clone)]
+pub(crate) struct ExtractedCheckpoint {
+    pub(crate) txid: Buf32,
+    pub(crate) wtxid: Buf32,
```
storopoli: These should be `RBuf32` because of #1815

crates/csm-worker/src/state.rs
```diff
@@ -209,65 +199,18 @@ mod tests {
...
-            run: SyncParams {
-                l1_follow_distance: 10,
-                client_checkpoint_interval: 100,
-                l2_blocks_fetch_limit: 1000,
-            },
-        })
+        Arc::new(strata_test_utils_l2::gen_params())
     }
 
     fn create_test_storage_and_status(
```
storopoli, line 213: If possible also change `Buf32` to `RBuf32` here as well.

storopoli (changes requested):
We need to STOP using `Buf32` for bitcoin txid and block hashes. Use `RBuf32`, since it reverses the hex-string in `Display`/`Debug` impls.

Can you also make sure that you're also using `RBuf32` in any database schema that uses `Buf32` for these use cases as well? 

Finally, can you make sure that we are logging all `RBuf32` txid and block hashes with `Debug` (e.g. `?txid`) instead of `Display` (e.g. `%txid` or just `txid`) inside tracing statements because of the hex-string truncation?

crates/csm-worker/src/processor.rs
```diff
@@ -109,42 +117,146 @@ fn update_client_state_with_checkpoint(
...
+    logs: &[AsmLogEntry],
+) -> anyhow::Result<()> {
+    let cur_state_snapshot = state.cur_state.clone();
 
-    // Update our tracked state
-    state.cur_state = Arc::new(next_state);
+    for log in logs {
+        if let Err(e) = process_log(state, log, &asm_block) {
+            state.cur_state = cur_state_snapshot;
+            return Err(e).with_context(|| format!("processing ASM log for block {asm_block}"));
```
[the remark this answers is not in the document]
bewakes: Fixed in https://github.com/alpenlabs/alpen/pull/1754/commits/8981725d2b1dd0dc909eeda97b4884bb4a99905b

bewakes:
@storopoli PTAL https://github.com/alpenlabs/alpen/pull/1754/commits/8d7ee88df994e332d8a1b1feece8bb338e21af3e

bewakes:
@codex review

bin/strata/src/rpc/node.rs
```diff
@@ -636,7 +636,13 @@ impl<P: OLRpcProvider> OLClientRpcServer for OLRpcServer<P> {
             .await
             .map_err(db_error)?;
         let confirmation_status = if let Some(obs) = l1_ref {
-            let l1_reference = RpcCheckpointL1Ref::new(obs.l1_commitment, obs.txid, obs.wtxid);
+            // `RpcCheckpointL1Ref` still carries `Buf32`; convert at this RPC
+            // boundary. `RBuf32`/`Buf32` share byte layout, only fmt/serde differ.
+            let l1_reference = RpcCheckpointL1Ref::new(
+                obs.l1_commitment,
+                Buf32::from(obs.txid.0),
+                Buf32::from(obs.wtxid.0),
```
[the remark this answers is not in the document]
storopoli: I know that this is a little bit out of the scope of the PR. But what do you think about changing `RpcCheckpointL1Ref` to use `RBuf32` as well? I did play around in `main` and it seems not a big lift.

crates/csm-worker/src/processor.rs
```diff
@@ -109,70 +117,181 @@ fn update_client_state_with_checkpoint(
...
+    let target_height = asm_block.height();
+
+    // Already processed (duplicate status) or behind us (reorg) — nothing to do.
+    if target_height <= last_height {
+        debug!(
+            %asm_block,
+            last_height,
+            "ASM status block not ahead of last committed block; skipping"
+        );
+        return Ok(());
```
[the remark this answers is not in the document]
bewakes: Check https://github.com/alpenlabs/alpen/pull/1754/commits/b26c6f055b8a0fa17b86b3bedd1b01704b035869. Reorg handling is left to subsequent ticket.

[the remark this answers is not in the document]
bewakes: @codex reveiw

crates/csm-worker/src/processor.rs
```diff
@@ -109,70 +117,199 @@ fn update_client_state_with_checkpoint(
...
+        debug!(%asm_block, "ASM status block matches last committed; skipping");
+        return Ok(());
+    }
 
+    // TODO(STRA-3466): Strictly behind — either a stale message or a deeper reorg.
+    if target_height < last_height {
+        warn!(
+            %asm_block,
+            last_height,
+            "ASM block strictly behind last committed; skipping (possible deeper reorg, see TODO)"
```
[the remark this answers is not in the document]
bewakes: This is left for STR-3466

bin/strata/src/rpc/node.rs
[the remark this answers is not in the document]
bewakes: PTAL at https://github.com/alpenlabs/alpen/pull/1754/commits/be005f35c0ef63a7897011e86a2ad7c675f7c2ff

storopoli (approved):
ACK 38fa7fae5fd91052cfe9ab3e3e7def6de1183ef4

crates/consensus-logic/src/csm_worker_context.rs
```diff
@@ -0,0 +1,126 @@
...
+            handle,
+            bitcoin_client,
+            params,
+            storage,
+            status_channel,
+        }
+    }
+}
+
+impl CsmWorkerContext for CsmWorkerCtx {
```
delbonis: This is kinda ambiguous, rename the impl to just be `CsmWorkerContextImpl` or similar.

crates/consensus-logic/src/csm_worker_context.rs
```diff
@@ -0,0 +1,126 @@
...
+        // ASM has already processed this block, so bitcoind must have it. Retry
+        // with backoff to absorb transient failures (RPC timeout, replica lag).
+        let hash = blockid.to_block_hash();
+        let backoff = ExponentialBackoff::new(200, 15, 10);
+        retry_with_backoff("csm_get_l1_block", 10, &backoff, || {
+            self.handle
+                .block_on(self.bitcoin_client.get_block(&hash))
+                .map_err(|e| anyhow::anyhow!("fetch L1 block {blockid}: {e}"))
+        })
+    }
```
delbonis, line 95: Ahh this is kinda crazy.  Maybe we should move towards abstracting away the low-level provider.

Not actionable, something we can think about later.

crates/consensus-logic/src/csm_worker_context.rs
delbonis, line 1: Change the `anyhow::anyhow!` usages to typed errors, wrapping the low-level errors where appropriate.

crates/csm-types/src/client_state.rs
```diff
@@ -106,17 +106,20 @@ impl CheckpointState {
 }
 
 /// Represents a reference to a transaction in bitcoin. Redundantly puts block_height a well.
+///
+/// `txid` and `wtxid` use [`RBuf32`] so their `Debug`/`Display` follow Bitcoin's
+/// reversed-byte hash convention.
 #[derive(
-    Clone, Eq, PartialEq, Arbitrary, BorshDeserialize, BorshSerialize, Deserialize, Serialize,
+    Clone, Debug, Eq, PartialEq, Arbitrary, BorshDeserialize, BorshSerialize, Deserialize, Serialize,
```
delbonis, line 113: Is it feasible for us to get rid of borsh here?

Is this important for consensus or should we aim for this to be future-compatible?

crates/csm-worker/src/checkpoint_extract.rs
```diff
@@ -0,0 +1,432 @@
...
+        checkpoint_state.sequencer_predicate(),
+        &envelope.envelope_pubkey,
+    )
+    .and_then(|_| {
+        verify_progression(
+            checkpoint_state.verified_tip(),
+            envelope.payload.new_tip(),
+            current_l1_height,
+        )
+    })?;
```
delbonis, line 138: Is this call necessary now that we've reworked `CheckpointState` to do more bookkeeping internally?

crates/csm-worker/src/processor.rs
```diff
@@ -2,18 +2,25 @@
...
+    state::CsmWorkerState,
+};
 
-pub(crate) fn process_log(
-    state: &mut CsmWorkerState,
+pub(crate) fn process_log<C: CsmWorkerContext>(
+    state: &mut CsmWorkerState<C>,
     log: &AsmLogEntry,
     asm_block: &L1BlockCommitment,
 ) -> anyhow::Result<()> {
```
delbonis: Can this just be a member fn on `CsmWorkerState`, to simplify the ergonomics?

Same with some of the rest of these.  Any fns that always do `state.ctx` and don't actually touch anything else in the state should *not* actually be merged, we can leave them as-is but change the args to take `ctx: &impl CsmWorkerContext`.  Think about the structure of the processing steps.

crates/csm-worker/src/processor.rs
```diff
@@ -109,73 +118,202 @@ fn update_client_state_with_checkpoint(
...
+        tip,
+        &mut checkpoint_state,
+        asm_block.height(),
+        &verified_aux_data,
+    )
+    .ok_or_else(|| {
+        anyhow::anyhow!(
+            "no checkpoint envelope tx in L1 block {asm_block} validated for epoch {}",
+            tip.epoch
+        )
```
delbonis: This and other places.

crates/csm-worker/src/processor.rs
```diff
@@ -109,73 +118,202 @@ fn update_client_state_with_checkpoint(
...
+    Ok(())
+}
 
-    // Update our tracked state
-    state.cur_state = Arc::new(next_state);
+/// Processes every log of a single ASM block and commits it as one unit.
+///
+/// All staged state is snapshotted up front and restored on any failure
+/// including a `commit_block` failure.
+fn process_block<C: CsmWorkerContext>(
```
delbonis: This isn't really processing a *block*.  It's processing an ASM manifest's logs (naturally corresponding to a block).  Consider passing around the manifest itself in more places instead of the slice of logs.

We should clarify/reframe the naming/docs around this.

crates/csm-worker/src/state.rs
```diff
@@ -7,32 +7,42 @@ use strata_identifiers::Epoch;
...
 
-    /// Node storage handle.
-    pub(crate) storage: Arc<NodeStorage>,
+    /// Last ASM block committed. Advanced only by a successful commit; never
+    /// rolled back.
+    pub(crate) last_asm_block: Option<L1BlockCommitment>,
+
+    /// State staged while processing the current ASM block's logs; committed as
+    /// a unit by `commit_block` or rolled back on any failure.
+    pub(crate) staged: StagedState,
```
delbonis: Why is this stored as part of the state instead of just being passed around ephemerally in the input processing fns?  Shouldn't we only store the last committed state here so we can expose it in `get_status`?

crates/csm-worker/src/state.rs
```diff
@@ -47,18 +57,15 @@ pub struct CsmWorkerState {
...
-    pub fn new(
-        params: Arc<Params>,
-        storage: Arc<NodeStorage>,
-        status_channel: Arc<StatusChannel>,
-    ) -> anyhow::Result<Self> {
+    ///
+    /// `params` and `storage` are read once at startup to bootstrap the
+    /// in-memory finalization queue; runtime persistence and L1 fetches go
+    /// through `ctx`.
+    pub fn new(params: Arc<Params>, storage: Arc<NodeStorage>, ctx: C) -> anyhow::Result<Self> {
```
delbonis: Rework this so that it loads everything from the context.  The red flag that there's something architecturally wrong here is that `Params` and `NodeStorage` are being accessed from both inside and outside the context trait, indicating that we're bypassing the abstraction we're defining and undermining it.

delbonis (changes requested):
Functionally looks good but lots of design cleanliness issues.

crates/csm-types/src/client_state.rs
bewakes: I tried it, but it drags in changes all the way upto the database. Let's tackle this in a separate and focused "remove borsh" task.

crates/csm-worker/src/checkpoint_extract.rs
bewakes: The bookkeeping seems to happen via `state.advance()`. However, in order to call that we need all of these calls. Trying to mimic how this is being done in asm here. Though it would be ideal if the verification was much simpler to use outside asm.

cc: @prajwolrg

crates/csm-worker/src/processor.rs
bewakes: Yes, it can be a method in `CsmWorkerState`. Kept the module as is and added a new `impl` block there. Moved major process functions into the impl. Check https://github.com/alpenlabs/alpen/pull/1754/commits/87727f18dbddb3249acca24162fcaeaa98c2da14

crates/csm-worker/src/processor.rs
bewakes: True, renamed the function name to `process_asm_logs` which is more accurate.

> Consider passing around the manifest itself in more places instead of the slice of logs.

This would be ideal, but the inputs from ASM don't include manifests themselves which is richer(also includes blk height and witness root). So, kept the arguments passed as they are.

crates/csm-worker/src/state.rs
bewakes: Actually not all that's contained are ephemeral. Those were grouped because I wanted to snapshot them prior to processing logs. But now I have split it more meaningfully - flattening out non-ephemeral fields and passing the ephemeral values.
Split in https://github.com/alpenlabs/alpen/pull/1754/commits/abb009e1677f188178d410047aad2270a6743888

crates/csm-worker/src/state.rs
bewakes: yeah, this is a real smell. Thanks for pointing out.
Fixed in https://github.com/alpenlabs/alpen/pull/1754/commits/f43dcf3ce0c825ac8902eb0e1c4342b7e082f33e

crates/consensus-logic/src/csm_worker_context.rs
bewakes: What do you think about doing this in a separate ticket that focuses on better errors representation in csm? This PR is already so big as some of the refactors got dragged in.
If that works for you I can create a ticket right away and prioritize over the coming sprint.

storopoli: Yes, separate ticket. In my previous review I almost flagged `anyhow!` twice but then I realized that this is out of the scope of the PR since `consensus-logic` already used a bunch of `anyhow!`. Can you please open a ticket and add a `// TODO(STR-XXXX)` comment here?

crates/csm-types/src/client_state.rs
storopoli: Same here `TODO(STR-XXXX)`

storopoli (approved):
ACK ed3d695f4cdf6f3953a4cbf55ef7847d32c6b454

delbonis (approved):

storopoli:
pinging @prajwolrg for review

functional-tests/tests/checkpoint/test_checkpoint_finalized.py
```diff
@@ -48,4 +50,20 @@ def main(self, ctx):
...
+            assert status["status"] == "finalized", (
+                f"epoch {target_epoch} not finalized in checkpoint info: {status}"
+            )
+            l1_ref = status["l1_reference"]
+            assert l1_ref["txid"] != ZERO_HASH, (
+                f"epoch {target_epoch} l1_reference has zero txid: {l1_ref}"
+            )
+            assert l1_ref["wtxid"] != ZERO_HASH, (
+                f"epoch {target_epoch} l1_reference has zero wtxid: {l1_ref}"
+            )
```
prajwolrg, line 68: Instead of just checking whether they are non-zero, we should fetch them from the bitcoin itself and see if they exist.

prajwolrg (dismissed):
Great work! LGTM!

Only a small blocking comment.

crates/csm-worker/src/checkpoint_extract.rs
prajwolrg: The `verify_progression` shouldn't be needed. Since we already know that the new tip is valid, we shouldn't need this. We can instead use the DB to fetch the manifest directly. They don't have to be `VerifiedAuxData` as well. I created a new [ticket](https://alpenlabs.atlassian.net/browse/STR-3532) to track this.

functional-tests/tests/checkpoint/test_checkpoint_finalized.py
storopoli: @bewakes I agree. This is the only blocker for this PR.

bewakes: Thanks for the suggestion! Done in https://github.com/alpenlabs/alpen/pull/1754/commits/2b286fdd0e95d8e84d9fa3c43771228f27f13f1e

storopoli (approved):
ACK 2b286fdd0e95d8e84d9fa3c43771228f27f13f1e

prajwolrg (approved):

PR #1757, opened 2026-05-06
(alpen-ee): Use correct bridge gateway account in sequencer config

bewakes:
## Description

Use correct bridge gateway account(`0x10`) instead of incorrect value (`1`) being used.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/alpen-ee/sequencer/src/block_builder/config.rs
+++ b/crates/alpen-ee/sequencer/src/block_builder/config.rs
@@ -7,7 +7,8 @@ const DEFAULT_BLOCKTIME_MS: u64 = 5_000;
 /// Default number of deposits to process per ee block.
 const DEFAULT_DEPOSITS_PER_BLOCK: NonZeroU8 = NonZeroU8::new(16).expect("16 is always NonZero");
 /// Default bridge gateway account on OL.
-const DEFAULT_BRIDGE_GATEWAY_ACCOUNT: AccountId = AccountId::special(1); // TODO: correct account id
+// TODO(STR-3358): Don't hardcode here
+const DEFAULT_BRIDGE_GATEWAY_ACCOUNT: AccountId = AccountId::special(0x10);
 /// Default time to wait on errors during block building.
 const DEFAULT_ERROR_BACKOFF_MS: u64 = 100;
 


crates/alpen-ee/sequencer/src/block_builder/config.rs
```diff
@@ -7,7 +7,8 @@ const DEFAULT_BLOCKTIME_MS: u64 = 5_000;
 /// Default number of deposits to process per ee block.
 const DEFAULT_DEPOSITS_PER_BLOCK: NonZeroU8 = NonZeroU8::new(16).expect("16 is always NonZero");
 /// Default bridge gateway account on OL.
-const DEFAULT_BRIDGE_GATEWAY_ACCOUNT: AccountId = AccountId::special(1); // TODO: correct account id
+const DEFAULT_BRIDGE_GATEWAY_ACCOUNT: AccountId = AccountId::special(0x10); // TODO(STR-3358): Don't
+                                                                            // hardcode here
```
storopoli: Why not use this guy here instead of having a `const` being defined (and out-of-sync) in two places?
https://github.com/alpenlabs/alpen/blob/34199d3696dda817b88c929599d0df7794ed030c/crates/ol/stf/src/constants.rs#L5-L6

storopoli (changes requested):

crates/alpen-ee/sequencer/src/block_builder/config.rs
storopoli: Actually we should just add it to `strata_acct_types` and update in `crates/ol/stf/src/constants.rs` and here also...

bewakes: That's because we don't want to have ol stf dependency in alpen client. See the TODO ticket.

storopoli: Yes, I do generally agree with that. But this is also not a good solution because we'll have the same issue in the future once more...

I'm approving this but I don't know if this is the right approach yet.

storopoli (approved):
ACK 7c76a117e8424a72d369222a92187882b55d89c2

crates/alpen-ee/sequencer/src/block_builder/config.rs
bewakes: There are other places where we would like to use acct constants from a common place. We can do that in a single ticket. See the discussion here: https://alpen-labs.slack.com/archives/C08G7PL1RBK/p1774511230877389?thread_ts=1774511230.877389&cid=C08G7PL1RBK

krsnapaudel (commented):
Nit about the two-line comment for TODO. Probably better to put it in a line above.

storopoli (approved):
ACK 1b18e95467860f67e83a48871bfed8a503257088

PR #1760, opened 2026-05-06
STR 2066 Revise and harden OL STF tests

krsnapaudel:
## Description

Refactor, revise and harden OL STF tests. Add tests to fill existing gaps. 
- **Module reorganization**: splits `tests/stf.rs` and `tests/edge_cases.rs` into themed modules (`verify_header`, `verify_body`, `staging_layers`, `sau_transfers`, `sau_messages`, `sau_limits`, `sau_sequences`, `sau_logs`, `chain_processing`, `asm_manifests`, `tx_constraints`, `partial_execution`, `stress`, `sau_validation`).
- **Behavior fixture layer**: introduces `OLStfFixture::builder()` + `OLStfFixtureBuilder` for pre-genesis setup, `FixtureBlockBuilder` for child-block sequencing and tx construction, SAU/GAM/manifest fixture builders, `StateSnapshot::assert_unchanged` for state-unchanged pinning, and `StateTestExt: IStateAccessor` for state-lookup ergonomics. Raw verifier/staging/tamper helpers stay as free functions over `&mut impl IStateAccessorMut`.
- **Production fixes**:
  - Rejects SAU updates with unconsumed proofs in `transaction_processing`.
  - Replaces broad `ExecError::ChainIntegrity` with finer variants: `AsmManifestHeightMismatch { expected, actual, index }` in `manifest_processing` and `HeaderEpochMismatch` / `ContextEpochMismatch` in `chain_processing`. Downstream callers updated.
- **Existing-test hardening**: positive controls on tamper tests, state-unchanged assertions via snapshots on SAU/inbox/ledger-reference validation tests, bridge-gateway silent-drop pinned across 5 drop variants (including limbo-amount checks), body-root mismatch revived, mid-block partial execution semantics pinned, zero-balance boundary asserted including transfer-after-drain, supply-cap tests use `BitcoinAmount::MAX_MONEY` and `BitcoinAmount::MAX` with intent comments, limbo bookkeeping asserted in deposit/withdrawal and ASM manifest tests.
- **New coverage**: chain processing error branches, GAM structural validation, ASM manifest processing boundaries (height gap, height ≤ last L1, max manifest count, malformed deposit log, unknown-serial deposit limboing, checkpoint tip update), log-root padding parity, epoch preseal and post-epoch diff verification, dependent SAU sequences, replay rejection across blocks, `TxConstraints` slot boundaries, ledger-reference proof rejection paths, multi-SAU same-block scenarios (distinct senders, sequential seqnos, duplicate seqno failure), staging-layer execution paths (`WriteTrackingState`, `IndexerState`), SAU and log boundary limits, SAU target-type and sequence-number bounds, block terminality verification, output message delivery to snark inboxes, SAU log emission contract, ignored stress tests for large inbox/tx batches.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Please review the doc below and the test added.

- [x]  Yes
- [ ]  No

[OL STF test cases](https://www.notion.so/OL-STF-test-cases-2e3901ba000f80598da0fceaeb62dd5e?source=copy_link)

## Checklist

## Related Issues

[STR-2066](https://alpenlabs.atlassian.net/browse/STR-2066)

[STR-2066]: https://alpenlabs.atlassian.net/browse/STR-2066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


krsnapaudel:
@barakshani Added tests with tx staging. See `staging_layers.rs`.

AlpenCodex:
@codex Please review this PR

(This is a test being run by @sampkaALP (security))

delbonis:
Merge conflict because of the limbo PR that just went in.

delbonis:
This is good work, but there's still a lot of repeating patterns with tests.  I would love it if we could think of some declaritive model for describing all the different test scenarios which would make it easier to write tests by handling the setup/asserts better, and when we make changes to how things work we don't have to change tons of tests that interact with it.  I am not sure what that would look like at the moment.

krsnapaudel:
> This is good work, but there's still a lot of repeating patterns with tests. I would love it if we could think of some declaritive model for describing all the different test scenarios which would make it easier to write tests by handling the setup/asserts better, and when we make changes to how things work we don't have to change tons of tests that interact with it. I am not sure what that would look like at the moment.

@delbonis Yes I am aware. I will try to come up with some ways to reduce repetition with AI help.

delbonis:
wow I would have assumed this would have caused more conflicts

crates/ol/stf/src/test_utils.rs
delbonis, line 1: REALLY need more docs on how these fixtures work and are expected to be used.  I can look at the existing tests, but that's trying to reconstruct the intended usage from examples, when it should be explained more thoroughly upfront.

crates/ol/stf/src/test_utils.rs
```diff
@@ -481,6 +490,951 @@ pub fn get_test_proof(variant: u8) -> Vec<u8> {
...
+        Self {
+            sender,
+            builder,
+            state_root: get_test_state_root(seqno as u8 + 2),
+            proof: get_test_proof(1),
+            should_advance_seqno: true,
+        }
+    }
+
+    fn unchecked(sender: AccountId, seqno: u64) -> Self {
```
delbonis: Should be named like `new_unchecked`.

crates/ol/stf/src/test_utils.rs
```diff
@@ -677,14 +1670,46 @@ impl SnarkUpdateBuilder {
             ledger_ref_claims: vec![],
             ledger_ref_proofs: vec![],
             extra_data: VariableList::default(),
+            next_msg_idx_override: None,
         }
     }
 
+    /// Creates an unchecked builder when no snark account state is available.
+    fn unchecked(seq_no: u64, old_msg_idx: u64) -> Self {
```
delbonis: Should be named like `new_unchecked`.

crates/ol/stf/src/test_utils.rs
```diff
@@ -809,14 +1843,31 @@ impl SnarkUpdateBuilder {
...
+pub fn lookup_snark_state(
+    state: &MemoryStateBaseLayer,
+    snark_id: AccountId,
+) -> &OLSnarkAccountState {
+    state
+        .get_account_state(snark_id)
+        .unwrap()
+        .unwrap()
+        .as_snark_account()
+        .unwrap()
```
delbonis: Calling these `lookup_snark_state` and `snark_id` is misleading.  I realize these's some naming uncertainty with `lookup_snark_account_state`.  Consider avoiding making overly-specific helpers and rework how things are defined, like adding local extension traits for common behavior.

Also, impl these for `IStateAccessor`/`IStateAccessorMut` instead of privileging a specific impl.  This is a common mistake I see in lots of places.

Also, these `.unwrap()`s maybe should print more elaborate error messages about *what* went wrong when something went wrong, including the account ID.

crates/ol/stf/src/test_utils.rs
```diff
@@ -481,6 +490,951 @@ pub fn get_test_proof(variant: u8) -> Vec<u8> {
...
+            proof: get_test_proof(1),
+            should_advance_seqno: true,
+        }
+    }
+
+    /// Adds a transfer effect.
+    pub fn transfer(mut self, dest: AccountId, amount: BitcoinAmount) -> Self {
+        self.builder = self.builder.with_transfer(dest, amount.to_sat());
+        self
+    }
```
delbonis, line 1318: This and the other fns are named like they either do something on their own or are accessors.  But they're actually returning `Self` like `with` fns would do.  Should make this more idiomatic.

crates/ol/stf/src/test_utils.rs
```diff
@@ -527,6 +1481,20 @@ impl InboxMmrTracker {
         }
     }
 
+    /// Returns the current raw proof for a tracked inbox entry.
+    pub fn raw_proof(&self, index: usize) -> RawMerkleProof {
```
delbonis: `get_idx_raw_proof` or something more clear

crates/ol/stf/src/test_utils.rs
```diff
@@ -587,32 +1555,51 @@ impl ManifestMmrTracker {
...
-        .expect("Should create snark account");
+    setup_genesis_with_snark_accounts(state, &[(snark_id, initial_balance)])
+}
+
+/// Creates SNARK accounts with initial balances and executes genesis.
+///
+/// Genesis carries an empty manifest container, matching production genesis.
+/// Returns the completed genesis block.
+pub fn setup_genesis_with_snark_accounts(
+    state: &mut MemoryStateBaseLayer,
```
delbonis: `&mut impl IStateAccessorMut`

crates/ol/stf/src/test_utils.rs
```diff
@@ -481,6 +490,951 @@ pub fn get_test_proof(variant: u8) -> Vec<u8> {
...
+        L1BlockId::from(Buf32::from([variant; 32])),
+        WtxidsRoot::from(Buf32::from([variant; 32])),
+        vec![deposit_log_for_account(
+            account_serial,
+            dest_subject,
+            amount,
+        )],
+    )
+    .expect("deposit manifest should be valid")
+}
```
delbonis: This and similar fns should have verby names, in general but *especially* here since "deposit" is a verb.  Should add `make_` prefixes or something.

crates/ol/stf/src/test_utils.rs
```diff
@@ -481,6 +490,951 @@ pub fn get_test_proof(variant: u8) -> Vec<u8> {
...
+    let body = encode_to_vec(&withdrawal).expect("withdrawal message should encode");
+    OwnedMsg::new(WITHDRAWAL_MSG_TYPE_ID, body)
+        .expect("withdrawal message should be valid")
+        .to_vec()
+}
+
+/// Creates a [`BitcoinAmount`] from satoshis.
+pub fn sats(amount: u64) -> BitcoinAmount {
+    BitcoinAmount::from_sat(amount)
+}
```
delbonis: Is this really necessary?

delbonis (changes requested):
There's a lot of good things here to improve ergonomics, however there's also a lot of smelly problems (like "being too specific without reason") and unidiomatic/misleading names that should be improved.

Also in most of the tests we can probably just do `use ...::test_utils::*;` instead of importing each and every item, since that just creates more work to keep updated (either for humans or LLMs).

crates/ol/stf/src/test_utils.rs
krsnapaudel: Not really. Removed.

crates/ol/stf/src/tests/asm_manifests.rs
```diff
@@ -0,0 +1,149 @@
...
+    let result = execute_block(
+        &mut state,
+        &BlockInfo::new_genesis(1_000_000),
+        None,
+        BlockComponents::new_manifests(vec![asm_manifest]),
+    );
+
+    match result {
+        Err(e) => assert!(
+            matches!(e.into_base(), ExecError::ChainIntegrity),
```
bewakes: `ChainIntegrity` seems to be a bit generic and less informative here because there are many ways chain integrity can be violated. Do you think we can have finer variants or wrapping other errors inside `ChainIntegrity` here?

crates/ol/stf/src/test_utils.rs
```diff
@@ -481,6 +490,951 @@ pub fn get_test_proof(variant: u8) -> Vec<u8> {
...
+#[derive(Debug, Clone)]
+pub struct FixtureAsmManifestBuilder {
+    height: L1Height,
+    variant: u8,
+    logs: Vec<AsmLogEntry>,
+}
+
+impl FixtureAsmManifestBuilder {
+    /// Starts building a deterministic ASM manifest at `height`.
+    pub fn new(height: L1Height) -> Self {
```
bewakes: Better name can be `new_at_height`. It's not clear what it does just by seeing `FixtureAsmManifestBuilder::new(1)`.

crates/ol/stf/src/tests/asm_manifests.rs
```diff
@@ -0,0 +1,149 @@
...
+    let state = output.fixture().state();
+
+    assert_eq!(output.log_count(), 0);
+    assert_eq!(state.cur_epoch(), 1);
+    assert_eq!(state.last_l1_height(), 1);
+    assert_eq!(state.asm_manifests_mmr().num_entries(), 1);
+}
+
+#[test]
+fn test_manifest_processing_skips_unknown_serial_deposit() {
```
bewakes, line 153: Should the deposit go to limbo balance?

crates/ol/stf/src/tests/asm_manifests.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    assert_eq!(
+        state.asm_recorded_epoch(),
+        &EpochCommitment::from_terminal(7, l2_commitment)
+    );
+    assert_eq!(state.last_l1_height(), 1);
+}
+
+#[test]
+fn test_manifest_processing_skips_malformed_deposit_log() {
```
bewakes, line 203: This should also check that limbo account gets updated.

crates/ol/stf/src/tests/asm_manifests.rs
```diff
@@ -0,0 +1,149 @@
...
+    AccountSerial, Buf32, EpochCommitment, OLBlockCommitment, OLBlockId, SubjectId,
+};
+use strata_ledger_types::{IAccountState, ISnarkAccountState, IStateAccessor};
+use strata_msg_fmt::MAX_TYPE;
+use strata_ol_chain_types_new::MAX_SEALING_MANIFEST_COUNT;
+
+use crate::{assembly::BlockComponents, context::BlockInfo, errors::ExecError, test_utils::*};
+
+#[test]
+fn test_manifest_processing_rejects_height_gap() {
```
bewakes, line 38: Perhaps also a good idea to think about/test when manifest has height <= `state.last_l1_height()` ?

crates/ol/stf/src/tests/chain.rs
```diff
@@ -0,0 +1,466 @@
...
+}
+
+#[test]
+fn test_process_chain_with_multiple_epochs() {
+    // Test that we can process a chain from genesis through epoch 3
+    // with 3 blocks per epoch (epochs 0, 1, 2, 3 = 12 blocks total)
+    let mut state = create_test_genesis_state();
+    const BLOCKS_PER_EPOCH: u64 = 3;
+    const _TARGET_EPOCH: u32 = 3;
+    const TOTAL_BLOCKS: usize = 12; // 4 epochs * 3 blocks each
```
bewakes: nit: derive the value from consts defined in lines above..

crates/ol/stf/src/tests/chain_processing.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+#[test]
+fn test_epoch_initial_rejects_epoch_mismatch() {
+    let mut state = create_test_genesis_state();
+    let context = EpochInitialContext::new(1, OLBlockCommitment::null());
+
+    let err = process_epoch_initial(&mut state, &context).unwrap_err();
+
+    assert!(matches!(err, ExecError::ChainIntegrity));
```
bewakes: Similar to above, might be a good idea to make the error more precise.

crates/ol/stf/src/tests/deposit_withdrawal.rs
```diff
@@ -1,227 +1,222 @@
...
+    );
+    assert_eq!(
+        withdrawal_log.selected_operator,
+        u32::MAX,
+        "Withdrawal operator should preserve the any-operator sentinel"
+    );
+}
+
+#[test]
+fn test_bridge_gateway_direct_transfer_is_silently_dropped() {
```
bewakes, line 96: Should also check limbo account update here.

crates/ol/stf/src/tests/deposit_withdrawal.rs
```diff
@@ -1,227 +1,222 @@
...
+        })
+        .execute_with_outputs();
+
+    assert_no_bridge_gateway_logs(&output);
+    assert_eq!(fixture.account_balance(snark_id), sats(90_000_000));
+    assert_eq!(*fixture.snark(snark_id).seqno().inner(), 1);
+}
+
+#[test]
+fn test_bridge_gateway_non_denomination_withdrawal_is_silently_dropped() {
```
bewakes, line 132: Need to check limbo amount is updated.

crates/ol/stf/src/tests/deposit_withdrawal.rs
bewakes, line 1: Most of the tests here also need to check limbo amount is updated instead of just silently dropped.

crates/ol/stf/src/tests/sau_limits.rs
```diff
@@ -0,0 +1,347 @@
...
+}
+
+#[test]
+fn test_snark_update_single_transfer_exceeding_max_bitcoin_suceeds() {
+    let snark_id = get_test_snark_account_id();
+    let recipient_id = get_test_recipient_account_id();
+
+    let more_than_max_bitcoin = 2_100_000_000_000_001u64;
+    let mut fixture = GenesisBuilder::new()
+        .with_snark_account(snark_id, |acct| acct.with_balance(sats(u64::MAX)))
```
bewakes: Actually this should not be allowed. The type is `BitcoinAmount` so I believe the construction should consider max supply. On a quick glance I can see that `BitcoinAmount` just accepts any u64.

crates/ol/stf/src/tests/sau_limits.rs
```diff
@@ -0,0 +1,347 @@
...
+}
+
+#[test]
+fn test_snark_update_rejects_aggregate_transfer_overflow() {
+    let snark_id = get_test_snark_account_id();
+    let recipient1_id = test_account_id(TEST_RECIPIENT_ID + 1);
+    let recipient2_id = test_account_id(TEST_RECIPIENT_ID + 2);
+
+    let mut fixture = GenesisBuilder::new()
+        .with_snark_account(snark_id, |acct| acct.with_balance(sats(u64::MAX)))
```
bewakes: Let's not use more than max bitcoin supply for balance.

crates/ol/stf/src/tests/sau_limits.rs
```diff
@@ -0,0 +1,347 @@
...
+    assert_eq!(
+        fixture.account_balance(recipient1_id),
+        sats(0),
+        "Recipient1 should have no balance after failed update"
+    );
+    assert_eq!(
+        fixture.account_balance(recipient2_id),
+        sats(0),
+        "Recipient2 should have no balance after failed update"
+    );
```
bewakes, line 186: nit: Not sure if its worth it but this portion is repeated and perhaps can be extracted into a helper method inside `fixture`, something like `fixture.assert_unchanged(acct_id, acct_state)`.

crates/ol/stf/src/tests/sau_limits.rs
```diff
@@ -0,0 +1,347 @@
...
+}
+
+#[test]
+fn test_snark_update_allows_max_balance_transfer() {
+    let snark_id = get_test_snark_account_id();
+    let recipient_id = test_account_id(TEST_RECIPIENT_ID + 1);
+    let second_recipient_id = test_account_id(TEST_RECIPIENT_ID + 2);
+
+    let mut fixture = GenesisBuilder::new()
+        .with_snark_account(snark_id, |acct| acct.with_balance(sats(u64::MAX)))
```
bewakes: This should be limited by max bitcoin supply?

crates/ol/stf/src/test_utils.rs
```diff
@@ -481,6 +490,951 @@ pub fn get_test_proof(variant: u8) -> Vec<u8> {
...
+        }
+    }
+
+    /// Sets the GAM payload.
+    pub fn with_payload(mut self, payload: Vec<u8>) -> Self {
+        self.payload = payload;
+        self
+    }
+
+    fn into_tx(self) -> OLTransaction {
```
bewakes: nit: since this is a builder, perhaps name this `build()`?

crates/ol/stf/src/test_utils.rs
```diff
@@ -481,6 +490,951 @@ pub fn get_test_proof(variant: u8) -> Vec<u8> {
...
+        self
+    }
+
+    /// Overrides the predicate proof bytes.
+    pub fn with_proof(mut self, proof: Vec<u8>) -> Self {
+        self.proof = proof;
+        self
+    }
+
+    fn into_tx_with_seqno(self) -> (OLTransaction, Option<u64>) {
```
bewakes: nit: since this is a builder, maybe name this to `build*()`?

crates/ol/stf/src/tests/stress.rs
```diff
@@ -0,0 +1,124 @@
...
+
+/// Representative large workload size for stress tests.
+const STRESS_BATCH_SIZE: usize = 1_000;
+
+fn indexed_payload(index: usize) -> Vec<u8> {
+    index.to_le_bytes().to_vec()
+}
+
+#[test]
+#[ignore = "stress test"]
```
bewakes: Why do we want to keep ignored tests?

bewakes (commented):
Great work! The tests are much cleaner. Some comments, nothing major blocking.
You might want to also update [this doc](https://www.notion.so/OL-STF-test-cases-2e3901ba000f80598da0fceaeb62dd5e) as necessary.

crates/ol/stf/src/tests/stress.rs
krsnapaudel: @bewakes The idea is these may take longer to run. We could actually make them run every time if we want.

crates/ol/stf/src/tests/sau_limits.rs
krsnapaudel: Yes it's already done.

crates/ol/stf/src/tests/sau_limits.rs
krsnapaudel: @bewakes This depends on what the test is trying to do. One tests checks overflow.

crates/ol/stf/src/tests/asm_manifests.rs
krsnapaudel: Yes added an assertion.

crates/ol/stf/src/tests/stress.rs
bewakes: I think we should have a separate flow to run stress tests. @storopoli what's your opinion here since you thumbs-downed above?

bewakes (commented):
My comments seem to be resolved. Just in case I might be missing something, I asked claude to check for possible missing tests. Here's what it recommends:

 1. **u32 overflow untested**. Balance/transfer overflow (u64) is covered exhaustively; chain coordinates (u32) — L1 height, epoch, slot increments — aren't. In the guest, a debug panic is a liveness failure and a release wrap is a soundness failure. Please add boundary tests near `u32::MAX`.
  2. **Logs-root padding barely exercised**. `MAX_LOGS_PER_BLOCK` is a power of two, so the max-logs and empty-logs tests both skip the padding loop in compute_logs_root. Needs a non-power-of-two log count. The same logic is duplicated inassembly — cover both and assert they agree.
  3. **`da_preseal_correctness` asserts "no error", not correctness**. The round-trip tests only check that reconstruction doesn't error; they never compare the rebuilt blob/state against the original. A DA scheme that silently drops a field would pass, which defeats the test's purpose. Please add an equality assertion.

  Items 1 and 3 share a pattern — a green test standing in for verification never actually done. Worth scanning the suite for others.

delbonis (approved):
Lots of improvements here.  Approving, although will need another since there's merge conflicts now.

Other comment is worth considering, although maybe should be done in a followup PR instead of this one since this has been open a while and is in a mature state.

krsnapaudel:
> My comments seem to be resolved. Just in case I might be missing something, I asked claude to check for possible missing tests. Here's what it recommends:
> 
> 1. **u32 overflow untested**. Balance/transfer overflow (u64) is covered exhaustively; chain coordinates (u32) — L1 height, epoch, slot increments — aren't. In the guest, a debug panic is a liveness failure and a release wrap is a soundness failure. Please add boundary tests near `u32::MAX`.
> 2. **Logs-root padding barely exercised**. `MAX_LOGS_PER_BLOCK` is a power of two, so the max-logs and empty-logs tests both skip the padding loop in compute_logs_root. Needs a non-power-of-two log count. The same logic is duplicated inassembly — cover both and assert they agree.
> 3. **`da_preseal_correctness` asserts "no error", not correctness**. The round-trip tests only check that reconstruction doesn't error; they never compare the rebuilt blob/state against the original. A DA scheme that silently drops a field would pass, which defeats the test's purpose. Please add an equality assertion.
> 
> Items 1 and 3 share a pattern — a green test standing in for verification never actually done. Worth scanning the suite for others.

Thanks @bewakes. I added checked arithmetic for the epoch increment in `manifest_processing.rs` and reused the same `EpochOverflow` error in `process_block_start`. I also added `SlotOverflow` for the block-start slot increment and covered both block-start overflow paths with targeted tests. I also added an explicit `AsmManifestHeightOverflow` path for expected manifest-height calculation.

bewakes (approved):

prajwolrg (approved):
utACK

PR #1761, opened 2026-05-07
STR 3156 Apply new branch during FCM reorg handling instead of rollback-only behavior

krsnapaudel:
## Description

Fixes a fork-choice bug: `apply_tip_update` rolled back to the pivot on a reorg but never consumed `reorg.apply_iter()`, so FCM's chain tip and the published `OLSyncStatus` stayed stuck at the pivot. `TipUpdate::LongExtend` had the same shape of bug. Also hardens the apply path against storage inconsistencies and adds a thin context abstraction so it can be tested in isolation.

### Fix
- New helper `advance_fcm_to_block`: validates the block header's parent against the current FCM tip *before* loading state, loads the persisted `OLState`, and advances FCM via `update_tip_block` (which drives `update_safe_tip`).
- `TipUpdate::Reorg` rolls back to the pivot (when there's a down branch), iterates `reorg.apply_iter()` through the helper, then post-checks the tip landed on `reorg.new_tip()`. `TipUpdate::LongExtend` iterates its intermediates + new block through the same helper, reusing the same post-check.
- Pre/post-condition failures are typed errors — `InvalidOLReorgPivot`, `InvalidOLReorgEmptyDownPivot`, `OLApplyBlockParentMismatch`, `OLApplyTipMismatch` — which can fire on storage corruption, not just developer bugs. Adds `MissingOLBlock` / `MissingOLState`.
- `revert_ol_state_to_block` now propagates the `update_tip_block` result instead of discarding it.

### Supporting refactor (no behavior change)
FCM is now generic over an `FcmContext` super-trait (`ChainController` + `CsmStatusReader` + `FcmStorage`) instead of holding concrete `ChainWorkerHandle` / `ServiceMonitor<CsmWorkerStatus>` / `NodeStorage`; the production adapter (`StrataFcmContext`) lives in `bin/strata/src/fcm.rs`, so `consensus-logic`'s `fcm/` module no longer names any node-wiring type and drops the `strata-chain-worker-new` dependency. `unfinalized_tracker.rs` is split into `unfinalized_tracker/{mod,l2,ol}.rs`. This is what makes the fixes testable without `NodeContext` / `bitcoind` / `AsmParams` scaffolding.

### Tests
- Reorg `A1/A2 → B1/B2/B3` — asserts the safe-tip update order `[pivot, B1, B2, B3]` and final state.
- Single-block up reorg.
- Corrupt tracker (forged parent link) → `OLApplyBlockParentMismatch`; FCM left at the rolled-back pivot.
- Missing `OLState` mid-loop → typed `MissingOLState`; FCM left at the partial-apply position (no compensating rollback).
- `process_fc_message` end-to-end: a new block on a competing fork triggers a reorg and publishes `OLSyncStatus.tip` at the new tip.
- LongExtend `G → X1 → X2 → X3 → X4` — direct apply (asserts the update is a `TipUpdate::LongExtend`) and `process_fc_message` end-to-end.

Tests run against an in-memory `StubFcmContext` (role-trait impls + call recorders) and an `execute_block`-based block fixture. Legacy `fork_choice_manager.rs` has the same bug class but is unreachable from `bin/strata`; left untouched.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-3156](https://alpenlabs.atlassian.net/browse/STR-3156)

[STR-3156]: https://alpenlabs.atlassian.net/browse/STR-3156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


krsnapaudel:
@AlpenCodex review

crates/consensus-logic/src/fcm/context.rs
delbonis, line 1: Can we make these non-dyn?  Not a huge priority.

crates/consensus-logic/src/errors.rs
```diff
@@ -48,6 +54,24 @@ pub enum Error {
     #[error("unexpected genesis OL block {0:?}")]
     UnexpectedGenesisBlock(OLBlockId),
 
+    /// Reorg has a non-empty down branch, but its pivot is not before the current tip.
+    #[error(
+        "invalid OL reorg pivot: pivot not before tip with non-empty down (pivot {0}, cur_tip {1})"
```
delbonis: formatting

crates/consensus-logic/src/fcm/test_utils.rs
```diff
@@ -0,0 +1,377 @@
...
+    params.rollup.cred_rule = CredRule::Unchecked;
+    Arc::new(params)
+}
+
+fn create_test_storage() -> anyhow::Result<Arc<NodeStorage>> {
+    Ok(Arc::new(create_node_storage(
+        get_test_sled_backend(),
+        ThreadPool::new(1),
+    )?))
+}
```
delbonis: Can we abstract away the storage context so that we don't need to pull in the db deps anymore?  At least we can try to incrementally move in this direction.

delbonis (changes requested):

krsnapaudel:
@storopoli I can't ask for review from @codex. I tried @AlpenCodex above. Seems it didn't work.

krsnapaudel:
@delbonis @bewakes Ended up being a good amount of refactor. PTAL.

crates/consensus-logic/src/fcm/service.rs
```diff
@@ -57,34 +63,53 @@ pub async fn start_fcm_service(
...
 
 #[derive(Clone, Debug, Serialize)]
 pub struct FcmStatus;
 
-impl Service for FcmService {
+impl<C, R, S> Service for FcmService<C, R, S>
+where
+    C: ChainController + 'static,
+    R: CsmStatusReader + 'static,
+    S: FcmStorage + 'static,
```
delbonis: If you declare the trait as `trait Foo: 'static` then you don't need this bound.

Also, it'd be more idiomatic to describe the bounds as

```rs
impl<C: ChainController, R: CsmStatusReader, S: FcmStorage> Service for FcmService<C, R, S>
```

Should update this throughout, there's a lot here that uses the more verbose style.  In some places it can even just be changed to `&impl Foo` as the arg type.

crates/consensus-logic/src/fcm/service.rs
```diff
@@ -393,16 +452,19 @@ pub fn check_ol_block_proposal_valid(
...
-async fn pick_best_block_async(
+async fn pick_best_block_async<S>(
     cur_tip: &OLBlockId,
     tips: &[OLBlockId],
-    ol_block_mgr: &OLBlockManager,
-) -> Result<OLBlockId, Error> {
+    storage: &S,
+) -> Result<OLBlockId, Error>
+where
+    S: FcmStorage + ?Sized,
```
delbonis, line 417: Like this example.  Sine `FcmStorage` doesn't have associated types that get referenced anywhere, we can just say `storage: &impl FcmStorage`.  You might have to add that `?Sized` bound to the trait decl or tweak some other things, but this is much more verbose than it needs to be in any case.

crates/consensus-logic/src/fcm/context.rs
delbonis, line 1: The way this works now sorta follows the context-based design in theory but not really in spirit.  It's adding the indirection but it's not actually decoupling the components, since we're still using all the same deps instead of splitting out these impls into downstream crates.

The goal should be to make these services unaware of the surrounding context they're operating in so that we can test them more independently and simplify the dependency hierarchy.

I'm not really sure how to proceed here with this, maybe it doesn't matter.

crates/consensus-logic/src/fcm/context.rs
```diff
@@ -1,58 +1,180 @@
...
+    ) -> Self {
+        Self::new(
+            nodectx.params().clone(),
+            nodectx.storage().clone(),
+            chain_worker,
+            csm_monitor,
+            nodectx.status_channel().clone(),
+        )
     }
 }
```
delbonis: This is an example of something that should be split out and defined closer to where it's actually called.

delbonis (changes requested):

crates/consensus-logic/src/fcm/context.rs
krsnapaudel: @delbonis I tried to be conservative in certain ways since this PR is not about this refactor. The refactor was a consequence of the ugliness around testing. I will try one more time. Hopefully the refactor is not too big.

storopoli:
@krsnapaudel you should call just [at]codex not [at]AlpenCodex (even if it suggest as autocomplete), see example below.

storopoli:
@codex review

krsnapaudel:
@delbonis I have done the full refactor. One major question is should we keep `impl FcmContext` part on `bin/strata` side or `fcm`? I have put this on `bin/strata` side. So everything is abstracted out.

delbonis:
@krsnapaudel Yes keep it in the downstream crate, this helps us move towards flattening the crate hierarchy.  The goal is that services' context is separated from the impl so they can be more cleanly tested in isolation.

delbonis (approved):
Substantively looks good now, I think.

storopoli (approved):
ACK e2a393c602ddde13206bafab22fa5a175d072372

Issue #112, opened 2024-07-11
Create toml config parser for rollup node


bewakes:
Those are passed through command line arguments at the moment. Would be easier if we have a toml config parser and a sample config file.

PR #1786, opened 2026-05-07
 cover VK rotation sync and derive checkpoint envelope key from ASM

voidash:
## Summary
- add a functional test covering late alpen-client fullnode sync after EE predicate rotation
- derive checkpoint envelope signing mode from latest ASM checkpoint state instead of startup rollup params
- reset pending L1 writer envelopes when the ASM-derived signing mode changes before reveal publication

## Non-goal
- does not change ASM same-block admin-update semantics; a SequencerUpdate and checkpoint tx included in the same L1 block still take effect across the normal ASM tx/message processing boundary

### DIFF
--- a/bin/strata/src/checkpoint_auth.rs
+++ b/bin/strata/src/checkpoint_auth.rs
@@ -1,0 +1,113 @@
+//! Runtime checkpoint authentication state derived from ASM.
+
+use std::{fmt, sync::Arc};
+
+use bitcoin::secp256k1::{Error as Secp256k1Error, XOnlyPublicKey};
+use strata_asm_common::Subprotocol;
+use strata_asm_proto_checkpoint::CheckpointSubprotocol;
+use strata_btcio::writer::{EnvelopeSigningMode, EnvelopeSigningModeProvider};
+use strata_db_types::errors::DbError;
+use strata_predicate::PredicateTypeId;
+use strata_storage::NodeStorage;
+
+/// Errors produced while resolving checkpoint envelope authentication.
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum CheckpointAuthError {
+    /// The latest ASM state could not be loaded from storage.
+    #[error("failed to fetch latest ASM state")]
+    FetchLatestAsmState(#[source] DbError),
+
+    /// No ASM state exists yet.
+    #[error("latest ASM state is not available")]
+    MissingLatestAsmState,
+
+    /// The checkpoint subprotocol section is missing from ASM state.
+    #[error("latest ASM state is missing checkpoint subprotocol state")]
+    MissingCheckpointState,
+
+    /// The checkpoint subprotocol section could not be decoded.
+    #[error("failed to decode checkpoint subprotocol state")]
+    DecodeCheckpointState(#[source] strata_asm_common::AsmError),
+
+    /// The checkpoint sequencer predicate type is unknown.
+    #[error("unknown checkpoint sequencer predicate {0}")]
+    UnknownPredicate(u8),
+
+    /// The checkpoint sequencer predicate has an invalid condition length.
+    #[error("Bip340Schnorr checkpoint sequencer predicate has {0} condition bytes")]
+    InvalidSchnorrConditionLength(usize),
+
+    /// The checkpoint sequencer predicate condition is not a valid x-only key.
+    #[error("invalid checkpoint sequencer x-only pubkey")]
+    InvalidSchnorrPubkey(#[source] Secp256k1Error),
+
+    /// The active checkpoint sequencer predicate cannot sign envelopes.
+    #[error("checkpoint sequencer predicate is NeverAccept")]
+    NeverAccept,
+
+    /// The active checkpoint sequencer predicate is not supported for envelope authentication.
+    #[error("checkpoint sequencer predicate cannot use Sp1Groth16")]
+    Sp1Groth16,
+}
+
+/// Resolves the active checkpoint sequencer key from the latest ASM state.
+#[derive(Clone)]
+pub(crate) struct CheckpointSequencerKeyProvider {
+    storage: Arc<NodeStorage>,
+}
+
+impl fmt::Debug for CheckpointSequencerKeyProvider {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("CheckpointSequencerKeyProvider")
+            .finish_non_exhaustive()
+    }
+}
+
+impl CheckpointSequencerKeyProvider {
+    pub(crate) fn new(storage: Arc<NodeStorage>) -> Self {
+        Self { storage }
+    }
+
+    fn resolve_signing_mode(&self) -> Result<EnvelopeSigningMode, CheckpointAuthError> {
+        let (_, asm_state) = self
+            .storage
+            .asm()
+            .fetch_most_recent_state_blocking()
+            .map_err(CheckpointAuthError::FetchLatestAsmState)?
+            .ok_or(CheckpointAuthError::MissingLatestAsmState)?;
+
+        let checkpoint_section = asm_state
+            .state()
+            .find_section(<CheckpointSubprotocol as Subprotocol>::ID)
+            .ok_or(CheckpointAuthError::MissingCheckpointState)?;
+
+        let checkpoint_state = checkpoint_section
+            .try_to_state::<CheckpointSubprotocol>()
+            .map_err(CheckpointAuthError::DecodeCheckpointState)?;
+
+        let predicate = checkpoint_state.sequencer_predicate();
+        let predicate_id = predicate.id();
+        let predicate_type = PredicateTypeId::try_from(predicate_id)
+            .map_err(|_| CheckpointAuthError::UnknownPredicate(predicate_id))?;
+
+        match predicate_type {
+            PredicateTypeId::AlwaysAccept => Ok(EnvelopeSigningMode::InProcess),
+            PredicateTypeId::Bip340Schnorr => {
+                let pubkey_bytes: [u8; 32] = predicate.condition().try_into().map_err(|_| {
+                    CheckpointAuthError::InvalidSchnorrConditionLength(predicate.condition().len())
+                })?;
+                let pubkey = XOnlyPublicKey::from_slice(&pubkey_bytes)
+                    .map_err(CheckpointAuthError::InvalidSchnorrPubkey)?;
+                Ok(EnvelopeSigningMode::External { pubkey })
+            }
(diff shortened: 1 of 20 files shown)


crates/btcio/src/writer/builder.rs
```diff
@@ -182,7 +184,7 @@ pub(crate) async fn build_envelope_txs<R: Reader + Signer + Wallet>(
 /// Builds envelope transactions using a temporary keypair and signs both commit and reveal
 /// in-process.
 ///
-/// Used when `CredRule::Unchecked` is configured — no external signer is needed.
+/// Used when `CredRule::Unchecked` is configured; no external signer is needed.
 pub(crate) async fn build_and_sign_envelope_txs<R: Reader + Signer + Wallet>(
```
prajwolrg, line 194: Are we still referencing `CredRule`?

prajwolrg (changes requested):
This is a good start, but we'd need to cover a few cases:

`SequencerUpdate` is expected to use confirmation depth 0 in practice, but the mechanism supports arbitrary confirmation depths. We should test the boundary conditions explicitly, especially for depth 0, and also cover depth > 0 where queued updates are processed before transactions at the activation height.

**1. `SequencerUpdate` (confirmation_depth=0) and checkpoint transaction from the old sequencer are included in the same block.**
Since the `SequencerUpdate` must be sent from the admin subprotocol to the checkpoint subprotocol, enactment happens in the second phase of the ASM. Therefore, the checkpoint transaction from the old sequencer should still be accepted when both transactions are in the same block. This should hold regardless of the ordering of the `SequencerUpdate` and `CheckpointTx` within the block.

2. `SequencerUpdate` is activated in a block before the checkpoint transaction settles.
If the SequencerUpdate is included in an earlier block than the checkpoint transaction settlement, the checkpoint should be invalid because the old sequencer is no longer authorized by the time the checkpoint settles. The checkpoint transaction may still have been published and considered valid from the old sequencer’s perspective. In that case, the new sequencer may need to rebroadcast the same checkpoint payload, but it must be re-signed using the SPS-51 envelope by the new sequencer.

We may need to delete the old valid checkpoint transaction that was published by the old sequencer, since `BundledPayloadEntry` is stored in the DB. In practice, however, a sequencer rotation would likely use a different device, so deleting the existing DB should not be necessary. The new sequencer may rebroadcast the same checkpoint payload, but it must be re-signed using the SPS-51 envelope by the new sequencer.

3. Depth > 0: queued `SequencerUpdate` activates at a later height.
For non-zero confirmation depth, the update is queued and activated at the target height. Queued updates are processed before any transactions at the activation height. For example, if the SequencerUpdate activates at height 100, it is processed before any transactions in block 100. Therefore, any checkpoint transaction that settles at height >= 100 and is signed by the old sequencer must fail.

purusang (commented):
If its not non-trivial please add test for checkpoint predicate rotation.

purusang (dismissed):
I do not actually like the home of such higher capabilities like predicate rotation code being something called `*-test-*`.

bewakes (approved):
LGTM - looked at func tests and just skimmed through other.

One good to have: the functional test steps are quite long, consider adding concise comments before each step or in the docstring.

voidash:
~~blocked rn on [ASM PR #112](https://github.com/alpenlabs/asm/pull/112)~~

Update: dropping the ASM PR after clarifying the semantics.

This PR no longer depends on changing ASM tx/message ordering. Same-block SequencerUpdate + checkpoint behavior remains unchanged: the update is visible in the next processing round, so a checkpoint from the old sequencer can still be accepted in that same L1 block.

purusang:
@storopoli Do you know why we put admin functionalities in test-cli? Should it be renamed?

storopoli:
> @storopoli Do you know why we put admin functionalities in test-cli? Should it be renamed?

Wait! This is for testing right @voidash?
If no, I agree with @purusang, we should have a separate `admin-cli` (or other naming) for that.

Cc @prajwolrg

voidash:
> Wait! This is for testing right @voidash?

Yes, this is for functional test only. This is not intended as operator admin tooling. @purusang .

bin/strata/src/checkpoint_auth.rs
delbonis, line 1: good code organization!  thank you!

bin/strata/src/checkpoint_auth.rs
```diff
@@ -0,0 +1,85 @@
...
+                let pubkey = XOnlyPublicKey::from_slice(&pubkey_bytes)
+                    .context("invalid checkpoint sequencer x-only pubkey")?;
+                Ok(EnvelopeSigningMode::External { pubkey })
+            }
+            PredicateTypeId::NeverAccept => {
+                Err(anyhow!("checkpoint sequencer predicate is NeverAccept"))
+            }
+            PredicateTypeId::Sp1Groth16 => Err(anyhow!(
+                "checkpoint sequencer predicate cannot use Sp1Groth16"
+            )),
```
delbonis: Should make this be its own error type though instead of using `anyhow::anyhow!`.

bin/strata/src/sequencer/rpc.rs
```diff
@@ -62,15 +61,15 @@ impl OLSeqRpcServer {
         blockasm_handle: Arc<BlockasmHandle>,
         envelope_handle: Arc<EnvelopeHandle>,
         fcm_handle: Arc<FcmServiceHandle>,
-        sequencer_pubkey: Option<Buf32>,
+        sequencer_key_provider: CheckpointSequencerKeyProvider,
```
delbonis: Should probably make this use the trait instead of using the concrete impl.

delbonis (changes requested):

storopoli (approved):
ACK 9f45309ff8d77b2400e3f2958292bb7289d891ef

irnb (approved):

bin/strata/src/checkpoint_auth.rs
voidash: Thanks. I kept the split and addressed the two follow-up review comments in `d4e3bc058`.

bin/strata/src/checkpoint_auth.rs
voidash: Addressed in `d4e3bc058`. `CheckpointSequencerKeyProvider` now resolves through a typed `CheckpointAuthError` with concrete variants and source errors, and only converts to `anyhow::Result` at the `EnvelopeSigningModeProvider` trait boundary.

bin/strata/src/sequencer/rpc.rs
voidash: Addressed in `d4e3bc058`. `OLSeqRpcServer` now stores `Arc<dyn EnvelopeSigningModeProvider>` and derives the optional current pubkey from `signing_mode()`, so the RPC path no longer depends on the concrete checkpoint key provider.

prajwolrg (approved):

functional-tests/tests/alpen_client/test_ee_predicate_fullnode_sync.py
```diff
@@ -0,0 +1,198 @@
...
+        admin_key_path = Path(strata_seq.props["datadir"]) / "bridge-operator_keys"
+        if not admin_key_path.exists():
+            raise AssertionError(f"admin key file not found: {admin_key_path}")
+        admin_xpriv = admin_key_path.read_text().strip()
+        if not admin_xpriv:
+            raise AssertionError(f"admin key file is empty: {admin_key_path}")
+        return admin_xpriv
+
+    @staticmethod
+    def _apply_predicate_update(
```
delbonis, line 167: I think these `@staticmethod`s can be split out as module-level fns, they're not providing anything by being on the class.

functional-tests/tests/checkpoint/test_checkpoint_predicate_rotation.py
delbonis, line 1: also here

delbonis (dismissed):
minor comments now, otherwise I think this looks good

purusang (approved):

storopoli (approved):
ACK 78a17a67d451c353a69d4632ef6a77a4c7871ae6

storopoli:
@krsnapaudel can you take a look on this? You're the final CODEOWNER so this can be added to the queue.

crates/btcio/src/writer/watcher/service.rs
```diff
@@ -280,20 +282,20 @@ impl<C: WatcherServiceContext> AsyncService for WatcherService<C> {
...
     /// and transitions directly to `Unpublished`.
     async fn handle_unsigned_or_needs_resign(
         &mut self,
         payloadentry: BundledPayloadEntry,
     ) -> anyhow::Result<()> {
         debug!(current_status=?payloadentry.status);
 
-        if self.ctx.needs_external_signing() {
-            match self
+        match self.ctx.signing_mode()? {
```
krsnapaudel: `signing_mode()` is `?`-propagated here, and the service worker treats a `process_input` error as fatal (`async_worker.rs` logs and then returns, with no retry). So a `NeverAccept`/`Sp1Groth16` predicate — or a transient `DbError` from `fetch_most_recent_state_blocking` — permanently kills the writer with no recovery, even though the ASM-derived mode is dynamic and could rotate back to a signable key.

Suggest warning + `Response::Continue` (matching the existing `is_retryable_envelope_error` handling just above) so the tick loop re-evaluates instead of terminating. Was hard-terminating intended here?

With help from Claude.

krsnapaudel (dismissed):
I think the error propagation is worth a second look.

storopoli (approved):
ACK ef87f46e1ab2faf3e027e87acb49d306a46d2d0d

crates/btcio/src/writer/watcher/service.rs
prajwolrg: Resolved in [ef87f46](https://github.com/alpenlabs/alpen/pull/1786/commits/ef87f46e1ab2faf3e027e87acb49d306a46d2d0d).

bin/strata/src/sequencer/rpc.rs
```diff
@@ -62,19 +60,28 @@ impl OLSeqRpcServer {
...
 }
 
+fn current_pubkey(
+    signing_mode_provider: &dyn EnvelopeSigningModeProvider,
+) -> anyhow::Result<Option<Buf32>> {
+    match signing_mode_provider.signing_mode()? {
+        EnvelopeSigningMode::InProcess => Ok(None),
+        EnvelopeSigningMode::External { pubkey } => Ok(Some(Buf32(pubkey.serialize()))),
+    }
+}
```
delbonis, line 83: nit: for clarity should be `get_current_sequencer_pubkey`

delbonis (approved):
I think this looks pretty good now, just one low-priority comment which could be skipped.

krsnapaudel (approved):

PR #1751, opened 2026-05-04
Parallelize EE DA chunk submission

irnb:
## Description

Implementation of https://www.notion.so/EE-DA-Chunked-Envelope-Redesign-344901ba000f80009316c16b340a8bb8?d=354901ba000f80a0b5b7001c8a143321#344901ba000f80d89b16e136903f372f

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

 the PR implements the EE DA tx skeleton proposed in the [design doc](https://www.notion.so/EE-DA-Chunked-Envelope-344901ba000f80009316c16b340a8bb8?source=copy_link) and the DaBlob proposed [here](https://alpen-labs.slack.com/archives/C08G7PL1RBK/p1777991214813059?thread_ts=1777991012.505009&cid=C08G7PL1RBK).

I'll create two SPS docs, one for the tx skeleton and another for EE L1 sync + DaBlob -- instead of updating the existing notion doc

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [ ] I have updated the documentation if needed.
- [ ] My changes do not introduce new warnings.
- [ ] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [ ] New and existing tests pass with my changes.
- [x] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

PR #1788, opened 2026-05-07
STR-1369: EE DA correctness in acct proofs

evgenyzdanovich:
## Description

- (done) EE DA **inclusion** statements
- (done) EE DA **correctness** statements 
- (WIP) range witness extractor is currently used in-place (should be persisted)

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist


crates/btcio/src/writer/chunked_envelope/reader.rs
```diff
@@ -0,0 +1,163 @@
...
+
+/// Errors lifting an envelope payload out of a reveal tx.
+#[derive(Debug, thiserror::Error)]
+pub enum ExtractRevealError {
+    #[error("reveal tx has no inputs")]
+    MissingInputs,
+    #[error("reveal tx witness has no tapscript leaf")]
+    MissingLeafScript,
+    #[error("envelope parse error: {0}")]
+    EnvelopeParse(#[from] EnvelopeParseError),
```
delbonis, line 32: `error` is redundant in error strings

crates/ee-acct-runtime/src/da_witness.rs
delbonis, line 1: Ahh this is wonderfully clean code.

I understand why we might not want this in the `private_inputs` module (too big) but ehh, maybe we could create a dir for the different parts?

crates/ee-acct-runtime/src/da_witness.rs
```diff
@@ -0,0 +1,297 @@
...
+
+#[archive_impl]
+impl DaBlockWitness {
+    pub fn raw_header(&self) -> &[u8] {
+        &self.raw_header
+    }
+
+    pub fn raw_coinbase_tx(&self) -> &[u8] {
+        &self.raw_coinbase_tx
+    }
```
delbonis, line 119: Maybe add `try_decode_header` and `try_decode_coinbase_tx` convenience fns like in the other `private_inputs` modules?

crates/proof-impl/alpen-acct/src/da_inclusion.rs
delbonis, line 1: Any way we could move parts of this logic into the EE account runtime so that it's reusable for whatever else that needs to check DA?

crates/proof-impl/alpen-acct/src/da_verify.rs
delbonis, line 1: Same here.

crates/ee-acct-runtime/src/da_witness.rs
```diff
@@ -0,0 +1,297 @@
...
+    wtxid: [u8; 32],
+
+    /// Merkle proof: reveal wtxid → witness root (committed in
+    /// coinbase OP_RETURN).
+    wtxid_to_witness_root_proof: BitcoinMerkleProof,
+
+    /// Envelope payload bytes lifted from the reveal's witness:
+    /// `chunk_header (37 bytes) || chunk_payload`. Same shape that
+    /// `alpen_ee_common::decode_da_chunk` consumes.
+    envelope_payload: Vec<u8>,
```
delbonis, line 167: I would expect this to be the encoded tx (where we'd decode the payload out of it), but it's the payload.  How does this work without needing the full tx?  Am I misunderstanding something?

crates/proof-impl/alpen-acct/src/da_verify.rs
```diff
@@ -0,0 +1,771 @@
...
+/// `raw_partial_pre_state` is empty so test fixtures and intermediate
+/// host states keep working until [`super::EeAcctProofInput`] is fully
+/// populated.
+pub fn verify_da_witness(
+    ee_input: &ArchivedEePrivateInput,
+    da_witness: &ArchivedDaWitness,
+) -> Result<Option<DaBlob>, DaVerificationError> {
+    if da_witness.blocks().is_empty() {
+        return Ok(None);
+    }
```
delbonis, line 126: Is this safe?  Does this enable skipping DA checks by putting an empty list of blocks?  Surely that's not a correct understanding of how this works.  Please correct my understanding.

crates/proof-impl/alpen-acct/src/da_verify.rs
```diff
@@ -0,0 +1,771 @@
...
+pub struct BatchIdMismatch {
+    pub blob_prev: Hash,
+    pub blob_last: Hash,
+    pub expected_prev: Hash,
+    pub expected_last: Hash,
+}
+
+/// Errors raised by [`verify_da_witness`].
+#[derive(Debug, thiserror::Error)]
+pub enum DaVerificationError {
```
delbonis, line 64: Put data in parens, like `blah blah blah (witness tip {witness_tip:?}, ledger tip {ledger_tip:?})`.

crates/proof-impl/alpen-acct/src/lib.rs
```diff
@@ -48,6 +73,20 @@ pub fn process_ee_acct_update(zkvm: &impl ZkVmEnvSerde, chunk_predicate_key: &Pr
...
+    let _da_blob = verify_da_witness(ee_input, da_witness).expect("DA witness verification failed");
+
+    // Bind the in-proof L1 anchor (`da_witness.l1_block_hash`) to the
+    // OL-facing anchor (highest-idx `LedgerRefs` entry) so a host can't
+    // pass mismatched tips between the inclusion checks and OL
+    // canonicality.
+    let pub_params = UpdateProofPubParams::from_ssz_bytes(upd_input.update_pub_params_ssz())
+        .expect("UpdateProofPubParams must be valid SSZ");
+    bind_da_witness_to_ledger_refs(da_witness, pub_params.ledger_refs())
+        .expect("DA witness tip and LedgerRefs tip mismatch");
```
delbonis, line 88: Maybe this *really isn't* necessary, but (as I gestured towards in the other comment) I would like to see some of this somehow integrated into the EE acct runtime.  We could add support for a check for "this state diff matches the state transition" and then *outside* the runtime exec we can add these checks against L1.  Then EEs that use alt DA could figure out on their own how to check it, but could use the same EE infra to check that it matches.

delbonis (changes requested):
This is fantastic work.  I have some design questions and other minor concerns, but in general this looks great.

crates/btcio/src/writer/chunked_envelope/reader.rs
```diff
@@ -0,0 +1,163 @@
+//! Inverse of the chunk-encoding side: given a reveal tx produced by
+//! [`super::builder::build_chunked_envelope_txs`], lift the
+//! `chunk_header || chunk_payload` bytes that
+//! `alpen_ee_common::decode_da_chunk` consumes.
+//!
+//! Used by the EE outer (acct) proof's host-side input assembly to plumb
+//! DA witness data into the prover.
+//
+// FIXME(#1751): the chunked-envelope wire format is being redesigned in
```
irnb, line 9: #1751 is merged

crates/ee-acct-runtime/src/da_witness.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    pub fn coinbase_to_header_proof(&self) -> &ArchivedBitcoinMerkleProof {
+        &self.coinbase_to_header_proof
+    }
+}
+
+/// Witness data for a single DA reveal tx.
+#[derive(Clone, Debug, Archive, Deserialize, Serialize)]
+#[rkyv(derive(Debug))]
+pub struct RevealWitness {
```
irnb, line 153: I think this witness shape is missing the critical binding we need. `wtxid` and `envelope_payload` are independent private fields, so the guest proves inclusion of one tx but reassembles DA from bytes supplied separately by the host. The guest should receive enough raw commit/reveal tx data to recompute txid/wtxid and extract the chunk payload from the included reveal tx itself.

crates/proof-impl/alpen-acct/src/da_verify.rs
```diff
@@ -0,0 +1,771 @@
...
+/// without requiring the host to plumb witnesses unconditionally. The
+/// state-diff consistency check is also skipped when
+/// `raw_partial_pre_state` is empty so test fixtures and intermediate
+/// host states keep working until [`super::EeAcctProofInput`] is fully
+/// populated.
+pub fn verify_da_witness(
+    ee_input: &ArchivedEePrivateInput,
+    da_witness: &ArchivedDaWitness,
+) -> Result<Option<DaBlob>, DaVerificationError> {
+    if da_witness.blocks().is_empty() {
```
irnb, line 124: This makes DA verification optional for any prover that supplies an empty `DaWitness`. For production acct proofs, an empty witness should only be valid for a true no-DA/genesis path. If chunks are present, the proof should require DA blocks and fail closed instead of returning `Ok(None)`.

crates/proof-impl/alpen-acct/src/da_verify.rs
```diff
@@ -0,0 +1,771 @@
...
+    }
+
+    let blob = reassemble_and_bind_batch_id(ee_input, da_witness)?;
+
+    let l1_block_hash = *da_witness.l1_block_hash();
+    for block in da_witness.blocks() {
+        verify_block_inclusion(block, l1_block_hash)?;
+    }
+
+    let raw_pre_state = ee_input.raw_partial_pre_state();
```
irnb, line 135: This also makes the state-diff binding optional. If DA exists, `raw_partial_pre_state` needs to be present; otherwise the proof can skip the check that `DaBlob.state_diff` applied to pre-state reaches the final chunk `tip_state_root`.

crates/btcio/src/writer/chunked_envelope/reader.rs
```diff
@@ -0,0 +1,163 @@
...
+    EnvelopeParse(#[from] EnvelopeParseError),
+}
+
+/// Extracts the chunked-envelope payload bytes (`chunk_header ||
+/// chunk_payload`) from a reveal tx's witness.
+///
+/// The reveal tx's first input must spend a tapscript path whose leaf
+/// is a chunked-envelope reveal script of the shape
+/// `<pubkey> OP_CHECKSIG OP_FALSE OP_IF <chunk_bytes> OP_ENDIF`.
+pub fn extract_chunk_envelope_payload(reveal: &Transaction) -> Result<Vec<u8>, ExtractRevealError> {
```
irnb, line 41: With the merged chunked-envelope format, the proof needs to start from the commit tx: verify OP_RETURN magic/version, identify reveal outputs by commit output order, then extract chunks from the matching reveal spends. Parsing a reveal in isolation is not enough.

crates/proof-impl/alpen-acct/src/da_verify.rs
```diff
@@ -0,0 +1,771 @@
...
+/// populated.
+pub fn verify_da_witness(
+    ee_input: &ArchivedEePrivateInput,
+    da_witness: &ArchivedDaWitness,
+) -> Result<Option<DaBlob>, DaVerificationError> {
+    if da_witness.blocks().is_empty() {
+        return Ok(None);
+    }
+
+    let blob = reassemble_and_bind_batch_id(ee_input, da_witness)?;
```
irnb, line 128: This `batch_id` binding is stale after #1751. 

After reassembling the blob, we should also verify the metadata fields. `update_seq_no` should match the update operation sequence number, and `evm_header` should match the final EVM block header summary for the batch. Right now the proof only uses `state_diff`.

irnb: we should also bind the blob to the chunk range through the state-diff/root check.

bin/alpen-client/src/prover/spec_acct.rs
```diff
@@ -307,12 +392,285 @@ impl ProofSpec for AcctSpec {
...
+        let raw_header_chain_to_tip: Vec<Vec<u8>> = ((height + 1)..=tip_height)
+            .map(|h| {
+                header_at_height
+                    .get(&h)
+                    .cloned()
+                    .expect("pre-fetched above")
+            })
+            .collect();
+
+        let mut reveals = Vec::with_capacity(r.txns.len());
```
irnb, line 581: `L1DaBlockRef.txns` now includes the commit tx first when present, followed by reveals sorted by vout. This loop treats every tx as a reveal and calls `lift_reveal_envelope`, so after rebasing onto #1751 it will try to parse commit txs as reveal txs.

crates/proof-impl/alpen-acct/src/da_verify.rs
```diff
@@ -0,0 +1,771 @@
...
+
+/// Cross-checks that `DaWitness.l1_block_hash` (the tip the in-proof
+/// inclusion checks anchored to) matches the highest-idx entry in
+/// [`LedgerRefs`] (the tip the OL canonicality check will anchor to).
+///
+/// Without this binding, a host could pass a `DaWitness` anchored to
+/// one tip and a [`LedgerRefs`] anchored to another, in which case
+/// inclusion verifies against tip A while OL canonicality verifies
+/// tip B — defeating the intended chain-of-custody.
+pub fn bind_da_witness_to_ledger_refs(
```
irnb, line 183: The LedgerRefs binding is useful, but it only aligns the proof’s L1 tip with the OL canonicality tip. It does not prove that the DA bytes came from the included Bitcoin txs. We still need the commit/reveal transaction binding described above.

irnb (changes requested):
Great effort, thanks. There are good building blocks here already, especially the Bitcoin header/wtxid inclusion checks and the state-root equality against the chunk transition.

The high-level goal I think we want is:

1. prove inclusion of the EE DA commit/reveal txs;
2. verify the commit OP_RETURN has the correct EE DA magic/version;
3. identify reveal spends by commit output order;
4. extract chunk bytes from the actual included reveal txs inside the proof;
5. concatenate/decode the `DaBlob`;
6. verify `update_seq_no` and `evm_header`;
7. apply `DaBlob.state_diff` to the proof-bound pre-state;
8. compare the result to the final chunk transition `tip_state_root`.

Right now the witness still provides DA bytes separately from the included txs, and parts of the flow are still pre#1751. I think aligning the proof around the commit/reveal binding above would close the mental-model gap.

irnb:
Another possible shape is to receive the `DaBlob` as witness data, encode and chunk it inside the proof using the protocol chunk size (`MAX_CHUNK_PAYLOAD = 395_000`), then check that those expected chunks match the payloads extracted from the included reveal txs. I think either direction is fine as long as the proof derives or verifies the DA bytes against the actual commit/reveal transactions, rather than trusting host-provided payload bytes independently.

PR #1793, opened 2026-05-08
fix!(strata): split sequencer admin RPC behind bearer auth

storopoli:
## Description

Splits signer-facing `strataadmin_*` RPC methods off the public Strata RPC listener and serves them from a separate admin RPC listener protected by bearer-token authentication.

This PR:
- adds `client.admin_rpc_host`, `client.admin_rpc_port`, and `client.admin_rpc_bearer_token`
- starts public and admin RPC listeners separately for sequencer nodes
- keeps `OLSequencerRpc` methods off the public listener
- authenticates admin HTTP/WebSocket handshakes with `Authorization: Bearer <token>`
- updates `strata-signer` to connect to the admin endpoint with the bearer token
- updates functional-test harness/configs to use the admin RPC endpoint
- hardens validation so public/admin listeners cannot share a port, including `0.0.0.0` vs `127.0.0.1`

### Type of Change

- [x] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The legacy `StrataAdminApi` was already removed from this branch because it was unused. This PR only moves the current signer-facing `strataadmin_*` methods from the public listener to the authenticated sequencer admin listener.

`STRATA_ADMIN_RPC_TOKEN` is applied directly to the parsed config instead of going through the generic override-string path, so the token is not stored in `override_strs`.

Non-sequence fullnodes do not start the admin RPC listener and are not required to configure an admin bearer token.

AI assistance notice: I used AI assistance to implement and validate this change.

- [ ]  Yes
- [ ]  No

## Checklist

## Related Issues

STR-3164


docker/configs/config.seq.toml
```diff
@@ -3,6 +3,9 @@
 [client]
 rpc_host = "0.0.0.0"
 rpc_port = 8432
+admin_rpc_host = "127.0.0.1"
```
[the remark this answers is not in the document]
storopoli: @codex check again this is fixed now

bin/strata/src/rpc/auth.rs
```diff
@@ -0,0 +1,124 @@
...
+
+impl<S, B> Service<Request<B>> for AdminAuthService<S>
+where
+    S: Service<Request<B>, Response = HttpResponse> + Send + Clone + 'static,
+    S::Future: Send + 'static,
+    B: Send + 'static,
+{
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
```
delbonis, line 59: I've messed with tower before, I'm pretty sure we can avoid having to make this dyn, but it's not a big deal so we don't *have* to change it.

delbonis (approved):
Looks good.

bin/strata/src/rpc/auth.rs
storopoli: The auth middleware can avoid boxing by returning a small enum future, but the solution is not elegant anyways...

storopoli:
@alexhui01 please take a look. We already have one CODEOWNER approval so this can be added to the queue, feel free to do so. Also check that the TOML configs in the OL full nodes have changed. This PR also fixes the docker example configs. But be aware with the "change me" values in https://github.com/alpenlabs/alpen/blob/f5f1b6579b6c60a1f22fd8e52047ce675712bc7d/docker/configs/strata-signer/config.toml#L2-L3

bin/strata/src/rpc/auth.rs
```diff
@@ -0,0 +1,124 @@
...
+    let Some(presented) = presented else {
+        return false;
+    };
+
+    constant_time_eq(presented.as_bytes(), expected.as_bytes())
+}
+
+fn unauthorized_response() -> HttpResponse {
+    HttpResponse::builder()
+        .status(StatusCode::UNAUTHORIZED)
```
krsnapaudel, line 87: Does this need

`.header(WWW_AUTHENTICATE, HeaderValue::from_static("Bearer"))`

bin/strata/src/rpc/auth.rs
```diff
@@ -0,0 +1,124 @@
...
+        .header(CONTENT_TYPE, HeaderValue::from_static("text/plain"))
+        .body(HttpBody::from("Unauthorized\n"))
```
krsnapaudel, line 90: Does it make sense to do this?
`.body(HttpBody::from(
            r#"{"jsonrpc":"2.0","error":{"code":-32001,"message":"Unauthorized"},"id":null}"#,
        ))`

docker/configs/config.fn.toml
```diff
@@ -2,6 +2,9 @@
 [client]
 rpc_host = "0.0.0.0"
 rpc_port = 8432
+admin_rpc_host = "127.0.0.1"
+admin_rpc_port = 8434
+admin_rpc_bearer_token = "dev-only-change-me"
```
krsnapaudel: Do full nodes need this?

bin/strata-signer/src/config.rs
```diff
@@ -13,7 +13,10 @@ pub(crate) struct SignerConfig {
     pub(crate) sequencer_key: PathBuf,
 
     /// WebSocket RPC URL of the sequencer node (e.g. ws://127.0.0.1:9944).
-    pub(crate) sequencer_endpoint: String,
+    pub(crate) sequencer_admin_endpoint: String,
+
+    /// Bearer token used to authenticate with the sequencer admin RPC.
+    pub(crate) sequencer_admin_bearer_token: String,
```
krsnapaudel: Claude suggestion:

`Config` derives `Debug` (via `ClientConfig`'s `Debug`), so anywhere the full config is formatted with `{:?}` — for example in an error context, a tracing event with `?config`, or a panic — the raw token bytes get printed.

```
#[derive(Clone, Deserialize, Serialize)]
#[serde(transparent)]
pub struct Secret(String);

impl Secret {
    pub fn expose(&self) -> &str { &self.0 }
}

impl fmt::Debug for Secret {
        f.write_str("Secret(***)")
    }
}
```
Then change the field to `pub admin_rpc_bearer_token: Option<Secret> `
Is this worth doing?

krsnapaudel (commented):
Mostly nits.

bin/strata-signer/src/config.rs
storopoli: Done in 9f54a75d6b5dca74722c51e1bb54fc6fea89dc43

docker/configs/config.fn.toml
storopoli: Nice catch, removed in 9f54a75d6b5dca74722c51e1bb54fc6fea89dc43

bin/strata/src/rpc/auth.rs
storopoli: I think we should keep this as an HTTP-level `401 Unauthorized` response rather than returning a JSON-RPC error object.

This middleware runs before the request reaches JSON-RPC dispatch, and it also protects the WebSocket handshake path, so the failure is at the transport/auth boundary rather than inside a JSON-RPC method call. I added `WWW-Authenticate: Bearer` to make the HTTP response clearer for clients in 9f54a75d6b5dca74722c51e1bb54fc6fea89dc43

bin/strata/src/rpc/auth.rs
storopoli: Yes, added in 9f54a75d6b5dca74722c51e1bb54fc6fea89dc43 Thanks

PR #1797, opened 2026-05-11
fix(btcio): keep insufficient funds errors retryable

irnb:
Preserve EnvelopeError::NotEnoughUtxos through the normal BTCIO envelope builder and signer paths instead of converting it into a generic anyhow error.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

## Related Issues

https://alpenlabs.atlassian.net/browse/STR-3406

### DIFF
--- a/crates/btcio/src/writer/builder.rs
+++ b/crates/btcio/src/writer/builder.rs
@@ -154,11 +154,15 @@ impl EnvelopeData {
 pub(crate) async fn build_envelope_txs<R: Reader + Signer + Wallet>(
     payload: &L1Payload,
     ctx: &WriterContext<R>,
-) -> anyhow::Result<EnvelopeData> {
-    let (network, utxos, fee_rate) = fetch_envelope_prereqs(ctx).await?;
-    let envelope_pubkey = ctx
-        .envelope_pubkey
-        .ok_or_else(|| anyhow::anyhow!("envelope_pubkey is required for envelope transactions"))?;
+) -> Result<EnvelopeData, EnvelopeError> {
+    let (network, utxos, fee_rate) = fetch_envelope_prereqs(ctx)
+        .await
+        .map_err(EnvelopeError::Other)?;
+    let envelope_pubkey = ctx.envelope_pubkey.ok_or_else(|| {
+        EnvelopeError::Other(anyhow!(
+            "envelope_pubkey is required for envelope transactions"
+        ))
+    })?;
     let env_config = EnvelopeConfig::new(
         ctx.btcio_params.magic_bytes(),
         ctx.sequencer_address.clone(),
@@ -168,7 +172,6 @@ pub(crate) async fn build_envelope_txs<R: Reader + Signer + Wallet>(
         Some(envelope_pubkey),
     );
     create_envelope_transactions(&env_config, payload, utxos)
-        .map_err(|e| anyhow::anyhow!(e.to_string()))
 }
 
 /// Builds envelope transactions using a temporary keypair and signs both commit and reveal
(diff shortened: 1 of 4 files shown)


crates/btcio/src/writer/watcher/service.rs
```diff
@@ -591,6 +608,9 @@ mod tests {
             _idx: u64,
             _entry: &BundledPayloadEntry,
         ) -> Result<(Buf32, Buf32), EnvelopeError> {
+            if self.fail_sign_not_enough_utxos {
+                return Err(EnvelopeError::NotEnoughUtxos(4096, 2658));
+            }
```
prajwolrg, line 622: Why are we hardcoding `4096` and `2658`?

crates/btcio/src/writer/watcher/service.rs
```diff
@@ -635,6 +655,23 @@ mod tests {
...
+        let entry = test_unsigned_entry();
+        ctx.stored.lock().unwrap().insert(0, entry.clone());
+
+        let mut state = WatcherState::new(ctx, 0);
+        state.handle_unsigned_or_needs_resign(entry).await.unwrap();
+
+        let stored = state.ctx.get_stored(0).unwrap();
+        assert_eq!(stored.status, L1BundleStatus::Unsigned);
+        assert_eq!(stored.commit_txid, Buf32::zero());
+        assert_eq!(stored.reveal_txid, Buf32::zero());
```
prajwolrg, line 680: This feels weird, but I guess is fine. Is it zero because we don't have any inputs and the transaction itself is invalid?

crates/btcio/src/writer/builder.rs
```diff
@@ -178,8 +181,10 @@ pub(crate) async fn build_envelope_txs<R: Reader + Signer + Wallet>(
 pub(crate) async fn build_and_sign_envelope_txs<R: Reader + Signer + Wallet>(
     payload: &L1Payload,
     ctx: &WriterContext<R>,
-) -> anyhow::Result<EnvelopeData> {
-    let (network, utxos, fee_rate) = fetch_envelope_prereqs(ctx).await?;
+) -> Result<EnvelopeData, EnvelopeError> {
+    let (network, utxos, fee_rate) = fetch_envelope_prereqs(ctx)
+        .await
+        .map_err(EnvelopeError::Other)?;
```
prajwolrg: let's use proper error if possible.

prajwolrg (changes requested):
Mostly looks good. Some nits and comments.

crates/btcio/src/writer/watcher/service.rs
irnb: These are only mock test fixture values for `EnvelopeError::NotEnoughUtxos`, not protocol/config values.

crates/btcio/src/writer/watcher/service.rs
irnb: In this test the entry remains `Unsigned`, so no commit/reveal tx has been built or persisted yet. I added a comment to make that explicit.

crates/btcio/src/writer/builder.rs
irnb: Agreed. I added explicit `MissingEnvelopePubkey` and `PrereqFetch` variants instead of using `Other` for those cases. `PrereqFetch` is still fatal for now; I added a TODO for the separate follow-up to make the OL node resilient when the Bitcoin node is unavailable.

prajwolrg (approved):

prajwolrg (approved):
LGTM. I'd also like @bewakes to take a look at this before merging this in.

bewakes:
@irnb, there doesn't seem to be existing code that errors out when there is not enough utxo. I can see that the not enough utxo paths all do `error!()` logs and not directly return. Does this solve the issue? Or am I missing something?

irnb:
> Does this solve the issue? Or am I missing something?

You’re right that the watcher already had a `NotEnoughUtxos` retry branch, but the issue was that the normal writer path never reached it with the typed error.

The actual failure path was:

`create_envelope_transactions(...)`
-> returns `EnvelopeError::NotEnoughUtxos`
-> `build_envelope_txs` / `build_and_sign_envelope_txs`
-> converted it to `anyhow!(e.to_string())`
-> signer `?` converted that generic anyhow into `EnvelopeError::Other`
-> watcher did not match `NotEnoughUtxos`
-> fell through to `e?`
-> critical `btcio_watcher` task exited.


This change preserves `EnvelopeError::NotEnoughUtxos` through builder -> signer -> watcher. After that, the watcher hits the existing retry branch, logs a warning, leaves the entry `Unsigned`, does not advance `curr_payloadidx`, and retries on the next tick.


@bewakes

bewakes (approved):

PR #1800, opened 2026-05-11
test(functional): add real bridge e2e

voidash:
## Summary
- add a functional e2e that runs an external strata-bridge checkout against core bitcoind/Strata/Alpen, deposits with alpen-cli into a no-prefund EE chainspec, waits for bridge completion, spends bridged EE funds, and waits through confirmed epoch 6
- add deterministic native Schnorr predicate/prover plumbing for local functional tests, including datatool predicate/operator-key overrides
- fix e2e blockers in core: OL inbox next-index handling for queried ranges and checkpoint DA witness generation from preseal execution

## Why
We need an e2e that uses the real bridge to fund an EE chain with no prefunded balances. This validates the actual DRT -> bridge DT -> ASM deposit -> OL inbox -> EE balance path instead of relying on devchain prefunds or debug deposit paths. Native Schnorr keeps the local harness non-permissive without requiring SP1 proving on a laptop.

## Verification
- python3 -m py_compile on the touched functional-test Python files
- cargo fmt --check on the touched Rust packages
- git diff --check
- cargo check -p strata-zkvm-hosts
- cargo check -p strata-proofimpl-checkpoint -p strata-proofimpl-alpen-acct -p strata-proofimpl-alpen-chunk -p strata-proofimpl-evm-ee-stf
- cargo check -p strata-datatool
- CARGO_INCREMENTAL=0 cargo check -p strata --features prover
- local real bridge e2e passed before final conflict cleanup: DRT 34d543216e426ca26e0dff805366483665cad2a362adfeac47f7e06c4f8370ce, bridge DT 7f2a4009ecdbe741d97b368ef43105cd4439199af63563c517e61051aac17bb1, EE balance 10000000000000000000, spend succeeded, confirmed epoch 6

Note: a later focused sequencer test rerun was blocked by local disk exhaustion while recompiling Reth dependencies, not by a code assertion failure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1131,7 +1131,7 @@ dependencies = [
  "threadpool",
  "tokio",
  "tracing",
- "zkaleido 0.1.0 (git+https://github.com/alpenlabs/zkaleido?tag=v0.1.0-alpha-rc25)",
+ "zkaleido 0.1.0 (git+https://github.com/alpenlabs/zkaleido?rev=f287d06b91a140bffff9ab7dbf037296b6915581)",
  "zkaleido-sp1-host",
 ]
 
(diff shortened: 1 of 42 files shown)


voidash:
there is no plan on doing this at all. This was the stepping stone for finding the bugs which is #1803 actually. there was a branch confusion which lead to this being opened.

PR #1801, opened 2026-05-11
fix(ol): correct checkpoint DA and inbox indexes

voidash:
## Summary
- compute checkpoint DA witness bytes from preseal-only OL block execution, matching the state root authenticated by checkpoint proofs
- keep full epoch replay for logs/header validation so manifest processing is still checked separately
- fix OL chain tracker inbox range bookkeeping so non-empty ranges report the next inbox index after the last returned block, while empty ranges preserve the before-range index
- remove stale direct `strata` prover dependencies that became unused after sharing the checkpoint DA helper

## Bug / context
This addresses the Slack-reported OL prover failure under STR-3412:

`DA witness does not match authenticated preseal state root: ChainIntegrity`

The old prover input path built DA by replaying full OL blocks through `DaAccumulatingState`. Full replay includes L1 manifest processing, so deposits, limbo deposits, or predicate-key updates could be included in the DA blob. The checkpoint proof verifies the DA blob against the terminal block's `preseal_state_root`, which is taken before manifest processing. In those cases the witness described `epoch_start -> post-manifest`, while the proof expected `epoch_start -> preseal`.

The inbox index fix is a separate local e2e finding: `get_inbox_messages` was deriving `next_inbox_msg_idx` from the block before the range instead of the last block in the returned range.

## Verification
- `cargo fmt --check -p strata -p strata-ol-checkpoint -p alpen-ee-sequencer`
- `git diff --check`
- `CARGO_INCREMENTAL=0 cargo check -p strata-ol-checkpoint --lib`
- `CARGO_INCREMENTAL=0 cargo check -p strata --features prover --no-default-features`
- `CARGO_INCREMENTAL=0 cargo check -p alpen-ee-sequencer --lib`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13927,7 +13927,6 @@ dependencies = [
  "strata-ol-state-provider",
  "strata-ol-state-support-types",
  "strata-ol-state-types",
- "strata-ol-stf",
  "strata-paas",
  "strata-params",
  "strata-predicate",
--- a/bin/strata/Cargo.toml
+++ b/bin/strata/Cargo.toml
@@ -24,8 +24,6 @@ debug-utils = [
 ]
 prover = [
   "dep:borsh",
-  "dep:strata-ol-state-support-types",
-  "dep:strata-ol-stf",
   "dep:strata-paas",
   "dep:strata-predicate",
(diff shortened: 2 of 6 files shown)


storopoli:
@codex review this pr

PR #1803, opened 2026-05-11
fix(ol): correct checkpoint DA and inbox indexes

voidash:
## Summary
- compute checkpoint DA witness bytes from preseal-only OL block execution, matching the state root authenticated by checkpoint proofs
- keep full epoch replay for logs/header validation so manifest processing is still checked separately
- fix OL chain tracker inbox range bookkeeping so non-empty ranges report the next inbox index after the last returned block, while empty ranges preserve the before-range index
- remove stale direct `strata` prover dependencies that became unused after sharing the checkpoint DA helper

## Bug / context
This addresses the Slack-reported OL prover failure:

`DA witness does not match authenticated preseal state root: ChainIntegrity`

The checkpoint proof authenticates the terminal block's `preseal_state_root`. That root is taken after normal OL block execution and before L1 manifest processing. The old prover input path built DA by replaying full OL blocks through `DaAccumulatingState`, which also includes terminal manifest processing. When a manifest mutates DA-tracked state, the witness describes the wrong transition:

```text
expected by proof: epoch_start -> preseal_state_root
old DA witness:    epoch_start -> post-manifest state
```

This is intermittent because empty manifests do not add extra writes. Real bridge activity makes the issue easier to hit because deposits, malformed deposits routed to limbo, and predicate-key updates can all change manifest-side OL state.

### DIFF
--- a/crates/ol/checkpoint/src/context.rs
+++ b/crates/ol/checkpoint/src/context.rs
@@ -8,10 +8,12 @@ use std::{
 use strata_asm_proto_checkpoint_types::CheckpointPayload;
 use strata_checkpoint_types::EpochSummary;
 use strata_identifiers::{Epoch, EpochCommitment, OLBlockCommitment};
-use strata_ol_chain_types_new::{OLBlock, OLBlockHeader, OLBlockId, OLLog};
+use strata_ol_chain_types_new::{OLBlock, OLBlockHeader, OLBlockId, OLLog, OLTxSegment};
 use strata_ol_state_support_types::{DaAccumulatingState, MemoryStateBaseLayer};
 use strata_ol_state_types::OLState;
-use strata_ol_stf::execute_block_batch;
+use strata_ol_stf::{
+    BlockContext, BlockExecInput, BlockInfo, execute_block_batch, execute_block_inputs,
+};
 use strata_primitives::nonempty_vec::NonEmptyVec;
 use strata_storage::NodeStorage;
 use tracing::{debug, warn};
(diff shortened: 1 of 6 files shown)


crates/ol/checkpoint/src/context.rs
```diff
@@ -375,22 +377,75 @@ fn replay_epoch_and_compute_da<C: CheckpointWorkerContext>(
...
+    Ok((da_bytes, logs, terminal_header))
+}
 
-    // Extract the DA blob from the accumulating layer.
-    let da_bytes = da_state
+/// Computes the epoch DA payload using only preseal block execution.
+///
+/// Manifest processing is proven by full block replay and terminal header
+/// equality. The DA witness is verified against the terminal preseal root, so
+/// L1-derived manifest mutations must not be included in this blob.
```
storopoli, line 396: non-blocking: make sure to add in the docstring that the argument `blocks: &[OLBlock]` takes ALL blocks including the terminal block. This is ambiguous given the current docstring.

storopoli (approved):
ACK 7cd253ee552a6d5f0f0d119713ca5330889ac9b2

storopoli:
@codex review this pr

voidash:
Superseded by draft PR #1806 on branch , which includes the same core fixes plus the canonical-valid epoch selection fix from SP1 e2e verification.

voidash:
Superseded by draft PR #1806 on branch `fix/ol-valid-canonical-epoch-selection`, which includes the same core fixes plus the canonical-valid epoch selection fix from SP1 e2e verification.

PR #1806, opened 2026-05-11
fix(ol): select valid canonical epoch commitments

voidash:
## Summary

This PR fixes OL canonical block and epoch commitment selection so callers prefer terminal blocks with `BlockStatus::Valid` instead of blindly taking the first stored commitment.

It also includes the previously validated checkpoint DA/inbox-index fixes needed for the real bridge e2e path.

## Bug Context

During the real bridge e2e run with staging-v2-like SP1 proving, the bridge deposit path succeeded, but the EE update path previously failed because the OL tracker/checkpoint prover could resolve an epoch terminal block from an invalid fork. That produced failures like `No OL state found for terminal block ...` even though the valid canonical block at the same slot existed.

The root cause is that epoch summaries/commitments can contain multiple candidates for the same epoch/slot, and the old lookup path selected the first stored entry. The first entry is not guaranteed to be valid.

## Fix

- Prefer `BlockStatus::Valid` in OL canonical block lookup.
- Prefer valid terminal blocks when RPC/prover code resolves canonical epoch commitments.
- Re-check the valid canonical epoch commitment after checkpoint proof completion before advancing the checkpoint proof cursor.
- Keep fallback behavior to the first stored entry when no valid block is available.

## Verification

- `cargo fmt --check`
- `cargo check -F sequencer -F strata/sp1 -F debug-utils -F test-mode -F debug-asm --bin strata --bin alpen-client --bin strata-dbtool`
- Local real bridge e2e in SP1 reserved-network mode: `test_real_bridge_deposit_withdraw` passed in datadir `functional-tests/_dd/12-1-tbtyw`. Deposits credited OL/EE, SAU landed, and withdrawal settled on regtest.

### DIFF
--- a/bin/strata/src/prover/spec.rs
+++ b/bin/strata/src/prover/spec.rs
@@ -8,6 +8,7 @@ use std::{fmt, sync::Arc};
 
 use async_trait::async_trait;
 use borsh::{BorshDeserialize, BorshSerialize, io::Error as BorshIoError};
+use strata_db_types::traits::BlockStatus;
 use strata_identifiers::{Epoch, EpochCommitment};
 use strata_ol_checkpoint::compute_epoch_preseal_da_diff;
 use strata_paas::{ProofSpec, ProverError as PaasError, ProverResult};
@@ -85,9 +86,7 @@ fn fetch_input_blocking(
     debug!(%epoch_index, "fetching checkpoint proof input (blocking)");
 
     // Ensure this task still matches the canonical commitment for the epoch.
-    let canonical_commitment = storage
-        .ol_checkpoint()
-        .get_canonical_epoch_commitment_at_blocking(epoch)?
+    let canonical_commitment = canonical_valid_epoch_commitment(&storage, epoch)?
         .ok_or(ProverError::EpochCommitmentNotFound(epoch_index))?;
     if canonical_commitment != task_commitment {
         return Err(ProverError::StaleTaskCommitment {
@@ -182,4 +181,25 @@ fn fetch_input_blocking(
         parent,
         da_state_diff_bytes,
     })
+}
+
+pub(crate) fn canonical_valid_epoch_commitment(
+    storage: &NodeStorage,
+    epoch: Epoch,
+) -> Result<Option<EpochCommitment>, ProverError> {
+    let commitments = storage
+        .ol_checkpoint()
+        .get_epoch_commitments_at_blocking(epoch)?;
+
+    for commitment in &commitments {
+        let block_id = *commitment.last_blkid();
+        if matches!(
+            storage.ol_block().get_block_status_blocking(block_id)?,
+            Some(BlockStatus::Valid)
+        ) {
+            return Ok(Some(*commitment));
+        }
+    }
+
+    Ok(commitments.first().copied())
 }
--- a/bin/strata/src/prover/spec.rs
+++ b/bin/strata/src/prover/spec.rs
@@ -8,7 +8,6 @@ use std::{fmt, sync::Arc};
 
 use async_trait::async_trait;
 use borsh::{BorshDeserialize, BorshSerialize, io::Error as BorshIoError};
-use strata_db_types::traits::BlockStatus;
 use strata_identifiers::{Epoch, EpochCommitment};
 use strata_ol_checkpoint::compute_epoch_preseal_da_diff;
 use strata_paas::{ProofSpec, ProverError as PaasError, ProverResult};
@@ -86,7 +85,8 @@ fn fetch_input_blocking(
     debug!(%epoch_index, "fetching checkpoint proof input (blocking)");
 
     // Ensure this task still matches the canonical commitment for the epoch.
-    let canonical_commitment = canonical_valid_epoch_commitment(&storage, epoch)?
+    let canonical_commitment = storage
+        .find_valid_epoch_commitment_at_blocking(epoch)?
         .ok_or(ProverError::EpochCommitmentNotFound(epoch_index))?;
     if canonical_commitment != task_commitment {
         return Err(ProverError::StaleTaskCommitment {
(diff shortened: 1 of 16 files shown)


bin/strata/src/rpc/provider.rs
```diff
@@ -70,10 +71,26 @@ impl OLRpcProvider for NodeRpcProvider {
         &self,
         epoch: Epoch,
     ) -> DbResult<Option<EpochCommitment>> {
-        self.storage
+        let commitments = self
+            .storage
             .ol_checkpoint()
-            .get_canonical_epoch_commitment_at_async(epoch)
-            .await
+            .get_epoch_commitments_at_async(epoch)
```
krsnapaudel: Why is this change needed? Is it because `ol_checkpoint` manager is not doing this correctly? Should `get_canonical_epoch_commitment_at_async` be updated?

bewakes: It seems like to get the correct canonical commitment, it needs to also check ol block db, which is not present inside the checkpoint manager. Ideally we would like to have this at one place so that other places can also use it.
I wonder how clean it would be to store a reference of OL block manger inside checkpoint manager.

krsnapaudel: Ah I see. Isn't it strange that we would seal an epoch with a block that is not valid?

bin/strata/src/prover/spec.rs
```diff
@@ -199,3 +182,24 @@ fn fetch_input_blocking(
...
+        let block_id = *commitment.last_blkid();
+        if matches!(
+            storage.ol_block().get_block_status_blocking(block_id)?,
+            Some(BlockStatus::Valid)
+        ) {
+            return Ok(Some(*commitment));
+        }
+    }
+
+    Ok(commitments.first().copied())
```
bewakes: If we haven't returned by this point then we don't have valid blocks. Let's just return `None` here.

crates/ol/checkpoint/src/context.rs
```diff
@@ -375,22 +377,75 @@ fn replay_epoch_and_compute_da<C: CheckpointWorkerContext>(
...
-    let mut da_state = DaAccumulatingState::new(ol_state);
 
-    let logs = execute_block_batch(&mut da_state, &epoch_blocks, &prev_terminal_header)
+    let mut replay_state = MemoryStateBaseLayer::new(ol_state_raw.clone());
+    let logs = execute_block_batch(&mut replay_state, &epoch_blocks, &prev_terminal_header)
         .map_err(|e| anyhow::anyhow!("epoch block replay failed: {e}"))?;
 
     let terminal_header = epoch_blocks.ensured_last().header().clone();
+    let da_bytes =
+        compute_epoch_preseal_da_diff(&ol_state_raw, &epoch_blocks, &prev_terminal_header)?;
```
bewakes: This duplicates the batch execution. Is there any reason we can't use `execute_block_batch` by passing `DaAccumulatingState`? Any state transition/computation fix should be inside that function.

crates/storage/src/managers/ol.rs
```diff
@@ -141,11 +141,17 @@ impl OLBlockManager {
         tip: Slot,
     ) -> DbResult<Option<OLBlockCommitment>> {
         let blocks = self.get_blocks_at_height_blocking(tip)?;
-        // TODO(STR-2105): determine how to get the canonical block. for now it is just the first
-        // one
+        for id in &blocks {
```
bewakes, line 146: I think the TODO comment is still valid. There can be scenario when there are multiple valid blocks.

crates/ol/checkpoint/src/context.rs
bewakes: At a closer look I thought that `compute_epoch_preseal_da_diff` is required to correctly compute DA because `execute_block_batch` updates the `DaAccumulatingState` with manifests as well, which we don't want for DA.

But then, `compute_epoch_preseal_da_diff` is just doing more or less what `execute_block_batch` does i.e. execute everything including manifests. DA still does not seem to be correctly computed using this new function either. A testcase that executes batch containing deposit manifests and compares the pre_seal root should help verify this.

bin/strata/src/rpc/provider.rs
```diff
@@ -70,10 +71,26 @@ impl OLRpcProvider for NodeRpcProvider {
...
+            .await?;
+
+        for commitment in &commitments {
```
bewakes: I think this can be consolidated into ol checkpoint manager. Right now it seems to be scattered: sync version is within prover code and async version is in binary. Doesn't need to be methods of OL checkpoint manager, just a function each for sync and async version should be fine for now.

bewakes (changes requested):

bin/strata/src/prover/spec.rs
```diff
@@ -199,3 +182,24 @@ fn fetch_input_blocking(
         da_state_diff_bytes,
     })
 }
+
+pub(crate) fn canonical_valid_epoch_commitment(
```
voidash: async function for this . so that this is reusable across and not just in prover/spec

bin/strata/src/rpc/provider.rs
```diff
@@ -70,10 +71,26 @@ impl OLRpcProvider for NodeRpcProvider {
         &self,
         epoch: Epoch,
     ) -> DbResult<Option<EpochCommitment>> {
-        self.storage
+        let commitments = self
+            .storage
             .ol_checkpoint()
```
voidash: struct OLBlockmanager {
  ops: ..,
cach: ...,
olblkmgr: Arc<OlBlockMgr>,
}

bin/strata/src/prover/spec.rs
delbonis: Should be `find_` to indicate *what* it's doing.

delbonis (changes requested):
nit, but significant

voidash:
The checkpoint DA/preseal state-root mismatch fix belongs to STR-3412, which Bibek is already tracking separately. I’m going to remove that part from this PR so #1806 stays focused on valid canonical epoch commitment selection.

voidash:
Closing this instead of reviving/rebasing it.

callers should not blindly treat the first stored epoch commitment as canonical when checkpoint storage can contain fork candidates.
- It is stale/conflicting against current `main` and would need a real rework, not a simple rebase.
- The final approach still equates "canonical" with "first `BlockStatus::Valid` candidate". That is better than selecting an invalid block, but it does not resolve the case where multiple valid blocks exist at the same slot/epoch. Canonical selection needs to come from fork-choice/finalization semantics, not insertion order.
- The earlier checkpoint DA/preseal/inbox-index work was already split out conceptually and should stay tracked separately rather than being coupled to this fix.

A replacement PR should start fresh from current `main`, define one canonical epoch-commitment resolution path, update all call sites that need canonical semantics, preserve explicit non-canonical/raw checkpoint lookup where needed, and include tests for both invalid-before-valid and multiple-valid fork candidates.

PR #1810, opened 2026-05-12
STR 3335 use `sequencer_predicate` for block sequencer signature check

krsnapaudel:
## Description

Verify OL block signatures against the ASM checkpoint sequencer predicate

### Summary
New-OL block signature checks in block assembly and fork choice still used the legacy `RollupParams::cred_rule`. This moves both to the Checkpoint subprotocol's `sequencer_predicate` (from ASM params), so a block accepted by the assembler is validated the same way when it reaches fork choice, and sequencer authorization is anchored in the ASM rather than the legacy rollup params.

### Changes
- **`strata-ol-chain-types-new`**: new `validation` module with `verify_sequencer_predicate_signature(&PredicateKey, &Buf32, &Buf64) -> bool` and `sequencer_predicate_requires_signature(&PredicateKey) -> bool`. The verifier dispatches on `PredicateTypeId` — `AlwaysAccept` accepts, `NeverAccept`/`Sp1Groth16`/unknown reject, `Bip340Schnorr` verifies via `strata_crypto::verify_schnorr_sig` against the predicate's x-only pubkey — mirroring how the ASM checkpoint subprotocol interprets the same predicate.
- **Block assembly**: extracts the configured `PredicateKey` once from `AsmParams::checkpoint_config()` and threads it through `BlockasmBuilder`/`BlockasmServiceState`; `check_completion_data` verifies `BlockCompletionData` signatures against it instead of `cred_rule`.
- **FCM**: `FcmContext` carries the sequencer predicate (extracted at construction — a missing checkpoint config is now a startup error); `check_ol_block_proposal_valid` takes `&PredicateKey`. Genesis-block rejection and the `MissingBlockSignature` / `InvalidBlockSignature` errors are preserved.
- **`datatool gen-asm-params`**: new `--seqkey` (sequencer master xpub) → emits `sequencer_predicate = Bip340Schnorr(pk)`, using the same x-only derivation as `datatool genparams --seqkey`. Functional tests thread the sequencer pubkey through so keyed env configs get matching `cred_rule`/`sequencer_predicate`; the `strata_unchecked` config stays `AlwaysAccept`.
- `CredRule` is removed only from the new-OL block-assembly and FCM paths; legacy L2 block credential checks are untouched.

### Compatibility
- The signed/verified message is unchanged: `OLBlockHeader::compute_blkid()` — what the OL signer already signs.
- `CredRule::Unchecked` ↔ `AlwaysAccept`; `CredRule::SchnorrKey(pk)` ↔ `Bip340Schnorr(pk)`. Networks must ensure their ASM checkpoint `sequencer_predicate` matches what `cred_rule` was; `datatool gen-asm-params --seqkey` produces the matching predicate.

### Design note
Uses the configured (genesis) `CheckpointInitConfig.sequencer_predicate`, not the live checkpoint-state predicate. The Checkpoint subprotocol supports predicate rotation; if a deployment rotates it, OL block validation would diverge from ASM until this is changed to read the predicate from current checkpoint state. Out of scope here .

### Testing
- New unit tests: `validation` helpers (AlwaysAccept / Bip340Schnorr valid+invalid / NeverAccept / requires-signature); `datatool::resolve_sequencer_predicate`.
- Migrated block-assembly service tests and FCM proposal-validation tests to predicates (AlwaysAccept-with-garbage-sig accepted, unsigned-rejected under Bip340Schnorr, valid-sig accepted, slot-0 genesis rejected).
- `cargo test` + `cargo clippy --all-targets` + `cargo fmt --check` green for `strata-ol-chain-types-new`, `strata-ol-block-assembly`, `strata-consensus-logic`, `strata-datatool`; `cargo check -p strata` passes; `ruff check` clean on the functional-test changes. Keyed functional env configs now exercise the `Bip340Schnorr` path.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-3335](https://alpenlabs.atlassian.net/browse/STR-3335)

[STR-3335]: https://alpenlabs.atlassian.net/browse/STR-3335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


delbonis (approved):
Some good cleanups here, looks good I think.

bin/strata/src/fcm.rs
```diff
@@ -140,6 +140,12 @@ pub(crate) fn start(
     csm_monitor: Arc<ServiceMonitor<CsmWorkerStatus>>,
 ) -> Result<FcmServiceHandle> {
     let checkpoint_state_rx = nodectx.status_channel().subscribe_checkpoint_state();
+    let sequencer_predicate = nodectx
+        .asm_params()
+        .checkpoint_config()
+        .ok_or_else(|| anyhow!("ASM checkpoint config required for FCM"))?
+        .sequencer_predicate
+        .clone();
```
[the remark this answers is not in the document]
storopoli: I don't think we can do this right now. Cc @prajwolrg

storopoli (approved):
ACK 03545b1f24ce1c95b45d973a059b162bfd0f868c

bin/datatool/src/cmd/asm_params.rs
```diff
@@ -181,3 +182,42 @@ pub(super) fn exec(cmd: SubcAsmParams, ctx: &mut CmdContext) -> anyhow::Result<(
...
+    let Some(seqkey) = seqkey.map(str::trim) else {
+        return Ok(PredicateKey::always_accept());
+    };
+
+    let xpub = Xpub::from_str(seqkey)?;
+    Ok(PredicateKey::new(
+        PredicateTypeId::Bip340Schnorr,
+        xpub.to_x_only_pub().serialize().to_vec(),
+    ))
+}
```
purusang, line 196: Not saying this is wrong, but I think we have inconsistency in generating params. While generating operator keys we [take priv key and generate pubkey](https://github.com/alpenlabs/alpen/blob/e49857f37e35882916d02347469daa794674dc3d/bin/datatool/src/cmd/params.rs#L64) for them in params while we take [pubkey for sequencer](https://github.com/alpenlabs/alpen/blob/e49857f37e35882916d02347469daa794674dc3d/bin/datatool/src/cmd/params.rs#L46).
Our current approach for generating sequencer schnorr key relies on an external step to generate pubkey properly which seems like an unnoticeable way to introduce mismatch if not done carefully. I am inclining towards taking priv key to generate pubkey instead of  taking Xpub to generate x_only_pub.

bin/strata/src/fcm.rs
```diff
@@ -140,6 +140,12 @@ pub(crate) fn start(
     csm_monitor: Arc<ServiceMonitor<CsmWorkerStatus>>,
 ) -> Result<FcmServiceHandle> {
     let checkpoint_state_rx = nodectx.status_channel().subscribe_checkpoint_state();
+    let sequencer_predicate = nodectx
+        .asm_params()
```
purusang, line 144: I might be wrong, but why do we depend on asm-params in OL? I know same binary runs both services and loads both ol and asm params but should not there be a separation of concern? 
I have seen similar thing in [ol rpc](https://github.com/alpenlabs/alpen/blob/e49857f37e35882916d02347469daa794674dc3d/bin/strata/src/rpc/mod.rs#L148) as well. 
cc: @storopoli @delbonis

bewakes (approved):
This looks good. But I think the points raised by @purusang are valid.

bin/strata/src/fcm.rs
storopoli: Yes, we should have proper guidance and policy on Params "boundaries".
Can we move this to a ticket?

krsnapaudel: @purusang We do have to get the predicate right? And that comes from `asm_params`.
[Created this ticket](https://alpenlabs.atlassian.net/browse/STR-3460). Probably for protocol team. cc: @prajwolrg

bin/datatool/src/cmd/asm_params.rs
krsnapaudel: @purusang It seems `genparams --seqkey <xpub>` uses pubkey. Do you want to make these changes now? It likely has other implications. Created [this ticket](https://alpenlabs.atlassian.net/browse/STR-3462). We may need a discussion on this.

PR #1809, opened 2026-05-12
chore(deps): bump SP1 to v6 and align strata-common, asm, and zkaleido

prajwolrg:
## Description

Bumps SP1 from `5.2.1` to `6.1.0` and lifts the rest of the prover stack to consume the corresponding releases:

- `strata-common`: `v0.1.0-alpha-rc19` → `v0.1.0-alpha-rc21`
- `asm`: `v0.1-alpha.7` → `v0.1-alpha.8`
- `zkaleido`: `v0.1.0-alpha-rc25` → `v0.1-beta.1`

The dependency bumps drag in a handful of breaking API changes, so the rest of the diff is mechanical adaptation rather than new behavior:

- **`SP1Host` init is now async.** `crates/zkvm/hosts/src/sp1.rs` swaps the `LazyLock<Arc<SP1Host>>` statics for `async fn checkpoint_host` / `alpen_chunk_host` / `alpen_acct_host` backed by `tokio::OnceCell`, and the prover deadline moves from a post-init `with_deadline` builder into `SP1HostConfig` passed at init time. Call sites in `bin/strata`, `bin/alpen-client`, and `bin/prover-perf` are updated accordingly. The `remote-prover` feature is no longer a separately-gated flag in `zkaleido-sp1-host`, so the workspace feature wiring drops it.
- **`SP1Groth16Verifier::load` gained two parameters** (`vk_root_bytes` and a `verify_vk_root` bool); `bin/datatool`'s acct/checkpoint predicate builders and `provers/sp1/build.rs`'s Groth16 condition computation pass `*sp1_verifier::VK_ROOT_BYTES, true`.
- **`AdministrationInitConfig` now takes a `ConfirmationDepths` struct** instead of a single depth. `bin/datatool/src/cmd/asm_params.rs` fans the existing `--confirmation-depth` value out across every field.
- **`prover-perf` switches from `PerformanceReport` to `ExecutionSummary`** — `ZkVmHostPerf`/`ZkVmProgramPerf` are no longer available, so programs are now run via `ZkVmProgram::execute` and the results table reports cycles + gas (the `success` column is gone since `execute` is fallible and propagates errors).
- **SP1 v6 guests target `riscv64im-succinct-zkvm-elf`** (was `32im`); `provers/sp1/build.rs` follows the rename and stops persisting per-program `.pk` files (SP1 v6 derives the proving key on demand via `blocking::ProverClient::setup`).
[...]

PR #1816, opened 2026-05-13
fix(datatool, provers/sp1): emit borsh-serialized verifier as Sp1Groth16 condition

voidash:
## Summary

#1809 bumped strata-predicate to rc21, whose `Sp1Groth16Verifier::parse_condition` deserializes a full `SP1Groth16Verifier` via borsh (verifying key + folded program_vk_hash + vk_root + vk_hash_tag + require_success). The producer sites in `bin/datatool` and `provers/sp1/build.rs` still emit `sp1_verifier.vk.to_uncompressed_bytes()`, so the on-chain `Sp1Groth16` predicate condition can't be parsed by the runtime verifier.

Switch the three producer sites to `borsh::to_vec(&sp1_verifier)` and add the borsh dep to `bin/datatool/Cargo.toml` (gated on `sp1-builder`) and to `provers/sp1` build-dependencies.

Validated locally by running `test_real_bridge_alpen_cli_e2e` with `SP1_PROVER=network` against the Succinct reserved cluster: seq_no=0 and seq_no=1 SAUs land on OL with zero `verify_claim_witness` failures. Without the fix, every SAU is rejected at parse_condition.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15433,6 +15433,7 @@ dependencies = [
  "argh",
  "bitcoin",
  "bitcoind-async-client",
+ "borsh",
  "rand_core 0.6.4",
  "reth-chainspec",
  "serde_json",
@@ -16709,6 +16710,7 @@ name = "strata-sp1-guest-builder"
 version = "0.3.0-alpha.1"
 dependencies = [
  "bincode",
+ "borsh",
  "cargo_metadata 0.19.2",
  "cfg-if",
  "once_cell",
--- a/bin/datatool/Cargo.toml
+++ b/bin/datatool/Cargo.toml
@@ -27,6 +27,7 @@ strata-predicate.workspace = true
 strata-primitives.workspace = true
 strata-snark-acct-runtime.workspace = true
 
+borsh = { workspace = true, optional = true }
 sp1-verifier = { workspace = true, optional = true }
 strata-sp1-guest-builder = { path = "../../provers/sp1", optional = true }
 zkaleido-sp1-groth16-verifier = { workspace = true, optional = true }
@@ -48,6 +49,7 @@ default = ["btc-client"]
 btc-client = ["dep:bitcoind-async-client", "dep:tokio"]
 sp1-builder = [
   "strata-sp1-guest-builder/sp1-dev",
+  "dep:borsh",
   "dep:sp1-verifier",
   "dep:zkaleido-sp1-groth16-verifier",
 ]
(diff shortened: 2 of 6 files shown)


prajwolrg (approved):
Thanks! I somehow missed this!

Just so we catch this kind of failure in the CI, could you create a follow up to do an actual alpen proof verification as part of the `prover-eval`? In the asm repo, we do that [here](https://github.com/alpenlabs/asm/blob/main/bin/prover-perf/src/programs/moho.rs#L49-L52). The moho proof verifies an actual Groth16 SP1 proof.

prajwolrg:
The verifier is defined in [`zkaleido-sp1-groth16-adapter`](https://github.com/alpenlabs/zkaleido/blob/main/adapters/sp1/groth16-verifier/src/verifier.rs#L26-L46).

The upstream only has `borsh` and `serde` feature-gated. I know we are trying to move away from `borsh`, but I preemptively used `borsh` encoding for the verifier to unblock v6 upgrade. Previously we could directly use the `Groth16VerifyingKey` and used it it's canonical uncompressed format. @delbonis do you think we should handle that appropriately or is fine as is?

delbonis (approved):
Does this have to be borsh?  Does the upstream sp1 deps force this or is this something we're doing in `strata-predicate`?  Kinda annoyed we're re-adding a dep for this that I've wanted to get rid of.

Approving anyways.

storopoli (approved):
ACK 63780f8264802e75d84c08dbcdb2320d4841b687

Agree with @delbonis about re-adding `borsh` that we want to get rid of. Can we have a spike ticket to investigate workarounds having `borsh` here?

storopoli:
> The verifier is defined in [`zkaleido-sp1-groth16-adapter`](https://github.com/alpenlabs/zkaleido/blob/main/adapters/sp1/groth16-verifier/src/verifier.rs#L26-L46).
> 
> The upstream only has `borsh` and `serde` feature-gated. I know we are trying to move away from `borsh`, but I preemptively used `borsh` encoding for the verifier to unblock v6 upgrade. Previously we could directly use the `Groth16VerifyingKey` and used it it's canonical uncompressed format. @delbonis do you think we should handle that appropriately or is fine as is?

Ooh I missed the comment, @prajwolrg can you please open a ticket to move away from `borsh` in `zkaleido`? @delbonis what should we bring instead as a serialization format (probably `cbor`?)?

prajwolrg:
> Does this have to be borsh? Does the upstream sp1 deps force this or is this something we're doing in `strata-predicate`? Kinda annoyed we're re-adding a dep for this that I've wanted to get rid of.
> 
> Approving anyways.

It doesn't have to be `borsh`. This is something we define [here](https://github.com/alpenlabs/zkaleido/blob/main/adapters/sp1/groth16-verifier/src/verifier.rs#L32). Created a [ticket](https://alpenlabs.atlassian.net/browse/STR-3471) to tackle this later.

PR #1817, opened 2026-05-13
prover: optional in-flight cap and race-safe task spawn

voidash:
## Summary

- **`prover-core`: cap in-flight proofs and fix a double-spawn race.**
  Under SP1 network backpressure, EE provers pile up in-flight and OOM small hosts. `ALPEN_MAX_CONCURRENT_PROVES=N` lets the operator bound concurrency. No default, so behavior matches main unless set. Also: claim a task as `Proving` before spawning its future, so the retry loop can't pick it up twice.

## Validation

`SP1_PROVER=network` against the Succinct reserved cluster. Two deposits land, two SAUs land on OL (`seq_no=0, 1`) with zero verify failures.

## Out of scope (for this PR) but still we need to fix / merge before deployment

- **borsh-serialized predicate condition.** [#1816](https://github.com/alpenlabs/alpen/pull/1816). Without it no SAU lands; 1816 lands first.
- **bitcoincore_rpc `next_request_id` race.** Upstream flake, kills the EE process mid-run.

### DIFF
--- a/bin/alpen-client/src/main.rs
+++ b/bin/alpen-client/src/main.rs
@@ -682,6 +682,29 @@ fn main() {
                     ext.custom_chain.genesis().config.clone().into_rsp()
                 };
 
+                // Optional cap on concurrent in-flight prove tasks per prover.
+                // Each task holds witness state in memory (heavy for the chunk
+                // path's reth range-witness extraction); without a cap, tasks
+                // can pile up when SP1 network proofs lag behind submission
+                // and the process OOMs on resource-constrained hosts.
+                //
+                // Unset by default — behavior matches main. Operators on
+                // small machines (laptops, CI runners) can opt in by setting
+                // `ALPEN_MAX_CONCURRENT_PROVES=N`; the cap then applies to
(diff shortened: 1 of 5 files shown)


evgenyzdanovich:
I am kind of against defining any in-flight caps / parallelization constraints on the `prover-core / paas`. this is conceptually a wrong direction for the crate imo.

the whole idea of **integrated** prover in the first place means that there should be _a clear contract_ between the caller (our case ee block production pipeline) and the prover (part of the binary, just another service). 

Ideally, ee block production must be halted if prover lags behind (e.g. _what is proven constantly lags behind and drifts from **the tip** further and further, accumulating proving backlog_).

Adding such limits with the idea of conceptually allowing **to bombard** (integrated) prover with requests is a bad idea imo. Instead, we should figure out the policy on ee block/chunk production so that we don't have a situation with **400 proving tasks** in the first place -- **and not because prover rejected them, but because caller didnt trigger the proof requests in the first place** (by observing that what he, caller, had asked for hasn't yet been proven).

voidash:
closing because this is bad design

PR #1821, opened 2026-05-14
  chore(ci): update functional tests to use large runners

alexhui01:
## Description

  The "Run functional tests" job in `functional.yml` intermittently fails on the
  standard `ubuntu-latest` runner due to resource exhaustion. The failures are
  non-deterministic — re-running the same job with no code changes passes — which
  points to runner capacity (CPU/memory contention) rather than test logic.

  This moves that single job to the larger `ubuntu-latest-m` GitHub-hosted runner
  to give the functional suite more headroom and cut down on flaky failures.

  Scope is intentionally limited to "Run functional tests" only. Every flaky
  incident collected in STR-3381 originates from that job; all other jobs stay on
  `ubuntu-latest` and can be migrated individually if they start showing the same
  symptoms.

  Also adds `.github/actionlint.yaml` to declare `ubuntu-latest-m` as a known
  runner label. It is not in actionlint's built-in label list, so without this the
  `actionlint` check fails on the unknown label. The label is confirmed available
  in the org runner settings.

  ### Type of Change

  - [x] Other — CI / infrastructure change

  ## Notes to Reviewers

  - No workflow logic changes beyond the `runs-on` label and the actionlint config.
  - `ubuntu-latest-m` is an org-configured larger runner, confirmed available.

  ## Related Issues

  Part of STR-3381 (tracking flaky tests).

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,4 @@
+# Custom GitHub-hosted larger runners not in actionlint's built-in label list.
+self-hosted-runner:
+  labels:
+    - ubuntu-latest-m
--- a/.github/workflows/functional.yml
+++ b/.github/workflows/functional.yml
@@ -11,7 +11,7 @@ permissions: {}
 jobs:
   extract-rust-version:
     name: Extract Rust toolchain version
-    runs-on: ubuntu-latest-m
+    runs-on: ubuntu-latest
     outputs:
       rust-version: ${{ steps.extract.outputs.rust-version }}
     steps:
@@ -26,7 +26,7 @@ jobs:
 
   lint:
     name: Lint test files
-    runs-on: ubuntu-latest-m
+    runs-on: ubuntu-latest
     timeout-minutes: 60
 
     steps:
@@ -53,7 +53,7 @@ jobs:
 
   run:
     name: Run functional tests
-    runs-on: ubuntu-latest-m
+    runs-on: ubuntu-latest
     needs: extract-rust-version
     timeout-minutes: 60
     env:
@@ -173,7 +173,7 @@ jobs:
 
   functional-tests-success:
     name: Check that all checks pass
-    runs-on: ubuntu-latest-m
+    runs-on: ubuntu-latest
     if: always()
     needs: [lint, run]
     timeout-minutes: 60
--- a/.github/workflows/functional.yml
+++ b/.github/workflows/functional.yml
@@ -11,7 +11,7 @@ permissions: {}
 jobs:
   extract-rust-version:
     name: Extract Rust toolchain version
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-m
     outputs:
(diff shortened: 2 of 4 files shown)


alexhui01:
The workflow actionlint is not true since we have ubuntu-latest-m label available. 
Screenshot 2026-05-14 at 20 52 57

storopoli:
This is failing the action lint as you said. Can you do some sort of commenting or config override to not fail? Once we have that we should be good to go to merge.

bewakes (commented):
~Not sure if `-m` is corrrect. Couldn't find it on docs. Plus there's check failure.~

bewakes (approved):

storopoli (approved):
ACK 4ae1fb7a4a3db4be1c75588248278b4e7c42e3ac

storopoli (changes requested):
@alexhui01 two things:

1. (Most important) All flaky incidents reported in [STR-3381](https://alpenlabs.atlassian.net/browse/STR-3381) are only from "Run  functional tests" job. This is the ONLY place we need a beefier machine. Can you please change to `ubuntu-latest-m` only in that job? If we start to get flakiness in other jobs, we can migrate them to `ubuntu-latest-m` as well. But let's do "Run functional tests" only for now.
2. Can you please add a proper PR description and revise your title? Since we use Squash Merge as the main and only PR merge strategy in this repo, the PR title will be the commit title and the PR body will be the commit message that gets committed into the `main` branch. Hence, it is important to have a proper PR description/body.  Also, here we try to adhere to conventional commits, so a candidate PR title would be: `chore(ci): update functional tests to use large runners`.

alexhui01:
> @alexhui01 two things:
> 
> 1. (Most important) All flaky incidents reported in [STR-3381](https://alpenlabs.atlassian.net/browse/STR-3381) are only from "Run  functional tests" job. This is the ONLY place we need a beefier machine. Can you please change to `ubuntu-latest-m` only in that job? If we start to get flakiness in other jobs, we can migrate them to `ubuntu-latest-m` as well. But let's do "Run functional tests" only for now.
> 2. Can you please add a proper PR description and revise your title? Since we use Squash Merge as the main and only PR merge strategy in this repo, the PR title will be the commit title and the PR body will be the commit message that gets committed into the `main` branch. Hence, it is important to have a proper PR description/body.  Also, here we try to adhere to conventional commits, so a candidate PR title would be: `chore(ci): update functional tests to use large runners`.

@storopoli Updated for both comments.

[STR-3381]: https://alpenlabs.atlassian.net/browse/STR-3381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

storopoli (approved):
ACK eb656bdf93d3c70823eb90d6e1b21e16284dd8da

PR #1826, opened 2026-05-16
chore: remove debug macros from production code

sashaphmn:
Replace dbg! and println! in production code with structured warn! and debug! via tracing.


bewakes:
Thanks for the PR. However, I am closing this since this is something we plan to do in a more focused bigger cleanup task.

PR #1827, opened 2026-05-17
chore(docker): remove stale local stack configs

voidash:
## Summary

- remove the obsolete local full-stack compose path that predates the signer split
- remove the old init-keys script that generated always-accept checkpoint params
- remove stale root docker configs superseded by the current signer-split stack
- update docker README and gitignore around the current generated layout

## Jira

STR-3127

## Validation

- bash -n docker/gen-params-and-elfs.sh docker/init-network.sh docker/strata/entrypoint.sh docker/alpen-client/entrypoint.sh docker/strata-signer/entrypoint.sh docker/bitcoind/entrypoint.sh
- docker compose config for compose-signet.yml
- docker compose config for compose-ol-el-seq.yml
- docker compose config for docker-compose-eest.yml
- docker compose config for docker-compose-p2p-test.yml
- git diff --check -- docker


storopoli (approved):
ACK 252e029ce0b7c085860970ac31abb527ad8e3097

purusang (approved):

PR #1830, opened 2026-05-18
feat: add `/healthz` endpoints for Strata services

storopoli:
## Description

Adds unauthenticated HTTP `GET /healthz` endpoints for Strata service health probes.

This includes a shared `strata-common` health middleware/helper, wires `/healthz` into the `strata` public RPC listener, and adds standalone health servers for `strata-signer` and `alpen-client` on configurable `0.0.0.0:8080` defaults. Docker compose healthchecks now use these endpoints.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

`strata` serves `/healthz` on the existing public RPC port as listener-liveness once the RPC server is accepting requests. `strata-signer` and `alpen-client` use standalone health servers and return `503` until their startup path marks them ready.

AI was used to assist this PR.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-3376


bin/alpen-client/src/main.rs
```diff
@@ -839,6 +851,14 @@ pub struct AdditionalConfig {
     #[arg(long, default_value_t = false)]
     pub dummy_ol_client: bool,
 
+    /// Host for the HTTP health check endpoint.
+    #[arg(long, default_value = DEFAULT_HEALTH_CHECK_HOST)]
+    pub health_check_host: String,
+
+    /// Port for the HTTP health check endpoint.
+    #[arg(long, default_value_t = DEFAULT_HEALTH_CHECK_PORT)]
+    pub health_check_port: u16,
```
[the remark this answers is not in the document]
storopoli: Ok good finding, addressed in 20aa8601932517c2d6d8ab51ae0d5d03a9f209f6

bewakes (approved):

bin/alpen-client/src/main.rs
```diff
@@ -115,6 +116,9 @@ use crate::{
 #[cfg(feature = "sequencer")]
 const ALPEN_EE_BLOCK_TIME_MS_ENV_VAR: &str = "ALPEN_EE_BLOCK_TIME_MS";
 
+const DEFAULT_HEALTH_CHECK_HOST: &str = "0.0.0.0";
```
delbonis, line 120: Should this be 127.0.0.1 so as to not listen on all interfaces by default?

delbonis (approved):
Looks good aside from one comment which might not be actionable.

bin/alpen-client/src/main.rs
storopoli: AFAIK this should be all interfaces. Cc @alexhui01 please advise.

alexhui01: This needs to be 0.0.0.0 for all health check.

bin/strata/src/rpc/mod.rs
```diff
@@ -264,7 +265,13 @@ async fn spawn_public_rpc(deps: RpcDeps) -> Result<()> {
     } else {
         CorsLayer::new()
     };
-    let http_middleware = ServiceBuilder::new().layer(cors);
+    // The public RPC health check is listener-liveness only. Once this server
+    // is accepting requests, `/healthz` stays ready and does not probe workers
+    // or external dependencies.
+    let health_check_state = HealthCheckState::ready();
```
alexhui01: Does this means it always ready once the listener bind.  the comment is correct as it can only be used for liveness.  How about readiness?

docker/compose-ol-el-seq.yml
```diff
@@ -34,10 +34,7 @@ services:
       - alpen_network
     entrypoint: ["/usr/local/bin/entrypoint.sh"]
     healthcheck:
-      test: >-
-        curl -sf -X POST -H 'Content-Type: application/json'
-        -d '{"jsonrpc":"2.0","method":"strata_protocolVersion","params":[],"id":1}'
-        http://localhost:8432
+      test: ["CMD-SHELL", "curl -sf http://localhost:8432/healthz"]
```
alexhui01: Why 8432 here, suppose all /healthz use 8080?

bin/alpen-client/src/main.rs
```diff
@@ -817,6 +828,7 @@ fn main() {
                 // TODO: post update to OL
             }
 
+            health_check_state.mark_ready();
```
alexhui01, line 831: What's difference we use mark_ready vs HealthCheckState::ready() between strata and alpen-client?

bin/alpen-client/src/main.rs
storopoli: solving this

bin/strata/src/rpc/mod.rs
storopoli: Yes, for `strata` this becomes ready once the public RPC listener is bound and accepting requests. So it is a liveness/listener-readiness signal, not a deeper application readiness check.

I kept it that way intentionally because the requested `/healthz` should not depend on Bitcoin/OL/EE peer state or transient sync/dependency conditions. For `strata-signer` and `alpen-client`, the standalone health server does start before readiness and only flips to `200` after their startup path finishes. For `strata`, the public RPC listener itself is the startup gate.

bin/alpen-client/src/main.rs
storopoli: `HealthCheckState::ready()` creates the state already set to ready. That is what `strata` uses because its `/healthz` is attached to the public RPC listener itself, so once the listener exists and can answer the request, the probe is ready.

For `alpen-client`, the health listener is separate from Reth’s public RPC listener and starts earlier. If we used `HealthCheckState::ready()` there, Kubernetes could get `200` while the node is still launching Reth and spawning Alpen services. So it starts as not-ready with `HealthCheckState::new()`, then calls `mark_ready()` only after the Reth node is launched and the Alpen background tasks are spawned.

docker/compose-ol-el-seq.yml
storopoli: Changed all ports to 8080 solving this

purusang (approved):
Look good.

storopoli:
Ok @alexhui01 i'mr merging this

PR #1831, opened 2026-05-18
feat(ci): automated SP1 E2E proving test

purusang:
## Description

Adds an automated end-to-end CI workflow that runs the full sequencer stack (OL + EE + signer) with real SP1 Groth16 proofs against an ephemeral signet. Validates that:

1. Checkpoint proofs are generated via Succinct network and confirmed on L1
2. EE batch proofs are accepted by the OL snark account system
3. No `AlwaysAccept` predicates are used at runtime

The workflow runs every 6 hours on main (skipping if no new image builds) and supports manual dispatch with a custom `image_tag`.

### Validation

Latest CI run: https://github.com/alpenlabs/alpen/actions/runs/26081688006/job/76684656063

SP1 proof requests visible at: https://explorer.reserved.succinct.xyz/requester/0xdcc666d28a96687dc4d1a421b2730a33a2026d78

### Key components

- **`.github/workflows/ci-sp1-e2e.yml`** — Workflow with auto-detection of latest image build from `ci-build.yml`
- **`.github/sp1-e2e/run-test.sh`** — Test orchestration: signet startup, params generation (via `init-network.sh`), sequencer stack launch, proof assertions
- **`.github/sp1-e2e/compose-override.yml`** — Swaps local builds for prebuilt ECR images
- **`.github/sp1-e2e/config.seq.toml`** — OL sequencer config with `[prover] backend = "sp1"`
- **`.github/sp1-e2e/sequencer.toml`** — Fast epoch config (5s block time, 12 slots/epoch)
- **`.github/sp1-e2e/.env.ci`** — SP1 network prover env vars

### Design decisions

- Reuses `docker/init-network.sh` for params generation — no code duplication
- Uses `public.ecr.aws/alpenlabs/signet:tmpconf2` (same as staging-v2) with MNEMONIC-derived `custom_signet` wallet
- Waits for coinbase maturity (>101 blocks) before proceeding so the wallet has UTXOs
- Preflight cleanup removes stale containers/volumes from previous runs
- Streams container logs to GH Actions stdout for real-time visibility
- Checkpoint and batch proof assertions run in a single interleaved poll loop
- Generated params printed in collapsible GH Actions groups
- Final state summary (chain status, snark state, BTC height, EE block, params) collected before teardown
- Scheduled dedupe filtered to `--event schedule` to avoid manual runs polluting the skip logic
- `ci-build.yml` also gets a 6hr schedule (offset by 1.5hr) so fresh images are available

### Also included

- `ci-build.yml`: added 6hr schedule with new-commit check gating both `prepare` and `build-sp1-artifacts` jobs

### Type of Change

- [x] New or updated tests

## Notes to Reviewers

The EE batch proof (`Program simulation failed`) is a known issue being investigated separately — the checkpoint proof pipeline is fully working.

The only change to existing code is in `docker/init-network.sh` (passing `--chain-config` and `-s` seqkey flags), which is backwards compatible.

- [x] No

## Checklist

**AI Assistance Notice**: This PR was developed with assistance from Claude Code (Anthropic).

## Related Issues

Closes STR-3440


.github/workflows/ci-build.yml
voidash, line 23: can we include `provers/sp1/**` and `docker/strata-datatool/**` in the trigger list?

.github/workflows/ci-sp1-e2e.yml
```diff
@@ -1,29 +1,194 @@
...
-  stub:
-    name: SP1 E2E (stub)
+  check-new-commits:
+    name: Check for new commits
     runs-on: ubuntu-latest
+    permissions:
+      actions: read
+    outputs:
+      should_run: ${{ steps.check.outputs.should_run }}
+      short_tag: ${{ steps.check.outputs.short_tag }}
```
voidash, line 34: let's also add another output for this 
`build_sha: ${{ steps.check.outputs.build_sha }}`

.github/workflows/ci-sp1-e2e.yml
```diff
@@ -1,29 +1,194 @@
...
+    timeout-minutes: 30
+    environment: prod
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v5
+        with:
+          persist-credentials: false
```
voidash, line 110: can you also add `ref: ${{ needs.check-new-commits.outputs.build_sha }}`

.github/workflows/ci-sp1-e2e.yml
```diff
@@ -1,29 +1,194 @@
...
+            --jq '.[0].headSha // ""')
+
+          if [ -z "${BUILD_SHA}" ]; then
+            echo "::error::No successful image build found on main"
+            echo "should_run=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          SHORT_TAG="${BUILD_SHA:0:7}"
+          echo "short_tag=${SHORT_TAG}" >> "$GITHUB_OUTPUT"
```
voidash, line 70: add `echo "build_sha=${BUILD_SHA}" >> "$GITHUB_OUTPUT"`

voidash (commented):
Docker images are selected by $IMAGE_TAG, which comes from last run ci build 
but our actions doesn't expliclty choose that specific commit which means there will be drift in scripts  
So script might run from one commit and images from another commit

.github/workflows/ci-sp1-e2e.yml
```diff
@@ -1,29 +1,194 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+          INPUT_IMAGE_TAG: ${{ inputs.image_tag }}
+          EVENT_NAME: ${{ github.event_name }}
+        run: |
+          if [ -n "${INPUT_IMAGE_TAG}" ]; then
+            # Manual dispatch with explicit tag
+            SHORT_TAG="${INPUT_IMAGE_TAG}"
+            SHORT_TAG="${SHORT_TAG:0:7}"
+            echo "short_tag=${SHORT_TAG}" >> "$GITHUB_OUTPUT"
```
voidash, line 47: add echo `build_sha=${INPUT_IMAGE_TAG}" >> "$GITHUB_OUTPUT"`

voidash:
There is also one mismatch
 
`CHAIN_SPEC=${CHAIN_SPEC:-dev}`
but our CI is generating params using `testnet-chain.json`

Let's make the change and see if this works. passing `CHAIN_SPEC=testnet`

.github/workflows/ci-build.yml
purusang: Its already in the same file https://github.com/alpenlabs/alpen/blob/672274afc60ced5efd1f204bf1a5c2dffeda32f4/.github/workflows/ci-build.yml#L155

.github/workflows/ci-sp1-e2e.yml
purusang: We cannot do that because I had to fix init-network.sh in my branch because that script lacked alpen predicate so we would not be able to generate proper params that would validate sp1 proofs. We should do this in a followup pr once my pr merges. Created a ticket for it. https://alpenlabs.atlassian.net/browse/STR-3529

.github/workflows/ci-sp1-e2e.yml
purusang: same

.github/workflows/ci-sp1-e2e.yml
purusang: same

.github/workflows/ci-sp1-e2e.yml
purusang: same

purusang:
>CHAIN_SPEC=${CHAIN_SPEC:-dev}

This is a very good catch @voidash . Thank you!

voidash:
Here is another interesting thing
 
  - Root lock: Cargo.lock#L14879-L14881 https://github.com/alpenlabs/alpen/blob/b241108fe186509990ab4f63bddd0857485b1d0c/Cargo.lock#L14879-L14881
  
  - Account guest: guest-alpen-acct/Cargo.lock#L5074-L5077
    https://github.com/alpenlabs/alpen/blob/b241108fe186509990ab4f63bddd0857485b1d0c/provers/sp1/guest-alpen-acct/Cargo.lock#L5074-L5077
    
  - Chunk guest: guest-alpen-chunk/Cargo.lock#L5073-L5076
    https://github.com/alpenlabs/alpen/blob/b241108fe186509990ab4f63bddd0857485b1d0c/provers/sp1/guest-alpen-chunk/Cargo.lock#L5073-L5076
    
  - Checkpoint guest: guest-checkpoint/Cargo.lock#L1951-L1953
    https://github.com/alpenlabs/alpen/blob/b241108fe186509990ab4f63bddd0857485b1d0c/provers/sp1/guest-checkpoint/Cargo.lock#L1951-L1953
 
 **alpha 8 on provers/sp1/\*\* vs  alpha 9 on root**

storopoli:
> Here is another interesting thing
> 
> * Root lock: Cargo.lock#L14879-L14881 https://github.com/alpenlabs/alpen/blob/b241108fe186509990ab4f63bddd0857485b1d0c/Cargo.lock#L14879-L14881
> * Account guest: guest-alpen-acct/Cargo.lock#L5074-L5077
>   https://github.com/alpenlabs/alpen/blob/b241108fe186509990ab4f63bddd0857485b1d0c/provers/sp1/guest-alpen-acct/Cargo.lock#L5074-L5077
> * Chunk guest: guest-alpen-chunk/Cargo.lock#L5073-L5076
>   https://github.com/alpenlabs/alpen/blob/b241108fe186509990ab4f63bddd0857485b1d0c/provers/sp1/guest-alpen-chunk/Cargo.lock#L5073-L5076
> * Checkpoint guest: guest-checkpoint/Cargo.lock#L1951-L1953
>   https://github.com/alpenlabs/alpen/blob/b241108fe186509990ab4f63bddd0857485b1d0c/provers/sp1/guest-checkpoint/Cargo.lock#L1951-L1953
> 
> **alpha 8 on provers/sp1/** vs alpha 9 on root**

I really hate and never understood why these `Cargo.{lock,toml}`s are not a part of the workspace where we can trivially do `asm.workspace = true`. Cc @prajwolrg

.github/sp1-e2e/config.seq.toml
storopoli, line 1: This is almost identical to `docker/configs/config.seq.toml` and will probably more than once lose sync with the original.
Is there a way to us to do an override (similar to what you did with `compose-override.yml`)?

.github/sp1-e2e/sequencer.toml
storopoli, line 1: This is almost identical to `docker/configs/sequencer.toml` and will probably more than once lose sync with the original.
Is there a way to us to do an override (similar to what you did with `compose-override.yml`)?

.github/workflows/ci-sp1-e2e.yml
```diff
@@ -1,29 +1,197 @@
...
+          # Scheduled: skip if we already tested this build
+          LAST_SUCCESS=$(gh run list -R "${REPO}" \
+            --workflow ci-sp1-e2e.yml \
+            --event schedule \
+            --status success \
+            --limit 1 \
+            --json headSha \
+            --jq '.[0].headSha // ""')
+
+          if [ "${LAST_SUCCESS}" = "${BUILD_SHA}" ]; then
```
storopoli, line 88: The workflow can record success for a different commit than the image it actually tested. `BUILD_SHA` is resolved from the latest successful `ci-build.yml` run, but the checkout later uses the workflow ref rather than that build SHA, and the scheduled dedupe compares the previous E2E run's `headSha` to `BUILD_SHA`. If main advances to commit C while the latest built image is still B, this run can test image B, record success with head SHA C, and then skip image C once it is built. The E2E run should either checkout the tested build SHA or key the dedupe on the actual tested image/build SHA instead of the workflow run head SHA.

Relevant lines: https://github.com/alpenlabs/alpen/blob/775ba98eedf45a2a06dbc26c117b6ce5a3e52e38/.github/workflows/ci-sp1-e2e.yml#L55-L89 and https://github.com/alpenlabs/alpen/blob/775ba98eedf45a2a06dbc26c117b6ce5a3e52e38/.github/workflows/ci-sp1-e2e.yml#L107-L110

.github/workflows/ci-build.yml
```diff
@@ -6,6 +6,8 @@
 name: CI — Build and Push Images
 
 on:
+  schedule:
+    - cron: '30 4,10,16,22 * * *'  # Every 6hr, ~1.5hr before SP1 E2E test
```
storopoli, line 10: This new scheduled build partly compensates for stale images, but the push trigger still misses inputs that define the SP1 artifacts produced by this workflow. The build job copies ELFs from `provers/sp1/guest-*` and builds/publishes `docker/strata-datatool`, but pushes that only touch those paths will not trigger `ci-build.yml`, so the scheduled SP1 E2E can keep testing stale prover artifacts until a later scheduled build happens. Please include at least `provers/sp1/**`, `docker/strata-datatool/**`, and the local SP1 runner override action path if changes there affect the artifact build.

Path-filter lines: https://github.com/alpenlabs/alpen/blob/775ba98eedf45a2a06dbc26c117b6ce5a3e52e38/.github/workflows/ci-build.yml#L13-L22
Relevant artifact build lines: https://github.com/alpenlabs/alpen/blob/775ba98eedf45a2a06dbc26c117b6ce5a3e52e38/.github/workflows/ci-build.yml#L155-L167 and https://github.com/alpenlabs/alpen/blob/775ba98eedf45a2a06dbc26c117b6ce5a3e52e38/.github/workflows/ci-build.yml#L304-L320

storopoli (changes requested):

barakshani:
>EE batch proofs are accepted by the OL snark account system

Shouldn't we go all the way through and check that they are verified as well (at least until https://alpenlabs.atlassian.net/browse/STR-3481)?

.github/workflows/ci-sp1-e2e.yml
purusang: @storopoli this was already raised by @voidash and I have responded there. PTAL.

purusang:
>Shouldn't we go all the way through and check that they are verified as well (at least until https://alpenlabs.atlassian.net/browse/STR-3481)?

@barakshani, `EE batch proofs are accepted by the OL` means its already verified. That ticket is about verifying update proof in EL before sending to OL.

barakshani:
> > Shouldn't we go all the way through and check that they are verified as well (at least until https://alpenlabs.atlassian.net/browse/STR-3481)?
> 
> @barakshani, `EE batch proofs are accepted by the OL` means its already verified.

Are you sure about it? verification does not happen on mempool, but during OL block building.

.github/sp1-e2e/run-test.sh
```diff
@@ -0,0 +1,337 @@
...
+                echo "PASS: SAU accepted by OL — seq_no ${seq_no}, epoch ${sau_epoch}"
+                sau_accepted=true
+            fi
+        fi
+
+        if [ "${sau_accepted}" = true ] && [ "${sau_confirmed}" = false ]; then
+            local confirmed
+            confirmed=$(get_confirmed_epoch)
+            if [ "${confirmed}" -ge "${sau_epoch}" ]; then
+                echo "PASS: SAU-containing epoch ${sau_epoch} confirmed — confirmed epoch ${confirmed}"
```
barakshani:
```suggestion
                echo "PASS: SAU with seq_no ${seq_no}, received in epoch ${sau_epoch}, confirmed — confirmed epoch ${confirmed}"

.github/sp1-e2e/config.seq.toml
purusang: @storopoli to override a toml file we have to write the whole toml file again so there is almost no benefit of overriding. If we use current approach then in worst case configs may change in structure/field and may fail the test early because of parse failure. We will know soon (~6hr) as it happens.

voidash (approved):

storopoli (approved):
ACK d4822f400f00e6c9f154819e077beffe5e0c54c6

prajwolrg:
> > Here is another interesting thing
> > 
> > * Root lock: Cargo.lock#L14879-L14881 https://github.com/alpenlabs/alpen/blob/b241108fe186509990ab4f63bddd0857485b1d0c/Cargo.lock#L14879-L14881
> > * Account guest: guest-alpen-acct/Cargo.lock#L5074-L5077
> >   https://github.com/alpenlabs/alpen/blob/b241108fe186509990ab4f63bddd0857485b1d0c/provers/sp1/guest-alpen-acct/Cargo.lock#L5074-L5077
> > * Chunk guest: guest-alpen-chunk/Cargo.lock#L5073-L5076
> >   https://github.com/alpenlabs/alpen/blob/b241108fe186509990ab4f63bddd0857485b1d0c/provers/sp1/guest-alpen-chunk/Cargo.lock#L5073-L5076
> > * Checkpoint guest: guest-checkpoint/Cargo.lock#L1951-L1953
> >   https://github.com/alpenlabs/alpen/blob/b241108fe186509990ab4f63bddd0857485b1d0c/provers/sp1/guest-checkpoint/Cargo.lock#L1951-L1953
> > 
> > **alpha 8 on provers/sp1/** vs alpha 9 on root**
> 
> I really hate and never understood why these `Cargo.{lock,toml}`s are not a part of the workspace where we can trivially do `asm.workspace = true`. Cc @prajwolrg

The guest codes are treated as a separate workspace mostly for the development. If it was part of the workspace, we'd need to build the ELF for all other workspace related things, slowing development/CI, and everything down the line.

PR #1835, opened 2026-05-19
fix(ee): preserve blockhash witness lookups

storopoli:
## Description

Fixes `BLOCKHASH(n)` witness handling across both pre-chunk ancestor headers and blocks executed earlier in the same chunk.

Ancestor headers are stored in a `BTreeMap`, so iteration is ascending by block number. The previous tuple-window validation treated `(parent, child)` as `(child, parent)`, causing multi-header ancestor witnesses to panic. The old pairwise map construction would also omit the highest ancestor header.

This PR now builds `block_hashes` with the direct invariant:

`block_hashes[n] = hash(header n)`

It also wires chunk execution to update the EVM partial state after each successfully executed block, so later blocks in the same chunk can resolve `BLOCKHASH` for earlier same-chunk block numbers instead of seeing the default zero hash.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The serialized `EvmPartialState` format is unchanged. The behavioral changes are:

- Ancestor header validation now follows parent-child ordering.
- Every included ancestor header contributes its own block hash to the witness lookup map.
- Chunk execution calls a post-block partial-state hook after successful merges.
- EVM partial state uses that hook to record executed block headers for subsequent same-chunk `BLOCKHASH` lookups.

Added tests cover:

- Single-ancestor witnesses populate the block hash map.
- Multi-header ancestor witnesses no longer panic.
- Every ancestor block number maps to that header’s own hash.
- Codec roundtrip preserves the complete `BLOCKHASH` lookup map.
- Invalid ancestor chains still panic during validation.
- The EVM post-block hook makes an executed block hash available for subsequent blocks.

Used AI to assist in the PR.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-3495

### DIFF
--- a/crates/evm-ee/src/types/partial_state.rs
+++ b/crates/evm-ee/src/types/partial_state.rs
@@ -22,7 +22,7 @@ use crate::{
 /// Partial state for EVM block execution.
 ///
 /// Contains the witness data needed to execute a block: the sparse Merkle Patricia Trie
-/// state, contract bytecodes, and ancestor block headers for BLOCKHASH opcode support.
+/// state, contract bytecodes, and ancestor block headers for `BLOCKHASH` opcode support.
 ///
 /// This struct pre-computes expensive operations (header hashing, block hash map) during
 /// construction to avoid repeated work when preparing witness databases.
@@ -36,7 +36,7 @@ pub struct EvmPartialState {
     /// Ancestor block headers with pre-computed hashes, indexed by block number.
     /// Headers are sealed once during construction to avoid repeated hash computations.
     ancestor_headers: BTreeMap<u64, Sealed<Header>>,
-    /// Pre-computed block hash lookup map for BLOCKHASH opcode.
+    /// Pre-computed block hash lookup map for `BLOCKHASH` opcode.
     /// Built once during construction from sealed ancestor headers.
     block_hashes: HashMap<u64, B256>,
 }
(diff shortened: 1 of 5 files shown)


crates/evm-ee/src/types/partial_state.rs
```diff
@@ -281,3 +233,33 @@ impl Codec for EvmPartialState {
         })
     }
 }
+
+/// Validates ancestor header continuity and builds the `BLOCKHASH` lookup map.
+///
+/// The lookup map stores every ancestor block number with that header's own hash,
+/// matching EVM `BLOCKHASH(n)` semantics.
+fn build_block_hashes(ancestor_headers: &BTreeMap<u64, Sealed<Header>>) -> HashMap<u64, B256> {
```
irnb, line 241: This fixes the pre-chunk ancestor BLOCKHASH case, but I think there is still a separate same-chunk case uncovered.

if block N+1 in the same chunk executes `BLOCKHASH(N)`, the guest `WitnessDB` still will not have that hash and can return the default zero hash.

crates/evm-ee/src/types/tests.rs
```diff
@@ -218,6 +257,49 @@ fn test_evm_partial_state_codec_roundtrip() {
     assert_eq!(decoded.ancestor_headers(), partial_state.ancestor_headers());
 }
 
+#[test]
+fn test_evm_partial_state_block_hashes_include_all_ancestors() {
```
irnb, line 276: Can we add a regression test for the single-ancestor case too?

The first post-genesis chunk can have `ancestor_headers = [genesis]`.

irnb (changes requested):

crates/evm-ee/src/types/partial_state.rs
storopoli: Ok done in an amend commit in 90799b50d86431d75e95758ccfff8816aff3e375

crates/evm-ee/src/types/tests.rs
storopoli: Added tests in 90799b50d86431d75e95758ccfff8816aff3e375

irnb (approved):

irnb (approved):

delbonis (approved):
I kinda hate this but it might be that there is no other way to accomplish it.  Approving since it's clean otherwise.  But I do have a maybe-actionable comment.

I would think that updating the block hash history buffer could be done before processing txs on the *next* block, couldn't it?  I don't remember how it's defined to be done in the standard Ethereum EVM flavor.  Doing it "after" tx exec has the potential to break the state root checks.  Is this definitely safe to do like this?

storopoli:
> I kinda hate this but it might be that there is no other way to accomplish it. Approving since it's clean otherwise. But I do have a maybe-actionable comment.
> 
> I would think that updating the block hash history buffer could be done before processing txs on the _next_ block, couldn't it? I don't remember how it's defined to be done in the standard Ethereum EVM flavor. Doing it "after" tx exec has the potential to break the state root checks. Is this definitely safe to do like this?

The `block_hashes`/`ancestor_headers` data is witness metadata, not part of the Ethereum state trie, so mutating it cannot affect the state root that `merge_write_into_state` checks. In the EVM path, the hook runs only after `merge_write_into_state` has applied the hashed post-state and verified the computed state root against the block header.

Semantically, this also keeps the timing right: block `N`'s txs execute without seeing `BLOCKHASH(N)`, then after block `N` is accepted we record its hash so block `N+1` can resolve `BLOCKHASH(N)`. Moving the update to "before processing the next block" would be equivalent as long as it happens before `execute_block_body` for `N+1`, but the post-success hook makes the invariant a little easier: only successfully processed blocks get added to the lookup metadata.

PR #1836, opened 2026-05-19
fix(ci): repair EE spec test workflows

voidash:
## Summary
- Repair the main and staging EE spec workflows so they install EEST dependencies and run against the Alpen EE RPC endpoint.
- Switch EEST from Shanghai to Prague to match the dev chainspec, which activates Shanghai/Cancun/Prague at genesis.
- Make `strata-zkvm-hosts` optional for `alpen-client` so the native sequencer-only CI build works without `sp1`.
- Deselect the remaining known Prague failure: `tests/frontier/opcodes/test_call.py::test_call_memory_expands_on_early_revert[fork_Prague-state_test]`.
- Increase the EEST transaction wait timeout for hosted runners, which were slower than local on transient-storage reentrancy cases.


.github/workflows/main-eest.yml
```diff
@@ -92,47 +94,43 @@ jobs:
...
+          screen -dmS alpen_ee_env uv run python entry.py --keep-alive alpen_ee
+          for _i in $(seq 1 60); do
+            if curl -sf -X POST http://localhost:30303 \
+              -H "Content-Type: application/json" \
+              -d '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}' > /dev/null 2>&1; then
+              echo "alpen-client RPC is up"
+              exit 0
+            fi
+            sleep 2
+          done
```
storopoli: clever, should we move this to a standalone script in `contrib/` so we can have full shellcheck lints on it?

.github/workflows/main-eest.yml
```diff
@@ -92,47 +94,43 @@ jobs:
...
       - name: Run tests
-        id: runtests
         run: |
           curl -LsSf https://astral.sh/uv/install.sh | sh
           git clone https://github.com/alpenlabs/execution-spec-tests
           cd execution-spec-tests
           uv python install 3.11
           uv python pin 3.11
           uv sync --all-extras
+          uv pip install solc-select
```
storopoli: Urgh! `pip` and `uv` in the same command... :/

.github/workflows/main-eest.yml
```diff
@@ -92,47 +94,43 @@ jobs:
...
           uv run execute remote \
             -m state_test \
             --fork=Prague \
-            --rpc-endpoint=http://localhost:12603 \
+            "--deselect=tests/frontier/opcodes/test_call.py::test_call_memory_expands_on_early_revert[fork_Prague-state_test]" \
+            --rpc-endpoint=http://localhost:30303 \
             --rpc-seed-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 \
             --rpc-chain-id 2892 \
-            --tx-wait-timeout 30 \
+            --tx-wait-timeout 120 \
```
storopoli: Is `120` enough?

storopoli (approved):
ACK 514152649f02b8ea61622593910312c57e3ff13b

non-blocking suggestions

.github/workflows/staging-eest.yml
```diff
@@ -92,19 +92,23 @@ jobs:
           uv python install 3.11
           uv python pin 3.11
           uv sync --all-extras
+          uv pip install solc-select
           uv run solc-select use 0.8.24 --always-install
           uv run execute remote \
             -m state_test \
-            --fork=Shanghai \
+            --fork=Prague \
+            "--deselect=tests/frontier/opcodes/test_call.py::test_call_memory_expands_on_early_revert[fork_Prague-state_test]" \
```
evgenyzdanovich: let's do it properly -- https://github.com/alpenlabs/execution-spec-tests/blob/main/skip_tests.yaml

.github/workflows/main-eest.yml
```diff
@@ -92,47 +94,43 @@ jobs:
...
             -v
-        continue-on-error: true
-
-      - name: Generate a proof of EE blocks execution.
-        working-directory: docker
-        run: |
-          chmod +x test_ee_proof.sh
-          ./test_ee_proof.sh local
```
evgenyzdanovich, line 123: the whole purpose of this entire pipeline, at least the way I designed it, was **to have our EE proofs (for now, chunk proofs) to be generated against the EF tests load** (e.g. against the blocks with all these crazy transactions). 
without this, we are quite literally testing pure Reth EVM against EF tests (which is pointless).

IMO, the essense of the effort should be directed towards restoring that logic (and re-adopting it for chunk proofs).

evgenyzdanovich (changes requested):

.github/workflows/main-eest.yml
voidash: running it multiple times. For the first time, 30 didn't work but 120 worked for me so...

delbonis (approved):

functional-tests/entry.py
```diff
@@ -301,6 +301,10 @@ def main(argv: list[str]) -> int:
         ),
         # Alpen-client (EE) environments
         "alpen_ee": AlpenClientEnv(enable_l1_da=True),
+        # Execution-spec tests only need a sequencer RPC endpoint. Keep this
+        # isolated from L1 DA/prover services so EEST failures reflect EE
+        # execution behavior rather than Bitcoin publishing.
+        "alpen_eest": AlpenClientEnv(fullnode_count=0, enable_l1_da=False),
```
evgenyzdanovich: I dont know why we'd want a **different** alpen client for tests.
IMO we want exactly the same env (with the same exact binary under the hood as for the usual ee tests)

functional-tests/entry.py
```diff
@@ -301,6 +301,10 @@ def main(argv: list[str]) -> int:
         ),
         # Alpen-client (EE) environments
         "alpen_ee": AlpenClientEnv(enable_l1_da=True),
+        # Execution-spec tests only need a sequencer RPC endpoint. Keep this
+        # isolated from L1 DA/prover services so EEST failures reflect EE
```
evgenyzdanovich: with l1 da disabled, I'm not sure if the batch would ever be "proven", even in native (according to the batch lifecycle, something should be posted on l1 before it transitions to proved).

contrib/assert-eest-prover-pipeline.sh
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+usage() {
+    cat <<'EOF'
+Usage: assert-eest-prover-pipeline.sh [options]
+
+Options:
+  --baseline-file FILE     File containing byte offset captured before EEST.
+  --log-file FILE          Alpen-client service.log to inspect.
```
evgenyzdanovich: i'd actually prefer a cleaner assertion path... a test-only RPC (if it doesnt exist yet) exposing parts that you need (chunk/batch/proof status, right from the storage) would not hurt.

we are overly relying on logs, and maintaining such fragile thing would be a nightmare.

P.S. we used to have `getProofStatus` or smth similar (at TN1 era) to actually assert stuff. Similar thing (whatever bits you need -- proof or chunk or batch status) can be easily exposed and used reliably.

bin/alpen-client/src/main.rs
evgenyzdanovich, line 1: can you provide a summary of what's happening here (as well as why we need changes)? 

frankly, I dont see a single reason to change ee binary as part of this work (besides perhaps adding test only RPC to live-introspect some specific state/storage bits).

storopoli (approved):
ACK c9c63c2d4168ed63f430767677b5f5cc60d37fdd

Nice to move all inline shell commands from the yml file into a `contrib/` shell script so we can have syntax highlighting, proper git tracking and shellcheck linting. Thanks!

storopoli:
@evgenyzdanovich your review comments were addressed please take a look

bin/alpen-client/src/main.rs
voidash: agree that that is a bad design. There is no longer alpen-client changes.

evgenyzdanovich:
@codex review

.github/workflows/main-eest.yml
```diff
@@ -81,57 +81,64 @@ jobs:
...
-          screen -dmS basic_env uv run python entry.py --keep-alive basic
-          sleep 20  # Wait for the service to start
+          screen -L -Logfile alpen_eest_env.screen.log -dmS alpen_eest_env uv run python entry.py --keep-alive alpen_eest
+          if ! ../contrib/wait-for-json-rpc.sh http://localhost:30303 240; then
+            cat alpen_eest_env.screen.log || true
+            [ ! -d _dd ] || find _dd -type f -name "service.log" -print -exec tail -100 {} \;
+            exit 1
+          fi
+
+      - name: Capture EEST start block
```
evgenyzdanovich: I dont think this step is required at all (neither here nor in staging counterpart) -- it'll be 1 in most of the cases (because we just waited for the chain to start).
IMO it's also logical to validate "from the start" without any skips (so from 1 or 0, basically hardcoded without capturing exactly when EEST has started)

.github/workflows/main-eest.yml
```diff
@@ -81,57 +81,64 @@ jobs:
...
+      - name: Assert EEST blocks reached confirmed EE/OL status
+        if: always() && steps.runtests.outcome != 'skipped'
         run: |
-          chmod +x test_ee_proof.sh
-          ./test_ee_proof.sh local
+          if ! ./contrib/assert-eest-prover-pipeline.sh \
+              --rpc-endpoint http://localhost:30303 \
+              --bitcoin-rpc-url http://user:password@localhost:18444 \
+              --bitcoin-wallet testwallet \
+              --advanced-from "${EEST_START_BLOCK}" \
```
evgenyzdanovich: again, the point is to make sure that we generated (native) proofs (**chunk proofs** to be precise -- the ones that validate the execution of evm blocks) that cover `EEST_START_BLOCK...EEST_END_BLOCK`. 
Each chunk covers a **_range_** of alpen EE blocks and our target is to make sure that the union of those **_ranges_** exactly covers the range above.

what's being asserted is not exactly that.
what's asserted is that our block production on EE sequencer has continued past EEST execution and at some point `EEST_END_BLOCK` block becomes confirmed/finalized. 
what's needed is to assert that chunk proofs has been generated successfully (in native mode) and that the union of ranges (covered by those chunks) covers the EEST range.

Now, in ideal world with ideal impl, block finality implies that it is covered by the proof: meaning that the block finality is derived from the corresponding proof -- in  other words, block may only ever become final if it's "covered" by the proof (specifically, acct proof that covers the chunk that covers the block; the whole thing is verified in the end by checkpoint proof that's controlled by OL, that's not even included in the fntests env -- AFAIU it's EE only). 

I'm not sure how it works for EE rn and how far we are from it, but the approach feels fragile: especially with proofs in native mode (native mode doesnt allow recursive proofs, e.g. acct proof native execution itself will skip the line where it verifies the chunk proof -- because it is empty) AND given it is still happening a test env through fntests (which is not fully-fledged strata entire stack, so the "ideal world storyline" - a paragraph above - does not suit here, but we seemingly rely on it implicitly).

-----------------

I'd insist on being very explicit and re-target the assertions around the proofs (via RPC to EE sequencer, that just returns the status of the proof from the db) and not block status.

P.S. in one of our testnet envs somewhere, our partners already encountered some messiness with the finality of our blocks (I cant find the ticket, but Stefano may find it): the point being it's not a very reliable mechanism (yet) to validate the prover pipeline in that kind of indirect manner (with all the complications mentioned).

P.S.S. The idea to verify through block finality isn't bad it itself -- quite the opposite -- I do like it, but the env should be prod-like and without native mode for sure, so it's perhaps out of the scope (and overall is much more heavy and complicated test that is closer to the "real env sp1 test" that Puru has created recently; so it  makes sense to create a "hybrid" -- real env with sp1 proving + eest test + block finality check implying that finality does mean that all the proofs across the stack are working).

evgenyzdanovich (changes requested):

.github/workflows/main-eest.yml
voidash: nice feedback. We need explicit proof coverage path then. Instead of `alpne_getBlockStatus`, I plan a small EE RPC that reads chunk status and reports the proof ready chunk ranges and CI will assert the ranges . 

For the real env sp1 test, there will be separate repo and puru has started the discussion regarding it. There we will have manual thing. for the time being do you think this is the right approach? 
FYI, I plan to work on whole orchestration pipeline ASAP while we don't deviate too much from the harness i created.

voidash:
@codex review

purusang (approved):
LGTM

crates/alpen-ee/rpc/server/src/block_status.rs
```diff
@@ -36,25 +40,33 @@ fn canonical_block_number<N: NodeTypesWithDB + ProviderNodeTypes>(
     }
 }
 
+fn hash_to_b256(hash: &[u8]) -> B256 {
```
evgenyzdanovich, line 43: very nit: the helper doesn't make sense, but brings more confusion imo, i'd inline

Cargo.toml
```diff
@@ -414,7 +414,10 @@ bitcoin = { version = "0.32.6", features = ["serde"] }
 bitcoin-bosd = { git = "https://github.com/alpenlabs/bitcoin-bosd", version = "0.10.0", tag = "v0.10.0", features = [
   "ssz",
 ] }
-bitcoind-async-client = { version = "0.10.6", features = ["29_0", "raw_rpc"] }
+bitcoind-async-client = { git = "https://github.com/voidash/bitcoind-async-client", rev = "22abae04fb85a15887410cd3ac9f2a0be9cda988", features = [
```
evgenyzdanovich: what does that client fork and why it is functionally necessary?

evgenyzdanovich (commented):
GREAT, functionally exactly what we need. a nit and a question

crates/alpen-ee/rpc/server/src/block_status.rs
voidash: sure i will inline this

Cargo.toml
voidash: concurrent RPC calls from EE prover repeatedly made the CI fail because it was returning empty response
setting it to 1 made it streamlined and it fixed the problem. 

https://github.com/alpenlabs/bitcoind-async-client/pull/114/changes
once it's merged we can use the new tag here

storopoli: If this needs a bump in `bitcoind-async-client` then we need to get #1860 in first.

evgenyzdanovich: does it mean you accidentally found a bitcoind bug? why would it return empty responses in the first place instead of explicitly pushing back somehow -- throwing rpc error or by any other means?
im afraid that the fix is a crutch and just hides the real problem.

perhaps outside of the scope of this specific PR (like cmon, I understand and can't ask you to fix everything without being ridiculous), but it would be nice to know the real cause of the problem.

evgenyzdanovich: to be precise -- im fine with this fix to unblock the main effort (in this PR), but I'd want us to take a more proper look into that (instead of defining a knob and restricting base layer -- `bitcoind client`)

storopoli: > does it mean you accidentally found a bitcoind bug?

No I don't think he found a bug. He pretty much used `bitcoind-async-client` in a manner that it is not supposed to be used:

> concurrent RPC calls from EE prover repeatedly made the CI

I'm still debating if the fix he proposed in https://github.com/alpenlabs/bitcoind-async-client/pull/114 is really needed for such extreme uncommon use case.

storopoli: @voidash what if you just throttled the RPC calls in this specific test?

evgenyzdanovich: how is it misused and how does throttling to 1 helps? I'd appreciate some context.

voidash: @evgenyzdanovich Here is more detail. 
So the problem is concurrent use of our bitcoin rpc client stack under this CI workload, so reducing concurrency helped with the problem. 

The chain is something like this
1. EEST drives thousands of EE block
2. now we also have lots of  L1 Da/ broadcaster / reader activity
3. multiple async service use same bitcoind_async_client::Client
4. `bitreq::Client` is used with a reused HTTP connection pool on bitcoind_async_client
5.  bitreq has this invariant counter `next_request_id` and `readable_request_id` and these counters turns out must stay ordered  i.e [`readable_request_id <= next request id`]( https://github.com/rust-bitcoin/corepc/blob/eb84c1185a20589cd9dbf0d884e37f94d8755eee/bitreq/src/connection.rs#L394-L401
)
6. now because of all this concurrency and shared connection we reach impossible internal state where `readable_request_id` is greater than `next_request_id`
invariant violated


Now if you look at the [permalink](https://github.com/rust-bitcoin/corepc/blob/eb84c1185a20589cd9dbf0d884e37f94d8755eee/bitreq/src/connection.rs#L394-L401) it is `debug_assert!`

Now proper fix will stay in low level bitcoin RPC client path. we have to do something with bitreq here. But who knows if we run this in release mode it might work :)

@storopoli i will try running this in release mode and if it works i will drop the bitcoind_async_client PR but it's worth noting i guess.

storopoli: Ok great. this might workaround the `debug_assert!`

Cargo.toml
```diff
@@ -414,7 +414,10 @@ bitcoin = { version = "0.32.6", features = ["serde"] }
 bitcoin-bosd = { git = "https://github.com/alpenlabs/bitcoin-bosd", version = "0.10.0", tag = "v0.10.0", features = [
   "ssz",
 ] }
-bitcoind-async-client = { version = "0.10.6", features = ["29_0", "raw_rpc"] }
+bitcoind-async-client = { git = "https://github.com/alpenlabs/bitcoind-async-client", tag = "v0.10.7", features = [
+  "29_0",
+  "raw_rpc",
+] }
```
storopoli: We don't need this diff now. Please restore back to crates.io instead of `git`

storopoli (changes requested):

storopoli (approved):
ACK 7d56e7d6b6e55d1a91e151caa017a941f1b4e926

Have the EEST tests on `release` mode passed?

Cargo.toml
voidash: The test passes with release mode. closing the bitcoind_async_client PR for now.

storopoli (approved):
ACK 43c1b43d975fbf9785f94b737f41e29f64696b48

evgenyzdanovich (approved):
great work, thank you for pushing it through iterations!

evgenyzdanovich:
now the only question is "wen staging counterpart" -- we used to have it run against the env that spins the env from the recent staging images (and does the rest of things the same, runs eest, validates the proof block coverage etc) -- `staging_eest.yml`

I wonder what are the plans on that (I'm fine with removing it if we don't need it anymore but it can be beneficial imo)

PR #1840, opened 2026-05-20
fix(csm): reshape L1Checkpoint and fix ClientState finality (STR-2438)

prajwolrg:
## Description

Fixes [STR-2438](https://alpenlabs.atlassian.net/browse/STR-2438). The CSM's `L1Checkpoint` carried a `BatchInfo` field that didn't reflect the data the CSM actually receives, and `ClientState` carried a `last_finalized_checkpoint` populated by an observation heuristic that could diverge from the depth-driven truth tracked in `CsmWorkerState`. This PR realigns the types with the real inputs and makes the depth-derived value the single source of finality.

### Type of Change

- [x] Refactor

## Notes to Reviewers

Best reviewed commit-by-commit. The first commit is type plumbing only; the second is the behavioral fix; the third is dead-code cleanup made possible by the second.

- [ ] Yes
- [x] No

## Checklist

AI assistance disclosure: Claude Code was used to help draft commits and this PR description; all changes were human-reviewed.

## Related Issues

[STR-2438]: https://alpenlabs.atlassian.net/browse/STR-2438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/csm-types/src/client_state.rs
+++ b/crates/csm-types/src/client_state.rs
@@ -64,7 +64,7 @@ impl ClientState {
     pub fn get_last_epoch(&self) -> Option<EpochCommitment> {
         self.last_seen_checkpoint
             .as_ref()
-            .map(epoch_commitment_for_tip)
+            .map(EpochCommitment::from)
     }
 
     /// Gets the last checkpoint that has already been finalized.
@@ -76,7 +76,7 @@ impl ClientState {
     pub fn get_declared_final_epoch(&self) -> Option<EpochCommitment> {
         self.last_finalized_checkpoint
             .as_ref()
-            .map(epoch_commitment_for_tip)
+            .map(EpochCommitment::from)
     }
 
     /// Gets the next epoch we expect to be confirmed.
@@ -86,10 +86,6 @@ impl ClientState {
             .map(|ck| ck.tip.epoch + 1)
             .unwrap_or(0u32)
     }
-}
-
-fn epoch_commitment_for_tip(checkpoint: &L1Checkpoint) -> EpochCommitment {
-    EpochCommitment::from_terminal(checkpoint.tip.epoch, checkpoint.tip.l2_commitment)
 }
 
 /// A [`ClientState`] wrapper used in StatusChannel.
@@ -171,6 +167,12 @@ impl fmt::Display for L1Checkpoint {
 impl L1Checkpoint {
     pub fn new(tip: CheckpointTip, l1_reference: CheckpointL1Ref) -> Self {
         Self { tip, l1_reference }
+    }
+}
+
+impl From<&L1Checkpoint> for EpochCommitment {
+    fn from(checkpoint: &L1Checkpoint) -> Self {
+        EpochCommitment::from_terminal(checkpoint.tip.epoch, checkpoint.tip.l2_commitment)
     }
 }
 
--- a/crates/csm-types/src/client_state.rs
+++ b/crates/csm-types/src/client_state.rs
@@ -4,11 +4,16 @@
 
 use core::fmt;
 
-use arbitrary::Arbitrary;
-use borsh::{BorshDeserialize, BorshSerialize};
+use arbitrary::{Arbitrary, Unstructured};
+use borsh::{
+    io::{Read, Result as BorshIoResult, Write},
+    BorshDeserialize, BorshSerialize,
+};
 use serde::{Deserialize, Serialize};
-use strata_checkpoint_types::BatchInfo;
-use strata_identifiers::{Epoch, EpochCommitment, L1BlockCommitment, L1BlockId, L1Height, RBuf32};
+use strata_asm_proto_checkpoint_types::CheckpointTip;
+use strata_identifiers::{
+    Epoch, EpochCommitment, L1BlockCommitment, L1BlockId, L1Height, OLBlockCommitment, RBuf32,
+};
 
 /// High level client's checkpoint view of the network. This is local to the client, not
 /// coordinated as part of the L2 chain.
@@ -59,7 +64,7 @@ impl ClientState {
     pub fn get_last_epoch(&self) -> Option<EpochCommitment> {
         self.last_seen_checkpoint
             .as_ref()
-            .map(|c| c.batch_info.get_epoch_commitment())
+            .map(epoch_commitment_for_tip)
     }
 
     /// Gets the last checkpoint that has already been finalized.
@@ -71,16 +76,20 @@ impl ClientState {
     pub fn get_declared_final_epoch(&self) -> Option<EpochCommitment> {
         self.last_finalized_checkpoint
             .as_ref()
-            .map(|ckpt| ckpt.batch_info.get_epoch_commitment())
+            .map(epoch_commitment_for_tip)
     }
 
     /// Gets the next epoch we expect to be confirmed.
     pub fn get_next_expected_epoch_conf(&self) -> Epoch {
         self.last_seen_checkpoint
             .as_ref()
-            .map(|ck| ck.batch_info.get_epoch_commitment().epoch() + 1)
+            .map(|ck| ck.tip.epoch + 1)
             .unwrap_or(0u32)
     }
+}
+
+fn epoch_commitment_for_tip(checkpoint: &L1Checkpoint) -> EpochCommitment {
+    EpochCommitment::from_terminal(checkpoint.tip.epoch, checkpoint.tip.l2_commitment)
 }
 
 /// A [`ClientState`] wrapper used in StatusChannel.
@@ -136,14 +145,20 @@ impl CheckpointL1Ref {
     }
 }
 
-#[derive(
-    Clone, Debug, Eq, PartialEq, Arbitrary, BorshDeserialize, BorshSerialize, Deserialize, Serialize,
-)]
+#[derive(Clone, Debug, Eq, PartialEq, BorshDeserialize, BorshSerialize, Deserialize, Serialize)]
 pub struct L1Checkpoint {
-    /// The inner checkpoint batch info.
-    pub batch_info: BatchInfo,
+    /// Tip published by the ASM checkpoint subprotocol for this checkpoint.
+    ///
+    /// `tip.l1_height` is the L1 view consumed by the OL for this epoch — distinct
+    /// from `l1_reference.l1_commitment`, which records where the checkpoint
+    /// envelope was observed on L1.
+    #[borsh(
+        serialize_with = "borsh_serialize_checkpoint_tip",
+        deserialize_with = "borsh_deserialize_checkpoint_tip"
+    )]
+    pub tip: CheckpointTip,
 
-    /// L1 reference for this checkpoint.
+    /// L1 reference for the envelope that carried this checkpoint.
     pub l1_reference: CheckpointL1Ref,
 }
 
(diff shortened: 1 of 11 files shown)


prajwolrg:
@codex review.

crates/csm-types/src/client_state.rs
```diff
@@ -136,14 +141,20 @@ impl CheckpointL1Ref {
...
+    /// Tip published by the ASM checkpoint subprotocol for this checkpoint.
+    ///
+    /// `tip.l1_height` is the L1 view consumed by the OL for this epoch — distinct
+    /// from `l1_reference.l1_commitment`, which records where the checkpoint
+    /// envelope was observed on L1.
+    #[borsh(
+        serialize_with = "borsh_serialize_checkpoint_tip",
+        deserialize_with = "borsh_deserialize_checkpoint_tip"
+    )]
+    pub tip: CheckpointTip,
```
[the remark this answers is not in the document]
prajwolrg: This isn't relevant as this is a breaking change. CC: @storopoli this came up again.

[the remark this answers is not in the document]
storopoli: Codex loves these kind of things. For now "we move fast and break stuff"... Ignore it.

crates/csm-worker/src/state.rs
```diff
@@ -79,14 +79,14 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
...
-            .map(|(epoch, _)| *epoch)
+            .map(EpochCommitment::from)
             .or(finalized_epoch);
 
         // Keep only non-finalized candidates for incremental tip-driven advancement.
         if let Some(finalized) = finalized_epoch {
             while observed_checkpoints
                 .front()
-                .is_some_and(|(epoch, _)| epoch.epoch() <= finalized.epoch())
+                .is_some_and(|checkpoint| checkpoint.tip.epoch <= finalized.epoch())
```
storopoli, line 120: Refresh ClientState when bootstrap advances finality

`CsmWorkerState::new` can derive a newer finalized epoch from stored L1 observations 

https://github.com/alpenlabs/alpen/blob/ea8ca76a1de9aac7dea78094482e287defac687b/crates/csm-worker/src/state.rs#L73-L76

and then prune the corresponding observation from `observed_checkpoints` 

https://github.com/alpenlabs/alpen/blob/ea8ca76a1de9aac7dea78094482e287defac687b/crates/csm-worker/src/state.rs#L85-L92

but `last_committed_state` is still the stale `ClientState` loaded from storage 

https://github.com/alpenlabs/alpen/blob/ea8ca76a1de9aac7dea78094482e287defac687b/crates/csm-worker/src/state.rs#L95-L99

If the node restarts after the observation was committed but before `refresh_finalized_checkpoint` persisted the refreshed `ClientState` (or after that write failed), bootstrap sets the internal `finalized_epoch` while consumers that read `ClientState::get_declared_final_epoch()` can remain behind 

https://github.com/alpenlabs/alpen/blob/ea8ca76a1de9aac7dea78094482e287defac687b/crates/chain-worker/src/service.rs#L134-L140

Because the finalized observation has already been removed from the queue, the service will not get another chance to write the missing finalized checkpoint until a later epoch finalizes. Please either make bootstrap persist/publish the refreshed `ClientState` for the derived finalized checkpoint, or keep the candidate queued until that refresh is durable.

storopoli (changes requested):

crates/csm-types/src/client_state.rs
```diff
@@ -136,14 +141,20 @@ impl CheckpointL1Ref {
     }
 }
 
-#[derive(
-    Clone, Debug, Eq, PartialEq, Arbitrary, BorshDeserialize, BorshSerialize, Deserialize, Serialize,
-)]
+#[derive(Clone, Debug, Eq, PartialEq, BorshDeserialize, BorshSerialize, Deserialize, Serialize)]
```
delbonis, line 141: Can we un-borsh this?

crates/csm-types/src/client_state.rs
```diff
@@ -154,10 +165,67 @@ impl fmt::Display for L1Checkpoint {
...
+fn borsh_deserialize_checkpoint_tip<R: Read>(reader: &mut R) -> BorshIoResult<CheckpointTip> {
+    let epoch = BorshDeserialize::deserialize_reader(reader)?;
+    let l1_height = BorshDeserialize::deserialize_reader(reader)?;
+    let l2_commitment = BorshDeserialize::deserialize_reader(reader)?;
+    Ok(CheckpointTip {
+        epoch,
+        l1_height,
+        l2_commitment,
+    })
+}
```
delbonis: Like this could just be SSZed, maybe wrap the field with `CodecSsz`.  Another reason to avoid `pub` fields to hide these wrappers.

crates/csm-worker/src/processor.rs
```diff
@@ -162,8 +160,10 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
 
     /// Folds checkpoint observations made during a successfully committed block
     /// into the worker's finality cursors.
-    fn apply_observations(&mut self, observations: Vec<(EpochCommitment, CheckpointL1Ref)>) {
-        for (commitment, observation) in observations {
+    fn apply_observations(&mut self, observations: Vec<L1Checkpoint>) {
```
delbonis: Would be better as `impl IntoIterator<Item = L1Checkpoint>`, maybe ref the checkpoint.

crates/csm-worker/src/processor.rs
```diff
@@ -179,23 +179,30 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
                 && !self
                     .observed_checkpoints
                     .iter()
-                    .any(|(epoch, _)| *epoch == commitment)
+                    .any(|existing| EpochCommitment::from(existing) == commitment)
             {
-                self.observed_checkpoints
-                    .push_back((commitment, observation));
+                self.observed_checkpoints.push_back(checkpoint);
             }
```
delbonis: Can break this body out as a `apply_checkpoint_observation` fn, so that we don't always need to call it as a part of a vec.

delbonis (changes requested):
Some cleanups to improve organization.

crates/csm-worker/src/service.rs
```diff
@@ -59,12 +61,22 @@ impl<C: CsmWorkerContext + 'static> SyncService for CsmWorkerService<C> {
             error!(%asm_block, err = ?e, "Failed to process ASM block");
         }
 
-        let finalized_changed = state.advance_finalization(asm_block.height());
+        let newly_finalized = state.advance_finalization(asm_block.height());
         let confirmed_changed = state.confirmed_epoch != prev_confirmed_epoch;
```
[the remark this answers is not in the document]
prajwolrg: Addressed in 827f1cd8. `process_input` now bails out before calling `advance_finalization` / `refresh_finalized_checkpoint` when `process_asm_block` returns `Err`, so no `ClientState` row is persisted at the failed block and bootstrap on restart still resumes from the last contiguous commit. Added `process_input_skips_finalization_when_process_asm_block_fails` to pin the behaviour.

crates/csm-worker/src/state.rs
prajwolrg: Addressed in 827f1cd8. `CsmWorkerState::new` now detects when the derived `finalized_epoch` outpaces the on-disk `ClientState`, rebuilds the state from the corresponding `L1Checkpoint`, and persists + publishes it before pruning the queue — so on restart `ClientState::get_declared_final_epoch()` no longer trails the worker. Extended the existing bootstrap test to verify the refreshed row lands on disk and added `test_state_new_does_not_refresh_when_baseline_matches` to guard the no-op case.

crates/csm-worker/src/processor.rs
prajwolrg: Done in 3aebea06. Kept the item as owned `L1Checkpoint` since `apply_checkpoint_observation` moves it into the queue — a borrowed variant would force a clone at the callsite for the same effect.

crates/csm-worker/src/processor.rs
prajwolrg: Done in 3aebea06 — extracted `apply_checkpoint_observation`; `apply_observations` is now a thin loop over it.

crates/csm-types/src/client_state.rs
prajwolrg: Update: landed a partial improvement in 8db14929 — `CheckpointTip` actually does have a Borsh impl via `impl_borsh_via_ssz_fixed!` in the pinned asm version, so the field-level `serialize_with`/`deserialize_with` shim and its standalone codec fns were dead weight. Dropped them; `L1Checkpoint` now uses a plain Borsh derive (no field attrs, no orphan-rule comment).

Full un-borsh of the chain (`ClientUpdateOutput` → `ClientState` → `L1Checkpoint`) still deferred — sled `ClientUpdateOutputSchema` uses `impl_borsh_value_codec!`, so going SSZ-native means switching that to `impl_codec_value_codec!` and `#[derive(Codec)]` on five types in csm-types. Worth its own PR.

crates/csm-types/src/client_state.rs
prajwolrg: Partial update in 8db14929: removed the standalone codec fns and field attrs entirely (turned out to be dead weight — `CheckpointTip` already has a Borsh impl via `impl_borsh_via_ssz_fixed!`).

crates/csm-worker/src/state.rs
storopoli: Thanks solving this

storopoli (approved):
ACK 8db149297ed6e228a80c9c453d26bf1ff2023625

crates/csm-types/src/client_state.rs
delbonis: Should make a ticket for this, but can defer from this PR.

crates/csm-types/src/client_state.rs
```diff
@@ -154,10 +162,43 @@ impl fmt::Display for L1Checkpoint {
...
+    pub fn new(tip: CheckpointTip, l1_reference: CheckpointL1Ref) -> Self {
+        Self { tip, l1_reference }
+    }
+}
+
+impl From<&L1Checkpoint> for EpochCommitment {
+    fn from(checkpoint: &L1Checkpoint) -> Self {
+        EpochCommitment::from_terminal(checkpoint.tip.epoch, checkpoint.tip.l2_commitment)
+    }
+}
```
delbonis, line 174: I would have honestly just done a `.get_epoch_commitment()` or something instead of impl `From` for this since it's not exactly an "obvious" conversion, but that's not a big deal.

delbonis (approved):
Some followups but looks good, approving.

crates/csm-worker/src/service.rs
```diff
@@ -53,18 +55,35 @@ impl<C: CsmWorkerContext + 'static> SyncService for CsmWorkerService<C> {
...
-        // Errors here are intentionally swallowed: gap-fill is idempotent and
-        // the next ASM status update will retry the missed blocks.
+        // On error we bail out before touching finality: `advance_finalization`
+        // and `refresh_finalized_checkpoint` would otherwise persist a
+        // `ClientState` row keyed on the failed block, and bootstrap
+        // (`fetch_most_recent_client_state`) would treat that as committed on
+        // restart — silently dropping the retry the gap-fill path is meant to
+        // guarantee. Per-block intermediate progress from successful gap-fill
+        // commits is already persisted and published by `commit_block`, so
+        // skipping the trailing finality work here doesn't lose it.
```
bewakes: nit: this comment seems a bit too verbose. Consider compacting/removing it.

crates/csm-worker/src/state.rs
```diff
@@ -132,37 +165,42 @@ fn load_observed_checkpoints<C: CsmWorkerContext>(
...
 }
 
-/// Returns the latest epoch commitment whose observation meets the depth threshold.
+/// Returns the latest observation whose L1 ref meets the depth threshold.
 ///
-/// Iterates forward; the last match wins (latest finalized).
-fn derive_finalized_epoch<'a, I>(
+/// Iterates forward; the last match wins (latest finalized). Returns the
+/// borrowed checkpoint so callers can both derive `EpochCommitment::from(&_)`
+/// and reconstruct the durable `ClientState` finality field from it.
```
bewakes: nit: redundant comment, describes what callers can do.

bewakes (approved):
This is great! Some nits on the "verbosy" comments. Nothing blocking.

storopoli (approved):
ACK 8e2b5a1bc32a96cc444af6f7402ddf35d8238128

PR #1842, opened 2026-05-20
STR-3414: admin API for prover tasks via strata-dbtool

prajwolrg:
## Description

Delivers an admin API for all three prover task stores in the Alpen stack via `strata-dbtool` (per [STR-3414](https://alpenlabs.atlassian.net/browse/STR-3414)): the OL integrated checkpoint prover, the EE chunk prover, and the EE acct/batch prover. None of these had an out-of-band path before — operators couldn't abandon stuck in-progress tasks across a restart, force a fresh re-prove, drop stale receipts after a guest-program upgrade, or queue a proof request "from outside." This PR adds offline admin verbs covering all of those for every store.

### Surface coverage

| Concern              | Lives in              | Subcommand prefix |
|----------------------|-----------------------|--------------------|
| OL checkpoint proofs | strata node datadir   | (no prefix)        |
| EE chunk proofs      | alpen-client datadir  | `ee-*` (`--kind chunk`) |
| EE acct/batch proofs | alpen-client datadir  | `ee-*` (`--kind acct`)  |

The two EE stores share one sled tree, kind-tagged by the first byte of every key (`b'c'` / `b'a'`), so `ee-*` task commands operate on opaque keys with a `--kind` filter on inspection and bulk-abandon. Receipt admin is kind-specific because the receipts live in distinct trees: chunk receipts keyed by task key, acct proofs typed by `BatchId` (`<prev_block_hex>:<last_block_hex>`, matching the `Display` format).

EE commands need `--ee-datadir` (the alpen-client's `--datadir`). It's opened lazily, so OL-only invocations keep working without the flag set.

### Verb semantics

| Verb       | Final status                                    | When to use                                               |
|------------|-------------------------------------------------|-----------------------------------------------------------|
| `abandon`  | `PermanentFailure { error: "abandoned via dbtool" }` | Stop the recovery scanner from respawning a stuck task while keeping an audit trail. |
| `reset`    | `Pending` (retry-after cleared)                 | Force a fresh prove attempt — drops accumulated retry count. |
| `delete`   | row removed                                     | Prefer `abandon` unless you really want no trace left.    |
| `backfill` | `Pending` (newly inserted)                      | Queue a proof request "from outside" — e.g. for an epoch the node never picked up. |
| `delete-*-receipt` / `delete-acct-proof` | row removed (acct also clears `ProofId → BatchId` index) | Drop a stale receipt after a guest-program upgrade so the next run re-proves it. |

Every mutating subcommand is a **dry run** unless `-f, --force` is passed — matching the existing `revert-ol-state` UX. Without `--force`, the command prints what *would* happen and a `Use --force to execute these changes.` hint; with `--force`, the mutation lands. See [bin/strata-dbtool/README.md](bin/strata-dbtool/README.md) for the full command list and usage.

### Type of Change

- [x] Refactor (extracted `CheckpointProofTask` and shared prover-task admin helpers to shared modules; added `delete_chunk_receipt` / `delete_acct_proof` to `EeProverDbSled`)
- [x] New or updated tests

## Checklist

## AI Assistance Disclosure

This PR was written by Claude Code (Opus 4.7), with me reviewing and steering the design.

## Related Issues

[STR-3414](https://alpenlabs.atlassian.net/browse/STR-3414)

[STR-3414]: https://alpenlabs.atlassian.net/browse/STR-3414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/strata-dbtool/README.md
+++ b/bin/strata-dbtool/README.md
@@ -32,6 +32,7 @@ strata-dbtool [OPTIONS] <COMMAND>
 ### Global Options
 
 - `-d, --datadir <path>` - Node data directory (default: `data`)
+- `--ee-datadir <path>` - Alpen-client data directory. Required for any `ee-*` subcommand; points at the alpen-client's `--datadir`, not the strata node's.
 
 ## Commands
 
@@ -440,11 +441,109 @@ upgrade.
 strata-dbtool delete-checkpoint-proof <epoch> --confirm
 ```
 
-### Not yet supported
+## EE Prover Task & Receipt Admin
 
-The EE prover task store (`alpen-client` datadir — chunk receipts, acct
-proofs, shared task tree) is not yet wired into this tool. Add a follow-up
-ticket if operator workflows need it.
+> [!WARNING]
+>
+> These commands mutate the EE prover store under the **alpen-client**
+> datadir (not the strata node's). Stop the alpen-client before using
+> them — concurrent writes from a running chunk/acct prover will conflict
+> with these edits and may corrupt state.
+
+The alpen-client maintains a separate sled instance for prover-side
+persistence — shared task tree (chunk + acct), chunk-receipt store, and
+typed acct-proof store. All `ee-*` subcommands require `--ee-datadir`,
+which points at the alpen-client's `--datadir`.
+
+### Which surface to use
+
+| Concern              | Lives in              | Subcommand prefix |
+|----------------------|-----------------------|--------------------|
+| OL checkpoint proofs | strata node datadir   | (no prefix)        |
+| EE chunk proofs      | alpen-client datadir  | `ee-*` (`--kind chunk`) |
+| EE acct/batch proofs | alpen-client datadir  | `ee-*` (`--kind acct`)  |
+
+Chunk and acct tasks share one tree, disambiguated by a single-byte
+**kind tag** at the start of every task key (`b'c'` for chunk, `b'a'`
+for acct). The `--kind` filter on the summary and bulk-abandon commands
+selects on that tag; single-key commands operate on opaque keys, so the
+kind comes from whatever the key starts with.
+
+### `ee-get-prover-task`
+Fetch a single EE prover task record by its hex-encoded key.
+
+```bash
+strata-dbtool --ee-datadir <path> ee-get-prover-task <key_hex> [OPTIONS]
+```
+
+### `ee-get-prover-tasks-summary`
+Aggregate counts by status, plus a bounded slice of matching entries.
+
+```bash
+strata-dbtool --ee-datadir <path> ee-get-prover-tasks-summary [--status <filter>] [--kind <kind>] [--limit <n>] [OPTIONS]
+```
+
+**Options:**
+- `--status <filter>` — same set as the OL summary command (`all`, `pending`, …, `terminal`).
+- `--kind <kind>` — one of `all` (default), `chunk`, `acct`.
+
+### `ee-abandon-prover-task`
+Mark a single EE task as `PermanentFailure { error: "abandoned via dbtool" }`.
+
+```bash
+strata-dbtool --ee-datadir <path> ee-abandon-prover-task <key_hex> --confirm
+```
+
+### `ee-abandon-prover-tasks`
+Bulk-abandon every `Pending`/`Proving` EE task, optionally restricted by kind.
+
+```bash
+strata-dbtool --ee-datadir <path> ee-abandon-prover-tasks --all-unfinished [--kind <kind>] --confirm [--dry-run]
+```
+
+### `ee-reset-prover-task`
+Flip an EE task back to `Pending` and clear its retry-after timestamp.
+
+```bash
+strata-dbtool --ee-datadir <path> ee-reset-prover-task <key_hex> --confirm
+```
+
+### `ee-delete-prover-task`
+Hard-delete an EE task record.
+
+```bash
+strata-dbtool --ee-datadir <path> ee-delete-prover-task <key_hex> --confirm
+```
+
+### `ee-backfill-prover-task-raw`
+Insert a `Pending` EE task record under a caller-provided raw key. EE
+task keys come from the chunk/acct spec encodings; raw is the only
+supported backfill path (no typed equivalent of `backfill-checkpoint-proof-task`).
+
+```bash
+strata-dbtool --ee-datadir <path> ee-backfill-prover-task-raw <key_hex> --confirm
+```
+
+### `ee-get-chunk-receipt` / `ee-delete-chunk-receipt`
+Inspect or remove a stored chunk-proof receipt by its task key. Use
+case: drop a stale receipt after a guest-program upgrade so the chunk
+prover re-proves it.
+
+```bash
+strata-dbtool --ee-datadir <path> ee-get-chunk-receipt <key_hex> [OPTIONS]
+strata-dbtool --ee-datadir <path> ee-delete-chunk-receipt <key_hex> --confirm
+```
+
+### `ee-get-acct-proof` / `ee-delete-acct-proof`
+Inspect or remove a stored acct/batch proof. The batch id is passed as
+`<prev_block_hex>:<last_block_hex>` (each 32 bytes), matching `BatchId`'s
+`Display` format — copy directly from the alpen-client's logs. Delete
+also clears the secondary `ProofId → BatchId` index.
+
+```bash
+strata-dbtool --ee-datadir <path> ee-get-acct-proof <prev_block>:<last_block> [OPTIONS]
+strata-dbtool --ee-datadir <path> ee-delete-acct-proof <prev_block>:<last_block> --confirm
+```
 
 ## Output Formats
 
--- a/bin/strata-dbtool/README.md
+++ b/bin/strata-dbtool/README.md
@@ -349,8 +349,10 @@ strata-dbtool revert-ol-state -f -d 858c390aaaabd7c457cb24c955d06fb9de0f6666d0b6
 > store. Stop the node before using them — concurrent writes from a running
 > prover will conflict with these edits and may corrupt state.
 
-Each mutating subcommand requires an explicit `--confirm` flag. Bulk variants
-also accept `--dry-run` to preview the change set without writing.
+Every mutating subcommand is a **dry run** unless `-f, --force` is passed —
+the same UX as `revert-ol-state`. Without `--force`, the command prints
+what *would* happen and a `Use --force to execute these changes.` hint;
+with `--force`, the mutation actually lands.
 
 ### Semantics — `abandon` vs `reset` vs `delete`
 
@@ -383,28 +385,29 @@ strata-dbtool get-prover-tasks-summary [--status <filter>] [--limit <n>] [OPTION
 Mark a single task as `PermanentFailure { error: "abandoned via dbtool" }`.
 
 ```bash
-strata-dbtool abandon-prover-task <key_hex> --confirm
+strata-dbtool abandon-prover-task <key_hex> --force
 ```
 
 ### `abandon-prover-tasks`
-Bulk-abandon every `Pending` or `Proving` task.
+Bulk-abandon every `Pending` or `Proving` task. Without `--force`, prints
+the change set as a dry run.
 
 ```bash
-strata-dbtool abandon-prover-tasks --all-unfinished --confirm [--dry-run]
+strata-dbtool abandon-prover-tasks --all-unfinished --force
 ```
 
 ### `reset-prover-task`
 Flip a task back to `Pending` and clear its retry-after timestamp.
 
 ```bash
-strata-dbtool reset-prover-task <key_hex> --confirm
+strata-dbtool reset-prover-task <key_hex> --force
 ```
 
 ### `delete-prover-task`
 Hard-delete a task row.
 
 ```bash
-strata-dbtool delete-prover-task <key_hex> --confirm
+strata-dbtool delete-prover-task <key_hex> --force
 ```
 
 ### `backfill-checkpoint-proof-task`
@@ -414,7 +417,7 @@ canonical commitment at the epoch and constructs the task key via the shared
 its next startup-recovery pass.
 
 ```bash
-strata-dbtool backfill-checkpoint-proof-task <epoch> --confirm
+strata-dbtool backfill-checkpoint-proof-task <epoch> --force
 ```
 
 ### `backfill-prover-task-raw`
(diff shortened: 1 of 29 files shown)


krsnapaudel:
@prajwolrg `strata-dbtool` has this option for `revert-ol-state`. 
> -f, --force - Force execution (without this flag, only a dry run is performed)

Should we have similar instead of `--confirm` flag? For consistency.

prajwolrg:
> @prajwolrg `strata-dbtool` has this option for `revert-ol-state`.
> 
> > -f, --force - Force execution (without this flag, only a dry run is performed)
> 
> Should we have similar instead of `--confirm` flag? For consistency.

We should definitely make things consistent. Handled in [3f29617](https://github.com/alpenlabs/alpen/pull/1842/commits/3f29617358e5c1fb489c12061de74e72c2f63224).

bin/strata-dbtool/README.md
```diff
@@ -32,6 +32,7 @@ strata-dbtool [OPTIONS] <COMMAND>
 ### Global Options
 
 - `-d, --datadir <path>` - Node data directory (default: `data`)
+- `--ee-datadir <path>` - Alpen-client data directory. Required for any `ee-*` subcommand; points at the alpen-client's `--datadir`, not the strata node's.
```
evgenyzdanovich: that's becoming messy

evgenyzdanovich:
the dbtool becomes much messier and would need its own refactor imo.

for the time being this seems ~alrightish, but ideally we'd want to have it (1) more structured (2) more _declarative_ and more friendly with storage (3) with much less boilerplate (so when we add a new proof type, it's automatically/almost automatically brought into dbtool with minimal efforts and little-to-no ctrl-c-ctrl-v).

P.S. in ideal world, our cli tools should resemble well known patterns for cli tools.
For example, when I interact with google cloud cli, it is very intuitive (because it is "layered" properly). it looks like this:
```
gcloud <api type> <resource type> <command type> <subcommand type>
```

examples:
```
gcloud compute vms <list|get|any other verbs>
gcloud compute disks <list|get|any other verb>
gcloud networking interfaces <list|get|any other verb>
```


Similarly, we can define
```
dbtool ol <ol resouce> <verb>
dbtool ee <resource> <verb>
dbtool proofs ...
etc
```
the "resource" can be checkpoint (for OL) / proof (for both) / state parts (specific to any) / etc.  

-------

In contract, rn we are flat ass with resources, commands, etc.

evgenyzdanovich:
overall, pre-ACK on the effort (it does move in a much needed direction).

prajwolrg:
Marking this for review. I'd defer the refactor out of this PR into a separate ticket instead, but the API you suggested look much better. CC: @krsnapaudel who I believe originally worked on the dbtool first.

bin/strata-dbtool/src/main.rs
```diff
@@ -38,6 +54,13 @@ fn main() {
     });
     let db = db.as_ref();
```
storopoli: This still opens the OL database before command dispatch, so every `ee-*` invocation attempts to open/lock `cli.datadir` even though those commands only need `--ee-datadir`. That breaks the documented EE-only operational path when the strata node is running, or when the default `-d data` path is missing/inaccessible, because execution exits before reaching `with_ee_db`.

Permalink: https://github.com/alpenlabs/alpen/blob/1756e76e076410bc47ccc1e731347790ff6162af/bin/strata-dbtool/src/main.rs#L51-L55

I think the fix is to make the OL DB lazy too: dispatch `ee-*` arms before calling `open_database`, or wrap OL-only arms in a helper that opens the OL DB only for commands that actually need it.

bin/strata-dbtool/src/cmd/ee_receipts.rs
```diff
@@ -0,0 +1,281 @@
...
+/// Parses `"<prev_block_hex>:<last_block_hex>"` into a [`BatchId`].
+///
+/// Each hex half is exactly 32 bytes (64 hex chars). Both parts are
+/// required — there is no fallback to a single-hash form because the
+/// underlying type literally is a pair of hashes.
+pub(crate) fn parse_batch_id(s: &str) -> Result<BatchId, DisplayedError> {
+    let (prev, last) = s.split_once(':').ok_or_else(|| {
+        DisplayedError::UserError(
+            "Expected batch id as <prev_block_hex>:<last_block_hex>".to_string(),
+            Box::new(s.to_string()),
```
storopoli, line 103: The acct-proof parser only accepts `<prev_block_hex>:<last_block_hex>`, but the codebase logs/display path for this value is `BatchId(prev_block=..., last_block=...)`. In particular, `BatchId::Display` emits that wrapper form and the acct proof hook logs it with `%batch_id`, so operators cannot copy the value they see in logs into `ee-get-acct-proof` / `ee-delete-acct-proof` as the docs/comments imply.

Permalinks:
- Parser: https://github.com/alpenlabs/alpen/blob/1756e76e076410bc47ccc1e731347790ff6162af/bin/strata-dbtool/src/cmd/ee_receipts.rs#L93-L108
- Display format: https://github.com/alpenlabs/alpen/blob/1756e76e076410bc47ccc1e731347790ff6162af/crates/alpen-ee/common/src/types/batch.rs#L17-L24
- Logged with `%batch_id`: https://github.com/alpenlabs/alpen/blob/1756e76e076410bc47ccc1e731347790ff6162af/bin/alpen-client/src/prover/hooks.rs#L84-L90

Please either parse the displayed `BatchId(...)` form too, or change the user-facing guidance/output so it consistently surfaces the accepted colon form.

bin/strata-dbtool/src/cmd/ee_prover_task.rs
delbonis, line 1: Comments before attributes!

bin/strata-dbtool/src/cmd/ee_receipts.rs
delbonis, line 1: Here too.

bin/strata-dbtool/src/cmd/prover_task.rs
delbonis, line 1: Here too.

delbonis (changes requested):
Implementation-wise this looks good, aside from the other comments.

But I'm wondering if it actually makes sense to put the EE-specific dbtool subcommands in a separate binary so that we aren't merging unrelated areas of concern.  Like doesn't this now depend on both Strata and Alpen database concepts now?

bin/strata-dbtool/README.md
krsnapaudel: Yes why can't we specify a different path with `--datadir` to run the command? Each command is standalone.

bewakes:
Functionally this looks good enough to me - I looked into the subcommands and they seem to support the operations that are needed.

I agree with comments on structuring the interface and making it less boilerplated.

Thinking out loud, not a concrete direction: one thing worth exploring as a separate task is a database <-> commandline abstraction layer which would primarily support the basic CRUD operations which account for majority of the dbtool access right now - most of the subcommands are plain read/writes of db tables. The open question is whether our database trees are uniform enough to support this. Supporting custom operations besides basic CRUD would be a cherry-on-top.

An example that comes to mind is how Django the web framework makes it reallyy easy to define CRUD interface for admin operations(declarative model registration -> free CRUD UI), at the same time being extremely customizable and flexible.

evgenyzdanovich:
agree with @bewakes on the idea! 

perhaps we can have a thin decl layer (through macro or otherwise) that's based off storage schema that delivers basic read/write logic (crud style). 
this should unify quite a lot of boilerplate across both ee&ol and allow for easier support in the future

bin/strata-dbtool/README.md
prajwolrg: Dropped `--ee-datadir` entirely in 7bde2159a — `-d` now covers both OL and EE invocations, and `main` lazily opens just the sled the chosen command needs (so `ee-*` invocations no longer try to also lock the OL sled). `strata-dbtool -d <alpen-client-datadir> ee-get-chunk-receipt ...` is the new path.

bin/strata-dbtool/src/main.rs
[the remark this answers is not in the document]
prajwolrg: Fixed in 7bde2159a — folded into the `--ee-datadir` removal. `main` now splits dispatch into `with_ol_db`/`with_ee_db` and only opens the sled the chosen command actually needs, so `ee-*` invocations no longer touch the OL sled at all.

prajwolrg: Fixed in 7bde2159a — see the reply on the sibling Codex comment. `main` now splits dispatch into `with_ol_db`/`with_ee_db` and opens exactly one sled per command (OL or EE), so the EE admin path no longer collides with a running strata node or a missing default `data/` directory. `--ee-datadir` is gone; `-d` is the single flag and operators point it at the alpen-client's `--datadir` for `ee-*` commands.

bin/strata-dbtool/src/cmd/ee_receipts.rs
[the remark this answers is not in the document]
prajwolrg: Fixed in f505c59de. Changed `BatchId::Display` to emit `<prev_hex>:<last_hex>` using `LowerHex` for full-form hashes — note that the previous wrapper form was lossy on top of being structurally different, since `Buf32::Display` truncates to a `prefix..suffix` form, so even parsing the wrapper wouldn't have round-tripped. The parser already accepts the `<prev>:<last>` shape, so `%batch_id` log values now copy-paste end-to-end. Other batch_id call sites use `?batch_id` (Debug) and are unaffected.

prajwolrg: Fixed in f505c59de — see the sibling Codex reply. tl;dr: switched `BatchId::Display` to `<prev_hex>:<last_hex>` (full hashes via `LowerHex`, since `Buf32::Display` was truncating to `prefix..suffix` and would have made the wrapper form unusable for copy-paste even if the parser accepted it). The acct hook's `%batch_id` log line now emits the exact shape the dbtool parser expects.

bin/strata-dbtool/src/cmd/ee_prover_task.rs
prajwolrg: Reordered in e32970ab5 — `///` now comes before `#[derive]`/`#[argh]` on every cmd struct in this file.

bin/strata-dbtool/src/cmd/ee_receipts.rs
prajwolrg: Same fix in e32970ab5 — reordered every cmd struct in this file too.

bin/strata-dbtool/src/cmd/prover_task.rs
prajwolrg: And here too in e32970ab5.

prajwolrg:
Pushed three commits addressing the open feedback:

@delbonis — on the question of whether the EE subcommands should live in a separate binary:

I think the right answer is "one binary, layered grammar" rather than "two binaries", but the architectural decision is exactly what the structural refactor needs to settle. I've filed **[STR-3564](https://alpenlabs.atlassian.net/browse/STR-3564) — Explore declarative/layered CLI structure for strata-dbtool** as a spike to design this end-to-end. The ticket captures both directions raised in this PR:

- @evgenyzdanovich's `dbtool <api> <resource> <verb>` grammar (gcloud/kubectl-style), which subsumes the "separate binary?" question — `dbtool ol …` vs `dbtool ee …` becomes a naming choice rather than a packaging one.
- @bewakes's declarative DB↔CLI abstraction (Django-admin-style resource registration → free CRUD with custom-verb opt-in), with the "are our trees uniform enough?" open question called out explicitly.

The spike will pick one combined direction, prototype against a single resource (probably checkpoints or prover tasks) to validate the boilerplate reduction is real, and lay out a phased migration so we don't have to do a single giant cutover. Happy to add either of you as watchers on the ticket.

[STR-3564]: https://alpenlabs.atlassian.net/browse/STR-3564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

bewakes:
[AI ASSISTED REVIEW]
A scan for security issues by claude surfaced the following. Please fix this if relevant:

`bin/strata-dbtool/src/cmd/prover_task.rs:288-319 · ee_prover_task.rs:349-378`

`reset-prover-task` has no terminal-state guard. `abandon_prover_task` explicitly refuses to act on a row in `is_terminal()` state (line 217 / 280); `reset_prover_task` blindly sets status to Pending and clears `retry_after_secs`, regardless of current state.

**Failure mode**

Operator runs `reset-prover-task <key> --force` against a Completed task (typo, stale dashboard, copy-paste from the wrong log line). Row flips to `Pending`. On node restart, the recovery scanner respawns the proof. The existing checkpoint receipt is still in `CheckpointProofSchema`, and OL's `put_proof` uses `EntryAlreadyExists` semantics `(crates/db/store-sled/src/prover/db.rs:22)` — so the re-prove succeeds in compute but fails on write, leaving the task in a stuck retry loop. For EE acct proofs, the receipt overwrites cleanly but the secondary `ProofId → BatchId` index transiently disagrees during the re-prove window.

**Fix**

Mirror the `is_terminal()` guard from `abandon_prover_task`. If resetting a Completed task is ever legitimate (it probably isn't — the typed flow would be `delete-checkpoint-proof` then `backfill-checkpoint-proof-task`), gate it behind an explicit `--allow-terminal` flag rather than silently allowing it.

prajwolrg:
Thanks for the careful trace, @bewakes — but after walking the code paths the failure mode is narrower than "stuck retry loop." `ProverError::Storage` isn't classified as transient ([`error.rs:41-43`](https://github.com/alpenlabs/alpen/blob/STR-3414-prover-dbtool/crates/prover-core/src/error.rs#L41-L43)), so `handle_error` takes the `else` branch and marks the task `PermanentFailure` on the first failed `put_proof` — one wasted re-prove plus a misleading status field, not an infinite loop. The receipt on disk stays valid and downstream consumers reading `CheckpointProofSchema` are fine.

Given that, guarding `reset-prover-task` in dbtool would be the wrong layer — the real asymmetry is that OL's `put_proof` refused to overwrite while EE's `put_acct_proof` already upserts ([`prover_db.rs:85-98`](https://github.com/alpenlabs/alpen/blob/STR-3414-prover-dbtool/crates/alpen-ee/database/src/sleddb/prover_db.rs#L85-L98) with the explicit "Upsert is fine: idempotent re-submits from paas replace the receipt" comment). Re-proves for the same epoch attest to the same statement, so overwriting is safe.

Fixed at the source in a1e9f99: OL `put_proof` is now an upsert, matching the EE side. Reset on a Completed task now re-proves and ends Completed normally — no PermanentFailure, no special-casing needed in dbtool.

storopoli (approved):
ACK a1e9f99fa11b84ac67ba0ee401ffecfc123dbcc2

bewakes (approved):
ACK https://github.com/alpenlabs/alpen/pull/1842/commits/a1e9f99fa11b84ac67ba0ee401ffecfc123dbcc2

delbonis (approved):

voidash (approved):

purusang (approved):

PR #1843, opened 2026-05-20
[DO NOT MERGE] feat(reth): enable Osaka and `OP_CLZ`

storopoli:
## Description

Enables the Osaka spec at genesis for the bundled Alpen chain specs so `OP_CLZ` is available in dev, devnet, and testnet configurations.

This also updates the bundled block gas limit to 36M (according to PRDs).

Adds an EE/OL functional test that deploys raw `OP_CLZ` bytecode, executes it, verifies the stored result, and waits for the post-execution snark update submission to OL. This covers host execution and proof-path re-execution consistency.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The Reth version is unchanged. Reth `v1.9.1` already supports `osakaTime` parsing and exposes the Osaka transaction cap constant used here.

Since we've changed the block gas limit the genesis block hash in functional tests had to be updated.

AI assistance was used to help implement and validate this change.

- [x]  Yes
- [ ]  No

- https://eips.ethereum.org/EIPS/eip-7939
- https://eips.ethereum.org/EIPS/eip-7935

## Checklist

## Related Issues

STR-2823


barakshani:
Nice!

Let's please put in draft until:
1. formal approval from product
2. more importantly, staging is stable. we don't want more bullet whistles right now.

barakshani:
>This also updates the bundled block gas limit to 60M

This should be discussed separately with product.

evgenyzdanovich (commented):
**NACK on the effort to be merged in any state up until EF tests and proofs are stabilized here:**

https://github.com/alpenlabs/alpen/pull/1836

evgenyzdanovich:
such changes do deserve:
1. its own task (not random `feat` that will be accidentally merged)
2. + clearly stating as [do not merge]
3. + (even before that) core team wide discussion/agreement
4. @barakshani to me, product approval is not enough here (well, they triggered it, any changes they won't approve?..). 

I'd actually demand engineers coming and looking at it in details (**in hopes that it will be postponed**) and realizing this is a ginormous change that must be carefully thought about first.

+CC @MdTeach @prajwolrg .

barakshani:
> such changes do deserve:
> 
> 1. its own task (not random `feat` that will be accidentally merged)
> 2. * clearly stating as [do not merge]
> 3. * (even before that) core team wide discussion/agreement
> 4. @barakshani to me, product approval is not enough here (well, they triggered it, any changes they won't approve?..).
> 
> I'd actually demand engineers coming and looking at it in details (**in hopes that it will be postponed**) and realizing this is a ginormous change that must be carefully thought about first.
> 
> +CC @MdTeach @prajwolrg .

@evgenyzdanovich maybe I am misunderstanding this change.

> that must be carefully thought about first.

Initially that's my reaction, as [reflected](https://alpen-labs.slack.com/archives/C05NCFNJ133/p1776006119915449?thread_ts=1775826001.556449&cid=C05NCFNJ133) in [discussions](https://alpen-labs.slack.com/archives/C05NCFNJ133/p1776094532115379) and on [Jira](https://alpenlabs.atlassian.net/browse/STR-2823). I read [Jose's message](https://alpen-labs.slack.com/archives/C05VA8PLCGM/p1779296016343569) as if that, in fact, we have already made all the required changes, also reflected in the minimal changes in this PR. Is this understanding wrong?

We have been burnt with the previous (Pectra) upgrade, so I'm definitely not in a hurry to merge this, but also will be good to understand your concerns, given the above.

evgenyzdanovich:
1. we never ever tested anything past Pectra. AND, Pectra was tested **on the old impl** of the EE.
2. new EE was never tested on neither of them (to the best of my knowledge).

so, even if we deploy our **new ee** **on Pectra** -- I personally can't guarantee it will work (and chain can always be proven).
To illustrate it -- https://alpenlabs.atlassian.net/browse/STR-3495 -- this would render our EE chain unprovable.
And trust me, we can have endless bugs of such types. 
So, bringing another eth upgrade into scope just exponentially increases the risk: with no EF tests, no analysis what's the diff between Osaka and Pectra (and what do we need to disable/tweak; for example, we did analysis and realized that Pectra Blob txns should be forbidden).

Moreover, looking deeper, at the **necessity** of this opcode -- it doesn't seems like it adds a lot of functional capabilities, but rather **optimizes the gas consumption** for bitwise operations in SC logic. 
So (if we don't deliver) -- best case they pay more gas, worst case they can rewrite their code to be compliant with Pectra (that does not use this opcode at all).

tldr: I dont see a clear reason of over-imposing ourselves with additional effort and unknown risk at this stage. The timing is way-off, and despite the change being small in size, it changes a lot.
I'm still **FOR** doing it at some point, especially to exercise upgradability path and see it working. But only if we ourselves decide it is indeed needed, not because somebody finds it as a nice-to-have.

evgenyzdanovich:
@barakshani
 
Concerns I have towards _this mindset and approach_ (of change being small and easy and already supported):
- have we thought of coverage for Osaka precompiles / opcodes in sp1 patches? Usually, `sp1` ships patches for specific proving logic (in case of EVM -- optimized patches for using precompiles / opcodes _in the guest code_ (so that prove would be efficient). I feel like it wasn't brought up or discussed at all +CC @prajwolrg @MdTeach.
 
- last time (when we used to be not in a hurry -- TN1 era), we spend a lot of time properly calculating the block limits, correlating it to prover capabilities, etc. In the end, it was company-side decision. What has changed in our approach of doing things?

- (from previous message) no clear analysis of Osaka upgrade (and a diff vs Pectra) and exactly what must be tuned/blacklisted/unsupported.

I'm by no means expert in EE; and other **engineers** may bring a lot more concerns to the table.

prajwolrg:
Agree with @evgenyzdanovich. Even though the code diff is small here, there are paths in the crates that we depend on that are never active without osaka enabled. Unless we are comfortable with those changes, we shouldn't rush into this. I think we need to have proper testing in place before migrating to this version.

> coverage for Osaka precompiles / opcodes in sp1 patches?

I don't think the sp1 patches relates to changes in EVM. sp1 patches are mostly for heavy cryptography. See [here](https://docs.succinct.xyz/docs/sp1/optimizing-programs/precompiles) for the full list. I don't think they have added any new sp1 patches after v3.

delbonis (approved):

MdTeach:
> Agree with @evgenyzdanovich. Even though the code diff is small here, there are paths in the crates that we depend on that are never active without osaka enabled. Unless we are comfortable with those changes, we shouldn't rush into this. I think we need to have proper testing in place before migrating to this version.
> 
> > coverage for Osaka precompiles / opcodes in sp1 patches?
> 
> I don't think the sp1 patches relates to changes in EVM. sp1 patches are mostly for heavy cryptography. See [here](https://docs.succinct.xyz/docs/sp1/optimizing-programs/precompiles) for the full list. I don't think they have added any new sp1 patches after v3.

[This](https://alpenlabs.atlassian.net/browse/STR-1505) should be the only remaning patch

PR #1844, opened 2026-05-21
feat(rpc): require auth for OL transaction submission

storopoli:
## Description

Moves `strata_submitTransaction` off the unauthenticated public Strata RPC listener and onto a dedicated authenticated submit RPC listener. The public OL RPC surface is now read-only, while the submit listener preserves the existing JSON-RPC method name and uses bearer-token auth.

This also adds submit RPC configuration, validation, alpen-client submit endpoint handling, and functional harness/docker updates so internal transaction submission uses the authenticated submit endpoint.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The public RPC listener no longer registers `strata_submitTransaction`. Sequencer deployments now require a non-empty submit RPC bearer token, and public/admin/submit RPC ports must all be distinct by port number.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-3548


bin/strata/src/rpc/auth.rs
delbonis, line 1: I *think* you could avoid having to roll this on your own, unless there's some annoying reason about how it's inside jsonrpsee: https://matze.github.io/axum-notes/notes/auth/with_tower_http/index.html

delbonis (approved):
One note.

krsnapaudel (approved):
LGTM. Thanks.

bin/strata/src/rpc/mod.rs
```diff
@@ -50,6 +50,9 @@ struct RpcDeps {
     admin_rpc_host: String,
     admin_rpc_port: u16,
     admin_rpc_bearer_token: Option<SecretString>,
+    submit_rpc_host: String,
```
bewakes, line 53: nit: might make sense to group these now.

bin/strata/src/rpc/mod.rs
```diff
@@ -301,6 +325,28 @@ async fn spawn_admin_rpc(deps: RpcDeps) -> Result<()> {
     Ok(())
 }
 
+/// Spawns the submit RPC server.
+async fn spawn_submit_rpc(deps: RpcDeps) -> Result<()> {
+    let module = build_submit_rpc_module(&deps)?;
+    let addr = format!("{}:{}", deps.submit_rpc_host, deps.submit_rpc_port);
```
bewakes, line 331: Since this is not admin related, why do we need a separate port to handle the submission endpoints? Shouldn't same port as non-admin read listeners be sufficient now that it has auth layer?

bewakes (commented):
This looks good generally. One config concern I see: a separate host/port per RPC category(admin, submit, general) feels redundant because there's more to configure, document and easier to misconfigure.

A port is worth it when we need network-level isolation. This is true for admin/debug related endpoints but not for read/submit neither of which are privileged. Isolating read/submit from each other doesn't buy much. I'd collapse these to two ports: admin/debug and non-admin(submit included) and make non-admin auth method-specific instead of port-specific.

The tradeoff here is that with method level auth we need to do it for every method we want to have auth whereas for port level, we just need to do it once.

Reminder to update the deployment docs after the PR lands.

bin/strata/src/rpc/mod.rs
storopoli: Why? Maybe you want to bind one to `0.0.0.0` and other to `127.0.0.1` etc.

bin/strata/src/rpc/mod.rs
storopoli: I'd like to keep this separated for the sake of the principle of separation of concerns. @alexhui01 opinions on having different ports for:

1. Public OL read RPC
2. Private auth-enabled EE RPC for submission of txs to OL mempool
3. Private auth-enabled OL sequencer RPCs

bewakes: I think that's reasonable then.

bin/strata/src/rpc/mod.rs
bewakes: I meant grouping admin related, submit related and general rpc related attributes but that should be fine for now.

bewakes (approved):

storopoli:
@purusang or @alexhui01 take a look at this? Testnet blocker

storopoli:
Rebased @purusang take a look

purusang (approved):

PR #1849, opened 2026-05-22
Str 1961 - checkpoint sync

bewakes:
## Description
**NOTE**
This PR *is* big and looks intimidating, but most of them are tests, around 900 lines are uv.lock changes, so the prod code is just about 1.6K lines. Out of that, ~1k is [Checkpoint Sync Service(CSS)](https://www.notion.so/OL-Checkpoint-Sync-309901ba000f80d98c1efc01f744b180), ~400 lines are chain-worker changes and the rest are some edits to RPC and downstream consumers.
This is created with the assistance of *Claude Code* under my direction.

Suggested review path:
**1.  `checkpoint_sync` module inside `consensus_logic` crate**

**2. `chain-worker-new` crate**
The chain worker now accepts a new message `ApplyCheckpoint` which replays state diff against previous epochal state. It also does the indexing of relevant data just like in block sync except for snark `AccountUpdateRecord` which contains `None` values for `UpdateMeta`(ol blkid + snark proof state) because those are not available in the checkpoint sync.

**3. Strata RPC changes and downstream change in alpen-client**
RPC for account epoch summary now returns an Optional next state root inside update data accounting for unavailability of that value for updates in Checkpoint sync node. This caused some downstream changes in alpen-client, especially change in the structure of `UpdateManifest` which makes the state root optional.

**4. Everything else, small changes**

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [x]  Yes
- [ ]  No

https://app.notion.com/p/OL-Checkpoint-Sync-309901ba000f80d98c1efc01f744b180

## Checklist


bewakes:
@codex review

crates/chain-worker-new/src/context.rs
```diff
@@ -370,6 +409,48 @@ fn build_indexing_writes(
...
+        .created_accounts()
+        .iter()
+        .map(|c| c.account_id())
+        .collect();
+
+    let mut account_updates: BTreeMap<AccountId, Vec<AccountUpdateRecord>> = BTreeMap::new();
+    for update in indexer_writes.snark_state_updates() {
+        let record = AccountUpdateRecord::new(
+            None,
+            *update.seqno().inner(),
```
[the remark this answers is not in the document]
bewakes: Resolved.

functional-tests/envconfigs/strata_seq_fullnode.py
```diff
@@ -58,35 +58,37 @@ def get_services(
...
         )
+        fullnode = fullnode_result.service
         fullnode.wait_for_ready(timeout=20)
 
         return {
             ServiceType.Bitcoin: bitcoind,
             ServiceType.Strata: sequencer,
             ServiceType.StrataSigner: signer,
-            STRATA_FULLNODE_SERVICE_NAME: fullnode,
+            ServiceType.StrataCheckpointNode: fullnode,
```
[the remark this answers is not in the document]
bewakes: Resolved.

crates/consensus-logic/src/checkpoint_sync/context.rs
```diff
@@ -0,0 +1,69 @@
...
+///
+/// The concrete implementation is assembled in the binary, keeping this module
+/// free of any dependency on `NodeContext`.
+pub trait CheckpointSyncCtx: Send + Sync {
+    /// Returns the rollup params.
+    fn rollup_params(&self) -> &RollupParams;
+
+    /// Fetches the current L1 chain tip height, if any. `None` is a normal
+    /// pre-sync condition (btcio reader hasn't ingested its first L1 block
+    /// yet) and callers must treat it as a wait, not an error.
```
storopoli: This is slightly harder for downstream implementers to honor.
What if you used the type system for that.

```rs
enum L1Tip { // better naming, I'm not creative today...
  Ready(L1Height),
  Pending
}
```

This makes very easy for you to force now your invariants and rules downstream to implementers.

crates/consensus-logic/src/checkpoint_sync/context.rs
```diff
@@ -0,0 +1,69 @@
...
+use strata_identifiers::Epoch;
+use strata_params::RollupParams;
+use strata_primitives::{EpochCommitment, L1Height, OLBlockCommitment};
+use strata_status::OLSyncStatus;
+
+/// Operations the checkpoint sync service needs from its environment.
+///
+/// The concrete implementation is assembled in the binary, keeping this module
+/// free of any dependency on `NodeContext`.
+pub trait CheckpointSyncCtx: Send + Sync {
```
storopoli: I'd really like if we could type the errors inside this trait instead of using `anyhow`. Maybe a tech-debt ticket with a `// TODO` comments?

crates/consensus-logic/src/checkpoint_sync/state.rs
```diff
@@ -0,0 +1,204 @@
...
+use strata_primitives::EpochCommitment;
+use strata_service::ServiceState;
+use strata_status::OLSyncStatus;
+use tracing::{debug, info};
+
+use crate::checkpoint_sync::{
+    context::CheckpointSyncCtx,
+    errors::{CheckpointSyncError, CheckpointSyncResult},
+    service::find_and_apply_unapplied_epochs,
+};
```
storopoli, line 14: This is confusing.

`state.rs` depends on `service::find_and_apply_unapplied_epochs` and `service.rs` depends on a ton of stuff from `state::*`.

What if we move the `find_and_apply_unapplied_epochs` to `state.rs` so that this dependency graph has some better logic?

crates/consensus-logic/src/checkpoint_sync/state.rs
```diff
@@ -0,0 +1,204 @@
...
+                debug!("no finalized epoch in CSM status, skipping");
+                return Ok(());
+            }
+            (None, Some(new_fin)) => {
+                info!(%new_fin, "first finalized epoch observed");
+                new_fin
+            }
+            (Some(prev), Some(new_fin)) => {
+                if prev == new_fin {
+                    debug!(%prev, "finalized epoch unchanged, skipping");
```
storopoli, line 57: Should this be a `warn!`?

crates/consensus-logic/src/checkpoint_sync/errors.rs
```diff
@@ -0,0 +1,55 @@
...
+    /// A database read failed.
+    #[error("db: {0}")]
+    Db(#[from] DbError),
+
+    /// Failure querying sync status (CSM status, L1 tip height).
+    #[error("sync status query failed")]
+    SyncStatusQuery(#[source] anyhow::Error),
+
+    /// Failure from a per-epoch chain-worker call (apply, update_safe_tip,
+    /// finalize).
```
storopoli:
```suggestion
    /// Failure from a per-epoch chain-worker call (`apply`, `update_safe_tip`,
    /// `finalize`).
```

crates/config/src/config.rs
```diff
@@ -104,8 +104,9 @@ pub struct ClientConfig {
     #[serde(default = "default_p2p_port")]
     pub p2p_port: u16,
 
-    /// Endpoint that the client will use for syncing blocks. In this case sequencer's rpc
-    /// endpoint.
+    /// OL peer endpoint to sync from.
+    /// NOTE: Currently unused. non-sequencer nodes sync from L1 checkpoints, not from a
+    /// peer RPC. Retained for config compatibility.
     pub sync_endpoint: Option<String>,
```
storopoli: We should remove this.
Maybe a `TODO` coment with a tech-debt ticket?

crates/consensus-logic/src/checkpoint_sync/service.rs
```diff
@@ -0,0 +1,293 @@
...
+                    required,
+                }) => {
+                    debug!(
+                        %epoch, depth, required,
+                        "checkpoint not reorg-safe yet, will retry on next CSM update"
+                    );
+                }
+                // Pre-sync: btcio reader hasn't published an L1 tip yet.
+                Err(CheckpointSyncError::L1TipNotReady) => {
+                    debug!("L1 tip not yet ready, will retry on next CSM update");
```
storopoli:
```suggestion
                    warn!("L1 tip not yet ready, will retry on next CSM update");
```

crates/consensus-logic/src/checkpoint_sync/service.rs
```diff
@@ -0,0 +1,293 @@
...
+                // Wait condition, not a failure: the L1 tip will advance and the
+                // next CSM update will re-trigger the scan.
+                Err(CheckpointSyncError::NotReorgSafe {
+                    epoch,
+                    depth,
+                    required,
+                }) => {
+                    debug!(
+                        %epoch, depth, required,
+                        "checkpoint not reorg-safe yet, will retry on next CSM update"
```
storopoli:
```suggestion
                    warn!(
                        %epoch, depth, required,
                        "checkpoint not reorg-safe yet, will retry on next CSM update"
```

crates/consensus-logic/src/checkpoint_sync/service.rs
```diff
@@ -0,0 +1,293 @@
...
+
+    // Publish initial OL sync status so the RPC is populated from startup.
+    match inner_state.last_finalized_and_applied() {
+        Some(epoch) => {
+            debug!(%epoch, "resuming from last finalized epoch");
+            let status = build_ol_sync_status(ctx.as_ref(), epoch).await?;
+            ctx.publish_ol_sync_status(status);
+        }
+        None => {
+            debug!("no finalized epoch found, doing nothing");
```
storopoli: I think both of these should be `INFO` after all we are starting CSS 

```suggestion
            info!(%epoch, "resuming from last finalized epoch");
            let status = build_ol_sync_status(ctx.as_ref(), epoch).await?;
            ctx.publish_ol_sync_status(status);
        }
        None => {
            info!("no finalized epoch found, doing nothing");
```

crates/consensus-logic/src/checkpoint_sync/tests.rs
```diff
@@ -0,0 +1,471 @@
...
+    l1_height: u32,
+) -> EpochSummary {
+    EpochSummary::new(
+        epoch,
+        cur_epoch.to_block_commitment(),
+        prev_epoch.to_block_commitment(),
+        make_l1_commitment(l1_height, l1_height as u8),
+        make_buf32(0xAA),
+    )
+}
```
storopoli, line 70: Can't we solve all of this using a `mockall` ting?

bin/strata/src/rpc/mod.rs
```diff
@@ -57,9 +57,9 @@ struct RpcDeps {
     max_headers_range: usize,
     storage: Arc<NodeStorage>,
     status_channel: Arc<StatusChannel>,
-    mempool_handle: Arc<MempoolHandle>,
-    #[cfg(feature = "sequencer")]
-    fcm_handle: Arc<FcmServiceHandle>,
+    /// [`None`] on checkpoint-sync nodes; `submit_transaction` returns an
+    /// unavailable error in that case.
+    mempool_handle: Option<Arc<MempoolHandle>>,
```
storopoli, line 62: Why are we removing this feature gate on `sequencer`. It is much cleaner if we keep it no?

bin/strata/src/rpc/errors.rs
```diff
@@ -47,6 +51,10 @@ pub(crate) fn map_mempool_error_to_rpc(err: OLMempoolError) -> ErrorObjectOwned
         | OLMempoolError::TransactionNotMature { .. }
         | OLMempoolError::UsedSequenceNumber { .. }
         | OLMempoolError::SequenceNumberGap { .. } => invalid_params_error(err.to_string()),
+        // Service unavailable on this node — not an error condition.
+        OLMempoolError::NotAvailable => {
+            ErrorObjectOwned::owned(NOT_AVAILABLE_ON_NODE_CODE, err.to_string(), None::<()>)
+        }
```
storopoli, line 57: We don't need this if we keep the feature gate on `sequencer`

bin/strata/src/rpc/provider.rs
```diff
@@ -23,17 +23,20 @@ use strata_storage::NodeStorage;
...
 /// Production provider that delegates to [`NodeStorage`], [`StatusChannel`],
 /// and [`MempoolHandle`].
+///
+/// `mempool_handle` is [`None`] on checkpoint-sync nodes; calls that need it
+/// (currently [`Self::submit_transaction`]) return an unavailable error.
 pub(crate) struct NodeRpcProvider {
     storage: Arc<NodeStorage>,
     status_channel: Arc<StatusChannel>,
-    mempool_handle: Arc<MempoolHandle>,
+    mempool_handle: Option<Arc<MempoolHandle>>,
```
storopoli, line 32: We don't need this change if we keep the feature gate on `sequencer`

bin/strata/src/run_context.rs
```diff
@@ -63,9 +63,13 @@ impl RunContext {
...
-    /// Returns the mempool handle.
-    pub(crate) fn mempool_handle(&self) -> &Arc<MempoolHandle> {
-        &self.service_handles.mempool_handle
+    /// Returns the mempool handle if this node runs the mempool.
+    ///
+    /// `Some` for sequencer nodes (and any future fullnode that admits txs);
+    /// `None` for checkpoint-sync nodes which never have the OL block bodies
+    /// the mempool needs.
+    pub(crate) fn mempool_handle(&self) -> Option<&Arc<MempoolHandle>> {
+        self.service_handles.mempool_handle.as_ref()
```
storopoli, line 73: We don't need this change if we keep feature gate on `sequencer`

bin/strata/src/run_context.rs
```diff
@@ -153,6 +156,18 @@ impl SequencerServiceHandles {
...
+/// Handle for whichever OL sync service the node runs.
+///
+/// A node runs exactly one: the fork-choice manager when it is a sequencer,
+/// the checkpoint sync service otherwise.
+pub(crate) enum SyncServiceHandle {
+    /// Fork-choice manager handle (sequencer nodes).
+    Fcm(Arc<FcmServiceHandle>),
+    /// Checkpoint sync service handle (non-sequencer nodes).
+    #[expect(dead_code, reason = "held to keep the service alive; not yet read")]
+    Css(Arc<CssServiceHandle>),
```
storopoli: I don't understand this `dead_code`

crates/alpen-ee/common/src/types/ol_account_epoch_summary.rs
```diff
@@ -1,27 +1,35 @@
...
+/// One snark-account update as fetched by the alpen-client OL tracker.
+///
+/// `final_state_root` is `None` when the source is a checkpoint-sync node:
+/// only the terminal epoch state root is recoverable from DA, not per-update.
+#[derive(Clone, Debug)]
+pub struct EpochUpdateOp {
+    pub seq_no: u64,
+    pub extra_data: Vec<u8>,
+    pub messages: Vec<MessageEntry>,
+    pub final_state_root: Option<Hash>,
```
storopoli: Do we need all the fields to be `pub`? Can we make then non `pub` add acessors and mutable accessors if needed?

functional-tests/envconfigs/el_ol_checkpoint_sync.py
```diff
@@ -0,0 +1,65 @@
...
+        bitcoin = strata_services[ServiceType.Bitcoin]
+
+        checkpoint_node = self._start_checkpoint_node(ectx, bitcoin)
+
+        alpen_services = AlpenClientEnv.get_services(
+            ectx,
+            self.alpen_env_params,
+            bitcoin_service=bitcoin,
+            ol_endpoint=sequencer.props["rpc_url"],
+        )
```
storopoli, line 37: This env starts the EE sequencer with only the public OL RPC URL, but `alpen-client` now requires a submit RPC URL when `--sequencer` is used. The regular `el_ol` env passes both `ol_submit_endpoint` and `ol_submit_token` from the Strata sequencer; this checkpoint-sync env should do the same. As-is, the PR's functional test fails before it reaches checkpoint sync with `--ol-submit-url is required with --sequencer when not using --dummy-ol-client`.

Permalinks:
- current call: https://github.com/alpenlabs/alpen/blob/d9b4988d0cff703a38b1dc0bced8363c398fb0a5/functional-tests/envconfigs/el_ol_checkpoint_sync.py#L31-L36
- working pattern: https://github.com/alpenlabs/alpen/blob/d9b4988d0cff703a38b1dc0bced8363c398fb0a5/functional-tests/envconfigs/el_ol.py#L80-L89
- requirement: https://github.com/alpenlabs/alpen/blob/d9b4988d0cff703a38b1dc0bced8363c398fb0a5/bin/alpen-client/src/main.rs#L262-L266

storopoli (changes requested):
There's some very good stuff here.
Some comments here are blocking others not.

crates/chain-worker-new/src/state.rs
```diff
@@ -367,12 +382,347 @@ impl ChainWorkerServiceState {
...
+    pre_seqnos: &HashMap<AccountSerial, Seqno>,
+) -> WorkerResult<Vec<SnarkAcctStateUpdate>> {
+    let mut next_seqno: HashMap<AccountSerial, Seqno> = HashMap::new();
+    let mut records = Vec::with_capacity(ol_logs.len());
+
+    for log in ol_logs {
+        let serial = log.account_serial();
+        let log_data: SnarkAccountUpdateLogData = decode_buf_exact(log.payload()).map_err(|e| {
+            WorkerError::Unexpected(format!("decode snark log for serial {serial:?}: {e}"))
+        })?;
```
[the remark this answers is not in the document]
storopoli: I think this is a valid finding @bewakes

crates/chain-worker-new/src/state.rs
```diff
@@ -367,12 +382,354 @@ impl ChainWorkerServiceState {
...
+
+    // Sanity check: indexer state root and final state root are same.
+    if final_state_root != indexer_state_root {
+        error!(
+            %epoch, %indexer_state_root, %final_state_root,
+            payload_tip_epoch = tip.epoch,
+            payload_tip_l1 = tip.l1_height(),
+            payload_tip_l2 = ?tip.l2_commitment(),
+            "epoch reconstruction state root divergence",
+        );
```
krsnapaudel: Should `apply_checkpoint_epoch` verify that the locally `reconstructed final_state_root` actually produces the checkpoint terminal block id?

crates/chain-worker-new/src/state.rs
```diff
@@ -367,12 +382,354 @@ impl ChainWorkerServiceState {
...
+    ctx: &impl ChainWorkerContext,
+    cur_epoch: Epoch,
+) -> WorkerResult<OLBlockCommitment> {
+    if cur_epoch == 0 {
+        return Err(WorkerError::Unexpected(
+            "received prev terminal request for unexpected epoch 0".to_string(),
+        ));
+    }
+    let target_epoch = cur_epoch - 1;
+    let prev_summaries = ctx.fetch_epoch_summaries(target_epoch)?;
```
krsnapaudel: Should this use `get_canonical_epoch_commitment` first?

bin/strata/src/rpc/node.rs
krsnapaudel: Can this use `chain_sync_status.confirmed_epoch`? instead of prev_epoch? CSS publishes `confirmed_epoch == finalized_epoch == current checkpoint epoch` in 
`crates/consensus-logic/src/checkpoint_sync/state.rs:190-194`.

crates/snark-acct-runtime/src/program_processing.rs
```diff
@@ -170,8 +175,10 @@ pub fn apply_update_unconditionally<P: SnarkAccountProgram>(
     // 6. Finalize state.
     program.finalize_state(state, extra_data)?;
 
-    // 7. Verify post-state matches manifest.
-    if state.compute_state_root() != manifest.new_state().inner_state() {
+    // 7. Verify post-state matches manifest, if provided.
```
krsnapaudel, line 178: Is it intentional that checkpoint-sync epoch operations apply intermediate updates with `final_state_root=None`, skipping per-update post-state checks until the terminal-per-account root?

functional-tests/tests/strata/test_checkpoint_sync_node.py
```diff
@@ -0,0 +1,114 @@
...
+    seq_updates = cast(list, seq_summary_d.pop("update_inputs", []))
+    node_updates = cast(list, node_summary_d.pop("update_inputs", []))
+
+    assert seq_summary_d == node_summary_d
+
+    for su, nu in zip(seq_updates, node_updates, strict=True):
+        s_root = su.pop("final_state_root")
+        n_root = nu.pop("final_state_root")
+        assert n_root is None or n_root == s_root, "final_state_root if present must match"
+        assert su == nu
```
krsnapaudel, line 115: Can this test assert that the last update for the queried account has `final_state_root` set and equal to the sequencer’s root? The current check allows every checkpoint-sync update root to be None, which would miss a bug where CSS never emits the terminal-per-account root that the EE tracker relies on for post-state validation.

crates/chain-worker-new/src/state.rs
```diff
@@ -367,12 +382,354 @@ impl ChainWorkerServiceState {
...
+            base_state.last_l1_height()
+        ))
+    })?;
+    let to_height = tip.l1_height();
+    if to_height < from_height {
+        return Err(WorkerError::Unexpected(format!(
+            "checkpoint payload L1 range inverted at {epoch}: from={from_height} to={to_height}"
+        )));
+    }
+    let range_len = (to_height - from_height + 1) as u64;
```
krsnapaudel: Is this better:
`let range_len = u64::from(to_height) - u64::from(from_height) + 1;`

crates/chain-worker-new/src/state.rs
```diff
@@ -367,12 +382,354 @@ impl ChainWorkerServiceState {
...
+                WorkerError::Unexpected(format!(
+                    "snark log references unknown account serial {serial:?}"
+                ))
+            })?;
+
+        let cur = next_seqno
+            .entry(serial)
+            .or_insert_with(|| pre_seqnos.get(&serial).copied().unwrap_or_else(Seqno::zero));
+        *cur = cur.incr();
+        let seqno = *cur;
```
krsnapaudel, line 793: `rebuild_snark_records_from_logs` reconstructs per-update seqnos by incrementing once per log. Where does the invariant come from? Should we add a debug/assertion that the reconstructed final seqno for each account matches the DA-applied terminal account seqno?

crates/ol/stf/src/verification.rs
krsnapaudel: Not related, but Claude found this typo.

krsnapaudel (commented):
Looks good for the most part. Some comments. Reviewed with help from Claude Code and Codex.

crates/chain-worker-new/src/state.rs
bewakes: Perhaps not. It's more characters and two conversions :D

crates/snark-acct-runtime/src/program_processing.rs
bewakes: Yes, that is intentional. checkpoint syncing cannot provide intermediate state roots.

crates/alpen-ee/ol-tracker/src/task.rs
```diff
@@ -154,14 +154,21 @@ pub(crate) async fn track_ol_state(
...
-    epoch_operations: &[UpdateInputData],
+    epoch_operations: &[EpochUpdateOp],
 ) -> Result<()> {
     for op in epoch_operations {
+        if op.final_state_root.is_none() {
+            warn!(
+                seq_no = op.seq_no,
+                "applying update without post-state check"
+            );
+        }
```
sapinb, line 179: Shouldn't the final state root be present for the end of the epoch during checkpoint sync ? Once per epoch ?
Option<> doesn't really convey this meaning.
Pass in the final state root separately and do a single check, either at the end of this function or in handle_extend_ee_state after this call.

bewakes: Yeah, this doesn't explicitly check the condition. Instead of passing the last root separately(which again would have to be an option because the updates might be empty in an epoch), I did an explicit check in the function that ensures the last update has root.
Please check https://github.com/alpenlabs/alpen/pull/1849/commits/9be02cf39fdfa2fb0e841a7e936cdb25ab691e5c

crates/consensus-logic/src/checkpoint_sync/context.rs
bewakes: The comment is misleading and assumes the shape of other components. I've updated the comment and left it as is. Enum would be a bit overkill imo. Just return None if not available

crates/consensus-logic/src/checkpoint_sync/tests.rs
bewakes: We could probably but since these tests are stateful and we don't yet require per-line changes, I'd keep these as they are. Plus the async interfaces might make it a bit harder to use.

bin/strata/src/rpc/mod.rs
bewakes: The fcm handle is moved inside seq deps, as it is only used for sequencer related RPCs.

bin/strata/src/run_context.rs
bewakes: The handle was not being used anywhere so it was unused. The "keeping the service alive" is not true. I've updated this now.

bin/strata/src/rpc/errors.rs
bewakes: Yeah, I thought quite about this and came to conclusion(sort of) not to make mempool sequencer specific(which is true now), the rpcs also have been designed leaving possibility for non-sequencer nodes to have mempool later.

bewakes:
@storopoli and @krsnapaudel thanks very much your feedback. Those are great and actionable!

@storopoli your requests have been addressed in https://github.com/alpenlabs/alpen/pull/1849/commits/d0f423f63ce0ad3843daf7d0d46ff26cf9540447.

@krsnapaudel your requests have been addressed in https://github.com/alpenlabs/alpen/pull/1849/commits/3b04f93dc14cafa776f46d036a6cb6fa1da7d1f4

crates/chain-worker-new/src/state.rs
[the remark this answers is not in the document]
bewakes: Yes, this is valid. Done in https://github.com/alpenlabs/alpen/pull/1849/commits/9be02cf39fdfa2fb0e841a7e936cdb25ab691e5c

crates/config/src/config.rs
bewakes: Do you think we might need to have it back? I am asking this because [here](https://www.notion.so/Core-system-mainnet-plan-1db901ba000f808fa25adeae5467238d?source=copy_link#1f7901ba000f80ae9c2fc5fed14a49aa) it indicates a possibility that we might have p2p sync(tbd). So, if that's the case, a NOTE could be fine now and we just use that when time comes.

krsnapaudel:
> @storopoli and @krsnapaudel thanks very much your feedback. Those are great and actionable!
> 
> @storopoli your requests have been addressed in [d0f423f](https://github.com/alpenlabs/alpen/commit/d0f423f63ce0ad3843daf7d0d46ff26cf9540447).
> 
> @krsnapaudel your requests have been addressed in [3b04f93](https://github.com/alpenlabs/alpen/commit/3b04f93dc14cafa776f46d036a6cb6fa1da7d1f4)

Looks good. `apply_checkpoint_epoch` now skips non-snark `ol_logs` during snark-record reconstruction. Could we add a checkpoint-sync regression test where the checkpoint sidecar contains both snark update logs and a non-snark OL log (e.g. bridge withdrawal intent log), to ensure reconstruction ignores the non-snark log and still matches full-sync output?

bin/strata/src/services.rs
```diff
@@ -258,8 +264,16 @@ pub(crate) fn start_strata_services(
         nodectx.status_channel().as_ref(),
     )?;
 
-    // Start mempool service
-    let mempool_handle = Arc::new(start_mempool(&nodectx)?);
+    // Mempool and the OL checkpoint builder are sequencer-only: a checkpoint-sync
```
barakshani: >Mempool ... sequencer-only

You already state above that you don't want to exclude non-sequencer nodes from running mempool. So at least for consistency I would phrase this differently; in the future, when we have more options for non-sequencer nodes this comment would be one less thing we need to change.

crates/alpen-ee/common/src/types/ol_account_epoch_summary.rs
```diff
@@ -1,27 +1,67 @@
-use strata_identifiers::EpochCommitment;
-use strata_snark_acct_types::UpdateInputData;
+use strata_acct_types::MessageEntry;
+use strata_identifiers::{EpochCommitment, Hash};
+
+/// One snark-account update as fetched by the alpen-client OL tracker.
+///
+/// `final_state_root` is `None` when the source is a checkpoint-sync node:
```
barakshani: 1. why is it called _final_ state root?
2. why `None`?

for both, don't we get an epoch update once per checkpoint? (if not, then when exactly?) so anyway we talk about _terminal-based_ increments, and final (if the meaning is that it's the epoch's final state root) is redundant.

crates/chain-worker-new/src/tests/apply_checkpoint.rs
```diff
@@ -0,0 +1,696 @@
+//! Cross-mode consistency tests for checkpoint-sync epoch reconstruction.
+//!
+//! Each test runs one epoch two ways — full sync (block-by-block STF) and
```
barakshani: why "full sync" and not "block sync". Intuitively, I don't know what full sync means.

crates/chain-worker-new/src/context.rs
```diff
@@ -344,7 +370,11 @@ fn build_indexing_writes(
 
     let mut account_updates: BTreeMap<AccountId, Vec<AccountUpdateRecord>> = BTreeMap::new();
     for update in indexer_writes.snark_state_updates() {
-        let meta = AccountUpdateMeta::new(commitment, update.state());
+        // Full-sync always knows the post-state root for every update.
+        let state_root = update
+            .state()
+            .expect("full-sync snark update must carry a state root");
```
barakshani: I'd rather not call it "full-sync", it's not intuitive.

crates/chain-worker-new/src/state.rs
```diff
@@ -166,16 +177,24 @@ impl ChainWorkerServiceState {
 
         // Handle epoch terminal if needed
         debug!(slot=%block.header().slot(), is_terminal=%block.header().is_terminal(), "Checking if block is terminal");
-        if block.header().is_terminal() {
-            self.handle_complete_epoch(&block, &output, &new_state)?;
-            // Send the epoch commitment to receiver
+        let completed_epoch = if block.header().is_terminal() {
```
barakshani: maybe _sealed_ instead of _completed_, for consistency just in case we use "sealed" across the board?

crates/chain-worker-new/src/state.rs
```diff
@@ -367,12 +382,456 @@ impl ChainWorkerServiceState {
...
+    let mut indexer_state = IndexerState::new(tracking_state);
+    apply_da_epoch::<_, OLDaSchemeV1>(&mut indexer_state, &epoch_info, da_payload, &manifest_cont)?;
+    let indexer_state_root = indexer_state
+        .compute_state_root()
+        .map_err(|e| WorkerError::Unexpected(format!("compute indexer state root: {epoch} {e}")))?;
+
+    // Replace the DA-collapsed snark records (one per account) with per-update
+    // records derived from the checkpoint sidecar's `ol_logs`.
+    let (rebuilt_snark_updates, derived_seqnos) =
+        rebuild_snark_records_from_logs(&indexer_state, sidecar.ol_logs(), &pre_seqnos)?;
```
barakshani: what is significant about seq_no? It's part of the snark account state like, balance, for example. Shouldn't we _apply da_ generically, get the output result (state) and persist it? Why seqnos get special treatment?

barakshani: I guess it's also related to `verify_snark_seqno_invariant` just below.

crates/chain-worker-new/src/state.rs
```diff
@@ -367,12 +382,456 @@ impl ChainWorkerServiceState {
...
+    Ok(AppliedEpochArtifacts {
+        terminal,
+        new_state,
+        summary,
+        output,
+    })
+}
+
+/// Validates the epoch's L1 range and builds the manifest container and
+/// [`EpochInfo`] used to drive [`apply_da_epoch`].
```
barakshani, line 560: I understand I don't have enough context, but I rather just ask: what exactly are we doing here that is different from how ASM verifies a checkpoint? The main thing I would like to know is that -- given we already have checkpoint verification in ASM -- can't we just call the _verify checkpoint_ procedures that ASM does, instead of re-doing it in here?

bewakes: This is not duplicating what ASM has done by verifying checkpoint. Here, in order to apply the checkpoint(which doesn't contain asm manifests and only state diffs), we want to create manifests in order to update the epochal state after applying state diffs.

crates/chain-worker-new/src/state.rs
bewakes: Yes, it is for the verification as well as it is also needed to create update records that contain extra_data which is needed for alpen-client to reconstruct state.

This is basically to construct [Update](https://github.com/alpenlabs/alpen/blob/32f7615e8d5c48590ac80b6f207a67074998903e/crates/ol/state-support-types/src/index_types.rs#L102) variant of [SnarkAccountUpdate](https://github.com/alpenlabs/alpen/blob/32f7615e8d5c48590ac80b6f207a67074998903e/crates/ol/state-support-types/src/index_types.rs#L161) struct which holds extra data. seq number got pulled in while creating this.

crates/alpen-ee/common/src/types/ol_account_epoch_summary.rs
bewakes: Yeah, the naming is a bit confusing. I have moved the  final state root one layer outside so that it now correctly represents the final state root at the end of an epoch.

The existing `final_state_root` has been renamed to `new_state_root` which represents the state root after each snark update. Since state root after each update is not recoverable in a checkpoint sync node, it is `None`.

bin/strata/src/services.rs
bewakes: Please check https://github.com/alpenlabs/alpen/pull/1849/commits/a4f16211fec03970b2624c38245d66d4a293c90b where I've done the renamings and some cleanups.

crates/config/src/config.rs
storopoli: Yes, eventually we'll have it. However, this is getting in the way right now. Why do you have an optional string for something that you may have 1 year from now?

bin/strata/src/rpc/mod.rs
storopoli: Ok that's not a hard blocker. Let's solve this.

bin/strata/src/rpc/errors.rs
storopoli: Ok

crates/config/src/config.rs
storopoli: This is my only blocker.

bin/strata/src/context.rs
```diff
@@ -137,8 +137,11 @@ fn get_config(args: Args) -> Result<Config, InitError> {
 }
 
 fn validate_config(config: Config) -> Result<Config, InitError> {
-    if !config.client.is_sequencer && config.client.sync_endpoint.is_none() {
-        return Err(InitError::MissingSyncEndpoint);
+    if config.client.sync_endpoint.is_some() {
+        warn!(
+            "client.sync_endpoint is set but ignored: non-sequencer nodes sync from L1 checkpoints, \
+             not from a peer RPC"
+        );
```
delbonis: What if we aren't a sequencer though?  This warning should probably be more specific.

crates/consensus-logic/src/checkpoint_sync/service.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+        Ok(Response::Continue)
+    }
+}
+
+/// Launches the checkpoint sync service and returns its monitor.
+///
+/// Takes the context and raw service inputs directly so this module needs no
+/// dependency on `NodeContext`; the binary assembles `ctx`.
+pub async fn start_css_service<C: CheckpointSyncCtx + 'static>(
```
delbonis: chain sync service service

Also can add `'static` to the ctx trait instead of needing to restate it here.

crates/consensus-logic/src/checkpoint_sync/errors.rs
```diff
@@ -0,0 +1,58 @@
...
+    #[error("epoch summary missing for {0}")]
+    MissingEpochSummary(EpochCommitment),
+
+    /// A database read failed.
+    #[error("db: {0}")]
+    Db(#[from] DbError),
+
+    /// Failure querying sync status (CSM status, L1 tip height).
+    #[error("sync status query failed")]
+    SyncStatusQuery(#[source] anyhow::Error),
```
delbonis: Don't wrap anyhow errors with thiserror errors, it garbles the error tracking.

crates/consensus-logic/src/checkpoint_sync/errors.rs
```diff
@@ -0,0 +1,58 @@
...
+    EpochOp {
+        epoch: EpochCommitment,
+        op: &'static str,
+        #[source]
+        cause: anyhow::Error,
+    },
+
+    /// Failure updating the chain worker's safe tip.
+    #[error("chain worker update_safe_tip")]
+    SafeTipUpdate(#[source] anyhow::Error),
```
delbonis: same here

crates/consensus-logic/src/checkpoint_sync/errors.rs
```diff
@@ -0,0 +1,58 @@
...
+    SyncStatusQuery(#[source] anyhow::Error),
+
+    /// Failure from a per-epoch chain-worker call (`apply`, `finalize`).
+    #[error("chain worker {op} at {epoch}")]
+    EpochOp {
+        epoch: EpochCommitment,
+        op: &'static str,
+        #[source]
+        cause: anyhow::Error,
+    },
```
delbonis, line 48: This is kinda unclear, also this should include the underlying error in the Debug message.

crates/consensus-logic/src/checkpoint_sync/errors.rs
```diff
@@ -0,0 +1,58 @@
...
+    #[error("L1 canonical chain tip not yet ingested")]
+    L1TipNotReady,
+
+    /// A finalized epoch has no L1 observation entry in the database.
+    #[error("finalized epoch {0} has no l1 observation entry")]
+    MissingL1Ref(EpochCommitment),
+
+    /// A finalized epoch's checkpoint is not buried deep enough to be
+    /// reorg-safe, despite a descendant epoch being finalized.
+    #[error("epoch {epoch} not reorg-safe: buried {depth} blocks, need {required}")]
```
delbonis: colons imply a new level of error hierarchy, what's after the colon here should be in parens

crates/consensus-logic/src/checkpoint_sync/service.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+/// Catches up on any unapplied finalized epochs at startup and returns the
+/// resulting last-applied epoch.
+///
+/// Also re-runs finalization on the last already-applied epoch found by the
+/// scan: if a previous run crashed between writing the summary and finalizing,
+/// the chain worker's `last_finalized_epoch` would otherwise stay behind
+/// silently. The re-finalize is idempotent.
+async fn initialize_css_inner_state(
```
delbonis, line 137: `load_`?

crates/consensus-logic/src/checkpoint_sync/state.rs
```diff
@@ -0,0 +1,293 @@
...
+        Self {
+            last_finalized_and_applied,
+        }
+    }
+
+    pub(crate) fn last_finalized_and_applied(&self) -> Option<EpochCommitment> {
+        self.last_finalized_and_applied
+    }
+}
+
```
delbonis: This could probably be flattened out into `CheckpointSyncState`, not a big deal.  `InnerState` is a bit of a dubious name.

bin/strata/src/css.rs
```diff
@@ -0,0 +1,153 @@
...
+                epoch,
+                op: "finalize_epoch",
+                cause: cause.into(),
+            })
+    }
+
+    fn publish_ol_sync_status(&self, status: OLSyncStatus) {
+        self.status_channel
+            .update_ol_sync_status(OLSyncStatusUpdate::new(status));
+    }
```
delbonis, line 125: This makes sense to work like this, but could we update downstream listeners to rely on a status monitor?

Could have a helper type like this to provide indirection, without needing to introduce new traits:

```rs
pub enum OLSyncMonitor {
  FullSync(StatusMonitor<...>),
  CheckpointSync(StatusMonitor<...>)
}

impl OLSyncMonitor {
  fn get_sync_tip(&self) -> ... {
    match self {
      Self::FullSync(...) => ...,
      Self::CheckpointSync(...) => ...
    }
  }
}
```

Possibly could be deferred in a followup PR, but open a ticket in that case.

crates/alpen-ee/common/src/types/ol_account_epoch_summary.rs
```diff
@@ -1,28 +1,75 @@
...
+///
+/// `new_state_root` is `None` for intermediate updates fetched from a
+/// checkpoint-sync source.
+#[derive(Clone, Debug)]
+pub struct EpochUpdateOp {
+    seq_no: u64,
+    extra_data: Vec<u8>,
+    messages: Vec<MessageEntry>,
+    new_state_root: Option<Hash>,
+}
```
delbonis, line 15: This naming is unclear, it suggests it's something about the update.  Should be something like `SnarkAccountUpdateInfo`.

Also *maybe* could include the new next message index so that we don't have to separately track it (or to give better sanity checks to whatever internal tracking we're doing ourselves).

Then we could also provide an iterator over the messages that provides their indexes too.

```rs
fn iter_messages_with_idxs(&self) -> impl Iterator<Item = (u64, &MessageEntry)> {
  let first_idx = self.new_next_msg_idx() - self.messages.len() as u64;
  self.messages.iter().enumerate().map(|(i, m)| (first_idx + i, m))
}
```

crates/alpen-ee/ol-tracker/src/task.rs
```diff
@@ -18,7 +19,8 @@ use crate::{
 #[derive(Debug)]
 pub(crate) struct OLEpochOperations {
     pub epoch: EpochCommitment,
-    pub operations: Vec<UpdateInputData>,
+    pub final_state_root: Hash,
+    pub operations: Vec<EpochUpdateOp>,
 }
```
delbonis, line 25: Similarly, `SnarkAccountEpochUpdates` or something.

crates/alpen-ee/ol-tracker/src/task.rs
```diff
@@ -154,14 +157,26 @@ pub(crate) async fn track_ol_state(
...
+        return Ok(());
+    }
+
     for op in epoch_operations {
+        if op.new_state_root().is_none() {
+            warn!(
+                seq_no = op.seq_no(),
+                "applying update without post-state check"
+            );
+        }
```
delbonis, line 179: This isn't really warn-worthy, imo.  Could be a `debug` but we'd expect to see a lot of these and the warning isn't actually actionable.  Warning logs should indicate something *seems* wrong but isn't attributable and we can proceed as normal.

crates/alpen-ee/ol-tracker/src/task.rs
```diff
@@ -170,6 +185,13 @@ pub(crate) fn apply_epoch_operations(
         .map_err(|e| OLTrackerError::Other(e.to_string()))?;
     }
 
+    let observed = state.compute_state_root();
+    if observed != final_state_root {
+        return Err(OLTrackerError::Other(format!(
+            "epoch terminal state root mismatch: observed {observed} expected {final_state_root}"
+        )));
```
delbonis: Convert to typed errors?

delbonis (changes requested):
I don't know for sure if I've covered everything but seems like a good start.  Overall the approach seems reasonable, I don't have any major design critiques, just cleanliness and organizational ones.

bin/strata/src/css.rs
bewakes: The enum already exists as `SyncHandle(see `bin/strata/src/run_context.rs`). We already have [a spike](https://alpenlabs.atlassian.net/browse/STR-3352) for replacing `StatusChannel`, so that should cover the downstream usage.

crates/alpen-ee/common/src/types/ol_account_epoch_summary.rs
bewakes: Yeah, the naming is unclear. Renamed to `SnarkAccountUpdateInfo`.

On adding new next msg index and iterator, skipping for now until a usecase actually appears. The message cursor divergence is already caught at the end of `apply_epoch_operations` where any change in msg index changes the state root. The iterator is a neat idea but it feels like it is not required yet.

crates/consensus-logic/src/checkpoint_sync/service.rs
bewakes: Since this is doing much more than just "loading" it from somewhere else (disk, memory, etc), I'd keep it as is.

bin/strata/src/context.rs
bewakes: Removed the `sync_endpoint` altogether.
cc: @storopoli

bewakes:
Thanks @delbonis for your suggestions. They have been resolved in https://github.com/alpenlabs/alpen/pull/1849/commits/3926a4118e1f44f330599fa1c87332d34bd40103 and https://github.com/alpenlabs/alpen/pull/1849/commits/c7a59fbab8b4955edfc30aa908281df28ed3b5dd

storopoli (approved):
ACK 3425a455a4b6b695896f221fdccb38455fdf3c8f

crates/chain-worker-new/src/state.rs
barakshani: If manifests is what we're missing, and given that ASM is already producing these manifests, why can't we have it taken from ASM?

bin/strata/src/rpc/node.rs
```diff
@@ -568,6 +572,11 @@ impl<P: OLRpcProvider> OLClientRpcServer for OLRpcServer<P> {
             .get_account_state(&account_id)
             .ok_or_else(|| not_found_error(format!("Account {account_id} not found")))?;
 
+        let snark_state = account_state.as_snark_account().map_err(|_| {
+            invalid_params_error(format!("Account {account_id} is not a snark account"))
+        })?;
+        let final_state_root: HexBytes32 = snark_state.inner_state_root().0.into();
```
barakshani, line 578: why `final`? This is just the (current) state root at the time of making the request, no?

bin/strata/src/rpc/node.rs
```diff
@@ -385,7 +389,7 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
                     pending.push(Pending {
                         block_commitment,
                         seq_no: r.seq_no(),
-                        final_state_root: meta.final_state_root(),
+                        final_state_root: meta.new_state_root(),
```
barakshani: something here just feels not right!

crates/chain-worker-new/src/state.rs
```diff
@@ -337,12 +352,413 @@ impl ChainWorkerServiceState {
         Ok(())
     }
 
-    /// Finalizes an epoch, merging write batches into finalized state.
+    /// Marks an epoch as finalized.
```
barakshani: finalized as buried or as in sealed (we've produced a terminal block)?

crates/chain-worker-new/src/state.rs
```diff
@@ -337,12 +352,413 @@ impl ChainWorkerServiceState {
...
+        .map(|s| *s.terminal())
+        .ok_or(WorkerError::MissingSummaryForEpoch(target_epoch))
+}
+
+/// Safety cap on the L1 block range an epoch's manifests can span.
+///
+/// Defends against an outsized range from a malformed payload allocating an
+/// unbounded `Vec<AsmManifest>`. Tuned generously: a healthy epoch spans at
+/// most a few hundred L1 blocks.
+const MAX_EPOCH_L1_RANGE: u64 = 100_000;
```
barakshani: What exactly is this being used for?

crates/chain-worker-new/src/state.rs
```diff
@@ -337,12 +352,413 @@ impl ChainWorkerServiceState {
...
+    let base_state = MemoryStateBaseLayer::new(base_state_raw);
+
+    let sidecar = payload.sidecar();
+    let terminal = *tip.l2_commitment();
+    let da_payload = decode_ol_da_payload_bytes(sidecar.ol_state_diff())
+        .map_err(|e| WorkerError::Unexpected(format!("decode OL DA payload: {e}")))?;
+    let (manifest_cont, epoch_info) =
+        assemble_da_inputs(ctx, epoch, &base_state, sidecar, tip, prev_terminal)?;
+
+    // Read pre-epoch seqno for every snark account touched by `ol_logs`.
```
barakshani: "pre-epoch" means...?

crates/chain-worker-new/src/state.rs
barakshani: Given where this is [being used below](https://github.com/alpenlabs/alpen/pull/1849/changes#diff-20e9a57dbb49eedbd62149ff53f123313545a7085994bdd1358a4b29abbff0feR551-R552), note that right now we have a max-manifest-per-epoch set to 2048: https://github.com/alpenlabs/alpen/blob/71d03a2a951aeb6a5dbd7611f2fd424a2ec789a2/crates/ol/chain-types/ssz/block.ssz#L8

crates/ol/rpc/types/src/account_summary.rs
```diff
@@ -17,6 +17,8 @@ pub struct RpcAccountEpochSummary {
     prev_epoch_commitment: EpochCommitment,
     /// Balance of account at the end of this epoch in sats.
     balance: u64,
+    /// Account's state root at the end of this epoch.
+    final_state_root: HexBytes32,
```
barakshani, line 21: if we don't call the previous field "final_balance", why do we call this one "final_state_root"?

functional-tests/tests/strata/test_checkpoint_sync_node.py
```diff
@@ -0,0 +1,121 @@
...
+        sequencer: StrataService = self.get_service(ServiceType.Strata)
+        checkpoint_node: StrataService = self.get_service(ServiceType.StrataCheckpointNode)
+        bitcoin: BitcoinService = self.get_service(ServiceType.Bitcoin)
+        btc_rpc = bitcoin.create_rpc()
+
+        # Wait for rpcs to be ready.
+        sequencer.wait_for_rpc_ready(timeout=20)
+        checkpoint_node.wait_for_rpc_ready(timeout=20)
+
+        # Walk epochs as the EE node posts updates, collecting epochs whose ALPEN
```
barakshani:
```suggestion
        # Walk epochs as the EE node posts updates, collecting epochs whose EE
```
This is a Strata binary/test, it does not aware of the EE internals.

functional-tests/tests/strata/test_checkpoint_sync_node.py
```diff
@@ -0,0 +1,121 @@
...
+            lambda st: st["finalized"]["epoch"] >= last_active,
+            error_with=f"checkpoint-sync node did not finalize epoch {last_active}",
+            timeout=120,
+        )
+
+        # Each active epoch's reconstructed account summary must be identical to
+        # the sequencer's, including the non-empty update inputs.
+        for epoch in active_epochs:
+            seq_summary = sequencer.get_account_epoch_summary(ALPEN_ACCOUNT_ID, epoch)
+            node_summary = checkpoint_node.get_account_epoch_summary(ALPEN_ACCOUNT_ID, epoch)
```
barakshani, line 93: I don't follow this. We are running an _OL_ checkpoint sync, so we should make sure the entire OL state is reconstructed properly. Why do we compare epoch summary of a specific snark account?

barakshani (commented):
>Is this PR addressing any specification, design doc or external reference document?

Should be proper to address this with a link to the design doc.

crates/alpen-ee/common/src/types/ol_account_epoch_summary.rs
delbonis: It would make troubleshooting easier since we can (more conveniently) print the messages indexes in logs.

crates/chain-worker-new/src/state.rs
delbonis: The phrasing "touched by `ol_logs`" is also a bit misleading.  The logs are a record of the changes that were made, were using them to infer the changes made (since we don't directly have them) rather than having the logs be the source of truth.

bin/strata/src/css.rs
delbonis: I'm starting to get more concerned about the amount of logic that's being added directly to the binary crate.  I understand that part of this is being forced, but the more we contribute to it the more difficult it is to break it out into more modular components in the future.

bin/strata/src/run_context.rs
```diff
@@ -158,6 +154,17 @@ impl SequencerServiceHandles {
...
 
+/// Handle for whichever OL sync service the node runs.
+///
+/// A node runs exactly one: the fork-choice manager when it is a sequencer,
+/// the checkpoint sync service otherwise.
+pub(crate) enum SyncServiceHandle {
+    /// Fork-choice manager handle (sequencer nodes).
+    Fcm(Arc<FcmServiceHandle>),
+    /// Checkpoint sync service handle (non-sequencer nodes).
+    Css { _handle: Arc<CssServiceHandle> },
```
delbonis: Should make this variant style consistent.

This should be moved into the consensus-logic crate, closer to the services that *it is a handle to*.

crates/alpen-ee/common/src/types/ol_account_epoch_summary.rs
```diff
@@ -1,28 +1,82 @@
...
 
 #[derive(Debug)]
 pub struct OLEpochSummary {
     epoch: EpochCommitment,
     prev: EpochCommitment,
-    updates: Vec<UpdateInputData>,
+    final_state_root: Hash,
+    updates: Vec<SnarkAccountUpdateInfo>,
 }
```
delbonis, line 75: This one too.

crates/chain-worker-new/src/state.rs
```diff
@@ -337,12 +352,413 @@ impl ChainWorkerServiceState {
...
+
+    /// Reconstructs an epoch's OL state from its checkpoint and persists it. Used by checkpoint
+    /// sync.
+    #[instrument(
+        skip_all,
+        fields(epoch = epoch.epoch(), slot = epoch.last_slot(), blkid = %epoch.last_blkid()),
+        err
+    )]
+    pub(crate) fn apply_checkpoint(&mut self, epoch: EpochCommitment) -> WorkerResult<()> {
+        let artifacts = apply_checkpoint_epoch(&self.ctx, epoch)?;
```
delbonis, line 388: After the discussion earlier today I'm worried now this might *also* be too verbose to log like this, especially considering that we have several other places where we log checkpointing/tip updates.  Should lower this at least down to debug.

crates/chain-worker-new/src/state.rs
delbonis: Move to top of file or somewhere else that more conspicuous.

crates/chain-worker-new/src/state.rs
```diff
@@ -337,12 +352,413 @@ impl ChainWorkerServiceState {
...
+    let range_len = (to_height - from_height + 1) as u64;
+    if range_len > MAX_EPOCH_L1_RANGE {
+        return Err(WorkerError::Unexpected(format!(
+            "epoch L1 manifest range too large at {epoch}: {range_len} blocks (max {MAX_EPOCH_L1_RANGE})"
+        )));
+    }
+    let manifests = ctx.fetch_l1_manifests(from_height, to_height)?;
+    let manifest_cont = OLL1ManifestContainer::new(manifests)
+        .map_err(|e| WorkerError::Unexpected(format!("build manifest container: {e}")))?;
+
```
delbonis, line 601: Add new error variants. :)

crates/chain-worker-new/src/state.rs
```diff
@@ -337,12 +352,413 @@ impl ChainWorkerServiceState {
...
+        error!(
+            %epoch,
+            expected_blkid = %terminal.blkid(),
+            %reconstructed_blkid,
+            "terminal header reconstruction blkid mismatch",
+        );
+        return Err(WorkerError::Unexpected(format!(
+            "terminal blkid mismatch at {epoch}: expected={} reconstructed={reconstructed_blkid}",
+            terminal.blkid()
+        )));
```
delbonis: Here too. :)

crates/chain-worker-new/src/state.rs
```diff
@@ -337,12 +352,413 @@ impl ChainWorkerServiceState {
...
+        })?;
+        if post != derived {
+            error!(
+                ?serial, %account_id, ?derived, ?post,
+                "snark seqno invariant violated",
+            );
+            return Err(WorkerError::Unexpected(format!(
+                "snark seqno mismatch for {serial:?} ({account_id}): derived={derived:?} post={post:?}"
+            )));
+        }
```
delbonis: And here. :)

crates/chain-worker-new/src/state.rs
```diff
@@ -337,12 +352,413 @@ impl ChainWorkerServiceState {
...
+        let account_id = state
+            .find_account_id_by_serial(serial)
+            .map_err(|e| {
+                WorkerError::Unexpected(format!("resolve post-state serial {serial:?}: {e}"))
+            })?
+            .ok_or_else(|| {
+                WorkerError::Unexpected(format!(
+                    "snark log references unknown account serial {serial:?}"
+                ))
+            })?;
```
delbonis: And here. :)

delbonis (changes requested):

crates/snark-acct-runtime/src/program_processing.rs
```diff
@@ -136,10 +136,15 @@ fn verify_coinput_and_process_message<P: SnarkAccountProgramVerification>(
 /// Applies an update unconditionally without verification.
 ///
 /// This is used outside the proof, after verifying the proof, to reconstruct
-/// the actual state from DA.  It decodes the extra data and messages from the
+/// the actual state from DA. It decodes the extra data and messages from the
```
barakshani, line 139: can you explain this? Why we need to reconstruct _something_ after the proof verification? what do actually verify? what is the "actual" state (is there _non-actual_ state??)?

crates/chain-worker-new/src/state.rs
bewakes: yes, we do take it from the database itself. Here we are just constructing appropriate inputs for da application.

crates/ol/stf/src/test_utils.rs
```diff
@@ -1978,3 +2001,122 @@ impl SnarkUpdateBuilder {
...
+    snark_id: AccountId,
+) -> (&OLAccountState, &OLSnarkAccountState) {
+    let snark_account = state.get_account_state(snark_id).unwrap().unwrap();
+    (snark_account, snark_account.as_snark_account().unwrap())
+}
+
+/// The inbox message a GAM block delivers and a snark update consumes in tests.
+pub fn snark_inbox_msg() -> MessageEntry {
+    snark_inbox_msg_with_data(b"inbox msg")
+}
```
barakshani, line 2057: This is the full OL state: https://github.com/alpenlabs/alpen/blob/20a23d395f8f0c5f51373d6c1a64b90af3c7d692/crates/ol/state-types/ssz/state.ssz#L10-L25

Do we have unit test showing the full state is reconstructed? I feel like the focus is on snark accounts.

functional-tests/tests/strata/test_checkpoint_sync_node.py
barakshani: See https://github.com/alpenlabs/alpen/pull/1849/changes#r3340577895

crates/alpen-ee/common/src/types/ol_account_epoch_summary.rs
bewakes: What about this one?

bewakes:
@delbonis PTAL at https://github.com/alpenlabs/alpen/pull/1849/commits/76b0b886ed8f13d31e8210afa64e04043e163455 and https://github.com/alpenlabs/alpen/pull/1849/commits/d0a2cda3fc8ea038e9d8a1d9485866cb774b459b

bin/strata/src/css.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// Node storage for checkpoint and L1 lookups.
+    storage: Arc<NodeStorage>,
+    /// Chain worker handle used to apply, advance and finalize epochs.
+    chain_worker: Arc<ChainWorkerHandle>,
+    /// Monitor exposing the current CSM worker status.
+    csm_monitor: Arc<ServiceMonitor<CsmWorkerStatus>>,
+    /// Status channel for publishing OL sync status updates.
+    status_channel: Arc<StatusChannel>,
+    /// Rollup params, used for the L1 reorg-safe depth.
+    rollup_params: RollupParams,
```
purusang, line 34: Will be removed soon https://github.com/alpenlabs/alpen/pull/1928

crates/alpen-ee/common/src/types/ol_account_epoch_summary.rs
delbonis: Sorry for not specifying.  The name is misleading like the other type names, change it like for `SnarkAccountUpdateInfo` to clarify that it's about a specific account rather than being a "summary of the epoch" as a whole.

bewakes: This is already `SnarkAccountUpdateInfo`. I think that's the name you are suggesting to change to?

crates/consensus-logic/src/checkpoint_sync/input.rs
delbonis, line 1: Maybe there's some reason for this but surely we can use the generic watch channel receiver?

crates/ol/stf/src/test_utils.rs
```diff
@@ -2005,3 +2028,122 @@ impl SnarkUpdateBuilder {
...
+pub const EPOCH_RUNNER_GENESIS_TIMESTAMP: u64 = 1_000_000;
+
+/// Per-slot timestamp step used by the shared epoch runner.
+pub const EPOCH_RUNNER_SLOT_TIMESTAMP_STEP: u64 = 1_000;
+
+/// L1 height of an epoch's terminal manifest under the shared epoch runner.
+///
+/// Genesis carries the manifest at height 1; the non-terminal blocks of the
+/// epoch under test carry none, so the terminal manifest is always at height 2.
+pub const EPOCH_RUNNER_TERMINAL_L1_HEIGHT: u32 = 2;
```
delbonis: Move to top of file.

delbonis (changes requested):
Okay aside from the other outstanding comments I think we're close to being able to move forwards with this.

crates/alpen-ee/ol-tracker/src/task.rs
```diff
@@ -160,20 +164,42 @@ pub(crate) async fn track_ol_state(
 
 pub(crate) fn apply_epoch_operations(
     state: &mut EeAccountState,
-    epoch_operations: &[UpdateInputData],
+    epoch_operations: &[SnarkAccountUpdateInfo],
+    final_state_root: Hash,
 ) -> Result<()> {
+    if epoch_operations.is_empty() {
+        return Ok(());
```
purusang: I think it should validate committed vs computed state roots regaradless of any operations 🤔

purusang (approved):
Looks good.

bin/strata/src/rpc/node.rs
bewakes: I can see this can be confusing, but the `final_state_root` here is the state root at the end of the epoch. So this is correct. All the other intermediate state roots for each update are named as `new_state_root`. The field docstring in `RpcAccountEpochSummary` indicates this.

storopoli (approved):
ACK e42cfd0be27b839900cb89ea13f869680827cb57

bin/strata/src/rpc/node.rs
barakshani: If all state root are the latest state roots at the time we query for the state root (ie. there is nothing relative), I'd claim we should just call it `state_root`. _new_ and _final_ just give the appearance that things are relative, which imo adds unnecessary confusion. If anything, I'd call it `epoch_state_root`, but I also deem it as unnecessary, if that is not the point.

crates/ol/rpc/types/src/account_summary.rs
barakshani: arguably, you should have dropped the `final`, not adding it to balance.

crates/alpen-ee/ol-tracker/src/task.rs
bewakes: Done!

bin/strata/src/css.rs
bewakes: Yeah, one of us will have to rebase. :)

crates/ol/stf/src/test_utils.rs
bewakes: Yes, we do check that state is reconstructed in `ol/stf/tests/da_epoch_reconstruction.rs` and in `ol/stf/tests/da_epoch_root_correctness.rs`.

We also test proper indexing is done in both modes in `chain-worker-new/str/tests/apply_checkpoint.rs`.

And in the functional tests, we test if an EE node can sync and properly operate using the data received from checkpoint sync node.

crates/snark-acct-runtime/src/program_processing.rs
bewakes: We need to reconstruct something(here state) after verifying the proof because we wouldn't want to reconstruct our state from any data that is not proven to be correct right?

"actual" here most likely indicates the state in DA to be the authoritative one.

Pinging @delbonis here as the original author of the comment.

crates/alpen-ee/common/src/types/ol_account_epoch_summary.rs
bewakes: Done

crates/chain-worker-new/src/state.rs
bewakes: This no longer exists now.

crates/chain-worker-new/src/state.rs
bewakes: Oh yes, this is redundant. We should be using the ssz const. Thanks!

crates/chain-worker-new/src/state.rs
bewakes: Finalized as buried - updated. We never refered to producing a terminal block as finalized as far as I remember.

bin/strata/src/rpc/node.rs
bewakes: Yes, the intermediate structure `Pending` contained the name `final_state_root` instead of `new_state_root`. Renamed the field name correctly.

functional-tests/tests/strata/test_checkpoint_sync_node.py
bewakes: OL state reconstruction is well tested as unit tests. In functional tests we test the functionality from outside - more like the interactions between the bigger components.

This is something that came up in the earlier closed PR as well.

crates/snark-acct-runtime/src/program_processing.rs
delbonis: The snark account's "materialized" inner state.  The OL doesn't care about it, but exec clients do, in order to be able to sync (from DA) properly.

krsnapaudel (approved):
LGTM

functional-tests/tests/strata/test_checkpoint_sync_node.py
barakshani: what is the lift to add this?

I feel this is fundamental. Do we have a unit test that does end-to-end validation?

delbonis (approved):

functional-tests/tests/strata/test_checkpoint_sync_node.py
bewakes: If you mean checking terminal block ids then there is no need for change in RPC responses, for state root checks/whole state checks, we'd need RPC changes.

Regarding your view of end-to-end validation, besides the unit tests that cover both cases, let's take this to DM - will create a ticket if necessary.

Resolving this for now.

PR #1850, opened 2026-05-22
fix(evm-ee): preserve bytecode code-hash keys in chunk witnesses

irnb:
## Description
Accessed-state records already identify bytecodes by the account `code_hash`, but range witness assembly dropped that key and rebuilt the `EvmPartialState` bytecode map by hashing materialized Bytecode values.

That is incorrect when the stored Bytecode contains revm legacy-analysis padding: 
the replay witness can be keyed by `keccak(runtime + padding)` instead of the account's real `keccak(runtime)`, causing chunk replay to miss bytecode that was actually accessed.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/bin/prover-perf/src/programs/alpen_chunk.rs
+++ b/bin/prover-perf/src/programs/alpen_chunk.rs
@@ -53,7 +53,15 @@ pub(super) fn prepare_input() -> EeChunkProofInput {
 
     let pre_state = EvmPartialState::new(
         witness.parent_state.clone(),
-        witness.bytecodes.clone(),
+        // This RSP fixture stores bytecodes as a Vec without original code-hash
+        // keys. Re-hashing keeps the fixture behavior; production range
+        // witnesses preserve the AccessedStateGenerator keys instead.
+        witness
+            .bytecodes
+            .clone()
+            .into_iter()
+            .map(|bytecode| (bytecode.hash_slow(), bytecode))
+            .collect(),
         witness.ancestor_headers.clone(),
     );
 
--- a/crates/evm-ee/src/execution.rs
+++ b/crates/evm-ee/src/execution.rs
@@ -264,11 +264,11 @@ impl BlockAssembler for EvmExecutionEnvironment {
 
 #[cfg(test)]
 mod tests {
-    use std::{fs, path::PathBuf};
+    use std::{collections::BTreeMap, fs, path::PathBuf};
 
     use alloy_consensus::Sealable;
     use reth_primitives_traits::Block as RethBlockTrait;
-    use revm::DatabaseRef;
+    use revm::{DatabaseRef, state::Bytecode};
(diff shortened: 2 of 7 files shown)


storopoli:
This still looks like it can preserve the padded revm representation as the bytecode *value*. The PR fixes the key side by carrying the `AccessedStateGenerator` code hash through range witness assembly, but here we persist `code.bytes().to_vec()` for the bytecode store: https://github.com/alpenlabs/alpen/blob/dd920aaac2d569ed2fb391298a863e485e285960/crates/reth/exex/src/accessed_state_exex.rs#L270-L274

For legacy analyzed bytecode, revm may append zero padding during analysis: https://github.com/bluealloy/revm/blob/main/crates/bytecode/src/legacy/analysis.rs#L42-L47. When the range witness later reconstructs the bytecode with `Bytecode::new_raw(code.into())`, those padded bytes become part of the reconstructed bytecode’s original length: https://github.com/alpenlabs/alpen/blob/dd920aaac2d569ed2fb391298a863e485e285960/crates/reth/witness/src/range_witness_extractor.rs#L203-L218

That means the preserved key can make `code_by_hash_ref(real_hash)` succeed, but replay may still diverge for contracts that observe code size or code bytes (`CODESIZE`, `EXTCODESIZE`, and related copy paths). Should this store `code.original_bytes().to_vec()` instead, so the bytecode value is also reconstructed from the real runtime bytes while this PR preserves the original code-hash key?

irnb:
@storopoli please check https://github.com/alpenlabs/alpen/pull/1850/changes/b9c92bff0e1e1e68f35409c7b73300d6b56078cd

storopoli (approved):
ACK f478b7262903463df2d9da87be8a50f02a0e4fc3

prajwolrg (approved):

prajwolrg:
@codex review

purusang (approved):
@irnb looks like we have two bugs, one of them is being fixed by this pr and second one about missing bytecode during proving if some bytecode added in some block and accessed in the same. If you could fix both of them in this one then we could verify that by ensuring the sp1 e2e ci workflow passes by modifying the ci.

irnb:
> looks like we have two bugs, one of them is being fixed by this pr and second one about missing bytecode during proving if some bytecode added in some block and accessed in the same.

@purusang there aren't two bugs, the main issue was `keccak(runtime + padding)`, and because of it we faced the "missing bytecode" issue. this fix is sufficient to solve it.

sapinb (approved):
ACK f478b7262903463df2d9da87be8a50f02a0e4fc3

We should add a test for this case if possible to prevent regression.

PR #1852, opened 2026-05-22
doc(ol,asm): OL/ASM withdrawal/deposits/assignments logs

storopoli:
## Description

More ASM/OL logs on deposit/withdrawals.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update
- [x] Logs

## Notes to Reviewers

Sister PR to https://github.com/alpenlabs/asm/pull/110

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-1405 and STR-3557

### DIFF
--- a/bin/alpen-client/src/payload_builder.rs
+++ b/bin/alpen-client/src/payload_builder.rs
@@ -15,7 +15,7 @@ use alpen_reth_node::{
 use eyre::{eyre, Context};
 use reth_node_builder::{ConsensusEngineHandle, PayloadBuilderAttributes};
 use reth_payload_builder::PayloadBuilderHandle;
-use tracing::{debug, warn};
+use tracing::{debug, info, warn};
 
 #[derive(Debug)]
 pub(crate) struct AlpenRethPayloadEngine {
@@ -76,6 +76,15 @@ impl PayloadBuilderEngine for AlpenRethPayloadEngine {
                 })
             })
             .collect::<Result<Vec<Withdrawal>, _>>()?;
+        for (deposit_index, withdrawal) in withdrawals.iter().enumerate() {
+            info!(
+                %parent,
+                deposit_index,
+                address = %withdrawal.address,
+                amount_gwei = withdrawal.amount,
+                "prepared deposit mint for payload attributes",
+            );
+        }
         let payload_attrs = AlpenPayloadAttributes::new_from_eth(PayloadAttributes {
             timestamp: build_attrs.timestamp(),
             // IMPORTANT: post cancun payload build will fail without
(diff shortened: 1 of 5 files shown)


delbonis (approved):
Wondering if any of these could be lowered to debug/trace, but looks good otherwise.

bin/alpen-client/src/payload_builder.rs
```diff
@@ -76,6 +76,15 @@ impl PayloadBuilderEngine for AlpenRethPayloadEngine {
...
             })
             .collect::<Result<Vec<Withdrawal>, _>>()?;
+        for (deposit_index, withdrawal) in withdrawals.iter().enumerate() {
+            info!(
+                %parent,
+                deposit_index,
+                address = %withdrawal.address,
+                amount_gwei = withdrawal.amount,
+                "prepared deposit mint for payload attributes",
+            );
```
barakshani, line 86: I assume this _withdrawals_ is Eth beacon chain withdrawals, so our deposits. If so, why `withdrawal.address` and not `deposit.address` (i.e. `withdrawal` --> `deposit`). Also curious, where do we get `deposit_index` at this stage? Is this necessarily the same deposit_index as in the bridge, or internal to reth? (or is it just 'local' indexing within this block?)
On the latter, just in case, see https://alpenlabs.atlassian.net/browse/STR-3567: the report says that the issue is in the RPC, but maybe it's deeper than that. (maybe we have some indexing issue at this level)

barakshani (commented):
Perfect, thanks!

@voidash is this the right place to add some "observability" aspects? I'm sure that at some point we'd like to track all minted funds and withdrawals.

bin/alpen-client/src/payload_builder.rs
bewakes: Yeah, using both `deposit_index` and `withdrawal` as reference is confusing. This should either be both indicating withdrawals(from evm's perspective) or both deposits(from Alpen's perspective) and clearly commented.

bewakes (approved):
Thanks for the logs! Agree with barak on the withdrawal/depsit namings.

prajwolrg (approved):

bin/alpen-client/src/payload_builder.rs
storopoli: The `deposit_index` is the index from the `withdrawals` container. `enumerate()` returns a tuple `(index, value)`.

barakshani:
@storopoli can we log the block slot as well? we do log the parent block hash, so it's not that critical.

PR #1855, opened 2026-05-25
fix(ci): prevent transitive skip propagation in ci-build

purusang:
## Summary

- On `workflow_dispatch`, `check-new-commits` is skipped (schedule-only). `prepare` and `build-sp1-artifacts` handle this with `if: always()`, but `build-and-push` and `dispatch-cd` inherited the transitive skip since they lacked the guard.
- Add explicit `if: always()` with success checks on direct dependencies so downstream jobs run when their inputs succeed regardless of upstream skips.

## Test plan

- [x] Trigger `ci-build.yml` via `workflow_dispatch` — `build-and-push` should no longer be skipped
- [ ] Verify scheduled runs still skip when `check-new-commits` says no new commits

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -192,6 +192,7 @@ jobs:
   build-and-push:
     name: "Build ${{ matrix.program }}"
     needs: [prepare, build-sp1-artifacts]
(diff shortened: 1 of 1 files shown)


purusang:
PoC for this pr https://github.com/alpenlabs/alpen/actions/runs/26385435693

alexhui01 (approved):

PR #1858, opened 2026-05-25
fix unbound decode allocation

sapinb:
## Description
Multiple decode paths allocate memory based on untrusted length values before validating input size. A crafted input with a large length prefix can trigger massive allocations (up to ~4 GiB from a single u32), causing OOM before any data is actually read.

Changes
- codec-utils: Cap CodecSsz and CodecBorsh decode allocations at 16 MiB
- da-framework: Add 1M entry cap to all collection encode/decode helpers (decode_vec, decode_map, decode_map_with, decode_vec_with and their encode counterparts)
- reth-statediff: Cap CodecBytes::decode at 2 MiB
- acct-types: Move MsgPayload::decode length check before allocation (the bound existed but was checked after vec![])
- codec: Tighten VarVec BOUND generics at declaration sites (WithdrawalMsgData, SimpleWithdrawalIntentLogData, SnarkAccountUpdateLogData, SubjTransferMsgData)

Note: Instead of sprinkling bound checks across all length-prefix serialization, should we consider add max bounds in the types themselves, similar to VarVec, etc and have the types explicitly decide max bounds, for all collection types we expect to use ? Adding checks just during serialization can potentially be incompatible with other parts of the code that create them, like block builder, etc.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
- VarVec from strata-common has a default bound of u32::MAX. This has been overridden in usages here, but we should consider if we should start with safer default instead. @delbonis ?
- Please check the max limits added here, if they are incompatible for our use cases.

This PR does not make changes in similar patterns used in the following locations:
- `evm-ee` codec impls: only used internally
-   `U16LenList` : type limits it to 65k entries, and a better approach would be to add const generic for max allowed size similar to VarVec
- `SimpleTransaction`: only used inside tests
- `alpen_reth_statediff::storage::BatchStateDiff` / `alpen_reth_statediff::storage::StorageDiff` : Not really possible to enforce meaningful limits without making more significant changes. Limits here will also need to be either large enough for any realistic batches, or factored into batch sealing logic. Will handle this one separately.
 
AI usage: refactors and change reviews.

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/codec-utils/src/borsh_shim.rs
+++ b/crates/codec-utils/src/borsh_shim.rs
@@ -3,6 +3,9 @@
 use borsh::{BorshDeserialize, BorshSerialize};
 use strata_codec::{Codec, CodecError, Decoder, Encoder, Varint};
 
+/// Maximum byte length for a single Borsh-encoded object (16 MiB).
+const MAX_BORSH_OBJECT_BYTES: usize = 16 << 20;
+
 /// Wraps a Borsh type so that it can be transparently [`Codec`]ed.
 #[derive(Clone, Debug, Hash, Eq, PartialEq, Ord, PartialOrd, BorshSerialize, BorshDeserialize)]
 pub struct CodecBorsh<T: BorshSerialize + BorshDeserialize>(pub T);
@@ -27,6 +30,10 @@ impl<T: BorshSerialize + BorshDeserialize> Codec for CodecBorsh<T> {
         let len = Varint::decode(dec)?;
         let len_usize = len.inner() as usize;
 
+        if len_usize > MAX_BORSH_OBJECT_BYTES {
+            return Err(CodecError::OverflowContainer);
+        }
+
         // Read a buffer of that size.
         let mut buffer = vec![0u8; len_usize];
         dec.read_buf(&mut buffer)?;
@@ -40,6 +47,10 @@ impl<T: BorshSerialize + BorshDeserialize> Codec for CodecBorsh<T> {
     fn encode(&self, enc: &mut impl Encoder) -> Result<(), CodecError> {
         // Encode convert the inner value to Borsh.
         let bytes = borsh::to_vec(&self.0).map_err(|_| CodecError::MalformedField("borsh"))?;
+
+        if bytes.len() > MAX_BORSH_OBJECT_BYTES {
+            return Err(CodecError::OverflowContainer);
+        }
 
         // First encode the length as a varint.
         let len = Varint::new_usize(bytes.len()).ok_or(CodecError::OverflowContainer)?;
--- a/crates/codec-utils/src/ssz_shim.rs
+++ b/crates/codec-utils/src/ssz_shim.rs
@@ -4,6 +4,9 @@ use ssz::{Decode, Encode};
 use ssz_derive::{Decode, Encode};
 use strata_codec::{Codec, CodecError, Decoder, Encoder, Varint};
 
+/// Maximum byte length for a single SSZ-encoded object (16 MiB).
+const MAX_SSZ_OBJECT_BYTES: usize = 16 << 20;
+
 /// Wraps an SSZ type so that it can be transparently [`Codec`]ed.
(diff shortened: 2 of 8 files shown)


sapinb:
@codex review please

sapinb:
@codex re-review please

crates/codec-utils/src/borsh_shim.rs
```diff
@@ -3,6 +3,9 @@
 use borsh::{BorshDeserialize, BorshSerialize};
 use strata_codec::{Codec, CodecError, Decoder, Encoder, Varint};
 
+/// Maximum byte length for a single Borsh-encoded object (16 MiB).
+const MAX_BORSH_OBJECT_BYTES: usize = 16 << 20;
```
prajwolrg, line 7: Curious why are we capping it at 16 MB? and not 32 MB? or other values.

crates/codec-utils/src/ssz_shim.rs
```diff
@@ -4,6 +4,9 @@ use ssz::{Decode, Encode};
 use ssz_derive::{Decode, Encode};
 use strata_codec::{Codec, CodecError, Decoder, Encoder, Varint};
 
+/// Maximum byte length for a single SSZ-encoded object (16 MiB).
+const MAX_SSZ_OBJECT_BYTES: usize = 16 << 20;
+
```
prajwolrg, line 9: Same as above.

prajwolrg (approved):
The changes themselves looks good. I'm curious about how the limits are chosen.

sapinb:
@codex re-review please

crates/codec-utils/src/borsh_shim.rs
sapinb: Some are pretty arbitrary, including this one. I'm hoping for recommendations on these limits in this PR 😄

crates/ol/msg-types/src/withdrawal.rs
```diff
@@ -34,15 +34,14 @@ pub struct WithdrawalMsgData {
     selected_operator: u32,
 
     /// Bitcoin Output Script Descriptor describing the withdrawal output.
-    // TODO idk why, but I can't make the MAX_WITHDRAWAL_DESC_LEN const generic work
-    dest_desc: VarVec<u8>,
+    dest_desc: VarVec<u8, { MAX_WITHDRAWAL_DESC_LEN }>,
```
delbonis: Maybe make a `type DestDescBufVec = VarVec<u8, { ... }>;` alias here since we also use it in a return type.

crates/ol/chain-types/src/log_payloads.rs
```diff
@@ -54,7 +61,7 @@ pub struct SnarkAccountUpdateLogData {
     pub new_msg_idx: u64,
 
     /// Extra data from the update operation.
-    pub extra_data: VarVec<u8>,
+    pub extra_data: VarVec<u8, { MAX_EXTRA_DATA_BYTES }>,
```
delbonis: Also would be good to make a type alias here too.

crates/ol/chain-types/src/log_payloads.rs
```diff
@@ -14,7 +21,7 @@ pub struct SimpleWithdrawalIntentLogData {
     pub amt: u64,
 
     /// Destination BOSD.
-    pub dest: VarVec<u8>,
+    pub dest: VarVec<u8, { MAX_DEST_BYTES }>,
```
delbonis: and here, maybe reusing the one from the other place

delbonis (changes requested):
Just some minor cleanliness suggestions.

storopoli (approved):
ACK a619fd7b28e32cbaab10e741466a70a5106393d9

bewakes (approved):
LGTM. Concur with trey's suggestions and prajwol's questions.

delbonis (approved):

storopoli (approved):
ACK 8d58d6c1a4712f7e5f169b75fc0fa45095830d26

PR #1832, opened 2026-05-18
STR-3155: bind snark account updates to specific seqnos

delbonis:
## Description

This PR fixes an audit finding by adding a `seq_no` field to the snark account update proof interface, binding the proofs to a specific update seqno.

Claude did most of the low-level work with me guiding it to minimize change blast radius.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Is this PR addressing any specification, design doc or external reference document?

- [ ]  Yes
- [x]  No

If yes, please add relevant links:

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [x] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [x] New and existing tests pass with my changes.
- [x] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

PR #1862, opened 2026-05-26
EE DA Inclusion Proof & OL MMR refactor

irnb:
## Summary
- Adds EE DA witness plumbing so acct proofs verify raw Bitcoin DA commit/reveal tx inclusion against public L1 block refs.
- Replaces acct `LedgerRefs` from ASM-manifest claims to reduced `{block_hash, wtxids_root}` L1 block-ref claims, with OL MMR/indexing support.
- Binds the DA blob to the proven EE batch via update `seq_no`, final chunk state root, EVM header summary, partial pre-state application, and bytecode checks.
- Adds host-side DA witness assembly from batch DA refs + Bitcoin block data, plus payload-builder retry handling for temporarily missing Reth parents.

### All EE DA–related components are now aggregated into three focused crates:
`alpen-ee-da-types`: DA types (DaBlob, DaWitness, etc.)
`alpen-ee-da-runtime`: host witness builder + guest verifier logic
`alpen-ee-da-provider` : DaBlob provider/integration layer

## Reviewer Notes
Review this in layers:
1. Public/protocol surfaces: `UpdateProofPubParams.seq_no`, `ChunkTransition.tip_state_root`, `tip_exec_header_summary`, and `LedgerRefs.l1_block_refs`.
2. OL binding: L1 block-ref MMR updates, proof indexing, and verifier rename from ASM manifest refs to L1 block refs.
3. Host witness path: Bitcoin block fetch, `wtxids_root` recomputation, tx inclusion proof construction, DA blob reassembly, and known-bytecode witness handling.
4. Guest verification: reduced ref membership, wtxid Merkle proofs, commit/reveal ordering, DA blob metadata, state-diff application, and bytecode hash checks.

## Breaking Changes
- SSZ/proof interfaces change: `UpdateProofPubParams`, `ChunkTransition`, `LedgerRefs`, `LedgerRefProofs`, and OL epochal state are not compatible with old serialized proof inputs/state.
- `EeAcctProofInput` now includes a DA witness; old acct proof input builders/perf fixtures need the new third private input.
- Non-empty acct batches now require DA witness data and partial pre-state; empty DA witness is only valid for empty/no-chunk cases.
- `L1DaBlockRef` now carries `L1DaBlockInfo { commitment, wtxids_root }`, not only `L1BlockCommitment`.

## Issues Resolved
- STR-1369 proof path no longer trusts host-selected DA bytes alone; the guest derives and verifies the DA blob from included Bitcoin transactions.
- Ledger refs now expose the Bitcoin-shaped commitment EE proofs need without pulling raw ASM manifest preimages into the EE proof boundary.
- Fixes payload build races where Reth has not imported the parent yet.
- Tightens functional test defaults around EE block wait timing and DA magic bytes.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

- [x]  Yes
- [ ]  No

- https://alpen-labs.slack.com/archives/C08G7PL1RBK/p1779696709736859
- https://alpen-labs.slack.com/archives/C08G7PL1RBK/p1779179252335769
- https://alpen-labs.slack.com/archives/C0B4MDUQDRV/p1779359008419229

## Checklist

## Related Issues

STR-1369


irnb:
@codex review

irnb:
@codex review

irnb:
sp1 e2e test passed after adding the OL `submit_rpc_port` in `.github/sp1-e2e/config.seq.toml`

https://github.com/alpenlabs/alpen/actions/runs/26493125550
https://github.com/alpenlabs/alpen/actions/runs/26495900004

crates/ol/state-support-types/src/write_tracking_layer.rs
```diff
@@ -236,23 +236,25 @@ where
...
+            .l1_block_refs_mmr
             .clone()
-            .unwrap_or_else(|| self.base.asm_manifests_mmr().clone());
+            .unwrap_or_else(|| self.base.l1_block_refs_mmr().clone());
 
-        use strata_acct_types::{StrataHasher, tree_hash::TreeHash};
-        let manifest_hash = <AsmManifest as TreeHash>::tree_hash_root(&mf);
-        strata_merkle::Mmr::<StrataHasher>::add_leaf(&mut mmr, manifest_hash.into_inner())
+        use strata_acct_types::StrataHasher;
+        use strata_snark_acct_types::l1_block_ref_leaf_hash;
```
bewakes: nit: use imports at top.

crates/snark-acct-types/src/update.rs
```diff
@@ -187,3 +222,30 @@ impl UpdateAccumulatorProofs {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn l1_block_ref_leaf_hash_matches_pinned_root() {
+        let block_hash = [1u8; 32];
+        let wtxids_root = [2u8; 32];
+        let expected = [
```
bewakes: can you briefly state how this value was obtained?

crates/alpen-ee/da/types/src/witness.rs
```diff
@@ -0,0 +1,323 @@
...
+}
+
+/// Bitcoin Merkle inclusion proof.
+///
+/// `siblings` is ordered leaf-first. `position` is the leaf index in the
+/// bottom layer; bit `i` selects whether `siblings[i]` is on the left or right
+/// of the running hash at level `i`.
+#[derive(Clone, Debug, Default, Archive, Deserialize, Serialize)]
+#[rkyv(derive(Debug))]
+pub struct BitcoinMerkleProof {
```
bewakes, line 181: There's a `MerkleProofB32` in `strata_merkle` which looks very similar in shape to this. Maybe that can be used?

crates/chain-worker-new/src/context.rs
```diff
@@ -427,14 +438,90 @@ fn index_inbox_mmr_writes(
...
+/// append into the DB-side proof index. The operation is idempotent for
+/// crash-restart retries.
+fn index_l1_block_ref_mmr_writes(
+    mmr_index_mgr: &MmrIndexManager,
+    output: &OLBlockExecutionOutput,
+    genesis_l1_height: u64,
+) -> WorkerResult<()> {
+    let handle = mmr_index_mgr.get_handle(MmrId::L1BlockRefs);
+    let prefill_count = genesis_l1_height
+        .checked_add(1)
```
bewakes: nit: You can just do `leaf_count..=prefill_count` below instead of doing this checked add here. This seems to just add confusion and non-required error handling.

crates/chain-worker-new/src/context.rs
```diff
@@ -427,14 +438,90 @@ fn index_inbox_mmr_writes(
...
+            )));
+        }
+    }
+
+    for write in output.indexer_writes().manifests() {
+        let expected_idx = write.height as u64;
+        let block_hash = *write.manifest.blkid().as_ref();
+        let wtxids_root = *write.manifest.wtxids_root().as_ref();
+        let expected_hash: Hash = l1_block_ref_leaf_hash(&block_hash, &wtxids_root).into();
+        let preimage = L1BlockRef::new(block_hash, wtxids_root).as_ssz_bytes();
```
bewakes: Shouldn't the hash be computed from preimage?

crates/chain-worker-new/src/context.rs
```diff
@@ -427,14 +438,90 @@ fn index_inbox_mmr_writes(
...
+            continue;
+        }
+
+        if expected_idx > leaf_count {
+            return Err(DbError::MmrIndexOutOfRange {
+                requested: expected_idx,
+                cur: leaf_count,
+            }
+            .into());
+        }
```
bewakes: Feels like these can be abstracted from the db/manager layer and not out here where it is being used. This is true for every mmr entry right, not only l1_ref mmrs.

bewakes (commented):
Neat work! Just looked at OL side code. Some nits and good to haves. Nothing blocking.

delbonis:
This is *ginormous*.  On a brief skim it looks good but I'll have to take a bit to review this all.

bin/alpen-client/src/prover/da_witness_build.rs
```diff
@@ -0,0 +1,417 @@
...
+fn last_commit_reveal_vout(commit: &Transaction) -> u32 {
+    commit
+        .output
+        .iter()
+        .enumerate()
+        .skip(1)
+        .take_while(|(_, output)| output.script_pubkey.is_p2tr())
+        .map(|(idx, _)| idx as u32)
+        .last()
+        .unwrap_or(0)
```
storopoli: This is BEAUTIFUL. ❤️ functional programming

crates/alpen-ee/database/src/serialization_types/batch.rs
```diff
@@ -100,14 +100,17 @@ impl TryFrom<DBBatch> for Batch {
 pub(crate) struct DBL1DaBlockRef {
     /// L1BlockCommitment serialized via its Borsh impl.
     block: L1BlockCommitment,
+    /// Witness transaction Merkle root for the L1 block.
+    wtxids_root: [u8; 32],
```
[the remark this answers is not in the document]
storopoli: We don't care about preserving db layouts. solving these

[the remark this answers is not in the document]
storopoli: Fuck migrations, kthxbye

crates/alpen-ee/da/src/envelope_provider.rs
```diff
@@ -262,16 +290,38 @@ impl ChunkedEnvelopeDaProvider {
...
+        "cannot compute wtxids root for empty block"
+    );
+    // NOTE: DA reveal txs are witness spends. If commit txs are also guaranteed
+    // to spend post-SegWit inputs, every DA block referenced here has witness
+    // data and the witness root matches the L1 block ref root committed by ASM.
+    // Without that funding-input guarantee, commit-only legacy blocks would
+    // require mirroring ASM's fallback to the txid merkle root.
+    let root = block
+        .witness_root()
+        .ok_or_else(|| eyre::eyre!("cannot compute wtxids root for empty block"))?;
```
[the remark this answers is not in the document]
storopoli: We don't care about legacy-only blocks. This is VERY VERY unlikely

crates/proof-impl/alpen-acct/src/da/blob.rs
```diff
@@ -0,0 +1,304 @@
...
+    pub block_num: u64,
+    /// Unix timestamp of the last EVM block.
+    pub timestamp: u64,
+    /// Base fee per gas of the last EVM block.
+    pub base_fee: u64,
+    /// Total gas consumed by the last EVM block.
+    pub gas_used: u64,
+    /// Gas limit of the last EVM block.
+    pub gas_limit: u64,
+}
```
storopoli: WHY?!?!? We have this already in https://github.com/alpenlabs/alpen/blob/5380125c912f7eef876af0a2ef9b49e3e4f708cc/crates/alpen-ee/common/src/types/da.rs?plain=1#L32-L43

crates/proof-impl/alpen-acct/src/da/blob.rs
```diff
@@ -0,0 +1,304 @@
...
+/// DA blob containing batch metadata and state diff.
+#[derive(Debug, Clone, Codec)]
+pub struct DaBlob {
+    /// Monotonic EE account update sequence number for this blob.
+    pub update_seq_no: u64,
+    /// EVM header context of the last block in this batch.
+    pub evm_header: EvmHeaderSummary,
+    /// Aggregated state diff for the batch.
+    pub state_diff: BatchStateDiff,
+}
```
storopoli: NO! We have this already in https://github.com/alpenlabs/alpen/blob/5380125c912f7eef876af0a2ef9b49e3e4f708cc/crates/alpen-ee/common/src/types/da.rs?plain=1#L51-L58

crates/proof-impl/alpen-acct/src/da/blob.rs
```diff
@@ -0,0 +1,304 @@
...
+
+/// Reassembles a [`DaBlob`] from raw chunk payloads.
+pub(super) fn reassemble_da_blob(chunks: &[Vec<u8>]) -> Result<DaBlob, CodecError> {
+    if chunks.is_empty() {
+        return Err(CodecError::MalformedField("no DA chunks provided"));
+    }
+
+    let blob: Vec<u8> = chunks.iter().flat_map(|c| c.iter().copied()).collect();
+    decode_buf_exact(&blob)
+}
```
storopoli: It is already in https://github.com/alpenlabs/alpen/blob/5380125c912f7eef876af0a2ef9b49e3e4f708cc/crates/alpen-ee/common/src/types/da.rs?plain=1#L104-L111

crates/proof-impl/alpen-acct/src/da/blob.rs
```diff
@@ -0,0 +1,304 @@
...
+    commit
+        .output
+        .iter()
+        .enumerate()
+        .skip(1)
+        .take_while(|(_, output)| output.script_pubkey.is_p2tr())
+        .map(|(idx, _)| idx as u32)
+        .last()
+        .unwrap_or(0)
+}
```
storopoli: You are using this function again. You are defined it in this same PR in https://github.com/alpenlabs/alpen/blob/45cfc23d224194ed705c5977572f867e34f307a9/bin/alpen-client/src/prover/da_witness_build.rs#L168-L178

crates/proof-impl/alpen-acct/src/da/blob.rs
```diff
@@ -0,0 +1,304 @@
...
+        return Err(DaVerificationError::MalformedCommitMarker);
+    };
+    if instructions.next().is_some() || push.as_bytes().len() != 8 {
+        return Err(DaVerificationError::MalformedCommitMarker);
+    }
+
+    let mut payload = [0u8; 8];
+    payload.copy_from_slice(push.as_bytes());
+    Ok(Some(payload))
+}
```
storopoli: Same here you've already defined this in this PR: https://github.com/alpenlabs/alpen/blob/45cfc23d224194ed705c5977572f867e34f307a9/bin/alpen-client/src/prover/da_witness_build.rs#L180-L198

crates/proof-impl/alpen-acct/src/da/blob.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    let leaf = input
+        .witness
+        .taproot_leaf_script()
+        .ok_or(DaVerificationError::RevealMissingLeafScript)?;
+    let script = leaf.script.into();
+    let chunk = parse_envelope_payload(&script)?;
+
+    Ok((vout, chunk))
+}
```
storopoli: Same here you've already defined this in this PR: https://github.com/alpenlabs/alpen/blob/45cfc23d224194ed705c5977572f867e34f307a9/bin/alpen-client/src/prover/da_witness_build.rs#L200-L221

crates/proof-impl/alpen-acct/src/da/inclusion.rs
```diff
@@ -0,0 +1,187 @@
...
+        root = if position & 1 == 0 {
+            bitcoin_hash_pair(root, *sibling)
+        } else {
+            bitcoin_hash_pair(*sibling, root)
+        };
+        position >>= 1;
+    }
+
+    root
+}
```
storopoli: Same here you've already defined this in this PR: https://github.com/alpenlabs/alpen/blob/45cfc23d224194ed705c5977572f867e34f307a9/bin/alpen-client/src/prover/da_witness_build.rs#L257-L276

functional-tests/common/config/constants.py
```diff
@@ -39,6 +39,10 @@
 # EE Timing
 # =============================================================================
 DEFAULT_EE_BLOCK_TIME_MS = 1_000
+# Debug Reth payload import can spend several seconds validating state roots
+# after DA/prover tasks are enabled, so wait budgets are intentionally looser
+# than the sequencer's target block cadence.
+DEFAULT_EE_BLOCK_WAIT_SECONDS = 20.0
```
storopoli: We are bumping the functional tests run time by 20x. WHY?!
We should use a `wait_until` function instead.

functional-tests/tests/alpen_client/test_fullnode_sync.py
```diff
@@ -36,8 +88,8 @@ def main(self, ctx):
 
         # Wait for initial sync
         logger.info("Waiting for initial sync...")
-        ee_sequencer.wait_for_peers(1, timeout=15)
-        ee_fullnode_0.wait_for_peers(1, timeout=15)
+        ee_sequencer.wait_for_peers(1, timeout=60)
+        ee_fullnode_0.wait_for_peers(1, timeout=60)
```
storopoli: Are you sure we need this? Why?

crates/proof-impl/alpen-acct/src/da/blob.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    // NOTE: known bytecodes are private reconstruction witness data for bytecodes
+    // deduped from the current L1 blob. The guest accepts them only after
+    // recomputing the EVM code hash, which proves identity for the account diff
+    // but not prior L1 publication.
+    //
+    // TODO(STR-1907): replace this with membership in an authenticated
+    // published-bytecode set, or explicit inclusion in the earlier DA blob that
+    // first carried the bytes.
+    for bytecode in known_bytecodes {
```
storopoli: This still accepts bytecode from an unauthenticated private witness. `known_bytecodes` are not part of the L1 DA blob and are only checked by re-hashing the bytes here, so the proof can pass for a state diff whose `code_hash` is not backed by bytecode in current DA or by a prior-publication proof. That weakens the DA guarantee because L1 reconstruction can miss the bytecode while the acct proof still accepts it.

Permalinks: implementation accepts the private bytes here: https://github.com/alpenlabs/alpen/blob/45cfc23d224194ed705c5977572f867e34f307a9/crates/proof-impl/alpen-acct/src/da/blob.rs?plain=1#L230-L238, and the test demonstrates the bypass shape by accepting a blob with no deployed bytecode when `DaBytecodeWitness` supplies it privately: https://github.com/alpenlabs/alpen/blob/45cfc23d224194ed705c5977572f867e34f307a9/crates/proof-impl/alpen-acct/src/da/tests.rs?plain=1#L210-L298. Until STR-1907 lands, this should require the bytecode in the current DA blob or include an authenticated prior-publication proof.

storopoli (changes requested):

bin/alpen-client/src/prover/da_witness_build.rs
```diff
@@ -0,0 +1,417 @@
...
+        .iter()
+        .enumerate()
+        .map(|(i, tx)| {
+            if i == 0 {
+                [0u8; 32]
+            } else {
+                tx.compute_wtxid().to_byte_array()
+            }
+        })
+        .collect();
```
krsnapaudel: This seems repeated in the next function: `compute_wtxids_root`.

crates/proof-impl/alpen-acct/src/da/blob.rs
```diff
@@ -0,0 +1,304 @@
...
+        return Err(DaVerificationError::CommitVersionMismatch {
+            expected: DA_BLOB_VERSION,
+            actual: actual_version,
+        });
+    }
+
+    Ok(())
+}
+
+fn commit_marker_payload(commit: &Transaction) -> Result<Option<[u8; 8]>, DaVerificationError> {
```
krsnapaudel: Summarizing @storopoli’s duplication comments: these are mostly semantic duplicates, not exact copies, but they sit on the proof boundary. Host and guest independently handle the same DA blob/header layout, commit-reveal parsing, and Merkle rules, so drift can cause proof failures or disagreement over the same L1 bytes.

Does it make sense to move the pure shared pieces into a small crate usable by both host and proof guest, e.g. `crates/ee-da-types` or `crates/alpen-ee/da-types`?

crates/chain-worker-new/src/context.rs
```diff
@@ -179,13 +180,23 @@ impl ChainWorkerContext for ChainWorkerContextImpl {
         {
             Ok(()) => {
                 index_inbox_mmr_writes(&self.mmr_index_mgr, output)?;
+                index_l1_block_ref_mmr_writes(
+                    &self.mmr_index_mgr,
+                    output,
+                    self.params.rollup.genesis_l1_view.blk.height() as u64,
```
krsnapaudel: The OL state’s L1 block refs MMR is prefilled from `OLParams.last_l1_block.height()`, while this local DB proof index is prefilled from `params.rollup.genesis_l1_view.blk.height()`. Are those guaranteed to be identical? Does it make sense to use the same source, i.e. `OLParams` here?

Similar comment for 
`self.params.rollup.genesis_l1_view` used below.

crates/chain-worker-new/src/context.rs
```diff
@@ -179,13 +180,23 @@ impl ChainWorkerContext for ChainWorkerContextImpl {
...
             Err(DbError::BlockIndexingConflict {
                 attempted,
                 last_applied,
                 ..
             }) if attempted == commitment && last_applied == commitment => {
                 index_inbox_mmr_writes(&self.mmr_index_mgr, output)?;
+                index_l1_block_ref_mmr_writes(
+                    &self.mmr_index_mgr,
+                    output,
+                    self.params.rollup.genesis_l1_view.blk.height() as u64,
```
krsnapaudel: Here

crates/proof-impl/alpen-acct/src/da/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+use self::{
+    blob::{
+        extract_da_chunks, reassemble_da_blob, verify_da_blob_metadata,
+        verify_state_diff_against_chunks,
+    },
+    inclusion::verify_block_witness,
+};
+
+/// Runs DA correctness checks for the outer proof.
+pub fn verify_da_witness(
```
krsnapaudel: This returns `Option<DaBlob>`, but callers only check success. Is the returned blob intended for future use?

crates/alpen-ee/sequencer/src/update_submitter/update_builder.rs
```diff
@@ -20,7 +20,7 @@ pub(super) async fn build_update_from_batch(
     batch: &Batch,
     da_refs: &[L1DaBlockRef],
     proof_id: &ProofId,
-    ol_client: &(impl SequencerOLClient + Send + Sync),
+    _ol_client: &(impl SequencerOLClient + Send + Sync),
```
krsnapaudel: Is there a reason to keep it around?

krsnapaudel (commented):
Great work overall. 

1. Test coverage suggestion: `verify_da_witness` has good happy-path and inclusion negative coverage, but several binding failures are not exercised end-to-end yet: seq-no mismatch, EVM header mismatch, state-root mismatch, missing/duplicate reveal, and reveal spending vout 0. Should we add a few focused negative tests?
2. Also does it make sense to rename `SauTxLedgerRefs.asm_history_proofs` to `l1_block_refs`?

More comments inline.

crates/alpen-ee/da/types/src/witness.rs
irnb: good catch; they are structurally identical. i left it as a separate type because `MerkleProofB32` is SSZ-only, while the rest of `DaWitness` is rkyv-encoded with `Archived*` zero-copy access. swapping in `MerkleProofB32` would mean either forking `strata-merkle` to add rkyv derives, or migrating the whole witness graph off rkyv. 

on top of that, `MerkleProofB32::verify` uses single-SHA-256 without odd-node duplication, so we'd still need a Bitcoin-specific verifier anyway, the only thing actually reused would be the two-field container.
https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#commitment-structure

supportive of moving this to strata-common later though.

crates/proof-impl/alpen-acct/src/da/blob.rs
irnb: This is a valid soundness gap and it's not the only one we have in the EE batch proof.

to fix it in this PR i'd have to either rip out the bytecode dedup flow or fix the soundness now, and i don't want to do either since both are out of scope for this effort. we'll tackle these soundness gaps later.

https://alpen-labs.slack.com/archives/C060N6GHHAL/p1779876511512249?thread_ts=1779872538.533149&cid=C060N6GHHAL

https://alpenlabs.atlassian.net/browse/STR-1907?search_id=d2b2c397-0556-4c96-93c4-1d3e5e6601f1

irnb:
thanks guys for the reviews 

@bewakes all your concerns are addressed in this commit: https://github.com/alpenlabs/alpen/pull/1862/changes/c9aee5ea42ce81587ec22b80ea3888b43bd72d21

@storopoli @krsnapaudel i was already working on a restructure to resolve the duplication, thanks for flagging it. resolved here: https://github.com/alpenlabs/alpen/pull/1862/changes/83bb517adb6c6571b03f993284cff70d29923d2c

@krsnapaudel great catch on the prefilling logic, thanks. resolved here: https://github.com/alpenlabs/alpen/pull/1862/changes/ae79a2c69958a45ab4d310efc1f7d432f5775ff2

<br>

> Also does it make sense to rename `SauTxLedgerRefs.asm_history_proofs` to `l1_block_refs`

yeah, since the OL MMR isn't based on the asm manifest anymore. will do it in the next commit.

crates/alpen-ee/da/types/src/witness.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    #[test]
+    fn da_witness_empty_roundtrips_through_rkyv() {
+        let witness = DaWitness::empty();
+
+        let bytes = rkyv::to_bytes::<RkyvError>(&witness).unwrap();
+        let archived = rkyv::access::<ArchivedDaWitness, RkyvError>(&bytes).unwrap();
+
+        assert!(archived.blocks().is_empty());
+    }
```
prajwolrg, line 228: nit: I wonder if the tests are really necessary since we are deriving `Archive`, and not manually implementing it.

crates/proof-impl/alpen-acct/src/da_inclusion.rs
prajwolrg, line 1: nit: it might be worth having function here where we generate the BitcoinMerkleProof and then verifying it, so it's better for visibility.

crates/snark-acct-types/src/update.rs
```diff
@@ -1,12 +1,45 @@
...
+            .expect("snark-acct-types: FixedBytes<32> is always 32 bytes")
+    }
+
+    /// Gets the block witness transaction Merkle root.
+    pub fn wtxids_root(&self) -> [u8; 32] {
+        self.wtxids_root
+            .as_ref()
+            .try_into()
+            .expect("snark-acct-types: FixedBytes<32> is always 32 bytes")
+    }
```
prajwolrg: nit: might make sense to use `RBuf32`. Not sure if we already have `strata-identifiers` as a dependency.

bin/alpen-client/src/prover/da_witness_build.rs
```diff
@@ -0,0 +1,417 @@
...
+    merkle_root(&leaves)
+}
```
prajwolrg: There's some repitition here between two functions which can be avoided.

prajwolrg: Also I feel like the generation and verification should be together somewhere.

crates/proof-impl/alpen-acct/src/da/inclusion.rs
prajwolrg: This I felt was also described somewhere else.

prajwolrg (approved):
LGTM! Great work! All my comments are nits, so approving this.

Note: I reviewed commit-by-commit so some of the comments below already seems to be addressed in the later commits.

krsnapaudel:
> > Also does it make sense to rename `SauTxLedgerRefs.asm_history_proofs` to `l1_block_refs`
> 
> yeah, since the OL MMR isn't based on the asm manifest anymore. will do it in the next commit.

@irnb The renaming is not done last time I checked.

bin/alpen-client/src/main.rs
```diff
@@ -712,11 +713,20 @@ fn main() {
                 .receipt_hook(ChunkReceiptHook::new(batch_storage_dyn.clone()))
                 .retry(RetryConfig::default());
 
+                let acct_range_witness_fn: Arc<AcctRangeWitnessFn> = {
+                    let extractor = range_witness_extractor.clone();
```
irnb: if this is service we firing it up from main in very wired approach 

first is this standard in this binary?
second there is no more generic executor that related to da or these stuff fire this witness generation or evven sequencer

crates/alpen-ee/da/src/verification/error.rs
```diff
@@ -0,0 +1,84 @@
...
+    PartialPreStateRootMismatch {
+        expected: [u8; 32],
+        actual: [u8; 32],
+    },
+    #[error("state-diff apply failed: {0}")]
+    StateDiffApply(#[from] ReconstructError),
+    #[error(
+        "post-apply state root does not match last chunk's tip_state_root: computed={computed:?}, expected={expected:?}"
+    )]
+    StateRootMismatch {
```
irnb: again didn't get the difference of StateDiffApply tjhs two maybe we need unified erro naming for ease of context

bin/alpen-client/src/prover/da_witness_build.rs
delbonis: definitely should try and organize this logic as a standalone component

bin/alpen-client/src/prover/spec_acct.rs
```diff
@@ -148,7 +169,9 @@ pub(crate) struct AcctSpec {
     chunk_receipts: Arc<dyn ReceiptStore>,
     batch_storage: Arc<dyn BatchStorage>,
     storage: Arc<EeNodeStorage>,
-    ol_client: Arc<dyn SequencerOLClient + Send + Sync>,
+    btc_client: Arc<BtcClient>,
+    state_diff_provider: Arc<dyn StateDiffProvider + Send + Sync>,
```
delbonis: Add the `Sync + Send` bounds to `StateDiffProvider` directly so we don't have to add it everywhere it's used like this.

bin/alpen-client/src/prover/spec_acct.rs
```diff
@@ -266,17 +296,28 @@ impl ProofSpec for AcctSpec {
...
+        let range_data =
+            task::spawn_blocking(move || (range_witness_fn)(first_block_hash, last_block_hash))
+                .await
+                .map_err(|e| {
+                    PaasError::TransientFailure(format!("batch witness extraction join: {e}"))
+                })?
+                .map_err(|e| {
+                    PaasError::TransientFailure(format!("batch witness extraction: {e}"))
+                })?;
+
```
delbonis, line 311: Can this and the other errors be converted to specific error variants so we don't lose the underlying error information?

(also applicable most places where `PaasError` are defined, this might not actually make sense to do, I'm not 100% sure how this error type is intended to be used downstream)

bin/alpen-client/src/prover/spec_acct.rs
```diff
@@ -423,6 +467,133 @@ fn da_refs_from_status(batch_id: BatchId, status: BatchStatus) -> ProverResult<V
     }
 }
 
+/// Builds raw DA transaction witnesses from the L1 blocks referenced by batch DA refs.
+async fn build_da_witness(
+    da_refs: &[L1DaBlockRef],
+    batch_block_hashes: &[Hash],
+    btc: &(impl Reader + Sync),
+    state_diff_provider: &(impl StateDiffProvider + Sync + ?Sized),
```
delbonis: Good places where merging the type constraints into the def would simplify downstream code.

bin/alpen-client/src/prover/spec_acct.rs
delbonis, line 1: I may be misunderstanding the design of this module but I see a red flag I wanted to raise.

It seems like we're merging multiple layers of concern here, in particular talking about bytecode stores at the same times we're walking over blocks looking for transactions.  I would expect that these would be separate layers, like that we'd talk about arbitrary byte blobs in transactions and be making proofs for those.  At the interface, we'd do some conversions between byte blobs that we've already established as being published and decode them into things like bytecode hashes we care about.

For example, it seems like it'd be harder to reuse the low-level indexing code if we wanted to migrate to a different execution environment design, since it's directly talking about EVM concepts.

This is a very abstract critique, as I have not deeply investigated the dataflow, so *maybe* this makes sense to be doing here and I am just not understanding it.

bin/alpen-client/src/payload_builder.rs
delbonis, line 1: May be able to use logging spans to simplify the logging, especially for `parent`, `timestamp`, `retry`, etc.

crates/alpen-ee/da-types/src/blob.rs
```diff
@@ -0,0 +1,77 @@
...
+///
+/// `chunks` must be in commit-output order.
+pub fn reassemble_da_blob(chunks: &[Vec<u8>]) -> Result<DaBlob, CodecError> {
+    if chunks.is_empty() {
+        return Err(CodecError::MalformedField("no DA chunks provided"));
+    }
+
+    let blob: Vec<u8> = chunks.iter().flat_map(|c| c.iter().copied()).collect();
+    decode_buf_exact(&blob)
+}
```
delbonis: I *think* it should be possible to write a decoder impl that walks over the slices directly instead of having to copy them into a contiguous buffer first.  That's just a requirement for being able to call `decode_buf_exact`.

Also, can probably make this a little more generic?

crates/alpen-ee/da-types/src/blob.rs
```diff
@@ -0,0 +1,77 @@
+//! DA codec types and format constants shared between producer and verifier.
+
+use alpen_reth_statediff::BatchStateDiff;
+use strata_codec::{decode_buf_exact, Codec, CodecError};
+
+/// Magic bytes in the EE DA commit transaction marker output.
+///
+/// TODO(STR-1907): derive this from authenticated EE proof context instead of
+/// baking the network value into runtime/proof code.
+pub const EE_DA_MAGIC_BYTES: [u8; 4] = *b"ALPN";
```
delbonis: ahhhhh plssssss derive this from configuration

crates/alpen-ee/da/types/src/bitcoin_merkle.rs
delbonis, line 1: These names are very ambigous.  Add `compute_` or `verify_` as prefixes.

crates/alpen-ee/da/types/src/bitcoin_merkle.rs
```diff
@@ -0,0 +1,143 @@
...
+
+use strata_crypto::hash::sha256d;
+
+/// Hashes one Bitcoin merkle-tree level as `SHA256(SHA256(left || right))`.
+pub fn bitcoin_hash_pair(left: [u8; 32], right: [u8; 32]) -> [u8; 32] {
+    let mut pair = [0u8; 64];
+    pair[..32].copy_from_slice(&left);
+    pair[32..].copy_from_slice(&right);
+    sha256d(&pair).0
+}
```
delbonis, line 17: `hash_pair_sha256d` maybe?

crates/alpen-ee/da-types/src/parse.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/// Extracts and orders DA payload chunks from a set of L1 transactions.
+///
+/// Returns the chunks in commit-output order. Caller is responsible for
+/// checking the commit marker's magic bytes / version against its own expected
+/// values; this parser is format-agnostic.
+pub fn extract_da_chunks<'a>(
+    txs: impl Iterator<Item = &'a Transaction>,
+) -> Result<Vec<Vec<u8>>, DaParseError> {
```
delbonis: Should clarify this is operating on a chained sequence of txs as a single blob instead of like, "all the txs in a block" and extracting whatever blobs there are in there, as was my first impression.

crates/alpen-ee/da-types/src/parse.rs
```diff
@@ -0,0 +1,160 @@
...
+/// commit transaction's vout-0 output, if present.
+///
+/// Returns:
+/// - `Ok(Some([u8; 8]))` if the first output is a well-formed OP_RETURN with exactly 8 pushed bytes
+///   (`magic_bytes || version_be_u32`).
+/// - `Ok(None)` if the first output is missing or not an OP_RETURN (i.e. this isn't a commit
+///   transaction).
+/// - `Err(MalformedCommitMarker)` if the OP_RETURN exists but has the wrong shape (extra ops, wrong
+///   push size, etc.).
+pub fn commit_marker_payload(tx: &Transaction) -> Result<Option<[u8; 8]>, DaParseError> {
```
delbonis: Misleading name since `commit` is a verb.

crates/alpen-ee/da-types/src/parse.rs
```diff
@@ -0,0 +1,160 @@
...
+    Ok(chunks_by_vout.into_values().collect())
+}
+
+/// Vout of the last P2TR reveal output on a commit transaction.
+///
+/// Scans outputs starting at vout 1 (vout 0 is the OP_RETURN marker) and
+/// returns the last contiguous P2TR vout. Returns 0 if there are no reveal
+/// outputs.
+pub fn last_commit_reveal_vout(commit: &Transaction) -> u32 {
+    commit
```
delbonis: Should return `Option<u32>` and return `None` if there are no reveal outputs.

Also, does this make sense to be `pub`?

crates/alpen-ee/da/src/verification/error.rs
```diff
@@ -0,0 +1,84 @@
...
+use strata_codec::CodecError;
+
+/// Result type used throughout DA verification. Defaults the success type to
+/// `()` for the common "verify and return nothing" case; pass a type parameter
+/// for helpers that produce data along the way (e.g. decoded chunk transitions).
+pub type DaVerificationResult<T = ()> = Result<T, DaVerificationError>;
+
+/// Errors raised while verifying EE DA witness data.
+#[derive(Debug, thiserror::Error)]
+pub enum DaVerificationError {
```
delbonis: Colons imply a new level of hierarchy to the errors.  Use parens for `foobar (expected {expected}, got {got})`.

crates/snark-acct-types/ssz/proof_interface.ssz
```diff
@@ -9,6 +9,9 @@ MAX_MESSAGE_ENTRIES = 2 << 10
 ### Public params that we provide as the claim the proof must prove the relate
 ### to each other correctly.
 class UpdateProofPubParams(Container):
+    ### Sequence number of the account update being proven.
+    seq_no: uint64
+
```
delbonis: Shit!  Please wait for this PR to be merged: https://github.com/alpenlabs/alpen/pull/1832

crates/alpen-ee/da/Cargo.toml
delbonis, line 1: This is some really complicated conditional complication here, what is the motivation for this?  Can we break components into multiple crates to avoid having such a wide scope?

crates/ee-chain-types/src/chunk.rs
```diff
@@ -2,18 +2,36 @@
...
-use crate::{ChunkTransition, ExecInputs, ExecOutputs};
+use crate::{ChunkTransition, ExecHeaderSummary, ExecInputs, ExecOutputs};
+
+impl ExecHeaderSummary {
+    pub fn new(opaque_bytes: Vec<u8>) -> Self {
+        Self {
+            opaque_bytes: opaque_bytes
+                .try_into()
+                .expect("exec header summary must fit within SSZ max length"),
+        }
```
delbonis: Make this accept the SSZ variable list directly and then have a fallible `from_vec` fn.

bin/alpen-client/src/prover/spec_chunk.rs
```diff
@@ -235,6 +235,11 @@ impl ProofSpec for ChunkSpec {
         let mut aggregated_inputs = ExecInputs::new_empty();
         let mut aggregated_outputs = ExecOutputs::new_empty();
         let mut tip_blkid = parent_blkid;
+        // Safe placeholder values: `block_hashes` was checked non-empty above
+        // and the witness block count must match it, so the loop below always
+        // overwrites these with the terminal block's verified metadata.
+        let mut tip_state_root = Hash::zero();
+        let mut tip_exec_header_summary = ExecHeaderSummary::new(Vec::new());
```
delbonis: Maybe add a `new_empty` constructor too?

crates/ee-acct-runtime/src/da_witness.rs
```diff
@@ -0,0 +1,323 @@
...
+/// re-hashes them before accepting the account diff.
+///
+/// This proves bytecode identity, not prior L1 publication.
+///
+/// TODO(STR-1907): prove membership in an authenticated published-bytecode set,
+/// or include explicit prior blob inclusion for the omitted bytecode.
+#[derive(Clone, Debug, Eq, PartialEq, Archive, Deserialize, Serialize)]
+#[rkyv(derive(Debug))]
+pub struct DaBytecodeWitness {
+    code_hash: [u8; 32],
```
delbonis: Why do we need to have the code hash here?  We can compute it from the bytecode.

crates/ee-acct-runtime/src/da_witness.rs
```diff
@@ -0,0 +1,323 @@
...
+pub struct DaWitness {
+    /// One per L1 block holding DA commit/reveal transactions for this batch.
+    blocks: Vec<DaBlockWitness>,
+
+    /// Private bytecodes omitted from the current DA blob because they were
+    /// already known locally from prior DA publication.
+    ///
+    /// TODO(STR-1907): replace this private witness path with an authenticated
+    /// prior-publication proof once bytecode DA dedupe is protocol-bound.
+    known_bytecodes: Vec<DaBytecodeWitness>,
```
delbonis: This should be reworked, maybe move to alpen proof itself if it cannot be made into a generic concept.  This is an EVM-specific concept and should not be in the general purpose EE acct runtime.

I think it's possible we could rework it to be some sort of way to provide generic "hash preimages" to the runtime, but it's important not to overspecialize this.

Also, I assume these should be sorted by hash so we can do binary searches when looking up the preimage (bytecode)?

crates/alpen-ee/da/types/src/witness.rs
```diff
@@ -0,0 +1,323 @@
...
+pub struct L1DaBlockInclusion {
+    /// Bitcoin block height.
+    l1_block_height: u32,
+
+    /// Bitcoin block hash in internal byte order.
+    l1_block_hash: [u8; 32],
+
+    /// Witness transaction ID Merkle root in internal byte order.
+    wtxids_root: [u8; 32],
+}
```
delbonis, line 66: Maybe should do a similar pattern to what we do in other places where we take something that's SSZ as an input so that we have an easier time mapping it to what we have commitments to in the public params.  Just so that we don't have to deal with expressing multiple different forms of the same semantic concept.

crates/ee-acct-runtime/src/da_witness.rs
```diff
@@ -0,0 +1,323 @@
...
+}
+
+/// Bitcoin Merkle inclusion proof.
+///
+/// `siblings` is ordered leaf-first. `position` is the leaf index in the
+/// bottom layer; bit `i` selects whether `siblings[i]` is on the left or right
+/// of the running hash at level `i`.
+#[derive(Clone, Debug, Default, Archive, Deserialize, Serialize)]
+#[rkyv(derive(Debug))]
+pub struct BitcoinMerkleProof {
```
delbonis: Maybe makes sense to define this somewhere else and import it?

crates/snark-acct-types/ssz/update.ssz
```diff
@@ -13,17 +13,26 @@ MAX_UPDATE_PROOF_BYTES = 2 << 20
 ### Max extra data bytes
 MAX_EXTRA_DATA_BYTES = 1024
 
+### L1 block reference committed into the OL L1 block refs MMR.
+class L1BlockRef(Container):
+    ### Bitcoin block hash for the L1 block that contains DA reveal txs.
+    block_hash: Bytes32
+
+    ### Witness transaction Merkle root for that L1 block.
+    wtxids_root: Bytes32
```
delbonis: Should probably be defined in acct types since this is principally part of the ledger, not the snark acct system.

Also, maybe call this `L1BlockRecord`

crates/snark-acct-types/src/update.rs
```diff
@@ -1,12 +1,49 @@
...
+
+    /// Computes the canonical OL L1 block refs MMR leaf hash.
+    pub fn leaf_hash(&self) -> [u8; 32] {
+        <L1BlockRef as TreeHash>::tree_hash_root(self).into_inner()
+    }
```
delbonis: `compute_commitment`

crates/snark-acct-types/src/update.rs
```diff
@@ -1,12 +1,49 @@
...
+}
+
+/// Computes the canonical OL L1 block refs MMR leaf hash.
+pub fn l1_block_ref_leaf_hash(block_hash: &[u8; 32], wtxids_root: &[u8; 32]) -> [u8; 32] {
+    L1BlockRef::new(*block_hash, *wtxids_root).leaf_hash()
+}
```
delbonis: Do we really need to have a fn like this?

delbonis (changes requested):
This PR is so enormous that I don't know for sure if I've been able to touch upon everything I would have wanted to, but I have managed to identify several potential design issues that we should consider some more.  I am pretty happy with how clean most of this is, but most of my concerns are around making sure we avoid muddying some concepts and that the various components are well-organized and designed to be reusable.

I have not been able to fully grasp the all of the dataflow and what the reasons for everything are, but I feel like I have a general understanding of the approach being taken here.  Some comments I made may not be considering constraints I have not been made aware of, so it's possible that the things I'm asking for in some places are unreasonable given the already-huge scope and timelines.

crates/proof-impl/alpen-acct/src/lib.rs
```diff
@@ -48,6 +54,21 @@ pub fn process_ee_acct_update(zkvm: &impl ZkVmEnvSerde, chunk_predicate_key: &Pr
...
+        .try_decode_update_pub_params()
+        .expect("failed to decode update public params");
+    let pre_account_state: EeAccountState = upd_input
+        .try_decode_pre_state()
+        .expect("failed to decode EE account pre-state");
+    let expected_pre_state_root = pre_account_state.last_exec_state_root().0;
+    verify_da_witness(
+        ee_input,
+        da_witness,
+        &update_pub_params,
```
irnb: mention this again maybe intead of proof detail realted things and apss big struct we could just pass the field da witness needs them and repect the boundry of crates

functional-tests/tests/alpen_client/test_fullnode_sync.py
storopoli: This is not solved yet.

functional-tests/common/config/constants.py
storopoli: This is not solved yet

crates/snark-acct-types/src/update.rs
storopoli: YES! please!

crates/snark-acct-types/ssz/proof_interface.ssz
storopoli: It's in the queue now

crates/alpen-ee/da-types/src/blob.rs
irnb: TODO(STR-1907)

crates/alpen-ee/da/types/src/witness.rs
irnb: yes we need those tests.

functional-tests/common/config/constants.py
irnb: This raises the ceiling, not the run time. So we already use wait_until; this just widens the budget so it doesn't flake when DA+prover IO pressure pushes block import past the 1s/block cadence. Fast runs finish just as fast.

functional-tests/tests/alpen_client/test_fullnode_sync.py
irnb: same.

bin/alpen-client/src/prover/spec_acct.rs
irnb: The boundary is intentionally string-at-the-edge.

crates/alpen-ee/da/types/src/witness.rs
irnb: This crosses the rkyv private-input boundary, so we can't carry the SSZ L1BlockRecord directly.
 It's not really a second form of the concept though.

functional-tests/common/config/constants.py
storopoli: Why not raise to 10 seconds then?

functional-tests/tests/alpen_client/test_fullnode_sync.py
storopoli: Why not raise to 30 seconds and see if we experience flakiness?

bin/alpen-client/src/prover/spec_acct.rs
```diff
@@ -266,17 +281,28 @@ impl ProofSpec for AcctSpec {
...
+        //    tip state root. This spans the full batch, not just one chunk.
+        let batch_block_hashes: Vec<Hash> = batch.blocks_iter().collect();
+        let first_batch_block = batch_block_hashes.first().copied().ok_or_else(|| {
+            PaasError::PermanentFailure(format!("batch {batch_id} has no execution blocks"))
+        })?;
+        let first_block_hash = B256::from(first_batch_block.0);
+        let last_block_hash = B256::from(batch.last_block().0);
+        let range_witness_fn = self.range_witness_fn.clone();
+        let range_data =
+            task::spawn_blocking(move || (range_witness_fn)(first_block_hash, last_block_hash))
```
evgenyzdanovich, line 303: that's quite heavy at the moment... can't we derive or return it from the chunk proof part somehow or just take it from the chunk witness storage maybe?

crates/reth/statediff/src/reconstruct.rs
```diff
@@ -20,6 +24,8 @@ use crate::{
 pub enum ReconstructError {
     #[error("MPT: {0}")]
     Mpt(#[from] strata_mpt::Error),
+    #[error("sparse MPT: {0}")]
+    SparseMpt(#[from] rsp_mpt::Error),
```
evgenyzdanovich, line 28: nit: do we need to distinguish between two above?

evgenyzdanovich (commented):
Conceptual LGTM on EE part (one concern, one nit), completely skipped review of OL part.

bin/alpen-client/src/prover/spec_acct.rs
irnb: each persisted `ChunkWitnessRecord.raw_partial_pre_state` is a sparse trie anchored at that chunk's pre-state root and only covers that chunk's touched accounts, not the whole batch. acct proof needs sparse trie of all touched accounts during the batch.
  
but agreed we shouldn't recompute this on the prover path. we can precompute it after batch sealing as part of the EE lifecycle and have it ready in db. 

if you are ok i can create a ticket to tackle this later.

irnb: https://alpenlabs.atlassian.net/browse/STR-3651

functional-tests/common/config/constants.py
storopoli: Solved by bf27b5161481ff1a95aef4c9463061cc44c07889

functional-tests/tests/alpen_client/test_fullnode_sync.py
storopoli: Solved by bf27b5161481ff1a95aef4c9463061cc44c07889

storopoli (approved):
ACK bf27b5161481ff1a95aef4c9463061cc44c07889

crates/ol/state-types/src/epochal.rs
```diff
@@ -47,23 +47,26 @@ impl EpochalState {
...
 
-        let manifest_hash = <AsmManifest as TreeHash>::tree_hash_root(&mf);
+        let l1_block_ref_hash =
+            l1_block_record_leaf_hash(mf.blkid().as_ref(), mf.wtxids_root().as_ref());
 
-        Mmr::<StrataHasher>::add_leaf(&mut self.manifests_mmr, manifest_hash.into_inner())
+        Mmr::<StrataHasher>::add_leaf(&mut self.l1_block_refs_mmr, l1_block_ref_hash)
             .expect("ol/state: MMR capacity exceeded");
         self.last_l1_block = L1BlockCommitment::new(height, *mf.blkid());
     }
```
delbonis: This is actually an important subtlety that I missed in the first go around.  If we're moving away from remembering the ASM manifests then it'd make sense to remove them from the interfaces around the state types.  This fn should accept the new `L1BlockRecord` type itself instead of passing it off to `l1_block_record_leaf_hash` to construct and then discard.

I'm also noticing that some of the state accessors *also* do this `add_leaf` call.  I'm wondering if it makes sense to have a `fn merge_l1_block_rec_into_mmr(mmr: &mut impl Mmr, rec: &L1BlockRecord)` in this crate that handles this reusably.

crates/ol/state-support-types/src/write_tracking_layer.rs
```diff
@@ -236,23 +238,22 @@ where
...
-            .unwrap_or_else(|| self.base.asm_manifests_mmr().clone());
+            .unwrap_or_else(|| self.base.l1_block_refs_mmr().clone());
 
-        use strata_acct_types::{StrataHasher, tree_hash::TreeHash};
-        let manifest_hash = <AsmManifest as TreeHash>::tree_hash_root(&mf);
-        strata_merkle::Mmr::<StrataHasher>::add_leaf(&mut mmr, manifest_hash.into_inner())
+        let l1_block_ref_hash =
+            l1_block_record_leaf_hash(mf.blkid().as_ref(), mf.wtxids_root().as_ref());
+        strata_merkle::Mmr::<StrataHasher>::add_leaf(&mut mmr, l1_block_ref_hash)
             .expect("MMR capacity exceeded");
```
delbonis: See here, this should try to reuse the same low-level MMR call.

crates/ledger-types/src/state_accessor.rs
```diff
@@ -95,9 +95,10 @@ pub trait IStateAccessorMut: IStateAccessor {
     /// Sets the current epoch.
     fn set_cur_epoch(&mut self, epoch: u32);
 
-    /// Appends a new ASM manifest to the accumulator, also updating the last L1
-    /// block height and other fields.
-    fn append_manifest(&mut self, height: L1Height, mf: AsmManifest);
+    /// Appends an accepted ASM manifest's L1 block ref to the accumulator.
+    ///
+    /// This also updates the last L1 block height and ID.
+    fn append_l1_block_ref_from_manifest(&mut self, height: L1Height, mf: AsmManifest);
```
delbonis: In accordance with other changes, this should probably be `fn append_l1_block_rec(&mut self, height: L1Height, rec: L1BlockRecord);`.

crates/ol/stf/src/manifest_processing.rs
```diff
@@ -113,7 +113,7 @@ fn process_asm_manifest<S: IStateAccessorMut>(
     }
 
     // 2. Accept the manifest into the ASM MMR.
-    state.append_manifest(real_height, mf.clone());
+    state.append_l1_block_ref_from_manifest(real_height, mf.clone());
```
delbonis: So here we'd construct the `L1BlockRecord` pass it off so the state code doesn't touch the ASM manifest.

delbonis (changes requested):
Ok this is a fairly minor change but hopefully it simplifies the deps a little bit too by making the state crate not have to care about manifests at all.

Otherwise looks good. :slightly_smiling_face:

krsnapaudel (approved):
LGTM

irnb:
@delbonis done.

storopoli (approved):
ACK dbd6950be56abbd8eb237d81fd5999a85e9761f8

delbonis (approved):

sapinb (approved):
ACK dbd6950be56abbd8eb237d81fd5999a85e9761f8

crates/reth/statediff/src/reconstruct.rs
storopoli: Solving this for now, feel free to move to a ticket Cc @irnb

PR #1859, opened 2026-05-25
chore(deps): bump `ssz-gen` to `0.16.0`

storopoli:
## Description

Bumps `ssz-gen` to `0.16.0` and friends along the way.

Had to update traits bounds on `ssz-gen`'s `TreeHasher` because of https://github.com/alpenlabs/ssz-gen/pull/82 and also had to update `asm`'s logging config because of https://github.com/alpenlabs/strata-common/pull/104 and https://github.com/alpenlabs/asm/pull/111

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [x] Dependency Update

## Notes to Reviewers

Is this PR addressing any specification, design doc or external reference document?

- [ ]  Yes
- [x]  No

If yes, please add relevant links:

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [x] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [x] New and existing tests pass with my changes.
- [ ] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

PR #1864, opened 2026-05-27
STR-3308: manifest handling refactor

delbonis:
## Description

This PR addresses the issue with manifests only being able to be posted in terminal blocks.  This makes it so we buffer ASM logs and handle those at the end of the epoch, but lets us update MMRs immediately instead of waiting until the end of the epoch to include them.  This lets us shorten the minimum latency for confirming execution updates on L1.

Claude did all the tedious refactoring work.

**PR is still in-progress.**

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist


delbonis:
@codex review

crates/proof-impl/checkpoint/src/statements.rs
```diff
@@ -308,9 +300,9 @@ fn execute_block_batch(
         parent = output.completed_block().header().clone();
     }
 
-    // Guaranteed by the l1_update_count == 1 assertion above.
-    let asm_manifests_hash =
-        asm_manifests_hash.expect("exactly one L1 update per epoch is enforced above");
+    let epoch_manifests = OLAsmManifestContainer::new(epoch_manifests)
+        .expect("epoch manifests should be within container limits");
```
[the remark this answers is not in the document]
delbonis: fixed

delbonis:
Another change I made here, since we're making changes to the state accessor system, is to simplify the snark account interface to only have `set_proof_state(_directly)`.  This required reworking some bits about how we handle logs, since we need to record the extra_data field from logs instead of capturing it in the indexer layer.  But this does simplify our indexing types.

In the process, I realized that OL logs were not being formatted according to msgfmt like I thought they were, and we were instead using the source account serial to signal how to interpret logs. (yikes!  I should have caught this a long time ago!) So I also made that change.  I will be making the corresponding change in the ASM repo shortly and will open a ticket to move these log types to the strata-common repo to avoid duplication.

delbonis:
Created ticket for future work: https://alpenlabs.atlassian.net/browse/STR-3629

storopoli (approved):
ACK 184a06a8d681a78a0bfdebab3a495b07efe947c4

delbonis:
@codex review

crates/ol/state-support-types/src/memory_state_layer.rs
```diff
@@ -167,6 +185,23 @@ impl IStateAccessorMut for MemoryStateBaseLayer {
...
 
+    fn try_append_pending_asm_log(&mut self, entry: PendingAsmLog) -> StateResult<()> {
+        let ssz_entry = entry.into();
+        if self
+            .state
+            .intraepoch_state_mut()
+            .try_append_pending_log(ssz_entry)
+        {
+            Ok(())
+        } else {
```
bewakes: `try_append_pending_log` returns bool using `.is_ok()`. It is not always guaranteed that the else here means `PendingAsmLogsFull`. Why not return `Result` from `try_append_pending_log` instead of `bool`?

crates/proof-impl/checkpoint/src/statements.rs
```diff
@@ -212,6 +214,28 @@ pub fn process_ol_stf_core(
     )
 }
 
+/// Artifacts collected while executing an epoch's block batch.
+struct EpochBatchExecution {
```
bewakes: The name is a bit confusing as `Execution` might also indicate the inputs for the execution, and `EpochBatch` is redundant. How about `EpochExecutionArtifacts`/`BatchExecutionArtifacts`?

crates/ol/chain-types/src/log_payloads.rs
```diff
@@ -210,6 +273,44 @@ mod tests {
         assert_eq!(decoded.extra_data.as_ref(), &vec![255u8; 250][..]);
     }
 
+    #[test]
+    fn test_log_envelope_round_trip() {
```
bewakes, line 314: Consider using`proptest` for roundtrip tests because the log is central to the protocol and just in case the single roundtrip might be missing something.

bewakes (commented):
Awesome refactor - especially with the removal of `preseal*` ceremony and separating out `extra_data` from account update state accessor entirely. A couple of nits and suggestions.

One question - is it correct to say that preseal check has always been redundant since we can always do a full state root after applying epoch update?

storopoli:
@delbonis rebase please :)

delbonis:
Rebased, will address review comments.

crates/proof-impl/checkpoint/src/statements.rs
delbonis: Renamed to `EpochExecTrace`.

crates/ol/state-support-types/src/memory_state_layer.rs
delbonis: Cleaned up and added more safety checks.

crates/ol/chain-types/src/log_payloads.rs
delbonis: Added proptests since this has more involved in it than just the `Codec`-generated logic.

bewakes (approved):

delbonis:
> is it correct to say that preseal check has always been redundant since we can always do a full state root after applying epoch update?

In a strict sense yes, but it was an additional sanity check that let us check state equality before processing manifests, so we only had to process manifests once.  There was some discussion a while ago about removing it to simplify some things, since actually doing the state equality check is more expensive than processing the manifests, so doing the manifest processing twice means we can compute the state root one fewer times.  So I finally got around to doing that now.

bewakes:
I get that computing state root is costlier than processing manifests and without preseal check we do state root computation one less time. But I don't get how manifest is processed twice.

storopoli (approved):
ACK 3307c16bd57f97b33a61724d6b186787d3558379

storopoli:
@delbonis I think you need to fix functional tests.
`test_terminal_block_assembly_fast_l1` fails with `no L1Manifest` error.

delbonis:
That's weird, it was passing before.

storopoli:
Let's rerun... Toggling a rerun now

storopoli:
If it passes please add `https://github.com/alpenlabs/alpen/actions/runs/26819950266/job/79071612703?pr=1864` to https://alpenlabs.atlassian.net/browse/STR-3381

Cc @bibek

storopoli:
@delbonis nope

`test_terminal_block_assembly_fast_l1` fails again with `AssertionError: terminal slot 64 has no L1 update` probably because of the `asm` bump in #1859.

delbonis:
Force merging due to triviality since last approval.

delbonis:
@storopoli The reason the test failed was because we changed an RPC field name and those tests were added to main since the previous run.  It's a real failure but it's really fixed now.

PR #1866, opened 2026-05-27
STR 3422 Sequencer signs two blocks per slot when two `GenerationTick`s race

krsnapaudel:
## Description

Prevent accidental OL block proposer equivocation

### Problem
The sequencer could generate, sign, and accept two OL blocks for the same slot on the same parent. Storage indexes `slot → Vec<OLBlockId>`, FCM checks only the signature, the unfinalized tracker accepts siblings, and the STF enforces `slot = parent.slot + 1` only per chain, so nothing downstream rejected it. The trigger: completing a template released the parent before the signed block became the visible tip, so a generation tick on the stale tip rebuilt a second template; a restart reopened the window.

### Fix
- `fix(ol/block-assembly)`: per-parent `Pending`/`Completed` tombstone; generation no-ops on `Completed`; tip-progression pruning, no wall-clock release.
- `feat(db)`: durable monotonic per-slot high-watermark.
- `fix(ol/sequencer)`: generation and duty extraction skip slots `≤` the high-watermark; completion persists via the atomic write; completion is split so the in-memory release happens only after the durable commit; FCM clears the high-watermark when that block is marked `Invalid`.
- `fix(strata)`: on startup, resubmit the high-watermark block to FCM if it is still `Unchecked`; clear the high-watermark if that block is already `Invalid`.
- `feat(strata-dbtool)`: expose OL blocks by slot for raw DB inspection.
- `fix(strata-dbtool)`: rollback OL block high-watermark on `revert-ol-state`.

### DB API + naming
Adds `get_block_high_watermark()` and `put_block_data_with_high_watermark(block)` to `OLBlockDatabase`, backed by a dedicated sled tree in the same `with_retry` transaction as block/status/height. The write rejects `slot ≤ current.slot` (`BlockHighWatermarkConflict`), else advances the mark and persists the block atomically. `put_block_data` is unchanged, so ingestion/sync paths are unaffected.

Also adds `clear_block_high_watermark(expected)` and `rollback_block_high_watermark(target)`. Clearing removes the singleton mark only if it currently equals `expected`, used when FCM invalidates that block. Rollback rewinds the mark to an existing target block during `strata-dbtool revert-ol-state`.

Naming is use-neutral on purpose: the DB knows only a monotonic-by-slot high-watermark over committed blocks, not that the caller treats it as an anti-equivocation signing floor. The persisted block keeps its `Unchecked` status until FCM processes it; the high-watermark is a separate singleton record, not a block status, and is not the canonical tip.

### Invalid Blocks
If the high-watermark block is marked `Invalid`, the mark is cleared so the existing replacement behavior can continue. In that case the DB may retain the invalid block plus a replacement at the same slot. The safety invariant is that the node does not accidentally accept/persist two competing non-invalid proposals because of the original race.

### strata-dbtool
Adds `get-ol-blocks-at-slot <slot>`. The node RPC is canonical-chain-only, so a non-canonical sibling at a slot is invisible over RPC. The command reads the raw per-slot block IDs plus a count from the OL block DB as JSON, so the functional test can assert exactly one block at the crash slot. It runs against a stopped node, since dbtool reads sled directly.

### Why no signer-side guard
`strata-signer` is a stateless signing oracle: key plus RPC, no DB. It signs whatever duty the node hands it. The guards live on the node, which controls what gets offered for signing and what gets accepted. A durable signer guard would require adding persistence to the signer and would only close the pre-commit crash window, where the stray signature is non-durable and never reaches storage or FCM.

### Tests
- DB proptests: same-slot put rejected with no partial write; plain `put_block_data` does not advance the mark; mark is monotonic under mixed puts; clearing only removes the expected mark; rollback rewinds only to an existing lower block; `get_blocks_at_height` returns all siblings at a slot.
- Block assembly: no rebuild after completion; tombstone pruning boundary; completion split, so the template stays cached until the durable commit; releasing an invalid completed-template tombstone allows same-slot regeneration.
- Unit: resume/startup decision covers `Unchecked` resubmit, `Valid` skip, and `Invalid` clear; generation and duty-extraction high-watermark filters; `NodeSequencerContext` releases an invalid completed-template tombstone and retries generation.
- Functional (`test_crash_fcm_new_block`): crash after the block is durably committed leaves one `Unchecked` block at the crashed slot; restart resubmits it; chain advances past it; the same block becomes `Valid` and remains unique at that slot.

### Scope
Stops an honest sequencer from accidentally producing two blocks per slot, including across restart.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Please check the issue flagged by Codex below about `Invalid` high-watermark blocks.
The high-watermark is intentionally not cleared when FCM marks the recorded block `Invalid`. `BlockStatus::Invalid` is not currently a proof that the proposal can be safely forgotten: FCM also maps execution/processing failures into `Invalid` (see the STR-2141 TODO around distinguishing invalid blocks from exec failures).

Clearing the high-watermark on `Invalid` would allow the sequencer to sign and persist a different block at the same slot after a transient local failure, which reopens accidental proposer equivocation. The safe behavior in this PR is to retry the same high-watermark block on startup. If it was transiently rejected, retry can recover; if it is genuinely invalid, the sequencer remains stuck and needs operator/protocol-level intervention rather than automatic same-slot replacement.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues
[STR-3422](https://alpenlabs.atlassian.net/browse/STR-3422)

[STR-3422]: https://alpenlabs.atlassian.net/browse/STR-3422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/strata/src/sequencer/block_producer.rs
+++ b/bin/strata/src/sequencer/block_producer.rs
@@ -2,10 +2,15 @@
 
 use std::{sync::Arc, time::Duration};
 
-use anyhow::{Result, anyhow};
+use anyhow::{Context, Result, anyhow, bail};
+use strata_consensus_logic::{FcmServiceHandle, message::ForkChoiceMessage};
+use strata_db_types::traits::BlockStatus;
+use strata_identifiers::{OLBlockCommitment, OLBlockId};
+use strata_ol_chain_types_new::OLBlock;
 use strata_ol_sequencer::{SequencerBuilder, SequencerServiceStatus};
 use strata_service::ServiceMonitor;
-use tracing::info;
+use strata_storage::NodeStorage;
+use tracing::{error, info};
 
 use super::node_context::NodeSequencerContext;
 use crate::run_context::RunContext;
@@ -25,6 +30,14 @@ pub(crate) fn start_block_producer(
         .ok_or_else(|| anyhow!("sequencer config required when block producer is enabled"))?
         .ol_block_time_ms;
 
+    runctx
+        .task_manager()
+        .handle()
+        .block_on(process_startup_high_watermark_block(
+            runctx.storage().as_ref(),
+            runctx.fcm_handle().as_ref(),
+        ))?;
+
     let context = Arc::new(NodeSequencerContext::new(
         handles.blockasm_handle().clone(),
         runctx.storage().clone(),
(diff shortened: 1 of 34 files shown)


krsnapaudel:
@codex review

bin/strata/src/sequencer/node_context.rs
```diff
@@ -54,6 +54,25 @@ impl SequencerContext for NodeSequencerContext {
...
         let tip_blkid = *parent_commitment.blkid();
+        let target_slot = parent_commitment.slot().saturating_add(1);
+
+        let high_watermark = self
+            .storage
+            .ol_block()
+            .get_block_high_watermark_async()
+            .await
+            .map_err(SequencerContextError::Db)?;
+        if target_slot_at_or_below_high_watermark(target_slot, high_watermark.as_ref()) {
```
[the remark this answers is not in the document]
krsnapaudel: This is an intentional safety/liveness tradeoff. The high-watermark is an anti-equivocation floor, not an execution-validity marker. Since `BlockStatus::Invalid` currently also covers transient execution failures, clearing the high-watermark or generating a replacement on Invalid would reopen accidental same-slot equivocation.

bin/strata/src/sequencer/block_producer.rs
```diff
@@ -42,3 +55,221 @@ pub(crate) fn start_block_producer(
...
+    high_watermark: Option<OLBlockCommitment>,
+    block: Option<&OLBlock>,
+    status: Option<BlockStatus>,
+) -> Result<HighWatermarkResumeDecision> {
+    let Some(high_watermark) = high_watermark else {
+        return Ok(HighWatermarkResumeDecision::SkipNoHighWatermark);
+    };
+
+    let block_id = *high_watermark.blkid();
+    let block = block.ok_or_else(|| anyhow!("high-watermark OL block {block_id} is missing"))?;
```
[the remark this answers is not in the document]
krsnapaudel: We will see with `strata-dbtool` functional tests for delete blocks. The test did fail. Adding rollback for high watermark block.

delbonis (approved):
Seems good, is clean.

I'm a little concerned about increasing the surface area of the database interface, it seems like the "high water mark" could be inferred on the fly, but maybe this is unavoidable.

crates/consensus-logic/src/fcm/service.rs
```diff
@@ -284,9 +316,7 @@ async fn handle_new_block<C: FcmContext>(
             .set_block_status(*blkid, BlockStatus::Valid)
             .await?;
     } else {
-        fcm_state
-            .ctx()
-            .set_block_status(*blkid, BlockStatus::Invalid)
+        set_block_status_and_clear_invalid_high_watermark(fcm_state, bc, BlockStatus::Invalid)
```
storopoli, line 322: This clears the anti-equivocation high-watermark whenever FCM records `BlockStatus::Invalid`, but the nearby `try_exec_block` branch still maps generic execution/processing failures into `Invalid` under the [STR-2141](https://alpenlabs.atlassian.net/browse/STR-2141) TODO:

https://github.com/alpenlabs/alpen/blob/6b286bd4ba35d0b038c72aed99616a6989623c81/crates/consensus-logic/src/fcm/service.rs?plain=1#L268-L279

After that clear, the canonical tip remains the parent and the next generation tick can sign and persist a different block for the same slot, so a transient local failure reopens the equivocation window this PR is trying to close.

This also conflicts with the PR note/earlier review resolution that `Invalid` is not currently enough proof that the proposal can be forgotten. The startup path has the same policy in `decide_startup_high_watermark_block_action`:

https://github.com/alpenlabs/alpen/blob/90850a27d71882c0feaeec45189633fb042ef939/bin/strata/src/sequencer/block_producer.rs#L162-L165

I think we should keep the watermark for generic `Invalid`, or first split consensus-invalid from local execution failure and only clear for the former.

storopoli (changes requested):

crates/consensus-logic/src/fcm/service.rs
krsnapaudel: @storopoli We are not trying to change the existing behavior for what happens when FCM finds a block `Invalid`. In that case, a new block was proposed replacing the `Invalid` one. The behavior stays the same.

"I think we should keep the watermark for generic Invalid, or first split consensus-invalid from local execution failure and only clear for the former."

Isn't this the task for TODO(STR-2141)? We can certainly revisit the high watermark logic after that is fixed.

storopoli: Yeah, let's add a `TODO` comment with the same ticket here as well? I feel this might fly under the radar.

storopoli (approved):
ACK 5ffa650eabd2702fb0deb5130c4c196464e2a561

PR #1868, opened 2026-05-27
fix(prover-perf): use a DA-backed checkpoint fixture

peter941221:
## Description

Restores a realistic checkpoint perf fixture by deriving the DA witness from the same replay path used by checkpoint generation.

The old fixture had fallen back to an empty-chain path because the guest now requires a valid DA witness. That made the reported checkpoint cycle counts understate the real checkpoint workload.

This change builds a transaction-bearing chain, replays the epoch through `DaAccumulatingState`, and takes `da_state_diff_bytes` from `execute_block_batch_preseal` instead of hand-building a slot-only DA payload. It also drops the DA helper dependencies that are no longer used by this path.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [x] Dependency Update

## Notes to Reviewers

This branch no longer carries the separate digest alias change. That toolchain-compatibility fix now lives only in `#1872`.

The release before/after numbers for this fixture use this patch together with `#1872`, because `rustc +succinct` still ICEs on the deprecated alias path without that separate unblock.

- [ ]  Yes
- [x]  No

## Checklist

- [x] I have disclosed my use of AI in the body of this PR.


peter941221:
1. I pushed 0df52734 to remove the digest alias change from this branch.

2. That toolchain-compatibility fix now lives only in alpen#1872. The live file list for this PR is back to the checkpoint fixture change plus its lockfile update.

3. I also tightened the PR body so the benchmark boundary is explicit: the release before/after numbers use this fixture change together with #1872 as the separate guest-toolchain unblock.

peter941221:
Rebased this onto current main and resolved the checkpoint fixture conflict.

This branch still only changes Cargo.lock, bin/prover-perf/Cargo.toml, and bin/prover-perf/src/programs/checkpoint.rs.

I reran cargo fmt --all --check, cargo test -p strata-proofimpl-checkpoint program::tests::test_statements_success -- --nocapture, and cargo check -p strata-provers-perf --bin strata-provers-perf in WSL after the rebase. All three passed.

PR #1869, opened 2026-05-27
test(functional-tests): add terminal block assembly coverage

krsnapaudel:
## Summary

Adds functional coverage for OL terminal block assembly at `slots_per_epoch = 64` under both fast and slow L1 timing.

The race fix is ordering-only and is applied in the external `strata-asm-worker` L1 block path: `crates/worker/src/service.rs::process_block` stores the L1 manifest before storing the committed anchor state.

Confirmed in pinned `strata-asm-worker` (`v0.1-alpha.9`, `dfb8e53`): both writes run sequentially in the single `SubmitBlock -> process_block` path. The fix is ordering-only, not a shared DB transaction; a crash can leave an orphan manifest, but not committed ASM state pointing at a missing manifest.

### Tests

- Added fast L1 terminal block test with 2s Bitcoin blocks, validating repeated terminal blocks and asserting multiple manifests are included.
- Added slow L1 terminal block test with 30s Bitcoin blocks to preserve baseline behavior.
- Shared Strata terminal block helpers under `functional-tests/tests/strata/helpers.py`.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-3337](https://alpenlabs.atlassian.net/browse/STR-3337)

[STR-3337]: https://alpenlabs.atlassian.net/browse/STR-3337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/functional-tests/tests/strata/helpers.py
+++ b/functional-tests/tests/strata/helpers.py
@@ -1,0 +1,115 @@
+"""Helpers for Strata functional tests."""
+
+import logging
+
+import flexitest
+
+from common.base_test import StrataNodeTest
+from common.config import EpochSealingConfig, ServiceType
+from common.rpc import JsonRpcClient
+from envconfigs.strata import StrataEnvConfig
+
+logger = logging.getLogger(__name__)
+
+
+def assert_terminal_block_l1_update(
+    rpc: JsonRpcClient,
+    target_slot: int,
+    min_l1_manifests_in_terminal_block: int,
+) -> None:
+    """Asserts a canonical terminal block exposes the expected L1 update."""
+
+    block = rpc.strata_getBlockBySlot(target_slot)
+    assert block is not None, f"terminal slot {target_slot} is missing from canonical chain"
+
+    header = block["header"]
+    assert header["slot"] == target_slot, f"unexpected terminal block header: {header}"
+    assert header["is_terminal"] is True, f"slot {target_slot} is not terminal: {header}"
+
+    l1_update = block.get("l1_update")
+    assert l1_update is not None, f"terminal slot {target_slot} has no L1 update: {block}"
+
+    manifest_count = l1_update.get("manifest_count")
+    assert (
+        isinstance(manifest_count, int) and manifest_count >= min_l1_manifests_in_terminal_block
+    ), (
+        f"terminal slot {target_slot} included {manifest_count!r} L1 manifests, "
+        f"expected at least {min_l1_manifests_in_terminal_block}: {block}"
+    )
+
+    status = rpc.strata_getChainStatus()
+    latest = status.get("latest")
+    assert isinstance(latest, dict), f"chain status missing latest epoch: {status}"
+    assert latest["last_slot"] >= target_slot, (
+        f"latest completed epoch did not advance past terminal slot {target_slot}: {status}"
+    )
+
+    logger.info(
+        "asserted terminal slot %s with %s L1 manifests",
+        target_slot,
+        manifest_count,
+    )
+
+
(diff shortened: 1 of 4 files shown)


functional-tests/tests/strata/helpers.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+class TerminalBlockAssemblyBase(StrataNodeTest):
+    """Base class for terminal block assembly checks."""
+
+    slots_per_epoch = SLOTS_PER_EPOCH
+    l1_interval_seconds: float
+    terminals_to_validate: int
+    timeout_seconds: int
+    min_manifest_count = 1
```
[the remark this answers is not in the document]
krsnapaudel: Tests have passed.

storopoli (approved):
ACK e3cc71174f76da88faca77773e71a385497782c1

functional-tests/tests/strata/helpers.py
```diff
@@ -0,0 +1,138 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def validate_terminal_block(
+    rpc: JsonRpcClient,
+    target_slot: int,
+    min_manifest_count: int,
+) -> None:
+    """Validates a canonical terminal block and its L1 update."""
```
delbonis: Can you clarify what you mean by "validate" here?  Is it just checking that it "correctly" looks like a terminal block?  Should change name to be more precise.

delbonis (changes requested):
Question on choice of phrasing.

functional-tests/tests/strata/helpers.py
```diff
@@ -0,0 +1,138 @@
...
+    def main(self, ctx):
+        bitcoin = self.get_service(ServiceType.Bitcoin)
+        strata = self.get_service(ServiceType.Strata)
+        rpc = strata.wait_for_rpc_ready(
+            timeout=self.rpc_ready_timeout,
+            method="strata_getChainStatus",
+        )
+
+        actual_slots_per_epoch = strata.props["slots_per_epoch"]
+        assert actual_slots_per_epoch == self.slots_per_epoch, (
```
bewakes: Why is `self.slots_per_epoch` needed if we are asserting the equality here and are already obtaining values from `strata.props["slots_per_epoch"]`?

functional-tests/tests/strata/helpers.py
```diff
@@ -0,0 +1,138 @@
...
+    )
+
+    logger.info(
+        "validated terminal slot %s with %s L1 manifests",
+        target_slot,
+        manifest_count,
+    )
+
+
+def mine_until_terminal_block(
```
bewakes: nit: The name is misleading. It secretly does the validation as well, so I suggest reflecting that on the name.

bewakes: Also, the function assumes the passed block is terminal. It might better to frame it as `validate/mine_until_next_terminal`, pass in the current slot and compute the terminal here.

bewakes (approved):
ACK. Some nits.

functional-tests/tests/strata/helpers.py
krsnapaudel: @delbonis This helper is only asserting RPC-visible test invariants for the terminal block and its L1 update. So renamed it to `assert_terminal_block_l1_update`.

delbonis (approved):

PR #1878, opened 2026-05-28
fix(rpc): source L1 tip height from ASM state instead of status channel

krsnapaudel:
## Description

`get_checkpoint_info` compared the persisted checkpoint observation height against `L1Status.cur_height`, which starts at 0 on restart and lags reader progress. Spurious "L1 tip below observed height" errors followed.

Switch the source to the latest persisted ASM state. It survives restarts and advances in lockstep with the checkpoint observation, so `tip >= observed_height` holds by construction.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-3672](https://alpenlabs.atlassian.net/browse/STR-3672)

[STR-3672]: https://alpenlabs.atlassian.net/browse/STR-3672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/strata/src/rpc/provider.rs
+++ b/bin/strata/src/rpc/provider.rs
@@ -188,8 +188,13 @@ impl OLRpcProvider for NodeRpcProvider {
         self.status_channel.get_ol_sync_status()
     }
 
-    fn get_l1_tip_height(&self) -> Option<L1Height> {
-        Some(self.status_channel.get_l1_status().cur_height)
+    async fn get_l1_tip_height(&self) -> DbResult<Option<L1Height>> {
+        Ok(self
+            .storage
+            .asm()
+            .fetch_most_recent_state_async()
+            .await?
+            .map(|(commit, _)| commit.height()))
     }
 
     async fn submit_transaction(&self, tx: OLTransaction) -> OLMempoolResult<OLTxId> {
--- a/crates/storage/src/managers/asm.rs
+++ b/crates/storage/src/managers/asm.rs
@@ -39,12 +39,16 @@ impl AsmStateManager {
     }
 
     /// Returns [`AsmState`] that corresponds to passed block.
-    pub fn get_state(&self, block: L1BlockCommitment) -> DbResult<Option<AsmState>> {
+    pub fn get_state_blocking(&self, block: L1BlockCommitment) -> DbResult<Option<AsmState>> {
         self.ops.get_asm_state_blocking(block)
     }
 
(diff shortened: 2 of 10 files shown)


storopoli (approved):
ACK 6c03e250e40f845329d0de5ebc0132cb2bb174c2

crates/storage/src/managers/asm.rs
```diff
@@ -29,6 +29,13 @@ impl AsmStateManager {
         self.ops.get_latest_asm_state_blocking()
     }
 
+    /// Returns [`AsmState`] that corresponds to the "highest" block.
+    pub async fn fetch_most_recent_state_async(
+        &self,
+    ) -> DbResult<Option<(L1BlockCommitment, AsmState)>> {
+        self.ops.get_latest_asm_state_async().await
+    }
```
delbonis, line 39: Should rename the other one to have `_blocking` so we don't make the above mistake that was made in node.rs of using the blocking call from the async fn.

bin/strata/src/rpc/provider.rs
```diff
@@ -188,8 +188,14 @@ impl OLRpcProvider for NodeRpcProvider {
...
-    fn get_l1_tip_height(&self) -> Option<L1Height> {
-        Some(self.status_channel.get_l1_status().cur_height)
+    async fn get_l1_tip_height(&self) -> Option<L1Height> {
+        self.storage
+            .asm()
+            .fetch_most_recent_state_async()
+            .await
+            .ok()
+            .flatten()
+            .map(|(commit, _)| commit.height())
```
delbonis: If there's an IO error we should propagate that instead of making it seem like there isn't a tip.

delbonis (changes requested):

bewakes (approved):
Great to see minimization of the use of status channel!

Concur with Trey on IO error handling. The rest looks good.

crates/storage/src/managers/asm.rs
krsnapaudel: @delbonis I renamed other similar ones with `_blocking` for consistency.

purusang (approved):

delbonis (approved):
lgtm

PR #1879, opened 2026-05-28
fix(alpen-client): publish finalized-only OL tracker updates

storopoli:
## Description

Fixes STR-3634 by making the Alpen EE OL tracker notify watchers when only the finalized OL epoch advances while the confirmed epoch is already synced.

Previously, `track_ol_state` treated this as a no-op. That meant the Alpen block builder did not observe the finalized checkpoint transition that should make a deposit mintable, so funds were only minted after a later confirmed checkpoint woke the tracker again.

This PR:

- adds a finalized-only tracker refresh path
- publishes watcher updates after finalized-only refreshes

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

AI was used to debug staging and help address the fix.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-3634

### DIFF
--- a/crates/alpen-ee/ol-tracker/src/service.rs
+++ b/crates/alpen-ee/ol-tracker/src/service.rs
@@ -16,7 +16,7 @@ use crate::{
     error::OLTrackerError,
     reorg::handle_reorg,
     state::OLTrackerState,
-    task::{handle_extend_ee_state, track_ol_state, TrackOLAction},
+    task::{handle_extend_ee_state, handle_refresh_finalized, track_ol_state, TrackOLAction},
 };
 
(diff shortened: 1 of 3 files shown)


sapinb (commented):
Fix looks good. But the new fn test fails consistently.

storopoli:
Yeah, I've dropped the functional test. We already have one unit test.

delbonis (approved):

bewakes:
Thanks for this fix @storopoli. It's already merged but it would be ideal to investigate why the functional test failed.

PR #1883, opened 2026-05-30
fix(alpen-ee): avoid invalid finalized forkchoice updates

peter941221:
## Description

Fixes `forkchoiceUpdated` handling when OL reports a finalized EE hash that this Reth node still sees as non-canonical.

Before this change, `crates/alpen-ee/engine/src/control.rs` canonical-checked `confirmed` / safe but forwarded `finalized` unchanged. During sync or fork-switch windows, that could send a non-canonical finalized hash into `forkchoiceUpdated`, which Reth rejects as `invalid forkchoice state`.

This patch canonical-checks `finalized_block_hash`, clears it to `B256::ZERO` when it is inconsistent with the chosen fork, and keeps finalized when the update is switching to OL's fork and the finalized block is an ancestor of the new target head. On the engine side it also stops retrying deterministic forkchoice validation failures while still retrying communication-style failures.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The serialized interface is unchanged.

The key behavioral boundaries are:
- finalized now drops to zero only when it is inconsistent with the chosen fork
- a finalized block on the OL target fork stays preserved across the fork switch
- `ExecutionEngineError::ForkChoiceUpdate(...)` no longer goes through the retry loop

- [ ]  Yes
- [x]  No

## Checklist

- [x] I have disclosed my use of AI in the body of this PR.

### DIFF
--- a/crates/alpen-ee/engine/Cargo.toml
+++ b/crates/alpen-ee/engine/Cargo.toml
@@ -12,6 +12,7 @@ alpen-reth-node.workspace = true
 strata-acct-types.workspace = true
 strata-common = { workspace = true, features = ["async"] }
 
+alloy-consensus.workspace = true
 alloy-primitives.workspace = true
 alloy-rpc-types-engine.workspace = true
 async-trait.workspace = true
--- a/crates/alpen-ee/engine/src/control.rs
+++ b/crates/alpen-ee/engine/src/control.rs
@@ -1,12 +1,13 @@
 use std::future::Future;
 
+use alloy_consensus::BlockHeader;
 use alloy_primitives::B256;
 use alloy_rpc_types_engine::ForkchoiceState;
 use alpen_ee_common::{BlockNumHash, ConsensusHeads, ExecutionEngine};
 use reth_node_builder::NodeTypesWithDB;
 use reth_provider::{
     providers::{BlockchainProvider, ProviderNodeTypes},
-    BlockHashReader, BlockNumReader, ProviderResult,
+    BlockHashReader, BlockNumReader, HeaderProvider, ProviderResult,
 };
 use tokio::{select, sync::watch};
 use tracing::{error, warn};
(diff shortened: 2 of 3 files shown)


peter941221:
Re-ran the engine-side validation after the final test-only follow-up commit.

Environment
- WSL2 temporary verification copy
- normalized copied crates/**/ssz/* inputs to LF in the Linux copy only
- branch head includes 5914edb1 test(alpen-ee): tighten forkchoice retry assertion

Commands
1. cargo test -p alpen-ee-engine update_consensus_state_maps_invalid_state_to_forkchoice_error_without_retry -- --nocapture
2. cargo test -p alpen-ee-engine update_consensus_state_retries_engine_unavailable_errors -- --nocapture
3. cargo test -p alpen-ee-engine -- --nocapture

Results
1. update_consensus_state_maps_invalid_state_to_forkchoice_error_without_retry: passed
2. update_consensus_state_retries_engine_unavailable_errors: passed
3. alpen-ee-engine package tests: 22 passed, 0 failed

The follow-up test change does not touch production logic. It only tightens the retry assertion from a sleep-and-abort shape to a bounded fail-once-then-succeed path with an exact attempt count.

peter941221:
Addressed the finalized-handling bug on fork switches.

Problem
- the previous patch evaluated finalized against the pre-update canonical view
- when OL safe was non-canonical and this update switched head to that OL fork, a finalized block on the same OL fork could be dropped to zero even though it would become valid after the head switch

Change
- when the update switches to OL's safe fork, keep finalized if it is an ancestor of the chosen target head on that fork
- still drop finalized when it belongs to a different branch
- kept the existing non-switching behavior for the current canonical fork

Validation
- cargo test -p alpen-ee-engine forkchoice_keeps_noncanonical_finalized_when_switching_to_same_ol_fork -- --nocapture
- cargo test -p alpen-ee-engine forkchoice_drops_finalized_from_different_noncanonical_fork -- --nocapture
- cargo test -p alpen-ee-engine -- --nocapture

Results
- both new targeted tests passed
- alpen-ee-engine package tests: 24 passed, 0 failed

This required adding alloy-consensus as a direct dependency for alpen-ee-engine so the header trait used by the provider-backed parent walk is in scope in this crate.

peter941221:
This branch is unchanged since the last test-only follow-up and the earlier review thread is resolved.

If someone wants another look, the latest head already includes the finalized fork follow-up and the package test rerun.

PR #1926, opened 2026-06-02
STR-3701: Reconcile stale checkpoint artifacts on startup

voidash:
## Summary
- Reconcile local checkpoint artifacts against ASM verified checkpoint tip during strata startup
- Delete unaccepted checkpoint payloads, checkpoint proofs, and checkpoint prover tasks before checkpoint/prover services start
- Expose prover task deletion through the storage manager

Jira: STR-3701

### DIFF
--- a/bin/strata/src/checkpoint_reconcile.rs
+++ b/bin/strata/src/checkpoint_reconcile.rs
@@ -28,8 +28,8 @@ pub(crate) fn reconcile_unaccepted_checkpoint_artifacts(nodectx: &NodeContext) -
 
     let deleted_payloads = storage
         .ol_checkpoint()
-        .del_checkpoint_payload_entries_from_epoch_blocking(first_unaccepted_epoch)
-        .context("delete unaccepted checkpoint payloads")?;
+        .del_local_checkpoint_payload_entries_from_epoch_blocking(first_unaccepted_epoch)
+        .context("delete unaccepted local checkpoint payloads")?;
     extend_missing(&mut cleanup_commitments, deleted_payloads.iter().copied());
 
     let mut deleted_proofs = 0usize;
--- a/bin/strata/src/checkpoint_reconcile.rs
+++ b/bin/strata/src/checkpoint_reconcile.rs
@@ -1,0 +1,146 @@
+//! Reconciles local checkpoint artifacts against ASM-accepted state.
+
+use anyhow::{Context, Result};
+use strata_asm_common::Subprotocol;
+use strata_asm_proto_checkpoint::CheckpointSubprotocol;
+use strata_checkpoint_types::CheckpointProofTask;
+use strata_identifiers::{Epoch, EpochCommitment};
+use strata_node_context::NodeContext;
+use tracing::{debug, info};
+
+/// Deletes local checkpoint artifacts after ASM's accepted checkpoint tip.
+///
+/// Checkpoint payloads, proofs, and prover tasks past the ASM verified tip are
+/// local candidate state. Rebuilding them on startup prevents a rotated OL
+/// image from reusing stale pre-rotation proof artifacts.
+pub(crate) fn reconcile_unaccepted_checkpoint_artifacts(nodectx: &NodeContext) -> Result<()> {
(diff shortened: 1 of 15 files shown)


bin/strata/src/checkpoint_reconcile.rs
```diff
@@ -0,0 +1,146 @@
...
+    };
+
+    let storage = nodectx.storage();
+    let mut cleanup_commitments =
+        checkpoint_commitments_from_epoch(nodectx, first_unaccepted_epoch)?;
+
+    let deleted_payloads = storage
+        .ol_checkpoint()
+        .del_checkpoint_payload_entries_from_epoch_blocking(first_unaccepted_epoch)
+        .context("delete unaccepted checkpoint payloads")?;
```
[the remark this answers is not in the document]
voidash: Addressed in f078fa356. Reconcile now uses a local-payload cleanup path that deletes locally-built payload/signing/unsigned state while preserving L1-observed checkpoint payloads and `l1_ref_tree` entries. Added a regression test covering an observed checkpoint that survives local cleanup.

Validation:
- `cargo fmt -- --check`
- `git diff --check`
- `cargo check -p strata-storage`
- `cargo test -p strata-db-store-sled ol_checkpoint --lib`
- `cargo check -p strata`

bin/strata/src/services.rs
```diff
@@ -261,6 +262,7 @@ pub(crate) fn start_strata_services(
         nodectx.ol_params(),
         nodectx.status_channel().as_ref(),
     )?;
+    reconcile_unaccepted_checkpoint_artifacts(&nodectx)?;
```
storopoli, line 271: After reconciliation deletes the unaccepted checkpoint payloads/proofs/tasks here, nothing appears to re-drive the already-persisted epoch summaries. On restart the chain-worker watch channel is initialized as `None` and only sends from `store_summary` when a terminal block is executed, while the checkpoint worker skips that initial `None` and the prover runner initializes `latest_epoch` from the same receiver as `0`. So if the node restarts specifically to clean stale artifacts and no later terminal epoch arrives, the deleted unaccepted checkpoint is not rebuilt until some future epoch completes.

Can we seed catch-up from storage after this cleanup, or initialize the checkpoint/prover runners from `get_last_summarized_epoch` instead of waiting only on new watch notifications?

Refs: https://github.com/alpenlabs/alpen/blob/f078fa35645e3ef6b4fd8a1e31149153d5d0676d/bin/strata/src/services.rs#L265, https://github.com/alpenlabs/alpen/blob/f078fa35645e3ef6b4fd8a1e31149153d5d0676d/crates/chain-worker-new/src/context.rs#L89, https://github.com/alpenlabs/alpen/blob/f078fa35645e3ef6b4fd8a1e31149153d5d0676d/crates/chain-worker-new/src/context.rs#L262, https://github.com/alpenlabs/alpen/blob/f078fa35645e3ef6b4fd8a1e31149153d5d0676d/crates/ol/checkpoint/src/service.rs#L47-L50, https://github.com/alpenlabs/alpen/blob/f078fa35645e3ef6b4fd8a1e31149153d5d0676d/bin/strata/src/prover/mod.rs#L199-L200

storopoli (changes requested):

delbonis (approved):

bin/strata/src/services.rs
voidash: Fixed in 60f15b9b7.

The checkpoint worker now drives a one-shot catch-up in `on_launch` (rebuilding payloads for already-summarized epochs), and the prover runner seeds `latest_epoch` from `get_last_summarized_epoch` instead of `0`. So after reconcile clears unaccepted payloads/proofs/tasks, they are rebuilt and re-proven on restart without waiting for a new terminal epoch.

voidash:
All review threads are resolved and CI is green on current head `97d94bf`.  Could you re-review when you get a chance? @storopoli

storopoli (approved):
ACK 97d94bfe3eb1ccf94b6df4ec73858136aa37c831

PR #1887, opened 2026-05-31
refactor: remove the legacy TN1 (Chainstate-based) stack

prajwolrg:
## Description

`bin/strata` now runs entirely on the new OLState fork-choice path (`start_fcm_service` over `strata-ol-chain-types-new`), so the original `Chainstate`/`L2Block` TN1 stack is dead code that was only kept compiling by lingering dependencies. This is the first cleanup pass for STR-2170: it removes the two foundational legacy crates this branch started from (`ol-chain-types`, `ol-chainstate-types`) along with their entire reverse-dependency closure that nothing live needs.

**Deleted crates:** `eectl`, `evmexec`, `chainexec`, `chaintsn`, `sequencer`, `chain-worker`, `sync`, the legacy `rpc/{api,types,utils}`, `test-utils/evm-ee`, plus the two foundational `ol-chain-types` / `ol-chainstate-types`.

**Stripped legacy code from surviving crates:** the legacy fork-choice/worker glue in `consensus-logic`, the L2-block / `Chainstate` / legacy-checkpoint DB layers across `db-types` / `db/store-sled` / `db/tests` / `storage`, the producer-less `Chainstate` field in `status`, and the `l2_database` benchmark.

**Three surviving crates referenced the deleted types and were repointed rather than removed:**

- `ExecSegment` is inlined into `proof-impl/evm-ee-stf` (still consumed via `reth/db`), preserving its Borsh/Serde proof-output layout.
- `strata-dbtool` now imports `AsmManifest` from `strata-ol-chain-types-new` (which re-exports it from `strata-asm-common`).
- `test-utils/l2` keeps only its non-legacy helpers.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This is delivered as a single commit because the removal is tightly coupled — deleting the foundational crates breaks the build until the whole closure is gone, so no smaller split compiles on its own. `cargo check --workspace` is green at HEAD.

[...]

PR #752, opened 2025-03-27
strata alpen rename

sapinb:
## Description
Rename alpen-evm specific code

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have commented my code where necessary.
- [ ] I have updated the documentation if needed.
- [ ] My changes do not introduce new warnings.
- [ ] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [ ] New and existing tests pass with my changes.

PR #1928, opened 2026-06-02
refactor!: remove RollupParams and the strata-params crate

prajwolrg:
## Description

The legacy TN1 (Chainstate-based) stack removal (#1887) left `RollupParams` / `Params` / `SyncParams` (the `strata-params` crate) almost entirely redundant with the newer `AsmParams` and `OLParams` artifacts. The node still loaded a `rollup-params.json` and threaded `Arc<Params>` through `NodeContext`, but only a handful of fields were ever read at runtime — and every one of those, except `l1_reorg_safe_depth`, already exists in `AsmParams`.

This PR removes `RollupParams`, `Params`, `SyncParams`, and the entire `strata-params` crate, re-sourcing the few live reads:

- **magic bytes, genesis L1 block/height, network** → `AsmParams` (`magic`, `anchor.block`, `anchor.network`)
- **sequencer pubkey** (checkpoint-envelope signing + RPC reporting) → the ASM checkpoint `sequencer_predicate`
- **`l1_reorg_safe_depth`** → a new `BtcioConfig` field (operational knob; no `AsmParams` equivalent)
- **alpen-cli** now reads `asm-params.json` for network/magic/denomination/recovery_delay

`SyncParams` was never read at runtime (its values flow from `Config`), and `checkpoint-types`' `verify_signed_checkpoint_sig` (the only real `CredRule` consumer) had no callers — both are dropped, along with the `--rollup-params` CLI flag.

datatool's `genparams` generator is removed; its shared genesis helpers move to `cmd/genesis_info.rs`, and a new `gen-checkpoint-predicate` subcommand exposes the checkpoint VK that `genparams` previously surfaced. The functional-test harness, docker deployment scripts, and the CI/CD VK-extraction workflows are updated to match.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests

## Notes to Reviewers

Commits are split by concern: core Rust removal (including datatool re-sourcing the genesis anchor from the ASM `L1Anchor` instead of the old `GenesisL1View`, and renaming the `genl1view` / `--genesis-l1-view-file` surface to `gen-l1-anchor` / `--l1-anchor-file`), functional-test harness, docker deployment, and the CI params / VK-extraction workflows.

`l1_reorg_safe_depth` defaults to `6` in `BtcioConfig`; the functional-test and docker configs set it explicitly so the node and dbtool stay in agreement. The terminal-block assembly tests additionally pin the depth they rely on (via the strata factory's `BtcioConfig`, written into `config.toml`) so their buried-manifest timing isn't coupled to — and can't be silently broken by — that default.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-2168

---

_AI assistance: this change was implemented with Claude Code (Opus 4.8)._


prajwolrg:
@codex review

crates/params/src/lib.rs
```diff
@@ -1,201 +0,0 @@
...
-            return Err(ParamsError::NoOperators);
-        }
-
-        // TODO maybe make all these be a macro?
-        if self.block_time == 0 {
-            return Err(ParamsError::ZeroProperty("block_time"));
-        }
-
-        if self.l1_reorg_safe_depth == 0 {
-            return Err(ParamsError::ZeroProperty("l1_reorg_safe_depth"));
```
purusang, line 82: Should we have this check somewhere? Looks like a regression.

docker/strata/entrypoint.sh
```diff
@@ -128,6 +120,12 @@ fi
 BITCOIN_NETWORK="${BITCOIN_NETWORK:-regtest}"
 CONFIG_OVERRIDES="${CONFIG_OVERRIDES} -o bitcoind.network=${BITCOIN_NETWORK}"
 
+# L1 reorg-safe depth now lives in the node's [btcio] config (it used to be a
```
purusang: If that is the case then why not update config file itself instead of ENV? It requires a complicated hack to inject Env var to a config file.

docker/strata/entrypoint.sh
```diff
@@ -76,30 +79,19 @@ if [ -n "${BITCOIND_RPC_URL}" ] && [ "${CURRENT_GENESIS_HEIGHT}" -eq 0 ]; then
...
-            '.l1_view.blk.height = $h | .l1_view.blk.blkid = $id' \
-            "${ASM_PARAMS_PATH}" > "${PATCHED_ASM}"
-        ASM_PARAMS_PATH="${PATCHED_ASM}"
-    fi
+    # Patch ol-params.json to mirror the same genesis L1 block.
+    PATCHED_OL="/app/data/ol-params.json"
+    jq --argjson h "${TIP_HEIGHT}" --arg id "${TIP_HASH}" \
+        '.last_l1_block.height = $h | .last_l1_block.blkid = $id' \
+        "${OL_PARAMS_PATH}" > "${PATCHED_OL}"
+    OL_PARAMS_PATH="${PATCHED_OL}"
```
purusang: I do not think its a good idea to patch params in entrypoint here, specially ol-params since genesis OL blkid is a function of both l1 height, blkid, and others so change in here should also be reflected to  asm params as well. Ideally, lets remove patching steps altogether.

purusang (changes requested):
Much needed refactor. Minor flags!

storopoli (approved):
ACK 1b457f71fff5b33687b2387ea3cf155437eca42a

storopoli:
You need to add STR-XXXX tickets to the new TODOs.

krsnapaudel (approved):
LGTM

crates/params/src/lib.rs
prajwolrg: A depth of `0` is intentional rather than a regression: it means the chain follows the L1 tip with no reorg buffer (a checkpoint finalizes as soon as its L1 block is the tip), so we don't want to reject it. The runtime already clamps the finality depth to a 1-block minimum and the field now carries a `#[serde(default)]` of 6, so a misconfiguration can't silently break finality. Rather than re-add the check I documented the `0` semantics on `BtcioConfig::l1_reorg_safe_depth` — abcf09673.

docker/strata/entrypoint.sh
prajwolrg: Good call — baked it into the committed config instead of round-tripping an env var. `l1_reorg_safe_depth` is now set directly under `[btcio]` in `config.seq.toml`, and the `-o btcio.l1_reorg_safe_depth` override is gone from the entrypoint. `L1_REORG_SAFE_DEPTH` stays only for the alpen-client (EE), which takes it as a real `--l1-reorg-safe-depth` CLI flag. fc0bd5b74

docker/strata/entrypoint.sh
prajwolrg: Agreed — removed the runtime genesis patching entirely. Params are now generated with their final L1 anchor at init time (pinned to `GENESIS_L1_HEIGHT` via `datatool gen-l1-anchor`), so the asm and ol params stay consistent by construction and the node consumes them as-is; init-network.sh's comments/warnings are updated to match. The default signet flow already produced a real anchor (height > 0) and skipped patching, so the only behavior change is regtest with `GENESIS_L1_HEIGHT=0`, which now pins genesis to L1 block 0 instead of patching to the tip. fc0bd5b74

docker/configs/config.seq.toml
```diff
@@ -24,6 +24,11 @@ rpc_user = "rpcuser"
 rpc_password = "rpcpassword"
 network = "regtest"
 
+[btcio]
+# Depth, in L1 blocks, after which an L1 block is considered safe from reorgs.
+# Set to 0 to follow the L1 tip with no reorg buffer.
+l1_reorg_safe_depth = 4
```
purusang, line 30:
```suggestion
l1_reorg_safe_depth = 6
```

purusang (approved):
LGTM

bewakes (approved):
This is great! alpen is getting fitter day by day shedding lines of code.

Just a note that deployment values/configs might have to change when this is included in the next deployment.
cc: @alexhui01

voidash (approved):

crates/primitives/src/l1/mod.rs
```diff
@@ -1,3 +1,68 @@
...
 pub use strata_btc_types::{payload, *};
 pub use strata_identifiers::{L1BlockCommitment, L1BlockId, L1Height};
+
+/// Number of confirmations an L1 block has under `current_tip`, counting the
+/// block itself as one confirmation.
+///
+/// A block at the tip has 1 confirmation; one block below tip has 2; etc.
+/// Observation heights above the tip saturate to 0.
+pub fn l1_confirmations(observed_height: L1Height, current_tip: L1Height) -> u32 {
+    if observed_height > current_tip {
```
delbonis: This should be named a little more self-explanatorily, like `compute_confirmation_depth`

Also, if `observed_height > current_tip`, that should probably return `None` since it's not actually confirmed in that case.

bin/alpen-cli/src/settings.rs
```diff
@@ -99,10 +107,10 @@ pub static CONFIG_FILE: LazyLock<PathBuf> = LazyLock::new(|| match var("CLI_CONF
...
 });
 
-pub static ROLLUP_PARAMS_FILE: LazyLock<PathBuf> =
+pub static ASM_PARAMS_FILE: LazyLock<PathBuf> =
     LazyLock::new(|| match var("STRATA_NETWORK_PARAMS").ok() {
-        Some(path) => PathBuf::from_str(&path).expect("valid rollup params path"),
-        None => PROJ_DIRS.config_dir().to_owned().join("rollup_params.json"),
+        Some(path) => PathBuf::from_str(&path).expect("valid asm params path"),
+        None => PROJ_DIRS.config_dir().to_owned().join("asm-params.json"),
     });
```
delbonis: Should define these at top of file.

bin/alpen-cli/src/settings.rs
delbonis, line 95: I'd think this shouldn't be in here, but an enum instance defining how to init it that we then use later.

bin/alpen-cli/src/settings.rs
```diff
@@ -84,7 +85,14 @@ pub struct Settings {
...
-    pub params: RollupParams,
+    /// L1 network, sourced from the ASM anchor.
+    pub network: Network,
+    /// SPS-50 magic bytes, sourced from the ASM params.
+    pub magic_bytes: MagicBytes,
+    /// At-rest deposit denomination, sourced from the ASM Bridge subprotocol config.
+    pub deposit_amount: Amount,
+    /// Deposit-request reclaim delay, sourced from the ASM Bridge subprotocol config.
+    pub recovery_delay: u16,
     #[cfg(feature = "test-mode")]
```
delbonis, line 107: Some of these seem like they should still be in a separate params file (baked into the binary?) since they're not really user-configurable params.

delbonis (dismissed):
Sorry this took a while to review, but this is really great.  I'm happy to see this cleanup/rework happening.

Just a few notes.

crates/primitives/src/l1/mod.rs
prajwolrg: Done in c8b0257ee — renamed to `compute_confirmation_depth` and it now returns `None` when `observed_height > current_tip` instead of saturating to 0. `is_l1_reorg_safe` treats `None` as not reorg-safe, and the checkpoint-sync caller defaults the `NotReorgSafe` error depth to 0.

bin/alpen-cli/src/settings.rs
prajwolrg: Done in 422c33042 — lifted the env-var keys (`PROJ_DIRS` / `CLI_CONFIG` / `STRATA_NETWORK_PARAMS`) and the default file names (`config.toml` / `asm-params.json`) into named `const`s at the top of the file.

bin/alpen-cli/src/settings.rs
prajwolrg: Agreed this would be cleaner. Since `signet_backend` is pre-existing (from #752) and the enum refactor touches ~10 `cmd/*` call sites, I kept it out of this PR to stay focused on the RollupParams removal and filed [STR-3728](https://alpenlabs.atlassian.net/browse/STR-3728) to track it.

bin/alpen-cli/src/settings.rs
prajwolrg: These already come from `asm-params.json` (loaded separately from the user's `config.toml`), so they aren't user-configurable knobs — but you're right that flattening them into `Settings` alongside real config blurs the line. Filed [STR-3729](https://alpenlabs.atlassian.net/browse/STR-3729) to track splitting the ASM-derived params into their own type. (Baking into the binary is probably not what we want since these differ per network, e.g. testnet vs mainnet.)

purusang (approved):

PR #1931, opened 2026-06-03
fix(alpen-cli): reserve gas for alpen drain

voidash:
## Summary
- reserve Alpen drain gas using wide `U256` arithmetic
- submit the drain transaction with the same explicit `gas_limit` and `gas_price` used for the balance calculation
- add focused tests for exact-fee, below-fee, and wide-arithmetic cases

## Bug Encountered
During the local k2 `alpen-cli` smoke run, `alpen drain --alpen-address 0x000000000000000000000000000000000000c0Fe` failed after the CLI wallet had been funded and had already sent an Alpen transfer. The RPC rejected the drain transaction with:

```text
insufficient funds for gas * price + value: have 998966468221608000 want 998979000000000000
```

The old code estimated gas for a 1-wei transaction, computed `balance - gas_estimate * gas_price`, then submitted a different max-value transaction while leaving the provider to fill gas fields again. That allowed the submitted transaction's upfront gas requirement to exceed the amount reserved by the drain calculation.

This PR fixes that by calculating the reserved fee with `U256`, returning early when there is no spendable balance after gas, and setting `gas_limit` and `gas_price` explicitly on the submitted drain transaction.

### DIFF
--- a/bin/alpen-cli/src/cmd/drain.rs
+++ b/bin/alpen-cli/src/cmd/drain.rs
@@ -153,52 +153,51 @@ pub async fn drain(
             .internal_error("Failed to fetch Alpen balance")?;
         if balance == U256::ZERO {
             println!("No Alpen bitcoin to send");
+        } else {
+            let estimate_tx = l2w
+                .transaction_request()
+                .from(l2w.default_signer_address())
+                .to(address)
+                .value(U256::from(1));
+
+            let gas_price = l2w
+                .get_gas_price()
+                .await
+                .internal_error("Failed to fetch Alpen gas price.")?;
+            let gas_estimate = l2w
+                .estimate_gas(estimate_tx)
+                .await
+                .internal_error("Failed to estimate Alpen gas")?;
+
+            if let Some(max_send_amount) = max_alpen_drain_value(balance, gas_estimate, gas_price) {
+                let tx = l2w
+                    .transaction_request()
+                    .to(address)
+                    .value(max_send_amount)
+                    .gas_limit(gas_estimate)
+                    .gas_price(gas_price);
+
+                let res = l2w
+                    .send_transaction(tx)
+                    .await
+                    .internal_error("Failed to broadcast strata transaction")?;
+
+                println!(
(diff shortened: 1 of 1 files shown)


storopoli (approved):
ACK cc7c5191fa91e9701f0c351be6bbe4a573f6ac1e

barakshani:
We seem to have issues with gas calculation in general, not only in drain -- I'm not sure if this general issue is only when we haven't had sufficient activity (tx history is totally empty or in some recent block range). I think we just don't query what is the minimal gas the network accepts.
Shouldn't we try to fix it more holistically?

voidash:
@barakshani 

> We seem to have issues with gas calculation in general, not only in drain -- I'm not sure if this general issue is only when we haven't had sufficient activity (tx history is totally empty or in some recent block range). I think we just don't query what is the minimal gas the network accepts.
Shouldn't we try to fix it more holistically?

This is only for drain specific invariant so i believe this is acceptable for narrow drain correctness fix. Activity history stuff is not the root issue
Before this pr we didn't set `gas_limit(gas_estimate)` and gas price. We sent with only `to` and `value`.
 This would have been broken even if the chain had lots of activity or no activity.

But i agree that we need something on `AlpenWallet` that is a shared transaction gas helper.

storopoli:
> But i agree that we need something on `AlpenWallet` that is a shared transaction gas helper.

Please open a ticket?

bin/alpen-cli/src/cmd/drain.rs
krsnapaudel, line 155: Does it make sense to return here?

krsnapaudel (approved):
LGTM, although an early return in the flagged line would have made the code readable. 
```
         if balance == U256::ZERO {
            println!("No Alpen bitcoin to send");
            Ok(())
        }
        
        // rest of the code instead of an else branch
```

And a ticket for the follow up work.

storopoli (approved):
ACK cbc97157a4f19c899b714c8d063492048cd56b2e

PR #1942, opened 2026-06-04
STR 3718 add snark acct update manifest rpc

krsnapaudel:
## Description

Adds `strata_getSnarkAcctUpdateManifest(account_id, seq_no)` to `OLClientRpc`. The response includes operation seqno, optional post-update inner state root, consumed inbox range `[prev_next_msg_idx, new_next_msg_idx)`, and optional extra data. Message payloads remain available through `strata_getSnarkAcctInboxMsgRange`.

The lookup scans existing account update records, with an early not-found return when the requested seqno is at or above the account's current tip seqno. Consumed range starts are persisted as `prev_next_inbox_idx`, so RPCs read `[prev_next_inbox_idx, next_inbox_idx)` directly instead of re-threading cursors.

Also updates block-sync/checkpoint-sync indexing to populate the new field and updates account-summary paths to use it.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-3718](https://alpenlabs.atlassian.net/browse/STR-3718)

[STR-3718]: https://alpenlabs.atlassian.net/browse/STR-3718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/strata/src/rpc/node.rs
+++ b/bin/strata/src/rpc/node.rs
@@ -23,7 +23,8 @@ use strata_ol_rpc_types::{
     RpcAccountEpochSummary, RpcAccountState, RpcBlockAccountChanges, RpcBlockEntry,
     RpcBlockHeaderEntry, RpcCheckpointConfStatus, RpcCheckpointInfo, RpcCheckpointL1Ref,
     RpcIndexedEntry, RpcMessageEntry, RpcOLBlockDetail, RpcOLBlockInfo, RpcOLBlockSummary,
-    RpcOLChainStatus, RpcOLTransaction, RpcOLTxDetail, RpcSnarkAccountState, RpcUpdateInputData,
+    RpcOLChainStatus, RpcOLTransaction, RpcOLTxDetail, RpcSnarkAccountState,
+    RpcSnarkAcctUpdateManifest, RpcUpdateInputData,
 };
 use strata_ol_state_types::OLState;
 use strata_primitives::{HexBytes, HexBytes32};
@@ -323,9 +324,8 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
     }
 
     /// Fetches per-(account, epoch) update and inbox records over
-    /// `[first_epoch, last_epoch]`. Updates are grouped by block commitment,
-    /// filtered to `block_commitments`. Out-of-chain and checkpoint-sync
-    /// records still advance the inbox cursor so in-chain slices stay accurate.
+    /// `[first_epoch, last_epoch]`. Updates are grouped by block commitment
+    /// and filtered to `block_commitments`.
     async fn fetch_records_in_epoch_range(
         &self,
         account_id: AccountId,
@@ -333,28 +333,8 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
         last_epoch: Epoch,
         block_commitments: &HashSet<OLBlockCommitment>,
     ) -> RpcResult<AccountRecords> {
-        // Seed the cursor once. If the account isn't found in the prior epoch's
-        // terminal state — either because `first_epoch == 0` (no prior epoch
-        // to query) or because it didn't exist as a snark account at that
-        // point — start at 0. Both cases mean "no prior record of the account
-        // in queryable state, so its inbox starts at 0."
-        let mut cursor: u64 = if first_epoch == 0 {
-            0
-        } else {
-            let (_, prev_ol_state) = self
-                .get_toplevel_ol_state_for_epoch(first_epoch - 1)
-                .await?;
-            prev_ol_state
-                .get_account_state(&account_id)
-                .and_then(|s| s.as_snark_account().ok())
-                .map_or(0, |s| s.next_inbox_msg_idx())
-        };
-
-        // Walk records across all epochs in one pass. Out-of-chain and
-        // checkpoint-sync rows advance the cursor without emitting; in-chain
-        // records emit a `Pending` triple capturing the inbox slice they
-        // consumed. Inbox indexing is monotonic per account across epoch
-        // boundaries, so the cursor flows through without re-seeding.
+        // Walk records across all epochs in one pass. In-chain records emit a
+        // `Pending` triple capturing the inbox slice they consumed.
         struct Pending {
             block_commitment: OLBlockCommitment,
             seq_no: u64,
@@ -374,10 +354,6 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
                 .map_err(db_error)?
             {
                 for r in records {
-                    let prev_cursor = cursor;
-                    let next_cursor = r.next_inbox_idx();
-                    cursor = next_cursor;
-
                     // Skip rows with no block attribution: checkpoint-sync,
                     // or CSS-terminal-stamped rows (root present but block
                     // absent). This endpoint is the block-scoped view.
@@ -412,8 +388,8 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
                         seq_no: r.seq_no(),
                         new_state_root: meta.new_state_root(),
                         extra_data,
-                        cursor_start: prev_cursor,
-                        cursor_end: next_cursor,
+                        cursor_start: r.prev_next_inbox_idx(),
+                        cursor_end: r.next_inbox_idx(),
                     });
                 }
             }
@@ -579,6 +555,19 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
 
         Ok((epoch_commitment, ol_state))
     }
+
+    /// Returns the account's current Snark update seqno at the tip epoch.
+    async fn current_snark_account_seq_no(
+        &self,
+        account_id: AccountId,
+        tip_epoch: Epoch,
+    ) -> RpcResult<Option<u64>> {
+        let (_, tip_ol_state) = self.get_toplevel_ol_state_for_epoch(tip_epoch).await?;
+        Ok(tip_ol_state
+            .get_account_state(&account_id)
+            .and_then(|state| state.as_snark_account().ok())
+            .map(|state| *state.seqno().inner()))
+    }
 }
 
 #[async_trait]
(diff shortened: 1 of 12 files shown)


bin/strata/src/rpc/node.rs
```diff
@@ -23,7 +23,7 @@ use strata_ol_rpc_types::{
     RpcAccountEpochSummary, RpcAccountState, RpcBlockAccountChanges, RpcBlockEntry,
     RpcBlockHeaderEntry, RpcCheckpointConfStatus, RpcCheckpointInfo, RpcCheckpointL1Ref,
     RpcOLBlockDetail, RpcOLBlockInfo, RpcOLBlockSummary, RpcOLChainStatus, RpcOLTransaction,
-    RpcOLTxDetail, RpcSnarkAccountState, RpcUpdateInputData,
+    RpcOLTxDetail, RpcSnarkAccountState, RpcSnarkAcctUpdateManifest, RpcUpdateInputData,
 };
```
delbonis, line 28: Just use a wildcard import at this point.

bin/strata/src/rpc/node.rs
```diff
@@ -558,6 +545,84 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
...
+        let extra_data = record
+            .extra_data()
+            .ok_or_else(|| {
+                internal_error(format!(
+                    "update record for account {account_id} epoch {epoch} seq_no {} \
+                     has no extra_data (DirectSet)",
+                    record.seq_no()
+                ))
+            })?
+            .to_vec();
```
delbonis: This *shouldn't* happen but it would be reasonable to make the field be `Option<HexBytes>` and set it to `None`/`null`.

bin/strata/src/rpc/node.rs
```diff
@@ -558,6 +545,84 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
...
+        // Checkpoint-sync rows currently have no update metadata. This RPC
+        // requires the post-update state root, so those rows cannot produce
+        // a manifest response.
+        let meta = record.update_meta().ok_or_else(|| {
+            internal_error(format!(
+                "update record for account {account_id} epoch {epoch} seq_no {} \
+                 has no update metadata",
+                record.seq_no()
+            ))
+        })?;
```
delbonis: This is actually a case where we *really should* return the field as `null`.  This is a case we *do* expect to happen (in checkpoint sync nodes), and the field is only expected to be used for doing optional sanity checks, but we still want to expose the data even if the Strata node didn't record a state root.

crates/ol/rpc/types/src/account_summary.rs
```diff
@@ -164,6 +164,66 @@ pub struct RpcUpdateInputData {
...
+    /// Sequence number of the update.
+    seq_no: u64,
+    /// Post-update proof state.
+    proof_state: RpcProofState,
+    /// Extra data posted with this update.
+    extra_data: HexBytes,
+    /// First inbox message index consumed by this update.
+    inbox_start_idx: u64,
+    /// First inbox message index after this update.
+    inbox_end_idx: u64,
```
delbonis: This `inbox_end_idx` field is redundant if we have `ProofState`'s `next_inbox_msg_idx`, they're always the same (or off by 1 or something depending on how you defined this field).

*However*, since we won't always have the inner state root, I think it makes sense to flatten out this struct to just be:

```rs
struct RpcSnarkAcctUpdateManifest {
  seq_no: u64,
  new_inner_state_root: Option<[u8; 32]>,
  prev_next_msg_idx: u64, // this is the first processed message
  new_next_msg_idx: u64, // this-1 is the last processed message
  extra_data: Option<HexBytes>,
}
```

bin/strata/src/rpc/node.rs
```diff
@@ -558,6 +545,84 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
...
+        // stores `upd.seq_no() + 1` as the account's post-update seqno. Index
+        // records capture that post-state value, while the published update
+        // manifest is keyed by `upd.seq_no()`.
+        record.seq_no().checked_sub(1).ok_or_else(|| {
+            internal_error(format!(
+                "update record for account {account_id} epoch {epoch} has invalid \
+                 post-state seq_no 0",
+            ))
+        })
+    }
```
delbonis: This does not take `self` and isn't a constructor for `Self`.

bin/strata/src/rpc/node.rs
```diff
@@ -558,6 +545,84 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
...
+
+    fn update_manifest_from_record(
+        account_id: AccountId,
+        epoch: Epoch,
+        record: &AccountUpdateRecord,
+        operation_seq_no: u64,
+        inbox_start_idx: u64,
+        inbox_end_idx: u64,
+    ) -> RpcResult<RpcSnarkAcctUpdateManifest> {
```
delbonis: Same for this actually.

delbonis (changes requested):

bin/strata/src/rpc/node.rs
delbonis: Also the name of this function is unclear, the `update` prefix implies it's updating a manifest from a record, which doesn't make sense.

bin/strata/src/rpc/node.rs
delbonis: Also the name of this fn is unclear, should be `extract` or something.  Maybe honestly even add a convenience accessor to `AccountUpdateRecord` that exposes this naturally (like `fn orig_acct_seq_no(&self) -> Option<Seqno>`) and then do the error handling in the RPC itself.

bin/strata/src/rpc/node.rs
krsnapaudel: Added a constructor for `RpcSnarkAcctUpdateManifest`: `from_account_update_record`.

delbonis (approved):

bin/strata/src/rpc/node.rs
```diff
@@ -903,6 +898,65 @@ impl<P: OLRpcProvider> OLClientRpcServer for OLRpcServer<P> {
...
+            })?;
+        let tip_epoch = self
+            .provider
+            .get_ol_sync_status()
+            .ok_or_else(|| internal_error("OL sync status not available"))?
+            .tip_epoch;
+        let mut cursor = self
+            .seed_account_inbox_cursor(account_id, creation_epoch)
+            .await?;
+
```
purusang: I think we should put a check of snark account's current seq no. to check if the requested seqno is lesser than accounts current seqno to avoid following nested loops.

purusang (changes requested):

bin/strata/src/rpc/node.rs
```diff
@@ -903,6 +898,65 @@ impl<P: OLRpcProvider> OLClientRpcServer for OLRpcServer<P> {
...
+                if operation_seq_no != seq_no {
+                    continue;
+                }
+
+                return Ok(RpcSnarkAcctUpdateManifest::from_account_update_record(
+                    &record,
+                    operation_seq_no,
+                    prev_next_msg_idx,
+                ));
+            }
```
purusang, line 992: Why are we not indexing `prev_next_msg_idx` if it is important?

crates/ol/rpc/api/src/lib.rs
```diff
@@ -71,6 +71,19 @@ pub trait OLClientRpc {
         &self,
         l1_height: L1Height,
     ) -> RpcResult<Option<HexBytes32>>;
+
+    /// Get published manifest metadata for a Snark account update.
+    ///
+    /// The consumed inbox range is half-open:
+    /// `[prev_next_msg_idx, new_next_msg_idx)`.
+    /// `prev_next_msg_idx` is the first message consumed by this update, and
+    /// `new_next_msg_idx` is the first message not consumed.
```
storopoli, line 92: What consumed inbox range?

```suggestion
    /// The consumed inbox range inside the resulting [`RpcSnarkAcctUpdateManifest`] is half-open:
    /// `[prev_next_msg_idx, new_next_msg_idx)`.
    /// `prev_next_msg_idx` is the first message consumed by this update, and
    /// `new_next_msg_idx` is the first message not consumed.
```

storopoli (approved):
ACK 2b5a9d01bb055e3bf4c2bcbb0462e5c74f9a5388

bin/strata/src/rpc/node.rs
krsnapaudel: @purusang Good point. It seems a few RPCs need this. @delbonis Any objection to adding `prev_next_msg_idx` to `AccountUpdateRecord`? Please check the second commit.

crates/ol/rpc/api/src/lib.rs
krsnapaudel: @storopoli The consumed range is I think consumed by the Snark account update.

storopoli: See how it was unambiguous :)
Anyways a non-blocking docstring nit.

bin/strata/src/rpc/node.rs
delbonis: Yes, we should definitely avoid needing to do a slow walk over the database to be able to give answers like this.

delbonis (approved):

bin/strata/src/rpc/node.rs
```diff
@@ -961,6 +933,70 @@ impl<P: OLRpcProvider> OLClientRpcServer for OLRpcServer<P> {
...
+
+                if operation_seq_no != seq_no {
+                    continue;
+                }
+
+                return Ok(RpcSnarkAcctUpdateManifest::from_account_update_record(
+                    &record,
+                    operation_seq_no,
+                ));
+            }
```
purusang, line 992: I think we need one more indexing to avoid this slow walking entirely, additional indexing account_id+seq_no.

purusang (approved):
Please resolve the comment or create a ticket for it if its good to do in a follow up or not at all.

storopoli (approved):
ACK 0b4fc32e4e3b580072b5026aede35b38e52e1b23

bin/strata/src/rpc/node.rs
krsnapaudel: @purusang Since this RPC is the only beneficiery, I have created [a ticket for it](https://alpenlabs.atlassian.net/browse/STR-3778).
.

Issue #139, opened 2024-07-15
Check for missed sync events when client state worker receives new EventInput


delbonis:
It's theoretically possible that the client sync worker might miss a sync event index, possibly due to maybe some sync issues during startup.  It's not clear, but in any case we can fairly easily handle this case.  When the worker receives an `EventInput` CSM message we should make sure to also execute any sync events that we might have missed between the previous sync event we processed and the one we're being told about.  These events we should process just as we would if we *did* receive the normal even input for them, nothing should be different there.

We should print a warning when we notice this happening, since it is an abnormal case that might signal a sync bug introduced in refactoring, but one that we can recover from well enough.

PR #1944, opened 2026-06-05
fix(btcio): recover stranded canonical L1 height on restart

voidash:
## Problem
btcio writes the L1 canonical-chain entry before notifying the ASM worker. If the
process crashes in that window, restart resumes from `canonical_tip + 1`, so the
stranded tip height has a canonical entry but no ASM anchor state/manifest.

The ASM worker does eventually self-heal this: when the next L1 block arrives it
walks back over Bitcoin header `prev_blockhash` (independent of the canonical DB)
to the last height with a stored anchor state and replays the gap, re-driving the
stranded height. So the stranding is **transient, not permanent** — but it only
clears once a *new* Bitcoin block is mined and submitted. Until then the canonical
tip stays unmaterialized: ~10 min on mainnet, and indefinitely on a quiet chain or
a restarted node that receives no further blocks.

## Tests
- New `test_l1_canonical_write_before_manifest_crash` + debug bail tag
  `btcio_after_l1_canonical_write`. The test asserts the stranded tip is
  materialized *after restart, before mining any further block*, which is exactly
  the window this change closes (without it the tip would wait for the next block).

### DIFF
--- a/crates/btcio/src/reader/query.rs
+++ b/crates/btcio/src/reader/query.rs
@@ -4,7 +4,7 @@ use std::{
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
-use anyhow::bail;
+use anyhow::{bail, Context};
 use bitcoin::{Block, BlockHash, Network};
 use bitcoind_async_client::{corepc_types::model::GetBlockchainInfo, traits::Reader};
 use strata_btc_types::{BlockHashExt, L1BlockIdBitcoinExt};
@@ -83,6 +83,12 @@ enum ReaderError {
     },
 }
 
+#[derive(Debug)]
+struct PollResult {
+    events: Vec<L1Event>,
+    is_at_client_tip: bool,
+}
+
 /// The main task that initializes the reader state and starts reading from bitcoin.
 pub async fn bitcoin_data_reader_task<E: BlockSubmitter>(
     client: Arc<impl Reader>,
@@ -133,6 +139,9 @@ async fn do_reader_task<R: Reader>(
 
     let poll_dur = Duration::from_millis(ctx.config.client_poll_dur_ms as u64);
     let mut state: Option<ReaderState> = None;
+    // Canonical storage advances before block delivery. Re-deliver one reconciled tip on startup
+    // so a crash between those operations cannot strand ASM until another L1 block arrives.
+    let mut startup_tip_delivered = false;
 
     loop {
         let mut status_updates: Vec<L1StatusUpdate> = Vec::new();
@@ -142,10 +151,31 @@ async fn do_reader_task<R: Reader>(
                 Err(err) => {
                     handle_poll_error(&err, &mut status_updates);
                 }
-                Ok(events) => {
-                    // handle events
-                    for ev in events {
+                Ok(result) => {
+                    for ev in result.events {
+                        let submitted_block = matches!(
+                            &ev,
+                            L1Event::BlockData(block_data, _)
+                                if block_data.block_num() >= ctx.btcio_params.genesis_l1_height()
+                        );
                         handle_bitcoin_event(ev, &ctx, event_submitter).await?;
+                        startup_tip_delivered |= submitted_block;
+                    }
+
+                    if !startup_tip_delivered
+                        && result.is_at_client_tip
+                        && reader_state.best_block_idx() >= ctx.btcio_params.genesis_l1_height()
+                    {
+                        let tip = L1BlockCommitment::new(
+                            reader_state.best_block_idx(),
+                            reader_state.best_block().to_l1_block_id(),
+                        );
+                        event_submitter
+                            .submit_block(tip)
+                            .await
+                            .with_context(|| format!("failed to re-submit L1 reader tip {tip}"))?;
+                        info!(%tip, "re-submitted L1 reader tip after startup reconciliation");
+                        startup_tip_delivered = true;
                     }
                 }
             }
@@ -375,7 +405,7 @@ async fn poll_for_new_blocks<R: Reader>(
     ctx: &ReaderContext<R>,
     state: &mut ReaderState,
     status_updates: &mut Vec<L1StatusUpdate>,
-) -> anyhow::Result<Vec<L1Event>> {
+) -> anyhow::Result<PollResult> {
     let chain_info = ctx.client.get_blockchain_info().await?;
     status_updates.push(L1StatusUpdate::RpcConnected(true));
     let client_height = chain_info.blocks as L1Height;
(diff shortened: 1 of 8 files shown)


delbonis (approved):

storopoli:
@codex review

crates/consensus-logic/src/asm_worker_context.rs
```diff
@@ -84,8 +87,17 @@ impl WorkerContext for AsmWorkerCtx {
...
+        let manifest_hash = manifest.compute_hash().into();
+
+        self.l1man.put_block_data(manifest).map_err(conv_db_err)?;
+
+        let mut pending = self.pending_manifest_mmr_leaf.lock().map_err(|_| {
+            WorkerError::Unexpected("ASM manifest MMR pending leaf lock poisoned".to_string())
+        })?;
+        *pending = Some((expected_idx, manifest_hash));
+        Ok(())
     }
```
prajwolrg: This should be handled appropriately by the ASM.

prajwolrg: https://github.com/alpenlabs/asm/pull/139 should handle all the reorgs. we shouldn't require this explicitly.

crates/consensus-logic/src/asm_worker_context.rs
```diff
@@ -114,12 +126,42 @@ impl WorkerContext for AsmWorkerCtx {
...
+                WorkerError::Unexpected(
+                    "ASM manifest MMR append requested without a pending manifest".to_string(),
+                )
+            })?;
+
+        if expected_hash != manifest_hash {
+            return Err(WorkerError::Unexpected(format!(
+                "ASM manifest MMR hash mismatch at height {expected_idx}: expected {expected_hash}, got {manifest_hash}"
+            )));
+        }
```
prajwolrg: This should also be fixed upstream. It is tracked in https://alpenlabs.atlassian.net/browse/STR-3239.

prajwolrg (changes requested):
A lot of these feels like it is already being handled upstream. I don't think this should be needed. Could you retry again and see where the failure happens.

voidash:
Rebased this on current `main` and removed the local ASM/MMR recovery path that Prajwol flagged.

What remains is narrower: btcio no longer mutates/reverts canonical L1 storage on restart. It only checks whether the stored canonical L1 tip has both its ASM manifest and anchor state; if not, it re-submits that canonical tip to ASM. That leaves reorg/replay/idempotency handling in ASM (including the upstream ASM PR #139 behavior) and only fixes the btcio-specific quiet-restart hole where no new Bitcoin block arrives to trigger ASM.

Validation:
- `cargo fmt --all -- --check`
- `cargo check -p strata-btcio --all-features`
- targeted functional test: `test_l1_canonical_write_before_manifest_crash`
- full GitHub CI is green, including functional tests

@prajwolrg can you re-review the current head? The prior `AsmWorkerCtx` comments are against old commit `fe7d7a91f` and that code is gone.

crates/btcio/src/reader/query.rs
```diff
@@ -104,6 +111,62 @@ fn calculate_target_next_block(
...
+    }
+
+    warn!(
+        %tip_height,
+        %tip_blockid,
+        has_manifest,
+        has_anchor_state,
+        "canonical L1 tip is not fully materialized; re-submitting tip to ASM"
+    );
+    block_submitter
```
bewakes: Only the tip is submitted here. I wonder if this is sufficient. It feels like the submission should be from `last_asm_synced_tip` to `tip`.

crates/btcio/src/reader/query.rs
```diff
@@ -104,6 +111,62 @@ fn calculate_target_next_block(
...
+/// [`handle_bitcoin_event`](super::handler::handle_bitcoin_event)). A crash
+/// between the canonical write and ASM materialization can therefore strand the
+/// canonical tip (or a short suffix) without a stored ASM anchor state. Restart
+/// computes the reader target from the canonical tip, so without this step that
+/// tip would be skipped until another Bitcoin block arrives.
+///
+/// The ASM worker can sync from a target block back to its latest stored anchor
+/// and replay any missing suffix itself, so the reader only needs to re-submit
+/// the stored canonical tip. It does not mutate the canonical L1 table during
+/// startup recovery.
```
bewakes: nit: I recommend limiting the function docstring to what the function exactly does and not the surrounding behavior which can be documented in call site.

bewakes (commented):
This is a good attempt. However, I feel like `btcio` should not depend on asm because it is downstream. Please consider using backfilling in ASM worker instead of the other way around.

voidash:
Updated this to address the layering concern from @bewakes.

What changed:
- Removed the restart recovery from `btcio`; it no longer checks ASM state or depends on ASM storage.
- Moved the recovery to ASM/consensus startup in `spawn_asm_worker`: after the ASM worker is launched, it compares the stored canonical L1 tip with the latest stored ASM anchor state. If ASM is behind or on a different tip, it submits the stored canonical tip to ASM and lets the ASM worker backfill from its latest anchor state through the canonical chain.
- Dropped the added async ASM state getter; it is no longer needed.
- Rebased onto current `main` (`175854f03`) so the PR is not conflicting.

Validation on the rebased head `20b0fc73b`:
- `cargo fmt --all -- --check`
- `cargo check -p strata-consensus-logic -p strata-btcio -p strata-storage --all-features`

I did not rerun the functional test locally after this last rebase because this machine is down to ~7GB free, which is too tight for a reliable functional-test build/run. The existing targeted test is still in the PR and should exercise this path in CI.

crates/consensus-logic/src/sync_manager.rs
```diff
@@ -152,13 +155,64 @@ pub fn spawn_asm_worker(
...
+    warn!(
+        %tip,
+        latest_asm_block = latest_asm_block
+            .as_ref()
+            .map(ToString::to_string)
+            .as_deref()
+            .unwrap_or("none"),
+        "canonical L1 tip is ahead of ASM state; re-submitting tip to ASM for startup backfill"
+    );
+    asm_handle
```
bewakes: Should this not be doing `submit_block()` from `latest_asm_block + 1` to `tip`? It just submits for `tip` now.

bewakes (changes requested):
One more concern. The rest looks good.

storopoli (approved):
ACK d422834ae3864cd380503e4792f53ff620f08500

bewakes (approved):

prajwolrg (approved):

crates/btcio/src/reader/query.rs
```diff
@@ -142,10 +151,31 @@ async fn do_reader_task<R: Reader>(
...
-                    for ev in events {
+                Ok(result) => {
+                    for ev in result.events {
+                        let submitted_block = matches!(
+                            &ev,
+                            L1Event::BlockData(block_data, _)
+                                if block_data.block_num() >= ctx.btcio_params.genesis_l1_height()
+                        );
                         handle_bitcoin_event(ev, &ctx, event_submitter).await?;
+                        startup_tip_delivered |= submitted_block;
```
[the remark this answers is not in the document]
voidash: this doesn't reproduce with this ASM worker . check please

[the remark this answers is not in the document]
voidash: @codex review

PR #1952, opened 2026-06-08
ci: stop patching the deployments repo from the CD workflow

alexhui01:
## Summary
Removes the cross-repo write coupling where the alpen CD workflow patched the deployments repo.

`cd.yml`'s **`update-helm-values`** job cloned `alpenlabs/deployments`, rewrote image tags + the SP1 VK with `yq`/`sed`, and **committed/pushed back** to a deployments branch (`BRANCH_OF_DEPLOYMENT_REPO`) using the `DEPLOYMENTS_REPO_WRITE` deploy key. Image-tag promotion is handled out of band now, so this job is removed.

## Changes
- Delete the `update-helm-values` job (clone deployments → yq/sed tag rewrite → commit/push).
- Drop the now-dead wiring that existed only to feed it:
  - the `build-and-push` job `outputs` block (`program`, `sp1_value`),
  - the `Extract SP1 Value` step,
  - the host-side `strata-datatool` build + `gen-checkpoint-predicate` extraction inside the build step (its output was only consumed by the Helm patch; the prover image build uses `--build-arg PROVER_FEATURES=sp1`, not the SP1 value).
- Update the header comment.

The workflow now only **builds and pushes the Docker image to ECR**. Net: +2 / −104 lines, valid YAML, single remaining job `build-and-push`.

## Follow-ups (not in this PR)
- The `DEPLOYMENTS_REPO_WRITE` secret and the `BRANCH_OF_DEPLOYMENT_REPO` / `CLUSTER_NAME` / `GH_ACTIONS_USER_NAME` repo vars are now unused and can be revoked/cleaned up in repo settings.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,6 +1,5 @@
-# This workflow is triggered when a new commit is made on main branch or manually triggered.
-# It builds and pushes the Docker images. Afterwards updates the Helm values with the new Docker
-# image tags and commits and pushes the changes to the deployments repo.
+# This workflow is manually triggered (workflow_dispatch).
+# It builds and pushes the Docker images to ECR.
 
 name: Continuous Deployment
 on:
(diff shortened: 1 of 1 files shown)


storopoli (approved):
ACK 52c9e1b5dbd71556b461dcdad3de5a810c2e0fdf

PR #1954, opened 2026-06-08
fix(btcio): resolve STR-3691 cleanup TODOs

storopoli:
## Description

Resolves the btcio cleanup items tracked by STR-3691.

This PR:
- Makes L1 intent bundling atomic by adding a storage-level `bundle_intent_payload` operation.
- Starts the L1 reader from the latest persisted client-state L1 anchor instead of the raw L1 DB tip.
- Replaces the reader's generic no-pivot failure with structured error context.
- Expands envelope builder test coverage around pubkey requirements, UTXO filtering, insufficient funds, reveal dust handling, and witness population.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The storage change intentionally keeps the existing intent index entry and only updates the intent row status during bundling, so bundling does not create duplicate intent index entries.

- [ ]  Yes
- [x]  No

## Checklist

AI assistance was used to implement the STR-3691 cleanup and split the changes into atomic commits.

## Related Issues

STR-3691

### DIFF
--- a/crates/btcio/src/reader/query.rs
+++ b/crates/btcio/src/reader/query.rs
@@ -13,6 +13,7 @@ use strata_primitives::l1::{L1BlockCommitment, L1Height};
 use strata_state::BlockSubmitter;
 use strata_status::StatusChannel;
 use strata_storage::NodeStorage;
+use thiserror::Error;
 use tokio::time::sleep;
 use tracing::*;
 
@@ -63,6 +64,20 @@ impl ReaderValidation {
             expected_l1_anchor,
         }
     }
+}
+
+#[derive(Debug, Error, PartialEq, Eq)]
+enum ReaderError {
+    #[error(
+        "btcio: unable to find common L1 block with Bitcoin client chain \
+         (client height {client_height}, reader best height {reader_best_height}, \
+         known depth {known_depth})"
+    )]
+    PivotNotFound {
+        client_height: L1Height,
+        reader_best_height: L1Height,
+        known_depth: usize,
+    },
 }
 
 /// The main task that initializes the reader state and starts reading from bitcoin.
@@ -269,9 +284,18 @@ async fn poll_for_new_blocks<R: Reader>(
             return Ok(vec![revert_ev]);
         }
     } else {
-        // TODO(STR-3691): make this case a bit more structured
-        error!("unable to find common block with client chain, something is seriously wrong here!");
-        bail!("things are broken with l1 reader");
+        let reader_best_height = state.best_block_idx();
+        let known_depth = state.iter_blocks_back().count();
+        let err = ReaderError::PivotNotFound {
+            client_height,
+            reader_best_height,
+            known_depth,
+        };
+        error!(
+            client_height,
+            reader_best_height, known_depth, "unable to find common block with client chain"
+        );
+        return Err(err.into());
     }
 
     debug!(%client_height, "have new blocks");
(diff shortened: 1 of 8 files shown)


delbonis (approved):

krsnapaudel (approved):
LGTM

krsnapaudel:
@storopoli the fn test run likely got canceled due to one hour limit.

storopoli:
> @storopoli the fn test run likely got canceled due to one hour limit.

Yeah, the workflow was ultimately cancelled by the 1h job timeout, but the functional tests had already started failing before that.

I reproduced `test_checkpoint_finalized` locally and found the issue: the reader was starting from the latest client-state height + 1 (`111`) even though the corresponding canonical L1 anchor at height `110` had not been persisted yet. That caused CSM startup to fail pivot lookup and shut down with `ReorgPastFinality`.

I've mad a fix that only trusts the client-state start height if the matching L1 canonical block is already stored; otherwise it falls back so the reader persists the missing anchor first.

The functional tests now passes and starts the reader at `target_next_block=110`.

bewakes:
@storopoli great work on the atomic bundling! One qq:

> Starts the L1 reader from the latest persisted client-state L1 anchor instead of the raw L1 DB tip.

What was the issue with the starting from raw L1 DB tip? How I look at the reader is that it is completely independent of existence of other components especially downstream components. The data flow is like this `reader -> ASM -> CSM` and I find it a bit weird that reader depends on the CSM's state for initialization. This might be fine since it is only during the startup, but I'd still like to know the reason for this.

And apologies for late review.

storopoli:
> @storopoli great work on the atomic bundling! One qq:
> 
> > Starts the L1 reader from the latest persisted client-state L1 anchor instead of the raw L1 DB tip.
> 
> What was the issue with the starting from raw L1 DB tip? How I look at the reader is that it is completely independent of existence of other components especially downstream components. The data flow is like this `reader -> ASM -> CSM` and I find it a bit weird that reader depends on the CSM's state for initialization. This might be fine since it is only during the startup, but I'd still like to know the reason for this.
> 
> And apologies for late review.

I agree the dependency looks a bit odd if we think of the reader as purely upstream of `ASM -> CSM`.

The reason is restart recovery. The raw L1 DB tip only tells us the reader persisted L1 chain-tracking data. It does not prove the downstream pipeline durably processed that block into client state. If the process crashes after the reader stores L1 data but before ASM/CSM catches up, restarting from the raw L1 DB tip can skip replaying blocks that downstream still needs.

So the client-state L1 anchor is being used as a durable downstream watermark during startup: start far enough back that anything not reflected in client state gets replayed through `reader -> ASM -> CSM`.

That said, your concern is valid. I found one edge case while debugging the functional failure: on a fresh functional-test DB, client state could point at the genesis anchor while the corresponding L1 canonical row was not stored yet. Blindly starting from client-state + 1 skipped that anchor and made CSM fail pivot lookup. I tightened the logic so the reader only trusts the client-state anchor if the matching canonical L1 block exists in storage; otherwise it falls back and persists the missing anchor first.

So the dependency is startup-only and is being used as a replay watermark, not as an ongoing reader dependency. Longer term, an explicit “downstream processed L1 height” watermark might make this layering clearer than reading it from client state. But it is out of scope of this PR. I can open a ticket if you want.

bewakes:
Thanks for the clarification @storopoli, I can now see why this was done in this particular way. I am still inclined towards making it independent of CSM's state even if it is only during startup. The reason I say this is that it's the responsibility of downstream components to appropriately backfill missed blocks. Also because let's say there's some other component(X) which depends on btcio reader and CSM caught upto the block but the component X did not. Relying on one downstream consumer's state for initialization just doesn't feel quite right and always correct to me.

storopoli:
Yeah, I agree with that direction.

For this PR I’m going to keep the narrow fix: prevent the reader from skipping the missing L1 anchor on startup, which is what was causing the functional failures. That keeps STR-3691 unblocked without expanding the scope into the broader recovery-boundary cleanup.

I’ll track the cleaner design separately: reader startup should depend only on reader/L1 state; ASM should backfill from canonical L1 storage to its own processed watermark; CSM should backfill from ASM state to its client-state watermark. That also handles your component X example better, since every downstream consumer owns its own catch-up from its own durable watermark instead of the reader depending on one consumer’s state.

Created [STR-3845](https://alpenlabs.atlassian.net/browse/STR-3845)

[STR-3845]: https://alpenlabs.atlassian.net/browse/STR-3845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

bewakes (approved):

PR #1956, opened 2026-06-08
STR 3370, STR 3580 FCM replay unvalidated blocks and improve logging 

krsnapaudel:
## Description

**STR-3370 — replay unvalidated OL blocks on startup.** A block persisted but not yet validated stays `Unchecked`; if the node restarted before validating it, the unfinalized-tree loader skipped it and nothing re-queued it, orphaning the block. The loader now collects `Unchecked` blocks (with data), in slot order, as replay candidates, attaches `Valid`, skips `Invalid`, and logs `None`/missing-data as a DB inconsistency. The candidates are stored on the FCM state, and the FCM service's `on_launch` hook reprocesses them through the normal validation path before the input loop starts — so they replay ahead of any live block/CSM message, without prefilling the command queue.

**STR-3580 — accurate finalization logging.** A CSM update with no new finalized epoch used to log `warn!` "invalid or out of order epoch" + "did not finalize epoch", falsely implying a problem. `handle_new_state_update` now returns early at `debug!` when the observed epoch already matches the applied finalized epoch, retries finalization when it only matches an epoch still pending locally, and routes ordering decisions through `record_observed_finalized_epoch`, which distinguishes an already-recorded epoch (`debug!`) from genuine out-of-order / inconsistent ordering (`warn!`).

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-3370](https://alpenlabs.atlassian.net/browse/STR-3370)
[STR-3580](https://alpenlabs.atlassian.net/browse/STR-3580)

[STR-3370]: https://alpenlabs.atlassian.net/browse/STR-3370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/consensus-logic/src/fcm/service.rs
+++ b/crates/consensus-logic/src/fcm/service.rs
@@ -164,7 +164,12 @@ async fn process_fc_message<C: FcmContext>(
                     // problem checking the block in general and it could be
                     // valid or invalid, but we're kinda sloppy with errors
                     // here so let's try to avoid crashing the FCM task?
-                    error!(%slot, %blkid, "error processing block, interpreting as invalid\n{e:?}");
+                    error!(
+                        %slot,
+                        %blkid,
+                        err = ?e,
+                        "error processing block, interpreting as invalid"
+                    );
                     false
                 }
             };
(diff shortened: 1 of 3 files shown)


crates/consensus-logic/src/fcm/service.rs
```diff
@@ -69,12 +70,33 @@ pub async fn start_fcm_service<C: FcmContext>(
...
+        fcm_tx
+            .send(ForkChoiceMessage::NewBlock(blkid))
+            .await
+            .context("failed to submit startup replay candidate to FCM")?;
+    }
+
+    info!(
+        replay_candidate_count,
+        "submitted startup replay candidates"
+    );
```
delbonis: Should probably be debug.

Also return early without printing this if is_empty so that we don't have spurious logs.

delbonis (changes requested):
I'm unsure if the approach taken is the best.  I would have assumed that the blocks that we're replaying at startup could be loaded into the state on init directly to be processed or we'd detect that there blocks that need to be replayed when we get the first input, instead of doing thing where we pre-stuff the command queue with blocks to process.

The smell that I'm seeing here is that this makes the fn that loads the initial FCM state also be responsible for an additional thing, which is to also load the list of blkids to reprocess.  Maybe this works fine, but it seems like it's making the init process more involved for something that *could* be directly integrated.

Want to hear your rationale.

krsnapaudel:
> I'm unsure if the approach taken is the best. I would have assumed that the blocks that we're replaying at startup could be loaded into the state on init directly to be processed or we'd detect that there blocks that need to be replayed when we get the first input, instead of doing thing where we pre-stuff the command queue with blocks to process.
> 
> The smell that I'm seeing here is that this makes the fn that loads the initial FCM state also be responsible for an additional thing, which is to also load the list of blkids to reprocess. Maybe this works fine, but it seems like it's making the init process more involved for something that _could_ be directly integrated.
> 
> Want to hear your rationale.

@delbonis I was also not happy with the approach to send new block commands. I did not push back on the Codex plan due to my limited understanding of async service. Apparently we can do this in `on_launch`.

delbonis:
@krsnapaudel Yeah the `on_launch` call is probably more appropriate for that, especially if they're not needed elsewhere in the setup.

krsnapaudel:
> @krsnapaudel Yeah the `on_launch` call is probably more appropriate for that, especially if they're not needed elsewhere in the setup.

@delbonis Updated.

storopoli (approved):
ACK 5e0cb964ca7c77d0fa4e77eb09434cf5d54a32df

bewakes (approved):

delbonis (approved):
Ok looks good now.

PR #1957, opened 2026-06-09
refactor!: drop temporary hardcoded test schnorr keys from primitives

prajwolrg:
## Description

`crates/primitives/src/utils.rs` exposed `get_test_schnorr_keys()` and `SchnorrKeypair` — a temporary stopgap (flagged `FIXME(STR-3689): remove`) that carried hardcoded Schnorr keypairs purely so tests could grab a fixed keypair. This put test-only material with hardcoded keys on `strata-primitives`' public surface.

This PR removes that module entirely and replaces the usages with a `test_schnorr_keypair()` helper in `strata-ol-chain-types-new`'s test utils. The helper derives a fresh random keypair on each call (mirroring the byte-clamping approach of `EvenPublicKey`'s `Arbitrary` impl in `strata-crypto`), so the returned `(secret_key, x_only_pubkey)` is self-consistent for sign/verify but no longer hardcoded. Callers in `consensus-logic` and `ol-block-assembly` tests, plus the local helper in `ol-chain-types`' `validation` tests, now use the shared helper, and `strata-primitives` is dropped as a dependency of `ol-chain-types`.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Breaking only in the sense that `strata_primitives::utils` (and its public `get_test_schnorr_keys`/`SchnorrKeypair`) is removed from the public API; it had no non-test users. Each call to `test_schnorr_keypair()` now returns a different keypair — every caller uses the pair self-consistently within a single call (signs with the secret key, builds the predicate from the public key), so none relied on a fixed or cross-call-stable value.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

STR-3689 (Jira).

---

_AI assistance notice: this change was prepared with the help of Claude Code._


bewakes (approved):

delbonis (approved):

storopoli (approved):
ACK cae04b2b77a4ccaa8628d750f31a54d19c56bab1

PR #1962, opened 2026-06-09
feat(evm): price bridgeout precompile gas

purusang:
## Description

Price bridge-out precompile execution in raw EVM gas instead of returning zero gas. The bridge-out precompile now computes a deterministic gas cost before validation or state/log side effects, rejects calls with insufficient gas, and returns the computed cost through `PrecompileOutput`.

The current implementation uses a protocol-constant proposal:

```text
bridgeout_gas = 10_000 + 16 * calldata_len
```

This is intentionally kept as code constants, not runtime config, because precompile gas accounting is consensus-visible. The Notion fee model says Alpen-only precompiles should be repriced/reviewed before launch, while DA/proving/OL overhead stay outside raw EVM gas as `non_execution_fee`.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Open question: are `BRIDGEOUT_BASE_GAS = 10_000` and `BRIDGEOUT_CALLDATA_BYTE_GAS = 16` acceptable launch constants, or should protocol/product choose different values?

The shape is `base + calldata-byte cost` so the fixed bridge-out operation is not free and longer BOSD/calldata payloads do not cost the same as short payloads. Non-execution costs from the Alpen L2 Fee Model are deliberately not included in this precompile gas value.

Validation:

- `cargo test -p alpen-reth-evm precompiles::bridge::`
- `cargo fmt --check --package alpen-reth-evm`

- [x]  Yes
- [ ]  No

- https://alpenlabs.atlassian.net/browse/STR-3676
- https://app.notion.com/p/Alpen-L2-Fee-Model-333901ba000f80d899b7c628cf0469d9

## Checklist

AI assistance notice: I used AI assistance to inspect the existing precompile and fee-model code, map the Notion fee model to raw precompile gas, and draft this implementation.

## Related Issues

https://alpenlabs.atlassian.net/browse/STR-3676

### DIFF
--- a/crates/reth/evm/src/precompiles/bridge.rs
+++ b/crates/reth/evm/src/precompiles/bridge.rs
@@ -6,6 +6,13 @@ use strata_primitives::bitcoin_bosd::Descriptor;
 
 use crate::{constants::BRIDGEOUT_PRECOMPILE_ADDRESS, utils::wei_to_sats};
 
+// REVIEW(STR-3676): Replace these draft values with protocol-approved launch constants.
+/// Fixed raw EVM gas charged for bridge-out precompile execution.
+const BRIDGEOUT_BASE_GAS: u64 = 10_000;
+
+/// Raw EVM gas charged per calldata byte handled by the bridge-out precompile.
+const BRIDGEOUT_CALLDATA_BYTE_GAS: u64 = 16;
+
 /// Custom precompile to burn rollup native token and add bridge out intent of equal amount.
 /// Bridge out intent is created during block payload generation.
 /// This precompile validates transaction and burns the bridge out amount.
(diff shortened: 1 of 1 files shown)


storopoli:
> Open question: are `BRIDGEOUT_BASE_GAS = 10_000` and `BRIDGEOUT_CALLDATA_BYTE_GAS = 16` acceptable launch constants, or should protocol/product choose different values?

The PRDs are ambiguous and just say:

> Additional precompiles enabling compatibility with the Strata bridge and enabling easier compatibility with bitcoin MAY be added.

and

> If the Bridge User's address does not have enough A-BTC to cover the amount specified in their withdrawal request, plus the gas fee for the transaction, then the transaction MUST fail.

Finally,

> By default, the Alpen Sequencer MUST NOT lose money. The Alpen Sequencer MAY optionally modify certain configuration settings or introduce extrinsic incentives to subsidize Alpen usage.

Hence, we have full discretion to choose how we price the precompile to make sure that the sequencer does not lose funds during operation.

I think it's fine to keep these as constants for now. I see in the future we moving these to a "params" style of maintaining values, but let's start simple: as constants.

storopoli (approved):
ACK b211e3cae39fac1ba45623c4859b0e8fc5a457cf

barakshani:
Nice, please also see https://alpenlabs.atlassian.net/browse/STR-3572.

The user should not only pay for the EVM storage slot changes (in the precompile, and its own state (nonce, BTC balance)), but also for the bitcoin footprint that the subsequent checkpoint will leave, due to withdrawal log. Luckily, this is independent of the amount (any call to the precompile will result in a single withdrawal log).

MdTeach (approved):

PR #1964, opened 2026-06-09
Paas proper retries

evgenyzdanovich:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/bin/strata/src/prover/errors.rs
+++ b/bin/strata/src/prover/errors.rs
@@ -47,6 +47,8 @@ impl From<ProverError> for PaasError {
             ProverError::StaleTaskCommitment { .. }
             | ProverError::EpochCommitmentNotFound(_)
             | ProverError::EpochSummaryNotFound(_) => PaasError::transient(e.to_string()),
+            // Infra: surface as a retryable error, not a domain verdict.
+            ProverError::Database(_) => PaasError::Storage(e.to_string()),
             _ => PaasError::permanent(e.to_string()),
         }
     }
--- a/bin/strata/src/prover/errors.rs
+++ b/bin/strata/src/prover/errors.rs
@@ -46,8 +46,8 @@ impl From<ProverError> for PaasError {
         match e {
             ProverError::StaleTaskCommitment { .. }
             | ProverError::EpochCommitmentNotFound(_)
-            | ProverError::EpochSummaryNotFound(_) => PaasError::TransientFailure(e.to_string()),
-            _ => PaasError::PermanentFailure(e.to_string()),
+            | ProverError::EpochSummaryNotFound(_) => PaasError::transient(e.to_string()),
+            _ => PaasError::permanent(e.to_string()),
         }
     }
 }
--- a/bin/strata/src/prover/errors.rs
+++ b/bin/strata/src/prover/errors.rs
@@ -38,18 +38,74 @@ pub(crate) enum ProverError {
 /// Classifies input-fetch failures as retriable or permanent for the paas
 /// service.
 ///
-/// Stale commitments and missing epoch metadata reflect expected race
-/// conditions — the orchestration layer resubmits the canonical epoch on
-/// its next tick, so those retry. Anything else is treated as permanent.
+/// Missing epoch metadata is a genuine not-ready-yet wait — the epoch's
+/// commitment/summary hasn't been produced yet — so those become transient
+/// (the paas layer parks them as `Blocked` and rechecks).
+///
+/// A *stale* commitment is different: the task was submitted for an epoch
+/// commitment that is no longer canonical (a same-epoch reorg replaced it). It
+/// can never become canonical again, so retrying/reblocking it forever would
+/// hang the checkpoint runner, which awaits `execute` inline. Mark it permanent
+/// (→ `Rejected`) so `execute` returns `Failed` and the runner's recovery path
+/// re-reads the now-canonical commitment and resubmits the replacement task.
 impl From<ProverError> for PaasError {
     fn from(e: ProverError) -> Self {
         match e {
-            ProverError::StaleTaskCommitment { .. }
-            | ProverError::EpochCommitmentNotFound(_)
-            | ProverError::EpochSummaryNotFound(_) => PaasError::transient(e.to_string()),
+            ProverError::EpochCommitmentNotFound(_) | ProverError::EpochSummaryNotFound(_) => {
+                PaasError::transient(e.to_string())
+            }
+            ProverError::StaleTaskCommitment { .. } => PaasError::permanent(e.to_string()),
             // Infra: surface as a retryable error, not a domain verdict.
             ProverError::Database(_) => PaasError::Storage(e.to_string()),
             _ => PaasError::permanent(e.to_string()),
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use strata_paas::FailureAction;
+
+    use super::*;
+
+    fn action_of(e: ProverError) -> FailureAction {
+        PaasError::from(e).action()
+    }
+
+    #[test]
+    fn missing_epoch_metadata_is_transient() {
+        // Not produced yet — the OL will fill it in; park as Blocked and recheck.
+        assert_eq!(
+            action_of(ProverError::EpochCommitmentNotFound(7)),
+            FailureAction::RetryResume
+        );
+        assert_eq!(
+            action_of(ProverError::EpochSummaryNotFound(7)),
+            FailureAction::RetryResume
+        );
+    }
+
+    #[test]
+    fn stale_commitment_is_permanent() {
+        // A same-epoch reorg replaced the commitment; it can never become
+        // canonical again. Must be permanent (-> Rejected) so the checkpoint
+        // runner's inline `execute` returns instead of blocking forever, letting
+        // it re-read the canonical commitment and resubmit.
+        assert_eq!(
+            action_of(ProverError::StaleTaskCommitment {
+                epoch: 3,
+                task: EpochCommitment::null(),
+                canonical: EpochCommitment::null(),
+            }),
+            FailureAction::Permanent
+        );
+    }
+
+    #[test]
+    fn database_errors_surface_as_storage() {
+        assert_eq!(
+            action_of(ProverError::DaComputation("boom".into())),
+            FailureAction::Permanent
+        );
     }
 }
--- a/crates/prover-core/src/config.rs
+++ b/crates/prover-core/src/config.rs
@@ -22,6 +22,11 @@ pub struct RetryConfig {
     /// fresh). Kept much smaller than `max_retries` because each resubmit
     /// re-runs the whole proof, whereas resume-class retries only re-poll.
     pub max_resubmits: u32,
+    /// Default recheck cadence for a `Blocked` task (waiting on a dependency),
+    /// in seconds. A steady poll — not exponential backoff — since blocking is
+    /// an expected wait, not a failure. A spec can override per task via
+    /// [`InputResolution::Blocked`](crate::InputResolution)'s `recheck_after`.
+    pub blocked_recheck_secs: u64,
 }
 
 impl Default for RetryConfig {
@@ -33,6 +38,7 @@ impl Default for RetryConfig {
             max_delay_secs: 3600,
             jitter_frac: 0.2,
             max_resubmits: 3,
+            blocked_recheck_secs: 10,
         }
     }
 }
--- a/crates/prover-core/src/config.rs
+++ b/crates/prover-core/src/config.rs
@@ -1,10 +1,46 @@
 //! Configuration.
 
+use std::time::Duration;
+
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Clone)]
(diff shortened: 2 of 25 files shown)


crates/prover-core/src/config.rs
```diff
@@ -22,17 +27,78 @@ impl Default for RetryConfig {
...
+    pub fn jittered_delay_secs(&self, retry_count: u32, seed: u64) -> u64 {
+        let base = self.calculate_delay(retry_count) as f64;
+        if self.jitter_frac <= 0.0 {
+            return base as u64;
+        }
+        // Map the seed deterministically into [0, 1).
+        let frac = (seed % 10_000) as f64 / 10_000.0;
+        let factor = 1.0 - self.jitter_frac + 2.0 * self.jitter_frac * frac;
+        (base * factor).clamp(0.0, self.max_delay_secs as f64) as u64
+    }
```
prajwolrg, line 180: This is really clever!

crates/prover-core/src/prover.rs
```diff
@@ -492,6 +493,17 @@ impl<H: ProofSpec> Prover<H> {
...
 
+/// Deterministic per-task backoff seed (FNV-1a over the key, mixed with the
+/// attempt count). Used to jitter retry delays so distinct tasks that failed on
+/// the same tick spread their wake-ups instead of retrying in lockstep.
+fn jitter_seed(key: &[u8], retry_count: u32) -> u64 {
+    let mut h: u64 = 0xcbf2_9ce4_8422_2325;
+    for &b in key {
+        h = (h ^ u64::from(b)).wrapping_mul(0x0000_0100_0000_01b3);
+    }
+    h.wrapping_add(u64::from(retry_count))
```
prajwolrg, line 640: nit: might be worth making the values constants, and(/or) adding references to where the values come from.

prajwolrg (approved):
LGTM! Great work!

delbonis (approved):
Very good improvements. <3

krsnapaudel (commented):
Found by Claude. This seems possible.

**Double-spawn of `run_task` on crash recovery.** `retry_after` is never cleared, so it survives the `→ Proving` transition (`update_status` → `set_status` touches only `status` + `updated_at`). Sequence:

1. Task fails transiently → `TransientFailure`, `schedule_retry` sets `retry_after = T`.
2. Scanner retries it → `run_task` flips to `Proving{..}`; `retry_after` still `T` (not cleared).
3. Process crashes mid-prove. On disk: `Proving{..}`, `retry_after = T`.
4. Restart, first `tick()`: `recover()` rewrites it to `TransientFailure` (retry_after still `T ≤ now`) and spawns `run_task` **A**.
5. Same tick: `list_retriable(now)` matches it (`wants_rescan()` + `retry_after ≤ now`) and spawns `run_task` **B**.

crates/prover-core/src/task.rs
```diff
@@ -15,16 +15,32 @@ use serde::{Deserialize, Serialize};
...
+    Proving {
+        retry_count: u32,
+        resubmit_count: u32,
+    },
     /// Proof completed successfully, receipt available.
     Completed,
+    /// Parked waiting for an input dependency (e.g. an upstream chunk proof
+    /// not yet produced). Not a failure: rechecked on a steady cadence via
+    /// `retry_after`, and does NOT consume the retry/resubmit budget.
+    Blocked { reason: String },
```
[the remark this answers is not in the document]
storopoli: Added a "DB schema breakage" label that I've created to DB schema breakage PRs

crates/prover-core/src/prover.rs
```diff
@@ -296,47 +312,97 @@ impl<H: ProofSpec> Prover<H> {
...
         async {
+            // Stage checkpoint: if proving already succeeded on a prior attempt
+            // the receipt is in the receipt store. Skip resolve_input and the
+            // (expensive) prove, and re-run only the post-prove hook + finish —
+            // so a transient receipt-hook failure never triggers a re-prove.
+            // Only applies to provers with a receipt store configured.
+            if let Some(receipt) = self
+                .receipt_store
+                .as_ref()
+                .and_then(|store| store.get(&key).ok().flatten())
```
storopoli: **Receipt-store read errors are swallowed → duplicate re-prove.** `.ok().flatten()` treats a transient `ReceiptStore::get` failure as "no receipt", so after a successful `put` followed by a transient hook failure, the retry can fall through this checkpoint and re-run the full (possibly remote, expensive) prove — contradicting the README's "a transient receipt-hook failure never triggers a re-prove" claim.

Consider propagating the error into `handle_error` (it would classify as `Storage` → `RetryResume`) instead of dropping it.

crates/prover-core/src/config.rs
```diff
@@ -22,17 +75,98 @@ impl Default for RetryConfig {
...
+    /// clamped to `max_delay_secs`.
+    pub fn jittered_delay_secs(&self, retry_count: u32, seed: u64) -> u64 {
+        let base = self.calculate_delay(retry_count) as f64;
+        if self.jitter_frac <= 0.0 {
+            return base as u64;
+        }
+        // Map the seed deterministically into [0, 1).
+        let frac = (seed % 10_000) as f64 / 10_000.0;
+        let factor = 1.0 - self.jitter_frac + 2.0 * self.jitter_frac * frac;
+        (base * factor).clamp(0.0, self.max_delay_secs as f64) as u64
```
storopoli, line 179: **No validation on `RetryConfig` values.** All fields are `pub` and deserialized from config, so e.g. `jitter_frac > 1.0` makes `factor` negative for many seeds; the delay clamps to `0` and the task retries on every tick instead of backing off. A very large `max_delay_secs` can also overflow `now_secs() + delay_secs` at 

https://github.com/alpenlabs/alpen/blob/bcc76f2c0b9c67b521331edda373b305c4cd50e7/crates/prover-core/src/prover.rs#L565

A validating constructor (or a `validate()` called on load) clamping `jitter_frac` to `[0, 1]` and sanity-bounding the delays would close both.

crates/prover-core/src/prover.rs
```diff
@@ -408,40 +474,95 @@ impl<H: ProofSpec> Prover<H> {
...
-                },
-            );
-        }
+    /// Park a task that is waiting on an input dependency.
+    ///
+    /// Sets [`TaskStatus::Blocked`] and schedules a steady recheck via
+    /// `retry_after`. Does not touch the retry/resubmit counters or notify
+    /// waiters — blocking is an expected wait, not a failure, so the scanner
+    /// re-spawns it (via [`TaskStatus::wants_rescan`]) when the recheck is due.
+    fn park_blocked(&self, key: &[u8], reason: String, recheck_after: Option<Duration>) {
```
storopoli: **Design question: `Blocked` tasks never expire.** A dependency that never materializes (e.g. a chunk receipt for an abandoned batch) now rechecks at `blocked_recheck_secs` forever, and since parking doesn't notify watchers, any `wait_for_tasks`/`execute` caller hangs indefinitely. Pre-PR the retry budget eventually forced `PermanentFailure` and unblocked waiters.

If indefinite waiting is intended, worth a sentence in the README; otherwise a max-blocked-duration (or max recheck count) backstop that promotes to `PermanentFailure` would be safer.

bin/strata/src/prover/errors.rs
```diff
@@ -46,8 +46,10 @@ impl From<ProverError> for PaasError {
         match e {
             ProverError::StaleTaskCommitment { .. }
             | ProverError::EpochCommitmentNotFound(_)
-            | ProverError::EpochSummaryNotFound(_) => PaasError::TransientFailure(e.to_string()),
-            _ => PaasError::PermanentFailure(e.to_string()),
+            | ProverError::EpochSummaryNotFound(_) => PaasError::transient(e.to_string()),
```
storopoli: **Checkpoint runner can hang forever on a same-epoch reorg.** `StaleTaskCommitment` is classified as transient here, and `InputResolution::from_result`

https://github.com/alpenlabs/alpen/blob/bcc76f2c0b9c67b521331edda373b305c4cd50e7/crates/prover-core/src/traits.rs#L146

now bridges every non-permanent failure to `Blocked`. But a stale commitment can never become canonical again, so each recheck re-blocks — and `park_blocked`

https://github.com/alpenlabs/alpen/blob/bcc76f2c0b9c67b521331edda373b305c4cd50e7/crates/prover-core/src/prover.rs#L483

never notifies watchers. The checkpoint runner awaits `execute(task)` inline:

https://github.com/alpenlabs/alpen/blob/bcc76f2c0b9c67b521331edda373b305c4cd50e7/bin/strata/src/prover/mod.rs#L289

so if the canonical commitment for the epoch changes between task submission and `resolve_input`, `execute` never resolves: the runner never re-reads the canonical commitment and never submits the replacement. Pre-PR, retries exhausted the budget → `PermanentFailure` → `execute` returned `Failed` and the runner recovered.

Suggest mapping `StaleTaskCommitment` to `PaasError::permanent` (→ `Rejected`) so `execute` returns `Failed` and the runner's existing recovery path (break, re-read canonical, resubmit) kicks in. `EpochCommitmentNotFound`/`EpochSummaryNotFound` staying transient (→ `Blocked`) is correct — those are genuine not-ready-yet waits.

storopoli (changes requested):
Some AI findings that I've validated myself

bewakes (approved):

purusang (approved):

irnb (approved):

crates/prover-core/src/prover.rs
```diff
@@ -479,6 +600,17 @@ impl<H: ProofSpec> Prover<H> {
     }
```
storopoli: **A transient status read in `notify` permanently loses the terminal notification.** After a task reaches `Completed`/`PermanentFailure`, `notify` re-reads the record and converts any store error into `None`.

On a read error the `let Some(result) … else { return }` bails out *without draining the watcher entry*. The task is already terminal, so `run_task` never runs again and `notify` is never re-invoked for that key — meaning an otherwise **successful** prove leaves every `execute`/`wait_for_tasks` waiter hung forever on one transient DB read, with the stale senders also leaking in the watchers map. Same `.ok().flatten()` pattern as the receipt-store issue I flagged at prover.rs:350, but here the failure mode is a permanent hang rather than a re-prove.

Since the caller of `notify` just computed the terminal status itself, consider passing the known `TaskStatus` (or the `TaskResult`) into `notify` instead of re-reading the store — that removes the fallible read entirely. Failing that, on a read error either retry or drain the watchers with an error result so waiters observe the failure instead of hanging.

crates/prover-core/src/prover.rs
storopoli, line 106: **PaaS `execute` deadlocks the tick loop on any recoverable first-attempt failure.** `Cmd::Execute` is awaited inline inside `handle_cmd`:

https://github.com/alpenlabs/alpen/blob/bcc76f2c0b9c67b521331edda373b305c4cd50e7/crates/paas/src/service.rs#L85-L88

and `process_input` handles both commands and ticks serially on the same service state, so no `Tick` can be processed while that await is pending. But `Prover::execute` blocks until the task is *terminal*:

https://github.com/alpenlabs/alpen/blob/bcc76f2c0b9c67b521331edda373b305c4cd50e7/crates/prover-core/src/prover.rs#L102-L107

and a first attempt that lands in `TransientFailure` or `Blocked` only ever gets rescheduled by `tick()`'s retriable scan:

https://github.com/alpenlabs/alpen/blob/bcc76f2c0b9c67b521331edda373b305c4cd50e7/crates/prover-core/src/prover.rs#L203-L223

So the retry that would unblock `execute` is queued behind `execute` itself: one transient failure (or one dependency wait) deadlocks the whole ticking service, and every subsequent `Submit`/`Execute`/`Tick` piles up behind it. Note this makes the checkpoint-runner hang I flagged earlier strictly worse — it doesn't even need a stale commitment, just a single transient error on the first attempt.

Suggest not awaiting terminal completion inside `process_input` — e.g. have `handle_cmd` do the `submit` synchronously but spawn the `wait_for_tasks` + completion-send onto the executor, so ticks keep flowing while callers wait.

storopoli (changes requested):

bin/strata/src/prover/errors.rs
evgenyzdanovich: done

crates/prover-core/src/prover.rs
evgenyzdanovich: thanks, I agree, done.

storopoli (approved):
ACK 1bcfc467a342fa0d351b87af07b1143ceb3c99b8

storopoli:
@evgenyzdanovich take a look at the recent codex findings.
Once you're done feel free to add it into the queue. Nice work!

evgenyzdanovich:
@storopoli Thanks! Filed a follow-up ticket to just merge and unblock for the faster feedback loop -- STR-4182. 

The findings addressed properly are definitely out of scope: **It is actually desirable to have compare-and-swap on a storage layer to properly fix all potential wild races that may occur.**

PR #1965, opened 2026-06-09
chore(deps): bump strata-common (rc25), asm (alpha.13), zkaleido (beta.4)

prajwolrg:
## Summary

Bumps the pinned dependencies and adapts the code to their API changes: **strata-common** `v0.1.0-alpha-rc23` → `v0.1.0-alpha-rc25`, **asm** `v0.1-alpha.12` → `v0.1-alpha.13`, and **zkaleido** `v0.1-beta.3` → `v0.1-beta.4` — across the workspace and the out-of-workspace SP1 guest manifests.

## Changes

**asm alpha.13 — `WorkerContext` split into concern traits.** The umbrella `WorkerContext` is now a blanket impl over `L1DataProvider`, `AnchorStateStore`, `ManifestMmrStore`, and `AuxDataStore`. `AsmWorkerCtx` is split into the four impls, adding the new `get_l1_block_header` and `manifest_mmr_leaf_count` methods. The manifest MMR write is now height-indexed (`put_manifest_hash(height, hash)`); since the local `MmrIndexHandle` is append-only, it is mapped positionally — append at the tip, or pop back to `height` then append on an L1 reorg (the worker rewrites leaves forward from the fork point, so dropped leaves belong to the abandoned branch). This pop-then-append overwrite is non-atomic but crash-safe: `apply_block` records the manifest before storing the anchor state (the commit point), so a crash mid-overwrite leaves the block uncommitted and the next sync re-runs it idempotently. A `NOTE` flags that a native height-indexed `put_leaf` would make it a single write.

**rc25 — `strata-predicate` default features.** rc25 drops `serde` from the predicate crate's defaults and renames the `all` feature. `PredicateKey` is (de)serialized across the workspace (ol-params, ol-rpc-types, alpen-chunk, …), which previously came for free via the default and now only compiled in full-workspace builds via feature unification. Re-enabled `serde` on the workspace `strata-predicate` dependency (matching how `strata-codec` / `strata-merkle` declare needed features there), and dropped the now-removed `all` feature from `strata-ol-stf`.

**rc25 — Sp1Groth16 predicate condition encoding.** rc25 switches the `Sp1Groth16` predicate condition from a borsh-serialized verifier (rc23) to the verifier's canonical uncompressed encoding (`SP1Groth16Verifier::parse` / `to_uncompressed_bytes`). The borsh form still compiles (the verifier is `BorshSerialize`) but produces a condition rc25 cannot parse, so every producer is switched to `to_uncompressed_bytes`: the datatool checkpoint/acct predicate builders, and `provers/sp1/build.rs`, which bakes the per-guest `*_VK_CONDITION` constants — including the chunk VK condition the alpen-acct guest verifies chunk proofs against. The now-unused `borsh` deps are dropped.

**zkaleido beta.4 — checkpoint proof bytes.** The groth16 verifier now accepts the SP1 verifying-key-hash prefix directly, so the checkpoint payload passes the stored proof bytes through unmodified instead of stripping the prefix, and drops the `zkaleido` / `zkaleido-sp1-groth16-verifier` deps from the checkpoint crate.

**SP1 guest manifests.** The `provers/sp1/guest-*` packages have their own workspaces and lockfiles, so they pin the git deps independently; bumped `strata-predicate` to rc25 and `zkaleido-sp1-guest-env` to beta.4 there to keep them in lockstep with the workspace.

## Testing

- `cargo check --workspace --all-targets --all-features` — clean
- `cargo clippy --workspace --lib --bins --examples --tests --benches --all-features -- -D warnings` — clean
- `cargo check --locked` — lockfile in sync

The SP1 guest packages build out-of-workspace and are not covered by the checks above. The `provers/sp1/build.rs` condition baking is `#[cfg(all(feature = "sp1-dev", not(debug_assertions)))]`, so it is not exercised by debug checks and needs the SP1 toolchain (release + `sp1-dev`) to compile; the change is type-identical to the compile-verified datatool fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15463,7 +15463,6 @@ dependencies = [
  "argh",
  "bitcoin",
  "bitcoind-async-client",
- "borsh",
  "rand_core 0.6.4",
  "reth-chainspec",
  "serde_json",
@@ -16603,7 +16602,6 @@ name = "strata-sp1-guest-builder"
 version = "0.3.0-alpha.1"
 dependencies = [
  "bincode",
- "borsh",
  "cargo_metadata 0.19.2",
  "cfg-if",
  "once_cell",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1140,7 +1140,7 @@ dependencies = [
  "threadpool",
  "tokio",
  "tracing",
- "zkaleido 0.1.0 (git+https://github.com/alpenlabs/zkaleido?tag=v0.1-beta.3)",
+ "zkaleido",
  "zkaleido-sp1-host",
 ]
 
@@ -1305,7 +1305,7 @@ dependencies = [
  "tokio",
  "tracing",
  "typed-sled",
- "zkaleido 0.1.0 (git+https://github.com/alpenlabs/zkaleido?tag=v0.1-beta.3)",
+ "zkaleido",
 ]
 
 [[package]]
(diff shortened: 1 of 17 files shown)


purusang (approved):

storopoli (approved):
ACK 1f0f3b47308cb5604b2511c6cee5a3feaf908d22

krsnapaudel (approved):

PR #1969, opened 2026-06-10
refactor(ol): make asm checkpoint-types source of truth for OL log payloads

prajwolrg:
## Description

The OL log-payload layer was defined twice — once in `strata-ol-chain-types-new` and once in the upstream `strata-asm-proto-checkpoint-types` (`alpenlabs/asm`): the `OLLogType` trait, `SimpleWithdrawalIntentLogData`, the `SIMPLE_WITHDRAWAL_INTENT_LOG_TYPE_ID` type id, and the log-decode error. These are a wire contract shared with checkpoint verification, so two independent copies can silently drift apart (e.g. a type-id or field change on one side makes withdrawal intents undecodable on the other).

This PR makes `strata-asm-proto-checkpoint-types` the single source of truth for those types. `strata-ol-chain-types-new` now re-exports them instead of redefining them. Because asm's `OLLogType` trait is minimal (just the type id), the typed-log encode/decode that previously lived on chain-types' trait moves onto chain-types' `OLLog` as `from_log`/`try_into_log`, mirroring asm's own `OLLog` API. Consumers in `ol/stf`, `ol/block-assembly`, and `chain-worker-new` are updated to the relocated API.

No wire formats change. `SimpleWithdrawalIntentLogData`'s `dest` bound goes from 255 to the asm default (`u32::MAX`); this only loosens decode-side validation and is byte-compatible for existing values.

### Type of Change

- [x] Refactor

## Notes to Reviewers

**Why `SimpleWithdrawalIntentLogData` belongs in asm.** ASM checkpoint verification has to *interpret* withdrawal-intent logs — it decodes `SimpleWithdrawalIntentLogData` out of the checkpoint sidecar's `OLLog`s to validate withdrawals. So that type (plus the trait, type id, and decode error it rides on) has to be defined where ASM can reach it, i.e. in `strata-asm-proto-checkpoint-types`. Keeping a second copy in the OL repo was the duplication this PR removes.

**Why other OL logs are not moved.** `SnarkAccountUpdateLogData` (type id `0x02`) is emitted within OL blocks but is *not* consumed by ASM checkpoint verification today, so it stays in `strata-ol-chain-types-new` and just implements asm's re-exported `OLLogType`. Only the subset ASM actually interprets has been hoisted upstream.

**`OLLog` is intentionally NOT deduplicated.** The two `OLLog` containers look identical but carry different payload caps and serve different layers:
- chain-types `OLLog`: `MAX_LOG_PAYLOAD_LEN = 4 KiB`, block-level logs — must hold snark-account updates whose `extra_data` reaches ~1 KiB.
- checkpoint-types `OLLog`: `MAX_LOG_PAYLOAD_LEN = 512 B`, checkpoint sidecar — withdrawal-only.

Collapsing chain's `OLLog` onto asm's would lower the block-level cap to 512 B and panic on real snark-update logs, and asm is an upstream dependency we can't widen from this repo. The existing chain→checkpoint `OLLog` conversion in `ol/checkpoint/src/state.rs` stays as the deliberate layer boundary.

**Open question for discussion.** As more OL log types appear, we need a clear convention for where each log-payload type lives: stays OL-local vs. hoisted into asm once ASM needs to interpret it. Right now the split is "asm owns what ASM interprets; OL owns the rest," but that's implicit. Worth deciding whether to (a) formalize this rule, (b) carve out a dedicated shared crate for the OL↔ASM log contract, or (c) document it in the relevant SPS (SPS-ol-chain-structures / SPS-62). Flagging here and starting a thread so we converge before the next log type lands.

- [ ] Yes
- [x] No

(Touches the OL↔ASM log contract described in SPS-ol-chain-structures / SPS-62, but does not change either spec.)

## Checklist

## Related Issues

### DIFF
--- a/crates/ol/chain-types/src/log_payloads.rs
+++ b/crates/ol/chain-types/src/log_payloads.rs
@@ -1,14 +1,16 @@
 //! Log payload types for orchestration layer logs.
+//!
+//! The canonical log-payload types shared with checkpoint verification
+//! ([`OLLogType`], [`SimpleWithdrawalIntentLogData`], the type-id namespace, and
+//! [`OLLogDecodeError`]) live in [`strata_asm_proto_checkpoint_types`] and are re-exported from the
+//! crate root. This module only holds the OL-block-local log types that are not consumed by
+//! checkpoint verification, such as [`SnarkAccountUpdateLogData`].
 
-use std::iter;
-
-use strata_codec::{Codec, CodecError, VarVec, decode_buf_exact};
-use strata_msg_fmt::{Msg, MsgRef, TypeId, try_encode_into_buf};
+use strata_asm_proto_checkpoint_types::OLLogType;
+use strata_codec::{Codec, VarVec};
+use strata_msg_fmt::TypeId;
 
 use crate::{SAU_MAX_EXTRA_DATA_BYTES, SauTxUpdateData};
-
-/// Maximum byte length for a withdrawal destination BOSD descriptor.
-const MAX_DEST_BYTES: u32 = 255;
 
 /// Maximum byte length for snark account update extra data (matches
 /// `SAU_MAX_EXTRA_DATA_BYTES` from the SSZ spec).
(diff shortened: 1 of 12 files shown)


crates/ol/chain-types/src/log_payloads.rs
```diff
@@ -1,115 +1,30 @@
...
-
-    /// Get the destination as bytes.
-    pub fn dest(&self) -> &[u8] {
-        self.dest.as_ref()
-    }
-}
+/// Shares the OL log type-id namespace with
+/// [`SIMPLE_WITHDRAWAL_INTENT_LOG_TYPE_ID`](strata_asm_proto_checkpoint_types::SIMPLE_WITHDRAWAL_INTENT_LOG_TYPE_ID).
+/// This log is emitted within OL blocks but is not consumed by checkpoint verification.
+pub const SNARK_ACCOUNT_UPDATE_LOG_TYPE_ID: TypeId = 0x02;
```
krsnapaudel, line 27: Question on the shared type-id namespace: the placement rule sorts out which crate a log type lives in, but the ids share one namespace defined in two repos that can't see each other (0x01 in `asm`, 0x02 here). As far as I can tell nothing catches a future 0x03 allocated independently in both.

Would it make sense to single-source the ids (not the types)?

krsnapaudel:
@prajwolrg I guess this should be closed now right?

prajwolrg:
Closing this after https://github.com/alpenlabs/alpen/pull/1969 which uses the deps from `strata-common` instead of `asm`.

PR #1930, opened 2026-06-02
draft: measure SP1 checkpoint capacity for 2048 manifests

krsnapaudel:
## Description

Measurement-only PR. Do not merge.

This PR triggers a branch-gated SP1 network proof job for
`CHECKPOINT_CAPACITY_MANIFEST_COUNTS=2048`, repeated 3 times, to evaluate whether
`MAX_SEALING_MANIFEST_COUNT=2048` has acceptable proof stability/runtime.

Also exercises max checkpoint log limit. For deposit log limit, it seems pending ASM logs limit kicks in limiting deposit logs per ASM manifest to 32.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update
- [x] Experitment 

## Notes to Reviewers

Is this PR addressing any specification, design doc or external reference document?

- [ ]  Yes
- [x]  No

If yes, please add relevant links:

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have commented my code where necessary.
- [ ] I have updated the documentation if needed.
- [ ] My changes do not introduce new warnings.
- [ ] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [ ] New and existing tests pass with my changes.
- [ ] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

## Related Issues

[STR-2750](https://alpenlabs.atlassian.net/browse/STR-2750)

[STR-2750]: https://alpenlabs.atlassian.net/browse/STR-2750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1975, opened 2026-06-11
STR 3716 Include manifests in OL blocks as soon as they are ready

krsnapaudel:
## Description

STR-3308 (#1864) let the OL STF buffer ASM manifest logs from any block, but block assembly still emitted ASM manifests only on terminal blocks. This PR includes buried/mature ASM manifests as soon as they mature, in any OL block, advancing the L1/MMR cursor earlier to reduce EE-update and withdrawal latency.

Manifests are still only included once buried below `l1_reorg_safe_depth` (STR-3555). Manifest effects still land at the epoch-terminal drain; only inclusion and cursor advancement move earlier.

### Changes

1. **refactor(ol/block-assembly): centralize epoch sealing decisions** — separates pure cadence from limit-aware sealing. `LimitAwareSealing` composes cadence with checkpoint-size and manifest-count limit rules, and terminality now comes from an explicit `EpochSealingDecision`, not manifest presence.

2. **feat(ol/block-assembly): include ASM manifests per block** — fetches buried ASM manifests every block and selects a prefix bounded by pending-ASM-log capacity and the epoch manifest budget. Epoch resource totals are threaded across blocks and reset at epoch boundaries. Non-terminal blocks may carry manifests; terminal blocks may have no manifests; empty manifest containers are omitted.

3. **fix(chain-worker-new): replay epoch manifests without per-block cap** — DA reconstruction passes epoch manifests as a `Vec<AsmManifest>` to `apply_da_epoch` instead of wrapping them in the per-block SSZ manifest container, so epochs with manifests spread across blocks or exceeding the per-block container bound reconstruct correctly. Empty L1 ranges are valid when an epoch seals without new buried manifests.

4. **fix(ol/state-types): raise pending ASM log buffer** — raises `MAX_PENDING_ASM_LOGS` from `1 << 16` to `1 << 18`, enough for the expected heavy-manifest case of 2048 manifests with 128 logs each. This changes the `OLState` SSZ root shape.

### Testing

Covers per-block inclusion, manifest-budget sealing, backlog draining across epochs, pending-log prefix limiting, epoch-resource reset, empty manifest containers, empty checkpoint-sync L1 ranges, and DA reconstruction with spread/over-cap manifests.

Targeted crate suites pass. The slow/fast terminal-block functional tests now validate manifest inclusion across the completed epoch instead of requiring manifests on the terminal block itself, and both targeted functional tests pass.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

We have tested 128 logs per manifest in #1930. So setting `MAX_PENDING_ASM_LOGS = 128 * MAX_SEALING_MANIFEST_COUNT = 1 << 18`.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-3716](https://alpenlabs.atlassian.net/browse/STR-3716)

[STR-3716]: https://alpenlabs.atlassian.net/browse/STR-3716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/ol/block-assembly/src/block_assembly.rs
+++ b/crates/ol/block-assembly/src/block_assembly.rs
@@ -16,7 +16,7 @@ use strata_ledger_types::{
 use strata_ol_chain_types_new::*;
 use strata_ol_mempool::MempoolTxInvalidReason;
 use strata_ol_state_support_types::{DaAccumulatingState, WriteTrackingState};
-use strata_ol_state_types::WriteBatch;
+use strata_ol_state_types::{MAX_PENDING_ASM_LOGS, WriteBatch};
 use strata_ol_stf::*;
 use strata_snark_acct_types as _;
 use tracing::{debug, error, warn};
@@ -25,12 +25,12 @@ use crate::{
     AccumulatorProofGenerator, BlockAssemblyResult, BlockAssemblyStateAccess, MempoolProvider,
     checkpoint_size::LogMetrics,
     context::BlockAssemblyAnchorContext,
-    da_tracker::AccumulatedDaData,
     epoch_sealing::{
         EpochSealingLimitAction, EpochSealingLimitVerdict, EpochSealingPolicy,
         EpochSealingResourceStats,
     },
     error::BlockAssemblyError,
+    resource_state::{AccumulatedDaData, EpochResourceState},
     types::{BlockGenerationConfig, BlockTemplateResult, FailedMempoolTx, FullBlockTemplate},
 };
 
@@ -45,6 +45,24 @@ struct ProcessTransactionsOutput<S: IStateAccessor> {
     /// Accumulated da data after processing all transactions.
     accumulated_da: AccumulatedDaData,
     /// Non-cadence limits requesting an epoch seal, if any.
+    sealing_limit_verdict: EpochSealingLimitVerdict,
+}
+
+/// Inputs consumed while finalizing a block template.
+struct BuildBlockTemplateInput<A> {
+    accumulated_batch: WriteBatch<A>,
+    output_buffer: ExecOutputBuffer,
+    successful_txs: Vec<OLTransaction>,
+    is_terminal: bool,
+    manifest_container: Option<OLAsmManifestContainer>,
+}
+
+/// ASM manifests selected for the candidate OL block.
+///
+/// The selected manifests are still raw manifests; the caller wraps them in an
+/// [`OLAsmManifestContainer`] when the sequence is non-empty.
+struct SelectedAsmManifests {
+    manifests: Vec<AsmManifest>,
     sealing_limit_verdict: EpochSealingLimitVerdict,
 }
 
@@ -148,8 +166,8 @@ pub(crate) struct ConstructBlockOutput<S> {
         expect(dead_code, reason = "only used by tests")
     )]
     pub(crate) post_state: S,
-    /// Accumulated DA data for the constructed block.
-    pub(crate) accumulated_da: AccumulatedDaData,
+    /// Resource state for the constructed block's epoch.
+    pub(crate) resource_state: EpochResourceState,
 }
 
 /// Generate a block template from the given configuration.
@@ -167,7 +185,7 @@ pub(crate) async fn generate_block_template_inner<C, E>(
     epoch_sealing_policy: &E,
     sequencer_config: &SequencerConfig,
     block_generation_config: BlockGenerationConfig,
-    parent_da: AccumulatedDaData,
+    resource_state_before_block: EpochResourceState,
     bridge_params: BridgeParams,
 ) -> BlockAssemblyResult<BlockTemplateResult>
 where
@@ -206,7 +224,7 @@ where
         block_slot,
         block_epoch,
         mempool_txs,
-        parent_da,
+        resource_state_before_block,
         bridge_params,
     )
     .await?;
@@ -214,7 +232,7 @@ where
     Ok(BlockTemplateResult::new(
         output.template,
         output.failed_txs,
-        output.accumulated_da,
+        output.resource_state,
     ))
 }
 
@@ -247,8 +265,9 @@ pub(crate) fn calculate_block_slot_and_epoch<S: IStateAccessor>(
 /// 2. Executes block initialization (epoch initial + block start)
 /// 3. Validates each transaction against accumulated state
 /// 4. Filters out invalid transactions (proof failures, execution failures)
-/// 5. Applies the sealing policy and fetches L1 manifests for terminal blocks
-/// 6. Builds the complete block with only valid transactions
+/// 5. Fetches and admits buried ASM manifests for this block
+/// 6. Applies the sealing policy
+/// 7. Builds the complete block with only valid transactions
 #[expect(clippy::too_many_arguments, reason = "can't get around the args")]
 #[tracing::instrument(
     skip_all,
@@ -266,7 +285,7 @@ pub(crate) async fn construct_block<C, E>(
     block_slot: Slot,
     block_epoch: Epoch,
     mempool_txs: Vec<(OLTxId, OLTransaction)>,
-    parent_da: AccumulatedDaData,
+    resource_state_before_block: EpochResourceState,
     bridge_params: BridgeParams,
 ) -> BlockAssemblyResult<ConstructBlockOutput<C::State>>
 where
@@ -299,9 +318,11 @@ where
     let output_buffer = ExecOutputBuffer::new_empty();
 
     // Phase 1: Execute block initialization (epoch initial + block start).
-    // `AccumulatedDaData` flows through each phase, accumulating state diffs and logs.
+    // Resource state flows through each phase, accumulating state diffs, logs, and manifest count.
+    let epoch_cumulative_da = resource_state_before_block.da().clone();
+    let epoch_cumulative_manifest_count = resource_state_before_block.manifest_count();
     let (accumulated_batch, accumulated_da) =
-        execute_block_initialization(parent_state.as_ref(), &block_context, parent_da);
+        execute_block_initialization(parent_state.as_ref(), &block_context, epoch_cumulative_da);
 
     // Phase 2: Process transactions, filtering out invalid ones.
     let ProcessTransactionsOutput {
@@ -309,7 +330,7 @@ where
         failed_txs,
         accumulated_batch,
         accumulated_da,
-        sealing_limit_verdict,
+        mut sealing_limit_verdict,
     } = process_transactions(
         ctx,
         epoch_sealing_policy,
@@ -319,11 +340,32 @@ where
         accumulated_batch,
         mempool_txs,
         accumulated_da,
+        epoch_cumulative_manifest_count,
         bridge_params,
     );
 
-    // Phase 3: Ask the sealing policy whether this block should be terminal. Fetch manifests for
-    // the terminal block (possibly empty if L1 is slow).
+    // Phase 3: Fetch buried ASM manifests for this block. Manifest selection may also request
+    // that this block becomes terminal.
+    let SelectedAsmManifests {
+        manifests: selected_asm_manifests,
+        sealing_limit_verdict: manifest_limit_verdict,
+    } = fetch_asm_manifests_for_block(
+        ctx,
+        epoch_sealing_policy,
+        parent_state.as_ref(),
+        epoch_cumulative_manifest_count,
+    )
+    .await?;
+    sealing_limit_verdict.merge(manifest_limit_verdict);
+    let selected_asm_manifest_count = u32::try_from(selected_asm_manifests.len())
+        .expect("selected ASM manifest count fits in u32");
+    let manifest_container = if selected_asm_manifests.is_empty() {
+        None
+    } else {
+        Some(OLAsmManifestContainer::new(selected_asm_manifests)?)
+    };
+
+    // Phase 4: Ask the sealing policy whether this block should be terminal.
     let sealing_decision =
         epoch_sealing_policy.should_seal_epoch(block_slot, &sealing_limit_verdict);
     let should_seal = sealing_decision.should_seal();
@@ -334,13 +376,8 @@ where
         should_seal,
         "epoch seal decision"
     );
-    let manifest_container = if should_seal {
-        fetch_asm_manifests_for_terminal_block(ctx, parent_state.as_ref()).await?
-    } else {
-        None
-    };
 
-    // Phase 4: Finalize block construction.
+    // Phase 5: Finalize block construction.
     // Clone output_buffer: the clone goes to build_block_template (which adds manifest logs
     // for the header), the original is consumed below to append this block's tx logs to DA.
     debug!(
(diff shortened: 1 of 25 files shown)


crates/ol/block-assembly/src/block_assembly.rs
```diff
@@ -318,82 +343,208 @@ where
...
+            // Pending ASM-log capacity is a hard admission bound, not a sealing trigger. Stop
+            // manifest inclusion and leave the remaining manifests for a later block; the epoch
+            // still seals only through cadence or another sealing limit.
+            break;
+        }
+
+        let candidate_manifest_count = epoch_cumulative_manifest_count
+            .saturating_add(selected_asm_manifests.len())
+            .saturating_add(1);
+        let stats = base_resource_stats.with_manifest_count(candidate_manifest_count);
```
storopoli: The manifest admission path still only checks the already-accumulated DA/log stats plus the epoch manifest count; it does not budget the DA/log output that these admitted ASM logs will create when the terminal block drains them. Here `base_resource_stats` comes from the pre-drain state:

https://github.com/alpenlabs/alpen/blob/7d2c655860e684174382ab8f35d6aa9260f056b2/crates/ol/block-assembly/src/block_assembly.rs#L348-L351

and the real effects are applied later in `process_block_manifests`/`process_epoch_terminal`:

https://github.com/alpenlabs/alpen/blob/7d2c655860e684174382ab8f35d6aa9260f056b2/crates/ol/block-assembly/src/block_assembly.rs#L812-L843

A block can therefore admit a large manifest prefix and only discover the resulting DA/log growth after the terminal decision. Checkpoint payload construction later enforces those final limits when building the sidecar:

https://github.com/alpenlabs/alpen/blob/7d2c655860e684174382ab8f35d6aa9260f056b2/crates/ol/checkpoint/src/state.rs#L177-L193

so the sequencer can produce a terminal OL block that cannot produce a checkpoint payload. Please either simulate/budget the terminal-drain effects before admitting the manifest prefix, or conservatively cap selected manifests by expected terminal DA/log output and add a regression with enough deposit-bearing manifests to cross the checkpoint DA/log limit.

storopoli (changes requested):

crates/ol/block-assembly/src/block_assembly.rs
krsnapaudel: @storopoli I have tried to address. PTAL.

storopoli (commented):
## Review

Solid, well-tested work that cleanly decouples sealing policy from manifest inclusion and correctly handles the DA-reconstruction edge cases. Below are the highlights and a ranked list of issues/risks. Nothing here looks like a clear blocking correctness bug, so this is **approve-with-comments** in spirit — but I'd like the author to address points 1–3 before merge.

### What's done well

- **Clean trait decomposition.** Separating cadence (`CadencePolicy`/`FixedSlotSealing`) from limit-aware sealing (`LimitAwareSealing`), and making terminality an explicit `EpochSealingDecision` rather than "block carries manifests ⇒ terminal", is a real improvement. The `EpochSealingLimitVerdict` merge model (most-restrictive action per limit) is tidy.
- **Projection correctness.** `process_pending_asm_log_effect` reuses the exact `process_asm_log` call that `process_epoch_terminal` uses per log, with a doc comment flagging the behavioral-alignment coupling. Right way to keep the sealing projection honest.
- **DA reconstruction fix is correct.** The `to_height == base_l1_height ⇒ empty Vec` branch correctly handles cadence-sealed epochs with no new manifests (previously a false `L1RangeInverted`), and the manifest-cap check is preserved. Dropping the per-block `OLAsmManifestContainer` bound on replay is the right call now that manifests can span blocks / exceed the per-block container.
- **Strong test coverage (97.7% patch).** New `EpochShape` fixtures cover empty-range, spread-across-blocks, and over-cap replay; assembly tests cover per-block inclusion, budget sealing, backlog drain, pending-log prefix limiting, resource reset, and empty containers. `assert_checkpoint_sidecar_builds_for_terminal_output` validating a real sidecar build on the DA-overflow path is a nice touch.
- Mechanical renames are complete — no dangling references to `EpochDaTracker`, `fetch_epoch_da_until_parent`, `slots_per_epoch()`, or `ManifestContainerBuild`.

### Issues & risks (ranked)

**1. Per-block re-projection of the entire pending-ASM-log buffer (performance — medium).**
`fetch_asm_manifests_for_block` runs on *every* block and calls `ProjectedAsmLogOutput::new`, which clones the parent state, applies the accumulated write batch, then re-drains **all** currently-pending ASM logs through real STF processing (`project_existing_pending_logs` over `0..pending_count`). Since the pending queue only drains at the terminal block, it grows across the epoch and is re-projected from scratch each block — roughly `O(pending × blocks_per_epoch)`, now bounded by the raised `1<<18` cap. Even a block carrying no new manifests pays this. Consider memoizing the projection incrementally across the epoch, or at minimum document the cost and confirm it's acceptable for the 2048-manifests/epoch target.

**2. Projection-overflow / base-reject becomes a sequencer halt (robustness — medium).**
Two paths return a hard `BlockAssemblyError::Other(..)`: `ProjectedAsmLogOutput::new` when already-buffered pending logs would exceed the terminal block log limit, and `fetch_asm_manifests_for_block` on a base `RejectCandidate` "before admitting new manifests". These are framed as invariants, with the soft 90% (`max_logs * 9 / 10`) threshold supposed to guarantee headroom (~410 logs). But that headroom must absorb a single block's combined tx-log + manifest-drain growth on top of a near-full buffer; a tx-heavy block on a parent whose pending buffer is already near the threshold could trip `new()` and halt block production rather than degrade gracefully. Suggest (a) a dedicated typed `BlockAssemblyError` variant instead of `Other(String)` (the repo's error guidance prefers typed enums), and (b) a written argument or test that the 90% headroom strictly dominates worst-case single-block log growth.

**3. `MAX_PENDING_ASM_LOGS` change is SSZ-root/consensus-affecting, but classified non-breaking (note).**
`1<<16 → 1<<18` changes the `OLState` SSZ list bound (generated from `state.ssz`) and therefore the state-root tree shape. The body acknowledges this, but the "Type of Change" checkboxes mark it non-breaking. Fine pre-launch — just confirm genesis/fixture roots are regenerated and flag it for anyone tracking state-format compatibility.

**4. Projection keys off `manifest.height()`; real buffering uses the cursor-derived height (low).**
`project_manifest` builds `PendingAsmLog::new(manifest.height(), …)`, whereas `handle_asm_manifest` buffers with `real_height` from `next_manifest_height(last_l1_height, index)`. These coincide in normal operation (and `process_asm_log` output appears height-independent for count/size, so the resource estimate stays valid), but the assumption is implicit — a one-line comment on why `manifest.height()` is a safe stand-in here would prevent a future footgun.

**5. API surface nits (low).**
- `EpochSealingPolicy` is `pub` and references `pub` stats/verdict types, but their constructors are `pub(crate)`, so no external crate can implement the trait. If sealed-by-design, a doc note helps; otherwise it's effectively dead public surface.
- `manifest_count_action` is test-only via `#[cfg_attr(not(test), expect(dead_code, …))]` — acceptable, mild smell.
- The 90% threshold `max_logs * 9 / 10` is a magic number; a named const with rationale would read better.

### Test coverage gap

`asm_log_projection.rs` patch coverage is ~82.5% — the uncovered lines are almost certainly the error arms (`BlockConstruction` mapping, the `project_existing_pending_logs` overflow `Err`, the non-overflow `Err(err)` arm). Given point 2 turns these into halt conditions, a targeted unit test driving `ProjectedAsmLogOutput::new` into the overflow error would be worth adding.

### Security

No new external input surface; manifest inclusion stays gated by `l1_reorg_safe_depth` and capped by `MAX_SEALING_MANIFEST_COUNT`, and reconstruction keeps range/cap validation. No unsafe code or secrets. The main concern is the halt-vs-degrade behavior in point 2, not a vulnerability.

---
_Generated by [Claude Code](https://claude.ai/code)_

storopoli (approved):
ACK 026a2fa80028b266afd0112eb9d92e9216b92196

crates/ol/block-assembly/src/block_assembly.rs
```diff
@@ -318,82 +354,276 @@ where
...
+        {
+            // Pending ASM-log capacity is a hard admission bound, not a sealing trigger. Stop
+            // manifest inclusion and leave the remaining manifests for a later block; the epoch
+            // still seals only through cadence or another sealing limit.
+            break;
+        }
+
+        let candidate_manifest_count = epoch_cumulative_manifest_count
+            .saturating_add(selected_asm_manifests.len())
+            .saturating_add(1);
```
bewakes, line 524: This seems to be an extra addition. Can you explain why is this needed?

bewakes (commented):
Thanks for this PR.

I think the logic for including manifests in non-terminal blocks can be much simpler. Processing a manifest contributes no DA-relevant size - manifets don't and shouldn't add to the state iff or OL logs(that'd be redundant with L1), and their only effective state update is `last_l1_height` (a register). `pending_asm_logs` are cleared at the end of terminal block anyway. The terminal-log projection and the DA accounting threaded through manifest selection seem to be guarding the things the manifest path can't actually modify. 

The one real constraint is that `pending_asm_logs` accumulates across the epoch and only drains at the terminal block, so it can fill mid-epoch. But that's a single capacity bound - a per-epoch(and optionally per-block) ASM log/manifest budget, checked as plain arithmetic. Beyond that, the sealing logic should stay more or less the same. Am I missing a case where manifest processing contributes to DA?

crates/ol/block-assembly/src/block_assembly.rs
krsnapaudel: @bewakes The +1 is the manifest currently being evaluated — it's the projected epoch count if this candidate is admitted (already-carried + already-selected-this-block + this one).

krsnapaudel:
> Thanks for this PR.
> 
> I think the logic for including manifests in non-terminal blocks can be much simpler. Processing a manifest contributes no DA-relevant size - manifets don't and shouldn't add to the state iff or OL logs(that'd be redundant with L1), and their only effective state update is `last_l1_height` (a register). `pending_asm_logs` are cleared at the end of terminal block anyway. The terminal-log projection and the DA accounting threaded through manifest selection seem to be guarding the things the manifest path can't actually modify.
> 
> The one real constraint is that `pending_asm_logs` accumulates across the epoch and only drains at the terminal block, so it can fill mid-epoch. But that's a single capacity bound - a per-epoch(and optionally per-block) ASM log/manifest budget, checked as plain arithmetic. Beyond that, the sealing logic should stay more or less the same. Am I missing a case where manifest processing contributes to DA?

I added the bookkeeping after @storopoli's comment (resolved above). Seems @bewakes is right.
Verified in code: **manifest processing contributes nothing to the checkpoint DA, checkpoint OL logs, or `MAX_LOGS_PER_BLOCK`.**

- The checkpoint payload is built **pre-drain**: `replay_epoch_and_compute_da` → `execute_block_batch_predrain` (`crates/ol/stf/src/assembly.rs:472`), which skips the terminal drain; the drain effects are re-derived from the L1 manifests during reconstruction (`apply_da_epoch`: "the deferred drain effects, which the DA diff does not carry"). So a manifest prefix can never push the checkpoint payload over a limit.
- ASM-log drain emits **no OL logs**: deposits only `add_balance` (`account_processing.rs:17-64`); only bridge-gateway/withdrawal messages emit logs, and those originate from txs, not manifests. CheckpointTip/predicate handlers emit none.

So @storopoli's concern (a terminal block whose drain overflows the checkpoint) can't actually occur — the payload excludes the drain. The terminal-drain DA/log projection (`BufferedAsmLogProjection`, `TerminalBlockLogCountRule`, the projection-fed `CheckpointSizeRule`) is guarding a path the manifest flow can't take.

The only real per-epoch bounds the manifest path touches are:
1. **`pending_asm_logs` capacity** — buffered across the epoch, drains at terminal; a plain arithmetic count bound.
2. **manifest count** (proving budget) — already enforced by `ManifestCountRule`.

Plan: drop the terminal-drain projection and revert to those two arithmetic bounds, as @bewakes suggested.
@storopoli Please check and let me know if this is sound.

storopoli:
True today but if a future ASM-log handler starts emitting an OL log (or a deposit ever resolves to the bridge-gateway serial), the overflow risk silently returns. Worth a comment at the drain plus a focused regression asserting the drain yields zero OL logs and no DA delta.

What you think?

Otherwise fine to go ahead.

bewakes:
It is quite unlikely that we would produce logs to post to L1 from manifests because that would be redundant. So I am inclined towards keeping things simple. We can have a proper note/comment instead.

crates/ol/block-assembly/src/da_tracker.rs
```diff
@@ -153,6 +162,31 @@ async fn fetch_state<C: BlockAssemblyAnchorContext>(
     Ok(ol_state)
 }
 
+/// Resource totals accumulated so far in an epoch.
+#[derive(Clone, Debug)]
+pub(crate) struct EpochResourceTotals {
+    da: AccumulatedDaData,
```
delbonis: Since this is just meant for tracking the "totals", we can just have this track the *encoded size* of the DA data instead of storing it wholly.  These could be a significant amount of data and we persist it anyways.  Since this struct is just used for decision-making in-memory, we don't need it here.

crates/ol/block-assembly/src/epoch_sealing.rs
```diff
@@ -1,22 +1,363 @@
...
+/// Resource stats used by sealing-limit rules.
+///
+/// Block assembly builds this snapshot incrementally before admitting a
+/// candidate resource, such as a transaction or manifest sequence. Rules ignore
+/// optional facts that are not present in a given snapshot.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct EpochSealingResourceStats {
+    da_diff_size: usize,
+    log: LogMetrics,
+    manifest_count: Option<usize>,
```
delbonis: Why not just `usize`?  Or really `u32`?

crates/ol/block-assembly/src/epoch_sealing.rs
```diff
@@ -1,22 +1,363 @@
...
+            log,
+            manifest_count: None,
+        }
+    }
+
+    /// Adds the epoch-cumulative ASM manifest count.
+    pub(crate) fn with_manifest_count(mut self, manifest_count: usize) -> Self {
+        self.manifest_count = Some(manifest_count);
+        self
+    }
```
delbonis: The `manifest_count` shouldn't be implicit in `new` and this is not a builder.

delbonis (changes requested):
This is pretty messy but that's not really the fault of this PR.  Just something to keep in mind for the future.

I have a few cleanup suggestions.

crates/ol/block-assembly/src/da_tracker.rs
krsnapaudel: @delbonis Good catch on the name. The `da` field isn't just decision stats here — it's the running epoch DA accumulator that gets threaded as the next block's `parent_da`, so it has to be cached either way. Pulling it out would just move the same accumulator into a second parallel map keyed by the same blocks (no memory saved, extra sync). So I'll rename the struct to `EpochResourceState`.

delbonis (approved):

PR #1982, opened 2026-06-12
storage manager cleanup

sapinb:
## Description
- [x] Move NodeStorage to own module
- [x] Add atomic cache length counter
- [x] Use caching in ClientStateManager

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
This PR is part of TODO cleanup efforts.

Length checks in cache is currently unused outside tests.

~ClientStateManager is only being read during application startup to get latest state to continue work on top of. Caching is useless for this, hence removed.~
  

- [ ]  Yes
- [ ]  No

## Checklist


crates/storage-common/src/cache.rs
```diff
@@ -59,35 +68,34 @@ impl<K: Clone + Eq + Hash, V: Clone, E: From<OpsError> + error::Error + Clone> C
...
         Self {
             cache: Mutex::new(lru::LruCache::new(size)),
+            len: AtomicUsize::new(0),
         }
     }
 
     /// Gets the number of elements in the cache.
-    // TODO(STR-3679): replace this with an atomic we update after every op
     #[allow(dead_code, clippy::allow_attributes, reason = "used for debugging")]
     pub async fn get_len_async(&self) -> usize {
```
krsnapaudel: Is it really async now?

krsnapaudel (approved):
Minor nit. Otherwise looks good.

crates/storage-common/src/cache.rs
```diff
@@ -335,6 +355,7 @@ impl<K: Clone + Eq + Hash, V: Clone, E: From<OpsError> + error::Error + Clone> C
                 let mut cache_guard = self.cache.blocking_lock();
                 trace!("re-acquired cache lock");
                 safely_remove_cache_slot(&mut cache_guard, k, &slot);
+                self.len.store(cache_guard.len(), Ordering::Relaxed);
```
storopoli, line 336: This cleanup path still calls `safely_remove_cache_slot`, whose [`cache.peek(key).unwrap()`](https://github.com/alpenlabs/alpen/blob/d049ec2d3245bf4513840e8e8638b0de37b7fd73/crates/storage-common/src/cache.rs#L373) can panic if the pending slot has already disappeared. The fetch paths push a `Pending` slot, then release the cache lock while the database fetch runs; during that window LRU pressure, `purge_*`, or `clear` can remove the key. If the fetch later returns an error, cleanup reaches that `unwrap()` instead of treating the missing key as "not removed by us".

Relevant paths: [`get_or_fetch` inserts then cleans up after reacquiring the lock](https://github.com/alpenlabs/alpen/blob/d049ec2d3245bf4513840e8e8638b0de37b7fd73/crates/storage-common/src/cache.rs#L227-L266), and the same issue exists in [`get_or_fetch_blocking`](https://github.com/alpenlabs/alpen/blob/d049ec2d3245bf4513840e8e8638b0de37b7fd73/crates/storage-common/src/cache.rs#L325-L358).

Could we make the helper tolerate a missing key?

```rust
let Some(current_slot) = cache.peek(key) else {
    return false;
};

let is_eq = Arc::ptr_eq(current_slot, slot);
if is_eq {
    cache.pop(key);
}
is_eq
```

storopoli (changes requested):

crates/storage-common/src/cache.rs
```diff
@@ -59,35 +68,34 @@ impl<K: Clone + Eq + Hash, V: Clone, E: From<OpsError> + error::Error + Clone> C
     pub fn new(size: NonZeroUsize) -> Self {
         Self {
             cache: Mutex::new(lru::LruCache::new(size)),
+            len: AtomicUsize::new(0),
```
delbonis, line 74: Is this definitely safe?  I wonder if there's some race condition that could crop up.  I guess if we only ever write to it while the lock is being held then it *should* be okay.

Regarding @krsnapaudel's comment: We could merge the two `get_len_` fns into one since they're the same.

delbonis (changes requested):
One minor comment.  Looks good otherwise though.

bewakes (approved):

bewakes:
@sapinb this needs rebase.

crates/storage-common/src/cache.rs
sapinb: removed get_len_blocking and get_len_async for just get_len

crates/storage-common/src/cache.rs
sapinb: updated.

crates/storage-common/src/cache.rs
sapinb: As of now, this is safe as it is only written while the lock is held. 
Added a comment documenting this invariant for potential future updates.

The two get_len_ fns are also merged into a single get_len()

crates/storage/src/managers/client_state.rs
```diff
@@ -172,41 +176,54 @@ impl ClientStateManager {
...
+            state,
+        })
     }
 
-    fn is_idx_better(&self, idx: L1Height) -> bool {
-        self.last_idx.is_none_or(|v| idx >= v)
+    fn is_height_better(&self, block: L1BlockCommitment) -> bool {
+        self.state
+            .as_ref()
+            .is_none_or(|s| block.height() >= s.last_block.height())
```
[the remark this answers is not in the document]
sapinb: resolved in 6abc027711e35360ead97df8dd763014903f8da8

sapinb:
@codex review

storopoli (approved):
ACK 6abc027711e35360ead97df8dd763014903f8da8

krsnapaudel (approved):

crates/storage/src/node_storage.rs
```diff
@@ -0,0 +1,511 @@
...
+            l1_writer_manager: self.l1_writer_manager.clone(),
+        }
+    }
+}
+
+impl NodeStorage {
+    /// Returns the raw database backend for direct access to databases without managers.
+    pub fn db(&self) -> &Arc<SledBackend> {
+        &self.db
+    }
```
delbonis, line 124: Ah crap I didn't realize this was being depended on by this crate, maybe we should remove this dep and expose NodeStorage as being generic over the backend?

delbonis (approved):
One comment that maybe should be part of a larger node crate refactor project.

crates/storage/src/node_storage.rs
sapinb: Agree that this should be part of a larger refactor. Out of scope for this cleanup.
Ideally we should have managers for all the operations and not need to use the ops directly in the services.

sapinb: https://alpenlabs.atlassian.net/browse/STR-3951

PR #1984, opened 2026-06-12
Str 3466 csm reorg handling

bewakes:
## Description
CSM previously only kept track of the last asm block processed and this made it unable to properly handle reorgs. The main change this PR makes is that it replaces the last processed asm blocks to a `recently_processed_asm_blocks: Vec<L1BlockCommitment>` which keeps track of processed asm blocks, anchored at the last finalized l1 block(which is the first item in the list).

Reorg handling does not goes beyond the finalized L1 block(the first item in the list).

**AI usase**: This was created with the help of claude code.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/csm-worker/src/processor.rs
+++ b/crates/csm-worker/src/processor.rs
@@ -16,7 +16,7 @@ use crate::{
     checkpoint_extract::{CheckpointVerificationContext, extract_matching_checkpoint},
     context::CsmWorkerContext,
     errors::{CsmWorkerError, CsmWorkerResult},
-    state::{CsmWorkerState, compute_reorg_floor_height, derive_state},
+    state::{CsmWorkerState, compute_reorg_floor_height, reload_client_state},
 };
 
 /// The in-flight CSM update produced by processing one ASM block's logs.
@@ -128,7 +128,7 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
             "received asm block skipping a height: tip {last}, block {asm_block}"
         );
 
-        let state = derive_state(&self.ctx, &asm_block, &next_state)?;
+        let state = reload_client_state(&self.ctx, &asm_block, &next_state)?;
         self.ctx
             .put_client_state_update(&asm_block, ClientUpdateOutput::new_state(state.clone()))?;
         self.recent_asm_blocks.push(asm_block);
@@ -276,7 +276,7 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
                 what: "client state at pivot block",
                 detail: pivot_block.to_string(),
             })?;
-        let new_clstate = derive_state(&self.ctx, &pivot_block, &pivot_clstate)?;
+        let new_clstate = reload_client_state(&self.ctx, &pivot_block, &pivot_clstate)?;
 
         // Re-persist at the pivot to overwrite the orphaned branch's row.
         self.ctx.put_client_state_update(
@@ -1365,18 +1365,19 @@ mod tests {
 
         let (mut state, storage) = create_test_state_with_ctx(|c| {
             // Fork lands at the anchor (height 100); 101 diverged. Epoch 2's
-            // checkpoint rode the orphaned branch, so its canonical commitment
-            // is gone after the reorg.
-            c.with_canonical_block(100, block_id_at(100))
+            // checkpoint rode the orphaned 101 block, so its L1 observation is
+            // not on the canonical chain (canonical 101 is a different block).
+            c.with_canonical_block(90, block_id_at(90))
+                .with_canonical_block(100, block_id_at(100))
                 .with_canonical_block(101, block_id_at(201))
-                .with_orphaned_epoch(2)
         });
         state.recent_asm_blocks = vec![anchor, orphan_tip];
 
         // CSM persisted a client state at the fork block in a prior run.
         seed_client_state_row(&storage, &anchor);
 
-        // Epoch 1 observed at L1 height 90; epoch 2 observed at the orphaned tip.
+        // Epoch 1 observed at canonical L1 height 90; epoch 2 observed at the
+        // orphaned 101 block (block_id_at(101) != canonical block_id_at(201)).
         let commitment_1 = seed_ol_checkpoint_observation(&storage, 1, 90);
         let commitment_2 = seed_ol_checkpoint_observation(&storage, 2, 101);
 
--- a/crates/csm-worker/src/processor.rs
+++ b/crates/csm-worker/src/processor.rs
@@ -124,7 +124,7 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
     ) -> CsmWorkerResult<()> {
         let state = next_state.as_ref().clone();
         self.ctx
-            .put_client_state_update(&asm_block, ClientUpdateOutput::new(state.clone(), vec![]))?;
+            .put_client_state_update(&asm_block, ClientUpdateOutput::new_state(state.clone()))?;
 
         // The list stays contiguous: a commit either extends the tip by one or
         // replaces it at the same height (reorg). It never skips a height.
@@ -292,8 +292,9 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
             return Ok(());
         }
 
-        // A clean forward extension keeps `last` on the canonical chain.
-        let last_still_canonical = self.ctx.get_canonical_l1_block(last.height())? == last;
+        // A clean forward extension keeps `last` on the canonical chain. A
+        // missing block (height above a reverted tip) means it diverged.
+        let last_still_canonical = self.ctx.get_canonical_l1_block(last.height())? == Some(last);
         let is_pure_extension = last_still_canonical && asm_block.height() > last.height();
         if !is_pure_extension {
             self.reorg_to_fork(&asm_block)?;
@@ -308,7 +309,12 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
 
         // Replay gap blocks, then the target.
         for height in (resume_height + 1)..asm_block.height() {
-            let gap_block = self.ctx.get_canonical_l1_block(height)?;
+            let gap_block = self.ctx.get_canonical_l1_block(height)?.ok_or_else(|| {
+                CsmWorkerError::MissingData {
+                    what: "canonical L1 block",
+                    detail: format!("height {height}"),
+                }
+            })?;
             let gap_state = self.ctx.get_asm_state(&gap_block)?;
             info!(%gap_block, "replaying ASM block skipped by status channel");
             self.process_asm_logs(gap_block, gap_state.logs())?;
@@ -358,12 +364,12 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
     }
 
     /// Highest list index whose block still matches the canonical L1 chain.
-    ///
-    /// `None` if even the finalized anchor (index 0) diverged.
     fn find_fork_index(&self) -> CsmWorkerResult<Option<usize>> {
         for idx in (0..self.recent_asm_blocks.len()).rev() {
             let block = self.recent_asm_blocks[idx];
-            if self.ctx.get_canonical_l1_block(block.height())? == block {
+            // A missing block (above a reverted tip) or a mismatched blkid both
+            // mean this entry diverged; keep walking down.
+            if self.ctx.get_canonical_l1_block(block.height())? == Some(block) {
                 return Ok(Some(idx));
             }
         }
@@ -480,7 +486,9 @@ mod tests {
     };
     use strata_asm_logs::constants::AsmLogTypeId;
     use strata_asm_params::AsmParams;
+    use strata_asm_proto_checkpoint_types::test_utils::create_test_checkpoint_payload;
     use strata_btc_verification::L1Anchor;
+    use strata_checkpoint_types::EpochSummary;
     use strata_csm_types::{CheckpointL1Ref, ClientState, ClientUpdateOutput, L1Checkpoint};
     use strata_db_store_sled::test_utils::get_test_sled_backend;
     use strata_identifiers::RBuf32;
@@ -488,7 +496,7 @@ mod tests {
     use strata_primitives::prelude::*;
     use strata_state::asm_state::AsmState;
     use strata_status::StatusChannel;
-    use strata_storage::create_node_storage;
+    use strata_storage::{NodeStorage, create_node_storage};
     use strata_test_utils::ArbitraryGenerator;
 
     use crate::{errors::CsmWorkerError, state::CsmWorkerState, test_utils::StubCtx};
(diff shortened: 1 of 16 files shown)


bewakes:
@codex review

crates/csm-worker/src/state.rs
```diff
@@ -48,113 +49,116 @@ pub struct CsmWorkerState<C: CsmWorkerContext> {
...
+        // Persist and publish only when the derived state actually changed.
+        if *new_clstate != recent_clstate {
+            let new_update = ClientUpdateOutput::new_state(new_clstate.as_ref().clone());
+            ctx.put_client_state_update(&recent_l1blk, new_update)?;
+            ctx.publish_client_state(new_clstate.as_ref().clone(), recent_l1blk);
         }
 
+        // The list starts at the committed block and refills as blocks are
+        // processed; pruning bounds it to the reorg-safe depth.
+        let recent_asm_blocks = vec![recent_l1blk];
```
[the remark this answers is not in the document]
bewakes: True. Fixed in https://github.com/alpenlabs/alpen/pull/1984/commits/61a3bb139c3a44d151eb138eea396bd0caa78be0

crates/csm-worker/src/processor.rs
```diff
@@ -249,60 +275,101 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
...
-                %asm_block,
-                last_height,
-                "ASM block strictly behind last committed; skipping (possible deeper reorg)"
-            );
-            return Ok(());
+        // A clean forward extension keeps `last` on the canonical chain.
+        let last_still_canonical = self.ctx.get_canonical_l1_block(last.height())? == last;
+        let is_pure_extension = last_still_canonical && asm_block.height() > last.height();
+        if !is_pure_extension {
+            self.reorg_to_fork(&asm_block)?;
```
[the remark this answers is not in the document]
bewakes: Done in https://github.com/alpenlabs/alpen/pull/1984/commits/71bfa64417eec2be07579ec39d64bbb53335b86a

crates/csm-worker/src/processor.rs
```diff
@@ -125,13 +125,39 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
...
-        self.last_asm_block = Some(asm_block);
+
+        // The list stays contiguous: a commit either extends the tip by one or
+        // replaces it at the same height (reorg). It never skips a height.
+        let last = self
+            .recent_asm_blocks
+            .last()
+            .expect("recent_asm_blocks is non-empty");
+        let extends_tip = asm_block.height() == last.height() + 1;
+        let same_height_reorg = asm_block.height() <= last.height();
```
purusang: Should this be `asm_block.height() == last.height();`?

crates/csm-worker/src/processor.rs
```diff
@@ -125,13 +125,39 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
...
+        debug_assert!(
+            extends_tip || same_height_reorg,
+            "commit skipped a height: tip {last}, block {asm_block}"
```
purusang: I think this message does not relay complete information.

krsnapaudel:
@bewakes Codex findings seem to be reasonable. Claude flagged the same issues. Will review after you update.

crates/csm-worker/src/state.rs
```diff
@@ -48,113 +53,160 @@ pub struct CsmWorkerState<C: CsmWorkerContext> {
...
+    // Confirmed: the latest observation seen on L1, else the finalized one.
+    let confirmed_ckpt = observed_checkpoints
+        .back()
+        .cloned()
+        .or_else(|| new_finalized_ckpt.clone());
+
+    let finalized_epoch = new_finalized_ckpt.as_ref().map(EpochCommitment::from);
+    let new_clstate = ClientState::new(new_finalized_ckpt, confirmed_ckpt);
+
+    // Keep only non-finalized candidates for incremental advancement.
```
purusang:
```suggestion
    // Drop candidates that are already finalized, leaving only newer observations 
    // for incremental finality advancement.
```

crates/csm-worker/src/processor.rs
```diff
@@ -255,54 +281,94 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
...
         }
 
-        // TODO(STR-3466): Strictly behind — either a stale message or a deeper reorg.
-        if target_height < last_height {
-            warn!(
-                %asm_block,
-                last_height,
-                "ASM block strictly behind last committed; skipping (possible deeper reorg)"
-            );
-            return Ok(());
```
purusang, line 275: Will we have a regression removing this check if some block lower than `last` is received?

crates/csm-worker/src/processor.rs
```diff
@@ -249,60 +285,124 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
...
+    ///
+    /// Errors if the fork lies at or below the finalized anchor (index 0): a
+    /// reorg past finality is a protocol violation the worker must not absorb.
+    fn reorg_to_fork(&mut self, incoming: &L1BlockCommitment) -> CsmWorkerResult<()> {
+        // No match means the divergence reaches past the finalized anchor.
+        let Some(fork_idx) = self.find_fork_index()? else {
+            return Err(CsmWorkerError::ReorgPastFinality {
+                finalized: self.recent_asm_blocks[0],
+                incoming: *incoming,
+            });
```
purusang, line 264: I think we should hard error this and crash the app.

crates/csm-worker/src/processor.rs
purusang: Also why it should only be debug assert?

crates/csm-worker/src/processor.rs
```diff
@@ -124,14 +124,50 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
     ) -> CsmWorkerResult<()> {
         let state = next_state.as_ref().clone();
         self.ctx
-            .put_client_state_update(&asm_block, ClientUpdateOutput::new(state.clone(), vec![]))?;
-        self.last_asm_block = Some(asm_block);
+            .put_client_state_update(&asm_block, ClientUpdateOutput::new_state(state.clone()))?;
```
purusang: Should we write db after the debug_assert that follows?

crates/csm-worker/src/state.rs
```diff
@@ -48,113 +53,160 @@ pub struct CsmWorkerState<C: CsmWorkerContext> {
...
-    /// All bootstrap reads (last client state, observed checkpoint refs, params)
-    /// go through `ctx`; runtime persistence and L1 fetches use the same
-    /// context after construction.
-    pub fn new(ctx: C) -> CsmWorkerResult<Self> {
-        // Load the most recent client state from storage
-        let (cur_block, cur_state) = ctx
+    /// Also eagerly updates, persists and publishes the client state.
+    pub fn bootstrap(ctx: C) -> CsmWorkerResult<Self> {
+        let (recent_l1blk, recent_clstate) = ctx
             .fetch_most_recent_client_state()?
```
purusang: On restarts, I think it may load client state based on orphaned/stale l1 blks 🤔 ?

purusang (dismissed):
Minor finds. Looks good mostly.

storopoli (commented):
I agree with @krsnapaudel on both Codex findings.

crates/csm-worker/src/state.rs
bewakes: yes, that was a possibility. This is handled in https://github.com/alpenlabs/alpen/pull/1984/commits/942043685570ec51d8f338d417986bbe11ab9974

crates/csm-worker/src/processor.rs
bewakes: yes, good catch! with `debug_assert!` the release code would have no effect on the ordering, but it's best to maintain proper order. Fixed now.

crates/csm-worker/src/processor.rs
bewakes: Right now the service silently swallows this error. I've made the service to exit on reorg-past-finality rather than panicing. This should trigger graceful shutdown of the services and also that having typed errors helps in testing.

crates/csm-worker/src/processor.rs
bewakes: No we won't have regression. This is handled by the reorg logic.

crates/csm-worker/src/processor.rs
bewakes: I've updated the message. This is debug assert because it is not as severe. Even if this somehow passes in the release build, the subsequent bootstrap handles this case.

crates/csm-worker/src/processor.rs
```diff
@@ -249,60 +285,124 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
...
+        self.confirmed_epoch = new_clstate.get_last_epoch();
+        self.finalized_epoch = new_clstate.get_declared_final_epoch();
+        self.observed_checkpoints = derived.observed_checkpoints;
+
+        // Re-persist at the fork to overwrite the orphaned branch's row, but
+        // publish only when the re-derived state actually changed.
+        self.ctx.put_client_state_update(
+            &fork_block,
+            ClientUpdateOutput::new_state(new_clstate.clone()),
+        )?;
```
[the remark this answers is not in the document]
bewakes: Fixed in https://github.com/alpenlabs/alpen/pull/1984/commits/942043685570ec51d8f338d417986bbe11ab9974

crates/csm-worker/src/processor.rs
```diff
@@ -249,60 +285,124 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
...
+        let Some(fork_idx) = self.find_fork_index()? else {
+            return Err(CsmWorkerError::ReorgPastFinality {
+                finalized: self.recent_asm_blocks[0],
+                incoming: *incoming,
+            });
+        };
+        let fork_block = self.recent_asm_blocks[fork_idx];
+
+        warn!(%fork_block, %incoming, "reorg detected; rewinding to fork point");
+        self.recent_asm_blocks.truncate(fork_idx + 1);
```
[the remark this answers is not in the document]
bewakes: Done in https://github.com/alpenlabs/alpen/pull/1984/commits/7f131cd82f39293ac7f9ffcbcbe067cf67a809c6

storopoli:
@codex review

crates/csm-worker/src/state.rs
```diff
@@ -48,113 +53,195 @@ pub struct CsmWorkerState<C: CsmWorkerContext> {
...
+/// Client state and surviving observation queue derived as of an L1 tip.
+pub(crate) struct DerivedState {
+    pub(crate) observed_checkpoints: VecDeque<L1Checkpoint>,
+    /// Client state rebuilt from the derived checkpoints.
+    pub(crate) new_clstate: ClientState,
+}
+
+/// Derives client state and newly observed checkpoints from storage as of `cur_block`.
+pub(crate) fn derive_state<C: CsmWorkerContext>(
+    ctx: &C,
```
delbonis: "Derived" is a very vague and imprecise term for this.  What does this really represent?  I can intuit it from reading the code, but that's not the point.  How do we compute new states?  Can we model it as a pure function or is it a bookkeeping structure like with services?  What do we do with these?

`new_clstate` is also a bit misleading in the current terminology since we don't have a "consensus later" anymore.

crates/csm-worker/src/state.rs
```diff
@@ -48,113 +53,195 @@ pub struct CsmWorkerState<C: CsmWorkerContext> {
...
+    /// Bootstraps a new CSM worker state from worker context.
     ///
-    /// All bootstrap reads (last client state, observed checkpoint refs, params)
-    /// go through `ctx`; runtime persistence and L1 fetches use the same
-    /// context after construction.
-    pub fn new(ctx: C) -> CsmWorkerResult<Self> {
-        // Load the most recent client state from storage
-        let (cur_block, cur_state) = ctx
+    /// Also eagerly updates, persists and publishes the client state.
+    pub fn bootstrap(ctx: C) -> CsmWorkerResult<Self> {
```
delbonis: Should be `init_from_context`.  This isn't really "bootstrapping".

crates/csm-worker/src/state.rs
```diff
@@ -48,113 +53,195 @@ pub struct CsmWorkerState<C: CsmWorkerContext> {
...
+    Err(CsmWorkerError::MissingData {
+        what: "canonical client state at or above reorg floor",
+        detail: candidate_blk.to_string(),
+    })
+}
+
+/// Builds the recent-ASM-blocks list from the reorg-safe floor (index 0) up to
+/// `tip`, the last block CSM committed client state for.
+fn init_recent_asm_blocks<C: CsmWorkerContext>(
+    ctx: &C,
```
delbonis, line 156: Should be `fetch_` or `load_`.  This name makes it look like the "ASM blocks" are the *target* of an init operation, but it's just fetching them from the database.

crates/csm-worker/src/state.rs
```diff
@@ -48,113 +53,195 @@ pub struct CsmWorkerState<C: CsmWorkerContext> {
...
+/// Deepest L1 height a reorg could reach under `tip` — the window's floor.
+///
+/// A block is reorg-safe once it is either buried `depth` deep under `tip` or at
+/// or below the last finalized checkpoint.
+pub(crate) fn reorg_floor_height<C: CsmWorkerContext>(
+    ctx: &C,
+    clstate: &ClientState,
+    tip: L1Height,
+) -> L1Height {
+    let depth = ctx.l1_reorg_safe_depth().max(1);
```
delbonis: This being prefixed with `reorg_` makes it look like it's *doing* something related to reorg operations, but it's really just doing some math.  Should be prefixed with `get_` or `compute_`.

crates/csm-worker/src/processor.rs
```diff
@@ -249,60 +284,130 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
...
+        if persisted.as_ref() != Some(&new_clstate) {
+            self.ctx
+                .publish_client_state(new_clstate.clone(), fork_block);
         }
+        self.last_committed_state = Arc::new(new_clstate);
         Ok(())
     }
 
+    /// Highest list index whose block still matches the canonical L1 chain.
+    fn find_fork_index(&self) -> CsmWorkerResult<Option<usize>> {
```
delbonis: What "list index"?  This function is unclear.

crates/csm-worker/src/processor.rs
```diff
@@ -249,60 +284,130 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
...
+        // Process the target.
+        self.process_asm_logs(asm_block, logs)?;
+        Ok(())
+    }
+
+    /// Rewinds the worker to the fork point shared with the chain leading to
+    /// `incoming`, re-deriving cursors and persisted state there.
+    ///
+    /// Errors if the fork lies at or below the finalized anchor (index 0).
+    fn reorg_to_fork(&mut self, incoming: &L1BlockCommitment) -> CsmWorkerResult<()> {
```
delbonis: This function is suspicious, and the dataflow around it is weird.  It leaves the database in a state that's not predictable from just the arguments, it implicitly does a bunch of chain bookkeeping but leaves the caller to process the chain forwards, from a state it has to figure out separately.

I think?  In any case, this is unclear.

Also the "fork point" is called the "pivot" block.

delbonis (changes requested):

crates/csm-worker/src/state.rs
```diff
@@ -48,113 +53,195 @@ pub struct CsmWorkerState<C: CsmWorkerContext> {
...
-        let current_l1_tip = cur_block.height();
-        let finality_depth = ctx.l1_reorg_safe_depth().max(1);
-        let baseline_finalized_epoch = cur_state.get_declared_final_epoch();
-        let observation_start_epoch = baseline_finalized_epoch
-            .map(|epoch| epoch.epoch().saturating_add(1))
-            .unwrap_or(0);
+        // The stored tip may be an orphan from a reorg the node missed while
+        // down; resolve down to the latest block still on the canonical chain.
+        let (recent_l1blk, recent_clstate) =
+            resolve_canonical_tip(&ctx, loaded_block, loaded_clstate)?;
```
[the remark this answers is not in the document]
krsnapaudel: Seems Codex is correct here on the finding and the suggestion.

crates/csm-worker/src/processor.rs
```diff
@@ -249,60 +284,130 @@ impl<C: CsmWorkerContext> CsmWorkerState<C> {
...
-            warn!(
-                %asm_block,
-                last_blkid = ?last.blkid(),
-                "same-height ASM block with different blkid; processing as reorged tip"
-            );
-            // Directly process the logs in one block reorg.
-            return self.process_asm_logs(asm_block, logs);
+        let is_pure_extension = last_still_canonical && asm_block.height() > last.height();
+        if !is_pure_extension {
+            self.reorg_to_fork(&asm_block)?;
```
krsnapaudel: Claude seems to have found an edge case.

**Duplicate window entry on a drop-tip (shorter) reorg**

`process_asm_block` re-commits the fork block when a reorg simply drops the tip and the new canonical tip is a block already in `recent_asm_blocks`.

Scenario: window `[40323, 40324, 40325]`, then 40325 is orphaned with nothing above 40324. ASM re-reports `cur_block = 40324`. Since canonical heights are unique, the incoming `(40324, blkid)` is byte-identical to the existing `good_24` entry, so `find_fork_index` matches at 40324 and `fork_block == incoming`.

`reorg_to_fork` already rewinds the window onto `good_24` (and deletes the orphan row) — the window is correct and `tip == incoming` at that point. But there's no guard afterward, so the empty gap loop is skipped and `process_asm_logs(incoming)` runs anyway; `commit_block` sees `same_height_reorg` and pushes `good_24` a second time:

```
recent_asm_blocks = [40323, 40324, 40324]   // good_24 duplicated
```

State stays correct (the observation re-apply is deduped by epoch), so it's not corrupting — but it leaves a duplicate in the window plus a redundant re-persist and re-publish of that block.

Reachable in production: Bitcoin losing its tip without immediate replacement is exactly the shorter-reorg case this PR targets.

The existing `shorter_chain_reorg_rewinds_to_fork` test doesn't catch it — it only asserts `last() == incoming` and `!contains(orphan_25)`, never window length. Adding `assert_eq!(state.recent_asm_blocks, vec![anchor, good_24])` fails with `[40323, 40324, 40324]`.

Suggested fix: after `reorg_to_fork`, short-circuit when the rewind already landed on `incoming` (e.g. `if asm_block == *recent_asm_blocks.last().unwrap() { return Ok(()) }`, or guard on `asm_block.height() <= resume_height`) before the replay/commit, and add a window-length assertion to the test.

krsnapaudel (commented):
Mostly correct. With Claude Code's help, identified one issue and verified one Codex finding.

crates/csm-worker/src/state.rs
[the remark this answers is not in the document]
bewakes: This is resolved in https://github.com/alpenlabs/alpen/pull/1984/commits/45ee50f39c9e2e60c83b7e1f3e42e030b2c93d06

crates/csm-worker/src/service.rs
```diff
@@ -98,10 +103,9 @@ fn refresh_finalized_checkpoint<C: CsmWorkerContext>(
 ) -> CsmWorkerResult<()> {
     let last_seen = state.last_committed_state.get_last_checkpoint();
     let refreshed = ClientState::new(Some(finalized), last_seen);
-    state.ctx.put_client_state_update(
-        &asm_block,
-        ClientUpdateOutput::new(refreshed.clone(), vec![]),
-    )?;
+    state
+        .ctx
+        .put_client_state_update(&asm_block, ClientUpdateOutput::new_state(refreshed.clone()))?;
```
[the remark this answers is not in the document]
bewakes: Resolved in the later changes.

crates/csm-worker/src/state.rs
bewakes: `DerivedState` is removed entirely and `derive_state` is renamed to `reload_client_state` and return a client state now.

crates/csm-worker/src/processor.rs
bewakes: Renamed and updated the return types.

bewakes:
@krsnapaudel , @delbonis thank you for your suggestions and comments. PTAL again, they have been resolved.

delbonis (approved):
looks good to move forwards with

storopoli (approved):
ACK a73a861148a78c338b7b0c8eb53f839ee9b0b96a

krsnapaudel (approved):
The `shorter_chain_reorg_rewinds_to_fork` test still only asserts `last() == incoming` and `!contains(orphan_25)` — not window length, but not blocking.

PR #1987, opened 2026-06-13
STR-3649: EE chunk witness hardness

evgenyzdanovich:
## Description

- deep EE chunk witness restructure -- per block partial MPT
- get rid of in-band range witness extractor for EE chunk proofs
- make sure that chunk witness capture happens per-block during block processing at depth-0 -- **using already calculated by reth witness** (as part of the block execution it is already there, we just didn't use it).

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist


crates/ee-chunk-runtime/src/chunk_processing.rs
```diff
@@ -91,39 +91,68 @@ impl<'c> IoTracker<'c> {
     }
 }
 
-/// Processes a chunk's blocks and updates the state, checking the IO against an
-/// expected IO trace.
+/// Processes a chunk's blocks and verifies each block transition, checking the
+/// IO against an expected IO trace.
+///
+/// Each block carries its own depth-0 partial-state witness (`block_states[i]`,
```
evgenyzdanovich: here is the crux of the change related to witness restructuring
@irnb @sapinb

bin/alpen-client/src/prover/spec_chunk.rs
```diff
@@ -154,100 +160,69 @@ impl ProofSpec for ChunkSpec {
...
-                PaasError::PermanentFailure(format!("decode block RLP for {chunk_id:?}: {e}"))
+        for (idx, block_hash) in block_hashes.iter().enumerate() {
+            // Per-block witness record: depth-0 witness + RLP block + parent
+            // header. Missing means production-time capture hasn't landed yet
+            // (or the record was deleted) — a transient failure so paas retries.
+            // Once the paas-retries `resolve_input` API lands, "not produced
+            // yet" should map to `Blocked` rather than a transient failure.
+            let bytes = self
+                .storage
+                .get_block_witness(*block_hash)
```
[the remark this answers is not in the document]
storopoli: We don't care about DB schema compat

crates/alpen-ee/sequencer/src/block_builder/task.rs
```diff
@@ -224,6 +232,32 @@ async fn block_builder_task_inner<TEngine: PayloadBuilderEngine>(
...
+
+    // Block acceptance blocks on witness production: now that the block is
+    // canonical at tip, synchronously compute + persist its depth-0 proof
+    // witness before the block is saved/advanced. A failure fails the block —
+    // the builder loop retries — so a block is never accepted without its
+    // witness, and the witness's multiproofs are always shallow.
+    witness_producer
+        .produce_block_witness(blockhash)
+        .await
+        .context("block_builder: produce block witness")?;
```
storopoli: This creates a partial-commit window: `update_consensus_state` canonicalizes the block in Reth before witness production, while `save_exec_block` / `ExecChainHandle::new_block` happen later. If `produce_block_witness` fails after the FCU succeeds, Reth has advanced its canonical head but Alpen storage has not accepted the block, and the retry loop continues from the old `next_block_target`. Please either avoid FCU-before-witness by capturing from the freshly built block/payload plus parent state/header, or add an explicit recovery/rollback path with a regression test for witness failure after successful FCU.

crates/ee-chunk-runtime/src/chunk_processing.rs
```diff
@@ -91,39 +91,66 @@ impl<'c> IoTracker<'c> {
...
-    state: &mut E::PartialState,
+    block_states: &mut [E::PartialState],
     chunk: &Chunk<'_, E>,
     verified_tip: Hash,
+    prev_state_root: Hash,
     expected_inputs: &ExecInputs,
     expected_outputs: &ExecOutputs,
 ) -> EnvResult<()> {
-    // 1. Check that the chunk is nonempty.
+    // 1. Check that the chunk is nonempty and the per-block witness list lines up with the blocks.
```
storopoli: "lines up" is too strong of a comment. This only checks that they are the same length.
For me "lines up" mean that they have the same things. I don't know it is ambiguous non-technical term anyways.

storopoli (changes requested):

crates/alpen-ee/sequencer/src/block_builder/task.rs
evgenyzdanovich: i think it's addressed in [f970928](https://github.com/alpenlabs/alpen/pull/1987/commits/f9709286802ff38b6b5fddb03ccc7e5eb49d5104)

crates/reth/node/src/block_witness.rs
```diff
@@ -0,0 +1,140 @@
...
+/// **no re-execution**.
+///
+/// Reads the witness directly out of the `executed_state` produced while the
+/// block was built — the live reth [`State`] right after
+/// `BlockBuilder::finish`. Reusing that state is the whole point of inline
+/// capture: the single production execution both commits state and yields its
+/// access set, so no second execution is needed.
+///
+/// `executed_state` supplies the access set (touched accounts/slots, loaded
+/// `codes`, BLOCKHASH range) via reth's [`ExecutionWitnessRecord`].
```
evgenyzdanovich, line 69: this is what we use to capture the "per-block-input" parts (for the chunk proof).
I think it's structurally the same thing and `reth`'s mechanism you mentioned during the sync @prajwolrg

crates/reth/node/src/payload_builder.rs
```diff
@@ -298,6 +304,27 @@ where
         ..
     } = builder.finish(&state_provider)?;
 
+    // Inline depth-0 proof-witness capture. The block was just executed into
```
evgenyzdanovich, line 307: FYI @storopoli -- the per block witness is derived here now

storopoli (approved):
ACK 22af86e76c14c9ddedd644e8f3e40c3d2d356d66

crates/ee-chunk-runtime/src/chunk_processing.rs
```diff
@@ -91,39 +91,66 @@ impl<'c> IoTracker<'c> {
...
+        }
+
+        // Verify the block itself against its own witness state.
         process_block(ee, state, cb)?;
 
+        // Post-root: after execution the witness must hash to this block's own
+        // header state root.
+        if state.compute_state_root()? != header.get_state_root() {
+            return Err(EnvError::MismatchedChainSegment);
+        }
```
delbonis: I intentionally didn't include per-block state root checks since computing the state root is very expensive in the proof (lots of hashing), although maybe we should benchmark/profile this to be sure.  We would just want to do it at the beginning and end of the chunk since that still asserts consistency, making sure not to do it extra times.

We could include this behind a `per-block-state-checks` feature flag if it's desired by downstream users.

crates/ee-chunk-runtime/src/private_inputs.rs
```diff
@@ -122,27 +113,42 @@ pub struct RawBlockData {
 
     /// SSZed `ExecOutputs`.
     exec_outputs_ssz: Vec<u8>,
+
+    /// Codec-encoded per-block `E::PartialState` — this block's depth-0
+    /// transition witness, anchored at the block's parent state root and
+    /// updatable to the block's own state root. The chunk witness is the
+    /// ordered list of these across the chunk's blocks.
+    raw_partial_pre_state: Vec<u8>,
```
delbonis: If we provide the pre-state at the chunk level and reuse it across all the blocks in the chunk then that reduces the amount of state that we need to provide.  It still ensures consistency when we do the chunk-level checks.

crates/reth/node/src/block_witness.rs
```diff
@@ -0,0 +1,140 @@
...
+        block_rlp,
+        parent_header_rlp,
+    })
+}
+
+/// Persisted per-block proof-witness, keyed by execution block hash.
+///
+/// Self-contained so the chunk prover can assemble a chunk proof input from the
+/// chunk's per-block records alone, without re-touching reth at proof time.
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize)]
```
delbonis: PLS NO MORE BORSH 😭 

If this is just a database bookkeeping structure then it's fine to just CBOR it imo.  If we want to we can zero-copy decode from the fields.

crates/ee-chunk-runtime/src/chunk_processing.rs
```diff
@@ -91,39 +91,68 @@ impl<'c> IoTracker<'c> {
...
         // Verify it builds on the previous block.
         if header.get_parent_id() != cur_verified_tip_blkid {
             return Err(EnvError::MismatchedChainSegment);
         }
 
-        // Verify the block itself.
+        // Pre-root: this block's witness must hash to the parent state root.
+        if state.compute_state_root()? != parent_state_root {
+            return Err(EnvError::MismatchedCurStateData);
+        }
```
prajwolrg: I wonder about the impact of this in the proof cycles.

crates/ee-chunk-runtime/src/chunk_processing.rs
```diff
@@ -91,39 +91,68 @@ impl<'c> IoTracker<'c> {
...
+
+        // Verify the block itself against its own witness state.
         process_block(ee, state, cb)?;
 
+        // Post-root: after execution the witness must hash to this block's own
+        // header state root.
+        if state.compute_state_root()? != header.get_state_root() {
+            return Err(EnvError::MismatchedChainSegment);
+        }
```
prajwolrg: I wonder about the impact of this in the proof cycles.

delbonis (changes requested):
Is there some reason we really want to provide per-block state and do per-block state consistency checks.  I realize that *not* doing it means that the intermediate headers can lie about state roots and confuse light clients, but this wouldn't be a security issue since we'd still assert correctness at the boundaries and when we have decentralized sequencing the blocks with bad state roots can be rejected by the PoS voting.

Maybe there's something I'm missing?

bin/alpen-client/src/main.rs
```diff
@@ -418,17 +420,19 @@ fn main() {
...
-                // Per-block accessed-state capture. Re-executes each
-                // committed block with a `CacheDBProvider` to record the
-                // read set + bytecodes, which the chunk-builder consumes
-                // at chunk-seal time to skip its own re-execution loop.
+                // Per-block accessed-state capture. The CHUNK proof's witness is
+                // now produced inline in block production (see `block_witness` /
+                // `RethBlockWitnessProducer`); this exex remains only to feed the
+                // ACCOUNT proof's batch-range witness (`RangeWitnessExtractor`
+                // reads `AccessedStateStore`). Retiring it is the separate acct
+                // migration tracked in experimental/evgeniy/ee-proper-witness.md.
```
prajwolrg: Might make sense to link a Jira ticket (or a PR) here for consistency.

prajwolrg (commented):
The changes themselves looks complete and sound. I do wonder about the implication of this in the prover's cost. We seem to be computing state root multiple times now, which I think is one of the most expensive part. I wonder if we can combine the `ExecutionWitness` for the chunk and pass it once, instead of per-block.

crates/ee-chunk-runtime/src/private_inputs.rs
evgenyzdanovich: yep, addressed, the logic in the chunk-runtime is now untouched

crates/ee-chunk-runtime/src/chunk_processing.rs
evgenyzdanovich: removed per-block state hashing

crates/ee-chunk-runtime/src/chunk_processing.rs
evgenyzdanovich: yep, addressed, the logic in the chunk-runtime is now untouched

crates/reth/node/src/block_witness.rs
evgenyzdanovich: sure, using `ciborium` for now, is this preferred?

evgenyzdanovich:
@prajwolrg @delbonis 
the changes to chunk-runtime were reverted, we prove the same stuff as before.

so the changes in the PR now are on the host side, not the runtime:
- we change the structure of how we provide the chunk's `pre_state` MPT -- we capture the execution witness **per-block** (while it is hot), and then do the assembly of the (chunk spanning) `EvmPartialState` by aggregating the per-block witnesses (into one trie that suffices RSP).
- previously we were doing chunk-spanning ETH multiproofs (over a potentially historical state) to grab that MPT (which is much less efficient, also due to how reth works)

evgenyzdanovich:
@codex review

crates/reth/node/src/block_witness.rs
delbonis: For container structures on non-protocol types where database migrations compatibility might be a concern.  The complicated stuff (inside the bufs) should try to avoid copies.

delbonis (approved):

bin/alpen-client/src/prover/spec_chunk.rs
```diff
@@ -154,100 +161,77 @@ impl ProofSpec for ChunkSpec {
...
```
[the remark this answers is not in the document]
evgenyzdanovich: backfill isnt a concern

prajwolrg (approved):

storopoli (approved):
ACK f6093feec4e9c815f85e142916ceceb21adea88a

functional-tests/tests/alpen_client/test_ee_prover_pipeline_alive.py
```diff
@@ -15,11 +15,14 @@
 
 Native mode runs the full guest program logic in plain Rust; only the
 zk proof generation itself is bypassed. So this test exercises:
-  - `ChunkSpec::fetch_input` (sled reads of the new ChunkWitnessRecord)
-  - `AccessedStateGenerator` exex (per-block accessed-state writes,
-    plus bytecode cache lookups via `BytecodeSchema`)
-  - The batch builder's seal-time witness extraction
-  - The `EeChunkProgram` guest (state transition checks, MPT validation)
+  - Inline per-block witness production (`RethBlockWitnessProducer` ->
+    `capture_block_witness` -> reth `ExecutionWitness` -> rsp
```
sapinb: `capture_block_witness` is not used anywhere in the current code. Is this comment out of date ?

evgenyzdanovich: done

crates/alpen-ee/sequencer/src/block_builder/task.rs
```diff
@@ -239,6 +241,24 @@ async fn block_builder_task_inner<TEngine: PayloadBuilderEngine>(
...
 
+    // Canonicalize the freshly built block LAST. The payload submission above
+    // only inserts the block into the engine tree as VALID; drive the
+    // forkchoice update here — instead of relying on the OL tracker's
+    // asynchronous FCU — so block acceptance is synchronous and self-contained,
+    // and the block is canonical at tip for the next build. Performing it after
+    // the storage save closes the partial-commit window: the engine's canonical
+    // head never advances past a block Alpen storage has not accepted, so a
+    // failure leaves the engine behind storage (safe, self-healing) rather than
+    // ahead. Safe/finalized are left zero to retain their current values.
```
sapinb: "partial-commit window" is based on a previous review comment and only makes sense in the context of the previous code commit (that will not exist after a squash-merge). 
Not relevant to the current state of the code and makes this confusing to read.

crates/alpen-ee/sequencer/src/block_builder/task.rs
```diff
@@ -239,6 +241,24 @@ async fn block_builder_task_inner<TEngine: PayloadBuilderEngine>(
...
+    payload_builder
+        .update_consensus_state(ForkchoiceState {
+            head_block_hash: B256::from_slice(blockhash.as_ref()),
+            safe_block_hash: B256::ZERO,
+            finalized_block_hash: B256::ZERO,
+        })
+        .await
+        .context("block_builder: canonicalize block via forkchoice update")?;
```
sapinb: This needs corresponding changes in engine control task to handle an edge case due to a race between this update and an OL tracker driven update. This will effectively cause the OL tracker's update to be lost.

The sequence of events:
last state: { head: H, confirmed: C, finalized: F }
1. block assembly -> { head: H + 1, confirmed: 0, finalized 0 } 
canonical head is updated to H + 1
 
2. OL update to engine ctrl -> { head: H (cached), confirmed: C + 1, finalized: F + 1 }
Reth, as of v1.9.1 will ignore updates that are already part of the canonical chain, so OL tracker's updated confirmed and finalized will not be applied.
Worse, as we run in `always_process_payload_attributes_on_canonical_head` mode, this will trigger a canonical chain unwind to H, losing block H + 1.
ref: https://github.com/paradigmxyz/reth/blob/3afe69a5738459a7cb5f46c598c7f541a1510f32/crates/engine/tree/src/tree/mod.rs#L1118

3. block builder triggers preconf update to engine ctrl -> { head: H + 1, confirmed: 0, finalized 0 } 
block H + 1 may or may not be recoverable here

sapinb (changes requested):
Need to handle a race between the FCU in block assembly and inside engine ctrl.

crates/alpen-ee/sequencer/src/block_builder/task.rs
evgenyzdanovich: makes sense, thank you for the finding! addressing it atm.

evgenyzdanovich: I removed the FCU, I no longer think this is actually needed.

evgenyzdanovich:
@sapinb please have another look.

delbonis (approved):

storopoli (approved):
ACK f2f8f93737473e8e3e4ffe71ba6b1a9d93cc0cd8

sapinb (approved):
ACK f2f8f93737473e8e3e4ffe71ba6b1a9d93cc0cd8

PR #1989, opened 2026-06-15
feat(alpen-ee/da): add fields in EvmHeaderSummary to reconstruct parent header

sapinb:
## Description
Add the remaining non-deterministic fields required to reconstruct the evm header of the previous block (last block of this batch). Full header of the previous block is necessary to resume block production and proving.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/bin/alpen-client/src/header_summary.rs
+++ b/bin/alpen-client/src/header_summary.rs
@@ -55,6 +55,11 @@ fn summarize_header(header: &reth_primitives::Header) -> eyre::Result<EvmHeaderS
         })?,
         gas_used: header.gas_used,
         gas_limit: header.gas_limit,
+        parent_hash: header.parent_hash.0,
+        transactions_root: header.transactions_root.0,
+        receipts_root: header.receipts_root.0,
+        logs_bloom: header.logs_bloom.0 .0,
+        withdrawals_root: header.withdrawals_root.unwrap_or_default().0,
     })
 }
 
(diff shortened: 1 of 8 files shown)


storopoli (changes requested):
NACK

Everything that you add to `EvmHeaderSummary` will be added to on-chain L1 data as in https://github.com/alpenlabs/alpen/blob/e54c01e005c44ee97cdcfaac0766176ce8ca59ce/crates/alpen-ee/da/types/src/blob.rs?plain=1#L24-L38

Hence, everything that you add here MUST be discussed and spec'ed out (or updated in the EE DA spec).

You are adding 384 bytes to every `EvmHeaderSummary` for no consensus-critical reason:

Field | Bytes
-- | --
parent_hash: [u8; 32] | 32
transactions_root: [u8; 32] | 32
receipts_root: [u8; 32] | 32
logs_bloom: [u8; 256] | 256
withdrawals_root: [u8; 32] | 32
Total added | 384

Current `EvmHeaderSummary` is 5 `u64`s = 40 bytes: 

https://github.com/alpenlabs/alpen/blob/e54c01e005c44ee97cdcfaac0766176ce8ca59ce/crates/alpen-ee/da/types/src/blob.rs?plain=1#L40-L59

This PR #1989 makes it 424 bytes more than 10x increase.

Can't we solve the 

> necessary to resume block production and proving

with databases or other things instead of 10x DA costs of `EvmHeaderSummary`?

Cc @irnb

sapinb:
@storopoli Closing this PR. But flagging that currently proof generation cannot be resumed with only the DA recovery data. So either we need to push full block header in the checkpoint or update the proof statements to not require the last block header. We will probably go for the latter.

storopoli:
Cc @irnb

PR #1993, opened 2026-06-16
STR-3238: fail closed on invalid withdrawal BOSD

storopoli:
## Description

Restricts withdrawal BOSD descriptors to a configurable maximum length, defaulting to 81 bytes including the type tag, and makes bridge-out withdrawal decoding fail closed. This makes possible to have 80-byte `OP_RETURN` withdrawal destinations.

This validates withdrawal descriptors at the EVM precompile boundary, when extracting bridge-out logs, and when OL processes bridge-gateway withdrawal messages. Malformed or oversized bridge-out logs now invalidate payload/execution processing instead of being silently dropped.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Commits are split by review area:

1. `BridgeParams` descriptor limit API/defaults.
2. Reth precompile and fail-closed withdrawal extraction.
3. OL bridge-gateway descriptor validation.
4. CLI/node config surfaces.

- [ ]  Yes
- [x]  No

## Checklist

AI assistance notice: This PR was implemented with AI assistance.

## Related Issues

STR-3238

### DIFF
--- a/crates/bridge-params/src/lib.rs
+++ b/crates/bridge-params/src/lib.rs
@@ -2,7 +2,6 @@
 
 #[cfg(feature = "arbitrary")]
 use arbitrary::Arbitrary;
-use borsh::{BorshDeserialize, BorshSerialize, io};
 use serde::{Deserialize, Serialize, de::Error as DeError};
 use ssz_derive::{Decode, Encode};
 use thiserror::Error;
@@ -31,13 +30,6 @@ struct BridgeParamsRaw {
     max_withdrawal_descriptor_len: u32,
 }
 
-/// Legacy raw mirror used to decode params encoded before descriptor limits.
-#[derive(Decode, Encode)]
-struct LegacyBridgeParamsRaw {
-    denomination: u64,
-    max_withdrawal_amount: Option<u64>,
-}
-
 impl<'de> Deserialize<'de> for BridgeParams {
     fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
         let raw = BridgeParamsRaw::deserialize(deserializer)?;
@@ -60,41 +52,13 @@ impl ssz::Decode for BridgeParams {
     }
 
     fn from_ssz_bytes(bytes: &[u8]) -> Result<Self, ssz::DecodeError> {
-        if let Ok(raw) = BridgeParamsRaw::from_ssz_bytes(bytes) {
-            return Self::new_with_descriptor_limit(
-                raw.denomination,
-                raw.max_withdrawal_amount,
-                raw.max_withdrawal_descriptor_len,
-            )
-            .map_err(|e| ssz::DecodeError::BytesInvalid(e.to_string()));
-        }
-
-        let raw = LegacyBridgeParamsRaw::from_ssz_bytes(bytes)?;
-        Self::new(raw.denomination, raw.max_withdrawal_amount)
-            .map_err(|e| ssz::DecodeError::BytesInvalid(e.to_string()))
-    }
-}
-
-impl BorshSerialize for BridgeParams {
-    fn serialize<W: io::Write>(&self, writer: &mut W) -> io::Result<()> {
-        BorshSerialize::serialize(&self.denomination, writer)?;
-        BorshSerialize::serialize(&self.max_withdrawal_amount, writer)?;
-        BorshSerialize::serialize(&self.max_withdrawal_descriptor_len, writer)
-    }
-}
-
-impl BorshDeserialize for BridgeParams {
-    fn deserialize_reader<R: io::Read>(reader: &mut R) -> io::Result<Self> {
-        let denomination = u64::deserialize_reader(reader)?;
-        let max_withdrawal_amount = Option::<u64>::deserialize_reader(reader)?;
-        let max_withdrawal_descriptor_len = u32::deserialize_reader(reader)?;
-
+        let raw = BridgeParamsRaw::from_ssz_bytes(bytes)?;
         Self::new_with_descriptor_limit(
-            denomination,
-            max_withdrawal_amount,
-            max_withdrawal_descriptor_len,
+            raw.denomination,
+            raw.max_withdrawal_amount,
+            raw.max_withdrawal_descriptor_len,
         )
-        .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e.to_string()))
+        .map_err(|e| ssz::DecodeError::BytesInvalid(e.to_string()))
     }
 }
 
@@ -341,19 +305,5 @@ mod tests {
         let encoded = ssz::Encode::as_ssz_bytes(&bp);
         let decoded = <BridgeParams as ssz::Decode>::from_ssz_bytes(&encoded).unwrap();
         assert_eq!(bp, decoded);
-    }
-
-    #[test]
-    fn ssz_legacy_decode_uses_default_descriptor_len() {
-        let legacy = LegacyBridgeParamsRaw {
-            denomination: 100_000_000,
-            max_withdrawal_amount: Some(1_000_000_000),
-        };
-        let encoded = ssz::Encode::as_ssz_bytes(&legacy);
-        let decoded = <BridgeParams as ssz::Decode>::from_ssz_bytes(&encoded).unwrap();
-        assert_eq!(
-            decoded.max_withdrawal_descriptor_len(),
-            DEFAULT_MAX_WITHDRAWAL_DESCRIPTOR_LEN
-        );
     }
 }
--- a/crates/bridge-params/src/lib.rs
+++ b/crates/bridge-params/src/lib.rs
@@ -2,6 +2,7 @@
 
 #[cfg(feature = "arbitrary")]
 use arbitrary::Arbitrary;
+use borsh::{BorshDeserialize, BorshSerialize, io};
 use serde::{Deserialize, Serialize, de::Error as DeError};
 use ssz_derive::{Decode, Encode};
 use thiserror::Error;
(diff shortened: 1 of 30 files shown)


crates/proof-impl/evm-ee-stf/src/executor.rs
```diff
@@ -31,6 +34,7 @@ pub fn process_block(mut input: EthClientExecutorInput) -> Result<EvmBlockStfOut
     let chain_spec: Arc<ChainSpec> = Arc::new((&input.genesis).try_into().unwrap());
     let evm_config =
         EthEvmConfig::new_with_evm_factory(chain_spec.clone(), AlpenEvmFactory::default());
+    let bridge_params = *evm_config.evm_factory().bridge_params();
```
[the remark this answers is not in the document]
storopoli: Ok good catch, done in 8714b8e6fc8715e41f843e8f009bd5153959fa09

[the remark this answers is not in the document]
storopoli: @codex review

crates/proof-impl/evm-ee-stf/src/executor.rs
```diff
@@ -27,10 +28,15 @@ pub const ACCRUE_LOG_BLOOM: &str = "accrue logs bloom";
...
+pub fn process_block(
+    mut input: EthClientExecutorInput,
+    bridge_params: BridgeParams,
+) -> Result<EvmBlockStfOutput, ClientError> {
     let chain_spec: Arc<ChainSpec> = Arc::new((&input.genesis).try_into().unwrap());
-    let evm_config =
-        EthEvmConfig::new_with_evm_factory(chain_spec.clone(), AlpenEvmFactory::default());
+    let evm_config = EthEvmConfig::new_with_evm_factory(
+        chain_spec.clone(),
+        AlpenEvmFactory::from_bridge_params(&bridge_params),
```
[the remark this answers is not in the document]
storopoli: Fixed in 45642fd61421700f70f83a623fbe39f986d77317. The EVM EE proof public output now commits the BridgeParams alongside the exec segments, with a regression test using a non-default descriptor limit.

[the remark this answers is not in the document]
storopoli: @codex review

crates/evm-ee/src/execution.rs
delbonis, line 1: I would think that importing this here now violates separation of concerns.  Maybe this is fine?  We do need to expose the limit somewhere and the fields in the struct are all used here, not sure what to think.

crates/ol/stf/src/account_processing.rs
```diff
@@ -7,6 +7,7 @@ use strata_ledger_types::*;
 use strata_msg_fmt::MsgRef;
 use strata_ol_chain_types_new::SimpleWithdrawalIntentLogData;
 use strata_ol_msg_types::OLMessageExt;
+use strata_primitives::bitcoin_bosd::Descriptor;
```
delbonis: Can we import the bosd crate directly instead of adding the dep on strata-primitives?  I want to rework that crate to be less of a key dep.

crates/reth/evm/src/utils.rs
```diff
@@ -27,47 +28,97 @@ pub(crate) fn wei_to_sats(wei: U256) -> (U256, U256) {
...
+                )
+            }
+            Self::InvalidDescriptorLength { txid, len, max } => write!(
+                f,
+                "withdrawal descriptor length {len} exceeds maximum {max} for tx {txid:?}"
+            ),
+            Self::DescriptorDecode { txid } => {
+                write!(f, "failed to decode withdrawal descriptor for tx {txid:?}")
+            }
+        }
```
delbonis: Surely we can just use thiserror here?

crates/bridge-params/src/lib.rs
```diff
@@ -2,34 +2,51 @@
...
+    #[serde(default = "default_max_withdrawal_descriptor_len")]
+    max_withdrawal_descriptor_len: u32,
+}
+
+/// Legacy raw mirror used to decode params encoded before descriptor limits.
+#[derive(Decode, Encode)]
+struct LegacyBridgeParamsRaw {
+    denomination: u64,
+    max_withdrawal_amount: Option<u64>,
 }
```
delbonis: This (and how it's used) is kinda hacky.  I see what's going on with it, but do we really need it here?  We control the whole scenario.

I wonder if SSZ even makes sense to use here.  I'd hope to get rid of borsh.  There's a bit of smell here where we're adding manual deserialization everywhere.

delbonis (changes requested):
Couple of design questions.

crates/ol/stf/src/account_processing.rs
storopoli: Addressed in ee0d243: OL STF now depends on `bitcoin-bosd` directly for `Descriptor` and drops the `strata-primitives` dependency.

crates/evm-ee/src/execution.rs
storopoli: Addressed in ee0d243: `EvmExecutionEnvironment` no longer stores `BridgeParams` or imports `strata-bridge-params` directly. Withdrawal extraction now reads the configured params from the `AlpenEvmFactory` already held by `EthEvmConfig`.

crates/reth/evm/src/utils.rs
storopoli: Addressed in ee0d243: `WithdrawalIntentExtractionError` now derives `thiserror::Error` instead of carrying manual `Display` and `Error` impls.

crates/bridge-params/src/lib.rs
storopoli: Addressed in ee0d243: removed the `BridgeParams` `Borsh` impls and the legacy SSZ fallback. The proof public output now uses a proof-local `BridgeParamsStatement` so `BridgeParams` stays SSZ/config focused.

delbonis (approved):
Ok I think this is good to move forwards with.

crates/proof-impl/evm-ee-stf/src/primitives.rs
```diff
@@ -27,14 +28,84 @@ impl ExecSegment {
     }
 }
 
-/// Public Parameters that proof asserts
-pub type EvmEeProofOutput = Vec<ExecSegment>;
+/// Bridge parameters committed by the EVM EE STF proof public statement.
+#[derive(
+    Clone, Copy, Debug, Eq, PartialEq, BorshSerialize, BorshDeserialize, Serialize, Deserialize,
+)]
+pub struct BridgeParamsStatement {
```
krsnapaudel, line 37: Should we say somewhere:
"must mirror policy-relevant `BridgeParams` fields"?

krsnapaudel (commented):
Claude's finding.

### Issue 1 — Production proof reads bridge params privately but never commits them
 Is this accurate?
 
`crates/proof-impl/alpen-chunk/src/lib.rs#L33-L41`

```rust
let withdrawal_ssz = zkvm.read_buf();
let bridge_params = BridgeParams::from_ssz_bytes(&withdrawal_ssz)   // ← private zkVM input
    .expect("failed to deserialize withdrawal params");
let evm_factory = AlpenEvmFactory::from_bridge_params(&bridge_params);
let ee = EvmExecutionEnvironment::new(chain_spec, evm_factory);

strata_ee_chunk_runtime::verify_input(&ee, input).expect("chunk verification failed");

zkvm.commit_buf(input.chunk_transition_ssz());   // ← public output: ONLY the chunk transition, params NOT committed
```

This is the production EE proof. `bridge_params` (including `max_withdrawal_descriptor_len`) configures the factory but is never committed to the public output — so a verifier can't pin which policy the prover used. Contrast with the non-production path the PR actually hardened, `crates/proof-impl/evm-ee-stf/src/primitives.rs#L65-L78`, where `EvmEeProofOutput` does commit `BridgeParamsStatement`. That's the gap: the fix landed on the wrong path.

### Issue 2 — Descriptor-length limit is node-config, not consensus-pinned
This seems less of a concern as long as we use the same env var.

Both enforcement points compare against `max_withdrawal_descriptor_len` read from the node's configured `BridgeParams`:

- **Precompile** (write-time): [crates/reth/evm/src/precompiles/bridge.rs:114-120](crates/reth/evm/src/precompiles/bridge.rs#L114-L120), value threaded from [evm.rs:74](crates/reth/evm/src/evm.rs#L74) → [factory.rs:29](crates/reth/evm/src/precompiles/factory.rs#L29) → [bridge.rs:41](crates/reth/evm/src/precompiles/bridge.rs#L41).
- **Extraction** (read-time, fail-closed): [crates/reth/evm/src/utils.rs:86-92](crates/reth/evm/src/utils.rs#L86-L92).

Default is `DEFAULT_MAX_WITHDRAWAL_DESCRIPTOR_LEN` ([evm.rs:40](crates/reth/evm/src/evm.rs#L40),[51](crates/reth/evm/src/evm.rs#L51)). Nothing pins this on-chain, so nodes with different configured limits can disagree on a block's validity — lowering it is effectively consensus-affecting and should be documented as monotonic / governance-controlled.

prajwolrg (approved):
The changes themselves looks good, but they aren't safe as raised by Dilli. This is a known issue and is tracked [here](https://alpenlabs.atlassian.net/browse/STR-1907).

crates/proof-impl/evm-ee-stf/src/primitives.rs
storopoli: Done thanks for the suggestion.

krsnapaudel (approved):
Approving since one of the issues raised is tracked by [STR-1907](https://alpenlabs.atlassian.net/browse/STR-1907).

crates/reth/evm/src/evm.rs
```diff
@@ -28,13 +28,17 @@ use crate::{
 pub struct AlpenEvmFactory {
     denomination_wei: U256,
     max_withdrawal_wei: Option<U256>,
+    max_withdrawal_descriptor_len: u32,
+    bridge_params: BridgeParams,
 }
```
purusang, line 24: Can we avoid storing both BridgeParams and separately cached policy fields? `denomination_wei` /`max_withdrawal_wei` are derived from BridgeParams, and `max_withdrawal_descriptor_len` is directly duplicated.

purusang (changes requested):
Looks mostly good.

crates/reth/evm/src/evm.rs
storopoli: Done in 5294b033c9b4f407c81aa9162035ddce53114857. `AlpenEvmFactory` now stores only `BridgeParams`; the precompile map captures that single policy object, and amount/BOSD validation derives from it at the validation boundary.

purusang (approved):

bewakes (approved):

purusang (approved):

storopoli:
@purusang this definitely need a backport. I'm opening one soon as this is merged. This will generate a configuration drift.

PR #1997, opened 2026-06-18
feat(dbtool): inspect EE DA blobs and test publication parity

storopoli:
## Description

Implements STR-3187 by adding a read-only `strata-dbtool ee-da-inspect` command for locally inspecting EE DA blobs, validating contiguous DA replay, and reporting the reconstructed EVM post-state root.

Also extends the EE DA functional test helpers to expose raw reassembled blob bytes and adds a functional test that compares L1-reassembled bytes against producer-local bytes, then verifies DA replay reconstructs the batch block state root.

### Type of Change

- [x] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The PR is split into independent commits by concern:

- dbtool command and replay logic
- raw EE DA blob reassembly helper
- publication parity/reconstruction functional test
- EE DA test boundary documentation

The functional test stops the sequencer before invoking `strata-dbtool` because sled takes an exclusive lock on the EE database.

- [ ]  Yes
- [x]  No

## Checklist

AI assistance notice: This PR was prepared with AI assistance.

## Related Issues

STR-3187


bin/strata-dbtool/src/cmd/ee_da.rs
```diff
@@ -0,0 +1,296 @@
...
+            .internal_error("Failed to apply EE DA state diff")?;
+    }
+    Ok(format!(
+        "0x{}",
+        hex::encode(reconstructor.state_root().as_slice())
+    ))
+}
+
+/// Wraps an operator-facing validation failure as a displayable user error.
+fn user_error(message: impl Into<String>, source: impl fmt::Debug + 'static) -> DisplayedError {
```
krsnapaudel: Is information repeated between message and source? It seems it is.

bin/strata-dbtool/src/cmd/ee_da.rs
```diff
@@ -0,0 +1,296 @@
...
+    dbs: &EeDatabases,
+    args: EeDaInspectArgs,
+) -> Result<(), DisplayedError> {
+    let envelopes = load_decoded_envelopes(dbs)?;
+    let info = inspect_envelopes(&envelopes, &args.chain, args.target_last_block)?;
+    output(&info, args.output_format)
+}
+
+/// Loads and validates every chunked-envelope DA entry from the EE database.
+fn load_decoded_envelopes(dbs: &EeDatabases) -> Result<Vec<DecodedEnvelope>, DisplayedError> {
```
krsnapaudel: Do we need to load the entire DA history? Should this be restricted to target?

bin/strata-dbtool/src/db.rs
```diff
@@ -45,3 +48,16 @@ pub(crate) fn open_ee_database(datadir: &Path) -> Result<Arc<EeProverDbSled>, Di
 
     Ok(prover_db)
 }
+
+/// Opens all EE database trees at `<datadir>/sled`.
+pub(crate) fn open_ee_databases(datadir: &Path) -> Result<EeDatabases, DisplayedError> {
```
krsnapaudel: `open_ee_databases` routes through `init_db_storage`, which opens all six EE DB types (node, witness, broadcast, chunked-envelope, prover, da_context) when this path needs only chunked-envelope. I see that `open_ee_database` only open prover store. 
So rename that to `open_ee_prover_database`
and add `open_ee_chunked_envelope_database`?

krsnapaudel (changes requested):
Some comments inline. The README for `strata-dbtool` also needs an update.

bin/strata-dbtool/src/cmd/ee_da.rs
storopoli: Good catch — it was. Removed the `user_error` helper and inlined `DisplayedError::UserError` at each call site, following the convention used elsewhere in the dbtool (`checkpoint.rs`, `ol_state.rs`): the message carries the context and the boxed source carries the distinct offending value (or the underlying error), with no duplication. The decode-failure case now reports the actual `reassemble_da_blob` error as the source instead of repeating `idx`.

Fixed in 4b8e7c832.

bin/strata-dbtool/src/cmd/ee_da.rs
storopoli: Restricted it to the target. `load_decoded_envelopes` now takes `target_last_block` and stops at the first envelope whose `last_block_num` covers it, instead of decoding the whole DA history. EE DA envelopes are append-only in DA-update order, so that first covering envelope is the minimal covering blob and `[0..=that]` is the full replay prefix.

Validation (magic/version/decode) and the continuity checks therefore apply only to that bounded prefix. Reads are batched, so the final batch may fetch a few rows past the target into memory, but at most one batch (rather than the whole history) is ever over-read, and nothing past the first covering envelope is decoded/validated/processed. I documented the semantics on the function and added a `load_stops_at_first_covering_envelope` test (with an in-memory DB whose poisoned trailing row would fail validation if it were ever processed) to lock this in.

Fixed in 4b8e7c832.

bin/strata-dbtool/src/db.rs
storopoli: Done as suggested. Renamed `open_ee_database` → `open_ee_prover_database`, added `open_ee_chunked_envelope_database` that opens only the chunked-envelope tree, and removed `open_ee_databases` (which routed through `init_db_storage` and opened all six EE DBs). `ee-da-inspect` now reads through the `L1ChunkedEnvelopeDatabase` trait directly, which also let me drop the `threadpool` dependency from the dbtool.

Fixed in 4b8e7c832.

storopoli:
About the README update in 4b8e7c832: added an `ee-da-inspect` section to the `strata-dbtool` README (purpose, options, example, output fields, and the read-only/exclusive-sled-lock note).

delbonis (approved):
Looks good

bin/strata-dbtool/src/db.rs
```diff
@@ -26,7 +29,9 @@ pub(crate) fn open_database(path: &Path) -> Result<Arc<SledBackend>, DisplayedEr
 /// the dbtool has no use for the other EE DBs (witness, broadcast,
 /// chunked-envelope, DA context), and skipping them keeps the cold-start
```
krsnapaudel: nit: docstring needs update

krsnapaudel (approved):

bin/strata-dbtool/src/db.rs
storopoli: What do you think now?

bin/strata-dbtool/src/db.rs
```diff
@@ -23,10 +26,12 @@ pub(crate) fn open_database(path: &Path) -> Result<Arc<SledBackend>, DisplayedEr
 /// `datadir` is expected to be the alpen-client's `--datadir`. Mirrors
 /// the alpen-client's [`alpen_ee_database::init_db_storage`] opener but
 /// only constructs the prover-task / chunk-receipt / acct-proof trees —
```
krsnapaudel: This is not true right? The mirroring [`alpen_ee_database::init_db_storage`] opener part

bin/strata-dbtool/src/db.rs
krsnapaudel: Mostly fine. Check my message about mirroring.

bin/strata-dbtool/src/db.rs
storopoli: Fixed in 2284d3d71. I removed the inaccurate “mirrors init_db_storage” wording and clarified that these helpers open the same EE sled directory while constructing only the low-level DB wrapper each dbtool path needs.

functional-tests/tests/alpen_client/ee_da/test_publication_parity_reconstruction.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+
+        target_block_num = max(tx_blocks.values())
+
+        all_envs: list[DaEnvelope] = []
+        target_l1_blob = None
+        mine_address = btc_rpc.proxy.getnewaddress()
+
+        for attempt in range(20):
+            time.sleep(5)
```
bewakes: What are these bare `time.sleep()`s?

bewakes: Also this attempting is what `wait_until` abstracts. Please consider using it here if feasible.

bewakes (changes requested):

functional-tests/tests/alpen_client/ee_da/test_publication_parity_reconstruction.py
storopoli: Fair enough. My oversight. Fixed in 73cbede9333d6cbd5e8fdd0a4e66fb605ef64395

storopoli:
Rebased. Can we get eyes on this. It's for damn 3 weeks

@prajwolrg @bewakes

irnb (approved):

bewakes (approved):

prajwolrg (approved):

PR #2000, opened 2026-06-19
STR 3788 strata-dbtool revert-ol-state should revert MMRs

krsnapaudel:
## Description

Ensure `revert-ol-state` keeps the MMR index DB consistent with the reverted OL state.

### What Changed
- Added MMR namespace enumeration in the DB layer.
- Added dbtool MMR inspection commands:
  - `get-mmr-summary`
  - `get-mmr-leaf`
- Consolidated dbtool MMR namespace handling:
  - canonical display IDs and CLI parsing
  - owner/account metadata
  - raw `MmrId` conversion
  - sentinel handling
  - typed preimage decode
- Added MMR index rollback planning/execution to `revert-ol-state`.
  - Dry run prints the MMR index namespaces that would be popped.
  - Forced execution pops OL-owned MMR index namespaces back to the target OL state's counts and peaks.
  - ASM-owned namespaces are skipped.
  - Behind-target namespaces, prefix mismatches, and L1 sentinel removal abort before mutation.
- Added functional coverage proving `revert-ol-state` rolls back MMR index state.
  - Advances both `l1-block-refs` and `snark-msg-inbox:<account>` past the target.
  - Runs an executed revert.
  - Verifies MMR leaf counts return to the target state.
  - Restarts the sequencer and verifies progress past the old tip.
- Refactored OL cleanup boundaries:
  - OL block data/status helpers live in `ol.rs`.
  - OL state row cleanup stays in `ol_state.rs`.
  - Checkpoint row deletion stays in `checkpoint.rs`.
- Hardened `delete-ol-block` so it refuses applied OL blocks with OL state instead of treating them as orphan-only block deletion.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-3788](https://alpenlabs.atlassian.net/browse/STR-3788)

[STR-3788]: https://alpenlabs.atlassian.net/browse/STR-3788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


prajwolrg:
reverting the MMRs/pruning is now also in https://github.com/alpenlabs/strata-common/pull/117.  After https://github.com/alpenlabs/alpen/pull/2001 we should be able to use some parts of it as well.

bin/strata-dbtool/src/cmd/mmr.rs
delbonis, line 1: I see `SledBackend` is used here explicitly, can this be replaced with `DatabaseBackend` to be generic?

Also, docs before attrs.

bin/strata-dbtool/src/cmd/mmr.rs
```diff
@@ -0,0 +1,1515 @@
...
+    output::{
+        mmr::{MmrLeafInfo, MmrOwner, MmrPreimageDecoded, MmrSummaryEntry, MmrSummaryInfo},
+        output,
+    },
+};
+
+const MMR_ID_ASM: &str = "asm";
+const MMR_ID_L1_BLOCK_REFS: &str = "l1-block-refs";
+const MMR_ID_SNARK_MSG_INBOX_PREFIX: &str = "snark-msg-inbox";
+const MMR_ID_SNARK_MSG_INBOX_PREFIX_WITH_SEPARATOR: &str = "snark-msg-inbox:";
```
delbonis: I would imagine these could be defined in a central location elsewhere and imported from there.

delbonis (changes requested):
Two main takeaways:

* I'm seeing a lot of replicated conditional logic for the different kinds of MMRs.  Maybe this could be reworked so that the high-level control flow is all the same but make it generic over a local trait to abstract over the low-level bookkeeping.
* This is the type of bookkeeping that the client should be doing on startup.  It's not bad to have a CLI tool to do it, but I would expect that the indexing service(s) could see that the MMRs are advanced beyond where they "should" be (ie more leafs and more indexing write batches) based on the current chain state and discard them (with a warning) before proceeding.  This should probably be a separate PR.

bin/strata-dbtool/src/cmd/mmr.rs
delbonis: Wait maybe not.

Also instead of using a version with a separator use `.split_once(":")`.

krsnapaudel:
> * This is the type of bookkeeping that the client should be doing on startup.  It's not bad to have a CLI tool to do it, but I would expect that the indexing service(s) could see that the MMRs are advanced beyond where they "should" be (ie more leafs and more indexing write batches) based on the current chain state and discard them (with a warning) before proceeding.  This should probably be a separate PR.

@delbonis Thanks. Created [this ticket](https://alpenlabs.atlassian.net/browse/STR-3846) for the follow up work.

storopoli (approved):
ACK 417c78483256de24bb9890b409b4ca1c60da3b4d

delbonis (approved):
Functionally this is fine.  There's some sloppyness things I'm a little dissatisfied with that I assume are Codex-isms.

Approving.

PR #2002, opened 2026-06-22
Commit bridge params in OL state

purusang:
## Description

Commit bridge parameters into `OLState` and read them from state at withdrawal validation time. This removes the need to thread bridge params through block assembly and checkpoint proving paths, while preserving validation at the `BridgeParams` construction/decode boundary.

This also removes the checkpoint proving soundness issue where externally supplied bridge params could diverge from the committed OL state being proven.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Bridge params are now part of the committed OL state. Withdrawal amount checks in both block assembly and checkpoint proving/replay read params from state instead of receiving them through call chains. In the checkpoint proving path, the guest now validates withdrawals against the params committed in the starting state, not against separately threaded host input.

`BridgeParams` remains a validated handwritten domain type rather than a generated raw SSZ struct. The SSZ layout is two `u64` fields, with `u64::MAX` documented as the uncapped-withdrawal sentinel.

- [ ]  Yes
- [x]  No

## Checklist

AI assistance notice: I used AI assistance to research the change, implement the refactor, and run validation.

## Related Issues

STR-3617


purusang:
@codex review

prajwolrg (commented):
Heads-up that this changes the deployment-params contract, and the genparams tooling doesn't yet account for it — flagging so it isn't a deployment-time surprise.

This PR commits `bridge_params` into `OLState` (`from_genesis_params` now stores `bridge_params: *params.bridge_params()`, and `test_bridge_params_are_committed_in_state_root` confirms the denomination changes the genesis state root, hence `genesis_ol_blkid`). Good change for proving soundness. But it makes the **denomination genesis-affecting**, and the params-generation flow currently assumes the opposite.

Concretely, in `.github/params/generate-params.sh`:
- `gen-asm-params --ol-params ol-params-raw.json` computes `genesis_ol_blkid` from the datatool **default** denomination (`gen-ol-params` has no denomination flag).
- `params-helper.py merge_ol_params` keeps the **template's** `bridge_params`, so the *deployed* `ol-params` denomination comes from the per-env template.

Pre-this-PR those can differ harmlessly (denomination isn't in genesis state — verified: regenerating staging-v2 at 1 BTC vs 2 BTC yields an identical `genesis_ol_blkid` `ca764ac4…`). **After this PR they must match**, or any env whose template denomination ≠ the datatool default (staging-v2 = 2 BTC, prod = 2 BTC) gets an `asm-params.genesis_ol_blkid` computed at the wrong denomination, and the node recomputes a different genesis at startup → bootstrap failure.

So this PR (or an immediate follow-up) should update the genparams flow so the denomination used for the `genesis_ol_blkid` computation is the **template** denomination — e.g. align the raw `ol-params` `bridge_params` to the template before `gen-asm-params` and pass the same value to `--deposit-sats`, or give `gen-ol-params` a `--deposit-sats` flag so the raw is correct directly. I had a draft of the former in #2003, which I'm closing as premature for the current build; it's the piece to revive alongside this. The per-env templates (`.github/params/templates/*/ol-params.json` + `asm-params.json`) also need their denomination set to the intended value, since that's now what genesis commits to.

One more downstream note: deployments that change denomination by editing the values YAML (staging-v2 just did, 1→2 BTC, relying on genesis-independence) will, after this PR, require a genesis regeneration instead. Worth a line in the description / a migration note so deployers don't hand-edit denomination post-merge.

storopoli (approved):
ACK 5c629750a9fa9c2f000ca3aeedc9c442cd11f712

Agree with @prajwolrg's assessment of the `genparams` issue

delbonis:
Hard NACK

This is not something that makes sense to be in the chainstate at this time, we *should* be threading it through the call chains, that is totally fine.  I would argue that really we should be moving `BridgeParams` to be part of the OL params type, instead of doing `BridgeParams::default()` at the top level separately.

Is there a proposed change to the PRD that stipulates that we should be able to change the bridge params without doing an upgrade?

Edit: oh yep it is in `OLParams`

storopoli:
> Is there a proposed change to the PRD that stipulates that we should be able to change the bridge params without doing an upgrade?

https://github.com/alpenlabs/product/pull/13 partially addresses admin-controlled bridge/operator updates, but I don’t see a PRD requirement that `BridgeParams` specifically be changeable without an upgrade.

> Edit: oh yep it is in `OLParams`

`OLParams` seems consistent with the product PRDs, but the stronger “mutable without upgrade” requirement is not explicitly present.

delbonis:
@storopoli 

Yeah I regret not picking up on this sooner.  I don't have the context as to why the `BridgeParams` were changed to being passed in as a private params in [this PR](https://github.com/alpenlabs/alpen/pull/1814#discussion_r3295746349).  They should always have been taken from `OLParams` baked into the proof binaries at build-time.

It seems that we're just not using `OLParams` after constructing the genesis state.  I should have laid out those code paths when I was reworking the STF last so that that would have clearly been the easier route.

delbonis:
Closing as we've decided on taking a different approach.

PR #2018, opened 2026-06-28
STR-3855 return unused gas on bridge pegout failure

barakshani:
## Description

The reported issue (str-3855) is that on failure on the bridge precompile contract the entire tx gas is consumed -- this is an incorrect behaviour, since as a precompile, gas consumption should be fixed (though dependent on the calldata length).
The crux of the issue is that in the previous implementation, when the contract throws an error it does not make sure to "revert" the call and "refund" the user for the additional gas (beyond the precompile setting). This is being handled in this PR.

The entire code was written by Claude, Opus 4.8 model. I verified it.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[str-3855](https://alpenlabs.atlassian.net/browse/STR-3855)

### DIFF
--- a/crates/reth/evm/src/precompiles/bridge.rs
+++ b/crates/reth/evm/src/precompiles/bridge.rs
@@ -14,42 +14,78 @@ const BRIDGEOUT_BASE_GAS: u64 = 10_000;
 /// Raw EVM gas charged per calldata byte handled by the bridge-out precompile.
 const BRIDGEOUT_CALLDATA_BYTE_GAS: u64 = 16;
 
-/// Solidity `Error(string)` selector: `bytes4(keccak256("Error(string)"))`.
-const ERROR_STRING_SELECTOR: [u8; 4] = [0x08, 0xc3, 0x79, 0xa0];
-
-/// Appends `value` as a 32-byte big-endian ABI word.
-fn push_abi_word(out: &mut Vec<u8>, value: u64) {
-    let mut word = [0u8; 32];
-    word[24..].copy_from_slice(&value.to_be_bytes());
-    out.extend_from_slice(&word);
-}
-
-/// ABI-encodes a human-readable reason as Solidity `Error(string)` revert data so
-/// standard tooling (ethers/web3/foundry) decodes it as `revert("...")`.
-fn abi_encode_error_string(reason: &str) -> Bytes {
-    let reason_bytes = reason.as_bytes();
-    let len = reason_bytes.len();
-    let padded_len = len.div_ceil(32) * 32;
-
-    let mut out = Vec::with_capacity(4 + 64 + padded_len);
-    out.extend_from_slice(&ERROR_STRING_SELECTOR);
-    push_abi_word(&mut out, 32); // offset to the string data (always 0x20)
-    push_abi_word(&mut out, len as u64); // string length
-    out.extend_from_slice(reason_bytes); // string bytes
-    out.resize(4 + 64 + padded_len, 0); // right-pad to a 32-byte boundary
-    Bytes::from(out)
-}
-
-/// Builds a gas-refunding revert carrying an ABI-encoded `Error(string)` reason.
+/// Machine-readable failure reasons returned by the bridge-out precompile.
 ///
-/// Unlike returning `Err(PrecompileError::other(..))` — which is an exceptional halt
-/// that burns all gas forwarded to the call — a revert refunds the unspent gas
-/// (only `gas_used` is charged) and surfaces `reason` as the call's return data.
-fn revert_with_reason(gas_used: u64, reason: &str) -> PrecompileResult {
-    Ok(PrecompileOutput::new_reverted(
-        gas_used,
-        abi_encode_error_string(reason),
-    ))
+/// Each variant is encoded as a Solidity ABI custom error: the 4-byte selector
+/// `bytes4(keccak256(signature))` followed by ABI-encoded parameters. The canonical
+/// definitions consumers decode against live in `IBridgeOut.sol` (next to this file).
+/// The selector bytes below are asserted against `keccak256` of the signatures in
+/// `test_custom_error_selectors_match_signatures`, so they cannot silently drift.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum BridgeOutError {
+    /// `IncorrectCallType()` — precompile was not reached via a direct CALL.
+    IncorrectCallType,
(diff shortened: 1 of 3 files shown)


barakshani:
@codex review

storopoli (approved):
ACK d5fd2f3ccfd2255e1414e6e269c21feed6bac3e5

delbonis (approved):
This looks great but I have one question that might change parts of this a bit:

There would be a lot of value in having the error messages be machine-readable, ie encoding them as a Solidity ABI custom error, like `IncorrectCallType()`, `MalformedCalldata()`, `MalformedCalldataBosd()`, `IncorrectAmount(uint256 /* denom */)`, `OversizeWithdrawal(uint256 /* max */)`, etc.

This *could* allow contracts to be adaptive for handling weird cases across upgrades, especially around the denominations, instead of just throwing their hands up.

barakshani:
> This looks great but I have one question that might change parts of this a bit:
> 
> There would be a lot of value in having the error messages be machine-readable, ie encoding them as a Solidity ABI custom error, like `IncorrectCallType()`, `MalformedCalldata()`, `MalformedCalldataBosd()`, `IncorrectAmount(uint256 /* denom */)`, `OversizeWithdrawal(uint256 /* max */)`, etc.
> 
> This _could_ allow contracts to be adaptive for handling weird cases across upgrades, especially around the denominations, instead of just throwing their hands up.

Feels like the usual Keagen's "let me send you to a sidequest". Let's see if I can easily do that. If not, we'll mark it out of scope.

storopoli:
> Feels like the usual Keagen's "let me send you to a sidequest".

Here's another sidequest: "you need to rebase"....

barakshani:
@codex review

@delbonis review (PTAL)

crates/reth/evm/src/precompiles/bridge.rs
```diff
@@ -14,6 +14,86 @@ const BRIDGEOUT_BASE_GAS: u64 = 10_000;
...
+            Self::OversizeWithdrawal { .. } => [0xb0, 0x70, 0x13, 0x77],
+        }
+    }
+
+    /// The single `uint256` parameter, if the error carries one.
+    const fn param(self) -> Option<u64> {
+        match self {
+            Self::OversizeBosd { max } => Some(max),
+            Self::IncorrectAmount { denomination } => Some(denomination),
+            Self::OversizeWithdrawal { max } => Some(max),
```
delbonis, line 62: Maybe make this return `U256` directly instead of relying on the consumer to convert it?

delbonis (approved):
Oh this is *so* beautiful, thank you! 🙏

functional-tests/tests/alpen_client/precompiles/test_bridgeout_cap.py
```diff
@@ -26,27 +30,92 @@
...
+
+def assert_reverted(rpc, receipt: dict, amount_sats: int, expect_selector: str):
+    """Assert a failing bridgeout reverted with gas refunded and the expected custom error."""
+    status = receipt["status"]
+    status = int(status, 16) if isinstance(status, str) else status
+    assert status == 0, f"expected failure status, got {status}"
+
+    # A revert refunds unspent gas; a gas-burning halt would consume ~GAS_LIMIT.
+    used = gas_used(receipt)
+    assert used < GAS_LIMIT // 2, (
```
bewakes, line 108: non-blocking: I find this a bit weird - the assert string below says the limit is `GAS_LIMIT` but the check is with `GAS_LIMIT // 2`. This might be true today but imo the correct assertion would be `used < GAS_LIMIT`.

barakshani: The values are set such that in practice it is binary -- either we we'd consume about 30k gas (less than half) or we'd consume all gas (more than half). I agree that technically even `GAS_LIMT-1` should work, but I think that the way it is, it is fine. Note that the precompile spends a fix amount of gas, so there's no real risk we will just be on the edge of half gas-limit.

bewakes: I see. I think mentioning this in the comment would be useful here. Because this check is assuming things that is not obvious from around the code itself.

barakshani: Clarification added, thanks for the comment.

crates/reth/evm/src/precompiles/bridge.rs
barakshani: done, please have a look.

barakshani (commented):
Let's get this in...

bewakes (approved):

crates/reth/evm/src/precompiles/bridge.rs
```diff
@@ -14,6 +14,80 @@ const BRIDGEOUT_BASE_GAS: u64 = 10_000;
...
+    /// The Solidity custom-error selector, `bytes4(keccak256(signature))`.
+    const fn selector(self) -> [u8; 4] {
+        match self {
+            Self::IncorrectCallType => [0x7a, 0x5e, 0x63, 0xdc],
+            Self::MalformedCalldata => [0x59, 0x17, 0x0b, 0xf0],
+            Self::MalformedCalldataBosd => [0xc8, 0xe4, 0x58, 0x92],
+            Self::OversizeBosd { .. } => [0x27, 0x25, 0xac, 0x73],
+            Self::NonIntegerAmount => [0xf7, 0x73, 0x8c, 0x57],
+            Self::IncorrectAmount { .. } => [0x88, 0x96, 0x7d, 0x2f],
+            Self::OversizeWithdrawal { .. } => [0xb0, 0x70, 0x13, 0x77],
```
delbonis, line 53: Maybe these should be consts?  Now this fn is just a bunch of magic values.  Not a big deal, can ignore this.

delbonis (approved):
Looks good!

crates/reth/evm/src/precompiles/bridge.rs
barakshani: ignored.

purusang (approved):

prajwolrg (approved):

sapinb (approved):
ACK a9172f24cc56c835f566ff722e285f37a863c617

PR #2022, opened 2026-06-29
fix(docker): make docker-compose-eest.yml runnable for EEST-on-docker

voidash:
The STR-3562 refactor turned `docker/alpen-client/entrypoint.sh` into an env-var arg builder that prepends its own flags, which collided with this compose's explicit `command:` list (duplicate clap args), and the compose mounted a `seq-p2p.hex` from the gitignored `configs/alpen-client/` dir that doesn't exist in a clean checkout. Neither path ever worked in CI.

- Override the entrypoint to `/usr/local/bin/alpen-client` so the explicit command list runs verbatim.
- Allow the alpen-client image to be injected via `ALPEN_CLIENT_IMAGE` so a registry/ECR ref can be substituted.
- Add a tracked `docker/eest/` fixture bundle (`.env.eest` with a deterministic dev/test sequencer keypair + matching `seq-p2p.hex`) and mount the key from there, so a clean checkout has everything the stack needs.

Enables the new `Docker Images Ethereum Spec Tests` workflow in alpenlabs/e2e-runner to run EEST against a Dockerized alpen-client pulled from ECR. Validated with `docker compose config`.


voidash:
Closing this because the EEST test coverage should live in the E2E runner instead of being duplicated in Alpen-specific Docker Compose wiring. The E2E runner will handle these tests properly, so keeping this path here would add parallel maintenance without giving us a better signal.

PR #2030, opened 2026-06-30
feat(dbtool): add EE batch rollback recovery command

irnb:
Add `ee-revert-batches` for offline EE recovery after stale block-witness generation. The command dry-runs by default, reports affected batches, chunks, exec blocks, prover rows, txs, and accepted-frontier state, and requires --force for mutation.

On force, it removes the unaccepted batch suffix, deletes the related exec-block suffix and prover task/proof/receipt rows, and can export raw txs for manual rebroadcast.

Also share chunk/acct prover task key encoding with alpen-client so dbtool uses the same keys as the runtime.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/bin/strata-dbtool/src/cmd/ee_revert.rs
+++ b/bin/strata-dbtool/src/cmd/ee_revert.rs
@@ -9,7 +9,7 @@ use std::{
 
 use alloy_eips::eip2718::Encodable2718 as _;
 use alpen_ee_common::{
-    encode_batch_task_key, encode_chunk_task_key, BatchId, BatchStatus, BatchStorage, Chunk,
+    encode_batch_task_key, encode_chunk_task_key, Batch, BatchId, BatchStatus, BatchStorage, Chunk,
     ChunkId, ChunkStatus, ChunkStorage, EnginePayload, ExecBlockPayload, ExecBlockStorage,
     OLBlockOrEpoch, Storage, StorageError,
 };
@@ -254,17 +254,7 @@ async fn build_plan(
             Box::new(first_batch.idx()),
         )
     })?;
-    let first_reverted_record = storage
-        .get_exec_block(first_reverted_hash)
-        .await
-        .internal_error("Failed to read first reverted EE exec block")?
-        .ok_or_else(|| {
-            DisplayedError::UserError(
-                "First reverted batch block is missing from EE exec-block storage".to_string(),
-                Box::new(hash_hex(first_reverted_hash)),
-            )
-        })?;
-    let first_reverted_block_height = first_reverted_record.blocknum();
+    let first_reverted_block_height = first_batch_blocknum(&first_batch)?;
 
     let (kept_batch, _) = storage
         .get_batch_by_idx(args.from_batch_idx - 1)
@@ -471,6 +461,8 @@ async fn build_plan(
     let mut reverted_batch_blocks_sorted =
         reverted_batch_blocks.iter().copied().collect::<Vec<_>>();
     reverted_batch_blocks_sorted.sort_by_key(|hash| hash_hex(*hash));
+    let mut missing_affected_block_count = 0usize;
+    let mut first_missing_affected_block = None;
     for block_hash in reverted_batch_blocks_sorted {
         if !seen_delete_hashes.contains(&block_hash) {
             if storage
@@ -487,12 +479,17 @@ async fn build_plan(
                     )
                 });
             } else {
-                warnings.push(format!(
-                    "affected batch block {} was not found in the unfinalized suffix to delete",
-                    hash_hex(block_hash)
-                ));
+                missing_affected_block_count += 1;
+                first_missing_affected_block.get_or_insert(block_hash);
             }
         }
+    }
+    if missing_affected_block_count > 0 {
+        warnings.push(format!(
+            "{} affected batch block(s) were not found in the unfinalized suffix to delete; this is expected when retrying after a partial rollback that already deleted exec blocks. first_missing={}",
+            missing_affected_block_count,
+            hash_hex(first_missing_affected_block.expect("missing count is non-zero")),
+        ));
     }
 
     affected_blocks.sort_by(|a, b| {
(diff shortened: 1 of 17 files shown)


crates/alpen-ee/common/src/types/prover_task_key.rs
```diff
@@ -0,0 +1,73 @@
...
+
+fn tagged_range_key(tag: u8, prev_block: Hash, last_block: Hash) -> Vec<u8> {
+    let mut buf = Vec::with_capacity(RANGE_TASK_KEY_BYTES);
+    buf.push(tag);
+    let prev: [u8; 32] = prev_block.into();
+    let last: [u8; 32] = last_block.into();
+    buf.extend_from_slice(&prev);
+    buf.extend_from_slice(&last);
+    buf
+}
```
sapinb, line 101: Encode is moved here, but decode is still inside `alpen-client` in `spec_chunk.rs` and `spec_acct.rs`.
We should keep these the encode and decode co-located. either all here or in their original places, albeit with some duplication.

bin/strata-dbtool/src/cmd/ee_revert.rs
```diff
@@ -0,0 +1,908 @@
...
+                "EE batch not found for --from-batch-idx".to_string(),
+                Box::new(args.from_batch_idx),
+            )
+        })?;
+    let first_reverted_hash = first_batch.blocks_iter().next().ok_or_else(|| {
+        DisplayedError::InternalError(
+            "Non-genesis batch unexpectedly has no blocks".to_string(),
+            Box::new(first_batch.idx()),
+        )
+    })?;
```
sapinb, line 261: batch type enforces that it has at-least 1 block.
Safe to keep it as a precaution though.

prajwolrg (approved):

prajwolrg (approved):
LGTM! 

Not-blocking comment: Let's also include a functional test showing that this works.

irnb:
> Let's also include a functional test showing that this works.

@prajwolrg do you mean add a functional test that lets the ee run, then removes the batch with this api, then lets the ee run again?

i'd like to add it in a separate pr.

crates/alpen-ee/common/src/types/prover_task_key.rs
irnb: fixed https://github.com/alpenlabs/alpen/pull/2030/changes/002ffe4eafa41c22608125008219099235d8aa18

bin/strata-dbtool/src/cmd/ee_revert.rs
```diff
@@ -0,0 +1,908 @@
...
+    let first_reverted_record = storage
+        .get_exec_block(first_reverted_hash)
+        .await
+        .internal_error("Failed to read first reverted EE exec block")?
+        .ok_or_else(|| {
```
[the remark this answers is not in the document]
irnb: fixed https://github.com/alpenlabs/alpen/pull/2030/changes/002ffe4eafa41c22608125008219099235d8aa18

storopoli:
@codex review

bin/strata-dbtool/src/cmd/ee_revert.rs
```diff
@@ -0,0 +1,953 @@
...
+/// all unfinalized exec blocks at or above the first reverted block height,
+/// and cleans exact prover rows derivable from the affected batch/chunk ids.
+/// Dry-run unless `--force` is passed. `--tx-export` writes raw affected
+/// transactions even during dry-run. If the rollback crosses the local
+/// accepted EE frontier, the command also rolls back EE account-state rows to
+/// the kept batch tip or blocks if that target state is unavailable.
+///
+/// Forced execution is intentionally not one cross-tree transaction. Batch
+/// metadata is reverted last so a failed pre-final mutation can be retried
+/// while the command can still reconstruct the affected batch/chunk plan.
```
bewakes: [Claude's report]
The comment here and on lines 137-139 state batch metadata is reverted last "so a failed pre-final mutation can be retried." But the genuinely irreversible step — rollback_ee_account_state (deletes account-state rows for epochs > target) — runs first, before any block/prover deletion.

Fix:
Correct the comment to describe the real invariant: "account-state rollback runs first and is irreversible; retry-safety relies on best_ee_account_state recomputing the frontier below the kept tip on re-run, not on ordering." Add a test that asserts a re-run after a simulated post-account-state failure is a no-op for account state.

bin/strata-dbtool/src/cmd/ee_revert.rs
```diff
@@ -0,0 +1,953 @@
...
+            },
+        },
+        account_state_rollback: account_state_rollback_plan,
+        block_deletions,
+        chunk_artifacts,
+        batch_artifacts,
+    })
+}
+
+fn first_batch_blocknum(batch: &Batch) -> Result<u64, DisplayedError> {
```
bewakes: [Claude]
`first_batch_blocknum` assumes batch blocks occupy a contiguous, gap-free height range.

It derives the first reverted height as `last_blocknum + 1 - blocks_iter().count()`, where `count = inner_blocks.len() + 1`. This is only correct if the batch's blocks are exactly the contiguous heights `[last_blocknum - inner_blocks.len() .. last_blocknum]`.

Fix:
Either document the contiguity precondition at the function, or derive the height from the actual first block's stored record rather than arithmetic — note the prior commit deliberately moved away from reading the first exec-block row to make retry-after-partial-delete work, so prefer an explicit invariant comment + a debug assert that `inner_blocks.len()+1 == last_blocknum - prev_batch.last_blocknum` when that data is available.

storopoli (approved):
ACK 002ffe4eafa41c22608125008219099235d8aa18

bewakes (approved):
Looks good from my side. Did a quick scan. Some comments by Claude.

bin/strata-dbtool/src/cmd/ee_revert.rs
irnb: fix https://github.com/alpenlabs/alpen/pull/2030/changes/e02fccf6e3fdf389f76a7fdaeccf48049d0261a8

PR #2031, opened 2026-06-30
Str 3857 checkpoint sync rpc fix

bewakes:
## Description
**AI Usage:** This PR was created with the assistance of claude and codex.

This PR fixes some of the issues surfaced in checkpoint sync node testing. The main change is the introduction of `OLBlockDataAccess: Available | Unavailable` enum to `OLRpcServer`. The rpc methods then use it to return appropriate response.

**Chain worker** now explicitly stores the new state root for an snark account for the terminal block in checkpoint sync path. This fixes the second issue in [the ticket](https://alpenlabs.atlassian.net/browse/STR-3857). 

**[Breaking Change(only one rpc method)]:** The `RpcCheckpointInfo` now splits `l2_range` into `l2_start: Option<OLBlockCommitment>` and `l2_end: OLBlockCommitment`. This is to make the method compatible with checkpoint sync node which cannot serve the start of the OL range(can only serve terminal block info). An alternative would be to have first value optional in the `l2_range` but although it keeps the response unchanged, it is a bit uglier. This fixes the first issue in [the ticket](https://alpenlabs.atlassian.net/browse/STR-3857).

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

### DIFF
--- a/crates/chain-worker/src/context.rs
+++ b/crates/chain-worker/src/context.rs
@@ -485,8 +485,14 @@ pub(crate) fn build_checkpoint_indexing_writes(
                 found: log.new_msg_idx(),
             });
         }
+        // Checkpoint-sync has no per-block attribution, but the terminal update
+        // of an epoch carries a recoverable post-epoch root; earlier updates
+        // have `None`.
+        let update_meta = update
+            .state()
+            .map(|root| AccountUpdateMeta::new(None, root));
         let record = AccountUpdateRecord::new(
-            None,
+            update_meta,
             *update.seqno().inner(),
             update.prev_next_read_idx(),
             update.next_read_idx(),
@@ -714,6 +720,66 @@ mod tests {
                 found: 0
             }
         ));
+    }
+
+    /// Builds an output with snark updates and one matching log per update.
+    fn checkpoint_output_with_snark_updates(
+        updates: impl IntoIterator<Item = (AccountId, AccountSerial, SnarkAcctStateUpdate)>,
+    ) -> OLBlockExecutionOutput {
+        let mut indexer_writes = IndexerWrites::new();
+        let mut logs = Vec::new();
+        for (_account_id, serial, update) in updates {
+            let log_data =
+                SnarkAccountUpdateLogData::new(update.next_read_idx(), vec![0xaa]).unwrap();
+            logs.push(OLLog::new(serial, log_data.encode_log().unwrap()));
+            indexer_writes.push_snark_acct_update(update);
+        }
+        OLBlockExecutionOutput::new(Buf32::zero(), WriteBatch::default(), indexer_writes, logs)
+    }
+
+    /// Regression: a checkpoint-sync update that carries a terminal post-epoch
+    /// root must surface that root in the index record. Today
+    /// `build_checkpoint_indexing_writes` drops it (`update_meta: None`),
+    /// which is what makes RPC `new_state_root` come back null.
+    #[test]
+    fn build_checkpoint_indexing_writes_keeps_terminal_root() {
+        let account_id = AccountId::from([7u8; 32]);
+        let serial = AccountSerial::from(7u32);
+        let root = Hash::from([9u8; 32]);
+
+        let update = SnarkAcctStateUpdate::new(account_id, Some(root), 0, 3, Seqno::from(5));
+        let output = checkpoint_output_with_snark_updates([(account_id, serial, update)]);
+
+        let writes = build_checkpoint_indexing_writes(&output).unwrap();
+        let records = writes
+            .account_updates()
+            .get(&account_id)
+            .expect("account update should be present");
+        assert_eq!(records.len(), 1);
+
+        let meta = records[0]
+            .update_meta()
+            .expect("terminal update must carry root metadata");
+        assert_eq!(meta.new_state_root(), root);
+        assert!(
+            meta.block_commitment().is_none(),
+            "checkpoint-sync rows have no block attribution"
+        );
+    }
+
+    /// Updates with no root (earlier updates in a multi-update epoch) stay
+    /// `update_meta: None` — the intermediate roots are genuinely unavailable.
+    #[test]
+    fn build_checkpoint_indexing_writes_leaves_non_terminal_root_absent() {
+        let account_id = AccountId::from([8u8; 32]);
+        let serial = AccountSerial::from(8u32);
+
+        let earlier = SnarkAcctStateUpdate::new(account_id, None, 0, 2, Seqno::from(1));
+        let output = checkpoint_output_with_snark_updates([(account_id, serial, earlier)]);
+
+        let writes = build_checkpoint_indexing_writes(&output).unwrap();
+        let records = writes.account_updates().get(&account_id).unwrap();
+        assert!(records[0].update_meta().is_none());
     }
 
     fn output_with_l1_block_records(
--- a/crates/chain-worker/src/context.rs
+++ b/crates/chain-worker/src/context.rs
@@ -454,9 +454,7 @@ fn debug_assert_contiguous_update_ranges(
 
 /// Builds an [`IndexingWrites`] payload for a DA-reconstructed epoch.
 ///
-/// Like [`build_indexing_writes`] but with no per-block attribution and no
-/// per-update state root: update records carry `update_meta: None`. The
-/// post-epoch root lives on the epoch summary.
+/// Like [`build_indexing_writes`] but with no per-block attribution.
 pub(crate) fn build_checkpoint_indexing_writes(
     output: &OLBlockExecutionOutput,
 ) -> WorkerResult<IndexingWrites> {
--- a/crates/chain-worker/src/state.rs
+++ b/crates/chain-worker/src/state.rs
@@ -17,7 +17,7 @@ use strata_asm_proto_checkpoint_types::{CheckpointSidecar, CheckpointTip};
 use strata_bridge_params::BridgeParams;
 use strata_checkpoint_types::EpochSummary;
 use strata_db_types::errors::DbError;
-use strata_identifiers::{AccountId, Buf32, Epoch, Hash, OLBlockCommitment};
+use strata_identifiers::{AccountId, Buf32, Epoch, OLBlockCommitment};
 use strata_ledger_types::{
     IAccountState, ISnarkAccountState, IStateAccessor, StateError, StateResult,
 };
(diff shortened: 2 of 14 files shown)


storopoli (approved):
ACK 1f7328c60de0793f2b7a439c2e93931e82391f70

crates/chain-worker/src/state.rs
```diff
@@ -841,9 +846,45 @@ fn rebuild_snark_records_from_logs<S: IStateAccessor>(
...
+        .ok_or_else(|| {
+            WorkerError::Unexpected(format!("post-state account {account_id} missing"))
+        })?;
+    acct.as_snark_account()
+        .map(|s| s.inner_state_root())
+        .map_err(|source| WorkerError::AccountStateRead {
+            stage: "post-state as_snark_account",
+            source,
+        })
+}
```
delbonis: This function simply reads an accounts inner state root, it doesn't care about the scenario it's being used in.  So remove all the phrasing about "post epoch state", it's misleading.

Also the error structure here is a bit messy, maybe we should remove `Unexpected` entirely.

crates/chain-worker/src/state.rs
```diff
@@ -841,9 +846,45 @@ fn rebuild_snark_records_from_logs<S: IStateAccessor>(
...
+        let root = post_epoch_inner_state_root(state, account_id)?;
+        let prev = &records[idx];
+        records[idx] = SnarkAcctStateUpdate::new(
+            account_id,
+            Some(root),
+            prev.prev_next_read_idx(),
+            prev.next_read_idx(),
+            prev.seqno(),
+        );
+    }
```
delbonis, line 837: This is messy, we're reconstructing the update over top of itself.  Surely we can inspect the inner state root as we're inserting the record in the above loop if we detect that this is the last log entry, instead of having to loop over it a second time.

crates/chain-worker/src/state.rs
```diff
@@ -796,6 +796,9 @@ fn rebuild_snark_records_from_logs<S: IStateAccessor>(
 ) -> WorkerResult<(Vec<SnarkAcctStateUpdate>, SnarkAccountCursors)> {
     let mut next_cursor: SnarkAccountCursors = HashMap::new();
     let mut records = Vec::with_capacity(ol_logs.len());
+    // Index of each account's last update, so we can stamp the recoverable
+    // post-epoch root onto it after the walk.
+    let mut terminal_idx: HashMap<AccountId, usize> = HashMap::new();
```
delbonis: The name `terminal_idx` "terminal index" is factually true, but it's misleading because the abbr `idx` almost always refers to an ordinal index in some sequence rather than an "index over a collection of data".  The term "terminal" is also factually true but misleading, since together this makes it seem like it could be talking about the slot of an epoch terminal.

~~A more precise name that avoids all this confusion could be `updated_acct_seqnos`.~~

crates/chain-worker/src/state.rs
```diff
@@ -841,9 +846,45 @@ fn rebuild_snark_records_from_logs<S: IStateAccessor>(
         ));
     }
 
+    // The post-DA state holds each account's final inner state root, which is
+    // the terminal update's post-state root. Earlier updates stay `None`.
+    for (account_id, idx) in terminal_idx {
+        let root = post_epoch_inner_state_root(state, account_id)?;
+        let prev = &records[idx];
+        records[idx] = SnarkAcctStateUpdate::new(
```
delbonis: Ah except I'm realizing that my last comment is incorrect!  Instead of doing `records[idx]` we could just do `records.last_mut()` and not need to keep track of that, just use a hashset.

delbonis (changes requested):

crates/chain-worker/src/state.rs
bewakes: Cleaned up the error mess. `Unexpected` is being used at other places so left it as is.

bewakes:
Thanks @delbonis for your comments. Although the second pass can't be avoided in this case, I've made other cleanups. PTAL at the [last commit](https://github.com/alpenlabs/alpen/pull/2031/commits/9cfc4278f55a4a34677ffe3623e4afa6947be737).

Second pass is unavoidable here because the logs contain updates from any account, so we can't know the last entry on a forward pass. we could try backward pass but then there's some bookkeeping with cursors which isn't very safe to modify at this point.

delbonis:
@bewakes Hmm I must have misunderstood the algorithm because GitHub hid parts of it in the fold.  I'll take another look.

crates/chain-worker/src/state.rs
```diff
@@ -841,9 +816,47 @@ fn rebuild_snark_records_from_logs<S: IStateAccessor>(
...
+    state: &S,
+    account_id: AccountId,
+) -> StateResult<Hash> {
+    snark_acct(state, account_id).map(|s| s.inner_state_root())
+}
+
+/// Gets an account's sequencer number.
+fn snark_acct_seq_no<S: IStateAccessor>(state: &S, account_id: AccountId) -> StateResult<Seqno> {
+    snark_acct(state, account_id).map(|s| s.seqno())
+}
```
delbonis: These should all be verby names.

But actually, do we really need these latter two?  They're both only used once and both only perform a mapping operation.  They don't restructure error variants like (what should be) `get_snark_acct`.

crates/chain-worker/src/state.rs
```diff
@@ -796,6 +772,9 @@ fn rebuild_snark_records_from_logs<S: IStateAccessor>(
 ) -> WorkerResult<(Vec<SnarkAcctStateUpdate>, SnarkAccountCursors)> {
     let mut next_cursor: SnarkAccountCursors = HashMap::new();
     let mut records = Vec::with_capacity(ol_logs.len());
+    // Index of each account's last update, so we can stamp the recoverable
+    // post-epoch root onto it after the walk.
+    let mut last_record_idx: HashMap<AccountId, usize> = HashMap::new();
```
delbonis: This var `last_record_idx` name is still misleading for half of the same reasons I mentioned before.

I'm still unhappy about this function.  The return type is very complicated and the data structure it returns is difficult to query since you need to scan backwards from the end to find the post state of any particular snark account you're looking for.

Should structs like this:

```rs
struct EpochSnarkAcctChanges { // not sure about name
  accounts: HashMap<AccountId, SnarkAccountChanges>
}

struct SnarkAccountChanges {
  new_updates: Vec<SnarkAcctStateUpdate>,
  final_seqno: Seqno,
  final_next_inbox_msg_idx: u64,
}
```

Also the name `rebuild_snark_records_from_logs` is misleading since it implies that it's rebuilding records about all kinds of snark operations.

crates/chain-worker/src/state.rs
```diff
@@ -753,34 +745,18 @@ fn collect_pre_snark_account_cursors<S: IStateAccessor>(
...
                 },
-            },
-            None => SnarkAccountCursor {
-                seqno: Seqno::zero(),
-                next_inbox_msg_idx: 0,
+                Err(StateError::MissingAccount(_)) => SnarkAccountCursor::new_empty(),
+                Err(source) => {
+                    return Err(acct_read_err("pre-state account read")(source));
+                }
             },
```
delbonis: It *would* make sense to split out this logic as a separate fn since it does two state queries and restructures the errors.

crates/chain-worker/src/state.rs
```diff
@@ -48,6 +50,15 @@ struct SnarkAccountCursor {
...
 }
 
+impl SnarkAccountCursor {
+    pub(crate) fn new_empty() -> Self {
+        Self {
+            seqno: Seqno::zero(),
+            next_inbox_msg_idx: 0,
+        }
+    }
+}
```
delbonis: Could be actually `#[derive(Default)]`.  The `new_empty` constructor implies that it's a container (and `new_empty` makes sense when there's a `new` constructor that takes all the fields but there's plausible empty values for some/all of them).

delbonis (changes requested):

crates/chain-worker/src/context.rs
krsnapaudel: Does this need an update?

krsnapaudel (approved):
One nit. Looks good otherwise.

crates/chain-worker/src/state.rs
bewakes: > I'm still unhappy about this function. The return type is very complicated and the data structure it returns is difficult to query since you need to scan backwards from the end to find the post state

I am not very proud of this function either and I partly agree. The return type is slightly complicated. But it is internal to this module and is not meant to be queried. What we are trying to do here is to construct the update records as they appear in the logs. Aggregating them into a hashmap per account won't buy us much. Also, the cursor already holds the latest sequence number and next inbox message, it just doesn't have the latest root. However, once we find out the latest root and update the records, we don't need to access it further.

I've created a different struct for the return type that aggregates the current tuple and I have renamed the function to `reconstruct_snark_acct_update_records` and the variable to `acct_last_record_idx` which I think are better. If you think otherwise please suggest precise names.

crates/chain-worker/src/context.rs
bewakes: I've updated the docstring to not assume what data comes in. Only mentions about non-existent block references.

crates/chain-worker/src/state.rs
delbonis: See my initial comment:

> abbr `idx` almost always refers to **an ordinal index in[to] some sequence** rather than an "index over a collection of data".

(emphasis added)

Better names:

* `last_record_idxs` - in reference to how the values *are* idxs
* `last_records_tbl` - indicates that it's basically a table, with the idx part implied

Adding `acct` to the name doesn't help to improve clarity (since we already know the function is doing something with acct update records based on its name), although it doesn't hurt either.

I realize I'm being a bit pedantic with this but it's important for communication.

delbonis: As far as the dataflow goes: Yeah after investigating it some more I don't see a good solution that doesn't involve changing a lot more to shake off more weirdness, which isn't worth it at this time.

bewakes: Renamed to `last_record_idxs` in https://github.com/alpenlabs/alpen/pull/2031/commits/cb487470bb219e437efc230066fd6916a072b4ff

delbonis (approved):

PR #2065, opened 2026-07-01
fix(params): skip ci-build lookup for manual genparams

purusang:
## Description

Manual `ci-genparams.yml` runs currently fail before params generation when the requested datatool image commit does not have a matching successful `ci-build.yml` run. This changes the workflow so manual params generation trusts the supplied commit as the datatool image tag source and proceeds directly to checkout, image pull, and params generation.

The workflow still derives the short image tag from the `commit` input, and `checkout_ref` remains available when callers need to generate params with scripts/templates from a specific ref. By default, checkout uses the workflow ref so branch/template changes are exercised when testing a workflow branch.

### Type of Change

- [x] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Validation:

- `actionlint .github/workflows/ci-genparams.yml`
- `git diff --check`
- Manual workflow run against this branch reached params generation path instead of failing in `Resolve image commit`: https://github.com/alpenlabs/alpen/actions/runs/28513430571/job/84519080551

- [ ]  Yes
- [x]  No

N/A

## Checklist

AI assistance notice: This PR was prepared with AI assistance.

## Related Issues

N/A

### DIFF
--- a/.github/params/README.md
+++ b/.github/params/README.md
@@ -1,6 +1,6 @@
 # CI Params Generation
 
-Generates deployment params (`ol-params.json`, `asm-params.json`) using the prebuilt datatool image tag from a given commit. Manual runs do not validate that the tag came from a successful `ci-build.yml` run; the workflow checks out the workflow ref by default, then proceeds directly to pulling the requested datatool image and generating params.
+Generates deployment params (`ol-params.json`, `asm-params.json`) using a prebuilt datatool image and the params scripts/templates from the checked-out repo ref. Manual runs validate that `datatool_image_commit` looks like a commit SHA, but they do not prove that the ECR tag came from a successful `ci-build.yml` run. A missing image tag still fails at `docker pull`.
 
 ## Templates
 
@@ -16,14 +16,29 @@ When adding a new environment or updating an existing one, commit the static val
 ```
 gh workflow run ci-genparams.yml --ref <branch> \
   -f env=<staging-v2|prod> \
-  -f commit=<short-sha> \
+  -f datatool_image_commit=<7-to-40-char-hex-sha> \
   -f genesis_l1_height=<height> \
   -f chain_config=<path-to-chainspec>
 ```
 
+Dispatch ref:
+
+| Argument | Description |
+|-------|-------------|
+| `--ref <branch>` | Branch/ref whose workflow file GitHub Actions runs. For normal runs, set this to the branch or commit to test and omit `checkout_ref`. |
+
+Workflow inputs:
+
+| Input | Description |
+|-------|-------------|
+| `datatool_image_commit` | Commit whose first 7 chars identify the prebuilt datatool image tag. Must be a 7-40 char lowercase hex SHA. |
+| `checkout_ref` | Optional override for the repo ref checked out inside the job for params scripts/templates. Use only when the workflow file should come from `--ref`, but params scripts/templates should come from a different ref. |
+
+When `checkout_ref` is omitted, the job checks out the workflow run commit (`github.sha`) for params scripts/templates. In the common case, this is the commit selected by `--ref`.
+
 Download the artifact:
 ```
-gh run download <run-id> -n params-<env>-<commit>
+gh run download <run-id> -n params-<env>-<datatool-image-tag>
 ```
 
 ## GitHub Environment Setup
--- a/.github/params/README.md
+++ b/.github/params/README.md
@@ -32,6 +32,8 @@ Workflow inputs:
 | Input | Description |
 |-------|-------------|
 | `datatool_image_commit` | Commit whose first 7 chars identify the prebuilt datatool image tag. Must be a 7-40 char lowercase hex SHA. |
+| `genesis_l1_height` | Genesis L1 block height. Must be a non-negative integer. |
+| `chain_config` | Path to a chainspec file. Must be single-line and exist in the checkout. |
 | `checkout_ref` | Optional override for the repo ref checked out inside the job for params scripts/templates. Use only when the workflow file should come from `--ref`, but params scripts/templates should come from a different ref. |
 
 When `checkout_ref` is omitted, the job checks out the workflow run commit (`github.sha`) for params scripts/templates. In the common case, this is the commit selected by `--ref`.
--- a/.github/params/README.md
+++ b/.github/params/README.md
@@ -1,6 +1,6 @@
 # CI Params Generation
 
-Generates deployment params (`ol-params.json`, `asm-params.json`) using the prebuilt datatool image from a given commit.
+Generates deployment params (`ol-params.json`, `asm-params.json`) using the prebuilt datatool image tag from a given commit. Manual runs do not validate that the tag came from a successful `ci-build.yml` run; the workflow checks out the workflow ref by default, then proceeds directly to pulling the requested datatool image and generating params.
 
 ## Templates
 
--- a/.github/workflows/ci-genparams.yml
+++ b/.github/workflows/ci-genparams.yml
@@ -74,6 +74,11 @@ jobs:
 
           # Optional caller-selected repo ref for params scripts/templates.
           if [ -n "${INPUT_CHECKOUT_REF}" ]; then
+            if [[ "${INPUT_CHECKOUT_REF}" == *$'\n'* || "${INPUT_CHECKOUT_REF}" == *$'\r'* ]]; then
+              echo "::error::checkout_ref must be a single-line repo ref"
+              exit 1
+            fi
+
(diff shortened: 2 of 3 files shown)


storopoli (approved):
ACK bc11879dac1f65ea462514b239a590b30eb54e6f

storopoli (changes requested):
The direction makes sense. The `ci-build` lookup this removes was never a real provenance check: it had no `--branch` filter, `ci-build.yml` builds and pushes from arbitrary refs on manual dispatch, and ECR tags are mutable — so a SHA match against run history never guaranteed the image bytes. What it actually provided was fail-fast on a bad tag, and `docker pull` still gives us that, just later in the job.

Two asks before merge:

1. **Validate the `commit` input.** `SHORT_TAG` is now raw dispatch input fed
   straight into `docker pull` as the tag. Restrict it to a hex SHA so the
   workflow can't be pointed at arbitrary named tags (`latest`, stale ones,
   etc.):

   ```bash
   if ! [[ "${INPUT_COMMIT}" =~ ^[0-9a-f]{7,40}$ ]]; then
     echo "::error::commit must be a 7-40 char lowercase hex commit SHA"
     exit 1
   fi
   ```
2. Drop the `commit_sha` output (or comment why it's kept). It's now just the raw input echoed back, and nothing in the workflow consumes it — `Checkout` uses `checkout_ref` and the pull/generate/artifact steps use `short_tag`. Keeping an unvalidated value named `commit_sha` invites someone to trust it later.

.github/workflows/ci-genparams.yml
```diff
@@ -46,56 +46,53 @@ concurrency:
...
           fi
 
-          set_output short_tag "${SHORT_TAG}"
-          set_output commit_sha "${BUILD_SHA}"
-          set_output checkout_ref "${CHECKOUT_REF}"
-          echo "Selected image build: ${SHORT_TAG} (${BUILD_SHA})"
-          echo "Checkout ref: ${CHECKOUT_REF}"
+          set_output datatool_image_tag "${DATATOOL_IMAGE_TAG}"
+          set_output datatool_image_commit "${DATATOOL_IMAGE_COMMIT}"
+          set_output params_source_ref "${PARAMS_SOURCE_REF}"
```
storopoli, line 110: Keep `checkout_ref` from overwriting the validated image tag

`checkout_ref` is raw dispatch input, and this helper emits it with single-line `$GITHUB_OUTPUT` syntax. A value like `main\ndatatool_image_tag=latest` is parsed as two outputs; because this call comes after the validated tag, the duplicate `datatool_image_tag` wins. Both Docker pull and params generation then use `latest`, bypassing the SHA-only guard from the prior review. Please reject CR/LF in `checkout_ref` (they are not valid refs) before writing it, or serialize it with a safe multiline output format.

.github/workflows/ci-genparams.yml
```diff
@@ -46,56 +46,53 @@ concurrency:
...
+            PARAMS_SOURCE_REF="${WORKFLOW_REF_SHA}"
           fi
 
-          set_output short_tag "${SHORT_TAG}"
-          set_output commit_sha "${BUILD_SHA}"
-          set_output checkout_ref "${CHECKOUT_REF}"
-          echo "Selected image build: ${SHORT_TAG} (${BUILD_SHA})"
-          echo "Checkout ref: ${CHECKOUT_REF}"
+          set_output datatool_image_tag "${DATATOOL_IMAGE_TAG}"
+          set_output datatool_image_commit "${DATATOOL_IMAGE_COMMIT}"
```
storopoli: Remove the still-unused commit output

This is the same unused output called out in the previous review, now renamed from `commit_sha` to `datatool_image_commit`. Nothing consumes `steps.resolve.outputs.datatool_image_commit`, and there is still no comment explaining a future consumer. Please drop it (the log line already records the validated full input), or document why the step contract needs it.

storopoli (changes requested):

.github/workflows/ci-genparams.yml
purusang: Yes, thanks! Done in https://github.com/alpenlabs/alpen/pull/2065/commits/6b916105fba78790a58ad9b38cefc13c6a165c79

.github/workflows/ci-genparams.yml
purusang: Nice, resolved in https://github.com/alpenlabs/alpen/pull/2065/changes/2b4dccdd86acefe3c2ca8b9785a7d9ab2ebbd5dc

.github/workflows/ci-genparams.yml
```diff
@@ -125,7 +145,7 @@ jobs:
       - name: Generate params
         env:
           DEPLOY_ENV: ${{ inputs.env }}
-          COMMIT: ${{ steps.resolve.outputs.short_tag }}
+          DATATOOL_IMAGE_TAG: ${{ steps.resolve.outputs.datatool_image_tag }}
```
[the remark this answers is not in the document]
purusang: Made backward compatible in https://github.com/alpenlabs/alpen/pull/2065/changes/9a85014079f0cbf1fa44f348bf7720e954dba9bd

purusang:
@codex review

.github/workflows/ci-genparams.yml
purusang: Force pushed to https://github.com/alpenlabs/alpen/pull/2065/changes/9a85014079f0cbf1fa44f348bf7720e954dba9bd

.github/workflows/ci-genparams.yml
```diff
@@ -46,56 +46,76 @@ concurrency:
...
-          BUILD_SHA=$(gh run list -R "${{ github.repository }}" \
-            --workflow ci-build.yml \
-            --status success \
-            --limit 100 \
-            --json headSha \
-            --jq ".[] | select(.headSha | startswith(\"${SHORT_TAG}\")) | .headSha" | head -n1)
-
-          if [ -z "${BUILD_SHA}" ]; then
-            echo "::error::No successful ci-build found for ${SHORT_TAG}"
+          fail() {
```
storopoli, line 67: nit: Validation errors never reach the log. `fail` writes the `::error::` message to stdout, but both validators run inside command substitutions (`DATATOOL_IMAGE_TAG=$(...)`, `PARAMS_SOURCE_REF=$(...)`), so the message is captured into the variable and discarded when the shell exits. A bad `datatool_image_commit` still fails the step (exit 1 under `bash -e`), but with a completely empty log — verified by executing this block with an invalid input.

Either send the message to stderr (`>&2` — plain text, no annotation, but visible) or run the validations as statements before the assignments so the `::error::` workflow command stays on the step's stdout.

.github/workflows/ci-genparams.yml
```diff
@@ -112,8 +132,8 @@ jobs:
 
       - name: Pull datatool image
         env:
-          COMMIT: ${{ steps.resolve.outputs.short_tag }}
-        run: docker pull "${DATATOOL_REPO}:${COMMIT}"
+          DATATOOL_IMAGE_TAG: ${{ steps.resolve.outputs.datatool_image_tag }}
+        run: docker pull "${DATATOOL_REPO}:${DATATOOL_IMAGE_TAG}"
 
       - name: Verify chainspec
```
storopoli, line 138: nit: `chain_config` is echoed raw here before any CR/LF check — that only happens later in `generate-params.sh`. A multiline value like `missing.json\n::error::forged` prints a runner command on its own line, letting a dispatcher forge annotations in the log. It can't execute anything or touch `$GITHUB_OUTPUT`, so purely cosmetic, but a single-line check before the echo (same as the one `checkout_ref` now gets) closes it.

.github/params/README.md
```diff
@@ -16,14 +16,31 @@ When adding a new environment or updating an existing one, commit the static val
...
+  -f datatool_image_commit=<7-to-40-char-hex-sha> \
   -f genesis_l1_height=<height> \
   -f chain_config=<path-to-chainspec>
 ```
 
+Dispatch ref:
+
+| Argument | Description |
+|-------|-------------|
+| `--ref <branch>` | Branch/ref whose workflow file GitHub Actions runs. For normal runs, set this to the branch or commit to test and omit `checkout_ref`. |
```
storopoli, line 28: nit: `gh workflow run --ref` only accepts a branch or tag name, not a commit SHA — dispatching with a raw commit fails. Suggest wording this as "branch (or tag) to test" and pointing at `checkout_ref` for pinning to a specific commit.

storopoli (approved):
ACK 9a85014079f0cbf1fa44f348bf7720e954dba9bd

non-blocking nits.

PR #2069, opened 2026-07-02
fix(strata): Decouple BTCIO startup from CSM state

storopoli:
## Description

Decouples BTCIO reader startup from client-state rows by deriving the next L1 read height only from stored L1 canonical storage and ASM genesis.

Adds startup reconciliation before CSM starts so stored L1 canonical state is checked against bitcoind, orphaned suffixes are rewound within the configured reorg lookback, and the reconciled tip is submitted to ASM. CSM startup now seeds from the reconciled/latest ASM state and relies on CSM’s canonical gap-fill instead of a capped historical ASM status list.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The BTCIO startup change is behavior-preserving because any previously accepted client-state target was already bounded by the stored L1 canonical tip.

- []  Yes
- [x]  No

## Checklist

## Related Issues

STR-3845

### DIFF
--- a/crates/consensus-logic/src/sync_manager.rs
+++ b/crates/consensus-logic/src/sync_manager.rs
@@ -26,8 +26,10 @@ use tracing::{debug, info, warn};
 
 use crate::{asm_worker_context::AsmWorkerCtx, csm_worker_context::CsmWorkerContextImpl};
 
-const STARTUP_RECONCILE_RPC_ATTEMPTS: usize = 12;
+#[cfg(not(test))]
 const STARTUP_RECONCILE_RPC_RETRY_DELAY: Duration = Duration::from_secs(5);
+#[cfg(test)]
+const STARTUP_RECONCILE_RPC_RETRY_DELAY: Duration = Duration::from_millis(1);
 
 /// Holds startup inputs that seed CSM before it subscribes to live ASM updates.
 struct CsmListenerStartup<'a> {
@@ -344,11 +346,11 @@ fn is_retryable_bitcoin_error(err: &anyhow::Error) -> bool {
     })
 }
 
-/// Runs a Bitcoin RPC operation with bounded retries for transient failures.
+/// Runs a Bitcoin RPC operation with retries for transient failures.
 ///
-/// Non-retryable errors fail immediately. Retryable errors are retried up to
-/// [`STARTUP_RECONCILE_RPC_ATTEMPTS`] with
-/// [`STARTUP_RECONCILE_RPC_RETRY_DELAY`] between attempts.
+/// Non-retryable errors fail immediately. Retryable errors are retried with
+/// [`STARTUP_RECONCILE_RPC_RETRY_DELAY`] between attempts so startup waits for
+/// bitcoind instead of aborting before the BTCIO reader is launched.
 async fn retry_bitcoin_startup_rpc<T, F, Fut>(
     operation: &'static str,
     mut f: F,
@@ -357,26 +359,23 @@ where
     F: FnMut() -> Fut,
     Fut: Future<Output = anyhow::Result<T>>,
 {
-    for attempt in 1..=STARTUP_RECONCILE_RPC_ATTEMPTS {
+    let mut attempt = 1;
+    loop {
         match f().await {
             Ok(value) => return Ok(value),
-            Err(err)
-                if attempt < STARTUP_RECONCILE_RPC_ATTEMPTS && is_retryable_bitcoin_error(&err) =>
-            {
+            Err(err) if is_retryable_bitcoin_error(&err) => {
                 warn!(
                     operation,
                     attempt,
-                    max_attempts = STARTUP_RECONCILE_RPC_ATTEMPTS,
                     %err,
                     "retryable Bitcoin RPC error during startup reconciliation"
                 );
                 sleep(STARTUP_RECONCILE_RPC_RETRY_DELAY).await;
+                attempt += 1;
             }
             Err(err) => return Err(err),
         }
     }
-
-    unreachable!("retry loop always returns from inside the bounded range")
 }
 
 pub fn spawn_asm_worker_with_ctx(nodectx: &NodeContext) -> anyhow::Result<AsmWorkerHandle> {
@@ -455,7 +454,10 @@ fn prefill_asm_mmr(handle: &MmrIndexHandle, target_count: u64) -> anyhow::Result
 
 #[cfg(test)]
 mod tests {
-    use std::collections::BTreeMap;
+    use std::{
+        collections::BTreeMap,
+        sync::atomic::{AtomicUsize, Ordering},
+    };
 
     use strata_db_store_sled::test_utils::get_test_sled_backend;
     use strata_identifiers::Buf32;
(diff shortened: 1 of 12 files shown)


crates/consensus-logic/src/sync_manager.rs
```diff
@@ -103,6 +102,256 @@ fn spawn_csm_listener(
...
+/// Returns the reconciled canonical ASM block CSM should use as its startup
+/// seed. `None` means startup had no stored L1 canonical target to submit.
+pub async fn reconcile_l1_storage_and_submit_to_asm(
+    storage: Arc<NodeStorage>,
+    bitcoin_client: Arc<Client>,
+    asm_handle: &AsmWorkerHandle,
+    genesis_l1_height: L1Height,
+    l1_reorg_safe_depth: u32,
+) -> anyhow::Result<Option<L1BlockCommitment>> {
+    let bitcoind_tip_height = retry_bitcoin_startup_rpc("fetching Bitcoin block count", || {
```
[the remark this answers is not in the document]
storopoli: Ok makes sense nice catch. Done in c3f54dbfb88ecbbf0419c9b2083d83ed93f38aab Please recheck @codex.

crates/consensus-logic/src/sync_manager.rs
```diff
@@ -2,39 +2,67 @@
...
 use strata_storage::{MmrId, MmrIndexHandle, NodeStorage};
 use strata_tasks::TaskExecutor;
-use tokio::runtime::Handle;
+use tokio::{runtime::Handle, time::sleep};
+use tracing::{debug, info, warn};
 
 use crate::{asm_worker_context::AsmWorkerCtx, csm_worker_context::CsmWorkerContextImpl};
 
+const STARTUP_RECONCILE_RPC_ATTEMPTS: usize = 12;
+const STARTUP_RECONCILE_RPC_RETRY_DELAY: Duration = Duration::from_secs(5);
```
delbonis: Should these be configurable?

delbonis (approved):

crates/consensus-logic/src/sync_manager.rs
storopoli: We could if this becomes an issue in the future. I'd say keep as it is for simplicity and avoiding configuration drift.

crates/consensus-logic/src/sync_manager.rs
```diff
@@ -103,6 +102,286 @@ fn spawn_csm_listener(
...
+
+        if height < stored_tip_height {
+            if bitcoind_is_behind && !observed_divergence {
+                info!(
+                    stored_tip_height,
+                    bitcoind_tip_height,
+                    matched_height = height,
+                    "bitcoind is behind stored L1 canonical tip; preserving stored suffix"
+                );
+                return Ok(Some(L1BlockCommitment::new(height, stored_blkid)));
```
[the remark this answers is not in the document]
storopoli: Ok fair enough done in cec92d95aa27b488b6e1098afdd511e2898f7c58 @codex please check

crates/consensus-logic/src/sync_manager.rs
```diff
@@ -103,6 +102,283 @@ fn spawn_csm_listener(
...
+    bitcoin_client: Arc<Client>,
+    asm_handle: &AsmWorkerHandle,
+    genesis_l1_height: L1Height,
+    l1_reorg_safe_depth: u32,
+) -> anyhow::Result<Option<L1BlockCommitment>> {
+    if !should_reconcile_l1_storage_at_startup(storage.as_ref(), genesis_l1_height).await? {
+        return Ok(None);
+    }
+
+    let bitcoind_tip_height = retry_bitcoin_startup_rpc("fetching Bitcoin block count", || {
```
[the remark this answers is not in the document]
storopoli: Done in f2b25bafa664ec2fcc41b16478270a250082926b and e74616a8943c89725348b8c2673f686cc81ca569 @codex check it again.

crates/consensus-logic/src/sync_manager.rs
```diff
@@ -103,6 +104,285 @@ fn spawn_csm_listener(
...
+                warn!(
+                    stored_tip_height,
+                    pivot_height = height,
+                    "stored L1 canonical suffix diverged from bitcoind; rewinding before startup"
+                );
+            }
+
+            storage
+                .l1()
+                .revert_canonical_chain_async(height)
```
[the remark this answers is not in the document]
storopoli: Ok done in 7bcf241a76093e92d8811224f12b0309fe49e6ee
Please check @codex

crates/consensus-logic/src/sync_manager.rs
```diff
@@ -103,6 +104,299 @@ fn spawn_csm_listener(
...
+        return Ok(None);
+    }
+
+    let max_rewind_depth = l1_reorg_safe_depth.max(1).saturating_sub(1);
+    let safe_rewind_floor = stored_tip_height
+        .saturating_sub(max_rewind_depth)
+        .max(genesis_l1_height);
+    let bitcoind_is_behind = bitcoind_tip_height < stored_tip_height;
+    if bitcoind_is_behind && bitcoind_tip_height < safe_rewind_floor {
+        bail!(
```
krsnapaudel, line 237: Should this be a wait rather than a bail? A successful `get_block_count` returning a low height isn't caught by `retry_bitcoin_startup_rpc`, so this exits the process when bitcoind is merely behind the stored tip.

crates/consensus-logic/src/sync_manager.rs
```diff
@@ -103,6 +104,299 @@ fn spawn_csm_listener(
...
+/// Returns `true` when startup reconciliation needs Bitcoin RPC data.
+///
+/// Empty L1 storage and pre-genesis stored tips do not require reconciliation,
+/// so startup avoids Bitcoin RPCs and lets the BTCIO reader perform its normal
+/// retry loop.
+async fn should_reconcile_l1_storage_at_startup(
+    storage: &NodeStorage,
+    genesis_l1_height: L1Height,
+) -> anyhow::Result<bool> {
+    let Some((stored_tip_height, _)) = storage.l1().get_canonical_chain_tip_async().await? else {
```
krsnapaudel, line 183: Nit: `should_reconcile_l1_storage_at_startup` and `reconcile_l1_canonical_storage` both fetch the canonical tip and re-check the pre-genesis condition. Consider having the gate return the tip `(Option<(height, blkid)>)` and passing it into `reconcile_l1_canonical_storage`, so the tip is read once and the check isn't duplicated.

crates/db/store-sled/src/asm/db.rs
```diff
@@ -65,6 +67,48 @@ impl AsmDatabase for AsmDBSled {
         }
     }
 
+    fn del_asm_entries_after(&self, block: L1BlockCommitment) -> DbResult<Vec<L1BlockCommitment>> {
+        let mut blocks_to_drop = BTreeSet::new();
```
krsnapaudel, line 71: This assumes a single writer, right? A concurrent ASM append after the scan would be missed. Worth a comment somewhere?

krsnapaudel (commented):
Some comments.

crates/db/store-sled/src/asm/db.rs
```diff
@@ -65,6 +67,48 @@ impl AsmDatabase for AsmDBSled {
         }
     }
 
+    fn del_asm_entries_after(&self, block: L1BlockCommitment) -> DbResult<Vec<L1BlockCommitment>> {
```
prajwolrg, line 70: The deletion of asm entries should also rollback the ASM manifests and the manifest MMR. 

Let's add the tests to see how anything breaks if they are not deleted i.e. the mechanism is self-recovered and then see that the property we like still holds.

prajwolrg (changes requested):

bewakes (changes requested):
I agree with the goal: btcio should not start from CSM/client-state progress, and the small btcio start-height decoupling plus bitcoind warmup retry handling look worth keeping.

But the size and shape of the diff are a concern: most of the ~1k LOC change is new consensus-side startup reconciliation machinery, while the core btcio decoupling is tiny. I do not think that reconciler is the right fix. It duplicates L1 reorg handling outside btcio, blocks startup before the btcio retry loop is running, and mutates stored L1/ASM state based on startup RPC reads. The ASM pruning API also seems unnecessary.

The actual gap seems to be btcio startup after an offline L1 reorg: the reader seeds its lookback window from bitcoind’s current chain, losing stored canonical history needed to emit `RevertTo`.

I’d move repair into btcio startup: seed from stored L1 canonical entries, compare against bitcoind, and use the existing live `RevertTo` path.

storopoli:
> I agree with the goal: btcio should not start from CSM/client-state progress, and the small btcio start-height decoupling plus bitcoind warmup retry handling look worth keeping.
> 
> But the size and shape of the diff are a concern: most of the ~1k LOC change is new consensus-side startup reconciliation machinery, while the core btcio decoupling is tiny. I do not think that reconciler is the right fix. It duplicates L1 reorg handling outside btcio, blocks startup before the btcio retry loop is running, and mutates stored L1/ASM state based on startup RPC reads. The ASM pruning API also seems unnecessary.
> 
> The actual gap seems to be btcio startup after an offline L1 reorg: the reader seeds its lookback window from bitcoind’s current chain, losing stored canonical history needed to emit `RevertTo`.
> 
> I’d move repair into btcio startup: seed from stored L1 canonical entries, compare against bitcoind, and use the existing live `RevertTo` path.

Yeah makes sense, this was getting too convoluted. Thanks for the review. Closing this and will open a more clean one

PR #2070, opened 2026-07-03
fix(ee): verify chunk block state roots

storopoli:
### Motivation
- A refactor removed per-block EVM post-state root verification and only checked the final chunk tip, allowing a malicious sequencer/prover to inject a forged intermediate `state_root` and expose the forged header hash to later blocks via `BLOCKHASH`.
- The change restores per-block safety by ensuring a block's computed post-state root matches the header before that header can be canonicalized and used by subsequent execution.

### Description
- Recompute the state root after applying the `WriteBatch` and reject the block if the computed root does not match the block header by returning `EnvError::InvalidBlock` in `process_block` (file: `crates/ee-chunk-runtime/src/chunk_processing.rs`).
- Update the chunk-processing comment to document that each block's post-state root is verified before the header can be used as a parent for later blocks.
- Add a regression test `process_chunk_blocks_rejects_forged_intermediate_state_root` that constructs a two-block chunk with a forged intermediate header `state_root` and asserts the chunk is rejected (file: `crates/ee-chunk-runtime/src/chunk_processing.rs`).

### Testing
- Ran `cargo test -p strata-ee-chunk-runtime process_chunk_blocks_rejects_forged_intermediate_state_root`, which passed.
- Ran the full package test `cargo test -p strata-ee-chunk-runtime`, all tests passed (4 passed, 0 failed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a3541ff17a08331beef7991d34d4a93)

### DIFF
--- a/crates/ee-chunk-runtime/src/chunk_processing.rs
+++ b/crates/ee-chunk-runtime/src/chunk_processing.rs
@@ -35,6 +35,10 @@ pub fn process_block<E: ExecutionEnvironment>(
 
     // Merge the changes and return the outputs.
     ee.merge_write_into_state(state, exec_outp.write_batch())?;
+    let computed_state_root = state.compute_state_root()?;
+    if computed_state_root != eb.get_header().get_state_root() {
+        return Err(EnvError::InvalidBlock);
+    }
     ee.update_partial_state_after_block(state, eb.get_header())?;
 
     Ok(())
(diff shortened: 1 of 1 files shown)


sapinb (approved):
ACK 543c189d0970682eb122ffde5d190794e15ec053

The post state root computation in verify_chunk_transition becomes redundant with this. May want to optimize it later, but not blocking.

PR #2071, opened 2026-07-03
[DO NOT MERGE] New gchain framework

delbonis:
## Description

This PR adds the core functionality of the "gchain" framework, which serves to unify all different sync modes across all different chains we sync.

This PR will node include actually integrating this functionality into services, as that should be in a followup PR.  At that point these crates should be moved to the strata-common repo, so the ASM repo can use them directly.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist


crates/gchain/types/src/processor.rs
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        let sb = s.as_bytes();
+        if sb.len() > PROC_ID_LEN {
+            return Err(());
+        }
+
+        let mut inner = [0; PROC_ID_LEN];
+        inner.copy_from_slice(sb);
+        Ok(Self(inner))
```
storopoli, line 52: This panics for any id shorter than 8 bytes — `copy_from_slice` wants exact lengths, so `"asm".parse::<ProcId>()` blows up. Needs `inner[..sb.len()].copy_from_slice(sb)`.

Also `""` parses fine and gives you an all-zero id, probably want to reject that.

crates/gchain/types/src/chain_provider.rs
```diff
@@ -0,0 +1,35 @@
...
+    ) -> anyhow::Result<Vec<LinkRef<Self::Spec>>>;
+
+    /// Fetches all the known link refs that are "backwards" in the graph from
+    /// the specified node.
+    ///
+    /// MUST match the behavior of `fetch_foward_links`.
+    fn fetch_backward_links(
+        &self,
+        nref: &NodeRef<Self::Spec>,
+    ) -> anyhow::Result<LinkRef<Self::Spec>>;
```
storopoli: Should this be `Vec<LinkRef<Self::Spec>>`? The doc comment says "all the known link refs" and asks for symmetry with `fetch_forward_links`, and in the DAG model a node really can have multiple incoming links (block vs checkpoint landing on the same state) — that seems like the whole point.

crates/gchain/types/src/chain_spec.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// Checks if the node ref matches the header.
+    ///
+    /// Default impl just called `get_header_ref` and checks equality, but there
+    /// may be more optimized impls.
+    fn check_ref_match_header(lref: &Self::LinkRef, lh: &Self::LinkHeader) -> bool {
+        Self::get_header_ref(lh) == *lref
+    }
+
+    /// Gets the header of a node.
+    fn get_link_header(n: &Self::LinkRef) -> Self::LinkHeader;
```
storopoli: I don't think this signature can work — a `LinkRef` is just an identifier, you can't recover the header from it (that's what `ChainProvider::fetch_link_header` is for). Did you mean to take `&Self::Link`? The `n`/`nh` param names in this trait also look like leftovers from before the node→link rename.

crates/gchain/types/src/processor.rs
```diff
@@ -0,0 +1,218 @@
...
+        output: &Self::Artifact,
+    ) -> anyhow::Result<()>;
+
+    /// Called when we are sure we will never try to roll back to before a
+    /// certain node so that we can perform cleanups and discard information we
+    /// no longer need.
+    ///
+    /// The provided node will become the oldest node.
+    fn prune_state_upto(self, nref: &NodeRef<Self::Spec>) -> anyhow::Result<()>;
+}
```
storopoli, line 157: Taking `self` by value means the executor loses the processor after the first prune. Intentional? Feels like this wants `&mut self`.

crates/gchain/types/src/chain_spec.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    /// Gets the in-protocol canonical previous node ref from the header (such
+    /// as a "parent block"), if there is one.  There may be other completely
+    /// valid previous nodes, but this may be across sync modes.
+    // TODO(trey): do we need this fn?
+    fn get_header_canonical_prev(nh: &Self::LinkHeader) -> Option<Self::LinkRef>;
```
storopoli, line 60: Doc says "previous *node* ref" but the type is `Option<Self::LinkRef>`. Given nodes-are-states, a header's parent pointer probably does identify the previous link, so maybe it's just the doc that needs fixing — but one of the two is wrong.

crates/gchain/executor/src/artifact_cache.rs
```diff
@@ -0,0 +1,65 @@
+use std::any::TypeId;
+use std::collections::*;
+use std::sync::Arc;
+
+use strata_gchain_types::*;
+
+/// Cached output from nodes that we've extracted and determined might be useful
+/// for later proc stages.
+pub struct ArtifactCache<S: GChainSpec> {
+    nodes: HashMap<LinkRef<S>, BTreeMap<TypeId, Arc<dyn ProcArtifact>>>,
```
storopoli: Keying by `TypeId` doesn't line up with `ProcDeps`, which keys by `ProcId` — two stages emitting the same artifact type collide here, and there's no way to ask for "stage X's output" specifically. `ProcId` (or `(ProcId, TypeId)`) as the key would match the deps model.

On the `unimplemented!()` downcast below: trait upcasting is stable now, so `(arc.as_ref() as &dyn Any).downcast_ref::<A>()` should just work, no type hacks needed.

crates/gchain/executor/src/dispatcher.rs
```diff
@@ -0,0 +1,16 @@
+/// Abstraction for how we execute low-level work.
+///
+/// Will be made more general in the future to support sending off to a threadpool.
+pub trait Dispatcher {
+    /// Dispatches a work operation to execute.
+    fn dispatch(&self, work: impl FnOnce() + Send + 'static);
```
storopoli, line 6: `impl FnOnce` here makes the trait not dyn-safe, which you'll probably regret once there's an `Arc<dyn Dispatcher>` holding the threadpool impl. Taking `Box<dyn FnOnce() + Send + 'static>` instead keeps it object-safe and costs basically nothing.

storopoli (commented):
Some findings. Nice work so war. Also checked https://app.notion.com/p/Generic-chain-processor-service-37a901ba000f80f2822af2540661eed5.

crates/gchain/types/src/chain_spec.rs
delbonis: Yeah I was rushing through this, will followup with more fixes.

crates/gchain/types/src/processor.rs
delbonis: oops!

bewakes (changes requested):
I think this is too abstract at the moment.

I can see the rough direction is to model sync more generically, but I don’t understand the relationship between nodes, links, processors, artifacts, and the executor clearly enough from this PR. Because of that, it’s hard to tell what problem each abstraction is solving.

It also feels like we are jumping pretty quickly into `dyn`, shims, and type erasure before seeing one concrete flow work end to end. How many distinct chains / sync flows do we actually expect this to support? If that number is small, which I believe is small, I’m not sure we need this much runtime-generic machinery from the start. A simpler static / compile-time checked design may be enough and easier to reason about.

It would help to have a short doc or PR note explaining the motivation: what existing sync problem this solves, what the main concepts are, and how the flow works from fetching a link through processing, artifact handling, commit, rollback, and prune.

I also think this PR should include one working simple chain implementation, even if it is just a toy chain in tests. It does not need to be one of the actual chains we use. The service integration can come as a follow-up, but the framework should prove at least one path before we merge the abstraction.

crates/gchain/types/src/processor.rs
```diff
@@ -0,0 +1,218 @@
...
+            return Err(());
+        }
+
+        let sb = s.as_bytes();
+        if sb.len() > PROC_ID_LEN {
+            return Err(());
+        }
+
+        let mut inner = [0; PROC_ID_LEN];
+        inner.copy_from_slice(sb);
```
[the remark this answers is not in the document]
delbonis: Resolved.

crates/gchain/types/src/processor.rs
delbonis: Resolved.  Empty IDs are weird and I don't expect any reason to use them, but I don't *really* see a problem with it (they're opaque anyways) so I'm leaving it.

crates/gchain/types/src/lib.rs
```diff
@@ -0,0 +1,17 @@
+#![expect(missing_debug_implementations, reason = "clippy is wrong")]
+#![expect(
+    type_alias_bounds,
+    reason = "bounds enforced by type aliases using associated types"
+)]
```
storopoli: add module level docs here. A suggestion would be to briefly introduce the modules in this types crate

crates/gchain/types/src/processor.rs
```diff
@@ -0,0 +1,321 @@
...
+use std::str::{self, FromStr};
+use std::sync::Arc;
+
+use crate::chain_spec::*;
+
+const PROC_ID_LEN: usize = 8;
+
+/// ID used to refer to a registered processor stage.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct ProcId([u8; 8]);
```
storopoli, line 34: shouldn't `8` here be the `PROC_ID_LEN`?

crates/gchain/types/src/processor.rs
```diff
@@ -0,0 +1,321 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::ProcId;
+    use std::str::FromStr;
+
+    #[test]
+    fn test_parse_short_proc_id() {
+        ProcId::from_str("foo").expect("test: parse ProcId");
+    }
```
storopoli, line 320: we need a test to validate the `ProcId` from empty strings. That should produce an error no? Making impossible to create one from empty strings...

crates/gchain/types/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "strata-gchain-types"
+version = "0.1.0"
```
storopoli, line 3: we should probably bump this to `0.3.0` to match all the `alpen` repo crates

crates/ol/gchain-impl/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "strata-ol-gchain-impl"
+version = "0.1.0"
```
storopoli, line 3: we should probably bump this to `0.3.0` to match all the `alpen` repo crates

crates/gchain/executor/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "strata-gchain-executor"
+version = "0.1.0"
```
storopoli, line 3: we should probably bump this to `0.3.0` to match all the `alpen` repo crates

crates/ol/gchain-impl/src/graph_types.rs
```diff
@@ -0,0 +1,114 @@
...
+
+impl OLStateNode {
+    pub fn epoch(&self) -> Epoch {
+        self.epoch.into()
+    }
+}
+
+// actually fine that these are the same here, for now
+impl GNodeRef for OLStateNode {}
+impl GNode for OLStateNode {}
```
storopoli, line 20: Yes actually but please add a reason to the comment. Something like, as long as the `OLStateNode` is `Copy` or something.

crates/gchain/executor/src/process.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    fn process_link(
+        &self,
+        lref: &LinkRef<S>,
+        link: &Link<S>,
+        cache: &ArtifactCache<S>,
+        prev_lref: Option<&LinkRef<S>>,
+    ) -> anyhow::Result<Arc<dyn DynProcArtifact>> {
+        let ctx = ProcContextImpl::<P>::new(cache, *lref, prev_lref.copied());
```
storopoli, line 94: `*lref` and `prev_lref.copied()` require `Copy`, but `GLinkRef` only guarantees `Clone`

crates/gchain/types/src/chain_spec.rs
```diff
@@ -0,0 +1,138 @@
...
+
+pub type NodeRef<S: GChainSpec> = <S as GChainSpec>::NodeRef;
+pub type Node<S: GChainSpec> = <S as GChainSpec>::Node;
+pub type LinkTable<S: GChainSpec> = HashMap<NodeRef<S>, Node<S>>;
+pub type LinkRef<S: GChainSpec> = <S as GChainSpec>::LinkRef;
+pub type LinkHeader<S: GChainSpec> = <S as GChainSpec>::LinkHeader;
+pub type Link<S: GChainSpec> = <S as GChainSpec>::Link;
+
+/// Toplevel trait describing a chain.
+pub trait GChainSpec {
```
storopoli, line 29: There's no way to map a link to its endpoint node refs. `GChainSpec` only produces link refs (`get_header_ref`, `get_header_canonical_prev`), and `ChainProvider` has no `fetch_node` or `link -> (origin, target)` query. This is the mechanism that makes the sync graph a graph: without it the executor can never detect that a checkpoint step and a block path converge, `LinkPath::try_push_link` can't ever assert canonicality, and `on_init(&NodeRef, &Node)` is uncallable since no `Node` is obtainable anywhere.

crates/gchain/executor/src/context.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use strata_gchain_types::*;
+
+use crate::artifact_cache::ArtifactCache;
+
+/// Context trait for a chain executor.  This should only be used by one
+/// executor at a time.
+pub trait ExecutorContext {
+    fn store_processor_output(&self, id: ProcId) -> anyhow::Result<()>;
+}
```
storopoli, line 13: Executor-owned persistence isn't expressible in the current types. The design doc puts artifact persistence on the executor, but `ProcArtifact` has `from_buf` with no "encode" counterpart, `ProcessorArtifactData` has private fields and no constructor (only reachable via `Deserialize`), and `ExecutorContext::store_processor_output(&self, id: ProcId)` takes no artifact at all. Decode-only round-tripping is a signature problem, not a missing body.

crates/gchain/executor/src/dispatcher.rs
storopoli: Bumping this

crates/gchain/types/src/processor.rs
```diff
@@ -0,0 +1,321 @@
...
+
+impl AsRef<str> for ProcId {
+    fn as_ref(&self) -> &str {
+        let idx = self
+            .0
+            .iter()
+            .enumerate()
+            .find_map(|(i, b)| (*b == 0).then(|| i))
+            .unwrap_or(PROC_ID_LEN);
+        unsafe { str::from_utf8_unchecked(&self.0[..idx]) }
```
storopoli, line 64: Missing `// SAFETY:` comment.

A later `derive(Deserialize)` would silently make it UB in a crate slated to be shared. Safe `from_utf8(...).expect(...)` costs nothing here. Also `ProcId::from_str("")` succeeds and displays as empty.

crates/gchain/executor/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+edition = "2024"
+
+[dependencies]
+strata-gchain-types.workspace = true
+
+anyhow.workspace = true
```
storopoli, line 9: I really don't like `anyhow` here and in the `types` crate.

For a framework crate destined for `strata-common`/`asm` repos, this forces `anyhow` on every implementor and denies the executor the ability to distinguish transient IO failure from processing failure. Repo convention is `thiserror` enums for library errors.

crates/gchain/executor/src/linear_executor.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn canonical_order(&self) -> &[ProcId] {
+        &self.canonical_order
+    }
+
+    /// Returns an iterator over the stages in canonical order.
+    fn iter_stages(&self) -> impl Iterator<Item = &Stage<S>> {
+        self.canonical_order()
+            .iter()
+            .map(|id| self.stages.get(id).unwrap())
+    }
```
storopoli, line 31: `stages: BTreeMap` + `canonical_order: Vec` with an `unwrap` joining them.
This is bad, we have two structures that must agree with nothing enforcing it. A `Vec<Stage>` in canonical order plus an index map removes the `unwrap` structurally.

crates/gchain/types/src/chain_spec.rs
```diff
@@ -0,0 +1,138 @@
...
+//! Nodes correspond to "at rest" states.  Blocks and checkpoints are different
+//! types of state transitions forming links between nodes.
+
+use std::collections::*;
+use std::fmt::Debug;
+use std::hash::Hash;
+
+pub type NodeRef<S: GChainSpec> = <S as GChainSpec>::NodeRef;
+pub type Node<S: GChainSpec> = <S as GChainSpec>::Node;
+pub type LinkTable<S: GChainSpec> = HashMap<NodeRef<S>, Node<S>>;
```
storopoli, line 23: `LinkTable` holds full nodes in memory (contrary to "nodes don't fit in memory"), contains no links despite the name, and is unused; delete it. `ArtifactCache` and `ProcHistory` are unbounded with nothing invoking eviction; the doc's "a few edges in memory" contract needs a stated budget somewhere.

crates/gchain/types/src/processor.rs
```diff
@@ -0,0 +1,321 @@
...
+        f.write_str(self.as_ref())
+    }
+}
+
+/// Generic chain processor stage.
+///
+/// Error variants on result types should *ONLY* be used to indicate that the
+/// processing *failed*, never that the node is invalid.  Nodes being invalid
+/// should be indicated through [`ProcArtifact::is_link_valid`].
+pub trait GChainProc: Sized + 'static {
```
storopoli, line 85: should we add a `type Version: ProcVersion` here?

`ProcessorArtifactData::exec_version()` exists, but processors have no method to report a version and nothing constructs the tracking data.

storopoli (changes requested):

crates/gchain/types/src/processor.rs
delbonis: Yeah I don't like the idea of empty string proc IDs.

delbonis:
Rebased onto current `main`.

delbonis:
Closing pull request because there's a bunch of stale stuff here.

PR #2072, opened 2026-07-06
STR-3186: Add OL DA checkpoint payload byte-parity test

bewakes:
## Description

This PR adds a functional test proving the OL DA bytes posted in checkpoint publication match the sequencer's locally encoded StateDiff bytes.

- `test_checkpoint_da_parity`: extracts the SSZ `CheckpointPayload` from the posted L1 envelope tx (anchored via the checkpoint's L1 reference txid) and asserts the inner `sidecar.ol_state_diff` bytes equal the SignCheckpoint duty payload's, with `new_tip` equality anchoring both to the same epoch.
- Python-side SSZ offset parsing is pinned by a fixture generated from the real `CheckpointPayload::as_ssz_bytes()` encoding (ignored test in `strata-test-utils-l2`).
- Replay coverage: `test_checkpoint_sync_node` now also asserts the reconstructed epoch commitment matches the published checkpoint's `l2_range` terminal instead of  a separate env spin-up.

**AI usage**: This was created with the use of codex and claude, which are good at writing tests and doing these byte-level machinaries.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist


crates/test-utils/l2/src/checkpoint.rs
```diff
@@ -331,3 +331,57 @@ impl CheckpointTestHarness {
...
+
+    /// Prints the fixture pinning the Python SSZ offset parsing in
+    /// `functional-tests/tests/checkpoint/checkpoint_payload_fixture.json`.
+    ///
+    /// Regenerate with:
+    /// `cargo test -p strata-test-utils-l2 dump_checkpoint_payload_fixture -- --ignored
+    /// --nocapture`
+    #[test]
+    #[ignore = "fixture generator, run manually"]
+    fn dump_checkpoint_payload_fixture() {
```
storopoli: This should be an example maybe?

storopoli (approved):
ACK beb188fb2fcadd7cee0b251dc3d5bb3cba154762

functional-tests/tests/checkpoint/test_checkpoint_da_parity.py
```diff
@@ -0,0 +1,91 @@
...
+        )
+
+        # Wait until this epoch's checkpoint is posted and finalized on L1.
+        mine_until_finalized_epoch(
+            bitcoin=bitcoin, strata=strata, strata_rpc=strata_rpc, target_epoch=epoch
+        )
+
+        # Anchor to the exact posted tx via the node's L1 reference rather than a
+        # broad tag scan, which could pick up a rejected checkpoint candidate.
+        info = strata_rpc.call("strata_getCheckpointInfo", epoch)
```
delbonis: Wait, why is this invoking the low-level `call` fn and not just doing `strata_rpc.strata_getCheckpointInfo` through the `__getattr__` tricks?

functional-tests/tests/strata/test_checkpoint_sync_node.py
```diff
@@ -115,6 +117,25 @@ def check_summaries_equivalent(seq_summary: AccountEpochSummary, node_summary: A
         assert su == nu
 
 
+def check_commitment_matches_checkpoint(seq_rpc, epoch: int, commitment: dict):
+    """Anchors the reconstructed epoch commitment to the published checkpoint.
+
+    The terminal blkid hashes the reconstructed header (which commits to
+    state_root), so equality proves replay yielded the expected post-state.
+    """
+    info = seq_rpc.call("strata_getCheckpointInfo", epoch)
```
delbonis: same here

crates/test-utils/l2/src/checkpoint.rs
delbonis: Or a separate tool entirely?  That's probably overkill, but doing it as a `#[ignore]`d test like this is definitely not the right approach.

delbonis (changes requested):
Maybe we should add a CLAUDE.md or more docs to the fntest support libraries.

bewakes:
@storopoli @delbonis example is very easy to drift and a separate tool is indeed an overkill.

I've converted the ignored test into a non ignored golden test and also fixed the raw rpc calls in functional tests in https://github.com/alpenlabs/alpen/pull/2072/commits/ca1fbf7af3b501c98f475f28f777b47e7ddd2af1

delbonis (approved):

PR #2074, opened 2026-07-06
fix(btcio): decouple reader startup from client state and repair offline L1 reorgs

storopoli:
## Description

The btcio reader's startup depended on the downstream CSM/client-state watermark, coupling the upstream L1 reader to one downstream consumer ([STR-3845](https://alpenlabs.atlassian.net/browse/STR-3845)). Startup also seeded the reader's lookback window from `bitcoind`'s *current* chain, so an L1 reorg that happened while the node was offline silently discarded stored canonical history - no `RevertTo` was ever emitted and stored L1/ASM/CSM state stayed on a stale fork.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- **Seed startup from stored canonical entries**: when a stored L1 canonical tip exists, `init_reader_state()` seeds the recent-blocks lookback queue from stored canonical entries (`get_canonical_blockid_range_async`) with `next_height = stored_tip + 1`. The first `poll_for_new_blocks()` iteration then detects any offline reorg via the existing `find_pivot_block()` comparison against bitcoind and emits `L1Event::RevertTo` through the normal handler. Startup never mutates stored L1 state. A fresh DB still seeds from bitcoind at the genesis/anchor height.
- **Start-height decoupling** (ported from #2069): `calculate_target_next_block()` no longer reads client state; it uses only the stored L1 canonical tip with a genesis fallback.
- **Bitcoind warmup retry** (ported from #2069): RPC `-28` (warmup) is classified as retryable in btcio and defers node startup checks instead of failing them.
- **Shorter-bitcoind-chain robustness**: with stored seeding the tracked tip can exceed bitcoind's tip. The pivot search skips `-8 Block height out of range` responses; a matching-prefix bitcoind (merely behind/syncing) yields no events and no rollback; only an actual hash divergence at or below bitcoind's tip triggers `RevertTo`. Deep reorgs beyond the lookback window still surface as `PivotNotFound`.

The seven commits are independent and reviewable in order:

1. warmup classification in btcio
2. startup-check deferral
3. start-height decoupling
4. stored-canonical seeding (the core fix)
5. pivot-search robustness for a shorter `bitcoind` chain
6. wait for a deeply lagging `bitcoind` that is a matching prefix of the stored chain (addresses [review feedback](https://github.com/alpenlabs/alpen/pull/2074#discussion_r3529289755))
7. honor the genesis clamp when seeding from the stored chain (addresses [review feedback](https://github.com/alpenlabs/alpen/pull/2074#discussion_r3529554107)).

- [ ]  Yes
- [x]  No

## Checklist

> **AI assistance notice**: this PR was implemented with Codex (GPT-5.5) orchestrated by Claude Code; all changes were human-reviewed.

## Related Issues

STR-3845

Supersedes #2069

[STR-3845]: https://alpenlabs.atlassian.net/browse/STR-3845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/btcio/src/reader/query.rs
+++ b/crates/btcio/src/reader/query.rs
@@ -227,52 +227,65 @@ async fn init_reader_state<R: Reader>(
         );
     }
 
+    let stored_canonical_tip = ctx.storage.l1().get_canonical_chain_tip_async().await?;
     let mut real_cur_height;
-    if let Some((stored_tip_height, _stored_tip_blkid)) =
-        ctx.storage.l1().get_canonical_chain_tip_async().await?
-    {
-        let start_height = stored_tip_height.saturating_sub(lookback);
-        let stored_blockids = ctx
-            .storage
-            .l1()
-            .get_canonical_blockid_range_async(start_height, stored_tip_height + 1)
-            .await?;
-        debug!(
-            %start_height,
-            end_height = %stored_tip_height,
-            entries = stored_blockids.len(),
-            "loaded reader init range from stored L1 canonical chain"
-        );
-
-        for blockid in stored_blockids {
-            init_queue.push_back(blockid.to_block_hash());
-        }
-
-        if init_queue.is_empty() {
-            bail!(
-                "btcio: stored L1 canonical chain tip at height {stored_tip_height} \
-                 was missing from the reader startup range"
+    match stored_canonical_tip {
+        Some((stored_tip_height, _stored_tip_blkid))
+            if stored_tip_height.saturating_add(1) >= target_next_block =>
+        {
+            let start_height = stored_tip_height.saturating_sub(lookback);
+            let stored_blockids = ctx
+                .storage
+                .l1()
+                .get_canonical_blockid_range_async(start_height, stored_tip_height + 1)
+                .await?;
+            debug!(
+                %start_height,
+                end_height = %stored_tip_height,
+                entries = stored_blockids.len(),
+                "loaded reader init range from stored L1 canonical chain"
             );
+
+            for blockid in stored_blockids {
+                init_queue.push_back(blockid.to_block_hash());
+            }
+
+            if init_queue.is_empty() {
+                bail!(
+                    "btcio: stored L1 canonical chain tip at height {stored_tip_height} \
+                     was missing from the reader startup range"
+                );
+            }
+
+            real_cur_height = stored_tip_height;
         }
+        stored_tip => {
+            if let Some((stored_tip_height, _)) = stored_tip {
+                info!(
+                    %stored_tip_height,
+                    %target_next_block,
+                    %genesis_height,
+                    "stored L1 canonical tip is below configured genesis; seeding reader from Bitcoin client"
+                );
+            }
 
-        real_cur_height = stored_tip_height;
-    } else {
-        let chain_tip = chain_info.blocks as L1Height;
-        let start_height = target_next_block
-            .saturating_sub(lookback)
-            .max(pre_genesis)
-            .min(chain_tip);
-        let end_height = chain_tip.min(pre_genesis.max(target_next_block.saturating_sub(1)));
-        debug!(%start_height, %end_height, "queried L1 client, have init range");
+            let chain_tip = chain_info.blocks as L1Height;
+            let start_height = target_next_block
+                .saturating_sub(lookback)
+                .max(pre_genesis)
+                .min(chain_tip);
+            let end_height = chain_tip.min(pre_genesis.max(target_next_block.saturating_sub(1)));
+            debug!(%start_height, %end_height, "queried L1 client, have init range");
 
-        // Loop through the range we've determined to be okay and pull the blocks we want to look
-        // back through in.
-        real_cur_height = start_height;
-        for height in start_height..=end_height {
-            let blkid = client.get_block_hash(height as u64).await?;
-            debug!(%height, %blkid, "loaded recent L1 block");
-            init_queue.push_back(blkid);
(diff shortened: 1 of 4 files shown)


storopoli:
@codex review

crates/btcio/src/reader/query.rs
```diff
@@ -370,7 +375,19 @@ async fn find_pivot_block(
...
+        let queried_l1blkid = match client.get_block_hash(height as u64).await {
+            Ok(block_hash) => block_hash,
+            Err(err) if is_missing_block_height_error(&err) => {
+                trace!(
+                    %height,
+                    %l1blkid,
+                    err = %err,
+                    "Bitcoin client does not have tracked block height while finding pivot"
+                );
+                continue;
```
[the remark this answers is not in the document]
storopoli: Addressed in c886b34fd: when `find_pivot_block()` finds no pivot and bitcoind's tip is below the lowest tracked height, `poll_for_new_blocks()` now compares bitcoind's tip hash against the stored canonical blockid at that height. A match (deeply lagging same-chain prefix) waits with no events instead of erroring; a mismatch or missing stored entry keeps the `PivotNotFound` behavior. Covered by two new regression tests.

crates/btcio/src/reader/query.rs
```diff
@@ -264,22 +227,53 @@ async fn init_reader_state<R: Reader>(
...
-    let start_height = target_next_block
-        .saturating_sub(lookback)
-        .max(pre_genesis)
-        .min(chain_tip);
-    let end_height = chain_tip.min(pre_genesis.max(target_next_block.saturating_sub(1)));
-    debug!(%start_height, %end_height, "queried L1 client, have init range");
+    let mut real_cur_height;
+    if let Some((stored_tip_height, _stored_tip_blkid)) =
+        ctx.storage.l1().get_canonical_chain_tip_async().await?
+    {
```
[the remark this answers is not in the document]
storopoli: Addressed in 13d0ade76: stored-canonical seeding is now used only when `stored_tip + 1 >= target_next_block` (the genesis-clamped target). A stored tip below the configured genesis falls through to the bitcoind-seeded path, which already clamps the init range to genesis — restoring the pre-refactor clamp behavior. Covered by a regression test (`init_reader_state_ignores_pregenesis_stored_tip`).

crates/btcio/src/reader/query.rs
```diff
@@ -264,22 +227,66 @@ async fn init_reader_state<R: Reader>(
...
+    let mut real_cur_height;
+    match stored_canonical_tip {
+        Some((stored_tip_height, _stored_tip_blkid))
+            if stored_tip_height.saturating_add(1) >= target_next_block =>
+        {
+            let start_height = stored_tip_height.saturating_sub(lookback);
+            let stored_blockids = ctx
+                .storage
+                .l1()
+                .get_canonical_blockid_range_async(start_height, stored_tip_height + 1)
```
krsnapaudel: `get_canonical_blockid_range` skips missing heights silently (`if let Some(id) = get(h) { push }`) and returns a shorter `Vec` — it doesn't guarantee contiguity.

Should we assert continguity?

crates/btcio/src/reader/query.rs
krsnapaudel, line 144: Seems we log errors twice here for some cases. `error!` inside `poll_for_new_blocks` and `warn!` inside `handle_poll_error`, but these are preexisting.

krsnapaudel (approved):
Some nits.

bewakes (changes requested):
One thing I think should be fixed before merge: the stored-chain seeding path assumes `get_canonical_blockid_range_async()` returns a contiguous range, but sled currently skips missing heights. If there’s a hole, `ReaderState` can end up with hashes aligned to the wrong heights. Can we enforce contiguity there or validate the returned range before constructing the state?

Also worth checking: if bitcoind is still catching up and `getblockhash(anchor_height)` returns `-8`, startup still seems to hard-fail even though the poll path now handles a lagging bitcoind. That may need to defer instead.

Non-blocking readability note: `init_reader_state()` has grown quite a bit. Splitting the stored-chain and bitcoind- seeded init paths into smaller helpers would make the startup invariants easier to follow.

crates/btcio/src/reader/query.rs
storopoli: Check 4f81347

crates/btcio/src/reader/query.rs
storopoli: Check 0924a18

storopoli:
> One thing I think should be fixed before merge: the stored-chain seeding path assumes `get_canonical_blockid_range_async()` returns a contiguous range, but sled currently skips missing heights. If there’s a hole, `ReaderState` can end up with hashes aligned to the wrong heights. Can we enforce contiguity there or validate the returned range before constructing the state?
> 
> Also worth checking: if bitcoind is still catching up and `getblockhash(anchor_height)` returns `-8`, startup still seems to hard-fail even though the poll path now handles a lagging bitcoind. That may need to defer instead.
> 
> Non-blocking readability note: `init_reader_state()` has grown quite a bit. Splitting the stored-chain and bitcoind- seeded init paths into smaller helpers would make the startup invariants easier to follow.

check the recent commits, c1b22d3 and 0924a18

bewakes (approved):

PR #2076, opened 2026-07-06
STR 3846 Reconcile MMR index DB against OL chain state on startup

krsnapaudel:
## Description

Adds shared OL MMR index comparison/reconcile planning and wires it into both `strata-dbtool` and `strata` startup so OL-owned MMR indexes stay consistent with persisted OL state after reverts, reorgs, interrupted writes, and promoted checkpoint-sync startup.

### 1. Shared OL MMR Index Crate

Introduces `strata-ol-mmr-index` (`crates/ol/mmr-index`) as the shared home for OL MMR index semantics:

- `MmrIndexEntry` represents a persisted MMR namespace and its current native `Mmr64` state.
- `find_divergent_ol_mmr_indexes` compares persisted index state against an OL target state.
- `DivergentOLMmrIndex` reports OL-owned divergences: ahead indexes, behind indexes, and same-count state mismatches.
- `build_mmr_index_reconcile_plan` converts comparison results into safe truncation work.
- `MmrIndexReconcilePlan::to_report` derives the reconciliation report used by startup reconciliation.

The crate owns comparison and planning only. Storage mutation, prefix preflight, truncation, reporting, and related-index cleanup stay with the caller.

### 2. `strata-dbtool` Revert Path

Updates `revert-ol-state` to use the shared OL MMR index plan. The dbtool path now reads persisted MMR namespaces and states, builds shared `MmrIndexEntry` values, plans reconciliation, preflights prefix compatibility, and truncates ahead indexes through the storage manager.

### 3. `strata` Startup Reconciliation

Adds startup reconciliation for OL-owned MMR indexes through `strata-consensus-logic`, with concrete storage adapters in `strata`.

FCM uses a dedicated `FcmStartupReconciler` trait for startup MMR reconciliation, while `FcmStorage` supplies only the FCM storage facts needed to select the target and history base.

On startup, `strata` now:
- seeds the L1 block refs genesis sentinel before comparison,
- reconciles checkpoint-sync nodes to the latest summarized OL state,
- reconciles sequencer/full-node startup to the FCM-selected tip after canonical index repair,
- handles promoted checkpoint-sync datadirs by stopping target-epoch history walks at the stored history base,
- removes rejected block-attributed OL state-indexing rows with exact block deletion,
- compares the MMR index DB against the selected persisted OL state,
- truncates ahead OL-owned indexes,
- rejects behind indexes, same-count state mismatches, and non-prefix ahead indexes,
- rolls OL state indexing back to the selected target.

### 4. Tests

Adds focused unit and functional coverage for divergence detection, reconcile planning, sentinel validation, prefix rejection, dbtool truncation, exact rejected-indexing cleanup, startup target selection, startup truncation/rejection paths, genesis prefill behavior, promoted history-base startup, and restart recovery after interrupted MMR index writes.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-3846](https://alpenlabs.atlassian.net/browse/STR-3846)

[STR-3846]: https://alpenlabs.atlassian.net/browse/STR-3846?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


krsnapaudel:
@codex review

krsnapaudel:
@codex review

prajwolrg (approved):
LGTM!

storopoli (approved):
ACK cdd5e22a9767fa7e80b02f02a65ee77366f616eb

bin/strata/src/mmr_reconcile.rs
```diff
@@ -0,0 +1,894 @@
...
+}
+
+/// Formats a raw namespace key as lowercase hex.
+fn format_raw_mmr_id(raw_mmr_id: &RawMmrId) -> String {
+    let mut out = String::with_capacity(raw_mmr_id.len() * 2);
+    for byte in raw_mmr_id {
+        write!(&mut out, "{byte:02x}").expect("write to String should not fail");
+    }
+    out
+}
```
delbonis: `hex::encode`?

bin/strata/src/mmr_reconcile.rs
delbonis, line 1: This *really* should not be directly in the Strata binary crate.  I would assume that it'd be part of the chain worker crate or alongside the FCM modules.

This pattern of stuffing init logic that doesn't have somewhere else to go into the Strata client crate directly needs to stop, it's becoming unmaintainable.

bin/strata/src/mmr_reconcile.rs
```diff
@@ -0,0 +1,894 @@
...
+            target.leaf_count
+        );
+    }
+
+    if final_state.peaks != target.peaks {
+        bail!(
+            "MMR {} final peaks do not match target state at leaf count {}",
+            display_ol_mmr_id(divergent_index.mmr_id()),
+            final_leaf_count
+        );
```
delbonis: Typed errors.

bin/strata/src/mmr_reconcile.rs
```diff
@@ -0,0 +1,894 @@
...
+use tracing::{debug, info};
+
+use crate::genesis::prefill_l1_block_refs_mmr_async;
+
+/// Source used to resolve the local OL state reconciliation target.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum OLMmrReconcileTargetSource {
+    CanonicalBlockTip,
+    EpochSummary,
+}
```
delbonis: This name is confusing, what's a `TargetSource`?  Isn't this just describing the MMR that we're trying to reconcile?

bin/strata/src/services.rs
```diff
@@ -271,9 +272,17 @@ pub(crate) fn start_strata_services(
...
+    reconcile_unaccepted_checkpoint_artifacts(&nodectx)?;
+    if !is_sequencer {
+        nodectx
+            .task_manager()
+            .handle()
+            .block_on(reconcile_ol_mmr_index_for_checkpoint_sync(
+                nodectx.storage().as_ref(),
+                nodectx.ol_params(),
+            ))?;
+    }
```
delbonis: This checks if we're not a sequencer and *then* does something for checkpoint sync.  These are distinct concepts.  We should explicitly be checking if we're in checkpoint sync mode instead of using "if we're a sequencer" to be a proxy for that.  It's totally possible to be a non-sequencing full node which would have to reset the dangling MMR entries like this.

delbonis (changes requested):
This PR does a lot of stuff directly working with the database handling instead of that being hidden behind a context trait.  That should be changed so we can hope to have more modular and reusable components, none of the code in this PR is able to be naturally adapted to fit a new scenario.

I realize that part of these restrictions is driven by how the bookkeeping around the MMR indexing works.  There's not a lot we can change without a much deeper refactor that this PR isn't the place for.  I hope that this can be replaced with the gchain system I'm cooking up in #2071 but that's out of scope for now.

bin/strata/src/mmr_reconcile.rs
krsnapaudel: @delbonis I realized this is not needed. Removing.

bin/strata/src/services.rs
krsnapaudel: @delbonis Moving the wiring through node role specific (css or fcm) context.

krsnapaudel:
> This PR does a lot of stuff directly working with the database handling instead of that being hidden behind a context trait. That should be changed so we can hope to have more modular and reusable components, none of the code in this PR is able to be naturally adapted to fit a new scenario.
> 
> I realize that part of these restrictions is driven by how the bookkeeping around the MMR indexing works. There's not a lot we can change without a much deeper refactor that this PR isn't the place for. I hope that this can be replaced with the gchain system I'm cooking up in #2071 but that's out of scope for now.

@delbonis I tried to clean up and reduce code duplication. PTAL.

crates/chain-worker/src/mmr_prefill.rs
delbonis, line 1: This module probably shouldn't be using the low-level `DbError` type should it?  This has to try to shove downstream error conditions into its model, which doesn't make sense.

crates/consensus-logic/src/ol_mmr_reconcile/mod.rs
```diff
@@ -0,0 +1,265 @@
...
+
+impl MmrIndexReconcileReport {
+    fn from_plan(plan: &MmrIndexReconcilePlan) -> Self {
+        Self {
+            inspected: plan.inspected(),
+            asm_owned_skipped: plan.asm_owned_skipped(),
+            indexes_truncated: plan.truncation_count(),
+            leaves_removed: plan.leaves_to_remove_count(),
+        }
+    }
```
delbonis: Probably makes more sense to be a `to_report` on the plan type.

crates/ol/mmr-index/src/mmr_divergence.rs
```diff
@@ -0,0 +1,437 @@
...
+
+    /// The index has the target count but different peaks.
+    PeaksMismatch {
+        /// Target state.
+        target: MmrStateView,
+
+        /// Persisted index peaks.
+        index_peaks: Vec<Hash>,
+    },
+}
```
delbonis: Should probably split out all of these arms as their own structs and make the variants be newtypes.  This way in `MmrIndexTruncation` we can just accept the struct as the arg and not return errors on the other variants.

bin/strata/src/fcm.rs
```diff
@@ -167,6 +175,14 @@ impl FcmStorage for StrataFcmContext {
             .get_canonical_epoch_commitment_at_async(epoch)
             .await
     }
+
+    async fn reconcile_ol_mmr_index(&self, target: OLMmrReconcileTarget) -> anyhow::Result<()> {
+        let mmr_reconcile_ctx =
+            StrataMmrReconcileCtx::new(Arc::clone(&self.storage), Arc::clone(&self.ol_params));
+
```
delbonis: `.clone()`

I keep seeing this pattern in different places, is this Codex getting confused by thinking it needs to be called like C++'s `std::shared_ptr`?  I never see Claude do this.

crates/consensus-logic/src/ol_mmr_reconcile/mod.rs
delbonis, line 1: Move all of these type/fn items into their own module(s).

crates/db/types/src/mmr_index.rs
```diff
@@ -59,6 +59,16 @@ pub fn num_leaves_to_mmr_size(leaves_count: u64) -> u64 {
...
 
+/// Read-only view of MMR state at a specific leaf count.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MmrStateView {
+    /// Number of leaves represented by this view.
+    pub leaf_count: u64,
+
+    /// MMR peaks in storage order.
+    pub peaks: Vec<Hash>,
+}
```
delbonis: I'm wondering if this should actually use native mmr types?  Or at least (optionally) implement the MmrStore trait?

crates/ol/mmr-index/src/mmr_divergence.rs
delbonis, line 1: This module *MUST NOT* accept `OLState` directly since the concept it represents will be much more restricted in the future.  It should *at least* accept `&impl IStateAccessor` so that we can leverage those abstractions.

crates/ol/mmr-index/src/mmr_reconcile.rs
```diff
@@ -0,0 +1,242 @@
...
+    asm_owned_skipped: u64,
+    truncations: Vec<MmrIndexTruncation>,
+}
+
+impl MmrIndexReconcilePlan {
+    /// Constructs a validated plan from persisted MMR index entries.
+    pub fn new(
+        target_state: &OLState,
+        entries: Vec<MmrIndexEntry>,
+    ) -> Result<Self, OLMmrIndexError> {
```
delbonis: This is *very* complex logic to put directly in a constructor.  A `new` fn should restrict itself to basic sanity checks and mostly just directly 1:1 map its arguments.  The way this is written also makes it more awkward to use in tests.

Also, see above comment about `&impl IStateAccessor`.

crates/storage/src/managers/mmr_algorithm.rs
```diff
@@ -132,6 +153,24 @@ pub(crate) fn compute_pop_plan(leaf_count: u64, table: &NodeTable) -> DbResult<O
     }))
 }
 
+pub(crate) fn compute_truncate_plan(
+    leaf_count: u64,
+    target_leaf_count: u64,
+    table: &NodeTable,
+) -> DbResult<TruncatePlan> {
```
delbonis: Maybe this could just be a fn on `NodeTable`?

And again, `DbError` feels like the wrong error type to be using in *this module* specifically.  Does this logic even make sense to directly be in this crate?  It seems mostly completely detached from the storage logic.

crates/storage/src/managers/mmr_algorithm.rs
```diff
@@ -81,6 +88,20 @@ pub(crate) fn compute_pop_fetch_positions(leaf_count: u64) -> Vec<NodePos> {
...
+    target_leaf_count: u64,
+) -> DbResult<Vec<NodePos>> {
+    if target_leaf_count > leaf_count {
+        return Err(DbError::MmrIndexOutOfRange {
+            requested: target_leaf_count,
+            cur: leaf_count,
+        });
+    }
+
+    Ok(iter_prune_after_positions(target_leaf_count, leaf_count).collect())
```
delbonis: This fn is weird, it just does a check and then forwards its arguments to `iter_prune_after_positions` and collects the result.

Similar thing about the errors.  My understanding was that that error variant was supposed to be returned by the database layer rather than something synthetically produced by the algorithms, but it's being *misused* to signal some internal inconsistency within the component.

delbonis (changes requested):
Unfortunately there's still a lot of organizational issues in this PR.

crates/chain-worker/src/mmr_prefill.rs
krsnapaudel: @delbonis This is a refactor. I guess we can turn the condition into an assert.

crates/db/types/src/mmr_index.rs
krsnapaudel: @delbonis This is a refactor. Do you mean to reuse `Mmr64` type? This seems to have other non-trivial implications.

I guess you mean `MmrState`, not `MmrStore`. Why not implement `MmrState` for `MmrStateView`:
`MmrState::iter_peaks()` is defined as lowest-height to highest-height.
`MmrStateView.peaks` is intentionally storage order, left-to-right.
Implementing `MmrState` would make a storage-order type pretend to be a native compact-MMR state. That is exactly the ordering ambiguity we’re trying to keep explicit.

crates/storage/src/managers/mmr_algorithm.rs
krsnapaudel: @delbonis I think this should stay in `strata-storage`. `NodeTable` is a DB type, while this helper is storage-manager mutation planning over a `NodeTable` snapshot. Moving it there would make `db/types` own MMR truncate planning before there is another caller that needs that API.

I agree with the error-boundary point, though. I’ll stop returning `DbError` from this pure algorithm module.

crates/storage/src/managers/mmr_algorithm.rs
krsnapaudel: @delbonis  Good point. The range check is redundant because `MmrIndexHandle` rejects `target_leaf_count > leaf_count` before calling this helper, so I’ll make `compute_truncate_fetch_positions` infallible and keep only a `debug_assert`.

I’ll still keep the helper itself: it names the truncate prefetch phase, which runs before the `NodeTable` exists. `compute_truncate_plan` then consumes that prefetched table to build the mutation plan.

krsnapaudel:
> Unfortunately there's still a lot of organizational issues in this PR.

@delbonis Addressed comments except for the `DbError`s in MMR algorithm. I agree that synthesizing `DbError` in the algorithm is not great. Cleanly splitting algorithm errors from DB errors is more involved than it looks: it means deciding what's a pure `MmrError` (bounds/structure from the inputs) vs a storage concern (a missing node is DB content, caught at the boundary — not an algorithm condition), applying that across append/pop/proof, and collapsing `DbError`'s flat `Mmr*` variants behind `Mmr(MmrError)`. I have created a [ticket for it](https://alpenlabs.atlassian.net/browse/STR-3979).

krsnapaudel:
@codex review

krsnapaudel:
@codex review

crates/db/types/src/mmr_index.rs
delbonis: Yes `MmrState`, used the wrong word.

But I don't understand that justification.  If it's an ordering thing we can make explicit in the interface.  This is part of why I avoid making fields `pub`, since it makes even interfaces on containers designed in a way that adding indirection is easy.

> That is exactly the ordering ambiguity we’re trying to keep explicit.

Is this an LLM response?

crates/chain-worker/src/mmr_prefill.rs
delbonis: Are these asserts reachable in normal conditions or do they require some nearby logic bug to trigger?

But yes that would be a refactor.

crates/ol/mmr-index/src/mmr_divergence.rs
```diff
@@ -0,0 +1,608 @@
...
+/// Adds missing OL-owned namespaces required by the target state.
+fn add_missing_target_ol_namespaces<S, I>(
+    entries: &mut Vec<MmrIndexEntry>,
+    target_state: &S,
+    found_namespaces: &HashSet<MmrId>,
+    target_snark_accounts: I,
+) -> Result<(), OLMmrIndexError>
+where
+    S: IStateAccessor,
+    I: IntoIterator<Item = AccountId>,
```
delbonis: Can just do `target_state: &impl IStateAccessor` and `target_snark_accounts: impl IntoIterator<Item = AccountId>`.

Same issue in `find_divergent_ol_mmr_indexes`.

crates/ol/mmr-index/src/mmr_divergence.rs
```diff
@@ -0,0 +1,608 @@
...
+/// Adds one missing namespace as an empty persisted index.
+fn add_missing_target_namespace(entries: &mut Vec<MmrIndexEntry>, mmr_id: MmrId) {
+    entries.push(MmrIndexEntry {
+        mmr_id,
+        state: MmrStateView {
+            leaf_count: 0,
+            peaks: Vec::new(),
+        },
+    });
+}
```
delbonis: This is a weird function.  There should be a function that returns the `MmrIndexEntry` created in this function and then the caller can just push it themselves.  That's a more idiomatic design.

crates/ol/mmr-index/src/mmr_divergence.rs
```diff
@@ -0,0 +1,608 @@
...
+
+/// Pushes a divergence when the persisted state differs from the target.
+fn add_divergent_ol_mmr_index(
+    divergences: &mut Vec<DivergentOLMmrIndex>,
+    entry: MmrIndexEntry,
+    target_mmr_state: MmrStateView,
+) {
+    let target_leaf_count = target_mmr_state.leaf_count;
```
delbonis: This fn does a lot of the same "weird thing" as the other one.  Have a separate fn that creates the index depending on the context and then the caller can push it to the vec.

crates/ol/mmr-index/src/mmr_divergence.rs
```diff
@@ -0,0 +1,608 @@
...
+    }
+}
+
+/// Returns the target inbox MMR state for a snark account.
+fn resolve_target_snark_inbox_mmr_view<S>(
+    target_state: &S,
+    account_id: &AccountId,
+) -> Result<MmrStateView, OLMmrIndexError>
+where
+    S: IStateAccessor,
```
delbonis: `impl IStateAccessor`

crates/ol/mmr-index/src/mmr_divergence.rs
```diff
@@ -0,0 +1,608 @@
...
+        // Compact roots are stored lowest-height first. Storage handles expose
+        // peaks in the left-to-right order returned by `get_state_at`.
+        peaks: mmr
+            .roots
+            .iter()
+            .rev()
+            .map(|peak| Hash::from(peak.0))
+            .collect(),
+    }
+}
```
delbonis: Sorta an example of "why do we have a separate `MmrStateView` type in the first place?".  Is it possible that we can make some of these fns generic over the trait?  At the very least we should remove `pub` from the fields and add helper constructors to the type so that we aren't having to muck around in the internals to keep up the invariants.

crates/ol/mmr-index/src/mmr_reconcile.rs
```diff
@@ -0,0 +1,322 @@
...
+    /// Builds a validated plan from persisted MMR index entries and target state.
+    pub fn build_for_target<S, I>(
+        target_state: &S,
+        entries: Vec<MmrIndexEntry>,
+        target_snark_accounts: I,
+    ) -> Result<Self, OLMmrIndexError>
+    where
+        S: IStateAccessor,
+        I: IntoIterator<Item = AccountId>,
+    {
```
delbonis: This is still a bad fn to have on the type, it's not something that's *intrinsic* to the `MmrIndexReconciliationPlan`.  Should split it out as a standalone routine.

Again, `impl Trait`.

delbonis (changes requested):
Also see other review comment replies.

crates/chain-worker/src/mmr_prefill.rs
krsnapaudel: @delbonis I mean I refactored this just to reduce duplicate code. This should only run during start up afaiu. So the assert should be sufficient.

crates/db/types/src/mmr_index.rs
krsnapaudel: @delbonis 2 things:
1. The type was pub with fields pub. It was in `storage`. I moved it to `db/types` for reuse in `ol/mmr-index` without depending on `storage`. Sure we can make fields private and add accessor apis.
2. About ordering and LLM response. I asked Claude and Codex both why implementing `MmrState` is not great for `MmrStateView`. This is the summary: `MmrState` is an accumulator interface. Not just leaf count and peaks. So yes the sentence was from LLM, but the idea is `MmrStateView` is different because it is a DB snapshot returned by `MmrIndexHandle::get_state_at`, with peaks in storage/read order. You will notice that I did use LLM for summary here as well :)

crates/ol/mmr-index/src/mmr_divergence.rs
krsnapaudel: @delbonis I agree about the `pub` fields. `MmrStateView` is the normalized DB/read-side comparison shape. Storage returns peaks in storage order from `get_state_at`, while `Mmr64` roots are native compact-MMR order (lowest peak height first), so `extract_ol_mmr_state_view` explicitly converts compact MMR into the storage-order. I have created[ a separate ticket](https://alpenlabs.atlassian.net/browse/STR-4035) to do this removal.

krsnapaudel:
@delbonis I have cleaned up after #2151 and #2149.

delbonis (approved):
Full disclosure: I have been unable to focus enough to review this PR again but it's been been through so many iterations that I assume that there aren't any immediately-pressing major problems, so approving as-is.

PR #2077, opened 2026-07-06
STR-3311 Refactor sensitive MsgPayload uses to keep a Coin

delbonis:
## Description

This PR adds a new `MsgPayloadCoin` type that works like `MsgPayload` but carries the value as a `Coin`, and uses it in various places where we were previously creating/destroying `Coin`s on an ad-hoc basis to be more deliberate about the bookkeeping.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

### DIFF
--- a/crates/ol/stf/src/account_processing.rs
+++ b/crates/ol/stf/src/account_processing.rs
@@ -2,7 +2,7 @@
 
 use bitcoin_bosd::Descriptor;
 use strata_acct_types::{
-    AccountId, BRIDGE_GATEWAY_ACCT_ID, BRIDGE_GATEWAY_ACCT_SERIAL, BitcoinAmount, MsgPayload,
+    AccountId, BRIDGE_GATEWAY_ACCT_ID, BRIDGE_GATEWAY_ACCT_SERIAL, MsgPayload,
 };
 use strata_ledger_types::*;
 use strata_msg_fmt::MsgRef;
@@ -11,7 +11,10 @@ use strata_ol_msg_types::OLMessageExt;
 use strata_snark_acct_sys as snark_sys;
 use tracing::*;
 
-use crate::{context::BasicExecContext, errors::ExecResult, output::OutputCtx};
+use crate::{
+    context::BasicExecContext, errors::ExecResult, msg_payload_coin::MsgPayloadCoin,
+    output::OutputCtx,
+};
 
 /// Processes a message by delivering it to its destination, which might involve
 /// touching the ledger state.
@@ -19,7 +22,7 @@ pub(crate) fn process_message<S: IStateAccessorMut>(
     state: &mut S,
     sender: AccountId,
     target: AccountId,
-    msg: MsgPayload,
+    msg: MsgPayloadCoin,
     context: &BasicExecContext<'_>,
 ) -> ExecResult<()> {
     match target {
@@ -30,19 +33,20 @@ pub(crate) fn process_message<S: IStateAccessorMut>(
 
         // Any other address we assume is a ledger account, so we have to look it up.
         _ => {
-            let coin = Coin::new_unchecked(msg.value());
-
             // Check if the account exists first.
             if !state.check_account_exists(target)? {
                 warn!(
                     %target,
                     %sender,
-                    value = %msg.value(),
+                    value = %msg.coin_amt(),
                     "limboing message to nonexistent target account",
                 );
-                handle_misplaced_funds(state, coin)?;
+                handle_misplaced_funds(state, msg.into_coin())?;
                 return Ok(());
             }
+
+            // Reconstitute the coin and a record payload for any snark inbox.
+            let (coin, record) = msg.into_coin_and_record();
 
             // Update the account within a closure.
             state.update_account(target, |acct_state| -> ExecResult<()> {
@@ -53,7 +57,7 @@ pub(crate) fn process_message<S: IStateAccessorMut>(
                 // for postprocessing.
                 if let Ok(sastate) = acct_state.as_snark_account_mut() {
                     // Call the handler fn now that we've increased the balance.
-                    handle_snark_account_message(sastate, sender, &msg, context)?;
+                    handle_snark_account_message(sastate, sender, &record, context)?;
                 }
                 // Empty accounts don't need any additional processing.
 
@@ -69,11 +73,9 @@ pub(crate) fn process_transfer<S: IStateAccessorMut>(
     state: &mut S,
     _sender: AccountId,
     target: AccountId,
-    value: BitcoinAmount,
+    coin: Coin,
     _context: &BasicExecContext<'_>,
 ) -> ExecResult<()> {
-    let coin = Coin::new_unchecked(value);
-
     match target {
         // Bridge gateway transfer, not permitted.
         BRIDGE_GATEWAY_ACCT_ID => {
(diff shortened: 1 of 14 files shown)


crates/ol/stf/src/manifest_processing.rs
```diff
@@ -236,10 +237,13 @@ fn process_deposit_log<S: IStateAccessorMut>(
 ) -> ExecResult<()> {
     let amt_btc = BitcoinAmount::from_sat(deposit.amount);
 
+    // Mint the coin once at the deposit ingress boundary; from here on the value
+    // is carried linearly and must be credited or swept to limbo exactly once.
+    let coin = Coin::new_unchecked(amt_btc);
```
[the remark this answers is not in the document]
bewakes: I think this assessment is correct. Coin should only be minted after a valid deposit.

bewakes (changes requested):
Looks good and safer than what was there before. I think the codex's review are valid and this needs to have more tests around correct coin handling between errors, especially the two failure cases codex has pointed out.

crates/ol/stf/src/transaction_processing.rs
```diff
@@ -174,31 +175,61 @@ fn apply_tx_effects<S: IStateAccessorMut>(
...
-        account_processing::process_transfer(state, source, t.dest(), t.value(), context)?;
+        let coin = split_effect_value(&mut remaining, t.value());
+        account_processing::process_transfer(state, source, t.dest(), coin, context)?;
     }
 
     for m in effects.messages_iter() {
-        account_processing::process_message(state, source, m.dest(), m.payload().clone(), context)?;
+        let coin = split_effect_value(&mut remaining, m.payload().value());
+        let msg = MsgPayloadCoin::new(coin, m.payload().payload_data().clone());
+        account_processing::process_message(state, source, m.dest(), msg, context)?;
```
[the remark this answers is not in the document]
delbonis: Oh there should be a unit test for this!

crates/ol/stf/src/manifest_processing.rs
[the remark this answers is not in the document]
delbonis: This is a good observation.

crates/ol/stf/src/transaction_processing.rs
```diff
@@ -174,31 +175,61 @@ fn apply_tx_effects<S: IStateAccessorMut>(
...
+    // effects.
+    let mut remaining = debit_source(state, source, total_sent)?;
 
-    // 2. Send funds out.
+    // 2. Send funds out, splitting the debited coin per effect so Rust enforces
+    // that the pieces sum to exactly what we took.
     for t in effects.transfers_iter() {
-        account_processing::process_transfer(state, source, t.dest(), t.value(), context)?;
+        let coin = split_effect_value(&mut remaining, t.value());
+        account_processing::process_transfer(state, source, t.dest(), coin, context)?;
```
krsnapaudel, line 213: This is similar to the line flagged below.

crates/ol/stf/src/account_processing.rs
krsnapaudel: Found by Claude.
`context.emit_typed_log(...)?` runs while `payload: MsgPayloadCoin` still owns the coin — it isn't consumed until line 181. `emit_typed_log` is genuinely fallible ("Returns an error if the block log cap would be exceeded", so a full log block on an otherwise-valid withdrawal → ? → payload drop → panic.

crates/ol/stf/src/account_processing.rs
```diff
@@ -88,7 +90,7 @@ pub(crate) fn process_transfer<S: IStateAccessorMut>(
             if !state.check_account_exists(target)? {
```
krsnapaudel: Same panic-on-error shape.

krsnapaudel (commented):
Some findings with help from Claude.

crates/ol/stf/src/account_processing.rs
```diff
@@ -30,20 +33,21 @@ pub(crate) fn process_message<S: IStateAccessorMut>(
...
-                handle_misplaced_funds(state, coin)?;
+                handle_misplaced_funds(state, msg.into_coin())?;
                 return Ok(());
             }
 
+            // Reconstitute the coin and a record payload for any snark inbox.
+            let (coin, record) = msg.into_coin_and_record();
+
             // Update the account within a closure.
             state.update_account(target, |acct_state| -> ExecResult<()> {
```
storopoli: Two more drop-on-error windows in `process_message`, same shape as the cases already flagged:

1. `state.check_account_exists(target)?` (L37) can return a `StateError` while `msg: MsgPayloadCoin` still owns the coin, so the `?` drops it and `Coin::drop` panics.
2. `state.update_account(target, ...)?` (L52) can fail *before* the closure ever runs (e.g. `MissingAccount` if the account vanishes between the existence check and the update, depending on the accessor impl). The coin has already been moved into the closure at that point, so the `Err` return drops it unconsumed.

The snark-handler `?` inside the closure is fine — `add_balance(coin)` has already consumed the coin by then. But whatever fix lands for the `apply_tx_effects` loop should probably cover these two as well, since they panic inside the callee rather than at the `remaining` coin.

crates/ol/stf/src/account_processing.rs
```diff
@@ -88,7 +90,7 @@ pub(crate) fn process_transfer<S: IStateAccessorMut>(
             if !state.check_account_exists(target)? {
                 warn!(
                     %target,
-                    %value,
+                    value = %coin.amt(),
                     "limboing transfer to nonexistent target account",
                 );
                 handle_misplaced_funds(state, coin)?;
```
storopoli: `handle_misplaced_funds` itself is a drop-on-error path, which is worth calling out because it's the designated safe-disposal sink for the linear coin: 

https://github.com/alpenlabs/alpen/blob/725cae922cf08647c1ffb437a9a3e9722486c69e/crates/ol/stf/src/account_processing.rs#L201-L207

implementations check limbo overflow *before* consuming the coin, see 

https://github.com/alpenlabs/alpen/blob/725cae922cf08647c1ffb437a9a3e9722486c69e/crates/ol/state-support-types/src/memory_state_layer.rs#L151-L155

and 

https://github.com/alpenlabs/alpen/blob/725cae922cf08647c1ffb437a9a3e9722486c69e/crates/ol/state-support-types/src/write_tracking_layer.rs#L239-L244

so a `LimboFundsOverflow` returns `Err` with the coin still live, and every `handle_misplaced_funds(state, coin)?` call site (this one and all the sweep-to-limbo branches) turns that into a `Coin::drop` panic.

Since limbo is where coins go when nothing else can take them, it may be worth making this path infallible by construction (consume the coin unconditionally and saturate/alarm on overflow) or defusing the coin on the error path — otherwise every future sweep site inherits the panic.

storopoli (changes requested):

crates/ol/stf/src/manifest_processing.rs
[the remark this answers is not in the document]
delbonis: Actually I have a solution that works even better, we figure out all the lookups in advance to figure out what we're doing with it and *then* create the `Coin` to handle it as needed, see commit in a bit.

crates/ol/stf/src/account_processing.rs
delbonis: Yeah these code paths should never be able to panic now.

delbonis:
Improved a bunch of code paths to be more conservative about when we construct `Coin`s.  Waiting on CI now.

storopoli (approved):
ACK f1aa6481981995880c773ae8fb17bb6636b671b9

krsnapaudel (approved):

bewakes (approved):

Issue #142, opened 2024-07-15
Add watchdogs on core components to shutdown the whole client on fatal error


delbonis:
As a crash safety measure, we should add a watchdog task that checks the handles for all of the core pipeline tasks/threads to ensure they're still running / making process.  If any fail / hard-stall, it would print an error with whatever information it knows and shut down the client as gracefully as it can (closing databases, flushing caches, completing duties, etc.).

It should also be able to handle panics in these core tasks.  If it encounters a panic or simpler error it should present that information on an RPC while it's trying to shut down so that it's more observable.

PR #2079, opened 2026-07-07
ci(build): mirror strata and alpen-client images to public ECR (STR-3894)

alexhui01:
## Summary

Publishes the `strata` and `alpen-client` node images to **public ECR** (`public.ecr.aws/alpenlabs`) alongside the existing private-ECR push, so external testnet operators can pull them without AWS credentials.

Depends on alpen-infra PR #142 (STR-3894), which created the public repos (`strata`, `alpen`, `checkpoint-sync`) and granted this repo's push role `ecr-public:*` access.

## Changes (`.github/workflows/ci-build.yml`)

- **New "Login to Amazon ECR Public" step**, gated to `strata` / `alpen-client`. Public ECR is a separate registry with a us-east-1-only auth token, so it needs its own `docker login public.ecr.aws`. Uses the same assumed role (`SHARED_ECR_ROLE_ARN` = `github-actions-ecr-push`), which already holds the `ecr-public:*` push perms — **no new secret**.
- **Multi-tag build/push** — one `buildx build --push` emits both the private tag and, for mapped programs, the public tag (same digest in both registries):
  - `strata` → `public.ecr.aws/alpenlabs/strata`
  - `alpen-client` → `public.ecr.aws/alpenlabs/alpen`
  - all other programs (`strata-signer`, `strata-datatool`) → private only, unchanged.

## Notes / known follow-ups (tracked in STR-3896)

- **`checkpoint-sync` is not covered here** — it isn't built in this repo. Its public repo exists but must be published from whatever repo builds it.
- **Public pushes are not gated to `main`** — this workflow also runs on cron/PR, matching current private-push behavior. A `main`-only guard for public tags is a deliberate open question.
- **No tag immutability on ECR Public** — SHA tags can be overwritten; consumers should pin by digest.
- Name flip `alpen-client` → `alpen` is encoded in a single `case` mapping.

STR-3896 tracks a standardization + security pass over this wiring (naming source-of-truth, hardcoded alias → var, main-only gating, image-content review, provenance).

Ticket: STR-3894

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -242,6 +242,17 @@ jobs:
         with:
           mask-password: "true"
 
+      # Public ECR is a distinct registry (public.ecr.aws) with its own auth
+      # token, which must be requested from us-east-1. The same assumed role
+      # (SHARED_ECR_ROLE_ARN) is authorized to push the public node images —
+      # see alpen-infra github-actions-image-push (STR-3894). Only the programs
+      # published publicly need this login.
+      - name: Login to Amazon ECR Public
+        if: matrix.program == 'strata' || matrix.program == 'alpen-client'
+        run: |
+          aws ecr-public get-login-password --region us-east-1 \
+            | docker login --username AWS --password-stdin public.ecr.aws
+
       # --- Download pre-built SP1 artifacts from shared job --------------------
       - name: Download SP1 artifacts
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-build.yml
purusang, line 97: @alexhui01 I do not recall if we have checkpoint-sync image building pipeline.

purusang (changes requested):

.github/workflows/ci-build.yml
alexhui01: Not yet, it's something @bewakes handle.

bewakes: I created a code fix [PR](https://github.com/alpenlabs/alpen/pull/2080) and [ci changes](https://github.com/alpenlabs/alpen/pull/2082) for building checkpoint image. Lmk what else needs to be done.

purusang (approved):

PR #1950, opened 2026-06-05
[DO NOT MERGE] STR 2404 EE DA verification tool

krsnapaudel:
## Description

Adds `bin/alpen-ee-da-tool`: a verifier that reconstructs the EE execution state root from L1 DA envelopes and can compare reconstruction against manual or OL-published roots.

Comparison sources:
- `--expected-root`: manual/debug comparison against the reconstructed EE execution state root.
- `ol_rpc_url` in config: product comparison against OL-published Snark account state by fetching:
  - `strata_getSnarkAcctUpdateManifest(ee_params.account_id(), update_seq_no)`
  - `strata_getSnarkAcctInboxMsgRange(...)`
- If neither source is configured, the tool runs reconstruction-only and reports no comparison result.

Report fields separate execution-root and Snark-account-root comparisons:
- `reconstructed_state_root`
- `expected_state_root`
- `state_root_matches_expected`
- `reconstructed_inner_state_root`
- `expected_inner_state_root`
- `inner_state_root_matches_expected`
- `inner_state_root_mismatch_update_seq_no`
- `soundness_notes`

Config:
- `ee_params` supplies the EE account id and genesis metadata.
- `ol_rpc_url` enables published inner-root comparison.
- No separate `ee_snark_account_id` config field.

New crates:
- `alpen-ee-da-l1-extraction` — bounded-range L1 fetch, cross-block commit/reveal scan, blob reassembly.
- `alpen-ee-da-state-replay` — ordered DA blob replay from chain-spec genesis or explicit pre-state snapshot.

Library changes in `alpen-reth-statediff`:
- Public `StateReconstructorPreState`, `from_prestate()`, and `to_prestate()` to support snapshot import/export.
- Behavior change: empty accounts are now deleted from the state trie on update (EIP-161). Diffs that empty an account now produce a different state root than before; the previous behavior left a stale entry.

CLI:
- `--custom-chain` selects the Reth chain spec and is checked against `ee_params`.
- `--snapshot` resumes replay from a pre-state snapshot.
- `--export-snapshot` writes the final replay state for later partial replay.
- `--expected-root` remains available as a manual execution-root comparison override.

This PR was created with help from Codex and Claude Code.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
[...]

PR #2087, opened 2026-07-08
fix(reth-statediff): add reconstructor prestate API and EIP-161 deletion

krsnapaudel:
## Description

Restructures `alpen-reth-statediff` around `rsp_mpt::EthereumState`, so tries are the single source of truth for reconstructed state. Also fixes two empty-account bugs and adds read helpers needed by downstream verifier code. Prerequisite for #1950 (STR-2404 EE DA verification tool).

- **Reuse `EthereumState` for reconstruction.** Removes `StateReconstructor`; adds `ethereum_state_from_genesis_accounts` and (feature-gated) `ethereum_state_from_chain_spec`. Genesis input uses `alloy_genesis::GenesisAccount`. `apply_batch_state_diff_to_ethereum_state` remains the sole apply boundary.
- **Preserve storage-only genesis accounts.** Main's `from_chain_spec` silently dropped accounts with empty state fields but declared storage. Skip only when both state fields and storage are empty.
- **`EthereumStateExt` read helpers.** Extension trait on `EthereumState`: `state_root_buf32`, `get_account_snapshot`, `get_storage_slot`. Doc comments note these are reconstructed-state accessors, not presence/absence proofs against an externally supplied state root.
- **EIP-161 empty-account deletion in apply.** When `apply_batch_state_diff_to_ethereum_state` produces an empty account, delete it from the state trie and remove its storage trie, matching Ethereum's post-Spurious Dragon semantics.

### Tests
- Genesis construction covers accounts with code, storage-only accounts, and zero-slot filtering.
- Existing cross-verification against a `strata_mpt`-backed test oracle preserved.
- `apply_batch_state_diff_to_ethereum_state` empty-account deletion covered.
- `EthereumStateExt` helpers exercised in genesis-construction tests.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

STR-2404 (#1950) needs the prestate API to power snapshot-based partial replay in `alpen-ee-da-tool`. Scoping the reth-statediff work here keeps review with reth-statediff owners; #1950 then only needs an EE DA reviewer.

- [ ]  Yes
- [x]  No

## Checklist

### DIFF
--- a/crates/reth/statediff/src/reconstruct.rs
+++ b/crates/reth/statediff/src/reconstruct.rs
@@ -1,17 +1,12 @@
 //! State reconstruction from batch diffs.
 
-#[cfg(test)]
-use std::collections::BTreeMap;
-use std::collections::HashMap;
-
+pub use alloy_genesis::GenesisAccount;
 #[cfg(feature = "chainspec")]
 use alpen_chainspec::chain_value_parser;
-use revm_primitives::{alloy_primitives::Address, B256, U256};
+use revm_primitives::{alloy_primitives::Address, B256};
 use rsp_mpt::EthereumState;
 use strata_da_framework::ContextlessDaWrite;
-#[cfg(feature = "chainspec")]
-use strata_mpt::KECCAK_EMPTY;
-use strata_mpt::{keccak, MptNode, StateAccount, EMPTY_ROOT};
+use strata_mpt::{keccak, StateAccount, EMPTY_ROOT, KECCAK_EMPTY};
 use thiserror::Error as ThisError;
 
 use crate::{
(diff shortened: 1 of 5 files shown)


crates/reth/statediff/src/reconstruct.rs
delbonis: I realize this is not even a touched line in this PR but this should really be two steps.  This shouldn't be setting up the whole reconstructor starting with an on-disk path.  How do you even test this unless you're also loading from file?

crates/reth/statediff/src/reconstruct.rs
```diff
@@ -30,14 +28,61 @@ pub enum ReconstructError {
...
+/// Maintains canonical account and storage maps alongside the MPTs to support
+/// cheap export through [`StateReconstructor::to_prestate`]. This roughly doubles
+/// state memory, and every mutation must keep the maps and MPTs in lock-step.
 #[derive(Clone, Default, Debug)]
 pub struct StateReconstructor {
     state_trie: MptNode,
     storage_trie: HashMap<Address, MptNode>,
+    accounts: BTreeMap<Address, StateAccount>,
+    storage: BTreeMap<Address, BTreeMap<U256, U256>>,
 }
```
delbonis: This seems like it's trying to do two things at the same time: track updates to the accounts and storage slots and *also* update the MPT trees that these are embedded in.  And it's a lot more complicated as a result, there's an enormous amount of tedious tiny checks that need to be perfectly consistent across this very large module.

Can't we just maintain something like a "`WriteBatch`" and then have a function that applies it to something representing the tries?

```rs
pub struct WriteBatch { // redefining this locally
  account_writes: BTreeMap<Address, StateAccount>, // maybe needs to be missing the updated storage root
  storage_writes: BTreeMap<Address, BTreeMap<U256, U256>>,
}

pub struct PartialState {
  accounts_trie: MptNode,
  storage_tries: BTreeMap<Address, MptNode>
}

impl PartialState {
  pub fn apply_writes(&mut self, wb: &WriteBatch) -> Result<..., ...> {
    // ...
  }
}
```

There might be some reason this can't work, related to how there *is* already a `BatchStateDiff` type separate from this module.

crates/reth/statediff/src/reconstruct.rs
```diff
@@ -30,14 +28,61 @@ pub enum ReconstructError {
...
+/// Canonical account and storage state used to initialize a [`StateReconstructor`].
+///
+/// The prestate stores account records separately from per-account storage. During
+/// initialization, the reconstructor recomputes each account's storage root from the
+/// supplied storage map before inserting the account into the global state trie.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct StateReconstructorPreState {
+    accounts: BTreeMap<Address, StateAccount>,
+    storage: BTreeMap<Address, BTreeMap<U256, U256>>,
+}
```
delbonis: What makes something a pre-state isn't what it's composed of but the context that it's used in.  So this totally plausibly could be used to represent a post-state (as nearly every pre-state is, for the previous block) if it wasn't for the name.  We can just call this `PartialState`.

crates/reth/statediff/src/reconstruct.rs
```diff
@@ -66,24 +111,102 @@ impl StateReconstructor {
...
         Ok(reconstructor)
     }
 
+    /// Exports the current canonical state as a reconstructor prestate.
+    ///
+    /// Empty accounts and zero-valued storage slots are not represented in the
+    /// returned prestate.
+    pub fn to_prestate(&self) -> StateReconstructorPreState {
+        StateReconstructorPreState::new(self.accounts.clone(), self.storage.clone())
+    }
```
delbonis: In ref to my above comment, this is contextually a post-state.

delbonis (changes requested):
I have a lot of comments about the approach taken in this PR, it seems like it wasn't designed intentionally and just kinda grew without a plan.

crates/reth/statediff/src/reconstruct.rs
krsnapaudel: @delbonis Split into two steps in this PR: `ethereum_state_from_genesis_accounts` takes a plain iterator (unit-testable without any file loading), and `ethereum_state_from_chain_spec` (feature-gated) parses the chain spec then delegates.

crates/reth/statediff/src/reconstruct.rs
krsnapaudel: @delbonis Agreed. Dropped it. `StateReconstructor` is removed from the public API, replaced by reusing `rsp_mpt::EthereumState` (which already has `state_trie` + `storage_tries` and cheap `state_root`). `BatchStateDiff` is applied via the existing `apply_batch_state_diff_to_ethereum_state` helper.

krsnapaudel:
@codex review

delbonis (approved):

storopoli (approved):
ACK a7e63d81340ef86c5f0e6372ee78282648118fe2

irnb:
@krsnapaudel i have a question related to <ins>**_EIP-161 empty-account deletion in apply._**</ins>

eip-6780 kinda disables selfdestruct, and it's enabled from genesis on our ee chain.
so in which situation do we actually face a deleted account?

krsnapaudel:
> @krsnapaudel i have a question related to **_EIP-161 empty-account deletion in apply._**
> 
> eip-6780 kinda disables selfdestruct, and it's enabled from genesis on our ee chain. so in which situation do we actually face a deleted account?

@irnb It's mostly defensive. Reth's diff producer normalizes destroyed accounts into `AccountChange::Deleted`, so the `Deleted` branch handles the real cases; our empty-check only fires if a diff arrives as `Updated` with signed deltas that zero the account — which the current producer shouldn't emit. EIP-6780 doesn't change this (it restricts `SELFDESTRUCT`; EIP-161 touch-cleanup still happens and still gets emitted as `Deleted`). Kept as insurance against producer edge cases and future changes, and matches EVM cross-verification.

krsnapaudel:
@sapinb @evgenyzdanovich Could one of you please review? Blocked.

krsnapaudel:
@storopoli Rebased onto `main` after #1997 and updated so `strata-dbtool` works.

irnb (approved):

storopoli (approved):
ACK 47f1de1283da0488aa27b4aedf7e83037f427679

storopoli:
We need approval from codeowners @evgenyzdanovic @sapinb and/or @MdTeach

MdTeach (approved):

PR #2095, opened 2026-07-10
feat(alpen-cli)!: source network params from the config file

prajwolrg:
## Description

The wallet CLI required a full `asm-params.json` alongside its `config.toml`, but only ever read four values from it: the L1 network, the SPS-50 magic bytes, the deposit denomination, and the deposit-request recovery delay. This PR flattens those four values into `config.toml` as required fields and removes the ASM params file plumbing (`STRATA_NETWORK_PARAMS` env var, `asm_params_path` config key) and the `strata-asm-params` dependency from the CLI.

Rationale beyond the smaller dependency surface: the params file is only the genesis config, so it was never authoritative post-genesis anyway (the operator set can rotate via the admin subprotocol), and the CLI already required consensus-critical values in its config (`bridge_pubkey`) that had to be kept consistent with the params file by hand. A single provisioning artifact is one fewer chance to mix up networks. This also removes a runtime dependency of the wallet CLI on an OL-generated artifact, which helps the planned EE repo split.

Since these values are consensus-critical (a wrong `recovery_delay` produces a deposit taproot the bridge won't recognize), the config should be generated, not hand-written: `gen-asm-params` gains a `--cli-config <path>` option that derives the CLI config snippet from the same inputs as the ASM params. Both crates pin the same byte-exact snippet literal in tests — datatool that it emits it, the CLI that it parses it into `SettingsFromFile` — so a field rename on either side fails a test. The snippet includes the aggregated MuSig2 `bridge_pubkey`, derived from the operator set with the same `strata_crypto::aggregate_schnorr_keys` the ASM bridge subprotocol uses for its N/N key (registration order, duplicates skipped), so the last hand-copied consensus-critical value in the CLI config is generated too. The command refuses to overwrite an existing file and rejects a `--cli-config` path equal to `--output`.

Breaking change for CLI deployments: `config.toml` must now provide `network`, `magic_bytes`, `bridge_denomination_sats`, and `recovery_delay`; `withdrawal_denomination_sats` is replaced by `bridge_denomination_sats`, which now feeds both the deposit amount and the withdrawal denomination (previously deposits read the ASM value while withdrawals read an unrelated config default). The migration is documented in the datatool README.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Review entry point: `bin/alpen-cli/src/settings.rs`. For the `bridge_pubkey` derivation, please double-check that `derive_bridge_pubkey` in `bin/datatool/src/cmd/asm_params.rs` exactly mirrors `OperatorTable::calculate_aggregated_key` in the ASM bridge subprotocol (aggregation in registration order, duplicates skipped, untweaked x-only result) — a mismatch there would emit a key deposits can't be recognized under.

Nothing in-repo (functional tests, docker, CI) provisions the wallet CLI, so the blast radius is external deployments only.

- [ ]  Yes
- [x]  No

## Checklist

## AI Assistance Notice

This PR was authored with the assistance of Claude Code.


bin/alpen-cli/src/settings.rs
```diff
@@ -66,14 +61,28 @@ pub struct SettingsFromFile {
...
+    pub network: Network,
+    /// SPS-50 magic bytes tagging protocol transactions on L1 (e.g. "ALPN").
+    ///
+    /// Must match the magic bytes in the ASM params.
+    pub magic_bytes: MagicBytes,
+    /// At-rest deposit denomination in satoshis, used for both deposits and
+    /// withdrawals.
+    ///
+    /// Must match the Bridge subprotocol denomination in the ASM params.
+    pub deposit_amount_sats: u64,
```
barakshani: let's call this "bridge denomination", as it's used for both deposits and withdrawals.

bin/datatool/src/cmd/asm_params.rs
```diff
@@ -184,6 +187,65 @@ pub(super) fn exec(cmd: SubcAsmParams, ctx: &mut CmdContext) -> anyhow::Result<(
...
+        magic_bytes: asm_params.magic,
+        bridge_pubkey: derive_bridge_pubkey(&bridge.operators)?,
+        deposit_amount_sats: bridge.denomination.to_sat(),
+        recovery_delay: bridge.recovery_delay,
+    };
+
+    let header = "# Alpen CLI network profile derived from the ASM params.\n\
+                  # Merge these fields into the CLI's config.toml.\n";
+    let body = toml::to_string(&profile)?;
+    fs::write(path, format!("{header}{body}"))?;
```
storopoli: `fs::write` silently truncates whatever already exists at this path. The other datatool outputs behave the same way, but those are standalone artifacts; this one is a snippet whose own header:

https://github.com/alpenlabs/alpen/blob/fb178e398721852f8b8afcd614aa346ba2d0818d/bin/datatool/src/cmd/asm_params.rs#L244-L245

says to merge it into the CLI's `config.toml`. The flag name makes it tempting to point it directly at a live `config.toml`, which would wipe every other setting in it (endpoints, seed config, ...). Consider refusing to overwrite an existing file, and bailing when the target collides with `--output`.

bin/datatool/src/cmd/asm_params.rs
```diff
@@ -221,13 +283,72 @@ fn resolve_safe_harbour_address(descriptor: &str) -> anyhow::Result<SafeHarbourA
...
+
     use super::*;
 
     /// X-only pubkey hex derived from the test xpriv
     /// `tprv8ZgxMBicQKsPd4arFr7sKjSnKFDVMR2JHw9Y8L9nXN4kiok4u28LpHijEudH3mMYoL4pM5UL9Bgdz2M4Cy8EzfErmU9m86ZTw6hCzvFeTg7`
     /// via `genseqpubkey`.
     const TEST_SEQ_PK: &str = "14ebfa9a90fee3020686b5334b297b675a9f29282f44b6c3a4ab1f0582021839";
 
+    #[test]
+    fn cli_network_profile_matches_cli_config_schema() {
```
storopoli, line 424: This test pins the datatool-local `CliNetworkProfile` against a datatool-local string, so it can't catch drift on the CLI side: renaming a field in `SettingsFromFile`

https://github.com/alpenlabs/alpen/blob/fb178e398721852f8b8afcd614aa346ba2d0818d/bin/alpen-cli/src/settings.rs#L37

would not fail any test. The PR body describes it as pinning the emitted field names to the CLI's config schema, which oversells it. Either add a test that merges the generated snippet into a full config and deserializes it with the actual CLI `SettingsFromFile`, or keep relying on the "must stay in sync" comments:

https://github.com/alpenlabs/alpen/blob/fb178e398721852f8b8afcd614aa346ba2d0818d/bin/datatool/src/cmd/asm_params.rs#L198-L200
and soften the claim in the PR description.

bin/datatool/src/args.rs
```diff
@@ -156,6 +156,12 @@ pub(crate) struct SubcAsmParams {
     )]
     pub(crate) output: Option<PathBuf>,
 
+    #[argh(
+        option,
+        description = "output file path for an alpen-cli network profile toml derived from these params"
+    )]
+    pub(crate) cli_config: Option<PathBuf>,
```
storopoli, line 163: Docs gap: the `gen-asm-params` example in the datatool README:

https://github.com/alpenlabs/alpen/blob/fb178e398721852f8b8afcd614aa346ba2d0818d/bin/datatool/README.md#L36-L44

doesn't show `--cli-config`, and the breaking CLI config migration (four new required fields; `STRATA_NETWORK_PARAMS` and `asm_params_path` removed; `withdrawal_denomination_sats` replaced by `deposit_amount_sats`) isn't documented anywhere. Worth adding to the README since external deployments have to migrate by hand.

storopoli (changes requested):

bin/alpen-cli/src/settings.rs
prajwolrg: Renamed to `bridge_denomination_sats` (and the runtime field to `bridge_denomination`) in 3b19a0dff.

bin/datatool/src/cmd/asm_params.rs
prajwolrg: Done in f51e335d5: `write_cli_network_profile` now refuses to overwrite an existing file, and `--cli-config` pointing at the same path as `--output` is rejected.

bin/datatool/src/cmd/asm_params.rs
prajwolrg: Done in fcaf00b29: both crates now pin the same byte-exact snippet literal — the datatool test asserts it emits it (header included), and a new alpen-cli test parses that literal with the real `SettingsFromFile` and asserts the values. A field rename on either side now fails one of the two tests. Also updated the PR description accordingly.

bin/datatool/src/args.rs
prajwolrg: Done in 23b85b2dd: the README example now shows `--cli-config`, and a new section documents the emitted fields plus the migration from `asm-params.json` (removed env var/config key, `withdrawal_denomination_sats` → `bridge_denomination_sats`).

storopoli (approved):
ACK 23b85b2dd0a76772f582b484e91b4c2b659704ca

bin/datatool/src/cmd/asm_params.rs
```diff
@@ -184,6 +187,81 @@ pub(super) fn exec(cmd: SubcAsmParams, ctx: &mut CmdContext) -> anyhow::Result<(
         println!("{params_buf}");
     }
 
+    if let Some(cli_config_path) = &cmd.cli_config {
+        anyhow::ensure!(
+            cmd.output.as_deref() != Some(cli_config_path.as_path()),
+            "--cli-config must not point at the same file as --output"
```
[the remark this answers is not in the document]
prajwolrg: Fixed in 474887103 — the same-path check now runs at the top of `exec`, before the `--output` write. Verified against a pre-existing `config.toml`: the run bails and the file is left untouched.

bin/datatool/src/cmd/asm_params.rs
```diff
@@ -184,6 +187,81 @@ pub(super) fn exec(cmd: SubcAsmParams, ctx: &mut CmdContext) -> anyhow::Result<(
...
+
+    let bridge = asm_params
+        .bridge_config()
+        .ok_or_else(|| anyhow::anyhow!("ASM params missing Bridge subprotocol config"))?;
+
+    let profile = CliNetworkProfile {
+        network: asm_params.anchor.network,
+        magic_bytes: asm_params.magic,
+        bridge_pubkey: derive_bridge_pubkey(&bridge.operators)?,
+        bridge_denomination_sats: bridge.denomination.to_sat(),
```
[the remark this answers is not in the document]
prajwolrg: Good catch, and it went further than the CLI: nothing tied the ASM deposit denomination to the OL one at all, so any non-default `--deposit-sats` produced a chain whose deposits could not be withdrawn — the CLI profile just made it visible.

Fixed in 9ce620216 by treating the OL params as the source of truth: `--deposit-sats` defaults to `ol_params.bridge_params.denomination` and is rejected when it differs. The profile reads the denomination from the OL params and now also carries `max_withdrawal_amount_sats` and `max_withdrawal_descriptor_len`, so the CLI validates batched withdrawals exactly as the OL STF does.

Note `gen-ol-params` still has no flag for the denomination (stuck at `BridgeParams::default()`), so a non-default network currently means hand-editing the OL params JSON; left a TODO there.

bin/datatool/src/cmd/asm_params.rs
```diff
@@ -47,6 +47,15 @@ const DEFAULT_MAX_SEQNO_GAP: NonZero<u8> = NonZero::new(10).expect("10 is non-ze
 /// Generates the ASM params for a Strata network.
 /// Either writes to a file or prints to stdout depending on the provided options.
 pub(super) fn exec(cmd: SubcAsmParams, ctx: &mut CmdContext) -> anyhow::Result<()> {
+    // Checked before anything is written: otherwise the `--output` write below would
+    // already have clobbered the CLI config by the time we bail.
+    if let Some(cli_config_path) = &cmd.cli_config {
+        anyhow::ensure!(
+            cmd.output.as_deref() != Some(cli_config_path.as_path()),
```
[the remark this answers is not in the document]
prajwolrg: Fixed in e890abe91 — both paths are resolved to the file they would write to before comparing: existing paths are canonicalized outright (which also collapses symlinks), and not-yet-existing outputs fall back to canonicalizing the parent directory and rejoining the file name. Verified `-o ./config.toml --cli-config config.toml` and a symlink alias against a live config: both are rejected before any write and the file survives.

bin/datatool/src/cmd/asm_params.rs
```diff
@@ -184,6 +190,124 @@ pub(super) fn exec(cmd: SubcAsmParams, ctx: &mut CmdContext) -> anyhow::Result<(
...
+struct CliNetworkProfile {
+    network: Network,
+    magic_bytes: MagicBytes,
+    bridge_pubkey: String,
+    bridge_denomination_sats: u64,
+    recovery_delay: u16,
+    /// Withdrawals are batched in multiples of the denomination, so the CLI needs
+    /// the OL's cap to reject amounts the OL STF would reject.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    max_withdrawal_amount_sats: Option<u64>,
```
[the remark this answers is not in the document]
prajwolrg: Fixed in e5c1e1614, though with a refusal rather than an encoding: TOML has no null and the CLI reads an absent `max_withdrawal_amount_sats` as the default 10 BTC cap, so the profile genuinely cannot express "uncapped" — any value emitted is either a wrong cap or a sentinel the CLI doesn't understand yet. `gen-asm-params --cli-config` now rejects OL params with `max_withdrawal_amount: null`, naming the field to set, and there's a TODO to teach the CLI config an explicit uncapped spelling so this can emit it instead.

The profile is also now built before any output is written, so a run that cannot emit it no longer leaves a half-generated `asm-params.json` behind.

bin/datatool/src/cmd/asm_params.rs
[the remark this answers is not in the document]
prajwolrg: Follow-up: the node-side half of this is now tracked in [STR-3971](https://alpenlabs.atlassian.net/browse/STR-3971) — `gen-asm-params` enforces the denomination invariant at generation time, but the node loads the two params files independently and should refuse to boot on a mismatched pair (there's a `TODO(STR-3971)` at the load site in `bin/strata/src/context.rs`). Reusing the ASM crate's `OperatorTable` for the emitted `bridge_pubkey` instead of mirroring it is [STR-3972](https://alpenlabs.atlassian.net/browse/STR-3972).

krsnapaudel (approved):

PR #2110, opened 2026-07-13
(fix): Correctly calculate `latest` epoch in checkpoint sync node + do not serve null epoch for genesis epoch

bewakes:
## Description
Also closes [STR-3957](https://alpenlabs.atlassian.net/browse/STR-3957).

While running alpen full node with checkpoint sync node, it was discovered that the `latest`(derived from `OLSyncStatus`'s `prev_epoch`) was behind the `confirmed/finalized` epochs and that alpen full node's ol tracker was erroring out.

This PR does the following:
- Renames the `prev_epoch` inside `OLSyncStatus` to `recently_complete_epoch` adhering to the docstring. The previous epoch framing was correct in the context of FCM only and with checkpoint sync, it is not correct.
- Correctly sets its value inside CSS's `build_ol_sync_status()` function to the finalized epoch instead of previous finalized epoch. This is because, for checkpoint sync node, the recently complete epoch is the finalized epoch itself as it has nothing beyond finalized checkpoint received from L1.
- Also fixes genesis commitment reported as null commitment(`000..`) instead of actual genesis commitment.
- `test_checkpoint_sync_node` now explicitly checks the ordering of the values in the RPC response.
- chore: makes `OLSyncStatus`'s fields private and exposes getters.

### Why was this not caught in the existing `test_checkpoint_sync_node` functional test?
- The behavior of this error is that it is treated as recoverable and thus even in the case of full node, the full node does nothing, but just continues in the loop trying to recover.
- In the functional test, we test that the alpen sequencer is able to sync from checkpoint sync node and submit updates to OL sequencer. This update path did not came across the check for `latest`/`confirmed` ordering which only happens in the tracker for full node. The component was erroring out and trying to recover, but since the update path was the one that determined the "pass"ing of the test, the issue was hidden.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

## Related Issues

[STR-3957]: https://alpenlabs.atlassian.net/browse/STR-3957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/strata/src/rpc/node.rs
+++ b/bin/strata/src/rpc/node.rs
@@ -742,7 +742,7 @@ impl<P: OLRpcProvider> OLClientRpcServer for OLRpcServer<P> {
         );
         let confirmed = chain_sync_status.confirmed_epoch;
         let finalized = chain_sync_status.finalized_epoch;
-        let latest = chain_sync_status.prev_epoch;
+        let latest = chain_sync_status.recently_complete_epoch;
 
         Ok(RpcOLChainStatus::new(tip, confirmed, finalized, latest))
     }
--- a/bin/strata/src/rpc/node.rs
+++ b/bin/strata/src/rpc/node.rs
@@ -241,9 +241,9 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
             .ok_or_else(|| internal_error("OL sync status not available"))?;
 
         Ok(match tag {
-            OLBlockTag::Latest => sync_status.tip,
-            OLBlockTag::Confirmed => sync_status.confirmed_epoch.to_block_commitment(),
-            OLBlockTag::Finalized => sync_status.finalized_epoch.to_block_commitment(),
+            OLBlockTag::Latest => sync_status.tip(),
+            OLBlockTag::Confirmed => sync_status.confirmed_epoch().to_block_commitment(),
+            OLBlockTag::Finalized => sync_status.finalized_epoch().to_block_commitment(),
         })
     }
 
@@ -300,7 +300,7 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
         let finalized_slot = self
             .provider
             .get_ol_sync_status()
-            .map(|css| css.finalized_epoch.last_slot())
+            .map(|css| css.finalized_epoch().last_slot())
             .unwrap_or(0);
 
         let mut chain = Vec::new();
@@ -735,14 +735,14 @@ impl<P: OLRpcProvider> OLClientRpcServer for OLRpcServer<P> {
             .ok_or_else(|| internal_error("OL sync status not available"))?;
 
         let tip = RpcOLBlockInfo::new(
-            *chain_sync_status.tip.blkid(),
-            chain_sync_status.tip.slot(),
-            chain_sync_status.tip_epoch,
-            chain_sync_status.tip_is_terminal,
+            *chain_sync_status.tip().blkid(),
+            chain_sync_status.tip().slot(),
+            chain_sync_status.tip_epoch(),
+            chain_sync_status.tip_is_terminal(),
         );
-        let confirmed = chain_sync_status.confirmed_epoch;
-        let finalized = chain_sync_status.finalized_epoch;
-        let latest = chain_sync_status.recently_complete_epoch;
+        let confirmed = chain_sync_status.confirmed_epoch();
+        let finalized = chain_sync_status.finalized_epoch();
+        let latest = chain_sync_status.recently_complete_epoch();
 
         Ok(RpcOLChainStatus::new(tip, confirmed, finalized, latest))
     }
@@ -843,7 +843,7 @@ impl<P: OLRpcProvider> OLClientRpcServer for OLRpcServer<P> {
             let is_finalized = self
                 .provider
                 .get_ol_sync_status()
-                .is_some_and(|sync_status| sync_status.finalized_epoch.epoch() >= epoch);
+                .is_some_and(|sync_status| sync_status.finalized_epoch().epoch() >= epoch);
 
             if is_finalized {
                 RpcCheckpointConfStatus::Finalized { l1_reference }
@@ -1035,7 +1035,7 @@ impl<P: OLRpcProvider> OLClientRpcServer for OLRpcServer<P> {
             .get_ol_sync_status()
             .ok_or_else(|| internal_error("OL sync status not available"))?;
         if let Some(current_seq_no) = self
-            .current_snark_account_seq_no(account_id, sync_status.tip)
+            .current_snark_account_seq_no(account_id, sync_status.tip())
             .await?
         {
             // Account state stores the next operation seqno. Published manifests
@@ -1046,7 +1046,7 @@ impl<P: OLRpcProvider> OLClientRpcServer for OLRpcServer<P> {
                 )));
             }
         }
-        for epoch in creation_epoch..=sync_status.tip_epoch {
+        for epoch in creation_epoch..=sync_status.tip_epoch() {
             let Some(records) = self
                 .provider
                 .get_account_update_records(epoch, account_id)
@@ -1246,7 +1246,7 @@ impl<P: OLRpcProvider> OLFullNodeRpcServer for OLRpcServer<P> {
             .provider
             .get_ol_sync_status()
             .ok_or_else(|| internal_error("OL sync status not available"))?
-            .tip
+            .tip()
             .blkid();
 
         let mut summaries = Vec::with_capacity(count as usize);
--- a/crates/consensus-logic/src/checkpoint_sync/state.rs
+++ b/crates/consensus-logic/src/checkpoint_sync/state.rs
@@ -151,15 +151,19 @@ pub(crate) async fn build_ol_sync_status(
     let terminal = *summary.terminal();
     let epoch_num = summary.epoch();
     let new_l1 = *summary.new_l1();
-    // Epoch 0 has no predecessor; `null` is the canonical genesis-prev value.
-    let prev_epoch = summary
-        .get_prev_epoch_commitment()
-        .unwrap_or(EpochCommitment::null());
+    // The recently complete epoch for checkpoint sync service is the finalized epoch.
+    let recent_epoch = epoch;
 
     // Checkpoint sync always lands on terminal blocks, and for it
     // confirmed == finalized (5th and 6th args).
     Ok(OLSyncStatus::new(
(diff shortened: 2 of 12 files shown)


bin/strata/src/rpc/node.rs
```diff
@@ -735,14 +735,14 @@ impl<P: OLRpcProvider> OLClientRpcServer for OLRpcServer<P> {
...
+            chain_sync_status.tip().slot(),
+            chain_sync_status.tip_epoch(),
+            chain_sync_status.tip_is_terminal(),
         );
-        let confirmed = chain_sync_status.confirmed_epoch;
-        let finalized = chain_sync_status.finalized_epoch;
-        let latest = chain_sync_status.prev_epoch;
+        let confirmed = chain_sync_status.confirmed_epoch();
+        let finalized = chain_sync_status.finalized_epoch();
+        let latest = chain_sync_status.recently_complete_epoch();
```
barakshani, line 745: Sounds to me that this _is_ what latest should be. However, "recently_complete_epoch" should encapsulate epochs that have not yet have a checkpoint (but not only), so semantically this is not the right term here. Moreover, I'm trying to understand if we always only see finalized epochs (checkpoints), which I asked about elsewhere.

crates/status/src/chain.rs
```diff
@@ -11,28 +11,32 @@ pub type OLSyncStatus = ChainSyncStatus;
...
-    pub tip_epoch: Epoch,
+    tip_epoch: Epoch,
 
     /// Whether the current chain tip is epoch terminal.
-    pub tip_is_terminal: bool,
+    tip_is_terminal: bool,
 
     /// The previous epoch (ie. epoch most recently completed).
-    pub prev_epoch: EpochCommitment,
+    recently_complete_epoch: EpochCommitment,
```
barakshani, line 23: These, I believe, are two different things, already discussed in this [message](https://alpen-labs.slack.com/archives/C09C6JLF76H/p1783884108787639?thread_ts=1780349990.107779&cid=C09C6JLF76H). We should not change this without proper discussion on the need for it (as asked before).

prev_epoch is always parent, while recently complete may not be a parent. Do we need the parent? why? Do we need the "latest" epoch? why?

barakshani (commented):
I am trying to understand whether checkpoint-sync RPC really need finalized/confirmed/latest to be different, or should all of them return the same value.

prajwolrg (approved):
LGTM!

functional-tests/tests/strata/test_checkpoint_sync_node.py
```diff
@@ -139,4 +139,16 @@ def check_commitment_matches_checkpoint(seq_rpc, epoch: int, commitment: dict):
...
+    check_epoch_ordering_invariant(status)
+    return status
+
+
+def check_epoch_ordering_invariant(status: ChainSyncStatus):
+    """The latest (recently complete) epoch can never lag confirmed or finalized."""
+    latest = status["latest"]["epoch"]
+    confirmed = status["confirmed"]["epoch"]
+    finalized = status["finalized"]["epoch"]
+    assert latest >= confirmed >= finalized, (
```
krsnapaudel: Do we also expect `tip.epoch()` to be greater than `latest`? At least `>=` right?

crates/consensus-logic/src/checkpoint_sync/state.rs
```diff
@@ -151,15 +151,19 @@ pub(crate) async fn build_ol_sync_status(
     let terminal = *summary.terminal();
     let epoch_num = summary.epoch();
     let new_l1 = *summary.new_l1();
-    // Epoch 0 has no predecessor; `null` is the canonical genesis-prev value.
-    let prev_epoch = summary
-        .get_prev_epoch_commitment()
-        .unwrap_or(EpochCommitment::null());
+    // The recently complete epoch for checkpoint sync service is the finalized epoch.
+    let recent_epoch = epoch;
```
krsnapaudel, line 155: Is this necessary?

krsnapaudel (approved):
Some nits.

functional-tests/tests/strata/test_checkpoint_sync_node.py
storopoli: Ideally this:

`tip ≥ latest ≥ confirmed ≥ finalized`

storopoli (approved):
ACK dcde2040f9d3e9a38852250d7521258367703940

bin/strata/src/rpc/node.rs
bewakes: > Moreover, I'm trying to understand if we always only see finalized epochs (checkpoints), which I asked about elsewhere.

Yes, checkpoint sync node always sees/processes finalized checkpoints.

crates/consensus-logic/src/checkpoint_sync/state.rs
bewakes: Strictly not necessary, but I just want to make it explicit in this case.

bin/strata/src/rpc/node.rs
bewakes: > "recently_complete_epoch" should encapsulate epochs that have not yet have a checkpoint (but not only), so semantically this is not the right term here.

It does, doesn't it? For sequencer, this holds the epoch which hasn't been checkpointed to L1 yet. And for checkpoint sync node, it holds the finalized epoch which is the recently complete epoch in the eyes of checkpoint sync node.

crates/status/src/chain.rs
bewakes: The problem here is that `latest` should not be smaller than the confirmed or finalized. Latest is derived from `OLSyncStatus.prev_epoch`, which behaves correctly for sequencer. But for checkpoint sync, latest is not previous of the tip - latest and tip are same. In case of CSS, it just doesn't make sense to pass in the tip(finalized epoch) value as previous epoch. So I thought about the intent of the latest and what makes sense for both kinds of nodes and concluded that it should be along the lines of 'recently complete epoch', which as I mention [here](https://github.com/alpenlabs/alpen/pull/2110#discussion_r3575940174) also encapsulates the epochs not checkpointed yet.

functional-tests/tests/strata/test_checkpoint_sync_node.py
bewakes: Yes, good point. Done in https://github.com/alpenlabs/alpen/pull/2110/commits/397bf2b350262c97ede5f666fe2413d23f0a9e7e

bin/strata/src/rpc/node.rs
barakshani: when I make an RPC call the an OL node, I should not get different answers, depending on whether the node is syncing from L1 or not. Nodes may not have all the info, so they can return some error message when something is missing, but they should not return answers that require context from the caller.

Moreover, also from the eyes of checkpoint-sync node, this is not the _recently completed_ epoch, this is the recent epoch they are aware of under this design (I am not even sure if that's true, I think they are aware of confirmed (valid) checkpoints as well, even if they don't fully process them), but checkpoint sync node also should know its limitations, which in this case is that there may be epochs it is not aware of (because of design, not because of drift).

voidash (approved):

crates/status/src/chain.rs
barakshani: I understand the issue. Why is your solution better than adding a new field instead of deleting the old one? Note, I am _not_ saying it is.

There is a value that is already used for quite some time. Should we just delete it in this innocent-looking PR? Have you checked what other implications on the system it could have?

To be clear, I agree that using `latest` for _previous_ is not proper.

bewakes: > Why is your solution better than adding a new field instead of deleting the old one?

The only place the `prev_epoch` is used is the rpc handler. I had thought of whether we should add it as a new field, but I didn't find a use case for that.

bin/strata/src/rpc/node.rs
bewakes: > when I make an RPC call the an OL node, I should not get different answers, depending on whether the node is syncing from L1 or not. Nodes may not have all the info, so they can return some error message when something is missing, but they should not return answers that require context from the caller.

I agree and that's a fair point. However, to me this seems kind of inevitable if we use checkpoint sync node for syncing alpen full nodes - and also in general in distributed systems(not all will be serving same values because of lags and all). I think from the RPC level, we should only guarantee about the response invariants and not the exact same values in the response. What do you think?

barakshani: That there are more experienced people in the team than me to answer that.

> not all will be serving same values because of lags and all

Note that I explicitly excluded drifts in my message. There's a difference between nodes that _sometimes_ return not-up-to-date values and nodes that by design always do.

bewakes:
@barakshani this PR has already been approved by the team leads and codeowners. I'd like this to get this in soon because it unblocks further steps for staging deployment/CSS testing and it also contains related fix sneaked in(STR-3957).

The main contention point here is the naming/semantics around `recently_completed_epoch`/`prev_epoch`/`latest` and an out of scope discussion on identical-ness of RPC responses from sequencer and checkpoint sync node. I say out of scope of this PR because the responses are already not identical beyond what this PR does. So I think this should be a separate discussion of its own and should not block this PR.

Can we move forward with this PR and continue the discussion on the contention points separately(already started in slack)?

barakshani:
> @barakshani this PR has already been approved by the team leads and codeowners. I'd like this to get this in soon because it unblocks further steps for staging deployment/CSS testing and it also contains related fix sneaked in(STR-3957).
> 
> The main contention point here is the naming/semantics around `recently_completed_epoch`/`prev_epoch`/`latest` and an out of scope discussion on identical-ness of RPC responses from sequencer and checkpoint sync node. I say out of scope of this PR because the responses are already not identical beyond what this PR does. So I think this should be a separate discussion of its own and should not block this PR.
> 
> Can we move forward with this PR and continue the discussion on the contention points separately(already started in slack)?

I am not blocking the merge. I've been expressing my own opinion, and trust you to apply your best judgment in deciding how to go about it. If you don't agree, I guess it's a no brainer. If you do agree, then you should know what is  changeable and what could be left as tech debt, and make sure we follow through.

bewakes:
Resolving the comments as we have dedicated discussions in slack for the topics raised here. We can follow up from there.

PR #2112, opened 2026-07-14
Fix: sequencer init logic should not run in non sequencer mode

sapinb:
## Description
Some sequencer specific init logic were gated behind sequencer feature flag, but not behind the --sequencer arg. So a non-sequencer (fullnode) mode run on a build with sequencer feature flag enabled runs these specific init logics.

One of the init steps involves a sequencer specific RPC call, that can fail when connected to OL fullnodes.

This fix gates all sequencer specific init logic behind both the feature flag and the --sequencer mode arg.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [x]  Yes
- [ ]  No

https://alpenlabs.atlassian.net/browse/STR-3968

## Checklist

### DIFF
--- a/bin/alpen-client/src/main.rs
+++ b/bin/alpen-client/src/main.rs
@@ -45,7 +45,7 @@ use alpen_ee_config::{AlpenEeConfig, AlpenEeParams};
 use alpen_ee_database::init_db_storage;
 use alpen_ee_engine::{create_engine_control_task, sync_chainstate_to_engine, AlpenRethExecEngine};
 #[cfg(feature = "sequencer")]
-use alpen_ee_exec_chain::init_exec_chain_state_from_storage;
+use alpen_ee_exec_chain::{init_exec_chain_state_from_storage, ExecChainState};
 #[cfg(feature = "sequencer")]
 use alpen_ee_genesis::ensure_finalized_exec_chain_genesis;
 use alpen_ee_genesis::{ensure_batch_genesis, ensure_genesis_ee_account_state};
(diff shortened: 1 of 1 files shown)


prajwolrg (approved):

bin/alpen-client/src/main.rs
```diff
@@ -542,6 +567,13 @@ fn main() {
 
                 use crate::gas_data_provider::RethGasDataProvider;
 
+                let SequencerBootState {
+                    ol_chain_tracker: ol_chain_tracker_state,
+                    exec_chain: exec_chain_state,
+                    batch_builder: batch_builder_state,
+                    batch_lifecycle: batch_lifecycle_state,
+                } = sequencer_boot_state.expect("sequencer boot state built in sequencer mode");
```
storopoli: The invariant behind this `expect` ("boot state is `Some` iff `ext.sequencer`") is established ~200 lines up in a separate `if ext.sequencer` block, so it's asserted rather than unrepresentable. Since this whole block is already the sequencer-only path, you could gate it on the state itself and drop the `expect`:

```rust
if let Some(SequencerBootState {
    ol_chain_tracker: ol_chain_tracker_state,
    exec_chain: exec_chain_state,
    batch_builder: batch_builder_state,
    batch_lifecycle: batch_lifecycle_state,
}) = sequencer_boot_state
{
    // sequencer specific tasks
    ...
}
```

That way a future refactor that decouples the two `ext.sequencer` checks can't turn this into a panic. Non-blocking.

storopoli (commented):
non-blocking nit

ACK c41fe2869383e0831f0fd10170554a3c03fe23b6

bin/alpen-client/src/main.rs
sapinb: updated in 8c9dcdece7f2eb839ff89e09ff7e03e64baafebf

storopoli (approved):
ACK 8c9dcdece7f2eb839ff89e09ff7e03e64baafebf

PR #2113, opened 2026-07-14
fix(alpen-client): chunk OL inbox-message fetch

sapinb:
## Description
`get_inbox_messages` sent the whole `[min, max]` slot span as one `get_blocks_summaries` call. The OL server rejects any request past its `max_headers_range` limit (default 5000), so a far-behind EE could stall with "Block range too big".

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
Low impact after #2112 as this will only be called by ee sequencer node, and sequencer node following trailing OL block tracking by 5k blocks (208 days) is very unlikely.

- [x]  Yes
- [ ]  No

https://alpenlabs.atlassian.net/browse/STR-3969

## Checklist

### DIFF
--- a/bin/alpen-client/src/rpc_client.rs
+++ b/bin/alpen-client/src/rpc_client.rs
@@ -26,6 +26,32 @@ use tracing::info;
 /// Max retries for startup RPC calls where the OL node may still be booting.
 const STARTUP_RPC_MAX_RETRIES: u16 = 10;
 
+/// Number of OL slots to advance per `get_blocks_summaries` request.
+const INBOX_FETCH_SLOT_WINDOW: u64 = 100;
+
+/// Splits an inclusive `[min_slot, max_slot]` slot range into contiguous
+/// inclusive windows spanning at most [`INBOX_FETCH_SLOT_WINDOW`] slots beyond
+/// their start.
+///
+/// Returns an empty list when `min_slot > max_slot`.
+fn inbox_slot_windows(min_slot: u64, max_slot: u64) -> Vec<(u64, u64)> {
+    let mut windows = Vec::new();
+    let mut window_start = min_slot;
+    while window_start <= max_slot {
+        let window_end = window_start
+            .saturating_add(INBOX_FETCH_SLOT_WINDOW)
+            .min(max_slot);
+        windows.push((window_start, window_end));
+        // Reached the end on this window; break before the increment so it
+        // can't overflow when `max_slot` is near `u64::MAX`.
+        if window_end == max_slot {
+            break;
+        }
+        window_start = window_end + 1;
+    }
+    windows
+}
+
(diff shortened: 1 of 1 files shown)


bin/alpen-client/src/rpc_client.rs
```diff
@@ -26,6 +26,32 @@ use tracing::info;
 /// Max retries for startup RPC calls where the OL node may still be booting.
 const STARTUP_RPC_MAX_RETRIES: u16 = 10;
 
+/// Number of OL slots to advance per `get_blocks_summaries` request.
+const INBOX_FETCH_SLOT_WINDOW: u64 = 100;
```
storopoli, line 35: Nit: since the windows are inclusive on both ends, each request actually spans `INBOX_FETCH_SLOT_WINDOW + 1` = 101 slots, not 100. Harmless at 101 << the default `max_headers_range` of 5000, but worth a word in the doc comment (e.g. "each window spans up to `INBOX_FETCH_SLOT_WINDOW + 1` slots inclusive") so nobody later sets this equal to a server limit and gets rejected by one slot.

bin/alpen-client/src/rpc_client.rs
```diff
@@ -26,6 +26,32 @@ use tracing::info;
...
+
+/// Splits an inclusive `[min_slot, max_slot]` slot range into contiguous
+/// inclusive windows spanning at most [`INBOX_FETCH_SLOT_WINDOW`] slots beyond
+/// their start.
+///
+/// Returns an empty list when `min_slot > max_slot`.
```
storopoli, line 41: Minor behavior change: an inverted range (`min_slot > max_slot`) previously reached the server and came back as an invalid-params error; now `get_inbox_messages` silently returns `Ok(vec![])`. All current callers go through `get_inbox_messages_checked`, which rejects inverted ranges first, so this is unreachable today — but it's a latent trap for a future direct caller. Consider returning an error (or noting this in the `get_inbox_messages` docs) instead of only documenting it here on the helper.

storopoli (approved):
non-blocking off-by-error and a behavior change.

ACK 55d5cd2ff1015af4128e59b5a198996715a584ed

bin/alpen-client/src/rpc_client.rs
sapinb: updated in 7801c32525a11a35e2afa32acd14b500f32d4757

bin/alpen-client/src/rpc_client.rs
sapinb: updated in 87474f4bf691f7a88a95ee2e0f31f18d490871b8

storopoli (approved):
ACK 87474f4bf691f7a88a95ee2e0f31f18d490871b8

PR #2124, opened 2026-07-16
chore(params): rotate testnet-prod OL sequencer pubkey (STR-3990) [draft]

madan-oss:


purusang:
@madan-oss I think it is not necessary to share private key to anyone. Not sure who else can see it. As long as its safe in secret manager, its not needed to be shared.

PR #2130, opened 2026-07-16
feat: OL sequencer restart from L1 DA

storopoli:
## Description

Lets a dead Strata sequencer be replaced from Bitcoin alone: a checkpoint-sync fullnode's `datadir` can be promoted into a working sequencer, with no copy of the old sequencer's block history.

Checkpoint sync and the sequencer already share the same sled schema, but a checkpoint-sync `datadir` was schema-compatible and data-incomplete. It holds finalized OL state, epoch summaries, and epoch indexing, and no OL block records past genesis. The sequencer path wants block records everywhere: startup walks the finalized chain, FCM links tips to the finalized base, block assembly and the chain worker read the parent block, and the checkpoint builder and prover derive their cursors from locally-authored payloads.

The way through is that the terminal block *header* is fully reconstructible from checkpoint DA even though the body is not. `OLBlockHeader` has eight fields; the sidecar's `TerminalHeaderComplement` plus the checkpoint tip plus the reconstructed state root produce all of them, and `verify_reconstruction` already did exactly this to check the block ID. So checkpoint sync now stores a header-only record at each epoch terminal, and an immutable `history_base` marker says "no block bodies at or below this". A one-shot, operator-requested, atomically-committed promotion turns such a `datadir` into a sequencer that builds its first block on the header-only anchor.

Recovery model, which is load-bearing and shouldn't be relaxed later:

- The old sequencer's `datadir` is assumed unrecoverable and is never read. No salvage path.
- The only input is L1-checkpoint-derived fullnode state.
- Anything not confirmed on L1 is permanently expendable, including a whole in-flight epoch. What isn't finalized on L1 isn't final.
- Promotion refuses during the confirmed-but-not-finalized window, and the error names the in-flight epoch, its checkpoint's L1 height, and the blocks remaining to the reorg-safe bury depth. Wait for finality, let checkpoint sync apply the epoch, promote at the later anchor.

Nothing changes for nodes without the marker. Every existing path, error, and message is preserved.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Nine commits, each self-contained and green, meant to be reviewed in order:

1. `feat(db)` — terminal-header schema, epoch-bearing `history_base` marker, slot-aware `BlockAvailability`, and the atomic `promote_to_history_anchor` primitive. Pure addition.
2. `feat(chain-worker)` — `reconstruct_terminal_header` extracted into `strata-checkpoint-types` (shared with dbtool), persisted on checkpoint apply before the summary so the summary stays the last-write applied marker.
3. `refactor` — `get_ol_header` added alongside full-block reads, and the consumers that only ever needed a header switched over. Byte-identical on full-history nodes.
4. `feat(strata)` — anchor-aware startup checks.
5. `feat(ol-checkpoint,prover)` — cursor floors.
6. `feat(dbtool)` — `backfill-terminal-headers` for `datadir`s predating commit 2.
7. `feat(strata)` — the `--bootstrap-from-checkpoint` flag.
8. `feat(rpc)` — pruned-history error semantics.
9. `test(functional)` — the end-to-end scenario.

Worth your attention:

- **Promotion atomicity** (commits 1 and 7). An earlier draft staged the writes and published the marker last. That is not crash-safe: a canonical tip pointing at a bodyless anchor without the marker is unbootable in every role. It's now one sled multi-tree transaction over the canonical and marker trees, so a crash leaves a plain checkpoint-sync `datadir`. Promotion validates everything before it writes anything, and every failure test asserts the marker is still absent and the canonical index untouched.
- **The terminal-state invariant** is `state.cur_epoch() == header.epoch() + 1`, not equality. Terminal processing advances the state into the next epoch, so equality would reject every valid anchor.
- **Cursor floors are derived, not seeded** (commit 5). There is no cursor row to seed, and copying the observed payload into the locally-authored table would insert an unsigned-checkpoint marker and resurface the already-finalized anchor as a fresh signing duty. Both cursors initialize as `max(last canonical local payload epoch, history_base epoch)`.
- **`tip == history_base` skips the parent read entirely** (commit 4). The anchor's parent has no record of any kind, so header fallback can't save it; the anchor is validated against its own summary, canonical epoch index, state root, and terminal-state invariant instead. No `BlockStatus` row is fabricated for it.

The functional test is the interesting one. It stops the sequencer, `rm -rf`s its `datadir`, promotes the checkpoint node with a replacement signer, and then asserts the first new block parents the header-only anchor, the dead fork's slots are re-authored with different blkids, epoch N+1's checkpoint finalizes with an L1 txid fetched back out of `bitcoind`, and finally that a fresh L1-only checkpoint node reconstructs the promoted epoch with matching summaries. That last phase closes the loop: the recovered sequencer's DA output is itself enough to bootstrap the next recovery.

Deferred on purpose: a functional test for the in-flight-checkpoint refusal. It needs mining paused inside the bury window and is timing-sensitive; the countdown error is unit-tested. Please advise if you have a good idea how to tackle this.

- [x]  Yes
- [ ]  No

- Design doc: [OL sequencer restart from L1 DA](https://app.notion.com/p/39d901ba000f809b9879f63735b43b42)

AI was used to plan and execute the several commits here with human checks and interventions on every step/commit.

## Checklist

## Related Issues

STR-3820.


crates/db/store-sled/src/ol/db.rs
```diff
@@ -155,6 +158,91 @@ impl OLBlockDatabase for OLBlockDBSled {
...
+    fn get_block_at(&self, commitment: OLBlockCommitment) -> DbResult<BlockAvailability> {
+        if let Some(block) = self.get_block_data(*commitment.blkid())? {
+            return Ok(BlockAvailability::Available(Box::new(block)));
+        }
+
+        match self.get_history_base()? {
+            Some(base) if commitment.slot() <= base.last_slot() => {
+                Ok(BlockAvailability::Unavailable)
+            }
+            _ => Ok(BlockAvailability::Missing),
```
prajwolrg: nit: `Unavailable` and `Missing` feels very similar. Maybe we can rename this to something distinct.

crates/storage/src/managers/ol.rs
```diff
@@ -275,6 +368,16 @@ impl OLBlockManager {
...
             .replace_canonical_suffix_from_blocking(start_slot, block_ids)
     }
+
+    /// Atomically publishes a checkpoint terminal as the immutable OL history anchor.
+    pub async fn promote_to_history_anchor_async(&self, anchor: EpochCommitment) -> DbResult<()> {
+        self.ops.promote_to_history_anchor_async(anchor).await
+    }
+
+    /// Atomically publishes a checkpoint terminal as the immutable OL history anchor.
+    pub fn promote_to_history_anchor_blocking(&self, anchor: EpochCommitment) -> DbResult<()> {
```
prajwolrg: nit: docstring mentions it's `publishing`. did you mean `promoting`?

crates/chain-worker/src/state.rs
```diff
@@ -867,6 +856,8 @@ pub(crate) struct AppliedEpochArtifacts {
     pub(crate) terminal: OLBlockCommitment,
```
prajwolrg: since we added `terminal_header`, do we still need `terminal` OLBlockCommitment?

crates/ol/block-assembly/src/resource_state.rs
```diff
@@ -199,6 +199,23 @@ async fn collect_epoch_blocks_until<C: BlockAssemblyAnchorContext>(
 
         let parent_id = *block.header().parent_blkid();
         blocks.push(block);
+
+        let parent_header = ctx
+            .fetch_ol_header(parent_id)
+            .await?
+            .ok_or(BlockAssemblyError::BlockNotFound(parent_id))?;
```
prajwolrg: nit: maybe have a new error for block header not found.

crates/ol/block-assembly/src/resource_state.rs
```diff
@@ -199,6 +199,23 @@ async fn collect_epoch_blocks_until<C: BlockAssemblyAnchorContext>(
...
+            if !parent_header.is_terminal() || parent_header.epoch() != epoch - 1 {
+                return Err(BlockAssemblyError::InvalidEpochBoundary {
+                    blkid: parent_header.compute_blkid(),
+                    expected_prev_epoch: epoch - 1,
+                    got_epoch: parent_header.epoch(),
+                    is_terminal: parent_header.is_terminal(),
+                });
+            }
+            break parent_header;
+        }
```
prajwolrg, line 218: the same logic is repeated twice. we should organize it to be better.

prajwolrg (approved):
Skimmed through the changes. Great work! Will do another pass once this is ready for review.

bin/strata/src/rpc/node.rs
krsnapaudel, line 1294: Does it make sense to call `self.history_unavailable_at_slot(start_height).await?;` here after the range check?

crates/db/store-sled/src/ol/db.rs
storopoli: Renamed `Unavailable` → `Pruned` in 84f77e7 — matches the pruned-history vocabulary the RPC layer and runbook already use; wire error messages unchanged.

crates/storage/src/managers/ol.rs
storopoli: Fixed in 84f77e7 — "publishes" → "promotes" on both variants.

crates/chain-worker/src/state.rs
storopoli: Good catch — it was redundant. Dropped the field in eb933e7; the commitment is now derived via an `AppliedEpochArtifacts::terminal()` accessor from the reconstructed header, so there is no stay-in-sync invariant anymore.

crates/ol/block-assembly/src/resource_state.rs
storopoli: Added a distinct `HeaderNotFound` variant in 2239a12; header-probe misses use it, full-body reads keep `BlockNotFound`.

crates/ol/block-assembly/src/resource_state.rs
storopoli: Deduplicated in 2239a12 — both branches now go through a single `ensure_epoch_boundary` helper.

bin/strata/src/rpc/node.rs
```diff
@@ -161,6 +171,48 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
...
+                Err(block_history_unavailable_error(history_base.last_slot()))
+            }
+            BlockAvailability::Missing => Err(not_found_error(format!(
+                "block not found: {}",
+                commitment.blkid()
+            ))),
+        }
+    }
+
+    async fn history_unavailable_at_slot(&self, slot: u64) -> RpcResult<()> {
```
krsnapaudel: Should this be named `ensure_history_available_at_slot`?

bin/strata/src/rpc/node.rs
krsnapaudel, line 249: Do we need `self.history_unavailable_at_slot(summary.terminal().slot()).await?;` here?

krsnapaudel, line 1357: Can this run into history unavailable? If we do, should we break?

krsnapaudel, line 371: Do we need to check history is available here?

bin/strata/src/startup_checks.rs
```diff
@@ -397,6 +554,157 @@ fn verify_previous_epoch_summary_for_tip(storage: &NodeStorage, tip_block: &OLBl
...
+    }
+    if tip_commitment.slot() == anchor_commitment.slot() && tip_commitment != anchor_commitment {
+        bail!(
+            "startup: canonical OL tip at history-base slot does not match anchor: expected {anchor_commitment}, got {tip_commitment}"
+        );
+    }
+
+    if tip_commitment == anchor_commitment {
+        let anchor_header = verify_history_anchor(storage, history_base)?;
+        return verify_previous_epoch_summary_for_tip_header(storage, &anchor_header);
```
krsnapaudel, line 699: `setup_history_anchor_with` hardcodes `epoch = 1`
   (`startup_checks.rs:1126`), so `previous_epoch = 0 = genesis` — always present.
   This check therefore passes vacuously and nothing exercises the real `H-1` lookup. Should we add a test test for H >1?

bin/strata-dbtool/src/cmd/terminal_header.rs
```diff
@@ -0,0 +1,382 @@
...
+                "Failed to store terminal-header record at epoch {epoch}"
+            ))?;
+        report.headers_written += 1;
+    }
+
+    Ok(report)
+}
+
+#[cfg(test)]
+mod tests {
```
krsnapaudel, line 159: `stored_state_root_mismatch_is_hard_error` Does it make sense to add a test for this?

krsnapaudel (approved):
Overall looks great. Nothing blocking. Some nits.

crates/db/store-sled/src/ol/db.rs
storopoli: Update I've squashed this into the original commits.

crates/storage/src/managers/ol.rs
storopoli: Update I've squashed this into the original commits.

crates/chain-worker/src/state.rs
storopoli: Update I've squashed this into the original commits.

crates/ol/block-assembly/src/resource_state.rs
storopoli: Update I've squashed this into the original commits.

crates/ol/block-assembly/src/resource_state.rs
storopoli: Update I've squashed this into the original commits.

bin/strata/src/rpc/node.rs
storopoli: Yeah, makes sense. Added the check right after the range validation, so a range that dips below the history base now fails immediately with the pruned-history error instead of blowing up halfway through the walk.

bin/strata/src/rpc/node.rs
storopoli: Renamed. The old name read like a predicate when it is really a guard.

bin/strata/src/rpc/node.rs
storopoli: We do. Without it, epochs at or below the anchor walk into pruned bodies and come back with a confusing "block not found", since the lookup is by id and cannot tell pruned from missing. Added the check on summary.terminal().slot() at the top.

storopoli: It can, yes. Made it break instead of erroring: the walk grabs the history base once up front and stops after the block just above it, because the next parent would be the bodyless anchor. You get a partial list back, same as when the walk reaches genesis.

storopoli: Added the check on start_slot before the walk. This one has to return the complete range to be useful, so failing fast beats handing back something partial.

bin/strata/src/startup_checks.rs
storopoli: Good catch, the check really was passing for free. Added a fixture that puts the anchor at epoch 2 with an actual epoch-1 summary behind it, plus a variant where that summary is missing so the error path finally fires for real.

bin/strata-dbtool/src/cmd/terminal_header.rs
storopoli: Done, added stored_state_root_mismatch_is_hard_error_without_write. It stores an anchor state whose root disagrees with the summary and checks the command bails without writing anything.

bin/strata/src/rpc/node.rs
```diff
@@ -190,6 +242,9 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
         &self,
         summary: &EpochSummary,
     ) -> RpcResult<L2BlockCommitment> {
+        self.ensure_history_available_at_slot(summary.terminal().slot())
+            .await?;
```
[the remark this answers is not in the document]
storopoli: Good catch, this one was a regression from the earlier review round. The l2_start branch now checks the history base and just returns None for epochs at or below it, which is what checkpoint-sync nodes already do in the same situation. The summary and l2_end keep being served. The test now pins the degraded response instead of the error.

crates/db/store-sled/src/ol/db.rs
```diff
@@ -209,6 +295,10 @@ impl OLBlockDatabase for OLBlockDBSled {
         Ok(self.blk_height_tree.get(&slot)?.unwrap_or(Vec::new()))
     }
 
+    fn get_highest_block_slot(&self) -> DbResult<Option<Slot>> {
+        Ok(self.blk_height_tree.last()?.map(first))
```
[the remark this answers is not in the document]
storopoli: Right, fixed in two places. del_block_data now drops the height row when the last block at a slot goes away, and get_highest_block_slot walks backward past any empty rows, so datadirs that picked up ghost rows from older deletes still classify fine. Tests cover both paths.

crates/consensus-logic/src/fcm/service.rs
```diff
@@ -573,8 +573,8 @@ where
     S: FcmStorage + ?Sized,
 {
     let mut best_tip = *cur_tip;
-    let mut best_block = storage
-        .get_ol_block(best_tip)
+    let mut best_header = storage
+        .get_ol_header(best_tip)
         .await?
         .ok_or(Error::MissingOLBlock(best_tip))?;
```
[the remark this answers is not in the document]
storopoli: Applicable, yes. A reorg that pivots at the anchor would have died in get_block_slot, since its fallback read the full block. It only ever needed the slot, so it now reads through get_ol_header and the terminal-header record covers the anchor case. Added a test where the pivot exists only as a header record.

bin/strata/src/rpc/node.rs
```diff
@@ -1248,13 +1344,16 @@ impl<P: OLRpcProvider> OLFullNodeRpcServer for OLRpcServer<P> {
             .ok_or_else(|| internal_error("OL sync status not available"))?
             .tip()
             .blkid();
+        let history_base = self.provider.get_history_base().await.map_err(db_error)?;
```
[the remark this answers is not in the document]
storopoli: Right, the walk only breaks after the first fetch succeeds, so a freshly promoted node with zero authored blocks trips over its own tip. Added a pre-loop check: when the sync-status tip is the anchor blkid, return an empty list, which matches the stop-before-anchor behavior once blocks exist. Test covers the tip-at-anchor case.

delbonis (approved):
Full disclosure: I did not read the code at all.

delbonis:
All of this will have to be removed in a couple months so I don't see a point in spending mental energy to review it if it is functionally complete.

bewakes (approved):

PR #2142, opened 2026-07-22
docs(alpen-cli): fix stale and inaccurate command help text

storopoli:
## Description

Fixes the stale `deposit` help text reported during testnet testing on Slack
(it claimed a hardcoded 10 BTC deposit), and cleans up other help text that
didn't match what the code does:

- `deposit`: the amount is one bridge denomination plus the bridge fee, both
  taken from params/config, not a fixed 10 BTC. The help now mirrors the
  wording `withdraw` already uses.
- `receive`: for `alpen` the command returns the wallet's default signer
  address, which never changes. Only the signet path derives a new address,
  so the help no longer promises a "new" one.
- `--fee-rate` (deposit, drain, recover, send): the docs said "must be >=1",
  but `get_fee_rate` clamps low values to the broadcast minimum instead of
  rejecting them. The help now says the effective rate is at least 1.

### Type of Change

- [x] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [x] Dependency Update

## Notes to Reviewers

Help-text only, no runtime changes.

- [ ]  Yes
- [x]  No

## Checklist

AI assistance: written with Claude Code and Codex CLI; all changes reviewed
by me.

### DIFF
--- a/bin/alpen-cli/src/cmd/deposit.rs
+++ b/bin/alpen-cli/src/cmd/deposit.rs
@@ -35,7 +35,8 @@ use crate::{
     signet::{get_fee_rate, log_fee_rate, SignetWallet},
 };
 
-/// Deposits 10 BTC from signet into Alpen
+/// Deposits one bridge denomination of BTC plus the bridge fee from signet into Alpen.
+/// The bridge denomination is configured in the `config.toml` file.
 #[derive(FromArgs, PartialEq, Debug)]
 #[argh(subcommand, name = "deposit")]
 pub struct DepositArgs {
@@ -44,7 +45,7 @@ pub struct DepositArgs {
     #[argh(positional)]
     alpen_address: Option<String>,
 
-    /// override signet fee rate in sat/vbyte. must be >=1
+    /// override signet fee rate in sat/vbyte; the effective rate is at least 1
     #[argh(option)]
     fee_rate: Option<u64>,
 }
--- a/bin/alpen-cli/src/cmd/drain.rs
+++ b/bin/alpen-cli/src/cmd/drain.rs
@@ -34,7 +34,7 @@ pub struct DrainArgs {
     #[argh(option, short = 'r')]
     alpen_address: Option<String>,
 
-    /// override signet fee rate in sat/vbyte. must be >=1
+    /// override signet fee rate in sat/vbyte; the effective rate is at least 1
     #[argh(option)]
     fee_rate: Option<u64>,
 }
--- a/bin/alpen-cli/src/cmd/recover.rs
+++ b/bin/alpen-cli/src/cmd/recover.rs
@@ -22,7 +22,7 @@ use crate::{
 #[derive(FromArgs, PartialEq, Debug)]
 #[argh(subcommand, name = "recover")]
 pub struct RecoverArgs {
-    /// override signet fee rate in sat/vbyte. must be >=1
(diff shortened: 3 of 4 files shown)


prajwolrg (approved):

bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -35,7 +35,8 @@ use crate::{
     signet::{get_fee_rate, log_fee_rate, SignetWallet},
 };
 
-/// Deposits 10 BTC from signet into Alpen
+/// Deposits one bridge denomination of BTC plus the bridge fee from signet into Alpen.
+/// The bridge denomination is configured in the `config.toml` fil.
```
pramodkandel: nit: misspelling of file

pramodkandel: To make it slightly more understandable, I'd rewrite it something like --
```
/// Deposits BTC from Signet into Alpen in the fixed bridge denomination amount,
///  plus the bridge fee.
/// The bridge denomination amount is configured in the `config.toml` file.
```

bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -44,7 +45,7 @@ pub struct DepositArgs {
     #[argh(positional)]
     alpen_address: Option<String>,
 
-    /// override signet fee rate in sat/vbyte. must be >=1
+    /// override signet fee rate in sat/vbyte; the effective rate is at least 1
```
pramodkandel, line 48: Why this change? For the CLI user, the previous one was probably clearer, saying the value to this argument must be >=1

storopoli: As I've said in the third bullet point in the PR description:

> `--fee-rate` (deposit, drain, recover, send): the docs said "must be >=1",
  but `get_fee_rate` clamps low values to the broadcast minimum instead of
  rejecting them. The help now says the effective rate is at least 1.

bin/alpen-cli/src/cmd/deposit.rs
storopoli: Will fix. Thanks for noticing. Good catch.

storopoli: Amended in 77176139ed66e634130d452c5561768f7a7290d1

pramodkandel (approved):

krsnapaudel (approved):

bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -35,7 +35,8 @@ use crate::{
     signet::{get_fee_rate, log_fee_rate, SignetWallet},
 };
 
-/// Deposits 10 BTC from signet into Alpen
+/// Deposits one bridge denomination of BTC plus the bridge fee from signet into Alpen.
+/// The bridge denomination is configured in the `config.toml` file.
```
barakshani, line 39: this needs to be in sync with the docs

barakshani: I think we'd want this to be consistent with the language [here](https://github.com/alpenlabs/alpen/blob/77176139ed66e634130d452c5561768f7a7290d1/bin/alpen-cli/src/cmd/withdraw.rs#L28).

PR #2144, opened 2026-07-22
STR-2653 deborsh v2

delbonis:
## Description

This PR is a sweeping removal of all Borsh uses in this repo.  Unfortunately there's still some usages in the asm repo that need to be removed separately, so it's not entirely purged, but it's most of the way there.

Most types were converted to use CBOR, which also gets us the benefit of supporting backwards-compatibility in future versions.  Protocol types were converted to use their native encodings in some places, or were wrapped in dumb byte container.

Some data types, like zkaleido's receipt types, were converted to be dumber byte containers so that `zkaleido` could be removed from the core database crate hierarchy.  The receipts are now parsed where they are used, which should not normally fail since these are only ever written by ourselves.

There's also 18 tickets that were sketched out by Claude in [this Notion doc](https://app.notion.com/p/DB-Refactor-2026-Index-3b4901ba000f80f3bf7efa627c4605b9) for followup fixes, some of them immediately pertinent to the ongoing upgradability work.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist


delbonis:
At this point I've done some of the preliminary work to survey what concretely needs to be done, but haven't started going through and fixing these issues identified.

While going through and doing survey work I removed borsh from a bunch of types and cleaned up *some* serde serializations, but there's still a lot of work to do here.

delbonis:
> +1,891 -4,471

:pray:

delbonis:
Okay this PR is in a nearly-mergeable state.

I see there's a CI issue that needs to be resolved (since we removed `strata-state`) and need to create tickets for all the DB Refactor tickets that we defined.

crates/db/store-sled/src/mmr_index/schemas.rs
```diff
@@ -14,13 +17,15 @@ define_table_without_codec!(
...
     /// MMR index leaf count storage: mmr_id -> leaf count
     (MmrIndexLeafCountSchema) RawMmrId => u64
 );
+impl_raw_bytes_key_codec!(MmrIndexLeafCountSchema, RawMmrId);
+impl_codec_value_codec!(MmrIndexLeafCountSchema, u64);
 
-impl_borsh_value_codec!(MmrIndexNodeSchema, Buf32);
-impl_borsh_value_codec!(MmrIndexPreimageSchema, Vec<u8>);
+impl_codec_value_codec!(MmrIndexNodeSchema, Buf32);
+impl_opaque_value_codec!(MmrIndexPreimageSchema, Vec<u8>);
```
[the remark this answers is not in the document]
delbonis: The purpose of this PR is to enable upgradability into the future.

crates/db/store-sled/src/asm/schemas.rs
```diff
@@ -1,30 +1,25 @@
...
-// ASM state per block schema and corresponding codecs implementation.
-//
-// `AnchorState` is SSZ-only — it carries no Borsh impls — so this table spells
-// out its codecs instead of using `define_table_with_seek_key_codec!`.
 define_table_without_codec!(
     /// A table to store ASM state per l1 block.
     (AsmStateSchema) L1BlockCommitment => AnchorState
 );
 impl_bincode_key_codec!(AsmStateSchema, L1BlockCommitment);
 impl_ssz_value_codec!(AsmStateSchema, AnchorState);
```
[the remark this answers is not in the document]
delbonis: The purpose of this PR is to enable upgradability into the future, we are expected to break database schemas right now.

Cargo.toml
```diff
@@ -409,7 +407,6 @@ bitcoind-async-client = { version = "0.10.8", default-features = false, features
   "raw_rpc",
 ] }
 bitvec = { version = "1.0.1", features = ["serde"] }
-borsh = { version = "1.6.1", features = ["derive"] }
```
storopoli, line 412: YES!!!!

functional-tests/tests/strata/test_crash_fcm_new_block.py
storopoli: this why func-tests are failing:


```suggestion
        assert_ol_block_status(datadir, block_id, "valid")
```

bin/alpen-client/src/prover/storage.rs
```diff
@@ -183,7 +206,8 @@ impl EeBatchProofDbManager {
     }
 
     pub(crate) fn get_proof_by_id(&self, proof_id: ProofId) -> Option<Proof> {
-        let receipt = self.db.get_acct_proof_by_id(proof_id).ok().flatten()?;
+        let entry = self.db.get_acct_proof_by_id(proof_id).ok().flatten()?;
```
storopoli: Not so severe, but this maps a malformed stored receipt to `None` via `.ok()?`. The corruption becomes indistinguishable from absence. Other receipt-decode sites propagate errors properly.

bin/alpen-client/src/prover/storage.rs
```diff
@@ -183,7 +206,8 @@ impl EeBatchProofDbManager {
...
+        let receipt = decode_receipt(&entry).ok()?;
```
storopoli: `decode_receipt(...).ok()?` silently maps a malformed stored receipt to `None`, while `has_proof` above reports readiness from row presence alone (and its `unwrap_or(false)` swallows DB errors too). Net effect in `batch_prover.rs`: `check_proof_status` returns `Ready` for a row that `get_proof` then answers `Ok(None)` for — callers poll an unretrievable "ready" proof forever, with no log line. This path is conspicuously the only unlogged one; there's already a `debug!` for the analogous "Completed but not in DB" case in `batch_prover.rs:110`.

Suggest at minimum a `warn!` on the decode-failure arm, and ideally making the two methods agree (e.g. `has_proof` validates decodability, or both return `ProverResult` and propagate).

crates/csm-types/src/payload.rs
```diff
@@ -182,35 +93,6 @@ impl L1Payload {
...
-    fn serialize<W: Write>(&self, writer: &mut W) -> io::Result<()> {
-        BorshSerialize::serialize(&self.data, writer)?;
-        BorshSerialize::serialize(&self.tag.subproto_id(), writer)?;
-        BorshSerialize::serialize(&self.tag.tx_type(), writer)?;
-        BorshSerialize::serialize(&self.tag.aux_data().to_vec(), writer)?;
-        Ok(())
-    }
-}
-
-impl BorshDeserialize for L1Payload {
```
storopoli, line 201: This deleted impl deliberately routed decode through `L1Payload::new` to preserve the `MAX_ENVELOPE_PAYLOAD_SIZE` invariant (per the comment above it). The derived `Deserialize` that replaces it constructs `data` directly, so the total-size check is now bypassed for every CBOR row read back from the writer/intent tables. (`TagData` still validates — its upstream `Deserialize` goes through `TagData::new` — so it's only the payload-size invariant that's lost.) A corrupt or hand-crafted row can materialize an oversized protocol payload.

`#[serde(try_from = "...")]` (or a manual `Deserialize` routed through `L1Payload::new`) would restore the parse-don't-validate property; an oversized-row rejection test would pin it.

crates/alpen-ee/database/src/serialization_types/account_state.rs
```diff
@@ -24,13 +24,15 @@ impl DBAccountStateAtEpoch {
...
 
-// TODO(STR-3421): Migrate EE account-state persistence away from this Borsh
-// mirror and store the SSZ account-state type directly, including any needed
-// DB compatibility/versioning path for existing local data.
-#[derive(Debug, Clone, BorshSerialize, BorshDeserialize, PartialEq)]
+// TODO(STR-3421): Migrate EE account-state persistence away from this serde mirror and store
+// the SSZ account-state type directly, including any needed DB compatibility/versioning path
+// for existing local data.
+#[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize)]
 pub(crate) struct DBEeAccountState {
```
storopoli, line 35: Two issues with this mirror:

1. The SSZ type bounds `pending_inputs`/`pending_fincls`, but the mirror uses unbounded `Vec`s and converts back through an infallible `From` into `EeAccountState::new`, whose `try_into().expect(...)` capacity checks then panic (`crates/ee-acct-types/src/state.rs:19`). So an oversized/corrupt CBOR row panics the node on a DB read instead of surfacing a decode error. The conversion should be fallible (`TryFrom`) with both list bounds enforced before construction.

2. The `[u8; 32]` mirror fields don't achieve what `TODO(db-refactor-part-17)` intends: serde serializes `[T; N]` via `serialize_tuple`, which ciborium forwards to `serialize_seq` — an array of integers, ~63 bytes for 32 bytes. That's barely better than the 66-byte hex string it replaced and far from a 34-byte CBOR byte string. `#[serde(with = "serde_bytes")]` is needed either way, at which point the mirror split buys nothing.

crates/db/types/src/chunked_envelope.rs
```diff
@@ -20,7 +21,8 @@ use crate::DbResult;
 /// (`magic + version`); each subsequent P2TR output funds a
 /// reveal whose tapscript carries one chunk under `<sequencer_pk> OP_CHECKSIG`.
 /// Reveals do NOT reference each other; entries are independent across batches.
-#[derive(Debug, Clone, PartialEq, BorshSerialize, BorshDeserialize)]
+// FIXME(trey): this merges information from different stages into a single entry, which creates issues where multiple services are writing to the same database object
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
 pub struct ChunkedEnvelopeEntry {
     /// Raw chunk payloads, ordered by commit-output index.
     pub chunk_data: Vec<Vec<u8>>,
```
storopoli: Heads-up on CBOR sizing for this field (and `RevealTxMeta.tx_bytes` below): ciborium has no `Vec<u8>` specialization — serde's `Serialize for Vec<T>` goes through `serialize_seq`, which emits a CBOR *array of integers*, ≈1.9 bytes per payload byte. This entry carries the largest blobs in the system (full DA chunk data), so the on-disk cost roughly doubles vs the old borsh encoding. The same applies to `BundledPayloadEntry`/`IntentEntry` via `L1Payload.data` (up to ~395 KB, stored in two tables).

`#[serde(with = "serde_bytes")]` on the byte fields — or an opaque/strata-codec value codec for the blob-bearing entries — gets proper CBOR byte strings. Worth measuring on a full-size DA payload before deciding to defer.

crates/db/store-sled/src/ol_checkpoint/db.rs
```diff
@@ -4,11 +4,21 @@ use strata_csm_types::CheckpointL1Ref;
...
 use super::schemas::*;
 use crate::define_sled_database;
+use crate::utils::conv_sled_err;
+
+/// The lowest [`EpochCommitment`] key belonging to `epoch`.
+///
+/// Keys encode as `epoch(4 BE) ‖ last_slot(8 BE) ‖ last_blkid(32)`, so tree order matches
+/// `(epoch, last_slot, last_blkid)` order and this value is the inclusive start of the range
+/// covering `epoch` and every epoch after it.
+fn epoch_lower_bound(epoch: Epoch) -> EpochCommitment {
```
storopoli, line 19: The new `range()`/`last()` queries in this file are correct only because `EpochCommitment`'s strata-codec key encoding is fixed-width big-endian — an invariant asserted in a comment in `schemas.rs` but pinned by no test. The existing range tests in `crates/db/tests/src/ol_checkpoint_tests.rs` only use epochs 0..4, which sort identically under LE and BE, so they can't catch a regression (contrast with the byte-layout pin tests added for `CheckpointProofTask`, `MmrId`, and `DBBatchId`/`DBChunkId` — same discipline wanted here).

Two asks: (a) a byte-layout pin test for `EpochCommitment` as a table key plus one range test spanning a byte boundary (e.g. epochs 255/256); (b) confirm the tie-break change in the `get_last_*` queries is intended — `last()` returns the greatest `(epoch, last_slot, last_blkid)` among competing same-epoch commitments, where the old scan kept the first max-epoch commitment encountered. That changes which reorged checkpoint gets reported; whoever owns checkpoint signing/inclusion should sign off, and a test with two commitments at the same epoch would pin it.

storopoli (changes requested):
This looks really good. Some human and AI findings that I've combed through and verified

crates/db/types/src/chunked_envelope.rs
storopoli: I know that this will probably be ported to `alpen-ee` but it's worth flagging since this might vanish if #2194 goes in, despite it not removing this current file.

bin/alpen-client/src/prover/storage.rs
delbonis: Oh you know what, it did this because this fn should have returned a Result in the first place.

crates/alpen-ee/database/src/serialization_types/account_state.rs
delbonis: ~~I should just have this store the SSZed structure, maybe.~~

This would actually make sense to use the new versioning system coming in a future PR.

bin/alpen-client/src/prover/storage.rs
delbonis: Yeah this should return `Result<Option<Proof>, ...>`.

crates/db/store-sled/src/ol_checkpoint/db.rs
delbonis: Ideally we could use the new `ArrayKey` stuff for this, but that should be done in a followup PR.

delbonis:
GitHub seems to have concluded that I wanted to leave those comments as a review on my own PR, which is naturally the most sensible thing for me to do.

crates/alpen-ee/database/src/serialization_types/account_state.rs
delbonis: But for now I'll convert things over to using `Buf32` or similar, which should "do the right thing".

crates/db/types/src/chunked_envelope.rs
delbonis: Fixed in recent commits, changed to use serde_bytes types.

crates/alpen-ee/database/src/serialization_types/account_state.rs
delbonis: Fixed in recent commits, changed to use serde_bytes types.

crates/csm-types/src/payload.rs
delbonis: Changed with custom deserialize impl.  This should be fixed more thoroughly in a future refactor of all this stuff.

bin/alpen-client/src/prover/storage.rs
delbonis: Fixed in recent commit with changes.

purusang (approved):

prajwolrg (approved):
LGTM! Thanks for removing a bunch of deprecated code.

crates/db/store-sled/src/ol_checkpoint/db.rs
storopoli: please don't forget to track this in a ticket and preferrable with a `TODO` comment?

storopoli:
> **MEDIUM — MMR leaf-count key encoding changed while node/preimage keys were deliberately kept on-disk compatible, so an existing index silently reports 0 leaves**
> 
> `crates/db/store-sled/src/mmr_index/schemas.rs:24`
> 
> `MmrIndexLeafCountSchema` moved from `define_table_with_seek_key_codec!` (bincode fixint/BE: `len(u64 BE) ‖ bytes`) to `impl_raw_bytes_key_codec!` (bytes verbatim), but the sibling `(RawMmrId, NodePos)` / `(RawMmrId, LeafPos)` key codecs in the same file were _intentionally_ hand-written to preserve the old framing — the comment at line 36 states "existing databases need no migration" and tests pin it. Result on upgrade of an existing datadir: node and preimage rows stay reachable, but every leaf-count row is orphaned, and `MmrIndexDb::get_leaf_count` (`mmr_index/db.rs:87`) fails **open** with `unwrap_or(0)`. The ASM manifest / snark-account-inbox / L1-block-refs MMRs then re-append from position 0 over live node data, producing a divergent accumulator root rather than a clean startup error.
> 
> Fix: either restore the framed key encoding for `MmrIndexLeafCountSchema` (`encode_scoped_key`-style `len(u64 BE) ‖ bytes`) so it matches the layout the rest of the file pins, or add an explicit datadir version marker that refuses to start on a pre-refactor MMR index instead of defaulting the count to zero.
> 
> **LOW — `impl_opaque_value_codec!` replaced `impl_borsh_value_codec!` for `Vec<u8>` values on tables whose keys did not change, so old rows decode with borsh's 4-byte length prefix as payload data**
> 
> `crates/db/store-sled/src/mmr_index/schemas.rs:28`
> 
> Borsh encodes `Vec<u8>` as `u32 LE length ‖ bytes`; the new identity codec returns the stored bytes verbatim. For `MmrIndexPreimageSchema` (key layout explicitly preserved) and for `BytecodeSchema` / `ExecBlockPayloadSchema` in `crates/alpen-ee/database/src/sleddb/schema.rs:111,58` (borsh and `strata-codec` both encode a 32-byte `Hash` key identically, so those rows are still reachable), a pre-upgrade row now reads back 4 bytes longer and corrupted. A cached contract bytecode read via `get_bytecode` no longer hashes to its content-addressed key, silently poisoning the chunk-prover witness; an MMR preimage read back this way yields a wrong leaf hash. Unlike the CBOR/SSZ codecs elsewhere in this PR, this path cannot fail closed.
> 
> Fix: keep a length-framed value codec for these tables, or gate startup on a datadir version so pre-refactor rows are rejected rather than reinterpreted.

We can disregard these since it is a DB compatibility that we Don't want to support. This is a breaking change.

storopoli:
> ### 💡 Codex Review
> https://github.com/alpenlabs/alpen/blob/92d7018cadba1dbb0aaf40ab5d24b701b15caf7b/crates/db/store-sled/src/ol_checkpoint/db.rs#L409-L412
> 
> 
> **P2 Badge Prune the checkpoint epoch index on cascaded L1-ref deletes**
> Once this path populates `l1_ref_epoch_index_tree`, every public removal of the same `l1_ref_tree` row has to update that index. `del_checkpoint_payload_entry` and `del_checkpoint_payload_entries_from_epoch` still cascade by removing from `l1_ref_tree` only, so after pruning a checkpoint payload that had an L1 ref, `get_observed_checkpoint_commitments_for_epoch` can return the stale candidate from the index even though `get_checkpoint_l1_ref` no longer has it. Please prune the epoch index in those cascade deletes (or filter indexed candidates against `l1_ref_tree`) so the DB API stays coherent.
> 
> ℹ️ About Codex in GitHub

This is pre-existing before this PR. How you wanna handle this @delbonis?

storopoli (approved):
ACK 92d7018cadba1dbb0aaf40ab5d24b701b15caf7b

Let's fix the ticketless `TODO` comments that the linter is failing on CI to get this merged.

crates/db/store-sled/src/ol_checkpoint/db.rs
storopoli: unresolving so we don't forget this

bewakes (approved):
This is a super nice cleanup!

krsnapaudel (approved):

delbonis:
Okay this should be good to go, if everything passes.

delbonis:
Aaaaand github shit the bed again it seems.

storopoli (commented):
ACK f9b7e23a66ff241a1333854338470fd56f814235

storopoli:
re-running CI again once it goes green we'll merge this.
Nice job! Thanks! :)

PR #2145, opened 2026-07-23
Alpen EE L1 Recoverability

irnb:
# EE sequencer recovery from L1

> [!NOTE]
> The goal of this PR and its sibling [e2e-runner PR #9](https://github.com/alpenlabs/e2e-runner/pull/9) is to demonstrate that we can reconstruct EE state from the data published on L1, materialize a fresh EE datadir, and continue sequencing from that state. This is a feasibility implementation. Some mitigations in this branch may not match our long-term node architecture.
>
> The e2e scenario uses native proof execution. The native hosts and SP1 guests call the same checkpoint, chunk, and account proof-program functions, so the run exercises the recovery-sensitive proof statements and OL acceptance path. It skips SP1 zkVM execution, Groth16 generation, remote prover integration, and verification-key deployment. Those SP1-specific layers do not change the state-reconstruction and sequencer-continuation result demonstrated here.
>
> The end-to-end run exposed major blockers in sparse startup, synchronization, batch processing, proof-input binding, and RPC history. We resolved each blocker far enough to complete the recovery flow and confirmed that EE DA on L1 contains the state data required to reconstruct and continue the sequencer. This answers the P0 feasibility question, so the remaining production work has moved to P2. I am keeping this PR as a reference implementation and a record of the blockers we found. The branch contains one atomic commit for the recovery foundation and one atomic commit for each blocker. The reviewer notes below describe each blocker and its current mitigation.
>
> The next recovery effort will focus on promoting a synced EE full node to sequencer. A full node already has the canonical execution state and history, which makes promotion a more practical recovery path than constructing a sparse sequencer database during an incident.
>
> The proof-statement changes in this PR will move into the next soundness epic. If we return to sparse L1 recovery, we can use this branch as the starting point and replace each feasibility mitigation with a production design that matches our long-term node architecture.

## Summary

This PR demonstrates that an EE sequencer can:

1. read the ordered finalized EE DA transaction IDs from a stopped sequencer database;
2. hydrate those transactions from Bitcoin and replay the published state diffs;
3. validate the reconstructed `EeAccountState` against the proof-backed `inner_state_root` accepted by OL;
4. materialize the reconstructed Ethereum state into a fresh sparse Reth database;
5. bootstrap the minimum EE metadata at the same recovery anchor; and
6. start a replacement sequencer that produces new blocks, batches, and OL-accepted account updates.

The state payload used for reconstruction comes from EE DA on Bitcoin. The current command still uses two indexes:

- the stopped source EE database provides the ordered finalized DA transaction IDs; and
- a synced OL node provides the accepted account state and execution commitment used to verify the result.

Scanning Bitcoin to discover and order EE DA transactions from scratch is outside this PR. The source database contributes transaction IDs, not execution state, to the recovered datadir.

The sibling [e2e-runner PR #9](https://github.com/alpenlabs/e2e-runner/pull/9) owns stack orchestration and the end-to-end assertions. This PR contains the recovery materializer and the runtime changes exposed by starting a sequencer from the sparse database.

## Recovery flow

```text
stopped source EE datadir
        │
        ├── ordered finalized EE DA transaction IDs
        ▼
Bitcoin commit/reveal transactions
        │
        ├── replay BatchStateDiff values
        ▼
reconstructed EeAccountState + Ethereum state
        │
        ├── verify against OL inner_state_root and execution commitment
        ▼
temporary Reth + EE databases
        │
        ├── validate paired anchors
        ├── publish output
        ▼
fresh sparse EE datadir
        │
        └── start replacement sequencer and continue proving
```

`strata-test-cli ee-state-recover` owns the full materialization operation:

1. extract transaction IDs through the requested update sequence;
2. fetch and decode the transactions from Bitcoin;
3. replay the DA state diffs;
4. load and verify the proof-backed target from OL;
5. import the Ethereum state and anchor header into a fresh Reth database;
6. bootstrap the paired EE database at the same anchor;
7. validate the completed output; and
8. publish the fresh datadir.

## End-to-end validation

The sibling e2e-runner scenario performs this sequence with native proof execution:

1. Start Signet, ASM, OL, and a source EE sequencer.
2. Submit a mock-bridge deposit and use the funded EVM account for value and storage-changing transactions.
3. Wait for an EE account update covering those transactions to reach OL and L1.
4. Stop the source EE sequencer.
5. Recover a fresh datadir with `ee-state-recover`.
6. Start a replacement sequencer from that datadir.
7. Verify the reconstructed contract storage and account balance.
8. Call `eth_gasPrice` before history warm-up or new transactions.
9. Produce and prove 256 canonical post-recovery blocks for the EVM `BLOCKHASH` visibility window.
10. Submit new bridge deposits to the recovered EE.
11. Run the selected Prague execution-spec tests against the recovered EE.
12. Submit a withdrawal, wait for its account update, and observe its ASM Bridge V1 assignment.

One completed run recovered target update sequence 19, started the replacement EE at execution block 61, and produced the canonical suffix through block 317. OL accepted the account update covering that warm-up at sequence 105. The complete recovery-focused Prague EEST selection finished with **281 passed and 23 skipped**. The recovered sequencer then submitted a withdrawal covered by OL update sequence 264, and ASM assigned it to a bridge operator.

A later run repeated the scenario with an immediate post-startup `eth_gasPrice` assertion, before warm-up, deposits, or EEST. This exercises the sparse gas-oracle boundary rather than allowing post-recovery traffic to hide it.

Native mode executes the same `process_ol_stf`, `process_ee_chunk`, and `process_ee_acct_update` functions used by the SP1 guests, with deterministic test predicates. The run therefore checks recovered proof inputs, chunk and account transitions, proof-statement consistency, and OL acceptance. SP1 zkVM execution, Groth16 generation, remote prover behavior, and verification-key deployment remain outside this test. They are SP1 integration and deployment coverage rather than missing coverage of the recovery data flow.

## Run the full flow locally

The flow requires Docker, Rust, `uv`, `jq`, and `curl`. Clone the two PR branches and the Alpen execution-spec-tests fork into one directory:

```bash
RECOVERY_WORKDIR=/absolute/path/to/ee-recovery-demo

mkdir -p "${RECOVERY_WORKDIR}"
git clone --single-branch --branch alpen-ee-recovery \
  git@github.com:alpenlabs/alpen.git \
  "${RECOVERY_WORKDIR}/alpen"
git clone --single-branch --branch test/ee-l1-recovery \
  git@github.com:alpenlabs/e2e-runner.git \
  "${RECOVERY_WORKDIR}/e2e-runner"
git clone https://github.com/alpenlabs/execution-spec-tests.git \
  "${RECOVERY_WORKDIR}/execution-spec-tests"
git -C "${RECOVERY_WORKDIR}/execution-spec-tests" checkout \
  af77c56ab299fdecfd6d80d11492aaa4472fc03a
```

Build the host-side recovery tools from the Alpen branch:

```bash
cd "${RECOVERY_WORKDIR}/alpen"
cargo build --locked --release \
  -F sequencer -F debug-utils -F test-mode -F debug-asm -F prover \
  --bin strata-test-cli \
  --bin strata-datatool
```

Run the recovery harness:

```bash
ALPEN_CORE_REPO="${RECOVERY_WORKDIR}/alpen" \
STRATA_TEST_CLI="${RECOVERY_WORKDIR}/alpen/target/release/strata-test-cli" \
STRATA_DATATOOL="${RECOVERY_WORKDIR}/alpen/target/release/strata-datatool" \
EEST_REPO="${RECOVERY_WORKDIR}/execution-spec-tests" \
"${RECOVERY_WORKDIR}/e2e-runner/harness/ee-recovery/run.sh"
```

The runner builds the OL and EE images from `ALPEN_CORE_REPO`, starts the local Signet stack, runs the source and recovered sequencers with native proof execution, and writes logs and datadirs below `e2e-runner/runs/ee-recovery-<timestamp>/`. Set `EE_RECOVERY_SKIP_IMAGE_BUILD=1` on later runs to reuse matching local images.

## Reviewer notes: blockers and current mitigations

| Commit | Blocker | Current mitigation |
|---|---|---|
| `47ebaae1` | The existing tools did not provide one operation from a stopped EE datadir to a complete fresh datadir. | Add `ee-state-recover` to extract, hydrate, reconstruct, validate, import, bootstrap, and publish the recovered datadir. |
| `fcea7c40` | Finalized execution-chain initialization assumed the first local block was genesis. | Initialize a trusted finalized anchor at its real execution height while preserving the block-zero check for genesis startup. |
| `aaf4b8f0` | Sequencer startup inserted genesis account, execution, and batch history into a sparse recovered database. | Inspect persisted local history first, preserve normal genesis initialization, and validate a complete recovery-anchored database without inserting genesis. |
| `06923e2f` | Reth synchronization scanned finalized EE history from block zero. | Start synchronization from the first locally retained finalized block and require the paired Reth database to contain the recovery anchor. |
| `a68dc331` | Batch lifecycle and chunk backfill assumed batch-zero history. | Treat the earliest contiguous retained batch as the local anchor and resume processing at the next batch. |
| `9f8c1b3a` | The first post-recovery proof could not bind an OL-authoritative block ID to a synthetic recovery header. | Bind the authoritative parent block ID and parent state root as separate chunk-transition values, and preserve authoritative ancestor hashes in witness input. |
| `16589c0d` | `eth_gasPrice` walked beyond the sparse history boundary and failed on the zero parent hash. | Fall back to the configured suggested fee only for `HeaderNotFound(B256::ZERO)` and propagate all other errors. |

#### Blocker

The individual reconstruction and import operations did not provide a single path from the source sequencer datadir to a complete replacement datadir. The caller had to coordinate transaction extraction, Bitcoin hydration, state reconstruction, Reth import, EE bootstrap, and validation.

#### Current mitigation

`ee-state-recover` owns the operation. It uses a temporary output directory, validates the paired Reth and EE anchors, and publishes the result only after every phase succeeds. It never mutates the source datadir or overwrites an existing output path.

#### Production boundary

The command proves that L1 state reconstruction and sparse restart are feasible. A production recovery tool still needs operator UX, recovery manifests, resumability, observability, deployment guidance, and a reviewed trust model for selecting the target update sequence.

#### Blocker

The recovered EE database begins at a trusted execution block above genesis. Finalized-chain initialization treated the first local block as block zero, so it could not represent the recovered anchor in the finalized-height index.

#### Current mitigation

Finalized-chain initialization records the supplied block at its real height. Normal genesis startup checks that its anchor is block zero. Repeating initialization with the same anchor remains idempotent, while an attempt to replace an initialized anchor fails.

#### Production boundary

The storage API now supports the required invariant. A production design should define anchor provenance and persistence as part of the node's recovery or promotion model instead of relying on the materializer and startup code to infer intent.

#### Blocker

Startup ran the account-state, execution-chain, and batch genesis helpers on every database. A sparse recovered database omits block-zero and batch-zero history, so startup attempted to mix generated genesis records with the recovered anchor.

#### Current mitigation

Startup inspects the three persisted histories before mutation:

- empty and genesis-anchored databases keep the existing genesis path;
- a complete post-genesis account, execution, and batch anchor skips genesis insertion; and
- mixed genesis and recovery component histories fail as inconsistent.

The branch does not add a new node or database type.

#### Production boundary

The inline classification keeps this feasibility change small and preserves existing startup behavior. A production design should use an explicit persisted recovery or promotion manifest so restart intent cannot depend on inference from sparse records.

#### Blocker

The EE engine searched Sled over `0..=best_finalized_height`. A recovered database has no entries below its anchor, so synchronization failed before it could confirm that Reth already contained the imported anchor.

#### Current mitigation

Execution storage exposes the first locally retained finalized block. Normal databases still replay from block zero. Sparse databases require the first retained anchor to exist in Reth, then replay any later finalized blocks through the existing payload and fork-choice path.

#### Blocker

The recovered EE database contains one trusted batch frontier at its original non-zero index and omits earlier batches and chunks. Batch lifecycle initialization required batch zero, and an empty chunk database began backfill at batch one. The recovered sequencer failed with `batch 1 not found during backfill`.

#### Current mitigation

Batch consumers resolve the earliest contiguous retained batch as their local anchor. Lifecycle and reorg frontiers start there. An empty chunk database begins from the anchor's ending execution block and requests `anchor_index + 1`. Ordinary databases still use batch zero and begin processing at batch one.

#### Production boundary

The mitigation reuses the existing `Genesis` batch status to mark the first trusted local batch. A production design should give the retained recovery frontier an explicit meaning and define its reorg guarantees.

#### Blocker

OL commits the recovery execution block ID and state root, while EE DA does not contain the complete canonical header preimage for that historical block. The materializer can build a header with the required execution metadata and import it into Reth under the OL-authoritative block ID, but hashing that synthetic header does not reproduce the committed ID. The chunk prover recomputed the parent ID from the header and rejected the first recovered chunk. Recomputing ancestor hashes from header RLP had the same ambiguity at the synthetic anchor.

#### Current mitigation

`ChunkTransition` commits the parent state root in addition to the authoritative parent execution block ID. Proof-input assembly reads the parent ID from the chunk boundary and the state root from the imported header. The chunk runtime binds both values, and the account runtime checks them against its current execution tip and state root. Reth witness capture stores each ancestor header's authoritative database hash with its RLP so `BLOCKHASH` linkage does not depend on hashing a synthetic header.

This changes the chunk proof public-value interface, not the Bitcoin DA format. The published `DaBlob` remains unchanged.

We will include the required mitigation for this blocker in the upcoming work to fix known proof soundness issues.

#### Blocker

Reth's gas oracle walks backward through recent transaction-bearing blocks. After recovered startup, the retained history ends at the imported anchor. If the local suffix has too few samples, the oracle follows the anchor's zero parent hash and returns `HeaderNotFound(0x00...00)`.

#### Current mitigation

Alpen runs Reth's gas-oracle calculation first. It uses the configured default suggested fee, or 1 Gwei when no default exists, only for `HeaderNotFound(B256::ZERO)`. Missing non-zero headers and other errors still propagate.

#### Production boundary

The fallback is narrow and preserves full-history behavior. A full-node promotion path should retain the fee-history suffix and avoid this boundary. If sparse recovery remains supported, the RPC contract should document the configured fallback.

## `BLOCKHASH` after sparse recovery

The recovered database does not contain the complete pre-anchor execution-header history. The e2e runner produces and proves 256 canonical blocks before it runs the selected EEST gate. Reth executes and stores these blocks through the normal sequencer path, and OL accepts the account updates that cover them. The runner does not inject block hashes or fabricate warm-up blocks.

After warm-up, every height visible to the EVM `BLOCKHASH` opcode belongs to the imported anchor or the new canonical suffix. Older heights return zero under the EVM's 256-block rule. A recovery-specific EEST checks both sides of that boundary: an expired pre-recovery height returns zero, and the preceding post-recovery block returns its non-zero canonical hash.

This warm-up is a feasibility strategy for sparse history. It does not change `BLOCKHASH` semantics or create a new Alpen node mode.

### Alternative considered: EIP-2935

We considered the EIP-2935 history-storage contract. Alpen's current chain specification activates Prague but does not predeploy the history contract, so existing chains have no contract code or historical ring-buffer state to reconstruct.

EIP-2935 does not prescribe an Alpen DA format. Under Alpen's state-diff DA model, each history-contract update would become part of the execution state diff published for reconstruction. The network would carry recent block IDs through DA, increasing the recurring payload and requiring a coordinated genesis or upgrade design. The contract also exposes history through a contract call; it does not redefine the `BLOCKHASH` opcode.

The warm-up approach adds no recurring DA cost or predeploy. It uses canonical blocks produced after recovery to establish the complete observable `BLOCKHASH` window. A future network can consider EIP-2935 as a separate protocol feature if contracts need direct access to historical hashes.

## Scope and non-goals

This PR demonstrates feasibility. It does not provide:

- raw Bitcoin scanning and ordering of EE DA transactions;
- a production operator recovery workflow;
- promotion of an existing full EE node to sequencer.

## Next steps

1. Design and test promotion of a synced EE full node to sequencer.
2. Move the sparse-anchor proof requirements into the proof-soundness epic.
3. Keep this PR as the blocker map and reference implementation.
4. When the time arrived revisit each atomic commit and replace its feasibility mitigation with the agreed production architecture.


storopoli:
Shouldn't we move this to `alpen-ee` repo @irnb?

bin/alpen-client/src/prover/spec_chunk.rs
```diff
@@ -233,7 +234,39 @@ impl ProofSpec for ChunkSpec {
...
+                    PaasError::PermanentFailure(format!(
+                        "decode ancestor header {ancestor_idx} for {block_hash:?}: {e}"
+                    ))
+                })?;
+                let authoritative_hash = record
+                    .ancestor_header_hashes
+                    .get(ancestor_idx)
+                    .copied()
+                    .unwrap_or_else(|| header.hash_slow());
+                let sealed = Sealed::new_unchecked(header, authoritative_hash);
```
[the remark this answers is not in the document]
delbonis: This has come up before and I don't think we have a good solution for this issue.  My instinct is that since it's difficult to (cheaply) constrain this to follow the upstream we should stub off the `BLOCKHASH` op and make it return 0 (or maybe the block number padded with 0 or something).

I am not sure how commonly used this op is (if I remember right, it's nonzero but fairly uncommon since it's only really useful for getting some randomness, people basically never make proofs against history like this and there's a more modern way you're supposed to do that anyways) so this might be a totally acceptable tradeoff.

crates/ee-chunk-runtime/src/chunk_processing.rs
```diff
@@ -171,7 +177,7 @@ pub fn verify_chunk_transition<E: ExecutionEnvironment>(
 
     // 2. Make sure the state matches the parent block's state root.
     let computed_pre_sr = state.compute_state_root()?;
-    if computed_pre_sr != prev_header.get_state_root() {
+    if computed_pre_sr != tsn.parent_state_root() {
         return Err(EnvError::MismatchedCurStateData);
     }
```
delbonis, line 176: Does this make sense to be done *here* or should we split up the checks?  Ie., have a step that makes the sure the pre-state matches the chunk transitions commitment and does the other consistency checks.

Is it redundant to pass `parent_exec_blkid` when `prev_header` is passed?  Can we constrain this somewhere else upstream?  What is it actually needed for that we can't just check from the passed header?

crates/ee-chain-types/ssz/chunk.ssz
delbonis, line 1: Alternative idea, can we get away with *only* specifying the parent state in this transition type?  Since that's what we really care about.  I know with out current system we want to enforce parents are consistent, but maybe we could loosen this *here* so we can specify it elsewhere.

delbonis (changes requested):
I might be premature with some of these comments.  Maybe we should create a ticket about these.

storopoli (changes requested):
Shouldn't we move parts of this PR to alpen-ee repo @irnb?

PR #2150, opened 2026-07-23
fix(btcio): stop re-posting settled checkpoints after sequencer restart

storopoli:
## Description

Restarting the sequencer while signed-but-unpublished checkpoints sat in the
writer queue made the node post every queued epoch to Bitcoin twice. Startup
reconciliation deleted the local checkpoint artifacts but never touched the
writer queue, so the queued originals published and were accepted while the
node rebuilt each checkpoint with fresh proof bytes. The rebuilds hash
differently (the dedup key covers proof bytes), so they were enqueued as new
work, published behind the originals, and rejected by ASM as stale. Each
duplicate cost real fees and delayed current checkpoints. This happened twice
on staging on 2026-07-17 (epochs 14–28, then 29–33).

Two fixes, plus the queue state to support them:

1. The writer now checks each checkpoint payload before signing. If its epoch
   is at or below the declared-final epoch it is abandoned; if it is only at or
   below the last-seen ASM tip it is deferred and re-checked next tick, since
   the last-seen tip can roll back on a reorg and abandonment cannot. Unknown
   tips and undecodable payloads publish as before, so a decoder bug cannot
   stall the writer.
2. Startup reconciliation now cancels the queued half of everything it deletes.
   It finds queue entries by decoding their payloads rather than trusting the
   signing-entry link (a crash can leave intents with no signing marker), and
   it also sweeps the broadcaster DB directly, invalidating stale checkpoint
   reveals and their commit transactions even when no bundle recorded the
   txids. Anything that already reached the network is left alone: published
   bundles, and also still-`Unpublished` bundles whose commit or reveal
   transaction escaped before the crash. The watcher then waits on those until
   the original finalizes, and the writer check from fix 1 handles the rebuild.
   A broadcast transaction cannot be recalled, so this is the only safe order.

The cancellation itself lives in `strata-btcio`, the crate that owns the
writer queue and the broadcaster, behind a `CheckpointCancelContext` trait in
the same shape as `WatcherServiceContext`. `bin/strata` keeps the node-level
part: read the ASM verified tip, pick the epoch side from prover config,
delete the local payloads, proofs and prover tasks, log the combined stats.
It has to run before those two services start, since the broadcaster resumes
publishing straight out of its database, which rules out driving this through
their handles.

Cancellation marks entries with a new terminal `Abandoned` status instead of
deleting them. The watcher drains by index and the bundler requires dense
predecessors, so deletion would stall both; there is even an existing test
asserting a missing intent is a fatal DB inconsistency. The bundler's startup
scan is anchored on finalization to match: it walks back to the first intent
bundled into a `Finalized` payload rather than stopping at the first bundled
one, since abandonment and intent-id aliasing can both leave an unbundled
entry below a bundled one.

## Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Review commit by commit. The first two commits are groundwork: the decode
helper has no callers yet, and the `Abandoned` commit adds the terminal state
plus the startup-scan bound it needs. Nothing there changes what gets posted
to L1, though the new scan bound does recover unbundled intents that the old
break-on-first-bundled scan could skip. The third and fourth commits are the
actual fixes.

`extract_payload_checkpoint_epoch` in commit 1 is not a duplicate of the
existing checkpoint parsing. `extract_checkpoint_from_envelope` and the
csm-worker extractor unwrap the taproot witness of a real Bitcoin
transaction; this
helper decodes the writer queue's pre-envelope `L1Payload` bytes, where no
transaction exists yet. Both import the tag constants and the
`CodecSsz<CheckpointPayload>` decode from the same upstream crate. A possible
follow-up is exporting a shared `decode_checkpoint_payload` from
`alpenlabs/asm` so both sides use one function.

Three things were considered and deliberately not done: no check at the
literal `send_raw_transaction` (with one sequencer draining one payload at a
time, the tip cannot advance between the writer check and broadcast); no
per-index intent schema rework for the id-aliasing quirk on resubmission,
which is documented at the type and pinned by a test as harmless beyond a
possible reorder; and no bitcoind probe in startup reconciliation for the
crash window between `send_raw_transaction` and the broadcaster's status
write, where a sent transaction still reads `Unpublished`. That window is
milliseconds wide, the epoch gate stops the rebuild if the original is
accepted, and the worst case is one rejected transaction's fees. It is
documented on `is_escaped_broadcast_status`.

AI assistance disclosure: this change was developed with Claude Code and
Codex. The plan, the implementation, and each review round were checked by a
human; all tests and lints were run locally before pushing.

- [ ]  Yes
- [x]  No
    
## Checklist

## Related Issues

STR-4015


bin/strata/src/checkpoint_reconcile.rs
```diff
@@ -96,6 +140,319 @@ fn checkpoint_commitments_from_epoch(
...
+                    .put_payload_entry_blocking(payload_idx, entry)
+                    .with_context(|| format!("abandon L1 writer payload entry {payload_idx}"))?;
+                stats.abandoned_bundles += 1;
+            }
+            WriterCancelAction::CancelAndInvalidate => {
+                stats.invalidated_txs += invalidate_unpublished_bundle_txs(storage, &entry)?;
+                if entry.status != L1BundleStatus::Abandoned {
+                    entry.payload_signature = None;
+                    entry.status = L1BundleStatus::Abandoned;
+                    writer
```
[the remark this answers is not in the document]
storopoli: Good catch, this was real. Fixed by checking the broadcaster entries before touching anything: if the commit or reveal already made it to Published/Confirmed/Finalized, reconcile now leaves the bundle, its intent, and both tx entries alone. The watcher then sits on the still-`Unpublished` bundle until the original finalizes, same as it already does for `Published` bundles, rather than skipping past an abandoned entry to the rebuild. The broadcaster sweep got the same guard, so a reveal whose commit escaped stays publishable and the envelope can still complete. Tests: `reconcile_deletes_artifacts_but_leaves_escaped_unpublished_bundle` plus a partial-escape case. Amended into the reconcile commit rather than a fixup.

bin/strata/src/checkpoint_reconcile.rs
```diff
@@ -96,6 +141,375 @@ fn checkpoint_commitments_from_epoch(
...
+                        payload_idx,
+                        epoch,
+                        ?commit_status,
+                        ?reveal_status,
+                        "leaving checkpoint bundle with escaped broadcaster transaction"
+                    );
+                    continue;
+                }
+
+                stats.invalidated_txs += invalidate_unpublished_bundle_txs(storage, &entry)?;
```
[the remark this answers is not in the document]
storopoli: You're right that this window exists: a crash between `send_raw_transaction` and the status write leaves a sent transaction reading `Unpublished`, and the DB alone can't tell the difference. We're accepting it rather than probing bitcoind from startup reconciliation. The window is milliseconds wide, and the failure it allows is bounded: if the original still gets mined, ASM accepts it and the writer's epoch gate abandons the rebuilt duplicate before signing, or ASM rejects the stale one and we pay one transaction's fees. No safety impact either way, and probing would add a bitcoind RPC dependency to a startup path that currently only reads the DB. Documented on `is_escaped_broadcast_status` and added to the deliberately-not-done list in the PR description.

bin/strata/src/checkpoint_reconcile.rs
```diff
@@ -96,6 +141,385 @@ fn checkpoint_commitments_from_epoch(
...
+        let Some(mut tx_entry) = broadcast_db
+            .get_tx_entry(tx_idx)
+            .with_context(|| format!("read L1 broadcaster transaction entry {tx_idx}"))?
+        else {
+            warn!(tx_idx, "L1 broadcaster transaction index has no entry");
+            continue;
+        };
+        if tx_entry.status != L1TxStatus::Unpublished {
+            continue;
+        }
```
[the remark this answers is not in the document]
storopoli: Confirmed, and this one is different from the accepted probe-only window: here the escape is recorded in the DB, so we fixed it rather than documenting it. The sweep now also decodes escaped (Published/Confirmed/Finalized) checkpoint entries into an epoch map, and a pre-publish bundle whose epoch has escaped orphan txs gets relinked instead of abandoned: the orphan commit/reveal txids are written into the bundle and it resets to `Unpublished`. The watcher then parks on it and the broadcast-status machinery drives it to finalization, so no second envelope is signed and no rebuild races the original. Covered by `cancellation_relinks_published_orphan_checkpoint_txs` plus a mismatched-epoch test that keeps the normal abandon path honest. Amended into the reconcile commit.

bin/strata/src/checkpoint_reconcile.rs
```diff
@@ -96,6 +145,431 @@ fn checkpoint_commitments_from_epoch(
...
+            .as_ref()
+            .is_some_and(|entry| is_escaped_broadcast_status(&entry.status))
+        {
+            debug!(
+                tx_idx,
+                epoch,
+                %commit_txid,
+                "leaving queued checkpoint reveal whose commit transaction escaped"
+            );
+            continue;
```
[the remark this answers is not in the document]
storopoli: Right again, the commit-only escape slipped past the relink map. Fixed: the commit-escaped branch of the sweep now records the (commit, reveal) pair into the escaped map before skipping, so the zero-txid bundle relinks to the in-flight envelope instead of being abandoned, and the still-`Unpublished` reveal remains publishable for the broadcaster to complete. Same first-wins duplicate handling as the reveal-escaped path, factored into a shared helper. Test: `cancellation_relinks_bundle_when_only_commit_escaped` asserts the relink, the untouched reveal, and the untouched published commit. Amended into the reconcile commit.

bin/strata/src/checkpoint_reconcile.rs
```diff
@@ -96,6 +145,455 @@ fn checkpoint_commitments_from_epoch(
...
+                    %err,
+                    "could not extract checkpoint from tagged L1 broadcaster transaction"
+                );
+                continue;
+            }
+        };
+        let epoch = checkpoint.payload.new_tip().epoch;
+        if epoch < first_unaccepted_epoch {
+            continue;
+        }
```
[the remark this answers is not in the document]
storopoli: Confirmed, and this one matters beyond the crash windows: it is the upgrade path. Nodes still carrying incident-era duplicate backlogs would have re-fired them from the broadcaster DB on the first restart with this code, since both reconcile loops filtered to epochs above the tip and the pre-sign gate never sees already-signed entries. Fixed by removing the below-tip filters: queued checkpoint entries at or below the verified tip now get the same treatment (invalidate + abandon, with the escaped-tx guards intact so nothing already on the network is touched and no commit output strands). Relinking stays restricted to unaccepted epochs since settled epochs never rebuild, and artifact deletion is unchanged. Three new tests cover below-tip invalidation/abandonment with artifact preservation, no-relink-below-tip, and the commit-escaped reveal staying publishable. Amended into the reconcile commit.

bin/strata/src/checkpoint_reconcile.rs
```diff
@@ -23,8 +45,41 @@ pub(crate) fn reconcile_unaccepted_checkpoint_artifacts(nodectx: &NodeContext) -
     };
 
     let storage = nodectx.storage();
+    let magic_bytes = *nodectx.asm_params().magic.as_bytes();
+    let stats = reconcile_unaccepted_checkpoint_artifacts_from_epoch(
+        storage,
+        first_unaccepted_epoch,
+        magic_bytes,
+    )?;
```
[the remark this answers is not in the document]
storopoli: Valid finding, fixed with one correction to the suggested shape. Splitting the whole queue cancellation out of the prover guard would stall proverless sequencers: cancelling an above-tip entry is only safe because artifact deletion triggers a rebuild that replaces it, and without a prover there is no rebuild, so abandoning the original would kill that epoch's only in-flight submission. The split is therefore by epoch side: every node now cleans the settled side (at-or-below-tip duplicates, the ones that republish from the broadcaster DB), while the unaccepted side plus artifact deletion stays prover-gated. Covered by `settled_only_reconcile_leaves_unaccepted_queue_and_artifacts_untouched`. Amended into the reconcile commit.

bin/strata/src/checkpoint_reconcile.rs
```diff
@@ -96,6 +154,541 @@ fn checkpoint_commitments_from_epoch(
...
+        L1BundleStatus::Unpublished | L1BundleStatus::Abandoned => {
+            WriterCancelAction::CancelAndInvalidate
+        }
+        L1BundleStatus::Published | L1BundleStatus::Confirmed | L1BundleStatus::Finalized => {
+            WriterCancelAction::Leave
+        }
+    }
+}
+
+type EscapedCheckpointTxs = BTreeMap<Epoch, (Buf32, Buf32)>;
```
delbonis: Make a container for this!

bin/strata/src/checkpoint_reconcile.rs
delbonis, line 1: All of these fn using `NodeStorage` should be generic over a context trait.

crates/btcio/src/writer/checkpoint_payload.rs
```diff
@@ -0,0 +1,74 @@
...
+use strata_codec_utils::CodecSsz;
+use strata_csm_types::L1Payload;
+use strata_identifiers::Epoch;
+
+/// Returns the checkpoint epoch encoded in a checkpoint-tagged [`L1Payload`].
+///
+/// Returns [`None`] for non-checkpoint payloads, multi-chunk payloads, or
+/// malformed checkpoint data.
+pub fn checkpoint_payload_epoch(payload: &L1Payload) -> Option<Epoch> {
+    let tag = payload.tag();
```
delbonis: This name is pretty misleading.  Like `extract_payload_checkpoint_epoch`?  I'm still not sure I understand why it's this specific.

crates/btcio/src/writer/checkpoint_payload.rs
```diff
@@ -0,0 +1,74 @@
...
+    if tag.subproto_id() != CHECKPOINT_SUBPROTOCOL_ID || tag.tx_type() != OL_STF_CHECKPOINT_TX_TYPE
+    {
+        return None;
+    }
+
+    let [encoded] = payload.data() else {
+        return None;
+    };
+    let decoded: CodecSsz<CheckpointPayload> = decode_buf_exact(encoded).ok()?;
```
delbonis: Wait couldn't we just decode the SSZ directly?  Is this correct in the context of how the checkpoint txs are structured?  Like I know the tests pass but the tests for this fn are written for this fn.

delbonis (dismissed):
Algorithmically looks good but some organizational comments.

bin/strata/src/checkpoint_reconcile.rs
storopoli: You mean a SSZ container?

bin/strata/src/checkpoint_reconcile.rs
```diff
@@ -96,6 +154,541 @@ fn checkpoint_commitments_from_epoch(
...
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
```
bewakes: Please add a docstring.

bin/strata/src/checkpoint_reconcile.rs
```diff
@@ -96,6 +154,541 @@ fn checkpoint_commitments_from_epoch(
...
+        }
+
+        match plan_writer_cancellation(&entry.status) {
+            WriterCancelAction::Cancel => {
+                if let Some(&(commit_txid, reveal_txid)) = escaped_checkpoint_txs.get(&epoch) {
+                    entry.commit_txid = L1TxId::from(commit_txid.0);
+                    entry.reveal_txid = L1TxId::from(reveal_txid.0);
+                    entry.payload_signature = None;
+                    entry.status = L1BundleStatus::Unpublished;
+                    writer
```
bewakes: I find this manipulation of writer db from checkpoint reconciler a bit concerning and breaking the abstraction. Please consider using handles that we already have and communicate via message passing. That would make this much cleaner, testable and concerns separated.

bewakes: Some methods might have to be added to the handle.

bin/strata/src/checkpoint_reconcile.rs
```diff
@@ -96,6 +154,541 @@ fn checkpoint_commitments_from_epoch(
...
+            .with_context(|| format!("read broadcaster transaction {txid}"))?
+        else {
+            continue;
+        };
+        if tx_entry.status != L1TxStatus::Unpublished {
+            continue;
+        }
+
+        tx_entry.status = L1TxStatus::InvalidInputs;
+        broadcast_db
```
bewakes: And same with broadcast db. Consider acting via handle.

bin/strata/src/checkpoint_reconcile.rs
```diff
@@ -143,3 +736,1245 @@ fn first_unaccepted_checkpoint_epoch(nodectx: &NodeContext) -> Result<Option<Epo
 
     Ok(Some(first_unaccepted_epoch))
 }
+
+#[cfg(test)]
+mod tests {
```
bewakes, line 334: The tests are humongous. Please consider if we can setup/builders that can be reused across cases. This it not a blocker though.

crates/btcio/src/writer/bundler/logic.rs
```diff
@@ -79,35 +79,29 @@ async fn bundle_unbundled_intent(ops: &EnvelopeDataOps, intent_idx: u64) -> anyh
...
     ops: &EnvelopeDataOps,
 ) -> anyhow::Result<Vec<PendingIntent>> {
-    let mut curr_idx = ops.get_next_intent_idx_blocking()?;
+    let next_intent_idx = ops.get_next_intent_idx_blocking()?;
     let mut unbundled = Vec::new();
 
-    while curr_idx > 0 {
-        curr_idx -= 1;
-        if let Some(intent) = ops.get_intent_by_idx_blocking(curr_idx)? {
+    for intent_idx in 0..next_intent_idx {
```
bewakes: This introduces linear scan from the 0 index right? This can be costly as the bundled items grow in database.
The earlier version also had the linear scan, but it was bound to exit early.

bewakes: The assumption earlier was that items are processed in order and once processed, there's no need to re-process them(or get back to them). Has this assumption now changed?

crates/btcio/src/writer/watcher/service.rs
```diff
@@ -284,6 +334,57 @@ impl<C: WatcherServiceContext> AsyncService for WatcherService<C> {
 }
 
 impl<C: WatcherServiceContext> WatcherState<C> {
+    fn stale_checkpoint(&self, payloadentry: &BundledPayloadEntry) -> StaleCheckpointAction {
```
bewakes: The name is a bit confusing. Perhaps something like : `get_stale_ckpt_action` or `get_stale_checkpoint_action`?

functional-tests/tests/checkpoint/test_checkpoint_no_double_submission_on_restart.py
```diff
@@ -0,0 +1,212 @@
...
+            first_epoch=2,
+            last_epoch=self.BACKLOG_LAST_EPOCH,
+        )
+
+        logger.info(
+            "restarting sequencer with checkpoint epochs 2..%s queued",
+            self.BACKLOG_LAST_EPOCH,
+        )
+        strata.stop()
+        time.sleep(self.RESTART_PAUSE_SECONDS)
```
bewakes: See if we can use `wait_until` here(maybe wait until till `protocolVersion` is returning value). With the right condition, we might be able to save some seconds/milliseconds.

functional-tests/tests/checkpoint/test_checkpoint_no_double_submission_on_restart.py
```diff
@@ -0,0 +1,212 @@
...
+                if not TestCheckpointNoDoubleSubmissionOnRestart._is_checkpoint_tx(tx):
+                    continue
+                txid = tx["txid"]
+                payload = extract_posted_checkpoint_payload(btc_rpc, txid)
+                epoch = parse_checkpoint_payload(payload).epoch
+                postings[epoch].append(txid)
+        return postings
+
+    @staticmethod
+    def _is_checkpoint_tx(tx):
```
bewakes: minor: This is a pretty common thing that we want to have. Consider moving it to more common place. Also, we might already have something like this.

bewakes (changes requested):
Nice work, and I like the smart functional test!
Some concerns over the interaction between reconciler and writer/broadcaster. Other are non-blocking comments.

bin/strata/src/checkpoint_reconcile.rs
storopoli: Not an SSZ container, just a named struct, which I think is what you meant. The `(Buf32, Buf32)` is now `EscapedEnvelopeTxs { commit_txid, reveal_txid }`, so the relink path reads `escaped.commit_txid` instead of a tuple whose field order you have to keep in your head.

It moved to `crates/btcio/src/writer/reconcile.rs` along with the rest of the cancellation code, see the thread on the context trait.

bin/strata/src/checkpoint_reconcile.rs
storopoli: Agreed, it shouldn't live in the binary. Fixed by moving the code rather than by adding a trait there.

The writer-queue and broadcaster cancellation is now `crates/btcio/src/writer/reconcile.rs`, behind a `CheckpointCancelContext` trait. Same arrangement as `WatcherServiceContext` and the broadcaster's context: trait next to the logic, in the crate that owns those databases, concrete impl alongside it. `bin/strata` calls `cancel_settled_checkpoint_submissions` / `cancel_unaccepted_checkpoint_submissions` and keeps only the node-level part, which is reading the ASM verified tip, picking the epoch side from prover config, and deleting local payloads, proofs and prover tasks. That leaves 214 lines on `NodeStorage` in the binary, down from about 740.

Two things fell out of the move. `bin/strata` no longer needs the `strata-l1-txfmt` dependency this PR had added to it, and the `#[cfg(feature = "sequencer")]` gating is gone from the logic, since btcio has no feature gates. The binary still gates the call.

This also covers @bewakes's comment below about the reconciler writing to the writer DB.

crates/btcio/src/writer/checkpoint_payload.rs
storopoli: Renamed to `extract_payload_checkpoint_epoch`.

On why it is this specific: the writer queue carries payloads for every subprotocol, so the SPS-50 tag check is the point of the function rather than a precondition on the side. Single-chunk is the same fact from the other end, since chunking is the EE DA path and checkpoints are always written as one chunk. Both are in the doc comment now, along with a pointer to where the bytes come from.

crates/btcio/src/writer/checkpoint_payload.rs
storopoli: Not directly, no. The bytes aren't bare SSZ: `CodecSsz` frames the SSZ body behind a `strata-codec` varint length prefix (`crates/codec-utils/src/ssz_shim.rs:34-48`), and the producer is the sequencer's `complete_checkpoint_signature` RPC calling `encode_to_vec(&CodecSsz::new(checkpoint_payload))` into a single-chunk `L1Payload` (`bin/strata/src/sequencer/rpc.rs:187-191`). `CheckpointPayload::from_ssz_bytes` on those bytes trips on the prefix.

Your point about the tests stands, though. They only prove the function agrees with itself. The independent evidence is elsewhere: the functional test's Python decoder reads the same varint before handing the rest to SSZ (`functional-tests/tests/checkpoint/helpers.py`), and the csm-worker and ASM extractors go through `CodecSsz` too. I put the producer reference in the doc comment so the next reader doesn't have to go find it.

bin/strata/src/checkpoint_reconcile.rs
storopoli: Added. The type moved to `crates/btcio/src/writer/reconcile.rs` in the process, and I documented the other new items there as well since they were all bare. `CheckpointEpochSide` now says why the two sides of the verified tip are handled separately.

bin/strata/src/checkpoint_reconcile.rs
storopoli: Half of this is fixed and half of it can't be, so let me take them separately.

Fixed: the binary no longer writes to the writer or broadcaster databases. That code moved into `crates/btcio/src/writer/reconcile.rs` behind a `CheckpointCancelContext` trait, so the crate that owns those queues does its own cleanup and `bin/strata` calls two functions. That also settles @delbonis's comment about these functions being generic over a context trait.

Can't: message passing needs the services to exist, and at this point they don't. Reconciliation runs at `services.rs:274`; the broadcaster and writer start at `services.rs:323`. That ordering is the fix rather than an accident, because the broadcaster resumes publishing straight out of its database on startup, so the queue has to be clean before it starts. Routing the cleanup through the handles would put it after the republish it exists to prevent.

bin/strata/src/checkpoint_reconcile.rs
storopoli: Same move: this is in `strata-btcio` now, behind the context trait, and the concrete implementation goes through `BroadcastDbOps` rather than reaching for `storage.db().broadcast_db()`. The handle itself still isn't reachable this early in startup, for the reason in the thread above.

bin/strata/src/checkpoint_reconcile.rs
storopoli: Fair. They were moving to btcio anyway, so they landed on a fixture instead of moving verbatim: `queue_checkpoint`, `put_envelope_txs` and `put_opaque_txs` for setup, `bundle`, `intent_status` and `tx_status` for assertions. Most cases went from 80-odd lines to 15-20. The three that stayed in `bin/strata` cover artifact deletion and got the same treatment.

crates/btcio/src/writer/bundler/logic.rs
storopoli: The assumption did change, and dropping the early exit was the wrong way to deal with it.

Two things break "processed in order, never revisited". Abandonment retires entries in the middle of the queue. And intent-id aliasing means resubmitting an abandoned intent refreshes the shared id-keyed entry, so an older index ends up tracking the retry. Either one leaves an `Unbundled` intent sitting below a `Bundled` one, which the old break-on-first-bundled scan walks straight past.

Rewritten to anchor on finalization instead: the scan walks backwards and stops at the first intent bundled into a `Finalized` payload. Payloads finalize in index order, so nothing below that point can still be waiting for a bundle. The cost is bounded by the unfinalized tail rather than by every intent ever written. Covered by `startup_scan_stops_below_finalized_payload`.

crates/btcio/src/writer/watcher/service.rs
storopoli: Renamed to `stale_checkpoint_action`. I left off the `get_` prefix since clippy and the Rust API guidelines discourage it, but the confusing part you flagged, that the name reads like a predicate when it actually returns a decision, is fixed. It has a docstring now covering the abandon/defer/publish split.

functional-tests/tests/checkpoint/test_checkpoint_no_double_submission_on_restart.py
storopoli: Removed the sleep rather than replacing it. flexitest's `ProcService.stop()` already calls `terminate()` and then `p.wait(timeout=...)`, so it blocks until the process is actually gone, and the `wait_for_rpc_ready` after `start()` already polls `strata_protocolVersion` through `wait_until`. The two seconds were buying nothing.

Test passes locally without it, with one expected connection-refused retry while the RPC comes back up.

functional-tests/tests/checkpoint/test_checkpoint_no_double_submission_on_restart.py
storopoli: Moved to `tests/checkpoint/helpers.py` as `is_checkpoint_tx`, built on a small `checkpoint_tag` helper. `extract_posted_checkpoint_payload` in that same file was doing the identical SPS-50 tag check inline, so it goes through it now too. I didn't find an existing one to reuse: the closest is the OP_RETURN parsing in the EE DA `codec.py`, which is a different layout.

crates/btcio/src/writer/reconcile.rs
```diff
@@ -0,0 +1,1307 @@
...
+            Err(err) => {
+                warn!(
+                    tx_idx,
+                    %err,
+                    "could not extract checkpoint from tagged L1 broadcaster transaction"
+                );
+                continue;
+            }
+        };
+        let epoch = checkpoint.payload.new_tip().epoch;
```
bewakes: Btcio should not be aware of the internals of checkpoint.

bewakes (commented):
I think now this crosses another boundary: `btcio` which is supposed to be a dumb interface to bitcoin, now depends on `asm` subprotocol.
I don't have concrete suggestion here, but the direction should be moving this checkpoint specific logic upwards from btcio.

And, quoting codex on a different issue which seems plausible:
> One concern: the “gate before signing” is not an atomic correctness
  boundary. It reads status, then later signs/broadcasts; ASM/CSM state can
  advance between those operations. The code deliberately fails open when
  decoding/status is unavailable, so it is a best-effort duplicate reduction—
  not a hard guarantee. That may be acceptable, but it should not be presented
  as the sole protection. Startup reconciliation plus idempotent/upstream
  checkpoint submission is the stronger protection.

crates/btcio/src/writer/reconcile.rs
storopoli: I've made btcio not be aware of checkpoint. I've moved the logic to `crates/ol/checkpoint/src/reconcile.rs` and the identification of checkpoints to `crates/ol/checkpoint/src/l1_tx.rs`.

Take a look.

bewakes (approved):
I think this is fine for now. Some of the complexities also come from the current writer/broadcaster design, which most likely will be refactored in future.

crates/ol/checkpoint/src/reconcile.rs
```diff
@@ -0,0 +1,1448 @@
...
+pub fn cancel_settled_checkpoint_submissions(
+    storage: &NodeStorage,
+    first_unaccepted_epoch: Epoch,
+    magic_bytes: [u8; 4],
+) -> Result<WriterCancelStats> {
+    cancel_checkpoint_submissions_for_epoch_side(
+        &StorageCancelContext::new(storage),
+        first_unaccepted_epoch,
+        magic_bytes,
+        CheckpointEpochSide::Settled,
```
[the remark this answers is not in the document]
storopoli: Ok this is fixed and amended in e1df38fac69165525b186b23a437844837c4622b

bin/strata/src/checkpoint_reconcile.rs
```diff
@@ -6,25 +6,84 @@ use strata_asm_proto_checkpoint::CheckpointSubprotocol;
...
+        last_settled_epoch,
+        first_unaccepted_epoch,
+        magic_bytes,
+    )?;
     let mut cleanup_commitments =
-        checkpoint_commitments_from_epoch(nodectx, first_unaccepted_epoch)?;
+        checkpoint_commitments_from_epoch(storage, first_unaccepted_epoch)?;
 
     let deleted_payloads = storage
         .ol_checkpoint()
```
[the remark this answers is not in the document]
storopoli: Solved and amended in d95dbb374394b9d01d7d127a91de29229f336fad

bin/strata/src/checkpoint_reconcile.rs
```diff
@@ -6,25 +6,84 @@ use strata_asm_proto_checkpoint::CheckpointSubprotocol;
...
+/// so a rotated OL image cannot reuse stale pre-rotation proof artifacts.
 pub(crate) fn reconcile_unaccepted_checkpoint_artifacts(nodectx: &NodeContext) -> Result<()> {
-    if nodectx.config().prover.is_none() {
-        return Ok(());
-    }
-
     let Some(first_unaccepted_epoch) = first_unaccepted_checkpoint_epoch(nodectx)? else {
         return Ok(());
     };
+    let last_settled_epoch = last_declared_final_checkpoint_epoch(nodectx)?;
```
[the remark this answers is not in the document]
storopoli: 💀 Done and amended 4563e6cfa109ed8bbe6c967e4b95754b3d8bb403

bin/strata/src/services.rs
```diff
@@ -244,6 +254,42 @@ mod sequencer_services {
...
+        loop {
+            let reconciled_block = loop {
+                let asm_block = asm_handle.monitor().get_current().cur_block;
+                let csm_block = csm_monitor.get_current().cur_block;
+                if let Some(asm_block) = asm_block
+                    && Some(asm_block) == csm_block
+                {
+                    break asm_block;
+                }
+                sleep(CHECKPOINT_RECONCILE_POLL_INTERVAL).await;
```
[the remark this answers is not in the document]
storopoli: Amended in c8b7a8d4f6ed089d047faf162f5c51191f81e821

krsnapaudel (approved):
LGTM

crates/btcio/src/writer/watcher/service.rs
```diff
@@ -520,24 +643,62 @@ impl<C: WatcherServiceContext> WatcherState<C> {
...
+    observed_status: L1BundleStatus,
+) -> L1BundleStatus {
+    if *current_status != L1BundleStatus::Retiring {
+        return observed_status;
+    }
+
+    match observed_status {
+        L1BundleStatus::Finalized => L1BundleStatus::Finalized,
+        L1BundleStatus::NeedsResign => L1BundleStatus::Abandoned,
+        _ => L1BundleStatus::Retiring,
```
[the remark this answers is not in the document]
storopoli: ok fixed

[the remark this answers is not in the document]
storopoli: 6f2aaa45252d519d198a8586e3cb3d0fe3fa3083

bin/strata/src/checkpoint_reconcile.rs
delbonis: Yes that is what I meant.

bin/strata/src/services.rs
```diff
@@ -244,6 +258,60 @@ mod sequencer_services {
...
+                    }
+                    sleep(CHECKPOINT_RECONCILE_POLL_INTERVAL).await;
+                };
+
+                reconcile(nodectx)?;
+
+                let asm_block = asm_handle.monitor().get_current().cur_block;
+                let csm_block = csm_monitor.get_current().cur_block;
+                if asm_block == Some(reconciled_block) && csm_block == Some(reconciled_block) {
+                    return Ok(());
```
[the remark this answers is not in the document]
storopoli: Solved

crates/ol/checkpoint/src/reconcile.rs
```diff
@@ -0,0 +1,1501 @@
...
+) -> Result<WriterCancelStats> {
+    let mut stats = WriterCancelStats::default();
+    let (invalidated_txs, escaped_checkpoint_txs) =
+        invalidate_checkpoint_txs_by_decoding(ctx, boundary_epoch, magic_bytes, epoch_side)?;
+    stats.invalidated_txs += invalidated_txs;
+
+    let next_payload_idx = ctx
+        .next_payload_idx()
+        .context("read next L1 writer payload index")?;
+    let first_unfinalized_payload_idx = first_payload_after_last_finalized(ctx, next_payload_idx)?;
```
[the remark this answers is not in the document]
storopoli: Solved

storopoli:
Tagging @prajwolrg can we merge this?

delbonis (approved):

crates/btcio/src/writer/watcher/service.rs
```diff
@@ -493,25 +654,35 @@ impl<C: WatcherServiceContext> WatcherState<C> {
...
 
-                self.ctx.report_status(&payloadentry, &new_status).await;
+                self.ctx
+                    .report_status(&payloadentry, &observed_status)
+                    .await;
+
+                if payloadentry.status == L1BundleStatus::Retiring
+                    && new_status == L1BundleStatus::Abandoned
+                {
+                    self.abandon_retiring_intent(&payloadentry).await?;
```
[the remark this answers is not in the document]
storopoli: Done in df2fe791a

crates/btcio/src/writer/watcher/service.rs
```diff
@@ -520,22 +698,117 @@ impl<C: WatcherServiceContext> WatcherState<C> {
...
+    ) -> anyhow::Result<()> {
+        let checkpoint = self.ctx.inspect_payload(&payloadentry.payload);
+        let PayloadCheckpointRef::Checkpoint { epoch, .. } = checkpoint else {
+            return Ok(());
+        };
+        let Some(seen_epoch) = self.ctx.seen_checkpoint_epoch() else {
+            return Ok(());
+        };
+        if epoch <= seen_epoch {
+            return Ok(());
```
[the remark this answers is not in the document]
storopoli: Fixed

crates/btcio/src/writer/watcher/service.rs
```diff
@@ -493,25 +670,42 @@ impl<C: WatcherServiceContext> WatcherState<C> {
...
+                    .await;
+
+                if payloadentry.status == L1BundleStatus::Retiring {
+                    match new_status {
+                        L1BundleStatus::Abandoned => {
+                            self.abandon_retiring_intent(&payloadentry).await?
+                        }
+                        L1BundleStatus::Finalized => {
+                            self.release_unaccepted_finalized_intent(&payloadentry, &rtx.status)
+                                .await?
```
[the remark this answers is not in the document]
storopoli: This is a very unlikely corner case.

storopoli:
> **Low — startup reconciliation rescans and decodes the entire writer/broadcaster history on every boot (unbounded, no pruning)**
> 
> `crates/ol/checkpoint/src/reconcile.rs:476` and `crates/ol/checkpoint/src/reconcile.rs:554`
> 
> The payload loop is correctly bounded by `first_payload_after_last_finalized`, but the intent loop (`0..next_intent_idx`) and the broadcaster sweep (`0..next_tx_idx`) are not. Every historical intent is loaded and SSZ-decoded via `inspect_l1_payload`, and every broadcaster entry passes the status filter at line 568 (`is_escaped_broadcast_status` is true for `Finalized`), so every historical checkpoint reveal is loaded raw, consensus-deserialized, and taproot-envelope-parsed. Neither table is pruned by the running node, and checkpoint payloads carry the proof and state-diff sidecar, so per-entry cost is large. Each boot runs this ≥3 times (settled + unaccepted from `cancel_queued_checkpoint_submissions`, plus `reconcile_settled_checkpoint_queue`), and `reconcile_at_caught_up_checkpoint_tip` can repeat the whole pass while the tips move. Startup work therefore grows linearly with sequencer lifetime, ahead of the broadcaster and watcher starting — a sequencer that has produced many epochs takes progressively longer to come back up after a restart, exactly when it is furthest behind.
> 
> Fix: bound both scans the way the payload scan already is. Derive a floor intent index from the last finalized bundle (or from `first_unfinalized_payload_idx`) and start the intent loop there; in the broadcaster sweep, skip terminal entries before touching `tx_raw()` (only `Unpublished`, `Published`, and `Confirmed` can still change — `Finalized` entries cannot be cancelled or relinked in a way that matters), or bound the sweep to the broadcaster index range at/after the last finalized bundle's transactions.

Fixed and amended into the existing commits.

storopoli:
Ok I don't have bandwidth to continue this closing this.

PR #2152, opened 2026-07-24
fix(btcio): Recover canonical heights missing ASM state

bewakes:
## Description

This PR does the following:

- Recovers L1 canonical heights that were persisted before their ASM state was materialized.
- Rewinds unmaterialized canonical entries within the configured reorg lookback window.
- Replays those heights through ASM after restart.
- Fails clearly when no matching ASM anchor exists within the recovery window.
- Adds crash-injection coverage and unit tests for both recovery and failure paths.

**Note:** Created with the help of codex and claude with close supervision from author.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist


crates/btcio/src/reader/query.rs
```diff
@@ -93,6 +93,14 @@ pub async fn bitcoin_data_reader_task<E: BlockSubmitter>(
     status_channel: StatusChannel,
     event_submitter: Arc<E>,
 ) -> anyhow::Result<()> {
+    // Keep twice the safe depth so startup can recover a crash near the tip.
+    reconcile_unmaterialized_canonical_tip(
+        storage.as_ref(),
+        btcio_params.genesis_l1_height(),
+        btcio_params.l1_reorg_safe_depth() as L1Height * 2,
```
storopoli: `l1_reorg_safe_depth = 0` is a valid config (the `BtcioConfig` docs explicitly allow it), but here it makes the lookback 0, so `earliest_height == tip_height` and the walk bails on the first iteration without ever checking the parent. That turns the exact crash this PR is fixing into a restart loop: crash between the canonical write and ASM materialization, restart, bail, repeat.

The recovery window shouldn't depend on the finality depth. Something like `l1_reorg_safe_depth().max(1).saturating_mul(2)` would do it, and saturating also removes the theoretical u32 overflow for huge configured depths. A depth-0 regression test would pin this down.

storopoli (changes requested):
We need to fix a potential restart loop with a corner-case rare but possible config.

crates/btcio/src/reader/query.rs
delbonis, line 1: Should refactor this to use a context trait instead of directly operating on `NodeStorage`.

delbonis (dismissed):
Seems fine but should take an opportunity to do some cleanup here.

crates/btcio/src/reader/query.rs
bewakes: This makes sense. Done in https://github.com/alpenlabs/alpen/pull/2152/commits/cb9b3a17bc2449d6b280c6d35185022e39820d4c

crates/btcio/src/reader/query.rs
bewakes: PTAL https://github.com/alpenlabs/alpen/pull/2152/commits/1b6c44dbfe86edf3421ae3724be56bea036e3b4b

storopoli (approved):
ACK 1b6c44dbfe86edf3421ae3724be56bea036e3b4b

crates/btcio/src/reader/query.rs
```diff
@@ -105,24 +149,84 @@ pub async fn bitcoin_data_reader_task<E: BlockSubmitter>(
...
+                    "reverted unmaterialized canonical L1 entries for ASM replay"
+                );
+            }
+            return Ok(());
+        }
+
+        if height == earliest_height {
+            bail!(
+                "no ASM anchor state found for canonical L1 heights {earliest_height} through {tip_height}; refusing to resume beyond the reorg lookback window"
+            );
```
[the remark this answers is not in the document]
bewakes: Done in https://github.com/alpenlabs/alpen/pull/2152/commits/4570b8abac95e3853223413c255aaa0ef15da2f2

voidash (approved):

krsnapaudel (approved):
LGTM

prajwolrg (approved):

PR #2153, opened 2026-07-26
feat(ol)!: explicit, proof-bound EE predicate rotations

prajwolrg:
## Description

This PR makes EE predicate-key (update VK) rotations explicit, consensus-committed, and proof-bound on the OL side. Previously, applying an admin `EeStfVk` update silently swapped the account's `update_vk` the moment the ASM log was processed, which invalidates every in-flight update still proven under the old key and gives the EE no deterministic event in its own history to derive a fork boundary from.

The model is account-sovereign: a snark account update may declare a `new_predicate`, and the OL applies it without restricting what the account rotates to — the declaration is bound into the update proof claim, so the account's current predicate authorizes its successor. Admin enactment no longer touches the key directly; it only queues a predicate-update message (type `0x20`, admin-sourced) into the target account's inbox as a signal. For the Alpen EE, declaring exactly the queued key — terminating the batch at that message — is the EE's own policy, enforced by its guest in the alpen-ee repo, not by the OL.

Review entry points: `crates/ol/stf/src/transaction_processing.rs` (declaration-driven activation), `crates/snark-acct-sys/src/verification.rs` (`compute_update_claim` binding), and the two `.ssz` schema changes (`ol/chain-types/ssz/transaction.ssz`, `snark-acct-types/ssz/outputs.ssz`). The optional fields use small `StableContainer` wrappers because the SSZ codegen only permits `Optional` there; the wrapper docstrings explain this.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

**Why this is sound.** The only way to rotate an account's key is to produce a valid update under its *current* key: the declared rotation rides in `UpdateOutputs`, which is part of `UpdateProofPubParams`, and the OL STF reconstructs the claim from the txid-committed `SauTxUpdateData::new_predicate`. A declaration stripped or altered anywhere between the EE and the OL produces a claim that does not match the proof's public params, so verification fails. The OL attaches no semantics to the queued admin message itself — consuming it does not rotate anything (there is a test pinning this) — so there is no second activation path to diverge from. Every failure mode is fail-closed: the worst outcome is an unprovable or rejected update, never wrong state.

**The guest side is handled in the alpen-ee repo, not here.** The Alpen EE's policy: the batch builder checks the inbox for a pending rotation message; if present it terminates the batch with that message as the last one consumed and declares the queued VK in the update it builds; the proof guest mirrors that by accumulating the rotation it consumed. The existing `check_obligations` equality in `ee-acct-runtime` (expected vs accumulated outputs) is the hook that enforces this policy inside the proof — a builder that declares a key it didn't consume (or vice versa) cannot produce a valid update. Until the alpen-ee side lands, nothing produces declarations, so all values are `None` and behavior is unchanged.

**DA reconstruction** (Codex catch): a declared rotation changes `update_vk`, which the snark-account DA diff didn't carry, so DA-reconstructing nodes would diverge at the first rotation. The diff now includes an update-VK register (unset registers add no bytes, so rotation-free diffs encode exactly as before, and the frozen V1 golden fixture still passes), and an epoch-reconstruction test pins that a rotated epoch reproduces the directly-executed state root.

Testing: STF tests cover declared-rotation activation, that consuming the queued message alone does not rotate, and DA epoch reconstruction across a rotation; SSZ roundtrip proptests cover both declared/undeclared arms; RPC conversion has unit tests for both. Full workspace unit suite plus clippy pass, and every commit in the stack compiles standalone.

AI disclosure: this PR was developed with Claude Code (Claude Fable 5), with design decisions directed and reviewed by the author.

- [x] Yes
- [ ] No

- [OL Upgradability](https://app.notion.com/p/3a4901ba000f8176ae3ed73346366781) (§3: switch at the referencing batch, terminate-at-boundary)
- [Live ELF/VK Switching](https://app.notion.com/p/390901ba000f8032aca7d5d4ddc8eb8d)
- Jira: STR-4000 (VK-update boundary/reflect seam, epic STR-3995)

## Checklist

## Related Issues

STR-3995 (Alpen Upgradability epic) / STR-4000.


crates/ee-acct-runtime/src/verification_state.rs
```diff
@@ -263,6 +263,16 @@ impl<'a, E: ExecutionEnvironment> EeVerificationState<'a, E> {
     /// were supposed to have been dealt with but weren't.
     pub(crate) fn check_obligations(&self) -> EnvResult<()> {
         // Check that the expected outputs match the ones we accumulated.
+        //
+        // Predicate rotations: the batch-building side (alpen-ee repo) detects
+        // a pending admin predicate-update message in the inbox, terminates
+        // the batch with that message as the last one consumed, and declares
+        // the new VK in the update outputs. This verification path must
+        // mirror that by accumulating the rotation when it processes the
+        // admin message; until it does, an update declaring a rotation fails
```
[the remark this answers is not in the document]
prajwolrg: Deliberately out of scope here: the accumulation mirror is part of the alpen-ee repo's batch-builder/guest implementation (detect queued rotation → terminate batch → declare queued VK → accumulate on consumption). Until it lands nothing produces declarations, so this equality is unaffected; see the PR notes and the comment above this check.

crates/ol/stf/src/transaction_processing.rs
```diff
@@ -134,7 +138,26 @@ fn process_update_tx<S: IStateAccessorMut>(
...
+    // TODO: enforce that contract here rather than trusting it: reject
+    // updates whose declared `SauTxUpdateData::new_predicate` disagrees with
+    // the rotation derived from the consumed messages, and require a consumed
+    // rotation message to be the last message the update consumes
+    // (terminate-at-boundary), so the committed tx data is authoritative for
+    // every snark account, not just those whose guest enforces it.
+    let consumed_rotations: Vec<PredicateKey> = op
+        .messages_iter()
+        .filter_map(parse_predicate_update_message)
+        .collect();
```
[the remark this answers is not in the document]
prajwolrg: Resolved by the restructured stack: activation is now driven solely by the txid-committed declaration (d09e35259), and the message-parsing activation path was removed — there is no second derivation left to diverge from. A test pins that consuming the queued message without declaring does not rotate.

crates/ol/stf/src/transaction_processing.rs
```diff
@@ -134,7 +138,26 @@ fn process_update_tx<S: IStateAccessorMut>(
...
     // 3. Actually take balance and write new account inner state.
     let serial = account_state.serial();
-    let upd = sau_payload.operation().update();
+    let op = sau_payload.operation();
+    let upd = op.update();
+    // Predicate rotations activate on consumption: if this update consumed
+    // admin predicate-update messages, the account's update VK rotates now.
+    // The update consuming the message is therefore the last one verified
+    // under the old key, per the Alpen upgrade design. Later rotations in
+    // the same update win, preserving admin ordering.
```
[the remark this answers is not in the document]
prajwolrg: With declaration-driven activation the OL attaches no semantics to the rotation message, so this hazard moved to the EE's own policy: terminating the batch at the rotation message is enforced by the alpen-ee guest (a builder that consumes past it produces outputs that fail the guest's expected/accumulated equality). See the PR notes.

crates/ol/stf/src/transaction_processing.rs
```diff
@@ -153,6 +154,21 @@ fn process_update_tx<S: IStateAccessorMut>(
...
+        // proof claim, so the account's current predicate authorizes its
+        // successor; this update is the last one verified under the old key
+        // and every later update must be proven under the new one. For the
+        // Alpen EE, admin enactment queues a predicate-update message into
+        // the account inbox, and declaring exactly the queued key (with the
+        // rotation message terminating the batch) is the EE's own policy,
+        // enforced by its guest in the alpen-ee repo — not by the OL.
+        if let Some(new_vk) = upd.new_predicate() {
+            info!(account_id = %target, "activating declared predicate key rotation");
+            acct_tstate.set_update_vk(new_vk.clone());
```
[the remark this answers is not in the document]
prajwolrg: Good catch — fixed in 0c5c0219a: the snark-account DA diff now carries an update-VK register (mirroring the account-init encoding; unset registers are byte-identical to the old format), apply_snark_diff sets the key on reconstruction, and test_apply_da_epoch_snark_update_with_rotation pins that a rotated epoch reproduces the directly-executed state root.

crates/ol/chain-types/ssz/transaction.ssz
```diff
@@ -30,12 +31,30 @@ class SauTxProofState(Container):
...
+### verifies every later update.
+###
+### This is a StableContainer because Optional fields are only allowed in
+### StableContainers, not plain Containers, so we need this wrapper to make
+### the predicate optional.  The spare capacity also lets future optional
+### fields slot in without changing the encoding of existing data.
+#~# derive: std::default::Default
+class SauTxNewPredicate(StableContainer[4]):
+    #~# external_kind: container
+    predicate: Optional[strata_predicate.PredicateKey]
```
delbonis, line 49: Should make this `StableContainer[8]` since it pads out to a full byte anyways.

crates/ol/da/src/types/snark.rs
delbonis, line 1: Make sure the field order matches that of the fields in the snark account.  It doesn't actually matter but it's good to keep it consistent.

crates/ol/msg-types/src/predicate_update.rs
delbonis, line 1: Just to clarify, this gets submitted at an epoch boundary so that the predkey doesn't get duplicately-persisted in DA right?

crates/snark-acct-types/ssz/outputs.ssz
```diff
@@ -20,10 +21,28 @@ class OutputMessage(Container):
...
+### verifies every later update.
+###
+### This is a StableContainer because Optional fields are only allowed in
+### StableContainers, not plain Containers, so we need this wrapper to make
+### the predicate optional.  The spare capacity also lets future optional
+### fields slot in without changing the encoding of existing data.
+#~# derive: std::default::Default
+class OutputNewPredicate(StableContainer[4]):
+    #~# external_kind: container
+    predicate: Optional[strata_predicate.PredicateKey]
```
delbonis, line 39: Should be `StableContainer[8]` since it pads out to a full byte anyways.

delbonis (changes requested):
Some cleanups requested but overall fairly clean.

crates/ol/chain-types/ssz/transaction.ssz
prajwolrg: Done in 4a9eb30bd (along with the outputs one).

crates/snark-acct-types/ssz/outputs.ssz
prajwolrg: Done in 4a9eb30bd.

crates/ol/da/src/types/snark.rs
prajwolrg: Reordered in f87f3f436 to `update_vk, proof_state, seq_no, inbox` to match `SnarkAccountState` (`DaProofStateDiff` already matched `ProofState`).

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -106,26 +106,32 @@ struct SnarkDelta {
     base_proof_state: DaProofState,
     final_proof_state: DaProofState,
     inbox: DaLinacc<InboxBuffer>,
+    base_update_vk: U16LenBytes,
+    final_update_vk: U16LenBytes,
```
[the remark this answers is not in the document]
prajwolrg: Fixed in 09019de4f — the per-diff loop now adds the rotated VK register's bytes (2-byte length prefix + the actual key length, which `SnarkDelta` already tracks). Also left a FIXME on the pre-existing gap where the new-account branch never counted VK bytes despite its comment claiming a `MAX_VK_BYTES` bound; fixing that properly needs the VK length tracked per `NewAccountRecord`.

crates/ol/msg-types/src/predicate_update.rs
prajwolrg: Right — it's persisted in DA only once. 

The admin log is just buffered during the block carrying the manifest; the inbox enqueue happens at the epoch-terminal drain (`process_epoch_terminal`), which runs on a non-DA-tracking accessor (`DaAccumulatingState` covers only pre-drain execution), and reconstruction replays the manifests/drain from L1 directly — so the queued message never enters the DA payload.

The one copy that *is* in DA is the `update_vk` register at activation. Strictly we could skip even that today, since the rotation originates on L1 and is re-derivable from the admin tx — but the register is what keeps the scheme generic: a snark account's update can declare a rotation that didn't originate from L1, so activation has to be persisted in DA in its own right.

crates/ol/da/src/types/snark.rs
```diff
@@ -173,50 +173,62 @@ impl CompoundMember for DaProofStateDiff {
...
+///
+/// Field order mirrors `SnarkAccountState` for consistency.
 #[derive(Debug)]
 pub struct SnarkAccountDiff {
-    /// Sequence number counter diff.
-    pub seq_no: DaCounter<CtrU64ByU16>,
+    /// Update predicate key (VK) register, set when an update declares a
+    /// rotation. Carries the serialized key bytes, as in
+    /// [`SnarkAccountInit`](super::ledger::SnarkAccountInit).
+    pub update_vk: DaRegister<U16LenBytes>,
```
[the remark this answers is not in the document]
storopoli: We don't care much about this right?

storopoli (changes requested):
Couple blockers:

* Most obvious: this is approach B of [OL Upgradability](https://app.notion.com/p/OL-Upgradability-3a4901ba000f8176ae3ed73346366781)
* The same doc says that DA costs is avoided which is not what this PR does
* In the doc, the ASM still reconstructs the claim from its own record, so the declared key is provably the one the genuine log carried. In the PR the OL doesn't check any correspondence between the declaration and the queued admin message at all; an account can rotate to anything its current predicate signs off on. It's a strictly weaker OL-side invariant than approach B states.

prajwolrg:
> Couple blockers:
> 
> * Most obvious: this is approach B of [OL Upgradability](https://app.notion.com/p/OL-Upgradability-3a4901ba000f8176ae3ed73346366781)
> * The same doc says that DA costs is avoided which is not what this PR does
> * In the doc, the ASM still reconstructs the claim from its own record, so the declared key is provably the one the genuine log carried. In the PR the OL doesn't check any correspondence between the declaration and the queued admin message at all; an account can rotate to anything its current predicate signs off on. It's a strictly weaker OL-side invariant than approach B states.

@storopoli you've mixed up things here. This PR isn't approach B of the [OL Upgradability](https://app.notion.com/p/3a4901ba000f8176ae3ed73346366781) doc, it's STR-4000 under the Alpen Upgradability epic: EE predicate/VK rotation, i.e. the Alpen EE account rotating its own update_vk.

Here, we allow EE account's own current predicate authorizes its successor via the update proof itself (declared new_predicate is bound into UpdateProofPubParams), so there's no ASM correspondence to check — the "genuine log" role is played by the account's own proof, not an ASM record.

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -361,6 +369,11 @@ impl EpochDaAccumulator {
         //   + if snark: 32 (state root) + 2 (vk len prefix) + vk bytes
         // We don't know VK size here so use MAX_VK_BYTES as upper bound.
         // In practice most VKs are much smaller, but this is a conservative estimate.
+        // FIXME: the VK bytes are not actually counted despite the comment
```
bewakes: Do we need a ticket here?

crates/snark-acct-types/ssz/outputs.ssz
```diff
@@ -20,10 +21,30 @@ class OutputMessage(Container):
...
 ### Collection of outputs from an update
 class UpdateOutputs(Container):
     ### Native asset transfers
     transfers: List[OutputTransfer, MAX_TRANSFERS]
 
     ### Messages with data payloads
     messages: List[OutputMessage, MAX_MESSAGES]
+
+    ### New update predicate declared by this update, if it rotates the key.
+    new_predicate: OutputNewPredicate
```
bewakes, line 50: I wonder if it makes more sense to make this `Optional` instead of what it contains. This is because, if a container is present then we'd expect it to contain the predicate key.

bewakes (commented):
Looks good. I've left some comments and one suggestion is to make the verbose comments concise - these are basically repeating what the design doc already describes.

crates/ol/da/src/types/snark.rs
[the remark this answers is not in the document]
bewakes: If we plan on testing upgrades with this change then it might matter. If this is more greenfield then doesn't matter imo.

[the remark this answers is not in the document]
prajwolrg: Yes, I don't think this is an issue yet.

crates/snark-acct-types/ssz/outputs.ssz
prajwolrg: Yes, this is `Optional`. The `OutputNewPredicate` is a `StableContainer` with a single `Optional` field. I followed the same pattern as in `SauTxLedgerRefs`.

crates/ol/state-support-types/src/da_accumulating_layer.rs
prajwolrg: No ticket needed — fixed directly in a8468ba88.

bewakes (approved):

storopoli (approved):
ACK 4f353806b84a96a57e694837f6e779735780f334

functional-tests/tests/alpen_client/test_ee_predicate_fullnode_sync.py
krsnapaudel, line 1: Do we need this env after deleting the fn tests? This is in `entry.py`.
```
        el_ol_ee_predicate_transition": EeOLEnv(
            pre_generate_blocks=110,
            admin_confirmation_depth=2,
            fund_test_cli_wallet=True,
        ),
```

Also this helper. This is in `test_cli.py`.
```
def create_ee_predicate_update(
    seq_no: int,
    predicate: str,
    admin_xpriv: str,
    btc_url: str,
    btc_user: str,
    btc_password: str,
    fee_rate: int = 2,
    commit_output_sats: int = 20_000,
) -> dict:
```

krsnapaudel (approved):
Some nits.

crates/ol/msg-types/src/predicate_update.rs
delbonis: Right that's fine.

crates/ol/da/src/types/snark.rs
delbonis, line 186: This is a bit of a nit but most predicates are likely to be smaller than 128 bytes so if we used a varint then we could save a byte in most cases (as opposed to using a u16 len prefix).

crates/ol/da/src/types/snark.rs
```diff
@@ -173,50 +173,62 @@ impl CompoundMember for DaProofStateDiff {
...
 
     /// Proof state diff.
     pub proof_state: DaProofStateDiff,
 
+    /// Sequence number counter diff.
+    pub seq_no: DaCounter<CtrU64ByU16>,
```
delbonis, line 192: Wait, how did we not already have this?

crates/ol/da/src/types/payload.rs
```diff
@@ -302,6 +302,12 @@ fn apply_snark_diff<T: IAccountStateMut>(
         next_seqno,
     );
 
+    if let Some(vk_bytes) = diff.update_vk.new_value() {
+        let buf = PredicateKeyBuf::try_from(vk_bytes.as_slice())
+            .map_err(|_| DaError::InvalidStateDiff("invalid predicate key in snark diff"))?;
```
delbonis: Make an error variant for this, this is opaque.

crates/ol/stf/src/manifest_processing.rs
```diff
@@ -386,6 +399,21 @@ fn process_ee_predicate_key_update<S: IStateAccessorMut>(
...
+/// [`PREDICATE_UPDATE_MSG_TYPE_ID`]. The source is the admin message account
+/// id, a reserved system id that no ledger account can occupy.
+fn build_predicate_update_message(new_vk: &PredicateKey, cur_epoch: u32) -> MessageEntry {
+    let body = ssz::Encode::as_ssz_bytes(new_vk);
+    let msg = OwnedMsg::new(PREDICATE_UPDATE_MSG_TYPE_ID, body)
+        .expect("predicate update message type id is in bounds");
+    let payload = MsgPayload::from_bytes_valueless(msg.to_vec())
+        .expect("predicate key fits in message payload");
+    MessageEntry::new(ADMIN_MSG_ACCT_ID, cur_epoch, payload)
+}
```
delbonis, line 444: Actually this is pretty important and I'm annoyed at myself for not catching it on the last review cycle:
* Should define a common MsgData type for this in a shared crate instead of directly encoding/decoding in-place on the fly.
  * Also it could make sense to separate the "making the update message" and "making the `MessageEntry`" parts since they're different layers of the stack.
* Should *not* use SSZ for this, it adds a useless length prefix since we already know the bounds of the container.

*In general*, SSZ should not be used for message/log bodies.

crates/snark-acct-types/src/outputs.rs
```diff
@@ -22,6 +55,7 @@ impl UpdateOutputs {
             messages: messages
                 .try_into()
                 .expect("messages should not exceed capacity"),
+            new_predicate: OutputNewPredicate::new_empty(),
         }
     }
```
delbonis: This is going to be annoying to keep expanding over time, make this constructor `pub(crate)` and then define a `new_empty`/`default`.

crates/snark-acct-types/ssz/outputs.ssz
delbonis, line 50: Ahhh crap crap crap I'm not liking this type structure, I wish SSZ was not as restrictive as it is.

This isn't properly backwards compatible, forget this comment, let's follow up on this later.

delbonis (changes requested):
Actually here's couple more important requests, I should have caught these on the last review cycle.

crates/ol/da/src/types/snark.rs
prajwolrg: In production we'll be using the SP1Groth16 predicate, which is larger than 128 bytes. So I don't think using varint is beneficial here.

functional-tests/tests/alpen_client/test_ee_predicate_fullnode_sync.py
prajwolrg: Good catch, removed both in 17f4da63c.

crates/ol/da/src/types/snark.rs
prajwolrg: Yeah, we already had this — `seq_no` diffing existed before this PR (it was already the first field in `SnarkAccountDiff`, see main). This hunk isn't adding it, it's just relocating the same field below the new `update_vk` so the order matches `SnarkAccountState`; git just renders the reorder as a remove+re-add since `update_vk` was inserted above it. Content is byte-identical, nothing new here.

crates/ol/da/src/types/payload.rs
prajwolrg: Added `DaError::InvalidPredicateKey(#[from] PredicateError)` and propagated in 44ece96ed.

crates/ol/stf/src/manifest_processing.rs
prajwolrg: Addressed in 6faa7d855:
- Added `PredicateUpdateMsgData` in `strata-ol-msg-types` (crates/ol/msg-types/src/predicate_update.rs), alongside `DepositMsgData`/`WithdrawalMsgData`, encoded via `strata-codec` instead of SSZ. It carries the predicate key's raw `[id: u8][condition: bytes...]` form in a `VarVec` (varint length prefix, not a fixed one), so the crate still doesn't need to depend on `strata-predicate`.
- Split `build_predicate_update_message` into `build_predicate_update_payload` (message bytes) + the thin `MessageEntry` wrapper, mirroring the deposit path's layering.
- Added `try_as_predicate_update` to `OLMessageExt` for symmetry with the other message types, for whenever the EE-side consumer needs to decode it.
- The one test that hand-rolled the SSZ body now calls the shared `build_predicate_update_payload` instead of duplicating the encoding.

crates/snark-acct-types/src/outputs.rs
prajwolrg: Made `new` `pub(crate)` and added `with_transfers`/`with_messages` builders alongside the existing `new_empty`/`with_new_predicate`, in b52abb0.

crates/ol/stf/src/manifest_processing.rs
```diff
@@ -365,16 +368,26 @@ fn process_ee_predicate_key_update<S: IStateAccessorMut>(
...
+            // The rotation is not applied here. It lands in the account's
+            // inbox and takes effect when an account update consumes the
+            // message (see `process_update_tx`). That makes the update that
+            // consumes the message the last one verified under the old key —
+            // the consensus-level fork boundary of the Alpen upgrade design —
+            // and gives the EE a deterministic position in its inbox
+            // ordering to derive the fork activation from. Applying the key
+            // immediately would instead reject every in-flight update still
+            // proven under the old key.
+            snark.insert_inbox_message(update_msg)?;
```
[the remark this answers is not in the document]
prajwolrg: Tracked as STR-4139: https://alpenlabs.atlassian.net/browse/STR-4139

crates/snark-acct-types/src/outputs.rs
prajwolrg: Update: removed `new` entirely rather than keeping it `pub(crate)` — only `new_empty` plus the `with_*` builders remain, in 22b78b5 (was b52abb0 before amending).

delbonis (approved):
(Approving un-re-reviewed.)

crates/ol/state-support-types/src/da_accumulating_layer.rs
```diff
@@ -106,26 +107,32 @@ struct SnarkDelta {
...
+            base_update_vk: base_update_vk.clone(),
+            final_update_vk: base_update_vk,
         }
     }
 
     fn update_from_state<T: ISnarkAccountState>(&mut self, state: &T) {
         self.final_seq_no = *state.seqno().inner();
         self.final_proof_state =
             DaProofState::new(state.inner_state_root(), state.next_inbox_msg_idx());
+        self.final_update_vk = U16LenBytes::new(state.update_vk().as_buf_ref().to_bytes());
```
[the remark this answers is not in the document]
storopoli: @prajwolrg

crates/ol/da/src/types/payload.rs
```diff
@@ -302,6 +301,11 @@ fn apply_snark_diff<T: IAccountStateMut>(
         next_seqno,
     );
 
+    if let Some(vk_bytes) = diff.update_vk.new_value() {
+        let buf = PredicateKeyBuf::try_from(vk_bytes.as_slice())?;
+        snark.set_update_vk(buf.to_owned());
```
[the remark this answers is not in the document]
prajwolrg: Fixed in 195a8bc: added a check against `MAX_CONDITION_LEN` before the owned conversion, in both call sites that go through `PredicateKeyBuf::try_from(...).to_owned()` (new-account init and account-diff rotation). Returns `DaError::InvalidPredicateKey` instead of panicking.

The root cause is also being fixed upstream in alpenlabs/strata-common#155 (`PredicateKeyBuf::try_from` will validate the condition length itself); once alpen bumps that pinned dependency, this local check can be dropped.

crates/ol/state-support-types/src/da_accumulating_layer.rs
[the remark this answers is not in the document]
prajwolrg: Fixed in 1e8bcb9: validated through the existing `try_as_buf_ref()` at both places an untrusted `PredicateKey` gets installed — the SAU-declared rotation in `transaction_processing.rs` (now rejected with `TxStructureCheckFailed`) and the ASM-log-declared rotation in `manifest_processing.rs` (now dropped, matching the existing convention for unusable manifest data). `da_accumulating_layer.rs` itself doesn't need a change since it only ever reads back a key that was already validated when it was written.

The root cause is also being fixed upstream in alpenlabs/strata-common#155, which removes the panicking `as_buf_ref` entirely in favor of `try_as_buf_ref`; once alpen bumps that pinned dependency this local guard becomes redundant.

PR #2166, opened 2026-07-30
fix(rpc): expose SAU operation seqno in OL summaries

krsnapaudel:
## Description

Fixes OL account summary RPCs to expose the SAU operation sequence number for update inputs instead of the stored post-state account sequence.

## Details

- Uses `AccountUpdateRecord::orig_acct_seq_no()` in block and epoch summary paths.
- Keeps `next_seq_no` as the post-state account value.
- Documents the operation `seq_no` vs post-state `next_seq_no` distinction.
- Adds regression coverage correlating epoch summary, block summary, and manifest by state root.

Commit 2 takes a different approach:
- Types `AccountUpdateRecord::seq_no` as `NonZeroU64`, making
  `orig_acct_seq_no()` infallible (`u64` rather than `Option<u64>`) and
  removing the per-call-site guards in the summary and manifest paths.
- Adds `WorkerError::SnarkUpdateZeroSeqno`, so a zero post-update seqno fails
  at indexing rather than surfacing as an RPC internal error.
- Drops the three tests that covered those guards; they could only be written
  by constructing a record the type now rejects.

If commit 2 is a better approach, we can fold 1 and 2. Basically keep commit 2 approach.

The PR was created with help from Codex and Claude.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-4102](https://alpenlabs.atlassian.net/browse/STR-4102)

[STR-4102]: https://alpenlabs.atlassian.net/browse/STR-4102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/strata/src/rpc/node.rs
+++ b/bin/strata/src/rpc/node.rs
@@ -10,7 +10,10 @@ use jsonrpsee::core::RpcResult;
 use ssz::{Decode, Encode};
 use strata_acct_types::MessageEntry;
 use strata_checkpoint_types::EpochSummary;
-use strata_db_types::{ol_block::BlockAvailability, ol_state_index::InboxMessageRecord};
+use strata_db_types::{
+    ol_block::BlockAvailability,
+    ol_state_index::{AccountUpdateRecord, InboxMessageRecord},
+};
 use strata_identifiers::{
     AccountId, Epoch, EpochCommitment, Hash, L1BlockCommitment, L1Height, L2BlockCommitment,
     OLBlockCommitment, OLBlockId, OLTxId, RBuf32,
@@ -123,6 +126,22 @@ fn local_inbox_message_range(
     })?;
 
     Ok(local_start..local_end)
+}
+
+/// Returns the SAU operation seqno for an indexed update record.
+///
+/// Indexed update records store post-state account seqnos. Missing operation
+/// seqnos indicate an indexing invariant violation and fail the RPC.
+fn require_operation_seq_no(
+    account_id: AccountId,
+    epoch: Epoch,
+    record: &AccountUpdateRecord,
+) -> RpcResult<u64> {
+    record.orig_acct_seq_no().ok_or_else(|| {
+        internal_error(format!(
+            "missing operation seq_no in update record for account {account_id} epoch {epoch}",
+        ))
+    })
 }
 
 impl<P: OLRpcProvider> OLRpcServer<P> {
@@ -422,7 +441,7 @@ impl<P: OLRpcProvider> OLRpcServer<P> {
         // `Pending` triple capturing the inbox slice they consumed.
         struct Pending {
             block_commitment: OLBlockCommitment,
-            seq_no: u64,
+            operation_seq_no: u64,
             new_state_root: Hash,
             extra_data: Vec<u8>,
             cursor_start: u64,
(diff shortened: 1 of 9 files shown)


bewakes (approved):

bin/strata/src/rpc/node.rs
```diff
@@ -125,6 +128,22 @@ fn local_inbox_message_range(
...
+    account_id: AccountId,
+    epoch: Epoch,
+    record: &AccountUpdateRecord,
+) -> RpcResult<u64> {
+    record.orig_acct_seq_no().ok_or_else(|| {
+        internal_error(format!(
+            "missing operation seq_no in update record for account {account_id} epoch {epoch}",
+        ))
+    })
+}
```
storopoli: Shouldn't this be a method in `AccountUpdateRecord` instead?

crates/ol/rpc/types/src/account_summary.rs
```diff
@@ -14,13 +14,17 @@ use strata_snark_acct_types::{ProofState, UpdateInputData};
...
+
     /// Balance of account at the end of this epoch in sats.
     final_balance: u64,
+
     /// Account's state root at the end of this epoch.
     final_state_root: HexBytes32,
-    /// Update inputs for this epoch if present
+
+    /// Update inputs for this epoch, if present.
     update_inputs: Vec<RpcUpdateInputData>,
```
storopoli: "if present" means `Option<T>`. Should we use that instead?

storopoli (changes requested):
Function-wise looks good. Some design considerations.

bin/strata/src/rpc/node.rs
krsnapaudel: @storopoli It's already right? This is a convenience helper to check the return value is not `None`.

crates/ol/rpc/types/src/account_summary.rs
krsnapaudel: @storopoli It's a `Vec`. Should be empty if not present. I will change the doc.

bin/strata/src/rpc/node.rs
storopoli: I've meant

```rs
impl AccountUpdateRecord {
  fn require_operation_seq_no()
}
```

I honestly don't know the answer.

crates/ol/rpc/types/src/account_summary.rs
storopoli: Ok

bin/strata/src/rpc/node.rs
krsnapaudel: The fn there does not make sense. We could change the type for `seq_no`. Operation seq no is the seqno for applied update. The type keeps post apply seqno. I will add this as a separate commit. So we can skip if it's too much.

krsnapaudel:
@storopoli I have updated the type to enforce non-zero semantics. If the second commit is approved, I will fold them into one. Or since the commits are squashed, this is not necessary.

delbonis (approved):

storopoli (approved):
ACK 45add0c135bfdd185b6499ab5a374164f0306586

PR #2189, opened 2026-08-10
fix(datatool): support bitcoin network keys

purusang:
## Description

Adds `bitcoin` as a supported `strata-datatool --bitcoin-network` value so `genxpriv` can generate Bitcoin mainnet extended private keys.

The datatool README now documents mainnet key generation with `--bitcoin-network bitcoin`. The parser has focused tests for the supported values and rejection cases.

### Type of Change

- [x] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

`genxpriv` is the command affected by the Bitcoin network selection because it controls the BIP32 extended private key network prefix. The pubkey derivation commands continue to derive raw public keys from the xpriv and do not need a network flag.

This PR was prepared with AI assistance.

- [ ]  Yes
- [x]  No

## Checklist

Tests run:

- `cargo fmt --package strata-datatool`
- `CARGO_TARGET_DIR=/tmp/alpen-target-datatool-mainnet cargo test -p strata-datatool parse_network`

## Related Issues

N/A

### DIFF
--- a/bin/datatool/src/args.rs
+++ b/bin/datatool/src/args.rs
@@ -17,7 +17,7 @@ use crate::{
 pub(crate) struct Args {
     #[argh(
         option,
-        description = "network name [bitcoin, signet, regtest]",
+        description = "bitcoin network name accepted by `bitcoin::Network`",
         short = 'b'
     )]
     pub(crate) bitcoin_network: Option<String>,
--- a/bin/datatool/src/args.rs
+++ b/bin/datatool/src/args.rs
@@ -15,7 +15,11 @@ use crate::{
 /// Args.
 #[derive(FromArgs)]
 pub(crate) struct Args {
-    #[argh(option, description = "network name [signet, regtest]", short = 'b')]
+    #[argh(
+        option,
+        description = "network name [bitcoin, signet, regtest]",
+        short = 'b'
+    )]
     pub(crate) bitcoin_network: Option<String>,
 
     #[argh(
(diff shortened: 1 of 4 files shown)


bin/datatool/src/util.rs
```diff
@@ -29,26 +29,30 @@ pub(crate) const SEQKEY_ENVVAR: &str = "STRATA_SEQ_KEY";
...
 }
 
+fn parse_network(network: &str) -> anyhow::Result<Network> {
+    match network {
+        "bitcoin" => Ok(Network::Bitcoin),
+        "signet" => Ok(Network::Signet),
+        "regtest" => Ok(Network::Regtest),
+        n => anyhow::bail!("unsupported network option: {n}"),
+    }
+}
```
prajwolrg: I don't think we need to implement this. There's a `FromStr` implementation for `Network` type.

prajwolrg: If we use that I don't think we'd even need all the tests.

prajwolrg (changes requested):

bin/datatool/src/util.rs
purusang: Yup, thats better.

bin/datatool/src/args.rs
```diff
@@ -17,7 +17,7 @@ use crate::{
 pub(crate) struct Args {
     #[argh(
         option,
-        description = "network name [bitcoin, signet, regtest]",
+        description = "bitcoin network name accepted by `bitcoin::Network`",
         short = 'b'
     )]
     pub(crate) bitcoin_network: Option<String>,
```
prajwolrg:
```
pub(crate) bitcoin_network: Option<Network>,
```
Let's  directly use the Network type here instead of parsing it later.

prajwolrg (changes requested):

prajwolrg (approved):
LGTM!

PR #2197, opened 2026-08-12
Do not post checkpoints twice - V2

bewakes:
## Description

 This PR prevents stale checkpoints from being re-posted after a restart when ASM lags behind the checkpoint builder,
  while preserving commit/reveal recovery across crashes. The core idea is similar to the predecessor(#2186) - an injectable `PublishPolicy`. This PR adds some corner cases on the design itself rather than patching them one by one later.

  - Adds a generic `PublishPolicy` hook to the `btcio` broadcaster. A policy can `Publish`, `Defer`, or `Abandon` a
  transaction immediately before `sendrawtransaction`.

  - Wires `CheckpointPublishPolicy` into that hook. It identifies checkpoint commit/reveal pairs, reads canonical ASM state, publishes checkpoints ahead of ASM, defers accepted checkpoints that are not yet reorg-safe, and abandons safe stale checkpoints.

  - Uses `Defer` when canonical state or publication linkage is temporarily unavailable, keeping the transaction
  `Unpublished` for reevaluation.

  - Persists writer linkage before exposing transactions to the broadcaster and inserts commit/reveal entries
  atomically, preventing crash-created half-pairs.

  - Extends startup reconciliation to consider every queued proof variant and delete checkpoint artifacts only when all
  corresponding writer and broadcaster entries can be safely abandoned.

  - Adds durable `Abandoned` handling, requeues rebuilt intents at the writer tail, and recovers noncontiguous queues.

  - Adds regression coverage for stale, deferred, in-flight, duplicate, commit/reveal, and crash-recovery paths.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
**[AI usage]**: Built with the help of codex with close supervision.

- [ ]  Yes
- [ ]  No

## Checklist


bewakes:
@codex review

bewakes:
@codex review

bewakes:
@codex review

bewakes:
@codex review

bewakes:
@codex review

bewakes:
@codex review

crates/ol/checkpoint/src/l1/publication.rs
```diff
@@ -0,0 +1,508 @@
...
+            ));
+        }
+        Ok(RevealKind::Other) => return Ok(Publication::Other),
+        Err(()) => {}
+    }
+
+    let Some(reveal_idx) = idx.checked_add(1) else {
+        return Ok(Publication::Unknown);
+    };
+    if reveal_idx >= broadcast.get_next_tx_idx()? {
```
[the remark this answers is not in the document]
bewakes: legacy is not an issue here.

krsnapaudel (approved):
LGTM

bin/strata/src/checkpoint_reconcile.rs
```diff
@@ -96,6 +136,98 @@ fn checkpoint_commitments_from_epoch(
...
+    }
+    Ok(all_cancelled)
+}
+
+#[cfg(feature = "sequencer")]
+/// Abandons one writer intent only when no associated transaction may be live on L1.
+///
+/// Unbundled intents are terminalized directly. Bundled intents are preserved if
+/// publication has started for either broadcaster entry.
+fn cancel_writer_intent(db: &impl DatabaseBackend, intent: IntentEntry) -> Result<bool> {
```
delbonis: Attributes after comments.

crates/ol/sequencer/src/checkpoint/publication.rs
delbonis, line 1: None of this logic should be included in the strata binary itself.  The sequencer crate is a much better place.

Also, reconsider every fn name in this module.  Most of them are extremely vague.  Module-level fns usually should be verbs.

crates/ol/sequencer/src/checkpoint/reconcile.rs
delbonis, line 1: Same here, actually.  Move to sequencer crate.

bin/strata/src/services.rs
```diff
@@ -90,16 +91,22 @@ mod sequencer_services {
             nodectx.storage().handle().clone(),
             broadcast_db,
         ));
+        let btcio_params = super::build_btcio_params(
```
delbonis: don't use `super::` for a ref like this

bin/strata/src/checkpoint_publish.rs
```diff
@@ -0,0 +1,512 @@
...
+use strata_l1_envelope_fmt::parser::parse_envelope_payload;
+use strata_l1_txfmt::{MagicBytes, ParseConfig};
+use strata_storage::NodeStorage;
+use tracing::warn;
+
+pub(crate) struct CheckpointPublishPolicy {
+    storage: Arc<NodeStorage>,
+    parser: ParseConfig,
+}
+
```
delbonis: Can this be made generic over the storage?

Also should probably not be in the strata binary crate.

bin/strata/src/checkpoint_publish.rs
```diff
@@ -0,0 +1,512 @@
...
+        Err(()) => Publication::Unknown,
+    })
+}
+
+enum Publication {
+    Checkpoint {
+        checkpoint: Box<CheckpointPayload>,
+        member: PairMember,
+        commit_status: Option<L1TxStatus>,
+        reveal_status: Option<L1TxStatus>,
```
delbonis: It almost never makes sense to have struct-like enum members (outside of like error types).  This is such a clear-cut case of where it'd make sense to have a `CheckpointPublication` type, as evidenced by how you noticed it was helpful to make a helper constructor for it.

crates/btcio/src/broadcaster/processor.rs
```diff
@@ -325,6 +326,41 @@ where
...
+/// Recovers a transaction whose submission may have crossed a crash boundary.
+async fn probe_ambiguous_entry<C>(
+    io: &C,
+    idx: u64,
+    txentry: &L1TxEntry,
+    txid: &Txid,
+    params: &BtcioParams,
+) -> BroadcasterResult<L1TxStatus>
+where
+    C: BroadcasterIoContext,
```
delbonis, line 334: `io: &impl BroadcasterIoContext`

crates/btcio/src/broadcaster/processor.rs
```diff
@@ -396,21 +432,40 @@ where
...
-async fn publish_tx<C>(
+/// Applies policy and durably records that submission started before calling Bitcoin.
+async fn publish_queued<C>(
+    io: &C,
+    idx: u64,
+    txentry: &L1TxEntry,
+    params: &BtcioParams,
+) -> BroadcasterResult<L1TxStatus>
+where
+    C: BroadcasterIoContext,
```
delbonis, line 441: `io: &impl BroadcasterIoContext`

crates/db/types/src/l1_writer.rs
```diff
@@ -167,6 +167,21 @@ pub enum L1BundleStatus {
...
+impl L1BundleStatus {
+    /// Returns whether the bundle status records publication on L1.
+    pub fn has_reached_l1(&self) -> bool {
+        matches!(self, Self::Published | Self::Confirmed | Self::Finalized)
+    }
+
+    /// Returns whether the writer has no more work for this bundle.
+    pub fn is_terminal(&self) -> bool {
+        matches!(self, Self::Finalized | Self::Abandoned)
+    }
```
delbonis, line 184: `is_finished` or `is_final` probably a little more clear just because "terminal" already has a meaning in our protocol.

delbonis (changes requested):

bewakes:
Thanks @delbonis for the comments. I’ve addressed the cleanup suggestions.

For module ownership, I considered the `strata-ol-checkpoint` crate as another candidate and believe it is the better fit. It already owns checkpoint creation and recovery, so placing reconciliation and checkpoint-specific publication policy there keeps the lifecycle together. The tradeoff is some additional sequencer-gated dependency and feature wiring.

I kept `CheckpointPublishPolicy` concrete over `NodeStorage` intentionally. It relies on high-level canonical ASM-state resolution, which filters orphaned state after reorgs; generic raw `DatabaseBackend` access would not provide equivalent guarantees. The lower-level publication and reconciliation helpers remain generic over `DatabaseBackend`. We could introduce a dedicated storage capability trait, but it would currently have only one implementation and add indirection without improving reuse or testing.

delbonis:
> For module ownership, I considered the strata-ol-checkpoint crate as another candidate and believe it is the better fit. It already owns checkpoint creation and recovery, so placing reconciliation and checkpoint-specific publication policy there keeps the lifecycle together. 

I'm not sure that's the right choice since it expands the scope of that crate a *lot* with unrelated functionality to what it's currently doing.  This is evidenced by all of the new deps that had to be added to it.

> It relies on high-level canonical ASM-state resolution, which filters orphaned state after reorgs; generic raw DatabaseBackend access would not provide equivalent guarantees. The lower-level publication and reconciliation helpers remain generic over DatabaseBackend.

I was ***not*** suggesting to use `DatabaseBackend`, you'd never want to do this *because* of the exact reasons you describe.  I am saying to use a context trait like services (should) do to hide the storage and make it easier to test in isolation, and also simplify the dependency tree.

>We could introduce a dedicated storage capability trait, but it would currently have only one implementation and add indirection without improving reuse or testing.

The immediate value is that it lets you have mocks to test it more easily in isolation and also helps avoid complicating the dependency hierarchy.  Putting it in the strata binary crate is declaring dependency tree bankruptcy because we have the whole dependency tree in scope there, but it's something we *really don't want to do* because it provides no guard against worsening tech debt.

In an ideal world, *nothing* would directly use `NodeStorage` outside of init logic.  Everything should go through context trait impls.

bewakes:
@codex review

bewakes:
@delbonis I agree. And I think it makes sense to have them in sequencer crate - at least this gets rid of those annoying feature gatings.
I've also updated the plumbing to use context trait. PTAL.

delbonis:
Will review this asap

crates/ol/sequencer/src/checkpoint/publication.rs
delbonis, line 1: Still need to fix the vague and unclear fn names.

There's also just *a lot going on here*, is it possible to break apart the logical components?  Especially since there is shared code between this and `reconcile`.

crates/ol/sequencer/src/checkpoint/reconcile.rs
```diff
@@ -0,0 +1,434 @@
...
+/// may already have been terminalized.
+fn cancel_queued_checkpoint(
+    context: &impl CheckpointReconcileContext,
+    intents: &[IntentEntry],
+) -> Result<bool> {
+    let mut all_cancelled = true;
+    for intent in intents {
+        all_cancelled &= cancel_writer_intent(context, intent.clone())?;
+    }
+    Ok(all_cancelled)
```
delbonis, line 152: This can `return Ok(false);` early if any of them are not cancelled and just `Ok(true)` at the end.

crates/ol/sequencer/src/checkpoint/reconcile.rs
delbonis, line 1: Some misleading names here too.

crates/db/store-sled/src/broadcaster/db.rs
```diff
@@ -30,6 +30,22 @@ impl L1BroadcastDBSled {
             None => Ok(0),
         }
     }
+
+    fn pair_indices(&self, commit_txid: Buf32, reveal_txid: Buf32) -> DbResult<Option<(u64, u64)>> {
```
delbonis, line 34: What does this fn do?  This name suggests it's pairing indices (what does it mean to pair them?  idk), but it's taking txids.

crates/ol/sequencer/src/checkpoint/context.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// Returns the next broadcaster index.
+    fn get_next_broadcast_idx(&self) -> CheckpointContextResult<u64>;
+
+    /// Returns the broadcaster entry at an index.
+    fn get_broadcast_entry(&self, idx: u64) -> CheckpointContextResult<Option<L1TxEntry>>;
+
+    /// Returns the broadcaster entry for a transaction ID.
+    fn get_broadcast_entry_by_id(&self, txid: Buf32) -> CheckpointContextResult<Option<L1TxEntry>>;
+}
```
delbonis, line 46: Why are some of these async but some of them aren't?  Is this used by different services?  If so, these should be split into separate traits.

crates/ol/sequencer/src/checkpoint/context.rs
```diff
@@ -0,0 +1,108 @@
...
+use strata_identifiers::{Buf32, Epoch, EpochCommitment};
+
+/// Error boundary for checkpoint publication and reconciliation infrastructure.
+#[derive(Debug, thiserror::Error)]
+pub enum CheckpointContextError {
+    #[error("database operation failed: {0}")]
+    Database(#[from] DbError),
+
+    #[error(transparent)]
+    Other(#[from] anyhow::Error),
```
delbonis, line 16: This is never correct error handling behavior.  There was a real finding in an audit report caused by this pattern missing errors that got re-wrapped, obscuring the root cause.  I am going to revive #1528 to followup on this.

crates/ol/sequencer/src/checkpoint/context.rs
```diff
@@ -0,0 +1,108 @@
...
+use async_trait::async_trait;
+use strata_db_types::l1_broadcast::L1TxEntry;
+use strata_db_types::l1_writer::{BundledPayloadEntry, IntentEntry};
+use strata_db_types::DbError;
+use strata_identifiers::{Buf32, Epoch, EpochCommitment};
+
+/// Error boundary for checkpoint publication and reconciliation infrastructure.
+#[derive(Debug, thiserror::Error)]
+pub enum CheckpointContextError {
+    #[error("database operation failed: {0}")]
```
delbonis, line 12: `db: {0}`

crates/ol/sequencer/src/checkpoint/context.rs
```diff
@@ -0,0 +1,108 @@
...
+    Database(#[from] DbError),
+
+    #[error(transparent)]
+    Other(#[from] anyhow::Error),
+
+    #[error("canonical {state} state is unavailable")]
+    CanonicalStateUnavailable { state: &'static str },
```
delbonis, line 19: `kind` instead of `state` is more precise, but this should be a tuple-like variant since it improves the ergonomics and is unambiguous

crates/ol/sequencer/src/checkpoint/publication.rs
```diff
@@ -0,0 +1,622 @@
...
+        .map(|envelope| RevealKind::Checkpoint(Box::new(envelope.payload)))
+        .map_err(|_| ())
+}
+
+fn is_envelope_reveal(tx: &Transaction) -> bool {
+    tx.input
+        .first()
+        .and_then(|input| input.witness.taproot_leaf_script())
+        .is_some_and(|leaf| parse_envelope_payload(&leaf.script.into()).is_ok())
+}
```
delbonis, line 230: This is kinda a smelly pattern since we go through all the effort of parsing the envelope payload and then just discard the result since we only care about true/false.  This suggests that the structure of the code paths could be improved so that we don't need to duplicate this effort.

crates/ol/sequencer/src/checkpoint/publication.rs
```diff
@@ -0,0 +1,622 @@
...
+    let tag = match parser.try_parse_tx(tx) {
+        Ok(tag) => tag,
+        Err(_) if is_envelope_reveal(tx) => return Ok(RevealKind::Other),
+        Err(_) => return Err(()),
+    };
+    let checkpoint_tag = OL_STF_CHECKPOINT_TX_TAG.as_ref();
+    if tag.subproto_id() != checkpoint_tag.subproto_id()
+        || tag.tx_type() != checkpoint_tag.tx_type()
+        || tag.aux_data() != checkpoint_tag.aux_data()
+    {
```
delbonis, line 217: Can we just do `tag == OL_STF_CHEKCPOINT_TX_TAG`?

crates/ol/sequencer/src/checkpoint/publication.rs
```diff
@@ -0,0 +1,622 @@
+//! Checkpoint-aware policy evaluated at the final Bitcoin publication boundary.
+
+use std::sync::Arc;
+
+use bitcoin::Transaction;
+use strata_asm_checkpoint_types::CheckpointPayload;
+use strata_asm_common::TxInputRef;
+use strata_asm_proto_checkpoint_txs::{extract_checkpoint_from_envelope, OL_STF_CHECKPOINT_TX_TAG};
```
delbonis, line 8: Ahhhhh this should not be a constant, it should be inferred from params.

delbonis (changes requested):
This is better, but there's still a lot of cleanliness/organizational issues.

PR #2240, opened 2026-08-13
STR-4219: Versioning with `-v1` suffixed crates and `V1` suffixed types

delbonis:
## Description

This PR rearranges several OL related crates by appending `-v1` suffixes onto their names and `V1` suffixes onto their types.  It also splits apart OL chain types (blocks) from tx types so that we can talk about them independently.

It also adds `imports_granularity = "Module"` to the `crates/ol/` subtree since it already touches almost every file there.

Claude did all the tedious work, including updating downstream deps.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist


storopoli (approved):
ACK 39693dcb575d97f92d314ff4f9e109cdf9a7386e

Just need to fix CI

bewakes (commented):
A couple of comments/qqs:

  1. Do we intend to group the versioned crates in the future? I think we should do that as we add more versions, since it would keep the directory cleaner - something like `crates/ol/v1/{stf,state-types,chain-types,tx-types}`. This probably depends on whether these crates are expected to version together or independently.

  2. I just want to confirm that the state types intended to evolve will use the `StableContainer`, right?
  3. With the state types structured as they are now, it seems that even a small field addition to the top-level state could require copying all of the state types into a new versioned crate, which feels redundant. Maybe this is intentional to keep state versions in sync with STF versions, but I don’t think they necessarily need to advance together.

     If a later STF version does not change the state schema, I think it could simply continue using `OLStateV1` - or at most expose an alias such as type `OLStateV4 = OLStateV1`. If only the top-level schema changes, could the new state type reuse the unchanged V1 nested types instead of copying the whole state-types crate?

  I also think we could have a top-level, versionable `OLState` wrapper:

 ```rust
enum OLState {
      V1(OLStateV1),
      V2(OLStateV2),
  }
```

  Since state migrations only happen at checkpoint/epoch boundaries, the new checkpoint STF could accept `OLState::V1`, migrate it to `OLState::V2` at the boundary, and then execute the new STF over V2. This way, the overall transition can remain `OLState -> OLState`, and the migration stays within the proved state transition rather than requiring separate conversion glue in the prover guest and client.

delbonis:
@bewakes 

1. These versions are all different version sequences, so actually *should not* be in a share dir.  I *expect* the STF version to increment the most often, followed by blocks/txs, and then the state.  It would be very misleading to have crates depend on each other across version subdirs.  However, the [abstraction traits coming in STR-4256](https://alpenlabs.atlassian.net/browse/STR-4256) will decouple the versions so that there will be minimal if any dependencies exposed anyways.  [See first iteration.](https://github.com/alpenlabs/alpen/pull/2098)
2. Some of them yes, not across the board, no.
3. We won't be doing type aliases, that's silly.  The crates are broken apart so they *don't* all have to advance in sync.  The abstraction traits will be immensely helpful here.

(re: `OLState` wrapper)

This will exist in the database layer but we'll be taking care to avoid needing them in other contexts.  See [STR-4220](https://alpenlabs.atlassian.net/browse/STR-4220).

I have thought this through several PRs down the road.

[STR-4220]: https://alpenlabs.atlassian.net/browse/STR-4220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

storopoli (approved):
ACK d8647191d1bb61e0d26899a50b71d988eb912ff9

> `imports_granularity = "Module"`

That's nice

krsnapaudel (approved):
LGTM. Needs rebase.
Stale `provers/sp1/guest-checkpoint/Cargo.lock` has been flagged. Worth updating.

delbonis:
Did the monster rebase.  Most of it was mechanical fixes to import names/formatting.

Now we wait for CI, on a great day to be dependent on GitHub Actions.

bewakes (approved):

PR #2245, opened 2026-08-24
ci: make cargo-audit scan dependencies

storopoli:
## Description

`cargo audit --file audit.toml` treats `audit.toml` as the lockfile to scan. CI therefore scanned zero dependencies and returned success even when `Cargo.lock` contained known vulnerabilities.

This moves the cargo-audit configuration to `.cargo/audit.toml`, where cargo-audit loads it automatically, and removes the incorrect `--file` argument from both workflow steps.

The first real scan found 14 vulnerabilities. The lockfile now uses patched releases for the affected AWS-LC, crossbeam, h2, ruint, rustls/webpki, lru, memmap2, rand, and Bitcoin dependencies. It also updates SP1 to 6.4.0 and `bitcoind-async-client` to 0.10.10. The latter added optional broadcast settings, so existing calls now pass `None` to preserve their current behavior.

The blocking audit scans 997 dependencies and reports no vulnerabilities. The non-blocking warnings step still reports 14 allowed warnings from unmaintained crates and upstream-constrained `lru 0.12.5` and `scc 2.4.0` soundness advisories.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [x] Dependency Update

## Notes to Reviewers

The workflow and config move are in the first commit. The second commit contains the dependency updates and the small broadcast API adjustment required by `bitcoind-async-client` 0.10.10.

Validation:

- `cargo audit --no-fetch`
- `cargo check --workspace --all-targets --locked`
- `cargo test -p strata-btcio --locked` (232 passed)
- `just lint-check-ws`
- `just fmt-ws`

- [x] Yes
- [ ] No

References:

- [SECFIND-779](https://alpenlabs.atlassian.net/browse/SECFIND-779)
- [alpenlabs/zkaleido#178](https://github.com/alpenlabs/zkaleido/pull/178)

## Checklist

AI assistance: Codex traced the finding, prepared the code and dependency updates, ran the checks listed above, and drafted this PR description.

## Related Issues

- [SECFIND-779](https://alpenlabs.atlassian.net/browse/SECFIND-779)

[SECFIND-779]: https://alpenlabs.atlassian.net/browse/SECFIND-779?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


storopoli:
> **Low — audit gate can silently pass without scanning the prover lockfiles (fail-open)** `.github/workflows/security.yml:57-66`
> 
> The prover-lockfile loop reads from `< <(find provers ...)`. A process substitution's exit status is not propagated and is not caught by the default `bash -e` shell, so if `provers/` is ever renamed, moved, or emptied, `find` errors to stderr, the loop body never executes, and the blocking "Check for vulnerabilities" step still exits 0 — reintroducing exactly the silently-passing-scan bug this PR fixes. Same applies to the warnings step, where `audit_status` simply stays 0.
> 
> Fix: assert the scan actually happened, e.g. before the loop add `[[ -d provers ]] || { echo "provers/ missing"; exit 1; }`, and count scanned lockfiles, failing if the count is 0:
> 
> ```shell
> scanned=0
> while IFS= read -r -d '' manifest; do
>   lockfile="${manifest%Cargo.toml}Cargo.lock"
>   [[ -f "$lockfile" ]] || continue
>   scanned=$((scanned + 1))
>   cargo audit --file "$lockfile"
> done < <(find provers -type f -name Cargo.toml -print0)
> [[ "$scanned" -gt 0 ]] || { echo "no prover lockfiles scanned"; exit 1; }
> ```
> 
> Notes on the rest of the diff: the `audit.toml` → `.cargo/audit.toml` move plus dropping `--file` is the correct fix and the ignore list was not expanded to paper over the 14 findings. The lockfile changes are upgrades only — no security-relevant crate is downgraded (`rustls` 0.21/`rustls-webpki` 0.101.7/`webpki-roots` 0.25.4 are fully dropped rather than retained), and `send_raw_transaction(tx, None)` / `submit_package(txs, None)` omit the new broadcast options, preserving the previous `maxfeerate`/`maxburnamount` behavior. `lru 0.12.5` and `scc 2.4.0` remain but are `unsound`-class informational advisories, so they surface only in the non-blocking warnings step, as the description states.

Ok, that's fair. I'm going to hard-code the Cargo.toml prover paths that we have so far. Instead of doing the shell script finding at runtime

.github/workflows/security.yml
```diff
@@ -37,9 +37,16 @@ jobs:
 
         #  HACK: not denying warnings as we depend on `yaml-rust` via `format-serde-error` which is unmaintained
       - name: Check for audit warnings
-        run: cargo audit -D warnings --file audit.toml
+        run: cargo audit -D warnings
```
robin-alpen, line 40: nit: IIUC the only ignore you currently have in audit.toml is RUSTSEC-2025-0055, but tracing-subscriber is now at 0.3.23 so that ignore should no longer be needed?

robin-alpen (approved):
Thanks for addressing this so quickly @storopoli ! Really nice that everything could be bumped! Looks great. Just a minor comment, see above.

.github/workflows/security.yml
storopoli: Nice catch let me try removing it in a follow-up commit and see if CI passes

storopoli: Ok done in f69b442a8943ac5034fc0764cddedf9cbe332ecd

Merging this

PR #2249, opened 2026-08-26
refactor(ol): make WriteBatch not generic (STR-4254)

storopoli:
## Description

Drops the account type parameter from `WriteBatch`/`LedgerWriteBatch`. The only type ever used was `OLAccountStateV1`, and the generic kept the batch defined in terms of a state-accessor type instead of a plain data structure we can persist on its own. The ledger batch now stores a dedicated `AccountStateWrite` type (a full-state replacement; not a DA type), and `IAccountState` gains an associated `Write` type plus a fallible `apply_write` so states apply writes to themselves.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Three commits, each compiles and passes tests on its own — easiest reviewed commit-by-commit:

1. additive: `AccountStateWrite`, `IAccountState::{Write, apply_write}`, impls, tests
2. pins every `WriteBatch<S::AccountState>` signature to the concrete type (shows the generic carried no weight)
3. drops the parameter and deletes the type args workspace-wide

Worth knowing:

- `LedgerWriteBatch` accessors still hand out `OLAccountStateV1`, so layers, RPC, and existing tests are untouched.
- `check_write_batch_safe` now preflights update serials against stored accounts (stricter than the old serial-threshold heuristic), keeping `apply_write_batch`'s "error leaves state unmodified" contract; regression test included.
- On-disk batch encoding is byte-identical, pinned by a byte-equivalence test and a non-empty DB round-trip test.
- `IndexerAccountStateMut` sets `Write = Infallible`: that wrapper can't describe a full-state replacement with its incremental index writes, so the call is unconstructible.

- [ ]  Yes
- [x]  No

## Checklist

## Related Issues

[STR-4254]: https://alpenlabs.atlassian.net/browse/STR-4254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/ol/block-assembly/src/block_assembly.rs
+++ b/crates/ol/block-assembly/src/block_assembly.rs
@@ -12,7 +12,7 @@ use strata_ol_chain_types_v1::*;
 use strata_ol_mempool::MempoolTxInvalidReason;
 use strata_ol_state_support_types::{DaAccumulatingState, WriteTrackingState};
 use strata_ol_state_types::{AccProofCheck, ISnarkAccountState, IStateAccessor, TxProofIndexer, *};
-use strata_ol_state_types_v1::{MAX_PENDING_ASM_LOGS, OLAccountStateV1, WriteBatch};
+use strata_ol_state_types_v1::{MAX_PENDING_ASM_LOGS, WriteBatch};
 use strata_ol_stf_v1::*;
 use strata_ol_tx_types_v1::*;
 use strata_snark_acct_types as _;
@@ -40,7 +40,7 @@ struct ProcessTransactionsOutput {
     /// Transactions that failed during block assembly.
     failed_txs: Vec<FailedMempoolTx>,
     /// Accumulated write batch after processing all transactions.
-    accumulated_batch: WriteBatch<OLAccountStateV1>,
+    accumulated_batch: WriteBatch,
     /// Accumulated da data after processing all transactions.
     accumulated_da: AccumulatedDaData,
     /// Non-cadence limits requesting an epoch seal, if any.
@@ -49,7 +49,7 @@ struct ProcessTransactionsOutput {
 
 /// Inputs consumed while finalizing a block template.
 struct BuildBlockTemplateInput {
-    accumulated_batch: WriteBatch<OLAccountStateV1>,
+    accumulated_batch: WriteBatch,
     output_buffer: ExecOutputBuffer,
     successful_txs: Vec<OLTransactionV1>,
     is_terminal: bool,
@@ -556,7 +556,7 @@ fn execute_block_initialization<S: BlockAssemblyStateAccess>(
     parent_state: &S,
     block_context: &BlockContext<'_>,
     accumulated_da: AccumulatedDaData,
-) -> (WriteBatch<OLAccountStateV1>, AccumulatedDaData) {
+) -> (WriteBatch, AccumulatedDaData) {
     let (accumulator, logs) = accumulated_da.into_parts();
     let write_state = WriteTrackingState::new_empty(parent_state);
     let mut da_state = DaAccumulatingState::new_with_accumulator(write_state, accumulator);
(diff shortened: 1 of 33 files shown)


crates/ol/block-assembly/src/block_assembly.rs
```diff
@@ -1018,7 +1010,7 @@ mod tests {
     use crate::test_utils::*;
     use crate::{FixedSlotSealing, LimitAwareSealing};
 
-    type OlWriteBatch = WriteBatch<<MemoryStateBaseLayer as IStateAccessor>::AccountState>;
+    type OlWriteBatch = WriteBatch;
```
krsnapaudel: Not introduced here, but we could make this `OLWriteBatch` for consistency.

krsnapaudel (approved):
A small nit. LGTM otherwise.

crates/ol/state-types-v1/src/write_batch.rs
```diff
@@ -2,14 +2,48 @@
...
+///
+/// Currently the only supported write is a full replacement of the account
+/// state; finer-grained writes can be added later. This is NOT a DA type, so
+/// DA considerations do not apply.
+#[derive(Clone, Debug)]
+pub struct AccountStateWrite(OLAccountStateV1);
+
+impl AccountStateWrite {
+    /// Creates an account state write.
+    pub fn new(state: OLAccountStateV1) -> Self {
```
bewakes, line 23: nit: Perhaps `new_sate(state)`? Just in case this might evolve into an enum when we support finer writes.

crates/ol/state-types-v1/src/write_batch.rs
```diff
@@ -320,7 +358,18 @@ impl Codec for EpochalStateWrites {
     }
 }
 
-impl<A: Encode + Decode + Clone> Codec for WriteBatch<A> {
+impl Codec for AccountStateWrite {
```
bewakes, line 361: Can we not just derive this?

crates/ol/state-types/src/errors.rs
```diff
@@ -66,6 +66,14 @@ pub enum StateError {
         in_table: AccountSerial,
     },
 
+    #[error(
+        "write not applicable to account state (state serial {in_state}, write serial {in_write})"
+    )]
+    InapplicableAcctWrite {
```
bewakes, line 72: Won't these two serials be the same? I am not sure what exactly this error represents.

bewakes (commented):
Nice cleanup! Some non-blocking comments only.

crates/ol/block-assembly/src/block_assembly.rs
storopoli: Good suggestion. Fixed and amended into the commits.

crates/ol/state-types-v1/src/write_batch.rs
storopoli: I'd hold off on renaming. If this grows finer-grained writes it becomes an enum, and at that point per-variant constructors (`replace(state)`, `set_balance(..)`, etc.) make more sense than a single `new_*`. Today the type is just a full-state replacement, so `new` says all there is. The rename also doesn't buy us anything forward-looking, since the enum change would rework the constructors anyway.

crates/ol/state-types/src/errors.rs
storopoli: They're equal exactly when the write is valid; the error exists for when they aren't. The serial is the account's identity, and a full-state replacement must carry the same serial as the account it's applied to. If it doesn't, the write was built for a different account or the batch is corrupt, and `check_write_batch_safe` rejects the whole batch before anything mutates. This is the "erroring if the write doesn't make sense" part of the ticket.

crates/ol/state-types-v1/src/write_batch.rs
storopoli: Not with the current derive. It requires every field to implement `Codec`, and `OLAccountStateV1` is SSZ-generated with no `Codec` impl, so the derive doesn't compile regardless of what we want the bytes to be. The alternatives are storing `CodecSsz<OLAccountStateV1>` in the field (then every accessor unwraps the shim) or hand-writing `Codec` for `OLAccountStateV1` itself (same delegation, moved down a level, and it gives an SSZ consensus type a second serialization trait). The manual impl seemed like the smallest option. It also happens to keep the bytes identical to what we stored before, which the round-trip test pins.

bewakes (approved):

PR #2250, opened 2026-08-27
fix(alpen-cli)!: derive the signet wallet via standard BIP39, not raw entropy

krsnapaudel:
## Description

- Derive the L1 signet wallet via standard BIP39 (entropy → mnemonic → PBKDF2 seed → BIP32), matching the L2 EVM path instead of feeding raw entropy directly into BIP32. A mnemonic from `alpen backup` is now recoverable in any standard BIP39 wallet (Sparrow, Electrum, hardware wallets), verified against the BIP-86 test vector.
- Remove backup args. Print the mnemonic in English.

### Breaking change
The fix changes key derivation for any seed generated before this PR. Existing local wallets will fail to load (`InvalidChangeSet(Mismatch(...))` from BDK, confirmed by testing) since the on-disk wallet cache no longer matches what's freshly derived from the same seed. No automatic migration is included — affected users need `alpen reset` and to recreate their wallet.

This PR was created with help from Claude.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Is this PR addressing any specification, design doc or external reference document?

- [ ]  Yes
- [ ]  No

If yes, please add relevant links:

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have commented my code where necessary.
- [ ] I have updated the documentation if needed.
- [ ] My changes do not introduce new warnings.
- [ ] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [ ] New and existing tests pass with my changes.
- [ ] I have [disclosed my use of AI](https://github.com/alpenlabs/alpen/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

## Related Issues

[SECFIND-773](https://alpenlabs.atlassian.net/browse/SECFIND-773)

[SECFIND-773]: https://alpenlabs.atlassian.net/browse/SECFIND-773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2251, opened 2026-08-27
fix(alpen-cli): derive deposit-request reclaim keys from the seed

krsnapaudel:
## Description

- Derive each deposit request's reclaim keypair deterministically from the seed (`m/<DRT_RECLAIM_PURPOSE>'/<counter>'`) instead of a fresh `OsRng` keypair, so the descriptor DB (`descriptor_db`) stops being the *only* copy of a live deposit's spending key — the mnemonic alone is now sufficient.
- `counter` is durable local state, tracked in its own sled tree (`DescriptorRecovery::next_reclaim_counter`) so it survives restarts and is never reused for a live deposit.
- `alpen recover` now automatically scans for deposits reconstructible from the seed (`counter = 0, 1, 2, ...`), checking each candidate address for *any* on-chain transaction history, and stops after `SEED_RECOVERY_GAP_LIMIT` (BIP44 gap-limit style). This is in addition to what is in the descriptor DB.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

- [ ]  Yes
- [ ]  No

## Checklist

## Related Issues

[SECFIND-772](https://alpenlabs.atlassian.net/browse/SECFIND-772)

[SECFIND-772]: https://alpenlabs.atlassian.net/browse/SECFIND-772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/alpen-cli/src/cmd/deposit.rs
+++ b/bin/alpen-cli/src/cmd/deposit.rs
@@ -16,7 +16,6 @@ use bdk_wallet::{
 };
 use colored::Colorize;
 use indicatif::ProgressBar;
-use rand_core::OsRng;
 use shrex::encode;
 use strata_asm_proto_bridge_v1_txs::deposit_request::DrtHeaderAux;
 use strata_cli_common::errors::{DisplayableError, DisplayedError};
@@ -27,10 +26,11 @@ use strata_primitives::crypto::even_kp;
 
 use crate::{
     alpen::AlpenWallet,
+    cmd::recover::reconstruct_reclaim_counter,
     constants::{ALPEN_EE_ACCT_SERIAL, SIGNET_BLOCK_TIME},
     link::{OnchainObject, PrettyPrint},
     recovery::DescriptorRecovery,
-    seed::Seed,
+    seed::{DrtReclaimKeypair, Seed},
     settings::Settings,
     signet::{get_fee_rate, log_fee_rate, SignetWallet},
 };
@@ -93,8 +93,10 @@ fn prepare_deposit_request(
     recover_delay: u16,
     alpen_address: AlpenAddress,
     bridge_in_amount: Amount,
+    reclaim_keypair: DrtReclaimKeypair,
 ) -> (DescriptorTemplateOut, BitcoinAddress, DrtHeaderAux, TxOut) {
-    let (secret_key, recovery_public_key) = even_kp(SECP256K1.generate_keypair(&mut OsRng));
+    let (secret_key, recovery_public_key) =
+        even_kp((reclaim_keypair.secret_key, reclaim_keypair.public_key));
     let recovery_public_key = recovery_public_key.x_only_public_key().0;
     let recovery_private_key = PrivateKey::new(secret_key.into(), network);
 
@@ -166,12 +168,34 @@ pub async fn deposit(
         alpen_address.to_string().cyan(),
     );
 
+    let mut desc_file = DescriptorRecovery::open(&seed, &settings.descriptor_db)
+        .await
+        .internal_error("Failed to open descriptor recovery file")?;
+    if desc_file
+        .current_reclaim_counter()
+        .internal_error("Failed to read the deposit reclaim counter")?
+        .is_none()
+    {
+        println!("Reconstructing the missing deposit reclaim counter...");
+        let reconstructed_counter = reconstruct_reclaim_counter(&seed, &settings).await?;
+        desc_file
+            .ensure_reclaim_counter_at_least(reconstructed_counter)
+            .await
+            .internal_error("Failed to save the reconstructed deposit reclaim counter")?;
+    }
+    let reclaim_counter = desc_file
+        .next_reclaim_counter()
+        .await
+        .internal_error("Failed to reserve a deposit reclaim key counter")?;
+    let reclaim_keypair = seed.drt_reclaim_keypair(reclaim_counter);
+
     let (bridge_in_desc, bridge_in_address, header_aux, deposit_output) = prepare_deposit_request(
         settings.bridge_musig2_pubkey,
         settings.network,
         settings.recovery_delay,
         alpen_address,
         drt_amount,
+        reclaim_keypair,
     );
 
     println!(
@@ -185,10 +209,11 @@ pub async fn deposit(
         .expect("valid chain tip")
         .height;
 
-    // Number of blocks after which the wallet actually enables recovery. This is mostly to account
-    // for any reorgs that may happen at the recovery height.
-    let recover_at =
-        current_block_height + settings.recovery_delay as u32 + settings.finality_depth;
+    let recover_at = compute_recover_at_height(
+        current_block_height,
+        settings.recovery_delay as u32,
+        settings.finality_depth,
+    );
 
     println!(
         "Using {} as bridge in address",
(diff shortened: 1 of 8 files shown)


bin/alpen-cli/src/cmd/recover.rs
```diff
@@ -98,64 +103,150 @@ pub async fn recover(
...
+        let mut recovery_wallet = Wallet::create_single(wallet_desc)
+            .network(settings.network)
+            .create_wallet_no_persist()
+            .internal_error("Failed to create recovery wallet")?;
+
+        let address = recovery_wallet.reveal_next_address(KeychainKind::External);
+        sync_wallet(&mut recovery_wallet, settings.signet_backend.clone())
             .await
-            .internal_error("Failed to broadcast signet transaction")?;
+            .internal_error("Failed to sync recovery wallet")?;
```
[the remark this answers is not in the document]
krsnapaudel: This is preexisting.

bin/alpen-cli/src/cmd/deposit.rs
```diff
@@ -93,8 +93,9 @@ fn prepare_deposit_request(
     recover_delay: u16,
     alpen_address: AlpenAddress,
     bridge_in_amount: Amount,
+    reclaim_keypair: (SecretKey, PublicKey),
```
storopoli: I think we should create a proper struct for this `reclaim_keypair` with the SK and PK and use `Zeroize`/`ZeroizeOnDrop` on it.

bin/alpen-cli/src/seed.rs
```diff
@@ -72,6 +73,27 @@ impl Seed {
...
     }
 
+    /// Derives the deposit-request reclaim keypair for the given counter.
+    ///
+    /// This is deterministic in the seed and `counter`, so a deposit's reclaim key can always be
+    /// recovered from the mnemonic alone, given the counter value it was created with. `counter`
+    /// must never be reused for a live deposit request: it is intended to come from
+    /// [`DescriptorRecovery::next_reclaim_counter`](crate::recovery::DescriptorRecovery::next_reclaim_counter),
+    /// which persists the increment before handing out a value.
+    pub fn drt_reclaim_keypair(&self, counter: u32) -> (SecretKey, PublicKey) {
```
storopoli: I think we should create a proper struct for this `reclaim_keypair` with the SK and PK and use `Zeroize`/`ZeroizeOnDrop` on it.

bin/alpen-cli/src/constants.rs
```diff
@@ -39,6 +45,20 @@ pub const SIGNET_BLOCK_TIME: Duration = Duration::from_secs(10 * 60); // 10 minu
...
+/// collide with anything registered.
+///
+/// Separates this key material from the wallet's BIP-86 path (`m/86'/0'/0'`). Each deposit derives
+/// `m/<DRT_RECLAIM_PURPOSE>'/<counter>'`, where `counter` is durable local state (see
+/// [`DescriptorRecovery::next_reclaim_counter`](crate::recovery::DescriptorRecovery::next_reclaim_counter)),
+/// making the reclaim key recoverable from the seed alone rather than only from the descriptor DB.
+///
+/// Don't change this. A deposit's reclaim key can only be reconstructed from the seed if this
+/// value is still the same as when that deposit was made.
+pub const DRT_RECLAIM_PURPOSE: ChildNumber = ChildNumber::Hardened { index: 43_000 };
```
storopoli, line 59: Is this arbitrary? Just curios why 43_000?

storopoli (changes requested):

bin/alpen-cli/src/constants.rs
krsnapaudel: @storopoli Yes it is arbitrary outside the reserved range.

bin/alpen-cli/src/constants.rs
```diff
@@ -10,6 +10,12 @@ pub const DEFAULT_FINALITY_DEPTH: u32 = 6;
 
 pub const RECOVERY_DESC_CLEANUP_DELAY: u32 = 100;
 
+/// Number of consecutive unused reclaim-key counters `recover --from-seed` tries before giving
+/// up, matching the conventional BIP44 address-gap-limit: there's no persisted "last used
+/// counter" to resume from when reconstructing purely from the seed, so this is the only signal
+/// for when to stop scanning.
+pub const SEED_RECOVERY_GAP_LIMIT: u32 = 20;
```
storopoli: Please increase this gap limit to `1_000`. Counters are reserved before transaction construction and broadcast, so failed deposit attempts create unused gaps. With a limit of 20, only 19 failed attempts can make mnemonic-only recovery stop before a later funded counter; 1,000 provides a substantially safer recovery bound.

This is quite cheap and can potentially help users recover funds in the case the bridge is offline and the user (who is always inclined to do stupid things) keeps trying to deposit non-stop and will probably accrue more than 19-20 failed attempts.

storopoli (changes requested):
minor blocking fix please

bin/alpen-cli/src/constants.rs
storopoli: Some additional context: 1,000 stop gap is the default in bitcoin core wallet.

storopoli: Done in [3001f74](https://github.com/alpenlabs/alpen/pull/2251/commits/3001f74f4be0b251ca328c3309389d820dc51d78)

storopoli (dismissed):
ACK 3001f74f4be0b251ca328c3309389d820dc51d78

storopoli:
pinging @Zk2u

Zk2u (changes requested):
**fix**

In one `alpen recover` run: the descriptor loop drains the DRT for counter
  k and broadcasts the spend (unconfirmed). The seed scan then re-derives k;
  if the backend has not indexed the mempool spend by the time that
  candidate syncs, the fresh wallet sees the DRT as confirmed-unspent,
  matured, `balance().confirmed > 0`, and broadcasts a second conflicting
  spend (to a different l1w external address — `reveal_next_address`
  advances).
- Verified the guard holds in the common case: bdk_chain excludes any
  output with `spent_by` set (even an unconfirmed spend) from
  `list_unspent` and from the balance, so a visible spend → confirmed
  balance 0 → skip. The window is the indexer/mempool lag between the
  broadcast and the candidate's sync.
- If it fires: both spends go to the user's own l1w, no loss; but the second
  broadcast fails (`bad-txns-inputs-missingorspent`) and `recover` exits
  with a confusing error mid-scan.
- Cheap fix: record the DRT script_pubkeys (or revealed addresses) drained by
  the descriptor loop in a set; skip seed-scan candidates whose revealed
  address is in that set. DB entries do not store the counter, so the
  address is the reliable join key.
- Classification: fix-now (low severity, funds path, small fix).

---

**minors/nits**

- The `DRT_RECLAIM_PURPOSE` docblock says "Don't change this" — good. But
  `drt_reclaim_keypair` does not state that the master is the raw-entropy
  BIP32 key, deliberately kept separate from the BIP39 PBKDF2 master #2250
  introduces. Someone who "cleans up" to reuse the signet-wallet master
  would silently break reconstruction of existing deposits. 
  
- `even_kp` and callers copy the `SecretKey` (a Copy type) before the
  `DrtReclaimKeypair` drops, so zeroize-on-drop clears one copy, not all.
  Defense in depth at best. Either soften the struct doc ("zeroizes its own
  copy") or drop the ceremony. Potentially not worth changing in this PR

storopoli:
@codex review

storopoli:
@Zk2u can you check. I've fixed your review comments.
Cc @barakshani

Zk2u (approved):

PR #2266, opened 2026-09-01
test(snark-acct): reject valid stale-state forks

storopoli:
## Description

This adds regression coverage for SECFIND-854.

The OL test starts at state s1, verifies and applies s1 to s2 and s2 to s3, then builds an alternative s2 to s4 proof using the sequence number expected at s3. The proof is valid against the stale s2 state. When it is submitted against current s3, the sequence check passes but proof verification fails because OL rebuilds the claim with the s3 root.

The runtime test checks the other side of the invariant: private pre-state data must match the public pre-state root.

These tests show that the reported fork does not work against current main. They do not establish whether the audited revision behaved differently.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The OL verifier is a focused mock that accepts the bytes of one expected claim. It lets the test check the claim that OL constructs without depending on a concrete proof backend.

Local checks:

- cargo test -p strata-snark-acct-runtime -p strata-snark-acct-sys --locked
- cargo fmt --all --check
- just lint-check-ws

- [x] Yes
- [ ] No

- https://alpenlabs.atlassian.net/browse/SECFIND-854

## Checklist

AI assistance: Codex wrote the tests and this PR description. The generated code was reviewed, and the checks listed above were run locally.

## Related Issues

- https://alpenlabs.atlassian.net/browse/SECFIND-854

### DIFF
--- a/crates/snark-acct-sys/src/verification.rs
+++ b/crates/snark-acct-sys/src/verification.rs
@@ -179,7 +179,7 @@ fn compute_update_claim(
 #[cfg(test)]
 mod tests {
     use strata_acct_types::{AccumulatorClaim, Hash, TxEffects};
-    use strata_ol_state_types::{ExecError, ProofVerifyError};
+    use strata_ol_state_types::{ExecError, ISnarkAccountStateMut, ProofVerifyError};
     use strata_ol_state_types_v1::OLSnarkAccountStateV1;
     use strata_predicate::PredicateKey;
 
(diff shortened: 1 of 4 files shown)


barakshani (commented):
Thanks for looking into this so quickly!

Here's what I had in mind for this test:
- start at state s_1
- move to state s_2, verifiably: s_1 --> s_2
- move to state s_3, verifiably: s_2 --> s_3
- attempt a new _valid_ state transition from s_2 --> s_4

I believe this case is what the report was about (we fork a valid chain).

Possibly we could do s_1 --> s_3 at the third state (without the fourth), but I'm not sure about the full details under the hood.

storopoli:
> Thanks for looking into this so quickly!
> 
> Here's what I had in mind for this test:
> 
> * start at state s_1
> * move to state s_2, verifiably: s_1 --> s_2
> * move to state s_3, verifiably: s_2 --> s_3
> * attempt a new _valid_ state transition from s_2 --> s_4
> 
> I believe this case is what the report was about (we fork a valid chain).
> 
> Possibly we could do s_1 --> s_3 at the third state (without the fourth), but I'm not sure about the full details under the hood.

Check the new commit bc162856eb111c7602215f969986dfc3fbee7917

crates/snark-acct-sys/src/verification.rs
```diff
@@ -175,3 +175,132 @@ fn compute_update_claim(
...
+            accepted_claim: state_2_claim.clone(),
+        };
+        let result =
+            verify_update_correctness(target, &current_state, &update_2_to_4, &mut verifier);
+        assert!(matches!(
+            result,
+            Err(ExecError::Acct(AcctError::InvalidUpdateSequence {
+                account_id,
+                expected: 2,
+                got: 1,
```
barakshani: Looks like we fail on this check: https://github.com/alpenlabs/alpen/blob/bc162856eb111c7602215f969986dfc3fbee7917/crates/snark-acct-sys/src/verification.rs#L18

This is good, but not sufficient. we can still fork from seq_no 1, but set the next seq_no to be (the expected) 3.

storopoli: Good catch. The previous version stopped at the sequence-number check, so it missed this interesting case.

I changed the forked update to use the sequence number expected at `s3` while keeping the proof bound to `s2`. The sequence check now passes, and full verification rejects it with `InvalidUpdateProof` because OL rebuilds the claim using the current `s3` root.

Fixed in 5cdc4db29.

barakshani (approved):
Ack https://github.com/alpenlabs/alpen/pull/2266/changes/5cdc4db29e4f4f87aaced15072d9b2253e79d4be.

Thanks, LGTM. Please check with the team whether the fact that seq_no {x} to seq_no {x+2} passes verification, in this case. I think it's fine (partly this is what the new tests show), but I prefer not to overlook this, so better bringing this up with the team.

storopoli:
> Thanks, LGTM. Please check with the team whether the fact that seq_no {x} to seq_no {x+2} passes verification, in this case. I think it's fine (partly this is what the new tests show), but I prefer not to overlook this, so better bringing this up with the team.

That is my opinion as well.
Cc @delbonis for your opinion and a review beg for CODEOWNERS

bewakes (approved):

PR #2273, opened 2026-09-04
fix(alpen-client): bound concurrent EE proof submissions

irnb:
Share a configurable submission budget between the chunk and account provers, defaulting to one concurrent preparation and submission.

Keep capacity-limited tasks pending for tick-based dispatch and restart recovery, release capacity after remote proof submission, and defer account proving until every chunk receipt is available.

Log permanently failed chunks that block account proof generation.

### DIFF
--- a/crates/prover-core/src/prover.rs
+++ b/crates/prover-core/src/prover.rs
@@ -3,7 +3,9 @@
 
 use std::{
     collections::HashMap,
-    fmt, slice,
+    fmt,
+    num::NonZeroUsize,
+    slice,
     sync::{
         atomic::{AtomicBool, Ordering},
         Arc,
@@ -12,7 +14,10 @@ use std::{
 };
 
 use parking_lot::Mutex;
-use tokio::{sync::oneshot, task::spawn_blocking};
+use tokio::{
+    sync::{oneshot, OwnedSemaphorePermit, Semaphore, TryAcquireError},
+    task::spawn_blocking,
+};
 use tracing::{error, info, info_span, warn, Instrument};
 use zkaleido::ZkVmHost;
 #[cfg(feature = "remote")]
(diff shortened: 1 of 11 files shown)


prajwolrg (dismissed):
LGTM!

sapinb (dismissed):
ACK bae7f20d36ed475800001c62dd5196a758b45413

LGTM

bewakes (approved):
Great! we needed something like this for quite some time now.

prajwolrg (approved):

crates/prover-core/src/prover.rs
```diff
@@ -44,6 +67,8 @@ pub struct Prover<H: ProofSpec> {
     task_store: Arc<dyn TaskStore>,
     receipt_store: Option<Arc<dyn ReceiptStore>>,
     receipt_hook: Option<Arc<dyn ReceiptHook<H>>>,
+    concurrency_budget: Option<ConcurrencyBudget>,
+    dispatch_lock: Mutex<()>,
```
delbonis, line 71: Don't love this pattern.  Can we put the lock on the thing that's used to do the dispatching?

Edit: I looked into how this is used and it doesn't seem like there really *is* just *one thing* that does the dispatching.  I'm wondering if this suggests we should break apart this `Prover` type into two layers, one part that does the low-level job handling and another part that handles orchestration and tracking.

delbonis (changes requested):

crates/prover-core/src/prover.rs
barakshani: @delbonis this isn't merge to main, just a hotfix for the release. main is planned a slightly different solution I believe.

storopoli (approved):
ACK bc3b774376a269e92930958a90cbeabc159fd4fa

barakshani:
merging this

PR #2277, opened 2026-09-08
test(consensus): [7/7] cover manifest authentication through persistence

storopoli:
## Description

Add signed-block regressions through the production FCM adapter, chain worker, and sled managers. Cover manifest persistence boundaries, delayed ASM output, restart recovery, and shallow L1 branch changes. Add checkpoint statement tests binding the manifest range.

L1/ASM outputs are seeded fixtures; these tests do not run Bitcoin RPC or ASM execution.

### Type of Change

- [x] New or updated tests

## Notes to Reviewers

Top of the stack. This diff contains the persistence regression suite and checkpoint tests. Earlier review discussions remain on this PR.

Part 7 of 7 for STR-4286. The former four-PR stack is split further for review. The combined code and regression tests are unchanged.

Merge order:

1. #2298: durable status pagination and retry ticks.
2. #2302: deferred execution and bounded recovery.
3. #2303: retry scheduling and durable discovery tests.
4. #2304: local failure and fork-choice recovery tests.
5. #2299: canonical manifest authentication.
6. #2300: restored unfinalized block authentication.
7. #2277: manifest persistence and checkpoint tests.

## Checklist

- [x] Self-reviewed the split and dependency boundaries.
- [x] Verified the repartitioned stack matches the reviewed fixes.
- [x] Formatting, whitespace, and locked workspace metadata checks passed locally.
- [x] GitHub CI passes for the rewritten branch. Full local tests were skipped as requested.
- [x] AI assistance: Codex prepared the implementation, tests, stack split, and PR text.

## Related Issues

[STR-4286](https://alpenlabs.atlassian.net/browse/STR-4286)

[STR-4286]: https://alpenlabs.atlassian.net/browse/STR-4286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/strata/src/fcm.rs
+++ b/bin/strata/src/fcm.rs
@@ -119,6 +119,17 @@ impl UnfinalizedOLBlockSource for StrataFcmContext {
 
 #[async_trait]
 impl FcmStorage for StrataFcmContext {
+    async fn scan_block_statuses(
+        &self,
+        after: Option<OLBlockId>,
+        limit: usize,
+    ) -> DbResult<Vec<(OLBlockId, BlockStatus)>> {
+        self.storage
+            .ol_block()
+            .scan_block_statuses_async(after, limit)
+            .await
+    }
+
     async fn set_block_status(&self, blkid: OLBlockId, status: BlockStatus) -> DbResult<bool> {
         self.storage
             .ol_block()
--- a/bin/strata/src/fcm.rs
+++ b/bin/strata/src/fcm.rs
@@ -66,6 +66,16 @@ impl ChainController for StrataFcmContext {
             Err(WorkerError::MissingPreState(_) | WorkerError::MissingOLBlock(_)) => Ok(
                 BlockExecutionOutcome::Deferred(ExecutionDeferral::Dependency),
             ),
+            Err(WorkerError::ManifestPending { height, reason }) => {
+                warn!(%block, height, ?reason, "deferring unauthenticated ASM manifest");
+                Ok(BlockExecutionOutcome::Deferred(
+                    ExecutionDeferral::Dependency,
+                ))
+            }
+            Err(WorkerError::ManifestStorage(err)) => {
+                warn!(%block, %err, "canonical manifest storage unavailable");
+                Ok(BlockExecutionOutcome::Deferred(ExecutionDeferral::Storage))
+            }
             Err(WorkerError::StfExecution(err)) => {
                 warn!(%block, %err, "rejecting invalid block execution");
                 Ok(BlockExecutionOutcome::Rejected)
--- a/bin/strata/src/fcm.rs
+++ b/bin/strata/src/fcm.rs
@@ -6,8 +6,8 @@ use anyhow::{Result, anyhow};
 use async_trait::async_trait;
 use strata_chain_worker::{ChainWorkerHandle, WorkerError};
 use strata_consensus_logic::{
-    BlockExecutionOutcome, ChainController, CsmStatusReader, ExecutionDeferral, FcmContext,
-    FcmServiceHandle, FcmStartupReconciler, FcmStorage,
+    BlockExecutionOutcome, BlockValidationOutcome, ChainController, CsmStatusReader,
+    ExecutionDeferral, FcmContext, FcmServiceHandle, FcmStartupReconciler, FcmStorage,
     ol_mmr_reconcile::{
         OLMmrReconcileResult, OLMmrReconcileTarget, reconcile_ol_mmr_index_to_target,
     },
@@ -37,6 +37,37 @@ struct StrataFcmContext {
     status_channel: Arc<StatusChannel>,
 }
 
+enum WorkerFailureOutcome {
+    Deferred(ExecutionDeferral),
+    Rejected(anyhow::Error),
+}
+
+fn classify_worker_failure(
+    block: OLBlockCommitment,
+    error: WorkerError,
+) -> anyhow::Result<WorkerFailureOutcome> {
+    match error {
+        WorkerError::MissingPreState(_) | WorkerError::MissingOLBlock(_) => Ok(
+            WorkerFailureOutcome::Deferred(ExecutionDeferral::Dependency),
+        ),
+        WorkerError::ManifestPending { height, reason } => {
+            warn!(%block, height, ?reason, "deferring unauthenticated ASM manifest");
+            Ok(WorkerFailureOutcome::Deferred(
+                ExecutionDeferral::Dependency,
+            ))
+        }
+        WorkerError::ManifestStorage(err) => {
+            warn!(%block, %err, "canonical manifest storage unavailable");
+            Ok(WorkerFailureOutcome::Deferred(ExecutionDeferral::Storage))
+        }
+        error @ WorkerError::StfExecution(_) => {
+            warn!(%block, %error, "rejecting invalid block inputs");
+            Ok(WorkerFailureOutcome::Rejected(error.into()))
+        }
+        error => Err(error.into()),
+    }
+}
+
 impl StrataFcmContext {
     fn new(
         storage: Arc<NodeStorage>,
@@ -63,24 +94,29 @@ impl ChainController for StrataFcmContext {
     ) -> anyhow::Result<BlockExecutionOutcome> {
         match self.chain_worker.try_exec_block(block).await {
             Ok(()) => Ok(BlockExecutionOutcome::Accepted),
-            Err(WorkerError::MissingPreState(_) | WorkerError::MissingOLBlock(_)) => Ok(
-                BlockExecutionOutcome::Deferred(ExecutionDeferral::Dependency),
-            ),
-            Err(WorkerError::ManifestPending { height, reason }) => {
-                warn!(%block, height, ?reason, "deferring unauthenticated ASM manifest");
-                Ok(BlockExecutionOutcome::Deferred(
-                    ExecutionDeferral::Dependency,
-                ))
-            }
-            Err(WorkerError::ManifestStorage(err)) => {
-                warn!(%block, %err, "canonical manifest storage unavailable");
-                Ok(BlockExecutionOutcome::Deferred(ExecutionDeferral::Storage))
-            }
-            Err(WorkerError::StfExecution(err)) => {
-                warn!(%block, %err, "rejecting invalid block execution");
-                Ok(BlockExecutionOutcome::Rejected)
-            }
-            Err(err) => Err(err.into()),
+            Err(error) => match classify_worker_failure(block, error)? {
+                WorkerFailureOutcome::Deferred(reason) => {
+                    Ok(BlockExecutionOutcome::Deferred(reason))
+                }
+                WorkerFailureOutcome::Rejected(_) => Ok(BlockExecutionOutcome::Rejected),
+            },
+        }
+    }
+
+    async fn validate_block_inputs(
+        &self,
+        block: OLBlockCommitment,
+    ) -> anyhow::Result<BlockValidationOutcome> {
+        match self.chain_worker.validate_block_inputs(block).await {
+            Ok(()) => Ok(BlockValidationOutcome::Authenticated),
+            Err(error) => match classify_worker_failure(block, error)? {
+                WorkerFailureOutcome::Deferred(reason) => {
+                    Ok(BlockValidationOutcome::Deferred(reason))
+                }
+                WorkerFailureOutcome::Rejected(error) => {
+                    Ok(BlockValidationOutcome::Rejected(error))
+                }
+            },
         }
     }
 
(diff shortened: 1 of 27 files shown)


bin/strata/src/fcm/tests.rs
delbonis, line 1: Weird formatting in this file.

crates/chain-worker/src/provenance.rs
delbonis, line 1: Weird formatting here too.

crates/consensus-logic/src/fcm/service.rs
```diff
@@ -81,15 +82,20 @@ pub async fn start_fcm_service<C: FcmContext>(
 pub(crate) struct FcmService<C: FcmContext>(PhantomData<C>);
 
 #[derive(Clone, Debug, Serialize)]
-pub struct FcmStatus;
+pub struct FcmStatus {
+    /// Number of blocks currently cached for dependency retries.
+    pub pending_blocks: usize,
```
delbonis: Make this not pub.

crates/consensus-logic/src/fcm/state.rs
```diff
@@ -31,6 +34,7 @@ pub(crate) struct FcmServiceState<C: FcmContext> {
     ctx: Arc<C>,
     sequencer_predicate: PredicateKey,
     inner_state: FcmInnerState,
+    pub(super) pending: PendingBlocks,
```
delbonis: Don't expose this, add wrapper fns for dealing with it.

crates/proof-impl/checkpoint/src/statements.rs
delbonis, line 1: Weird formatting here too.

delbonis (changes requested):
There's a lot of really weird formatting in this PR, like all repeated newlines were coalesced into a single newline.  It's not really actionable but it makes it a bit harder to read.

crates/consensus-logic/src/fcm/service.rs
storopoli: Made `pending_blocks` private and added a `pending_blocks()` accessor.

bin/strata/src/fcm/tests.rs
storopoli: Fixed. I restored the missing blank lines between the test fixtures, impl blocks, helper methods, setup steps, and assertions.

crates/proof-impl/checkpoint/src/statements.rs
storopoli: Fixed. I added spacing between setup, block construction, execution, and assertions in the checkpoint test.

crates/consensus-logic/src/fcm/state.rs
storopoli: Made `pending` private and routed scheduler operations through focused methods on `FcmServiceState`.

crates/chain-worker/src/provenance.rs
storopoli: Fixed. I separated the manifest validation phases with blank lines.

crates/chain-worker/src/state.rs
```diff
@@ -324,6 +340,25 @@ fn fetch_block_with_parent(
     Ok((block, parent_header, parent_commitment))
 }
 
+/// Authenticates all block inputs without executing or persisting state.
+fn validate_block_inputs(
+    ctx: &impl ChainWorkerContext,
+    block: &OLBlockV1,
+    parent_header: Option<&OLBlockHeaderV1>,
+    parent_commitment: OLBlockCommitment,
+) -> WorkerResult<MemoryStateBaseLayer> {
```
delbonis: Kinda weird that this fn both validates the "block inputs" and *also* loads the state.  Surely we can just pass in the state as an argument (as a trait) so that we can access `last_l1_height` from that.  A fn named `validate_` we'd expect to safely "have no side effects" or fail.

crates/chain-worker/src/traits.rs
```diff
@@ -17,6 +18,15 @@ use crate::{OLBlockExecutionOutput, WorkerResult};
...
 /// as the worker runs on a dedicated thread pool.
 pub trait ChainWorkerContext: Send + Sync + 'static {
+    /// Returns the configured depth beyond which L1 reorgs are unsupported.
+    fn l1_reorg_safe_depth(&self) -> u32;
+
+    /// Reads the canonical Bitcoin tip, which may lead ASM materialization.
+    fn canonical_l1_tip_height(&self) -> DbResult<Option<u32>>;
+
+    /// Resolves the uncached canonical ID then reads its ASM-generated manifest.
+    fn canonical_manifest(&self, height: u32) -> DbResult<Option<AsmManifest>>;
```
delbonis: Should we use `L1Height` for these?

crates/consensus-logic/src/fcm/context.rs
```diff
@@ -12,10 +12,46 @@ use crate::{
...
+
+/// Distinguishes authenticated stored blocks from retryable and rejected inputs.
+#[derive(Debug)]
+pub enum BlockValidationOutcome {
+    /// The stored block inputs match currently available canonical data.
+    Authenticated,
+    /// Canonical data or another local dependency is not ready yet.
+    Deferred(ExecutionDeferral),
+    /// Available canonical data proves that the stored block inputs are invalid.
+    Rejected(anyhow::Error),
```
delbonis: Don't really like this being an anyhow error.  "Rejected" implies that there's a validation error that we should be more closely tracking.

crates/consensus-logic/src/fcm/context.rs
```diff
@@ -29,6 +65,13 @@ pub trait CsmStatusReader: Send + Sync {
 /// Storage operations required by FCM.
 #[async_trait]
 pub trait FcmStorage: UnfinalizedOLBlockSource {
+    /// Reads a bounded status page to refill the in-memory retry queue.
+    async fn scan_block_statuses(
+        &self,
+        after: Option<OLBlockId>,
+        limit: usize,
+    ) -> DbResult<Vec<(OLBlockId, BlockStatus)>>;
```
delbonis: `OLBlockCommitment` maybe for the selector?  Maybe it doesn't matter.

crates/consensus-logic/src/fcm/input.rs
delbonis, line 1: Can we use a composition of the generic service input types?  I *think* we should be able to construct the same semantics from a few of those.

crates/consensus-logic/src/fcm/pending.rs
```diff
@@ -0,0 +1,248 @@
...
+    last_warned: Instant,
+}
+
+/// Holds only a bounded cache; status scans recover evicted entries without a restart.
+#[derive(Debug, Default)]
+pub(super) struct PendingBlocks {
+    entries: BTreeMap<PendingKey, PendingEntry>,
+    pub(super) scan_cursor: Option<OLBlockId>,
+}
+
```
delbonis, line 36: Shouldn't this be `PendingBlockCache` then?  And maybe have it do some of its own validation behavior so that `scan_cursor` isn't exposed.

There's functions in this impl block that do a lot of complicated things and their names are not very helpful in understanding what they do.  Should have more docs describing what's going on here.

crates/consensus-logic/src/fcm/state.rs
```diff
@@ -31,9 +36,49 @@ pub(crate) struct FcmServiceState<C: FcmContext> {
...
+
+    pub(super) fn set_pending_scan_cursor(&mut self, cursor: Option<OLBlockId>) {
+        self.pending.scan_cursor = cursor;
+    }
+
+    pub(super) fn discover_pending_block(&mut self, block: &OLBlockV1) {
+        self.pending.discover(block);
+    }
+
+    pub(super) fn due_pending_blocks(
```
delbonis: `get_`?  Name implies it's a simple accessor.

crates/consensus-logic/src/fcm/state.rs
```diff
@@ -31,9 +36,49 @@ pub(crate) struct FcmServiceState<C: FcmContext> {
...
+        &mut self,
+        progress: bool,
+        limit: usize,
+    ) -> Vec<(Slot, OLBlockId)> {
```
delbonis: Should totally use `OLBlockCommitment` here! (and in the underlying fn)

delbonis (changes requested):
There's a lot going on here, most of it looks good but there's a few subtleties I had comments on.

crates/chain-worker/src/state.rs
storopoli: Moved parent-state loading into `fetch_parent_state`. Both callers pass the loaded state to `validate_block_inputs` through `IStateAccessor`; validation returns `WorkerResult<()>` and does no state writes.

crates/chain-worker/src/traits.rs
storopoli: Changed the canonical tip and manifest lookup heights to `L1Height`, including the implementations and manifest authentication error. The reorg depth stays `u32` because it counts blocks.

crates/consensus-logic/src/fcm/context.rs
storopoli: `Rejected` now carries `ExecError`. The worker adapter preserves the STF error variant and its fields, so callers can inspect the validation failure. Operational failures still propagate through the outer `Result`.

crates/consensus-logic/src/fcm/context.rs
storopoli: Changed this to `OLBlockCommitment` in `3a2bf315`. To fix the security review's historical-read finding, the scan now uses the slot index and skips finalized slots before loading bodies. The cursor follows `(slot, block ID)` order; the status table itself is still keyed by block ID.

crates/consensus-logic/src/fcm/input.rs
storopoli: Replaced the custom select loop with `TickingInput<SelectInput<TokioMpscInput<_>, TokioWatchInput<_>>>`. `FcmInput` just maps their output to FCM events. It still skips missed ticks and aborts when either channel closes; I added tests for event delivery and both closure paths.

crates/consensus-logic/src/fcm/pending.rs
storopoli: Renamed it to `PendingBlockCache` and made the scan cursor private. The cache advances the cursor from completed status pages and wraps it on an empty page. I also renamed the retry-selection and insertion methods and documented eviction, parent ordering, fairness, and backoff.

crates/consensus-logic/src/fcm/state.rs
storopoli: Renamed this to `select_pending_retry_batch` and the cache method to `select_retry_batch`. The docs now say that selection updates each block's last-attempt time.

crates/consensus-logic/src/fcm/state.rs
storopoli: Changed this and the underlying cache method to take `OLBlockCommitment`. Cache keys, retry batches, removal, and the retry function now use the same type.

purusang (approved):

bin/strata/src/fcm.rs
```diff
@@ -36,6 +38,37 @@ struct StrataFcmContext {
...
+) -> anyhow::Result<WorkerFailureOutcome> {
+    match error {
+        WorkerError::MissingPreState(_) | WorkerError::MissingOLBlock(_) => Ok(
+            WorkerFailureOutcome::Deferred(ExecutionDeferral::Dependency),
+        ),
+        WorkerError::ManifestPending { height, reason } => {
+            warn!(%block, height, ?reason, "deferring unauthenticated ASM manifest");
+            Ok(WorkerFailureOutcome::Deferred(
+                ExecutionDeferral::Dependency,
+            ))
```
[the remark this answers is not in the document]
storopoli: Confirmed mismatches now return `ExecError::AsmManifestContentMismatch` in #2299 and are rejected through the cleanup path. Only missing or unburied provenance defers. The persistence regressions in this PR verify that a signed mismatch releases the watermark and permits a replacement at the same slot.

bin/strata/src/fcm/tests.rs
```diff
@@ -0,0 +1,973 @@
...
+        WorkerError::Database(DbError::Other("injected database failure".into())),
+        WorkerError::ApplyWriteBatch {
+            commitment: block,
+            source: StateError::InsufficientState,
+        },
+        WorkerError::WorkerExited,
+    ] {
+        assert!(matches!(
+            classify_worker_failure(block, error).unwrap(),
+            WorkerFailureOutcome::Deferred(ExecutionDeferral::Storage)
```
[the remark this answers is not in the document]
storopoli: Fixed in 555527e1. This test now checks only recoverable database failures. ApplyWriteBatch and WorkerExited remain fatal, with separate assertions for propagation in worker_failure_tests.

prajwolrg (approved):
I haven't thoroughly looked through the PR stack but the tests looks good here. Unblocked from my side, but someone should own looking at everything thoroughly.

PR #2281, opened 2026-09-10
refactor(ol): centralize epoch DA computation

purusang:
## Description

Centralizes canonical epoch DA computation behind `compute_epoch_da`. Checkpoint construction and prover-input assembly now share the same pre-drain replay and DA finalization path instead of reproducing it independently.

Also aligns the local rustdocs recipe with CI by enabling the coherence-only next-generation trait solver.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

`compute_epoch_da` intentionally uses pre-drain execution: epoch DA excludes ASM-manifest-derived effects because reconstruction obtains and applies those manifests separately from L1.

Validation:

- `just pr-lite`
- `test_epoch_progression`
- `test_checkpoint_sync_node`

This PR was developed and its body was written with Codex AI assistance.

- [ ] Yes
- [x] No

## Checklist

## Related Issues

None.


storopoli (approved):
ACK 6f70397f63e008e052f36cd3417cf2d530a296b8

crates/ol/checkpoint/src/epoch_da.rs
```diff
@@ -0,0 +1,150 @@
...
+    #[error("no DA blob produced after epoch replay")]
+    MissingBlob,
+}
+
+/// Epoch DA bytes and OL logs computed from an epoch's blocks.
+#[derive(Debug)]
+pub struct EpochDaOutput {
+    da_bytes: Vec<u8>,
+    ol_logs: Vec<OLLog>,
+}
```
delbonis: This feels like it's kinda mixing levels of abstraction (one field is encoded, the other is not).  I'm not sure if it makes sense to change it or not.

It'd make sense if this was clearly defined to be a version-agnostic container where the DA format is contextually-defined.

crates/ol/checkpoint/src/epoch_da.rs
```diff
@@ -0,0 +1,150 @@
...
+        previous_terminal,
+        runtime_params,
+    )
+    .map_err(EpochDaError::BlockReplay)?;
+
+    let da_bytes = da_state
+        .take_completed_epoch_da_blob()
+        .map_err(EpochDaError::Accumulation)?
+        .ok_or(EpochDaError::MissingBlob)?;
+
```
delbonis: Honestly this should be a panic.  The error case has no way to gracefully recover from it.  This should only occur if there was no terminal provided, which should be caught by earlier calls.

In the future `DaAccumulatingState` is going to be rewritten so the interface makes more sense.

crates/ol/checkpoint/src/epoch_da.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    let mut da_state = DaAccumulatingState::new(MemoryStateBaseLayer::new(pre_epoch_state));
+    let ol_logs = execute_block_batch_predrain(
+        &mut da_state,
+        epoch_blocks,
+        previous_terminal,
+        runtime_params,
+    )
+    .map_err(EpochDaError::BlockReplay)?;
```
delbonis, line 98: This ignores any logs that would be emitted during the checkin phase.  For now, there aren't any, but this should be clearly flagged so we can rework things when we do.

crates/ol/checkpoint/src/epoch_da.rs
```diff
@@ -0,0 +1,150 @@
...
+/// ASM manifests are already available from L1, so epoch DA does not duplicate
+/// the state effects reconstructed from them. The computation therefore
+/// replays the epoch's blocks without applying the terminal ASM-log drain.
+///
+/// # Errors
+///
+/// Returns an error if the blocks do not form a complete epoch, block replay
+/// fails, or the accumulated DA diff cannot be encoded.
+pub fn compute_epoch_da(
+    pre_epoch_state: OLStateV1,
```
delbonis: This MUST accept a `impl IStateAccessorMut`, not a full state.  We won't be able to just pass in a whole state in the future because it won't necessarily fit comfortably in memory.

delbonis (changes requested):

storopoli (approved):
ACK 55da1fc0ad909af38fff61f9173b5b265d345f21

crates/ol/checkpoint/src/epoch_da.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// The supplied epoch contains no blocks.
+    #[error("cannot compute epoch DA without blocks")]
+    NoBlocks,
+
+    /// The final supplied block is not an epoch terminal.
+    #[error("cannot compute epoch DA: final block is not terminal")]
+    FinalBlockNotTerminal,
+
+    /// A terminal block appears before the end of the supplied epoch.
+    #[error("cannot compute epoch DA: terminal block appears before the end")]
```
krsnapaudel, line 28: "cannot compute epoch DA; terminal block appears before the end"

NOTE the semi-colon. ":" is used for another error source

crates/ol/checkpoint/src/epoch_da.rs
```diff
@@ -0,0 +1,158 @@
...
+
+/// Errors produced while computing epoch DA.
+#[derive(Debug, Error)]
+pub enum EpochDaError {
+    /// The supplied epoch contains no blocks.
+    #[error("cannot compute epoch DA without blocks")]
+    NoBlocks,
+
+    /// The final supplied block is not an epoch terminal.
+    #[error("cannot compute epoch DA: final block is not terminal")]
```
krsnapaudel, line 24: "cannot compute epoch DA; final block is not terminal"

crates/ol/checkpoint/src/epoch_da.rs
```diff
@@ -0,0 +1,158 @@
...
+    #[error("DA accumulation failed: {0}")]
+    Accumulation(#[source] DaAccumulationError),
+}
+
+/// Version-agnostic artifacts computed by replaying an epoch's blocks.
+///
+/// The DA encoding is intentionally opaque. Consumers interpret `encoded_da`
+/// using the STF and DA scheme selected for the epoch's protocol version.
+#[derive(Debug)]
+pub struct EpochReplayArtifacts<Log> {
```
krsnapaudel, line 45: Is the generic necessary?

krsnapaudel (approved):
Some nits and a question. Not blocking.

crates/ol/checkpoint/src/epoch_da.rs
purusang: It is needed to make it version agnostic.

PR #2284, opened 2026-09-11
perf(storage): batch sentinel MMR prefill at startup

storopoli:
## Description

On a fresh datadir the node seeds two height-indexed MMR indexes, `MmrId::Asm` and `MmrId::L1BlockRefs`, with one sentinel leaf per L1 height up to genesis. Both prefills appended one leaf at a time, and each append costs a leaf-count read plus two sled transactions. With testnet-prod's genesis at L1 height 307,049 that is about 614k transactions per index and roughly ten minutes of silent startup each, twenty minutes in total (STR-3703). Mainnet-prod pays about 31 minutes per index (STR-4326). The first gap sits inside `spawn_asm_worker`, before the ASM worker task even launches, and the second inside `init_ol_genesis` between building and persisting the genesis block.

This PR adds `MmrIndexHandle::prefill_repeated_leaves(_blocking)`. It computes the all-sentinel MMR, which has one distinct hash per level, and writes it in 8192-leaf chunks through the existing `MmrBatchWrite` and `apply_update` path. Each chunk is a single transaction guarded by a leaf-count CAS and an empty-slot precondition, so a crash between chunks leaves a valid prefix and the next start resumes from it. Before writing anything, the stored peaks are checked against the expected repeated prefix. A mismatch fails with the new `DbError::MmrPrefillPrefixMismatch` and writes nothing.

Both call sites now go through the helper: `prefill_asm_mmr` in `strata-consensus-logic` and `prefill_l1_block_refs_mmr_blocking` in `strata-chain-worker`. Public signatures, the on-disk schema, hashing and proof generation are unchanged. The ASM worker's own `prefill_manifest_mmr` trait default keeps running afterwards and stays a no-op.

Release-mode benchmark on an on-disk sled, 307,050 leaves, on a dev VM:

| Path | Time |
|---|---|
| per-leaf appends (current `main`) | 189.5 s |
| batched prefill (this PR) | 6.9 s |

Scaling by the ratio between the per-leaf time observed on testnet-prod and on this VM, both prefills together should take under a minute on testnet-prod and a couple of minutes on mainnet-prod.

Chunk size does not change the speed, but it decides how much memory a single sled transaction holds. Peak RSS of the release test binary on an on-disk sled, measured with `/usr/bin/time -v`:

| Leaves | Chunk | Time | Peak RSS |
|---|---|---|---|
| 307,050 | 8192 | 8.0 s | 229 MB |
| 307,050 | 131,072 | 8.0 s | 1.37 GB |
| 307,050 | all at once | 6.1 s | 2.87 GB |
| 1,000,000 | 8192 | 28.8 s | 500 MB |
| 1,000,000 | 65,536 | 27.8 s | 1.10 GB |
| 1,000,000 | all at once | killed | OOM above 6 GB |

That is roughly 9 KB of transaction memory per leaf, so writing the whole prefix in one transaction is not an option at mainnet size and the 8192 chunk stays.

The PR also rewrites a stale comment in `strata-ol-genesis` that claimed the ASM MMR starts at `genesis_l1_height + 1` with the first post-genesis manifest at leaf 0.

### Type of Change

- [x] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The correctness argument rests on `test_prefill_repeated_leaves_matches_per_leaf_appends`. For 17 sizes that straddle chunk boundaries it compares the batched index against per-leaf appends node by node, then compares the MMR state and inclusion proofs, then appends a real leaf on both sides and compares again. Separate tests cover resuming from a partial prefix (including mid-chunk and exactly on a boundary), no-op when the target is already met, and the two failure cases where the existing prefix is not all-sentinel or a peak is missing. Both failure tests assert that nothing was written.

The helper relies on the invariant that at height `h` exactly `floor(count / 2^h)` nodes are complete, so a chunk advancing the count from `a` to `b` writes indices `[a >> h, b >> h)` at each height. That is the part worth checking by hand.

The trick: every one of those placeholder leaves is identical, so the tree built on top of them is completely predictable. Each row of the tree has a single repeated hash, and we can work out ahead of time exactly which slots a batch of new leaves fills in. So instead of appending one leaf at a time and letting the database recompute parents, the helper computes the answer up front and writes it in big chunks. Tests confirm the result is byte-for-byte the same tree the old one-at-a-time path produced, including across chunk boundaries, after a restart mid-way, and when the existing data doesn't match what's expected.

We can upstream the helper to `strata-common`.

Two `#[ignore]`d benchmarks are kept under `bench_prefill_*` for future tuning. They add `sled`, `typed-sled` and `tempfile` as dev-dependencies of `strata-storage` so they can run against an on-disk database. Run them with `cargo nextest run --release -p strata-storage bench_prefill --run-ignored ignored-only --no-capture`. `bench_prefill_chunk_size` reads `PREFILL_LEAVES` and `PREFILL_CHUNK` so one process runs one configuration; wrapping the test binary in `/usr/bin/time -v` reproduces the memory table above.

This change does not fix the other defect in STR-4326, where the CSM can persist client state before the genesis block is written. It shrinks that window from minutes to seconds. Closing it needs a startup ordering change and should be its own PR.

- [ ]  Yes
- [x]  No

## Checklist

AI disclosure: the diagnosis, design and review were done with Claude Code, and the implementation, tests and benchmarks were generated by OpenAI Codex from a written spec. This PR body was drafted by Claude Code and I reviewed and heavily edited it. All tests, lint and the release benchmark were re-run by hand before opening.

## Related Issues

Fixes [STR-3703](https://alpenlabs.atlassian.net/browse/STR-3703) and [STR-4305](https://alpenlabs.atlassian.net/browse/STR-4305). Partially addresses [STR-4326](https://alpenlabs.atlassian.net/browse/STR-4326) (bulk MMR prefill item).

[STR-3703]: https://alpenlabs.atlassian.net/browse/STR-3703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/db/types/src/errors.rs
+++ b/crates/db/types/src/errors.rs
@@ -6,7 +6,7 @@ use thiserror::Error;
 #[cfg(feature = "proxies")]
 use tokio::task::JoinError;
 
-use crate::mmr_index::{LeafPos, NodePos};
+use crate::mmr_index::{LeafPos, NodePos, RawMmrId};
 
 #[derive(Clone, Debug, Error)]
 pub enum DbError {
@@ -101,6 +101,10 @@ pub enum DbError {
     /// MMR node not found at the given tree position.
     #[error("MMR node not found at position {0:?}")]
     MmrNodeNotFound(NodePos),
+
+    /// Existing peaks do not commit to the requested repeated sentinel prefix.
+    #[error("MMR {mmr_id:?} does not have the repeated prefill prefix at {leaf_count} leaves")]
+    MmrPrefillPrefixMismatch { mmr_id: RawMmrId, leaf_count: u64 },
 
     /// MMR index batch precondition failed.
     #[error("MMR precondition failed for {mmr_id:?}: {detail}")]
--- a/crates/storage/src/managers/mmr_index.rs
+++ b/crates/storage/src/managers/mmr_index.rs
@@ -20,6 +20,9 @@ use tokio::task::spawn_blocking;
 
 use crate::ops::mmr_index::MmrIndexOps;
 
+/// Bounds transaction memory while reducing the 307,050-leaf genesis prefill to 38 writes.
+const REPEATED_LEAF_CHUNK_SIZE: u64 = 8192;
+
 /// Retry behavior for optimistic CAS-style MMR updates.
 #[derive(Debug, Clone, Copy)]
 pub struct MmrIndexRetryConfig {
@@ -199,6 +202,102 @@ impl MmrIndexHandle {
         run_with_precondition_retries(self.max_retries, || {
             self.append_leaf_once_blocking(hash, None)
         })
+    }
+
+    /// Extends a height-indexed MMR with a repeated sentinel prefix.
+    ///
+    /// Returns without writing when the count already reaches `target_leaf_count`.
+    /// Otherwise, verifies the existing peaks commit to repeated `leaf_hash` leaves
+    /// before extending the prefix. Each chunk commits a complete prefix, allowing
+    /// startup to resume after a crash. Batching reduces STR-3703's roughly 614,000
+    /// per-leaf writes across the two genesis MMRs to tens of transactions.
+    ///
+    /// Requires a single writer during startup. Count and empty-slot preconditions
+    /// reject conflicting writes without retrying. Returns
+    /// [`DbError::MmrPrefillPrefixMismatch`] without writing if a peak is missing
+    /// or differs from the repeated prefix.
+    pub fn prefill_repeated_leaves_blocking(
+        &self,
+        leaf_hash: Hash,
+        target_leaf_count: u64,
+    ) -> DbResult<()> {
+        self.prefill_repeated_leaves_with_chunk_blocking(
+            leaf_hash,
+            target_leaf_count,
+            REPEATED_LEAF_CHUNK_SIZE,
+        )
+    }
+
+    /// Allows benchmarks to vary transaction size without changing the public API.
+    fn prefill_repeated_leaves_with_chunk_blocking(
+        &self,
+        leaf_hash: Hash,
+        target_leaf_count: u64,
+        chunk_size: u64,
+    ) -> DbResult<()> {
+        assert!(chunk_size > 0, "prefill chunk size must be positive");
+        let mut current = self.get_leaf_count_blocking()?;
+        if current >= target_leaf_count {
+            return Ok(());
+        }
+
+        // This is the same hash ladder as Mmr::new_repeated and write_plan's
+        // completed subtrees, including the node hasher's domain separation.
+        // Retain zero hashes too: they are valid stored nodes, not absent peaks.
+        let mut level_hashes = [leaf_hash.0; 64];
+        for height in 1..=target_leaf_count.ilog2() as usize {
+            let child = level_hashes[height - 1];
+            level_hashes[height] = Sha256Hasher::hash_node(child, child);
+        }
+        for peak in peak_positions(current) {
+            let expected = Hash::from(level_hashes[usize::from(peak.height())]);
+            if self.get_node_blocking(peak)? != Some(expected) {
+                return Err(DbError::MmrPrefillPrefixMismatch {
+                    mmr_id: self.mmr_id_bytes(),
+                    leaf_count: current,
+                });
+            }
+        }
+
+        while current < target_leaf_count {
+            let next = current + (target_leaf_count - current).min(chunk_size);
+            let mut batch = MmrBatchWrite::default();
+            let mmr_batch = batch.entry(self.mmr_id_bytes());
+            mmr_batch.set_expected_leaf_count(current);
+            mmr_batch.add_node_precond(LeafPos::new(current).to_node_pos(), None);
+
+            // At height h, exactly floor(count / 2^h) nodes are complete.
+            for height in 0..=next.ilog2() as u8 {
+                let start = current >> height;
+                let end = next >> height;
+                if start == end {
+                    break;
+                }
+                let hash = Hash::from(level_hashes[usize::from(height)]);
+                for index in start..end {
+                    mmr_batch.put_node(NodePos::new(height, index), hash);
+                }
+            }
+            mmr_batch.set_leaf_count(next);
+            self.ops.apply_update_blocking(batch)?;
+            current = next;
+        }
+        Ok(())
+    }
+
+    /// Extends the sentinel prefix on a blocking task.
+    ///
+    /// Applies the invariant and single-writer contract of
+    /// [`Self::prefill_repeated_leaves_blocking`].
+    pub async fn prefill_repeated_leaves(
+        &self,
+        leaf_hash: Hash,
+        target_leaf_count: u64,
+    ) -> DbResult<()> {
+        let this = self.clone();
+        spawn_blocking(move || this.prefill_repeated_leaves_blocking(leaf_hash, target_leaf_count))
+            .await
+            .map_err(DbError::from)?
     }
 
     /// Appends a caller-provided leaf hash and stores its preimage bytes.
@@ -805,8 +904,15 @@ where
 
 #[cfg(test)]
 mod tests {
+    use std::time::{Duration, Instant};
+
+    use sled::Config;
     use strata_db_store_sled::test_utils::get_test_sled_backend;
+    use strata_db_store_sled::{SledBackend, SledDbConfig};
     use strata_merkle::{Mmr, Mmr64B32, MmrState};
+    use strata_ol_state_types_v1::MMR_SENTINEL_DUMMY_LEAF_HASH;
+    use tempfile::tempdir;
+    use typed_sled::SledDb;
 
     use super::*;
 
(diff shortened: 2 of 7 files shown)


crates/storage/src/managers/mmr_index.rs
```diff
@@ -201,6 +204,102 @@ impl MmrIndexHandle {
...
+        &self,
+        leaf_hash: Hash,
+        target_leaf_count: u64,
+        chunk_size: u64,
+    ) -> DbResult<()> {
+        assert!(chunk_size > 0, "prefill chunk size must be positive");
+        let mut current = self.get_leaf_count_blocking()?;
+        if current >= target_leaf_count {
+            return Ok(());
+        }
```
delbonis, line 251: Should fail if current is nonzero.  It is invalid to attempt to prefill if there are ANY leafs written, as the upper branches of the MMR will be inconsistent.

crates/storage/src/managers/mmr_index.rs
```diff
@@ -20,6 +20,9 @@ use tokio::task::spawn_blocking;
 
 use crate::ops::mmr_index::MmrIndexOps;
 
+/// Bounds transaction memory while reducing the 307,050-leaf genesis prefill to 38 writes.
+const REPEATED_LEAF_CHUNK_SIZE: u64 = 8192;
```
delbonis: Can this go any higher safely?

delbonis (changes requested):

crates/storage/src/managers/mmr_index.rs
storopoli: I tried this first: one transaction for the whole prefix and a hard error on any nonzero count. It doesn't fit in memory. sled holds about 9 KB per leaf inside a transaction, so 307,050 leaves peaked at 2.9 GB and 1,000,000 leaves was OOM-killed above 6 GB on an 8 GB VM. With 8192-leaf chunks the same runs stay at 229 MB and 500 MB, at the same speed. The table is in the PR body.

So the write has to be chunked, and a crash between chunks leaves a valid partial prefix on disk. Failing on it would leave a fresh datadir unable to start until someone wipes the DB. The inconsistency you're describing is what the check just below this line guards against. Before writing anything, the helper reads every peak of the current count and compares it against the expected all-sentinel hash for that height. The peaks commit to every leaf below them, so if they match, the prefix really is all sentinels and the new nodes join it consistently. Anything else, including a missing peak or a real manifest already in the tree, returns `MmrPrefillPrefixMismatch` without writing. `test_prefill_repeated_leaves_rejects_prefix_mismatch_without_writes` and `..._rejects_missing_peak_without_writes` cover both cases.

TLDR; I'd rather keep resume instead of failing

crates/storage/src/managers/mmr_index.rs
storopoli: No, and I have numbers now. Speed is flat across chunk sizes, memory is roughly linear at 9 KB per leaf per transaction: 8,192 gives 229 MB at testnet size and 500 MB at a million leaves, 131,072 gives 1.37 GB at testnet size, and a single transaction is 2.87 GB at testnet size and gets OOM-killed above 6 GB at a million. The 67 MB estimate on STR-4305 is the raw hashes; sled's transaction bookkeeping is roughly 100x that.

I put this on the constant's doc comment and made `bench_prefill_chunk_size` read `PREFILL_LEAVES` and `PREFILL_CHUNK`, so wrapping the test binary in `time -v` reproduces the table. If anything we could go lower, but 8,192 keeps testnet at 38 transactions and comfortably under 250 MB.

crates/storage/src/managers/mmr_index.rs
```diff
@@ -820,6 +936,246 @@ mod tests {
...
+                .unwrap();
+            assert_eq!(handle.get_leaf_count_blocking().unwrap(), count);
+            assert_eq!(snapshot_nodes(&handle, count), before);
+        }
+        // A completed prefill remains a no-op after real manifests arrive.
+        handle.append_leaf_blocking(Hash::from([0x42; 32])).unwrap();
+        handle
+            .prefill_repeated_leaves_blocking(MMR_SENTINEL_DUMMY_LEAF_HASH, count)
+            .unwrap();
+        assert_eq!(handle.get_leaf_count_blocking().unwrap(), count + 1);
```
krsnapaudel, line 1040: This checks count only not a no op. Needs something like this:
`assert_eq!(snapshot_nodes(&handle, count), before);`

crates/storage/src/managers/mmr_index.rs
```diff
@@ -820,6 +936,246 @@ mod tests {
...
+            .entry(MmrId::Asm.to_bytes())
+            .del_node(NodePos::new(1, 0));
+        manager.apply_update_blocking(batch).unwrap();
+        assert_prefill_prefix_rejected(&handle);
+    }
+
+    #[test]
+    fn test_prefill_repeated_leaves_resumes_at_chunk_boundary() {
+        let expected = setup_handle();
+        let actual = setup_handle();
```
krsnapaudel, line 1082: The use of `expected` and `actual` in tests is kind of confusing since they are handles returned by the same setup and later call `append_sentinels` and `prefill_repeated_leaves_blocking`, which do the same thing in different ways. We should be able to validate the MMR with sentinels without having to build another using append leaf right?

crates/storage/src/managers/mmr_index.rs
```diff
@@ -820,6 +936,246 @@ mod tests {
...
+                let leaf = actual.get_leaf_blocking(index).unwrap().unwrap();
+                assert!(<Mmr64B32 as Mmr<Sha256Hasher>>::verify(
+                    &state, &proof, &leaf.0
+                ));
+            }
+        }
+    }
+
+    #[test]
+    fn test_prefill_repeated_leaves_matches_per_leaf_appends() {
```
krsnapaudel, line 987: Is this useful? Don't the benchmark tests below test this?

krsnapaudel (commented):
Some nits on tests.

crates/storage/src/managers/mmr_index.rs
storopoli: The final case now snapshots the nodes after appending the real leaf and compares them again after prefill. Both snapshots use `count + 1`, so they include the real leaf. The count assertion stays too.

Updated in dd096dc044083b27d87ae19a1dbe7d8d035b7f38.

crates/storage/src/managers/mmr_index.rs
storopoli: Renamed the handles to `per_leaf` and `batched` throughout the prefill tests and comparison helper. `Mmr::new_repeated` can give us the expected state, but this test also compares every stored node and sampled inclusion proofs. I kept the append-built reference to check that batching preserves the existing index layout.

Updated in dd096dc044083b27d87ae19a1dbe7d8d035b7f38.

crates/storage/src/managers/mmr_index.rs
storopoli: I kept this test. Both benchmarks are `#[ignore]`, and their assertions cover MMR state without comparing every stored node or inclusion proofs. This test runs with the regular suite, covers tree and chunk boundaries, and checks that appending a distinct leaf after prefill still works.

storopoli:
@irnb can you please review this? @delbonis is off, @krsnapaudel will be off starting tomorrow and @bewakes is also off.

crates/storage/src/managers/mmr_index.rs
```diff
@@ -820,6 +936,248 @@ mod tests {
...
+            assert_mmr_equivalent(&per_leaf, &batched, count + 1);
+        }
+    }
+
+    #[test]
+    fn test_prefill_repeated_leaves_resumes_per_leaf_prefix() {
+        let per_leaf = setup_handle();
+        append_sentinels(&per_leaf, 20000);
+        for prefix in [1, 5, 8192, 12345] {
+            let batched = setup_handle();
```
krsnapaudel: `per_leaf`?

krsnapaudel (approved):
You can see the confusion with `batched` and `per_leaf` in the remaining nit. Anyway, fix the typo and should be fine.

crates/storage/src/managers/mmr_index.rs
storopoli: Right, that handle starts with per-leaf appends and only the tail is batched, so neither name fit. Renamed the resume tests to `reference` and `resumed`, and the helper's parameters to `reference` and `candidate`. `per_leaf` and `batched` stay where they are accurate. Amended into 56b72326.

crates/db/types/src/errors.rs
```diff
@@ -102,6 +102,10 @@ pub enum DbError {
     #[error("MMR node not found at position {0:?}")]
     MmrNodeNotFound(NodePos),
 
+    /// Existing peaks do not commit to the requested repeated sentinel prefix.
+    #[error("MMR {mmr_id:?} does not have the repeated prefill prefix at {leaf_count} leaves")]
+    MmrPrefillPrefixMismatch { mmr_id: RawMmrId, leaf_count: u64 },
```
delbonis, line 109: "expected prefill prefix" or something might be more clear?

crates/storage/src/managers/mmr_index.rs
```diff
@@ -20,6 +20,16 @@ use tokio::task::spawn_blocking;
...
 
+/// Leaves written per prefill transaction.
+///
+/// Chunk size does not affect prefill speed, because sled's per-key work dominates,
+/// but a transaction holds roughly 9 KB per leaf in memory until it commits. On an
+/// on-disk sled, 8192-leaf chunks peak near 230 MB for 307,050 leaves and 500 MB
+/// for 1,000,000 leaves, whereas a single transaction for 307,050 leaves peaks near
+/// 2.9 GB and one for 1,000,000 leaves exceeds 6 GB. Chunking also lets a crashed
+/// prefill resume from the last committed prefix.
+const REPEATED_LEAF_CHUNK_SIZE: u64 = 8192;
```
delbonis, line 31: 9K per leaf is insane overhead wtf

I did not expect that

crates/storage/src/managers/mmr_index.rs
```diff
@@ -201,6 +211,104 @@ impl MmrIndexHandle {
...
+        )
+    }
+
+    /// Allows benchmarks to vary transaction size without changing the public API.
+    fn prefill_repeated_leaves_with_chunk_blocking(
+        &self,
+        leaf_hash: Hash,
+        target_leaf_count: u64,
+        chunk_size: u64,
+    ) -> DbResult<()> {
```
delbonis, line 246: Some parts of this fn could maybe be factored out.

delbonis (changes requested):
Looks good.

crates/db/types/src/errors.rs
storopoli: Reworded to "MMR {mmr_id} peaks at {leaf_count} leaves do not match the expected prefill prefix". Folded into 3344678b.

crates/storage/src/managers/mmr_index.rs
storopoli: That number is measured, from the RSS sweep in the PR body, so I have no model for where it goes. Each leaf is roughly two node puts, so it is about 4.5 KB per put held in sled's transaction buffers until commit. Chunking keeps it flat, so I did not chase it further.

crates/storage/src/managers/mmr_index.rs
storopoli: Split it into three helpers: `repeated_leaf_ladder` builds the hash ladder, `check_repeated_prefix_blocking` checks the existing peaks against it, and `repeated_prefix_chunk` builds one chunk's batch. The prefill method is now the early return, the check, and the loop. Same writes in the same order, and the six prefill tests pass. Folded into 3344678b.

storopoli:
@delbonis can we merge this now?

delbonis (approved):

PR #2290, opened 2026-09-17
fix(build): [0/4] match CI rustdoc solver settings

purusang:
## Description

Local `just pr-lite` fails during rustdoc generation on the pinned SSZ dependency because the local recipe omits the solver flag already used in CI. Add `-Znext-solver=coherence` to the rustdoc recipe so local documentation checks use the same setting.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Stack position **00/04**; PR 00 is the build prerequisite. Base branch: `main`.

| Index | Draft PR | Base branch |
|---|---|---|
| 00 | [Rustdoc prerequisite](https://github.com/alpenlabs/alpen/pull/2290) | `main` |
| 01 | [Admission log budgets](https://github.com/alpenlabs/alpen/pull/2291) | `fix/str-4402-00-rustdoc-solver` |
| 02 | [Assembly deferral](https://github.com/alpenlabs/alpen/pull/2292) | `fix/str-4402-01-admission` |
| 03 | [Mempool snapshots](https://github.com/alpenlabs/alpen/pull/2293) | `fix/str-4402-02-assembly-deferral` |
| 04 | [Protocol log capacities](https://github.com/alpenlabs/alpen/pull/2294) | `fix/str-4402-03-mempool-snapshots` |

- This is the one-line build prerequisite for the STR-4402 draft stack. It changes no runtime behavior.

Validation:

- `just rustdocs` on this branch. No runtime tests were added for this configuration-only change.

AI assistance: Codex generated the recipe change and this PR description; the author reviewed and directed the changes.

Is this PR addressing any specification, design doc or external reference document?

- [x] Yes
- [ ] No

Reference: [STR-4402](https://alpenlabs.atlassian.net/browse/STR-4402).

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [x] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added (where necessary) tests that prove my changes are effective or that my feature works.
- [ ] New and existing tests pass with my changes.
[...]

PR #2291, opened 2026-09-17
fix(ol): [1/4] reject oversized updates at mempool admission

purusang:
## Description

A snark-account update can fit its transaction schema while emitting more logs than an empty checkpoint can accommodate. Reject those updates at mempool admission and return structured RPC errors identifying the exceeded log resource, actual usage, and limit.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Stack position **01/04**; PR 00 is the build prerequisite. Base branch: `fix/str-4402-00-rustdoc-solver`.

Stack index: [all five draft PRs](https://github.com/alpenlabs/alpen/pull/2290).

Prerequisite: https://github.com/alpenlabs/alpen/pull/2290. Merge after that PR.

- Review the first commit as the epoch-sealing refactor, then the admission-policy commit.
- The policy shares bridge-message classification with the STF and counts the update log plus valid withdrawal logs. It checks before replacement or persistence, preserving an existing usable update when a replacement is oversized.
- DA/inbox and full-envelope budgets remain outside this log admission check.

Validation:

- `just pr-lite` on this branch.
- Admission/STF log measurements, resource boundaries, typed RPC errors, and replacement preservation.

AI assistance: Codex generated code, tests, and this PR description; the author reviewed and directed the changes.

- [x] Yes
- [ ] No

Reference: [STR-4402](https://alpenlabs.atlassian.net/browse/STR-4402).

## Checklist

## Related Issues

Related to [STR-4402](https://alpenlabs.atlassian.net/browse/STR-4402).

[STR-4402]: https://alpenlabs.atlassian.net/browse/STR-4402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/ol/tx-policy/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[dependencies]
+strata-acct-types.workspace = true
+strata-asm-checkpoint-types.workspace = true
+strata-bridge-params.workspace = true
+strata-codec.workspace = true
+strata-ol-chain-types-v1.workspace = true
+strata-ol-stf-v1.workspace = true
+strata-ol-tx-types-v1.workspace = true
+thiserror.workspace = true
```
storopoli, line 17: empty spaces for outside/inside deps to separate them.

```suggestion
strata-ol-tx-types-v1.workspace = true

thiserror.workspace = true
```

storopoli (approved):
ACK 931701d7332024f96bda7b1f84d84632c9b61247

The claude and codex finding on existing txs in DB is not valid IMO. (they are the same finding)

storopoli (approved):
ACK fcab591b6aa0f147ac12d6bd2cf8b80fa28e8566

crates/ol/tx-policy/Cargo.toml
storopoli: Hey @purusang can you do this minor cosmetic/organization nit?

crates/ol/block-assembly/src/checkpoint_size.rs
```diff
@@ -12,19 +12,11 @@
...
 
-/// Maximum total OL log payload size per checkpoint (16 KiB per SPS-ol-chain-structures).
-///
-/// Set well below [`MAX_CHECKPOINT_PAYLOAD_SIZE`] to reserve room for the
-/// checkpoint's other components (e.g. state diff), by bounding logs in
-/// aggregate rather than lowering `MAX_LOG_PAYLOAD_LEN` (per-log) or
-/// `MAX_OL_LOGS_PER_CHECKPOINT`. It is intentionally separate from,
-/// and inconsistent with, those two limits.
-pub(crate) const MAX_TOTAL_LOG_PAYLOAD_BYTES: usize = 16 * 1024;
-
```
krsnapaudel, line 27: This limit is for checkpoint. So probably makes sense to leave it here.

crates/ol/block-assembly/src/epoch_sealing.rs
krsnapaudel, line 83: Should we also update this to something like
```
pub struct EpochSealingLimitActions {
    checkpoint_size: EpochSealingLimitAction,
    manifest_count: EpochSealingLimitAction,
}
```
So the enum above can go and interleaving of `EpochSealingLimit` and `EpochSealingLimitAction` is not possible?

crates/ol/tx-policy/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+        .update()
+        .get_log_data()
+        .expect("SSZ update extra data fits the account-update log bound");
+    usage.add_log(&update_log)?;
+
+    for message in tx.data().effects().messages_iter() {
+        if message.dest() != BRIDGE_GATEWAY_ACCT_ID {
+            // Non-bridge messages emit no additional OL logs; the account-update
+            // log is already counted above. Inter-EE inbox writes contribute to
+            // checkpoint DA.
```
krsnapaudel: What does the last sentence mean? Is this also already counted? Does it produce a log? Seems like the sentence starts and topic and does not complete it.

crates/ol/tx-policy/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+            continue;
+        }
+        if let Ok(log) = parse_bridge_withdrawal(
+            message.payload().value().to_sat(),
+            message.payload().data(),
+            bridge_params,
```
krsnapaudel, line 135: So bridge withdrawal is not part of update logs?

crates/ol/tx-policy/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn check_limits(self) -> Result<Self, TxLogBudgetError> {
+        let count_limit =
+            (MAX_LOGS_PER_BLOCK as usize).min(MAX_OL_LOGS_PER_CHECKPOINT as usize - 1);
+        if self.count > count_limit {
+            return Err(TxLogBudgetError::LogCount {
+                actual: self.count,
+                limit: count_limit,
+            });
+        }
+        let payload_limit = MAX_TOTAL_LOG_PAYLOAD_BYTES - 1;
```
krsnapaudel, line 83: Do we need this? Why not directly compare with the constant in the next line?

crates/ol/tx-policy/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+///
+/// This threshold is exclusive: total log payload size must be strictly below it.
+pub const MAX_TOTAL_LOG_PAYLOAD_BYTES: usize = 16 * 1024;
+
+/// Failures when measuring or checking a transaction's own log budget.
+#[derive(Debug, Error)]
+pub enum TxLogBudgetError {
+    /// The emitted log count exceeds the inclusive per-update limit.
+    #[error("update emits {actual} logs, exceeding limit {limit}")]
+    LogCount { actual: usize, limit: usize },
```
krsnapaudel, line 37: Suggest adding a blank line between variants.

crates/ol/tx-policy/src/lib.rs
```diff
@@ -0,0 +1,137 @@
+//! Sequencer admission policy for transaction log budgets.
```
krsnapaudel: Is the policy for mempool admission? If so, do these changes belong inside `ol/mempool`?

krsnapaudel (changes requested):

krsnapaudel:
@purusang I actually see two other issues.
1. Should mempool do anything with checkpoint size in mind? Afaiu this is not mempool concern in the current design. If mempool and block/batch assembly were one component, then checkpoint limit checks would make sense.
2. Block assembly already checks log count and payload size per tx. It does that using epoch-cumulative count and size metrics. Even if we add a check to mempool, we need to consider how code will be reused.

crates/ol/tx-policy/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+/// account messages emit no logs and pass this policy unchanged.
+///
+/// The log count must fit [`MAX_LOGS_PER_BLOCK`] and remain strictly below
+/// [`MAX_OL_LOGS_PER_CHECKPOINT`]. Combined encoded log payload bytes must remain
+/// strictly below [`MAX_TOTAL_LOG_PAYLOAD_BYTES`].
+///
+/// # Panics
+///
+/// Panics if the SSZ extra-data bound stops fitting the account-update log bound.
+pub fn check_tx_log_budget(
```
bewakes, line 108: Minor suggestion: Can this function be reused? This checks the usage irrespective of remaining budget. By passing the remaining budget, I think this can be reused inside block assembly. For this case, we would just pass in the full capacity.

crates/ol/block-assembly/src/checkpoint_size.rs
purusang: This has been moved to [stf](https://github.com/alpenlabs/alpen/pull/2294/changes#diff-54185addbadfed6c24dddd1104b3e175dcd56e61846dcd7c140446b78536cdf5R4) by https://github.com/alpenlabs/alpen/pull/2294. We can discuss on where to move this constant in that pr. That PR makes this a consensus epoch-log limit, so the STF enforces it and admission/assembly reuse the same rule.

crates/ol/tx-policy/src/lib.rs
purusang: The distinction is:

- Account-update log: one log describing the snark account update.
- Withdrawal logs: separate logs emitted by the bridge gateway for each valid withdrawal.

Both belong to the transaction’s total OL logs. An update with three valid withdrawals therefore emits 1 + 3 = 4 logs.

crates/ol/tx-policy/src/lib.rs
purusang: >What does the last sentence mean?

It means inter-ee msgs do not produce additional logs but it contributes to checkpoint's state diff.
Yup last part of the sentence was unnecessary, removing it.

crates/ol/tx-policy/src/lib.rs
purusang: Its refactored by [2294](https://github.com/alpenlabs/alpen/pull/2294/changes#diff-44d8b441344fdd9b7ec5c9e1ae9b29aa54219cad7d9d6ee568812965c5df9e2dR99)

crates/ol/tx-policy/src/lib.rs
purusang: #2292 also uses this check in block assembly. It measures standalone transaction log usage independently of mempool state, so we keep it in a shared crate. I’ll clarify the module description.

crates/ol/tx-policy/src/lib.rs
purusang: Yeah, can do it but it has already been reusable as designed by #2294. Please check that as welll.

purusang:
>@purusang I actually see two other issues.

>Should mempool do anything with checkpoint size in mind? Afaiu this is not mempool concern in the current design. If mempool and block/batch assembly were one component, then checkpoint limit checks would make sense.
Block assembly already checks log count and payload size per tx. It does that using epoch-cumulative count and size metrics. Even if we add a check to mempool, we need to consider how code will be reused.

@krsnapaudel Admission checks whether a transaction can fit an empty epoch, independently of current epoch usage. This avoids retaining an update that can never be included and gives the sender an immediate rejection reason. Assembly still checks cumulative usage. #2292 reuses the standalone check, and #2294 centralizes the underlying log-limit rule for admission and execution.

crates/ol/tx-policy/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+    LogPayloadBytes { actual: usize, limit: usize },
+
+    /// A typed log could not be encoded.
+    #[error("cannot encode transaction log: {0}")]
+    Encoding(#[from] CodecError),
+}
+
+/// Log usage of a transaction that passes the standalone admission budget.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct TxLogUsage {
```
krsnapaudel, line 50: This looks very similar to
```
pub(crate) struct LogMetrics {
    pub count: usize,
    pub total_payload: usize,
    pub ssz_size: usize,
}
```
It would be good to avoid the duplication.

